PR #116, opened 2011-09-07
Clean project improvements

dragan:
Changed the clean logic to run async instead of sync and I added an event to notify those who care that a build target was cleaned.


xamarin-release-manager:
Fixed in version 7.6.0.816 (master)

Pull Request #4806 merged by: Kirill Osenkov
Author: mono/dev/kirillo
Commit: 925430fa1f268b10d536893ca82299103fffa29d (mono/monodevelop)

PR #126, opened 2011-11-07
Generate equality operators

paszczi:
Patch adds equality operators generation to "Equality members" refactoring action. For value types (structs) additional Equals method is generated with parameter matching the enclosing type. This method is called from within equality operators body (avoids unnecessary boxing/unboxing) as well as from the Equals (object obj) method's body. Equals (SomeType type) can also be called from user code, which also improves performance by avoiding boxing when comparing two instances of the same value type.


slluis:
This needs to be redone for Roslyn.

PR #312, opened 2013-06-22
Implement basic delete from movement actions.

rcorre:
Allows actions like dw (delete word).

In order to do this, I established a framework where a command map may
associate a "transformer" with a nested command map. The transformer is
a delegate that converts the action retrieved from the nested command map into an action with
different behavior when it is finally executed.

For example, the 'd' key (for delete) contains the nested map of
motions. Associated with this is a transformer that changes a motion
into a delete action using that motion.

This is currently just a proof of concept - there are many bugs.

For one, commands like dj and dk do not behave as they should, and may
need special handling. It might be a good idea to separate motions that
operate within one line (h,j,k,l,w,b,e, ect.) into a different command
map than motions that operate between lines (j,k).

In addition, repeated deletion commands do not undo all at once. For
example, 3dw will take 3 presses of u to undo fully.


slluis:
The vi mode is gone.

PR #319, opened 2013-07-01
Fix coding style

rcorre:
Implements various features for the "old" Vi Mode.

Based on feedback, code reformats were squashed into each commit.


slluis:
The vi mode has been removed.

PR #367, opened 2013-08-02
C# Split project

luiscubal:
Fixed typo in Makefile.am
Added missing commit (ProjectTemplate is internal again; ProjectOperations now allows creation of new projects from a template ID)


slluis:
This feature should be published as an add-in, not as part of the core MD.

PR #414, opened 2013-10-07
Fix external console bug

benpye:
This commit fixes a bug with gnome-terminal in conjunction with the external console option. Gnome have removed the --disable-factory argument and as such, gnome-terminal appears to be unusable with Monodevelop, this will make it so xterm is used on Gnome desktops which does work. Refer to https://bugzilla.xamarin.com/show_bug.cgi?id=12684 for more detail.


greenspray:
In in Fedora 23 x86_64 and this issue still exists.

benpye:
This might be fixable if new terminal servers are started, see http://unix.stackexchange.com/questions/201900/run-true-multiple-process-instances-of-gnome-terminal . I never did get around to fixing this by adding a setting for the terminal unfortunately.

PR #437, opened 2013-11-16
Fixed a bug where XSP would fail to launch when using .NET 4 configuration

ThreeToes:
On Windows. Path generator would generate the correct path, and then switch it out for the 4.5 library path instead. XSP isn't packaged into that directory on Windows, so MonoDevelop would throw an exception when it couldn't find it


xamarin-release-manager:

Author: mono

PR #485, opened 2014-02-19
SearchEntry filter menu Y offset miscalcualted

DavidKarlas:
I will let image talk(this are all usages of SearchEntry I could find (left is previous right is new):
image


xamarin-release-manager:

Author: mono

PR #487, opened 2014-02-23
Command description is now set on CommandMenuItems as ToolTip

DavidKarlas:
Main reason why I wanted to change this is because I have 2 MonoDevelop(Main) solutions under Recent solutions(my repo and upstream) both display as Main so I had to guess...
This change has effect on all items in "MenuStrip"... I'm adding SS to display changes notice that before was no tooltips on any of this.
(first 2 images left without change right after change)
image


xamarin-release-manager:

Author: mono

PR #488, opened 2014-02-25
[Core] Fix TargetFramework race

ddunkin:
If something tries to request an unknown TargetFramework during
TargetRuntime initialization, ensure all the runtimes are initialized
and try again.

We have MonoTouch and MonoAndroid projects that reference PCL projects. Our CI builds would fail about half the time because the TargetFramework for the PCL project wasn't yet loaded when the MonoTouch project was loaded, resulting in the reference failing with the error:
`WARNING: The reference 'Libronix.Utility-Portable' is not valid for the target framework of the project.`


xamarin-release-manager:

Author: mono

PR #491, opened 2014-02-28
Directory properly moved.

Hedgehike:
https://bugzilla.xamarin.com/show_bug.cgi?id=17481&GoAheadAndLogIn=1 is now resolved. When a directory gets moved in the project explorer, the original gets deleted now, or, if it contains files not in the current project, those files will be kept.


slluis:
The bug has already been fixed. Closing.

PR #496, opened 2014-03-07
Fix ParsedTemplate to enable camelCase attributes

PiotrZierhoffer:
As for now, directive.Attributes is filled with "attName.ToLower()".

"Extract" method works on unchanged values and TemplatingEngine asks
for camelCase.

VisualStudio disregards the casing of parameters, so we can compare with
lower case.

This can be verified by using "linePragmas" attribute.

According to mhutch suggestion, fixed to use
StringComparer.OrdinalIgnoreCase.


xamarin-release-manager:

Author: mono

PR #497, opened 2014-03-10
[UnitTests] Only show test projects in active config

Therzok:
Only showing test projects that are built in the active configuration.

That means, we're not going to be showing nodes for Subversion.Win32 under Mac and Subversion.Unix on Windows.


xamarin-release-manager:

Author: mono

PR #508, opened 2014-03-22
[Win32Debug] Steping over Propeties and NonUserCode

DavidKarlas:
So anyone who did some developing with XS Win32 debugger know what pain it is to step into methods sometimes you have to press F11 10+ times to step into method because stepper goes somewhere dunno where...

Solution around property stepping over came from SoftDebugger this includes bool autoStepInto and copy-pasted method IsPropertyOrOperatorMethod. Also when this is handled is same. So after we step inside property we call step out and set autoStepIn so after stepping out we step in again.

After that I went after multiple F11 needed to step into method after some debugging. I noticed two problems stepping inside method that does not exist(autoimplmented constructor for example) this is solved by:

``` csharp
if (symbols == null || symbols.SequencePointCount == 0)
```

and multiple F11 when stepping inside existing method. This one was a bit trickier because there is no documentation laying around on how to to this so I more or less brute forced this two IFs:

``` csharp
if (seqFirst.Offset > offset || (seqFirst.Offset == offset && mappingResult == CorDebugMappingResult.MAPPING_APPROXIMATE))
and
if (seqLast.Offset < offset || (seqLast.Offset == offset && mappingResult == CorDebugMappingResult.MAPPING_APPROXIMATE)) {
```

First one is to check at beginning of method and second is at end of method. I came to this based on debugging and checking of values.

If you want to test difference I suggest you to copy paste this code before this commit and after with pressing F11... Difference is huge :)

``` csharp
using System;
using System.Linq;
using System.Collections.Generic;

public class Program
{
    {
        ClosuresWithLambdas a = new ClosuresWithLambdas();

        List<Func<int>> fs = a.GetNumberFunctions();

        a.Prop = 23;
        Console.WriteLine (a.Prop);

        Func<int> increments_i_func = fs[0];
        Func<int> just_returns_i_func = fs[1];

        //prints 1
        a.PrintReturnValue(increments_i_func);
        //prints 2
        a.PrintReturnValue(just_returns_i_func);
    }

    public class ClosuresWithLambdas
    {
        int prop=0;
        public int Prop {
            get{
                return prop;
            }
            set{
                prop = value;
            }
        }

        public List<Func<int>> GetNumberFunctions()
        {
            //this variable gets included in the closure
            int j = 1;

            //we use this to return two functions
            List<Func<int>> fs = new List<Func<int>>(2);

            //don't get confused by the postfix operator
            //the f lambda returns j BEFORE incrementing it
            Func<int> f = () => j++;

            //g just returns the variable j
            Func<int> g = () => j;

            //add the two functions to the list and return it
            fs.Add(f);
            fs.Add(g);
            return fs;
        }

        public void PrintReturnValue(Func<int> f)
        {
            Console.WriteLine(f());
        }
    }
}

```

Code is copy pasted from http://iformattable.blogspot.com/2008/01/lambdas-closures-in-c-30.html with added property
Tnx again ;)

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger.Win32/MonoDevelop.Debugger.Win32/CorDebuggerSession.cs
+++ b/main/src/addins/MonoDevelop.Debugger.Win32/MonoDevelop.Debugger.Win32/CorDebuggerSession.cs
@@ -14,6 +14,7 @@ using Microsoft.Samples.Debugging.Extensions;
 using Mono.Debugging.Backend;
 using Mono.Debugging.Client;
 using Mono.Debugging.Evaluation;
+using System.Linq;
 
 namespace MonoDevelop.Debugger.Win32
 {
@@ -28,6 +29,7 @@ namespace MonoDevelop.Debugger.Win32
 		CorStepper stepper;
 		bool terminated;
 		bool evaluating;
+		bool autoStepInto;
 		int processId;
 
 		static int evaluationTimestamp;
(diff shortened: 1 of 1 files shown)


xamarin-release-manager:
Fixed in version 8.5.0.197 (master)

Author: therzok
Commit: d7e4936559cc69cce17f9072c7cbfdf851f81c88 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.4.0.1560 (release-8.4)

Author: therzok
Commit: d7e4936559cc69cce17f9072c7cbfdf851f81c88 (mono/monodevelop)

PR #510, opened 2014-03-26
Registring Windows to CommandManager for keyboard shortcuts

DavidKarlas:
So I intentionally created 3(could be 2) commits 1st is Removing dead code I think that is clear...
Last "Registering DebugValueWindow..." as well since it only uses new method plus closing DebugValueWindow on Escape...
So commit "Adding Reg..."
So I'm not sure how this stuff should be handled in GTK if you want to be able to hook keys events before and after window I created two event hanlers one with [Gtk.ConnectBefore] and one without.

So why are 3 handlers? Well root window wants CommandManager to always handle keys before root window and we keep all code the same this way no possibility for regressions...
With "subwindows" like DebugWindow we don't want that because for example there is Ctrl+F command which is handled by TreeView inside DebugValueWindow so we want that to be handled after if subwindow didn't handle that key already...
So what is that OnWindowKeyPressedBefore doing? Well lets say we have chord Ctrl+W|E for opening errors window you press Ctrl+W and now you press E well if there was no OnWindowKeyPressedBefore E would be typed inside our subwindow(which will most likely handle this key) and result in chord not being finish... So that is there only so it can finish chords...

Main reason for starting this changes was annoying fact that when I left mouse cursor unattandant in middle of screen and I was stepping trough the code or pressing F3(next search) DebugValueWindow would appear if cursor became on top of variable and my F11 would stop reacting(or F3)... Well now if F11 is not handled by window and new step is executed(or search)....
Seem also that I'm used from VS that after viewing values with DebugValueWindow I don't bother closing it but instead just start pressing F11...

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/DebugValueWindow.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/DebugValueWindow.cs
@@ -79,6 +79,7 @@ namespace MonoDevelop.SourceEditor
 			this.Decorated = false;
 
 			TransientFor = (Gtk.Window) editor.Toplevel;
+			IdeApp.CommandService.RegisterWindowForKeyboardShortcuts (this);
 			
 			// Avoid getting the focus when the window is shown. We'll get it when the mouse enters the window
 			AcceptFocus = false;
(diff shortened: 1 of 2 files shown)


DavidKarlas:
https://github.com/mono/monodevelop/commit/b627818f26dfb37b39b1a6ac60299f09bd83b7ed
Fixes main reason for this PR with debugger popup and stepping keyboard shortcuts not working...
So I'm closing this...

PR #512, opened 2014-03-27
Fixed 'Bug 14530 - GUI Thread hangs a while while xml serialization'

x2bool:
> Xml Serialization is likely the >slowest< way to serialize data and shouldn't be used.
> (btw. nrefactory contains an optimized fast binary serializer - that should be used instead)

I changed the serialization method to the ICSharpCode.NRefactory.Utils.FastSerializer.
It stays backward compatible with the previous implementation (xml serialization) when there is a need for this.


xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #328 merged by: Lluis Sanchez
Author: mono
Commit: a92bfa0237cadfe0946dc902853306c2f020bbcc (xamarin/monodevelop)

PR #562, opened 2014-05-24
Added initial gnome-terminal-server support for gnome 3

gliwka:
Fix for bug 16804
https://bugzilla.xamarin.com/show_bug.cgi?id=16804

### DIFF
--- a/main/src/addins/GnomePlatform/GnomePlatform.cs
+++ b/main/src/addins/GnomePlatform/GnomePlatform.cs
@@ -200,7 +200,7 @@ namespace MonoDevelop.Platform
 		}
 		
 		
-		delegate string TerminalRunnerHandler (string command, string args, string dir, string title, bool pause);
+		delegate string TerminalRunnerHandler (string command, string args, string dir, string title, bool pause, Guid applicationId);
 		delegate string TerminalOpenFolderRunnerHandler (string dir);
 
 		string terminal_command;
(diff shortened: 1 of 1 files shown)


slluis:
approve

ircecho:
I tested this and it is not working. Reason is, that the gnome-terminal-server is located at "/usr/libexec/gnome-terminal-server" instead of the path you used. I am using GNOME Terminal 3.14.2 on Fedora 20, but this does not seem to be specific to Fedora (see https://wiki.gnome.org/Apps/Terminal/Debugging#Running_a_separate_instance).

Other than that, the patch is working. I fixed the path and now running in gnome-terminal is working perfectly.

Therzok:
I'm up for merging this.

slluis:
But what's then the correct location of gnome-terminal-server?

Therzok:
Looking at the Gnome docs, it seems to be

`/usr/libexec/gnome-terminal-server`

https://wiki.gnome.org/Apps/Terminal/Debugging

gliwka:
I've looked into it and have found two possible paths:
/usr/lib/gnome-terminal/gnome-terminal-server (found in Arch Linux, Debian, OpenSUSE and most other distributions)
/usr/libexec/gnome-terminal-server (found in Fedora and derivates)

slluis:
@Therzok can you verify this works?

mkrueger:
Couldn't merge the pull request - manually ported it over.

@gliwka thanks - y you're right that's the only way it can now work with the gnome terminal.

new PR:

https://github.com/mono/monodevelop/pull/2574

PR #629, opened 2014-08-15
Type System changes for JavaScript support

harsimranb:
These changes will enable the JavaScript project to cache JavaScript parsed documents, as well as, enable background parsing for JavaScript files.


slluis:
It needs to be redone for current master.

PR #653, opened 2014-10-06
Bug 20530 - Git addin should not stage new files by default

Therzok:


slluis:
@Therzok should we revive this?

Therzok:
Yes. I'll take a look

jmatthiesen:
@sevoku - is this PR something we can revive with our focus on the Git backend right now?

PR #684, opened 2014-12-08
Feature pin tabs

netonjm:
Adding feature to pin the opened tabs in XS, and new tab contextual menu to close all except pinned.


Therzok:
This is a useful feature, we should consider updating this patch and fixing the issues with it.

netonjm:
Sorry I thought this feature was rejected!

netonjm:
This is the new updated and functional branch:
https://github.com/mono/monodevelop/compare/feature_pintabs

PR #686, opened 2014-12-10
Dependent files duplicates

DavidKarlas:
@slluis can you please review this?

Commit 6d8ebb1 fixes:
[Bug 21642](https://bugzilla.xamarin.com/show_bug.cgi?id=21642) - After moving XAML file to other folders the solution sometimes shows 2 cs files

In commit 6d8ebb1I removed SapNodeBuilder because it seems to me like it doesn't have any function. But it inherits ProjectNodeBuilder and [FileAddedToProject is executed 2x](https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectNodeBuilder.cs#L72) it adds 2 instances of children in NodeTree.

Commit d6ba834 actually fixes 2 bugs:
[Bug 20704](https://bugzilla.xamarin.com/show_bug.cgi?id=20704) - Delete option does not remove the file(for backing file like XAML page) permanently
[Bug 20695](https://bugzilla.xamarin.com/show_bug.cgi?id=20695) - Copying an existing XAML file and pasting it into the same folder doesn't work right

Change in ProjectOperations.cs is made so when "projectFile.DependsOn = targetParent.Name;" is called in loop project of file is already set so it can execute [UpdateDependency ](https://github.com/mono/monodevelop/blob/6d8ebb11be56df278c1a9148b62a7646f74f62d1/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectFile.cs#L329).

Change in Project.cs is made so [DepedentChildren of parent file is updated](https://github.com/mono/monodevelop/blob/6d8ebb11be56df278c1a9148b62a7646f74f62d1/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectFile.cs#L324).

Change in ProjectFileNodeBuilder.cs is made because change in Project.cs which modifies DependentChildren and foreach throws exception... So ToArray() is to create copy which doesn't change...


slluis:
Looks good to me.

PR #687, opened 2014-12-10
[Shell] Propose telling addins to use the other overload.

Therzok:
This creates a Gtk.Menu on Mac, which means we don't make use of the awesome native Menu we have implemented in the other overload.
Propose hiding this API (not fully removing it).

/cc @slluis @mhutch @alanmcgovern @bratsche @DavidKarlas


Therzok:
Bump.

slluis:
Before obsoleting this, we need an alternative method for showing context menus, maybe with a generic object model (or maybe using xwt). The overload with CommandEntrySet is not a replacement, since it can't be used to show arbitrary menus.

Therzok:
Closing as the API now takes a `Control` and we have better alternatives with using `ContextMenu`.

PR #688, opened 2014-12-11
[MSBuild] Avoid unnecessary project reevaluation

mhutch:
Avoid unnecessarily re-evaluating MSBuild projects multiple times by creating them with the global properties already set.

Difficult to test performance since it's extremely variable.


slluis:
Looks good to me

mhutch:
IIRC this caused some evaluation issues so I'll just close it for now.

PR #689, opened 2014-12-12
[Core] PointToLocation starts/ends at start/end of char instead middle of char

DavidKarlas:
This logic was added here: https://github.com/mono/monodevelop/commit/f7f17fe73066d3b181d7a0f2a87653b3e75d69fd#diff-2d27ddec65e6cd76bcb8bd8083311dd0R1830
I have no idea why it's good to start/end text Location in middle of character...(I'm maybe making regression, hence PR)

@mkrueger Any idea?

So what this change does it calculates position like this:
image
image

How I noticed this... If you try to hover mouse to see debug value of variable 'l' and you are just one pixel to right of middle(so still above 'l', but one pixel to right of middle) debug tooltip won't show up because IsNearX1 will be false and make index++ resulting into space(between 'l' and '=') instead of 'l', hence not displaying debug tooltip...

Red rectangle indicates hit area for debug tooltip(before/after).


mkrueger:
I remember that code. I did that to improve the mouse user interaction when selecting text with a mouse.

However it *might* be obsolete today becuse when I implemented that the rendering of the text and calculation was different. I'll test that.

PR #693, opened 2014-12-19
Default output filename and ResolveAssemblyReference fixups.

tomkcook:
This request encompasses two separate changes, but I can't figure out how to create separate requests for them.

The change to TemplateGenerator.cs provides a minimal implementation of the ResolveAssemblyReference method.  This uses the reference paths provided on the command-line to attempt to resolve assembly references if they do not already refer to an absolute path.  If the reference cannot be resolved in this way, it returns the assembly name unchanged.  This causes Mono to search for it as normal, either finding it by some other mechanism (as happens for System.\* etc) or reporting an error.

The change to TextTransform.cs removes the requirement for the output file to be named on the command-line.  If the output file is not named, it defaults to the same as the input file.  This is then further modified by a line such as:

```
<#@ output extension=".f90" #>
```

in the template.  This change is to match the behaviour of TextTransform.exe in Visual Studio 2013 on Windows.

### DIFF
--- a/main/src/addins/TextTemplating/TextTransform/TextTransform.cs
+++ b/main/src/addins/TextTemplating/TextTransform/TextTransform.cs
@@ -76,8 +76,14 @@ namespace Mono.TextTemplating
 			}
 			
 			if (string.IsNullOrEmpty (outputFile)) {
-				Console.Error.WriteLine ("No output file specified; using default.");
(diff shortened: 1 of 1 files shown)


mhutch:
Yes, if you'd like to contribute those changes I'd be happy to integrate them.

PR #703, opened 2015-01-06
Icons renaming

vancura:
I moved and renamed icons, to make it simpler to use. People were confused by the -light- snippet in filenames when it was not used in the real path but only in solution file. Here I fix all of these issues.

Also take a look at https://github.com/mono/monodevelop/commit/8eb783c9ce2ee7b23a73d35703cd7a71e6e19a96 where I rename some icons to unify filenames.

One thing can complicate the stuff here: in https://github.com/fsharp/fsharpbinding/blob/master/monodevelop/MonoDevelop.FSharpBinding/Templates/FSharpNUnitLibraryProject.xpt.xml#L8 there's old reference to `project-nunit-overlay-light-32.png` and if we rename it in monodevelop, this template will have a red square icon until it's fixed. I can't change it there, I don't have rw access (/c @7sharp9) -- the change is simple, just remove `-light` so the new filename is `project-nunit-overlay-32.png`


alanmcgovern:
For now we could include both the original image name and new image name to avoid breaking the addins.

However that does raise an interesting question - are these images supposed to be considered public API or private implementation detail? I would hope it's the latter and so we can safely rename them at will.

vancura:
Maybe the icon should be bundled in https://github.com/fsharp/fsharpbinding. This dependency stinks.

vancura:
@alanmcgovern see above.

7sharp9:
We can use our own icons in the F# addin if we wish, lluis requested the icons match the C# ones.

vancura:
@7sharp9 so I guess it would be good to copy that icon to your repo and then I'd remove the duped reference.

mhutch:
Image names are not public APIs, but stock IDs are. I don't see any stock ID changes from a quick skip, so it should be okay.

mhutch:
Because it's so big, it's difficult to review and make sure nothing was missed and nothing else was touched. Marius mentioned in the email thread that he was going to make a script to check this, we should probably wait for that before merging.

PR #704, opened 2015-01-08
[T4] all templates processing for Solution/Project

xplicit:
This is new feature, which allows to update all T4 templates in Solution or selected Project.
To use it, select in Solution View Solution or Project node, right click to open context menu, select Tools/Process Teplates. 

It's very usefull when you have base template and alot of derived templates from it. 
Before this feature is implemented you have to change every derived T4 template and resave it to regenerate the content after changing base template.
With this feature you can do it in one click.

Knowing issue: after regeneration "Tool output" window is created for every templat was processed. 
It would be much better if all information about tool generation process be created in one "Tool output" window, but I don't know now how to do it inside monodevelop. If someone point me how it can be done, I'll do it.

### DIFF
--- a/main/src/addins/TextTemplating/MonoDevelop.TextTemplating/GenerateCommandHandler.cs
+++ b/main/src/addins/TextTemplating/MonoDevelop.TextTemplating/GenerateCommandHandler.cs
@@ -28,6 +28,8 @@ using MonoDevelop.Components.Commands;
 using MonoDevelop.Projects;
 using MonoDevelop.Ide;
 using MonoDevelop.Ide.CustomTools;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace MonoDevelop.TextTemplating
 {
@@ -42,20 +44,33 @@ namespace MonoDevelop.TextTemplating
 		{
 			IWorkspaceObject wob = IdeApp.ProjectOperations.CurrentSelectedItem as IWorkspaceObject;
 
+			IEnumerable<ProjectFile> files = null;
+
 			Solution solution = wob as Solution;
 			if (solution != null) {
 				foreach(Project p in solution.GetAllProjects ())
-					UpdateFilesInProject(p);
+					files = (files == null) ? GetFilesToUpdate (p) : files.Concat (GetFilesToUpdate (p));
 			} else {
 				Project proj = wob as Project;
 
 				if (proj != null) {
-					UpdateFilesInProject (proj);
+					files = GetFilesToUpdate (proj);
 				}
 			}
 
+			if (files != null)
+				CustomToolService.Update (files, true);
+
 			base.Run ();
 
+		}
+
+		private IEnumerable<ProjectFile> GetFilesToUpdate(Project project)
+		{
+			return project.Files.Where (
+				f => f.Generator == typeof(TextTemplatingFileGenerator).Name
+				|| f.Generator == typeof(TextTemplatingFilePreprocessor).Name
+			);
 		}
 
 		private void UpdateFilesInProject(Project project)
--- a/main/src/addins/TextTemplating/MonoDevelop.TextTemplating/GenerateCommandHandler.cs
+++ b/main/src/addins/TextTemplating/MonoDevelop.TextTemplating/GenerateCommandHandler.cs
@@ -1,0 +1,77 @@
+﻿//
+// GenerateCommand.cs
+//
+// Author:
+//       Sergey Zhukov <>
+//
+// Copyright (c) 2015 Sergey Zhukov
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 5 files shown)


mhutch:
Probably the solution to this would be alter CustomToolService.Update so that it takes an IEnumerable<ProjectFile> and uses the same monitor for all of the tasks.

Therzok:
@mhutch Wouldn't that break API?

xplicit:
I can add overloaded `public static void Update (IEnumerable<ProjectFile> files, bool force)`  method, it should not break API and in this case it would be possible to use single-file update via `public static void Update(ProjectFile file, bool force)` and multiple-files update via new method.

xplicit:
I added the the overloaded method for updating several files, generated by a tool.

     public static void Update (IEnumerable<ProjectFile> files, bool force)

"Process Templates" menu item now outputs all information about templates generation into one tool window.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CustomTools/CustomToolService.cs
```diff
@@ -102,22 +102,91 @@ static ISingleFileCustomTool GetGenerator (string name)
...
-		
-		public static void Update (ProjectFile file, bool force)
+
+		public static void Update (IEnumerable<ProjectFile> files, bool force)
 		{
-			var tool = GetGenerator (file.Generator);
-			if (tool == null)
+			var monitor = IdeApp.Workbench.ProgressMonitors.GetToolOutputProgressMonitor (false);
+
+			var fileEnumerator = files.GetEnumerator ();
```
Therzok: Eh? Why not `foreach (var file in files)`?

xplicit: @Therzok because we call then an asynchroneous method Update() and output of the tool.Generate() method should  be corresponded to the appropriate monitor.BeginTask() header. If we use foreach we don't have a guarantee that the first file is generated before the second. And in this case we can get into two issues: the output of the second will go to the output of the first, or if they run in separate threads they can produce race conditions in non-thread safe LogViewProgressMonitor class.
Current implementation generate files one by one (in synchroneous way) asynchroneously without blocking UI thread.

main/src/addins/TextTemplating/MonoDevelop.TextTemplating/GenerateCommandHandler.cs
```diff
@@ -0,0 +1,92 @@
...
+
+				if (proj != null) {
+					files = GetFilesToUpdate (proj);
+				}
+			}
+
+			if (files != null)
+				CustomToolService.Update (files, true);
+
+			base.Run ();
```
mhutch: This is redundant, base does nothing.

main/src/addins/TextTemplating/MonoDevelop.TextTemplating/GenerateCommandHandler.cs
```diff
@@ -0,0 +1,92 @@
...
+			} else {
+				Project proj = wob as Project;
+
+				if (proj != null) {
+					files = GetFilesToUpdate (proj);
+				}
+			}
+
+			if (files != null)
+				CustomToolService.Update (files, true);
```
mhutch: IMO if there are no files to update we should still provide some feedback instead of silent failure. Maybe pass an empty array to CustomToolService.Update and it can report it to the monitor? Or report directly to the status bar monitor?

main/src/addins/TextTemplating/MonoDevelop.TextTemplating/GenerateCommandHandler.cs
```diff
@@ -0,0 +1,92 @@
...
+				if (file.Generator == typeof(TextTemplatingFileGenerator).Name
+					|| file.Generator == typeof(TextTemplatingFilePreprocessor).Name) {
+					CustomToolService.Update (file, true);
+				}
+			}
+		}
+
+		protected override void Update (CommandInfo info)
+		{
+			base.Update (info);
```
mhutch: This is also redundant. However, it should be easy to provide a useful implementation:

    var wob = IdeApp.ProjectOperations.CurrentSelectedItem;
    info.Enabled = wob is Solution || wob is Project;

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CustomTools/CustomToolService.cs
```diff
@@ -102,22 +102,91 @@ static ISingleFileCustomTool GetGenerator (string name)
...
+
+			if (fileEnumerator.MoveNext())
+				Update (monitor, fileEnumerator, force, false, false);
+		}
+
+		private static void Update(IProgressMonitor monitor, IEnumerator<ProjectFile> fileEnumerator, bool force, bool hasErrors, bool hasWarnings)
```
mhutch: Minor style note - please remove redundant modifiers (private) and place a space before (.

mhutch:
A couple of minor issues but apart from that it looks good!

main/src/addins/TextTemplating/MonoDevelop.TextTemplating/GenerateCommandHandler.cs
xplicit: @mhutch How can I get currently selected ProjectFile? (I want to enable this menu for templates too). `IdeApp.ProjectOperations.CurrentSelectedItem` returns null for files, it's not null only for Project and Solution.
I also tried to use `CurrentSelectedWorkspaceItem` but don't know how can I convert WorkspaceItem to ProjectFile type.

xplicit:
Added generation info for empty IEnumerable collections, fixed styling, removed redudant calling of base methods.

xplicit:
@mhutch Could you review the latest changes?

xplicit:
I made misspelling fixes without last one (files -> file), waiting for an answer to the question above.

xplicit:
@mhutch Fixed all misspellings including files -> file.

mhutch:
Pushed with some local fixup, thanks!

PR #705, opened 2015-01-09
[ErrorReporting] Add ApplicationException as whitelist.

Therzok:
ApplicationExceptions should be non-fatal errors and informative to the user.

Should we also add a check in the LoggingService where we filter this exception as non-fatal and always show it as an Error Dialog?


slluis:
ApplicationException "serves as the base class for application-defined exceptions". For example, TargetInvocationException is a subclass of it. It is certainly not an informative exception.

PR #708, opened 2015-01-12
Langugae defaults to C# + unit test

tomkcook:
As discussed with @mhutch in another thread, this change makes 'C#' the default language setting.

I've also had a stab at a unit test for this change, however I can't get the unit tests to compile, apparently because I'm missing some of the stuff in main/external.  If I figure out how to get this set up, I'll check the unit tests and report back here.


monoadmin:

tomkcook:
I have now confirmed that the unit test I've suggested passes.

mhutch:
whitelist

PR #710, opened 2015-01-12
Upgrade csproj files to TargetFramework net4.5 and ToolsVersion 4.0

tpokorra:
this is another step to make Monodevelop compile with mono from git master


tpokorra:
ok, it seems I should not modify the csproj files in main/tests/test-projects/

tpokorra:
hmm, the pull request was against master from June 2014.
will do a clean version again...

PR #711, opened 2015-01-12
Upgrade csproj files to TargetFramework net4.5 and ToolsVersion 4.0

tpokorra:
this is another step to make Monodevelop compile with mono from git master


mhutch:
All of those addins are abandoned, you should not be building them (except the win32 debugger, but that doesn't need to compile with Mono).

Therzok:
How is the Gtk support addin abandoned?

mhutch:
GtkCore != GtkCore2

GtkCore2 was a SoC project a few years back that was a major overhaul of the GTK# designer, but unfortunately Lluis did not want to review/merge it, and now it's bitrotted.

Therzok:
Oh, I just noticed it's not in the solution file. o.O

PR #713, opened 2015-01-16
Bug 25586 - Unable to access Git log for Storyboard files

Therzok:
We need to use a display binding that we know is a File for this to work.

The downside on this commit? We're attaching a source view _and_ we're not attaching the designer view for things like the iOS/Android designer, instead a Source view. (That means the user will have to close the document then reopen it.)

This should be handled differently:
1. We need a VCS display binding in case the Document we're attaching to is a File Document which won't include anything else other than the VCS tabs?
2. We need to make the designers only become non-file documents when more files are open.


Therzok:
Need input on this.

PR #714, opened 2015-01-16
[Core] Added ContextMenu abstractions that build the context menu UI at ...

sgmunn:
...the time of the menu being shown.

Largely copied from Xwt.Menu but because Xwt decides on the backend at time of creation, and it defaults to Gtk, we don't get chance to use Mac native menus.
This will allow us to construct custom context menus (that are not based on command extensions) with arbitrary click handlers that will use the appropriate UI backend when shown.


main/src/core/MonoDevelop.Ide/MonoDevelop.Components/ContextMenuItem.cs
```diff
@@ -0,0 +1,127 @@
...
+		public ContextMenuItem ()
+		{
+			this.Visible = true;
+			this.Sensitive = true;
+
+			if (!IsSeparator)
+				UseMnemonic = true;
+		}
+
+		public ContextMenuItem (Xwt.Command command) : this()
```
slluis: I don't think we need this.

slluis:
Looks good!

PR #721, opened 2015-01-23
Introduce xUnit.NET testing framework support

bratsche:
Since @x2bool deleted his repository, here is his work on xunit testing. This was previously in https://github.com/mono/monodevelop/pull/633

I've just applied the patch from that pull request, so he still retains authorship if we merge this.

/cc @slluis


slluis:
This has been made available in the add-in repo.

PR #722, opened 2015-01-24
[AddIns] Disable Gtk addin by default.

Therzok:
It's a common source of memory leaks and it will take quite some time to patch up. Disable it by default for now, let's not make others suffer because of an AddIn they'll never use.

### DIFF
--- a/main/src/addins/MonoDevelop.GtkCore/AddinInfo.cs
+++ b/main/src/addins/MonoDevelop.GtkCore/AddinInfo.cs
@@ -6,6 +6,7 @@ using Mono.Addins.Description;
 [assembly:Addin ("GtkCore", 
 	Namespace = "MonoDevelop",
 	Version = MonoDevelop.BuildInfo.Version,
+	EnabledByDefault = false,
 	Category = "IDE extensions")]
 
 [assembly:AddinName ("GTK# Visual Designer")]


mhutch:
+1. I've had this disabled locally for 5+ years.

alanmcgovern:
I've disabled this and so has my wife!

slluis:
I don't like the idea of disabling such a major feature on Linux.

Therzok:
Should I change DisabledByDefault (or add a option) which besides the value, has a PlatformInclude/PlatformExclude parameter?

decriptor:
@slluis A thought about the feature and Linux. I'm not entirely sure what the addin does, but is it tied to gtk-sharp2? If so, its usefulness might be dwindling because of the move to gtk-sharp3.

mhutch:
TBH this shouldn't really be necessary - in theory it has negligible impact on projects that don't reference gtk-sharp and don't have designed widgets. If these checks aren't working correctly then we should fix them.

The perf impact is very noticeable on the MD/XS solution because it has so many projects that *do* have designed GTK widgets. But for most people it shouldn't be an issue.

When working on MD you can easily toggle it with the DISABLE_STETIC env var.

Therzok:
Superseded by #1404 .

PR #724, opened 2015-01-26
[build] Make Mavericks the minimum supported version of MacOSX

alanmcgovern:
The vast majority of our users are using Mavericks or higher so
we are going to make Mavericks the minimum requirement for future
releases.

This should be clearly communicated before we merge and ship this.


Therzok:
I approve of this with all my :heart:.

PR #726, opened 2015-02-02
[VersionControl] Fix thread hanging for recursive queries.

Therzok:


slluis:
This looks a bit over-engineered to me. What about this:
* AcquireLockForQuery(path) checks if a query already exists. If it does, it returns it. If it doesn't it creates the query with its own ManualResetEvent.
* GetDirectoryVersionInfo() calls AcquireLockForQuery to get/create the query, then waits in the event.

All the other changes are then not required. If more than one thread calls GetDirectoryVersionInfo() with the same path, only one query will be created and queued, and all will be waiting on the same event.

Therzok:
My solution was trying to diminish the waiting time by having two queries to the same path return the same result. I'll go ahead with your solution.

Therzok:
I want to note that what you said will not work if the query has already been started on the query thread. That is, the query has been copied to the RecursiveDirectoryQueryQueueClone.

Therzok:
Fixed now, should work just fine.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs
```diff
@@ -212,20 +212,36 @@ public IEnumerable<VersionInfo> GetVersionInfo (IEnumerable<FilePath> paths, Ver
...
 				if (recursive) {
-					using (var mre = new ManualResetEvent (false)) {
-						var rq = new RecursiveDirectoryInfoQuery {
+					RecursiveDirectoryInfoQuery rq;
+					lock (queryLock)
+						rq = AcquireLockForQuery (localDirectory);
+
+					if (rq == null) {
+						var mre = new ManualResetEvent (false);
+						rq = new RecursiveDirectoryInfoQuery {
```
slluis: The creation and queuing of RecursiveDirectoryInfoQuery should be done inside a lock, otherwise several threads could end creating the query at once.

Therzok:
How's it now? Probably the Interlocked.Increment/Decrement are not needed now, but this should do it.

slluis:
Yeah, the count is probably not necessary. I don't think it is necessary to explicitly dispose event objects.
Other than that, it looks good to me now.

Therzok:
Well, the ManualResetEvent has a finalizer that will cleanup, but we can't guarantee the same for .NET. I'd say this works.

PR #731, opened 2015-02-08
Add a command to sort the selected lines in the source editor

mhutch:


mkrueger:
btw. unit tests are missing. For each text editor command there should be at least 1 test.

PR #732, opened 2015-02-08
Use a wizard for the ASP.NET project templates

mhutch:


slluis:
A few comments:
* The design of the wizard is inconsistent with all other wizards.
   * The group of widgets should be vertically centered.
   * Unrelated options (the unit test option) should be separated from the other options using an horizontal separator, and using the same vertical spacing.
   * There should be no white box at the right.
   * We'll need an image for the right side of the template selector. I can ask Vaclav.
* I haven't been able to build any template out of the box
   * All projects fail with unknown System.Web.Http namespace
   * The default Web Forms template has a file using the System.Web.Mvc namespace.
* The test project should include at least one test class with a sample test method.
* I find it confusing to have MVC and WebForms templates and then offering to add references for them in the wizard. Why is this useful? If a user is going to work on an MVC project it is unlikely he'll need Web Forms too. And if there is need, it is very easy to just add a reference to the project.

mhutch:
bump

mhutch:
> The design of the wizard is inconsistent with all other wizards.
> The group of widgets should be vertically centered.
> Unrelated options (the unit test option) should be separated
> from the other options using an horizontal separator, and using
>  the same vertical spacing.

Not a big deal, it looks fine. And there is zero public helper code for this.

> There should be no white box at the right.

I tried that, it looks better moving through the wizard pages when it has the box.

> We'll need an image for the right side of the template selector. I can ask Vaclav.

Sure, not a blocker though.

> I haven't been able to build any template out of the box
> All projects fail with unknown System.Web.Http namespace

Probably a Mono bug, it works fine on Windows. Someone with a Mac will have to look into this.

> The default Web Forms template has a file using the System.Web.Mvc namespace.

Where?

> The test project should include at least one test class with a sample test method.

Any suggestions as to WHAT it should test? This are near-EMPTY templates, any dummy test would just have to be removed by the user.

> I find it confusing to have MVC and WebForms templates and then offering to add references for them in the wizard. Why is this useful? If a user is going to work on an MVC project it is unlikely he'll need Web Forms too. And if there is need, it is very easy to just add a reference to the project.

Yeah, the web forms option doesn't do anything, I just put that in for consistency with the VS wizard and so that other templates that use with wizard can use it. It could be removed from the `<SupportedParameters>` for these templates, but kept in the wizard.

It's NOT trivial to retroactively add MVC and Web API. The wizard sets up routing, config, etc.

mhutch:
I finally got around to trying this on my Mac.

I get the "unknown System.Web.Http namespace" if I create a template without WebApi support. It seems the FileTemplateReference in the project template doesn't get passed the variable values from the project template. IMO this is a bug in the template engine. I've fixed it.

In other cases, though it can build, the compiled site won't run. On Mono 3.12 it fails due to the System.Web.HttpTaskAsyncHandler class missing in Mono's System.Web. On Mono 4.0 it fails due to not being able to write to the registry. I added a workaround for the registry issue for this.

So currently on Mono 4.0 the projects created by these templates work.

mhutch:
build

main/src/core/MonoDevelop.Core/MonoDevelop.Core.StringParsing/StringTagModelExtensions.cs
```diff
@@ -0,0 +1,76 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+
+namespace MonoDevelop.Core.StringParsing
+{
+	public static class StringTagModelExtensions
```
slluis: This logic for evaluating conditions and values is very specific to the template engine. It should not be exposed in the public api.

slluis:
@mhutch would you mind pushing this branch to the monodevelop repository? It will make it easier to build in wrench so that QA can test, and other people can help polishing the new features.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.StringParsing/StringTagModelExtensions.cs
mhutch: In what way is it specific? It looks pretty general purpose to me.

Anyway, fixing FileTemplateReference needed the logic to be available in both MonoDevelop.Core and MonoDevelop.Ide and I didn't think copy-pasting it was a good idea because it would probably get out of sync.

mhutch:
Done!

main/src/core/MonoDevelop.Core/MonoDevelop.Core.StringParsing/StringTagModelExtensions.cs
slluis: The syntax supported by EvaluateCondition is specific to the template engine. Other features may need a different syntax. I just don't like the IStringTagModel interface to be polluted with methods which are feature-specific. If we need that method both in Core and Ide, then we can expose it elsewhere. Maybe in a TemplateConditionEvaluator class? @mrward, up to you.

mhutch: There's nothing fundamentally templating-specific about it and it seems like the kind of thing that could be useful elsewhere, especially if it's extended to support boolean logic (right now it's just `PropertyName` or `!PropertyName`).

slluis: The evaluation code was implemented for and it is used only by the template system. We have no other uses for it right now, so it is premature to expose it as a generic feature. And if we have to do it, it would be better to expose it through an independent evaluator class. There is no need to tightly couple it to IStringTagModel.

mhutch:
Any progress on this?

obadz:
Hey side question: how does one build MonoDevelop without the AspNet addin?

obadz:
@mhutch or anyone else.. would one of you be able to point me in the right direction?

PR #735, opened 2015-02-16
[Ide] Resolve full path of fileName before opening in Workbench

knocte:
Opening the realpath of a file prevents problems with debugging symlinked
files, such as the case https://bugzilla.xamarin.com/show_bug.cgi?id=27098

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
@@ -1285,7 +1285,7 @@ namespace MonoDevelop.Ide.Gui
 				return fileName;
 			}
 			set {
-				fileName = value.CanonicalPath;
+				fileName = FileService.ResolveFullPath (value.CanonicalPath);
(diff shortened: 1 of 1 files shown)


monoadmin:

slluis:
whitelist

slluis:
This should be fixed in the overload that takes the FileOpenInformation argument.

knocte:
> This should be fixed in the overload that takes the FileOpenInformation argument.

Oh good point. But, actually, looking at it more deeply, how about calling FileService.ResolveFullPath() in the setter of FileOpenInformation's FileName property?

DavidKarlas:
Is this also fixing https://bugzilla.xamarin.com/show_bug.cgi?id=21617?

knocte:
> Is this also fixing https://bugzilla.xamarin.com/show_bug.cgi?id=21617?

Quite possibly that bug is a duplicate of bug 27098 yes.

knocte:
@slluis updated PR, note I went with the setter solution in order to have less mutation in the FileOpenInformation class.

slluis:
Thanks!

PR #736, opened 2015-02-16
[FindInFiles] Use relative paths in the Path column, when possible

knocte:
Relative paths to the solution base directory are easier to read
without horizontal scrolling.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
@@ -39,6 +39,7 @@ namespace MonoDevelop.Ide.FindInFiles
 {
 	public enum PathMode {
 		Absolute,
+		Relative,
 		Hidden
 	}
 
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/ISearchProgressMonitor.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/ISearchProgressMonitor.cs
@@ -34,7 +34,6 @@ namespace MonoDevelop.Ide.FindInFiles
 {
 	public interface ISearchProgressMonitor: IProgressMonitor
 	{
-		void SetBasePath (string path);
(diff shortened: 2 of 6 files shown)


monoadmin:

slluis:
whitelist

slluis:
This doesn't look right to me. I often use Find in File to search in directories outside the currently open solution, and it doesn't make sense to show relative paths in this case. Maybe we could show relative paths when using the "Whole Solution" option, but even in this case, it will look wrong if there is more than one solution open.

Therzok:
This also won't work properly for Shared Projects.

knocte:
> I often use Find in File to search in directories outside the currently open solution, and it doesn't make sense to show relative paths in this case.

Good point! I'll make sure to not pervert this case.

> Maybe we could show relative paths when using the "Whole Solution" option, but even in this case, it will look wrong if there is more than one solution open.

I'll create a new PR that changes "Whole Solution" to "All Solutions" when there are more than one solution open. And then after that is merged, I could make this relative paths change only work when there's only one solution open.

> This also won't work properly for Shared Projects.

Why not?

Therzok:
> This also won't work properly for Shared Projects.
> Why not?

Solution A
- Shared Project C

Solution B
- Shared Project C

You're always taking the first defined project, so that means you'll only display the item relatively to Solution A, never to B.

Therzok:
@knocte @slluis SearchResult.Projects isn't set when not using an in-IDE search.

knocte:
> You're always taking the first defined project, so that means you'll only display the item relatively to Solution A, never to B.

But your case falls in the case already commented above: multiple solutions, in which case I plan to not use relative paths (in a 2nd iteration of this PR).

Therzok:
Yeah, Late edit marked that I didn't read above, thought I had read them. Sorry once again for being misleading.

knocte:
Updated PR so it only shows relative paths now, when solutions.Count == 1

Therzok:
This no longer cleanly applies. Can you rebase?

knocte:
Sure, will do later this week.

knocte:
Rebased.

Therzok:
LGTM, will merge when it's green.

knocte:
Green :)

PR #739, opened 2015-02-17
More PropertyGrid fixup

mhutch:
* Support editing items with standard values that are not convertible from strings
* Support RefreshPropertiesAttribute
* Allow entering arbitrary build actions
* Some more minor fixes


mhutch:
build

PR #740, opened 2015-02-18
[FindInFiles] Only use "Whole solution" when there is only one sol open

knocte:
MonoDevelop can open multiple solutions at once (or workspace documents
that can contain more than one solution), so using "Whole solution" as
the wording for this search option always is not ideal.

The enum element is renamed too, for clarity.


slluis:
Thanks!

knocte:
> If Solutions count is 0, should we still show All Solutions

There's already an error message for that particular case, in case the user tries to search this way.

PR #742, opened 2015-02-18
[Ide] Fix solution pad collapsing bug

slluis:
Fixes bug #20932 - Tapping on a file or folder in the solution pad of a SAP
causes the project to collapse.


mhutch:
build

mhutch:
(NuGet timed out)

PR #743, opened 2015-02-18
[FindInFiles] Hide the "Path" column if scope is current doc or selection

knocte:
There's no much point in showing the Path column when the scope of
the FindInFiles search is "Current document" or "Selection" because
its value is going to be the same across all rows.

This is developed with a PathMode enum which will be useful too, for
a future commit that will allow using relative paths in certain
situations.


Therzok:
:+1:

PR #748, opened 2015-02-23
[Core] Add FileCopied and FileMoved events to FileService

sandyarmstrong:
These make it a lot easier to perform some action when (for example)
solution nodes are dragged or copied around.


sandyarmstrong:
@abock @alanmcgovern @slluis had to open a new PR for this, sorry for the trouble.

alanmcgovern:
lgtm, but i don't actively work on MD so I don't want to merge it

PR #749, opened 2015-02-23
 [Core] Resolve full path of workspaceItem before reading

knocte:
Opening the realpath of a file prevents problems when using solutions
inside symlinked paths, such as:
- compilation problems [1].
- rendering of ugly relative paths (cosmetic) [2].
    
This is a similar fix to [3].
    
[1] https://bugzilla.xamarin.com/show_bug.cgi?id=27097
[2] https://bugzilla.xamarin.com/show_bug.cgi?id=27331
[3] https://github.com/mono/monodevelop/commit/32e25ea


knocte:
Uploaded new fix which is a bit more elegant.

knocte:
Oops, corrected bug number.

knocte:
Updated PR to avoid conflict.

PR #751, opened 2015-02-25
[Ide] show warning info if project for unsupported framework is created

loganek:
Add SelectPanel method with three arguments. Section Id can be treated as
identifier only inside one section. For searching through a few sections
level, parent Id might be necessary.

This commit fixes bug #24458

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/OptionsDialog.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/OptionsDialog.cs
@@ -314,6 +314,16 @@ namespace MonoDevelop.Ide.Gui.Dialogs
 				}
 			}
 		}
+
+		public void SelectPanel (string id, string parentId)
+		{
+			foreach (OptionsDialogSection section in pages.Keys) {
+				if (section.Id == id && section.Parent != null && section.Parent.Id == parentId) {
+					ShowPage (section);
+					return;
+				}
+			}
+		}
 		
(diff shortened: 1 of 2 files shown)


monoadmin:

mhutch:
The code looks good but I'm not sure this behaviour is the best way to fix the issue. It could lead to the case where, where a user opens a project from elsewhere, they are unable to close the options dialog unless they change the framework. Worse, if no compatible frameworks are installed they may not be able to change the dialog at all.

Also:

* The project's "supported frameworks" are mostly a hint, since the MSBuild targets can do *anything* in terms of framework resolution and invoking arbitrary compiler.
* The supported frameworks are hardcoded as the supported frameworks of the Mono VB.NET compiler, not of .NET. And soon we'll be using .NET's Roslyn compiler anyway.
* If the user opens an existing project, it could have all sorts of values or custom targets, and we can only determine whether it works by building it.

The problem here is really that when a user *creates* a project in XS, we want it to be able to build without issues. We don't know for sure what the compiler etc on the system support, however we should be able to assume a baseline version e.g. .NET 4. So IMO, the real fix is to make the templating system use the highest framework supported by the project, if the default framework is not supported.

mhutch:
We may also be able to bump the VB.NET binding to support .NET 4.5, but the templating fix would make sense anyway.

loganek:
Fixing templating seems to be very flexible solution, I'll do new pull request soon.

slluis:
The check should be done in the panel that allows changing the framework, not in ProjectOperations.cs.

loganek:
I've improved my previous patch https://github.com/mono/monodevelop/pull/752 
I think this pull request can be closed.

PR #754, opened 2015-02-28
[Core/Ide] Sanitize file:// URIs in FilePath ctor

knocte:
This file:/// translation pattern happens in many places:

 if (filename.StartsWith ("file://", StringComparison.Ordinal))
     filename = new Uri(filename).LocalPath;

All of them use strings for paths, which is a smell
(http://c2.com/cgi/wiki?PrimitiveObsession) so we can kill two
birds with one stone if we change them to use FilePath instead,
and move this path sanitization to the FilePath ctor.


slluis:
Thanks!

PR #766, opened 2015-03-03
[Ide] Fix tree icon rendering issue

slluis:
CellRendererImage can't take null images. CellRendererImage.NullImage has
to be used instead.

Fixes bug #27323 - Lots of GtkTreeView errors due to new debugger visualizers


DavidKarlas:
@slluis Thank you, I wasn't aware of this possibility

PR #783, opened 2015-03-10
[Windows] Disable WPF usage if running under Mono.

Therzok:


slluis:
We'll likely introduce more WPF use in the future, so we can't really support MD running on Mono.

Therzok:
At least d8412bd and 17dfeed can go in?

This was for testing purposes, @DavidKarlas wanted to see how XS worked on mono windows.

slluis:
No, we should not have conditions based on the clr.

PR #791, opened 2015-03-11
Bug 27236 - Debugging tooltips aren't being displayed

DavidKarlas:


DavidKarlas:
I can't think of any regression that this can cause so I think it should be merged into 5.9

PR #795, opened 2015-03-13
Last 5.9 XMToolbar fixes

Therzok:
There are a few items remaining, but I can't fix them:
* Undo/redo for NSTextField
* https://bugzilla.xamarin.com/show_bug.cgi?id=27920

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/EditCommands.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/EditCommands.cs
@@ -91,6 +91,34 @@ namespace MonoDevelop.Ide.Commands
 			IdeApp.Workbench.ShowDefaultPoliciesDialog (null);
 		}
 	}
+
+	internal class DefaultUndoHandler : CommandHandler
+	{
+		protected override void Run ()
+		{
+			// Nothing. Will be forwarded to gtk
+		}
+
+		protected override void Update (CommandInfo info)
+		{
+			// Forward to gtk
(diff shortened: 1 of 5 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/EditCommands.cs
```diff
@@ -91,6 +91,34 @@ protected override void Run ()
 			IdeApp.Workbench.ShowDefaultPoliciesDialog (null);
 		}
 	}
+
+	internal class DefaultUndoHandler : CommandHandler
+	{
+		protected override void Run ()
+		{
+			// Nothing. Will be forwarded to gtk
```
slluis: Why is this necessary?

Therzok: Because, if don't do this, we'll end up [here](https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs#L405), that means we'll be passing the native implementation that _we_ should be able to handle it, tried handling it and couldn't.

With this change, the command is routed to the native implementation in case we couldn't handle it.

Therzok: I'm pretty convinced that undo/redo is broken somewhere at GtkQuartzWindow-or-similar level, regarding the NSUndoManager. At least, this should route to the native implementation if we fix that.

PR #799, opened 2015-03-14
[Ide] When window is unfocused, let most system commands fall through

bratsche:
When our main window isn't focused, most system commands should
just pass through. For example, Cmd-Shift-G in the file open
dialog should open the "Go to folder" sheet.

Exceptions to this are made for SelectAll, Cut, Copy, Paste.

Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=27940

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -374,12 +374,26 @@ namespace MonoDevelop.Components.Commands
 			}
 			
 			bool bypass = false;
+			List<string> catchInAllWindows = new List<string> {
+				"MonoDevelop.Ide.Commands.EditCommands.SelectAll",
+				"MonoDevelop.Ide.Commands.EditCommands.Cut",
+				"MonoDevelop.Ide.Commands.EditCommands.Paste",
+				"MonoDevelop.Ide.Commands.EditCommands.Copy"
+			};
 			for (int i = 0; i < commands.Count; i++) {
 				CommandInfo cinfo = GetCommandInfo (commands[i].Id, new CommandTargetRoute ());
 				if (cinfo.Bypass) {
 					bypass = true;
 					continue;
 				}
+
+				var focusSkip = !catchInAllWindows.Contains (commands [i].Id.ToString ()) && !IdeApp.Workbench.HasToplevelFocus;
+
+				if (focusSkip) {
+					bypass = true;
+					continue;
+				}
+
 				if (cinfo.Enabled && cinfo.Visible && DispatchCommand (commands[i].Id, CommandSource.Keybinding))
 					return result;
 			}


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -374,12 +374,26 @@ bool ProcessKeyEvent (Gdk.EventKey ev)
...
+				"MonoDevelop.Ide.Commands.EditCommands.Copy"
+			};
 			for (int i = 0; i < commands.Count; i++) {
 				CommandInfo cinfo = GetCommandInfo (commands[i].Id, new CommandTargetRoute ());
 				if (cinfo.Bypass) {
 					bypass = true;
 					continue;
 				}
+
+				var focusSkip = !catchInAllWindows.Contains (commands [i].Id.ToString ()) && !IdeApp.Workbench.HasToplevelFocus;
```
Therzok: Since we use commands often, I'd rather see something in the lines of:
 * A static Dictionary in which we check contains, not a LIst.
 * Probably cache by CommandId?

Therzok: *Why* I think a Dictionary should be used: http://www.dotnetperls.com/dictionary-time

Therzok: Also, is it necessary to verify `HasTopLevelFocus` on each iteration?

bratsche: Well we need to know if it has focus. Alternately we could listen for focus changes, but this seemed easier.

Therzok: No, I meant in the `for` loop. Commands are run on the UI thread, so I don't think that value can change mid-command check.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -374,12 +374,26 @@ bool ProcessKeyEvent (Gdk.EventKey ev)
 			}
 			
 			bool bypass = false;
+			List<string> catchInAllWindows = new List<string> {
```
slluis: This should be a static field. There is no point in creating the list at every key press.
It should be an array of objects containing the command ids as enums. No need to use strings to compare.

bratsche:
Thanks, added another commit here https://github.com/mono/monodevelop/commit/6dc1b16b79d398ce2153b3b73c0f9f9f07a82510

Therzok:
This breaks things. If a debugger popup is shown, then disappears, we can't properly process `step over` for example.

Therzok:
Improvements in https://github.com/mono/monodevelop/pull/803

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #803 merged by: Marius Ungureanu
Author: mono
Commit: 244725ed3f30cfa71b9222d4480588c3747c748a (xamarin/monodevelop)

PR #803, opened 2015-03-16
Fix to unfocused workbench fallthrough.

Therzok:
This PRs goal is to fix #799, in case of no focused window in the application.

That means, with #799, if we opened a popup, then closed it and tried a command, it wouldn't get processed if it weren't an editor command.

Now we route the implementation to the window and test if it can perform the event first.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -31,6 +31,7 @@ using System;
 using System.Reflection;
 using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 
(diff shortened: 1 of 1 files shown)


Therzok:
If we want the PR description in the commit message, please tell.

slluis:
This doesn't look right, MD should have a chance to run the command before the default handler of the window.

Therzok:
The bug that this fixes is that MD was capturing the command and doing something else when we had another window on top (see https://bugzilla.xamarin.com/show_bug.cgi?id=27940).

With this, the window will try to do its default actions first, and if the window cannot, then we trace back to XS.

Therzok:
Updated with essay.

slluis:
With this patch, if a native window has the focus and does not handle the key event, then it will fallback to the command system, and we don't want that.

PR #805, opened 2015-03-16
[IDE] Align the images in the previous/next buttons

iainx:
The images were sitting very high in the tabstrip, this centres them vertically.


Therzok:
:+1:

Therzok:
Can you also apply it to the dropdown button on the far right?

Therzok:
This looks like it introduces a regression on Windows. (Not sure about Linux)

See that it's not properly centered now.

PR #820, opened 2015-03-22
[configure] Move the full check for pkg-config

knocte:
Checking for pkg-config was being done twice, and the first time
that was done was incomplete.


Therzok:
:+1:

PR #823, opened 2015-03-23
[Core] Fix terminating error when creating new solution.

mrward:
Fixed bug [28323](https://bugzilla.xamarin.com/show_bug.cgi?id=28323) - MonoDevelop terminating error when creating new solution with solution already opened.

The DataReceivedEventArgs passed to the event handler for the Process.ErrorDataReceived event can have a null Data property on Windows.


mrward:
Looks like this fix was cherry-picked from master to the monodevelop-5.9-branch. Closing this pull request.

PR #825, opened 2015-03-23
[ide] Folding persistence on disk, + Fix bug;

jpbruyere:
- Fix Code folding.
- Add Code folding persistence on disk, with fixed cache to 1mb, + add an entry in the editor option panel

### DIFF
--- a/main/src/addins/MacPlatform/MainToolbar/SelectorView.cs
+++ b/main/src/addins/MacPlatform/MainToolbar/SelectorView.cs
@@ -217,7 +217,7 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 				return menu;
 			}
 
-			void CreateMenuItem (NSMenu menu, IRuntimeModel runtime)
+			NSMenuItem CreateMenuItem (NSMenu menu, IRuntimeModel runtime)
 			{
 				NSMenuItem menuItem;
 				string runtimeFullDisplayString;
@@ -265,6 +265,7 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 					};
 				}
 				menu.AddItem (menuItem);
+				return menuItem;
 			}
 
 			public PathSelectorView (CGRect frameRect) : base (frameRect)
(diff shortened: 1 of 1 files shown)


mhutch:
approve

jpbruyere:
Next step is to add user preferences related to this function
- save on disk or not
- clear cache
- max space allowed for caching, removing older files by access time order.

jpbruyere:
If carret is positionned by a search task, the reloaded settings override position. I'll fixe that.

mkrueger:
However the default imports folding would need to be implemented in the c# folding parser therefore I don't know if it's a good thing. The folding parser is used when opening files - it's much faster than the real c# parser and prevents the file from jumping around when foldings that are folded are added.

For folding persistence I don't think we need an option. Can you remove that option ?

There are 2 problems:

a) The foldings are updated a while after loading the file currently. Therefore you would have the 'jump around' effect. How is that solved ? Wouldn't it be better to store the fold tree in the settings ?

b) It's needed to check that the file that is opened is the same that it was when the fold states are saved. Otherwise fold states and file wouldn't match.

jpbruyere:
a) I solved the jumping effect by duplicating offset setup, one for init and load of folding states, and the normal property is kept in all other cases. (still need exhaustive test cases to ensure it's bug free)
I also had a look at the Xml config store, but would size increase not be a problem for several fold trees? (note: binary serializer seems not to optimize space as I had thought)

I have also implemented a cache size check, (with fixed size to 1mb, (create new option?)).
the cache is also deleted if fold persistence is disabled .

b) time consistency is now checked.(source.writeUtc > folding.writeUtc => delete) I have to ensure
it's not disturbed by autosave recovering...If you have some steps for folding improvment that you think should be done, I could try to see if I'm able to complete some...

Foldings reload takes care also of caret position, and let open fold containing it. 

I personaly like folding persistence, specialy during creation from scratch time. For debug and code learning, it's maybe not so true. I will set default option to 'false' in next commits.
I may also try to set it as an addin, but I dont feel confident in doing so...

Using directives folding opion has been hidden (checkbox.visible=false, I should maybe completely remove the default behaviour, but keep the possibility to fold using in the editor), Maybe I could update csharpfoldingparser to parse them, I'll give it a try.

jpbruyere:
I've moved the CodeTemplate fix to another branch, and also the multiselect fix in reference add. To keep only folding code in this branch.

PR #826, opened 2015-03-23
Monodevelop 5.9 branch bug 19640

sgmunn:
Fixes bug 19640 - Resource.designer.cs not recompiled on close-and-save

When prompting the user to save the file if the user closes the document, we don't send a file change event. We should really be calling Document.Save() rather than replicating part of the save logic here.

Also addresses a minor error where, if the file is unsaved, and the user cancels the SaveAs, we get an ArgumentNullException.


slluis:
whitelist

PR #827, opened 2015-03-25
[Ide] Use the caret position, not window origin, to determine the screen...

bratsche:
... geometry

Fixes bug #27911


mhutch:
whitelist

PR #829, opened 2015-03-26
[Core] Fix fast-build-check issue

slluis:
Mark a project as dirty for fast-build-check when it is cleaned.
Fixes bug #28449


directhex:
build

PR #830, opened 2015-03-26
Set search pattern as title for search pad window

jpbruyere:
When multiple search windows are opened, having a brief reminder of what was the initial search text is missing for me.

The search pad title appears to me the best place to hold the initial pattern.


slluis:
whitelist

PR #832, opened 2015-03-27
Update UserInterfaceTests with template tests

manish:
Fixes UserInterfaceTests

Tests templates for templates in file `MonoDevelopTemplatesTest.cs`
* Console Project
* Gtk# 2.0 Project
* Library
* NUnit Library Project

Steps taken for all the template tests
* Open `New Solution` dialog
* Select `.NET`
* Select the template type
* Click `OK`
* Enter `Project Name`
* Enter `Solution Name`
* Enter `Location`
* Asserts that `Create a project within the solution directory` is enable and visible. If yes, then unchecks it.
* Asserts that `Create a .gitignore file to ignore inessential files` is enabled and visible. If yes, then unchecks it.
* calls `beforeBuild` action, which is the action to be taken before the solution is built. In case of NUnit Project, that delegate blocks waiting for `Package updates are available.` to be visible in Status Bar
* Builds the project
* Check if there any error
* Closes the solution
* Tries to delete the solution folder.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestSession.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestSession.cs
@@ -326,6 +326,20 @@ namespace MonoDevelop.Components.AutoTest
 			});
 		}
 
+		public object GetPropertyValue (string propertyName)
+		{
+			return Sync (delegate {
+				PropertyInfo propertyInfo = CurrentObject.GetType().GetProperty(propertyName);
+				if (propertyInfo != null) {
+					var propertyValue = propertyInfo.GetValue (CurrentObject);
+					if (propertyValue != null)
+						return propertyValue;
+				}
+
+				return false;
+			});
+		}
+
 		public bool SetPropertyValue (string propertyName, object value, object[] index = null)
 		{
 			return (bool)Sync (delegate {
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestSession.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestSession.cs
@@ -32,6 +32,7 @@ using System.Threading;
 using System.Collections.Generic;
 using MonoDevelop.Core.Instrumentation;
 using MonoDevelop.Ide;
+using MonoDevelop.Ide.Tasks;
 
 namespace MonoDevelop.Components.AutoTest
 {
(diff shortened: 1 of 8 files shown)


manish:
@mhutch @iainx

iainx:
approve

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestSession.cs
```diff
@@ -137,11 +138,32 @@ public void TypeText (string text)
...
+				var itemSelected = SelectWidget (treeViewName, true);
+				if (!itemSelected)
+					return false;
+
+				bool parentFound = after == null;
+				var treeView = currentObject as Gtk.TreeView;
+				bool result = false;
+				if (treeView != null) {
+					treeView.Model.Foreach ((model, path, iter) => {
+						var iterName = (string)model.GetValue (iter, 0);
```
slluis: We'll have to expose a semantic model of the tree. Item names are not always in column 0.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestSession.cs
```diff
@@ -406,7 +461,7 @@ public void WaitForWindow (string windowName, int timeout)
 
 			GLib.Timeout.Add ((uint) pollTime, () => {
 				var window = GetFocusedWindow (false);
-				if (window != null && window.Name == windowName) {
+				if (window != null && window.GetType ().FullName == windowName) {
```
slluis: The window type name is an implementation detail. We should avoid using those names in the tests.

main/tests/UserInterfaceTests/CreateBuildTemplatesTestBase.cs
```diff
@@ -0,0 +1,120 @@
...
+			var solutionParentDirectory = Util.CreateTmpDir (projectName);
+			string actualSolutionDirectory = string.Empty;
+			try {
+				var newProject = new NewProjectController ();
+				newProject.Open ();
+
+				SelectTemplate (newProject, kind, category, categoryRoot);
+
+				Assert.IsTrue (newProject.Next ());
+				Thread.Sleep (3000);
```
slluis: Tests should not use fixed waits. Instead, they should wait for events from the UI.

main/tests/UserInterfaceTests/NewProjectController.cs
```diff
@@ -0,0 +1,120 @@
...
+
+		public bool UseGit (bool useGit, bool useGitIgnore = true)
+		{
+			var gitSelectionSuccess = SetCheckBox ("useGitCheckBox", useGit);
+			if (gitSelectionSuccess && useGit)
+				return gitSelectionSuccess && SetCheckBox ("createGitIgnoreFileCheckBox", useGitIgnore);
+			return gitSelectionSuccess;
+		}
+
+		public bool SetCheckBox (string checkBoxName, bool active)
```
slluis: This method and the others below are not specific to NewProjectController, so they should be in some other reusable class.

slluis:
I added some comments about stuff that should be fixed in the test framework. However, I'm ok about committing this as is until we have a new testing infrastructure.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestSession.cs
manish: I have not yet proposed it in inclusion as I did not want it to drop huge amount of code, but I have it working in my private branch

PR #833, opened 2015-03-28
[ide] Update PinnedWatchWidget position in TextArea when grabbed with mo...

jpbruyere:
It was impossible to move pinned watches.


slluis:
Watches are moved here: https://github.com/mono/monodevelop/blob/master/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SourceEditorView.cs#L1275.

Something else must be wrong.

slluis:
Thanks!

PR #837, opened 2015-03-30
[code] Fix Generic collection argument type return for code template

jpbruyere:
Fix codeTemplate **GetComponentTypeOf** bug for generic collections.

**note**
Default 'foreach' template does not handle automatic type by default.
to add it you have to:
- add type to template:  'foreach($type$ $id$ in $collection$)'
- set function parameter for type to: GetComponentTypeOf("collection")

I'm not sure where to add this in the MD/src tree, changes to 'MonoDevelop-templates.xml' does not seem to have effect.

*Details:*
IType.GetAllBaseType() extention method returns 'System.Collections ' prior to 'System.Collections.Generic'
which causes fetching Generic Collection parameter's type to fail in CodeTemplate mechanic.

	modified:         core/MonoDevelop.Ide/MonoDevelop.Ide.CodeTemplates/ExpansionObject.cs


mhutch:
whitelist

slluis:
Thanks!

PR #847, opened 2015-04-04
T4 - use lowercased template directive names

mnaoumov:
Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=28767

PR #854, opened 2015-04-06
T4 - use lowercased template directive names

mnaoumov:
Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=28767

Recreated #847 to pull request on top of master

### DIFF
--- a/main/src/addins/TextTemplating/Mono.TextTemplating/Mono.TextTemplating/ParsedTemplate.cs
+++ b/main/src/addins/TextTemplating/Mono.TextTemplating/Mono.TextTemplating/ParsedTemplate.cs
@@ -263,7 +263,7 @@ namespace Mono.TextTemplating
 	{
 		public Directive (string name, Location start)
 		{
-			this.Name = name;
+			this.Name = name.ToLowerInvariant ();
 			this.Attributes = new Dictionary<string, string> (StringComparer.OrdinalIgnoreCase);
 			this.StartLocation = start;
 		}


mhutch:
IMO this would be cleaner to do where we resolve the directive names? That way the parsed document would still reflect the literal values of the document.

i.e. `switch (dt.Name.ToLowerInvariant ()) {`

https://github.com/mono/monodevelop/blob/master/main/src/addins/TextTemplating/Mono.TextTemplating/Mono.TextTemplating/TemplatingEngine.cs#L200

mnaoumov:
@mhutch, if we follow the approach you suggested we'll have to modify in many other places where **Directive.Name** are used, e.g. https://github.com/mono/monodevelop/blob/master/main/src/addins/TextTemplating/Mono.TextTemplating/Mono.TextTemplating/ParsedTemplate.cs#L135, https://github.com/mono/monodevelop/blob/master/main/src/addins/TextTemplating/Mono.TextTemplating/Mono.TextTemplating/ParsedTemplate.cs#L156, https://github.com/mono/monodevelop/blob/master/main/src/addins/TextTemplating/Mono.TextTemplating/Mono.TextTemplating/TemplatingEngine.cs#L401 etc,  So the implementation I suggested seems to be the easiest one.

mhutch:
I just don't like the idea that the AST won't match the actual text. This mean e.g. if we show a tooltip, the case will be wrong.

mhutch:
I've implemented my version in https://github.com/mono/monodevelop/commit/3f673215d714d480084a971957e3c60bdb07f339

PR #855, opened 2015-04-06
Add catch and output for all exceptions

mnaoumov:
Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=28768

Recreated #848 to pull request on top of master


mhutch:
Merged, thanks!

In future please begin tag your commit message with the area it affects, e.g. `[T4] Add catch and output for all exceptions`. And note that the MonoDevelop coding style places a space before the `(` in method calls.

PR #849, opened 2015-04-04
Append dll for assembly reference if needed

mnaoumov:
Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=28773

PR #856, opened 2015-04-06
Append dll for assembly if needed

mnaoumov:
Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=28773

Recreates #849 to pull request on top of master

### DIFF
--- a/main/src/addins/TextTemplating/Mono.TextTemplating/Mono.TextTemplating/TemplateGenerator.cs
+++ b/main/src/addins/TextTemplating/Mono.TextTemplating/Mono.TextTemplating/TemplateGenerator.cs
@@ -204,6 +204,9 @@ namespace Mono.TextTemplating
  				if (System.IO.File.Exists (path))
  					return path;
  			}
+
+			if (!assemblyReference.EndsWith (".dll", StringComparison.OrdinalIgnoreCase) && !assemblyReference.ToLowerInvariant ().EndsWith (".exe", StringComparison.OrdinalIgnoreCase))
+				return assemblyReference + ".dll";
 			return assemblyReference;
 		}
 		
--- a/main/src/addins/TextTemplating/Mono.TextTemplating/Mono.TextTemplating/TemplateGenerator.cs
+++ b/main/src/addins/TextTemplating/Mono.TextTemplating/Mono.TextTemplating/TemplateGenerator.cs
@@ -28,6 +28,7 @@ using System;
 using System.Collections.Generic;
 using System.CodeDom.Compiler;
 using System.IO;
+using System.Reflection;
 using System.Text;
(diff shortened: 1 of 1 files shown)


mhutch:
According to MSDN, this can be an assembly name or a filename. In the case of an assembly name, appending ".dll" would be incorrect.

https://msdn.microsoft.com/en-us/library/gg586946.aspx

I would suggest only appending .dll for the file search, and extending the file search to search the input file's directory. If the search fails, return the unmodified string.

mnaoumov:
@mhutch, as far as I see the file search code already checks for file existence https://github.com/mnaoumov/monodevelop/blob/f30e033ef1779fd6fe0a08ad2b60b3f7d41dbfb9/main/src/addins/TextTemplating/Mono.TextTemplating/Mono.TextTemplating/TemplateGenerator.cs#L205 and I suggest to append **.dll** if the file couldn't be found and is not **.dll** or **.exe**

I decompiled Microsoft's **TextTransform.exe** implementation and the trick they are doing instead of the one I suggested

      string location = GlobalAssemblyCacheHelper.GetLocation(assemblyReference);
      if (!string.IsNullOrEmpty(location))
        return location;
      return assemblyReference;

And their **GlobalAssemblyCacheHelper** is a black magic. It is using P/Invoke to *fusion.dll* to get a list of GAC assemblies.

I don't think it is acceptable in Mono so I don't see any other options as to the one I suggested.

Well, there is one approach that also works but I don't like it because of multiple reasons

    Assembly assembly = null;
    try
    {
        assembly = Assembly.LoadWIthPartialName(referenceAssembly);
    }
    catch
    {
        // ignore all errors
    }

    if (assembly != null)
        return assembly.Location;

I don't like this solution because of two main reasons.

1. Assembly.LoadWithPartialName is obsoleted since .NET 2.0
2. I don't like loading assembly just to find its location and then keep in AppDomain forever... Of course, we could load assembly into a separate helper AppDomain but this seems to be way too much for such a simple problem.

So adding *.dll* as a suffix seems to be the easiest and shortest solution that just works

mhutch:
Since this is a cleanly reverse engineered project, I can't accept PRs from people who've viewed decompiled source from the MS implementation :(

mhutch:
Note that another option would be to parse the assembly name with the AssemblyName class, and only append .dll if there is no Version component.

mnaoumov:
@mhutch , I've implemented a check that you suggested. You can accept my PR now as I erased my memory and forgot everything I saw in Microsoft's implementation :)

mhutch:
Well... black-box reverse engineering doesn't really work like that, unfortunately. Once you've seen it, it can't be "unseen".

But since your commit implements exactly what I suggested, which was not based on MS code, I think I can merge this as-is.

Thanks!

PR #858, opened 2015-04-07
5.9 toolbar fixes

Therzok:


Therzok:
f304d927c1c42b0455512fbc74c0d67eba4a24e6 is part of the fix for https://bugzilla.xamarin.com/show_bug.cgi?id=28730

abock:
This should fix [bxc#28829](https://bugzilla.xamarin.com/show_bug.cgi?id=28829), which prevents XS from starting up at all.

PR #860, opened 2015-04-07
MacPlatform: bxc#28829: fix crash in toolbar

abock:
When `Window.Screen` is null, which appears possible initially when running with multiple screens, skip the `RealignTexts` logic.

Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=28829.


Therzok:
I already submitted a PR

Therzok:
https://github.com/mono/monodevelop/pull/858

PR #861, opened 2015-04-08
[Mac] Fix build with Xamarin.Mac 2.0

mhutch:


alanmcgovern:
I presume our bots will have to be updated to xammac 2.x before this can be merged, which means md-addins will need porting and XamarinVS will have to test all their stuff to make sure it still works with the newer xammac?

jstedfast:
I think it would actually be fine, because the APIs that take CGRect exist in Xamarin.Mac 1.12 afaik, it's just that 1.12 *also* has compatibility APIs to massage RectangleF into CGRect.

I could be wrong though... @chamons ?

mhutch:
The buildbot seems happy with it.

mhutch:
Though I guess it's only building Windows?

alanmcgovern:
Yeah, it only builds linux afaik

chamons:
:+1:

PR #863, opened 2015-04-08
[Core] Remove SupportFeature as the new project dialog handles this diff...

Therzok:
...erently.

The widgets are not deleted, in case someone wants to port them to New Project Dialog wizards.

I could delete them, and whoever wants to make a wizard can git log on the directory.


Therzok:
Manually merged in master.

PR #865, opened 2015-04-09
[Ide] Handle native menus from ExtensibleTreeView's context menu

bratsche:
This adds an optional Action to handle when the context menu is closed so that the treeview can maintain the 'pushed' state on the gear button.

Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=26957


bratsche:
My original idea was to change `CommandManager` and `ContextMenu` around so that we could do:

```csharp
// Create ContextMenu instead of Gtk.Menu
ContextMenu menu = IdeApp.CommandService.CreateContextMenu (entryset, this);
menu.Hidden += HandleMenuHidden;
menu.Show();
```

And then let `ContextMenu` internally handle building a native or GTK menu, and have a `Hidden` event on `ContextMenu`.

This is a pretty big refactoring though, so instead I'm proposing this PR instead where we pass an Action in when we want to be notified that the menu is hiding. I'd like to come back to the refactoring later though.

bratsche:
Actually, better would be just to do `ContextMenu menu = new ContextMenu (entryset, this);`

slluis:
I think we should just implement your original idea.

PR #866, opened 2015-04-10
Bug 28856 - [Windows] Unable to open .sln file or .csproj file when we r...

Therzok:
...ight click on it and select open with XS


knocte:
If you put a "ResolveFullPath()" method in FilePath class, I bet the overall API would look confusing because there's already a "FullPath" property in it.

Therzok:
This will be changed. I put it up for review.

knocte:
Even if you drop ResolveFullPath() from FileService, FilePath would still have both `FullPath` property and `ResolveFullPath()` method. What's the difference between them? (A consumer of the API would wonder...)

Therzok:
Check the naming. It's not ResolveLinks (), so that means it'll resolve symlinks.

knocte:
Sorry, don't quite understand your last message.

Therzok:
`ResolveFullPath` has been renamed to `ResolveLinks`.

knocte:
Oh I see. No concerns from now then. Thanks!

Therzok:
Merged manually.

PR #867, opened 2015-04-10
Bug 28856 - [Windows] Unable to open .sln file or .csproj file when we r...

Therzok:
...ight click on it and select open with XS


knocte:
The commit message explains the **what**, but not the **why**, and since the bug is not public, I cannot really see what the problem is?

Therzok:
Parametrized opening of MonoDevelop was crashing with Addin Engine not being initialized.

i.e. right click .cs file on Windows -> Open With or mono MonoDevelop.exe <myfile.cs>

Therzok:
I'm updating the commit message on the other pull request.

PR #869, opened 2015-04-10
Bug 28983 - Xamarin Studio crash on when trying to open in full screen m...

Therzok:
...ode.


knocte:
Sorry if it sounds picky but the commit message doesn't respect the pattern that most commit messages follow, which is: start with the main area of what you are changing, in square brackets, then a small summary, and then from the third line give a more detailed description (if needed), where you can also mention the bug number and/or bug URL.

So instead of:

`Bug 28983 - Xamarin Studio crash on when trying to open in full screen mode.`

It should probably be:

```
[Mac] Fix crash when opening full screen mode

Skip updates with a null Window. Only crashes on Mavericks.
The View gets updated once again when the window resize finishes.

Fixes: https://bugzilla.xamarin.com/show_bug.cgi?id=28983
```

Therzok:
I will change the commit message in the PR targeting master.

These needed to go in 5.9 fast, so commit messaging wasn't really important.

PR #870, opened 2015-04-12
Fixed 'Bug 28691- [SourceEditor] Minimum value for code completion rows'

gitexperience:
https://bugzilla.xamarin.com/show_bug.cgi?id=28691
Minimum value for code completion rows is set to 5.


mkrueger:
Thanks - but we decided to remove that option - it doesn't really make sense.

PR #872, opened 2015-04-13
5.9 - Fix New Project dialog closing when auto-hide window closes

mrward:
Fixed bug #28971 - New project window disappears if pads
(e.g. Application output, solution window) set to auto hide
https://bugzilla.xamarin.com/show_bug.cgi?id=28971

Simplest way to reproduce the problem is to configure the
solution window to be auto-hide and then right click the solution
and select Add - Add New Project. The New Project dialog will
open and then as the Solution window hides the New Project dialog
will be closed. This happens on the Mac but not on Windows.


slluis:
whitelist

PR #875, opened 2015-04-14
Test framework

iainx:
The basics of the test framework:

Currently supports
Operations: Type, Property, Model, Marked, Text, NextSiblings, Index
Actions: Select, Click, TypeKey, Toggle

Moved waiting for timers/counters into the MD process, rather than creating a proxy MarshalByRef remote object for the counter.

Added an attribute that can be used to add metadata about GtkTreeModels. It can either be added to a class, or to an instance.

Still needs done:
- Cocoa support
- Need a better solution for AutoTestSession.IsBuildSuccessful. It seems too specific to be in the AutoTestSession. Maybe a more generic way to check the error count?
- The use of GetGlobalValue seems wrong. We need to work out a less generic "access anything you want" way.


main/tests/UserInterfaceTests/NewProjectController.cs
```diff
@@ -40,43 +40,42 @@ public class NewProjectController
...
-			Session.WaitForWindow ("MonoDevelop.Ide.Projects.GtkNewProjectDialogBackend");
-			Thread.Sleep (delay);
+			Session.WaitForElement (c => c.Window ().Marked ("MonoDevelop.Ide.Projects.GtkNewProjectDialogBackend"));
 		}
 
 		public bool SelectTemplateType (string type, string category)
 		{
-			return Session.SelectTreeviewItem ("templateCategoriesTreeView", type, category);
+			return Session.SelectElement (c => c.TreeView ().Marked ("templateCategoriesTreeView").Model ("templateCategoriesListStore__Name").Text (type));
 		}
```
manish: This would not work for non-MD templates.

Can you update this to use category too such that if we have

* Cross-platform
  * App
  * Tests
* Mac
  * App
  * Tests

and type is "App" and category is "Mac", then this would select "App" under "Cross-platform" instead of under "Mac"

If you have 
```
[Test]
public void TestAndroidUITest ()
{
    CreateBuildProject ("DroidApp", "Android App", "App", "Android", EmptyAction);
}
```

then it fails at the line 103 of `SelectTemplate`
```
Assert.IsTrue (newProject.SelectTemplate (kind));
```

since `newProject.SelectTemplateType (category, categoryRoot)` selected `App` under different category

Therzok:
whitelist

slluis:
Operation classes are placed in the wrong directory. Instead of:

MonoDevelop.Components.AutoTest\MonoDevelop.Components.AutoTest.Operations\Operation.cs

It should be:

MonoDevelop.Components.AutoTest.Operations\Operation.cs

main/tests/UserInterfaceTests/NewProjectController.cs
iainx: Fixed this, although I'm not entirely happy with the method.

The problem is that in the model the category name is marked up as <span font_weight='bold'>Mac</span>, so simply doing a Text operation with "Mac" fails and I don't want to require the test writer to know how the markup is written.

I've added a Contains operation, which is similar to Text but does not require an exact match.

We'll need to revisit this after I get back, but this should be enough to get it working in the meantime.

iainx:
@slluis I've moved the Operations and Results subdirectories.

slluis:
Looks good.

PR #877, opened 2015-04-15
[VersionControl] Fix for interleaved repositories, so we don't always ch...

Therzok:
...oose Subversion over other vcs.

There's no Unit Test for this as the framework does not support it.


slluis:
LGTM

PR #878, opened 2015-04-15
[Ide] check for invalid characters for filename

gitexperience:
Provides creation of filename with only valid characters.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileDescriptionTemplate.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileDescriptionTemplate.cs
@@ -28,6 +28,7 @@
 
 using System;
 using System.IO;
+using System.Linq;
 using System.Xml;
 using System.Collections.Generic;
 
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileDescriptionTemplate.cs
```diff
@@ -77,9 +77,11 @@ static void OnExtensionChanged (object s, ExtensionNodeEventArgs args)
 
 		internal string CreateCondition { get; private set; }
 		
+		static readonly char [] InvalidFileNameCharacters = "&<*;?>%:#|".ToCharArray();
+		
 		public virtual bool IsValidName (string name, string language)
 		{
-			return name.Length > 0 && name.IndexOfAny (Path.GetInvalidFileNameChars ()) == -1;
```
Therzok: Can you point out which of the invalid chars are extra to the path given ones?

See this for instance, you're missing out lots of chars from here. :)

https://github.com/mono/mono/blob/master/mcs/class/corlib/System.IO/Path.cs#L542-L553
https://github.com/mono/referencesource/blob/mono/mscorlib/system/io/path.cs#L88

Therzok: If there's anything extra to the ones in there, you can:

`Path.GetInvalidFileNameChars ().Concat("mycharshere").ToArray()` instead.

gitexperience:
Excuse me for that blunder, here is an updated patch.

PR #881, opened 2015-04-16
Remove check for UserName equal to IntPtr.Zero

Indomitable:
Hi,
    I have a problem with PasswordService. I can add a password but can not retrieve it. Here is my pull request to fix the problem. 
More info: I'm using Windows 8.1 and when I use the method PasswordService.AddWebPassword(url, password); MonoDevelop creates credentials with empty UserName. When I try to get back the password it is empty string this is because ncred.UserName is equal to IntPtr.Zero.


Therzok:
Thanks!

PR #882, opened 2015-04-16
Call IdeApp.Exit() instead of letting Process to kill it

manish:
* GetCommandArrayInfoValues (Command) returns the arrayInfo values
    of a command as a List<object>
* Added support to pass the dataItem argument with ExecuteCommand
* Ide.GetSolutionConfigurations () to get list of configurations
    available for a solution created by the template

---

Call IdeApp.Exit() instead of letting Process to kill it

    When IdeApp.Exit () is called, MonoDevelop goes through
    proper code paths to properly shut down itself. Going
    through Process.Kill () is not the correct method since
    it does not hit the correct code paths. There were instances
    where csproxy was not exiting and ended up holding a lot of
    memory thereby hitting 15.96/16GB of memory.

    Thanks to Alan for tracking down the issue

---

Update timeouts

    * Package update timeout is now 120s to account for network slowness
    * Build timeout is at 60s to account for complex solutions

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestClientSession.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestClientSession.cs
@@ -127,9 +127,9 @@ namespace MonoDevelop.Components.AutoTest
 			session.ExitApp ();
 		}
 
-		public void ExecuteCommand (object cmd)
+		public void ExecuteCommand (object cmd, object dataItem = null)
 		{
-			session.ExecuteCommand (cmd);
+			session.ExecuteCommand (cmd, dataItem);
 		}
 
 		/*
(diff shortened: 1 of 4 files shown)


manish:
adding @slluis and @Therzok

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestClientSession.cs
```diff
@@ -114,12 +114,27 @@ public void Stop ()
 			if (service != null)
 				service.DetachClient (this);
 			else
+				try {
 				process.Kill ();
+			} catch (InvalidOperationException invalidExp) {
```
Therzok: Indent screwed up here. :P

Therzok:
Otherwise, looks good.

slluis:
I don't think this approach is correct. UI tests should "test" the UI, that is, they should do verifications on what is being shown to the user. A call like GetCommandArrayInfoValues() is using the internal api to get a list of configurations, but there is no guarantee that those configurations are being correctly shown. The test should check the list of configurations that is shown in the main menu (Project -> Active configuration) and on the toolbar (although that may need more work since it is a custom widget).

manish:
Would it be possible to cherry-pick merge f0930a1 and e8c9eb7 commits in this case?

manish:
To get a list of configuration, i'll probably goto Solution Options > Configurations and fetch the list shown there.

slluis:
LGTM

PR #883, opened 2015-04-16
Create a new test target named uitest to run UITests

manish:
Instead of updating the existing `test` make target, created a new `uitest` make target which can be used to run UserInterfaceTests

### DIFF
--- a/main/tests/Makefile.am
+++ b/main/tests/Makefile.am
@@ -81,6 +81,11 @@ uitest:
 		done; \
 	fi
 
+	@if ! test -n "$(assembly)"; then \
+		echo "You must pass 'assembly' argument"; \
+		exit 1; \
+	fi
+
 coverage:
 	if ! test -n "$(assembly)"; then \
 		for asm in $(TEST_ASSEMBLIES); do \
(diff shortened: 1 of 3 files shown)


manish:
cc: @slluis @Therzok

main/tests/Makefile.am
```diff
@@ -69,6 +72,15 @@ test:
...
 	fi
 
+uitest:
+	@if test -n "$(assembly)"; then \
+		for asm in $(UITEST_ASSEMBLIES); do \
+			if test `basename $$asm` = $(assembly); then \
+				($(RUN_TEST) -xml=TestResult_`basename $$asm`.xml -labels $$asm) || exit $?; \
+			fi; \
+		done; \
+	fi
```
Therzok: It's missing the fallback for running all the UITests.

Therzok: See above code.

manish: The goal isn't to run all the tests at once, but the way it would be run on wrench would be by passing `assembly=UITest.dll` and also using categories so that all the tests can be in one assembly, but each step would run a different category of test.

Therzok: Okay then, you should fail the test if we're passing an invalid assembly. :P

slluis:
LGTM

PR #884, opened 2015-04-17
Bug 29181 - Avoid changing case of project guid in <ProjectReference> when saving.

BeeWarloc:
Description from https://bugzilla.xamarin.com/show_bug.cgi?id=29181
--

This is not really a big deal, just a trivial annoyance.

When first opening and then saving a (.csproj) project in monodevelop that has
a `<ProjectReference>` the casing of existing guid inside <Project> element will
be forced to upper case.

```
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\ClassLibrary1\ClassLibrary1.csproj">
-      <Project>{f40a92b2-99f3-44e7-a1b8-02584de2bfc7}</Project>
+      <Project>{F40A92B2-99F3-44E7-A1B8-02584DE2BFC7}</Project>
       <Name>ClassLibrary1</Name>
     </ProjectReference>
   </ItemGroup>
```

Visual Studio 2013 saves the guids as lowercase, so fixing this will reduce
noise in source control.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.Formats.MSBuild/MSBuildProject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.Formats.MSBuild/MSBuildProject.cs
@@ -809,11 +809,11 @@ namespace MonoDevelop.Projects.Formats.MSBuild
 			SetMetadata (name, value ? "True" : "False");
 		}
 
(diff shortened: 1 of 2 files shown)


slluis:
This is already fixed in the new-project-model branch, which does extensive changes in the project model. It will be merged to master in the next weeks. Thanks anyway!

PR #886, opened 2015-04-17
[UITest] Add WaitForElement before accessing it

manish:
It is important to wait for an Element before accessing it. I had to add this because I noticed that when calling `SelectTemplateType`+`SelectTemplate` in quick succession, the widgets end up being accessed  even before the widgets are drawn and ready for interaction.


manish:
cc @Therzok @slluis

Therzok:
LGTM.

slluis:
LGTM, although we may be able to fix this in a more generic way, maybe by running an additional UI loop iteration before returning a remote call.

manish:
I am not sure what you mean by "maybe by running an additional UI loop iteration".

PR #887, opened 2015-04-18
[Ide] Configure 'Rename' operation

gitexperience:
'Rename' operation is not working properly and it shows 'invalid characters error' all the time , this patch fixes this.


Therzok:
Could you tell me what was the error? Which of the chars was invalid?

gitexperience:
It was taking every changed character as an invalid character and hence throw a warning . I think it was probably because newPath has whole path of the file that may contain special characters whereas newName has just the name of the file.So, i changed it to newName , and its done!! :)

Therzok:
Okay, makes sense, apparently this is a windows only thing with FullPaths containing : and unix paths containing '/', but I wonder why the semantics changed...

gitexperience:
One more thing i have just encountered is that ,files and folders names can contain an empty period which should not be allowed ,i have seen this feature in visual studio. So ,prepared a patch to fix this and just submitted. .

Therzok:
The breakage was a mistake on my behalf. I fixed it properly now.

PR #888, opened 2015-04-18
[Ide] Prevent creation of files and folders names with empty period

gitexperience:
Though this feature is present when you create a new file , but is absent when you rename it , this patch fixes it.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
@@ -193,7 +193,7 @@ namespace MonoDevelop.Ide.Gui.Pads.ProjectPad
 				} else if (newProjectFile != null && newProjectFile != file) {
 					// If there is already a file under the newPath which is *different*, then throw an exception
 					MessageService.ShowWarning (GettextCatalog.GetString ("File or directory name is already in use. Please choose a different one."));
-				} else if (newPath.FileNameWithoutExtension =="") {
+				} else if (newPath.FileNameWithoutExtension == "" ) {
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
```diff
@@ -193,6 +193,8 @@ public override void RenameItem (string newName)
 				} else if (newProjectFile != null && newProjectFile != file) {
 					// If there is already a file under the newPath which is *different*, then throw an exception
 					MessageService.ShowWarning (GettextCatalog.GetString ("File or directory name is already in use. Please choose a different one."));
+				} else if (newPath.FileNameWithoutExtension == "" ) {
```
Therzok: This is wrong. These files are valid entries, on both Windows and Linux, yet Explorer won't let you do that on Windows.

gitexperience:
I have seen that feature in visual studio, so i feel like to include it in XS too , but if empty periods are valid then i think i need to close that PR.. :)

Therzok:
You don't test empty periods like that. "." and ".." are invalid (I think), so I think the fix is different.

gitexperience:
Yes, it is showing error for filenames like "." and ".." but not for ".cs" ..

gitexperience:
I request you to review this screencast and please give a comment on it ,why is it so? that would help me..  :)
and sorry for that "empty period" , it was actually "leading period" above.
http://screencast.com/t/kRXSm0kRy

Therzok:
Yes, but those files are perfectly valid. I'd rather not have platform specific code in action. ".gitignore" is an example of a file that is perfectly valid to create, and maybe I'd want to create it from the IDE?

FWIW, I can do this:

Unix
```
touch .asdf
```

PowerShell
```
ni .asdf -type file
```

Batch
```
copy NUL .asdf
```

So this is something I'd rather not have changed. Leading dot files are really important, and VS 2015 even writes to `.vs` for storing its per-solution preferences.

PR #896, opened 2015-04-23
Update SqlQueryView.cs

elsupergomez:
This prevent a bug that position the second character typed at first of the line on new query because reload the mime type.
Initialize the Document.Text property set the mime type before any character typed.


elsupergomez:
Thanks!

Therzok:
:+1:

PR #900, opened 2015-05-03
[Build] Don't fail tests for windows if dll is not found.

Therzok:
Matches Mac and Linux behaviour from makefiles.

### DIFF
--- a/main/wintest.sh
+++ b/main/wintest.sh
@@ -10,6 +10,8 @@ else
 			arg="build/tests/$arg"
 		fi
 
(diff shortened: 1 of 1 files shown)


alanmcgovern:
We should probably change this logic to echo out a message to @MonkeyWrenh if the tests aren't run. This way the message will show up on the build page

alanmcgovern:
We should also change the make file logic to do the same :)

Therzok:
I can surely do the logic in master, not sure if we need that for 5.9. I just want the tests green.

PR #901, opened 2015-05-05
Add screenshots to the unit tests for Mac platform

manish:
Adds support for Mac screenshots using native `Foundation` and `AppKit` support. There are instances where the tests are failing in XS, so having screenshots is the only reliable way to track down the issues

Thanks to Chris Hamons for the `DllImport` section.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestSession.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestSession.cs
@@ -104,11 +104,14 @@ namespace MonoDevelop.Components.AutoTest
 
 		public void ExitApp ()
 		{
-			try {
-				IdeApp.Exit ();
-			} catch (Exception e) {
-				Console.WriteLine (e);
-			}
+			Sync (delegate {
+				try {
+					IdeApp.Exit ();
+				} catch (Exception e) {
+					Console.WriteLine (e);
+				}
+				return true;
+			});
 		}
 
 		public object GlobalInvoke (string name, object[] args)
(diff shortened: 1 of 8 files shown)


slluis:
Looks good to me, although you should not include unrelated changes into the same commit.

manish:
Lluis, the IdeApp.Exit was needed along with this because XS was crashing once in a while during trying out screenshots. WRT to other commit, all of it is related to the screenshots.

iainx:
Nice :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestSession.cs
```diff
@@ -38,6 +38,12 @@ namespace MonoDevelop.Components.AutoTest
 {
 	public class AutoTestSession: MarshalByRefObject
 	{		
+		[System.Runtime.InteropServices.DllImport ("/System/Library/Frameworks/QuartzCore.framework/QuartzCore")]
+		static extern IntPtr CGDisplayCreateImage (int displayID);
```
Therzok: Copied from comment on wrong commit:

Not really important, but both `int displayID` and `int MainDisplayID` should be in fact `uint`s.

Therzok: Also, isn't the desired behaviour to grab screenshots of all screens on which XS root windows exist?

iainx: Well, my take is that as this is for the testing and not a general purpose screenshot facility, we can control the screens the root windows exist on and make sure they only appear on one?

Therzok: My intention was that if we ever get to the point that we want to test tabs in extended displays (aka grab a editor and move it to another screen), we'd need this functionality. So that's why I was thinking of using ListTopLevels() from XS to get all the screens it's got windows on and screenshot them.

PR #905, opened 2015-05-12
[AutoTest] Add support for ComboBoxes.

iainx:
If the combobox has been created with Gtk.ComboBox.NewText or the Gtk.TreeModel has a string as its first column then the Text operator will operate on it directly, using that first column.
If the combobox has been created by specifying a different type of model, or the required string is not the first column, then the Model operator will have to be used, and the combobox's Gtk.TreeModel will have to have the SemanticModelAttribute added to it.


iainx:
@manish Does this help?

manish:
:+1: 

works with ComboBox not explicitly created using a Model.

PR #908, opened 2015-05-18
[Ide] Don't call Present() on the window here

bratsche:
Calling Present() causes windows that have been snapped to the edge on Win32 to revert to their normal size.

Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=22570

@mhutch thought the reason for this `Present()` was maybe to focus XS in the event of double-clicking a .cs file in Windows Explorer, but after testing it I found it didn't actually do that. I tried replacing it with `DesktopService.GrabDesktopFocus (window)` but that also caused window snapping to break.

Does anyone know of anything else that might be affected by this change?


akoeplinger:
https://github.com/mono/monodevelop/commit/962d0619b07f898f565290e9203a389d3a80dadc suggests it has something to do with floating windows?

PR #909, opened 2015-05-19
[AutoTest] Add support for passing CommandSource to CommandManager.DispatchCommand

manish:
During writing UITest it is important to more closely imitate the working of the UI. `CommandManager.DispatchCommand` takes another argument named `CommandSource source` which tells from the source of the request like `MainMenu` or `ContextMenu`. This is important as some commands can only be called from one source and not from another


manish:
Review requested: @iainx

PR #917, opened 2015-05-27
Revert "[AspNet] Load bundled ASP.NET MVC assemblies"

obadz:
This reverts commit 12d610fb3f6dce121df538e36f21d8c2eeb0a6e3.

This commit broke xbuild's ability to find DLLs using pkg-config.

### DIFF
--- a/main/src/addins/AspNet/MonoDevelop.AspNet.csproj
+++ b/main/src/addins/AspNet/MonoDevelop.AspNet.csproj
@@ -452,34 +452,6 @@
     <None Include="Templates\Projects\WebApplication.xpt.xml">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
-    <None Include="..\..\..\packages\Microsoft.AspNet.Mvc.5.2.2\lib\net45\System.Web.Mvc.dll">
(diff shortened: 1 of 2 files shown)


obadz:
Cc @mhutch

mhutch:
Can you explain why this change is necessary?

None of these assemblies are available from pkg-config, and we need them.

obadz:
If you write <Reference Include="System.Blah"><HintPath>../packages/blah/System.blah.dll</ ..>, xbuild takes the liberty to look for System.Blah.dll in a bunch of places starting with ../packages/Blah, but then also moving on to PKG_CONFIG_PATH (if you've elected to put it there) and to the Mono framework & gac.

CopyToOutputDirectory does not have that behavior so by using it you are forcing everyone who builds MD to have these DLLs in the packages folder, either by using NuGet or by copying them there by hand.

I'm packaging MonoDevelop for NixOS, and we're trying to build all the dependencies from source so we'd like to have the option not to go to NuGet.

Could you explain why the original change was needed?  If you do use NuGet, xbuild will still pick the dlls up in ../packages but by removing the CopyToOutputDirectory you leave the option to those who do not wish to use NuGet to still build MonoDevelop (as was the case in version 5.7.*).

Thanks

PR #919, opened 2015-05-27
[Text Editor] Add Delete to start of line action

iainx:
Add an action to delete from the caret to the start of the line


iainx:
- Just something that's always annoyed me by its absence :)

PR #922, opened 2015-05-28
[Core] Remove Newtonsoft.Json submodule and replace with nuget package

sgmunn:
See similar PR for md-addins.


alanmcgovern:
merged

PR #928, opened 2015-06-03
Added interpreter path to configure.sh for freebsd builds

PaulGWebster:
Without this freebsd will treat the file as a /bin/sh file


PaulGWebster:
A more portable way would be #!/usr/local/bin/env bash ... Kinda submitted it before thinking about it :)

slluis:
@directhex what's your take on this?

PaulGWebster:
The correct way is #!/usr/bin/env bash 
it turns out

PR #930, opened 2015-06-03
[Test Framework] Export the widget hierarchy via HTTP

iainx:
This might be a bit of a weird one, so I'll defer judgement to @slluis 

Create a web server that listens on port 12698 and exports the widget hierarchy to any client that requests /AutoTest/ to aid in writing tests.

An example of the client is https://github.com/iainx/UIBuddy 

screen shot 2015-06-03 at 12 27 00 pm


iainx:
It's sort of like what Xam.Inspector does, but simpler.

manish:
I guess this can be closed.

PR #931, opened 2015-06-04
[Test Framework] Add a wait parameter to ExecuteOnIdle(AndWait)

iainx:
Sometimes the result of clicking a button is a modal dialog opening. In these cases ExecuteOnIdleAndWait will timeout because button.Click () will not return until after the dialog has been closed. By setting wait = false the idle completion will not be waited on, and the method will return immediately

Fixes XB #30730


manish:
build

directhex:
build

manish:
:+1: 

Works as expected :)

PR #933, opened 2015-06-04
[PropertyPad] Query tooltip on whole row

Therzok:
the previous logic was unintuitive, so now show the tooltip while hovering the whole row
Bug 30545 - Properties Pad could use a tool tip for long property values


Therzok:
cc @sgmunn @slluis

PR #934, opened 2015-06-05
[Coverage] Update the coverage options for the new option names in mono master

iainx:
Replace file with covfilter-file


slluis:
:+1:

PR #940, opened 2015-06-09
[Test Framework] Delay the test framework startup until MD has finish…

iainx:
…ed starting

When the AutoTestSession starts early in the startup process, there is a possible race condition if the test that is running attempts to do something before Monodevelop has finished starting. This patch starts AutoTestSession as the final part of the start up process


manish:
There are indeed a few failures
https://wrench.internalx.com/Wrench/ViewLane.aspx?lane_id=1902&host_id=233&revision_id=556631

`SetUp : System.Exception : Could not connect to application`

iainx:
@manish Interesting, it seems delaying the startup is now hitting the 15second timeout. We should increase that and see if it improves

PR #941, opened 2015-06-10
Reset listen_socket if connecting to existing endpoint fails.

lukaszunity:
Fixes issue with a new window always being created on Windows if the first instance of MonoDevelop was started with an open file argument.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
@@ -197,6 +197,15 @@ namespace MonoDevelop.Ide
 					// Reset the socket
 					if (null != socket_filename && File.Exists (socket_filename))
 						File.Delete (socket_filename);
+
+					if (options.IpcTcp) {
+						try {
+							listen_socket.Close();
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
```diff
@@ -197,6 +197,15 @@ int Run (MonoDevelopOptions options)
 					// Reset the socket
 					if (null != socket_filename && File.Exists (socket_filename))
 						File.Delete (socket_filename);
+
+					if (options.IpcTcp) {
```
slluis: Why only in the IpcTcp case?

lukaszunity: This issue only happens on Windows when IpcTcp is used, so I assume that the lines above that reset it correctly for when socket_filename is used on OSX.

lukaszunity: Would you prefer another solution to this?

lukaszunity:
This solution is incomplete, I will open another PR with all the necessary changes required to make this work.

PR #942, opened 2015-06-12
bump cecil - Bug 30463

DavidKarlas:
Bug 30463 - Debugger operation failed. Field not found: Mono.CompilerServices.SymbolWriter.LineNumberEntry.Column

Problem is that on .Net if GAC and local dll are both signed and have same version, .Net picks up GAC version instead of local... Bump is adding just https://github.com/mono/cecil/commit/98ad03e858675862bc4985950f7590c66bd9c3bd commit which changes version(to something random). I also created lanes and tested on Windows and Mac...

When Roslyn branch is merged... We should consider switching to NuGet.


alanmcgovern:
We can't nuget cecil because we custom patch it

DavidKarlas:
I checked commits on this branch and all are already merged into upstream master(aka. NuGet). But I guess question is... do we even want to lose ability to change Cecil rapidly if needed(via git submodule) just to have NuGet and save few seconds on compile time?

btw.. this conversation is unrelated to this PR :)(I know, I know, I started it)

jbevain:
@alanmcgovern I've been reading over the custom patches you have in [mono/cecil](https://github.com/mono/cecil).

I'm all for closing the gap and having upstream provide you with what you need.

The only remaining item I've identified is:

* https://github.com/mono/cecil/commit/a868b2c94e39c958e01eb5f55323cfeda031f946 I'm not quite sure what this is or how it used, but I'll gladly discuss how to provide the functionality you need.

Let me know if I'm missing anything.

DavidKarlas:
MD submodule is pointing to different branch that doesn't include that patch. And I'm pretty sure NuGet binary would be fine for MD needs...

So it's more question of flexibility to bugfix something if we have git submodule vs not compiling Cecil every time MD is built...

PR #944, opened 2015-06-13
[AutoTest] Use Button.Label for GtkWidgetResult's MarkedOperation

manish:
It would be useful to have Gtk.Button's label for filtering in MarkedOperation.


Therzok:
build

Therzok:
Git tests are failing on Linux again, I'll look at it on Monday.

iainx:
It doesn't handle the markup case at the moment. We could probably add something that just dropped the markup tags as they're just <></> tags

manish:
I have not tried but maybe using `Contains ()` should handle such a situation

iainx:
Contains will handle the simple markup cases like <bold>Cancel</bold> but not if there's something like <span>Cancel</span> Operation

Therzok:
build

PR #945, opened 2015-06-14
[AutoTest] Match AppQuery.Text (text) on TreeModel with TreePath

manish:
At the moment there is a small shortcoming with the current API for selecting `TreeView` rows. The only way to select a row if to know the string value of one of the `TreeModel` row.

There are scenarios, where none of the `TreeModel`/`TreeStore` rows are string or we do not know the string value of the row we are going to select. In such a case indexed selection is all we can do. In this patch, we use TreePath value as another way of selecting a row. We can use `0`, `1`, `2` etc to select first, second or third row respectively or even select child rows by using colon separated values like `2:3:1:2` which would select the third row, then it's fourth child row, then it's second child row, then its third child row


iainx:
it would be Model ().Children().Index(2).Child...etc
because Model().Index(2) would be "the second model returned by model", which is meaningless

but yes, this sounds like a better idea

PR #949, opened 2015-06-15
[AutoTest] Return the child TreeIter for TreeView when Children () is called

manish:
If a `Children ()` operation is called on a TreeView, then instead of returning null, return its immediate children, subsequent `Children ()` operation would be needed to child the child of the currently selected TreeIter


manish:
We can close https://github.com/mono/monodevelop/pull/945 now that this is the preferred method

PR #953, opened 2015-06-17
MDEntry widget (native context menu for entry widgets)

bratsche:
This is related to this bug:

https://bugzilla.xamarin.com/show_bug.cgi?id=29751

This is a subclass of GtkEntry that uses our ContextMenu component to give us native menus on OSX.


slluis:
Do we really need a subclass for this? We could have an extension method that subscribes the event and shows the new context menu. Also, we could have a method that iterates over all entries of a window and fixes the context menu for all of them.

bratsche:
See #958

PR #955, opened 2015-06-17
[WIP] Coverity fixes

Therzok:
* [ ] Guard against exceptions in GettextCatalog.Get(Plural)String, so we don't end up with lots of false positives.
* [ ] More to come.

### DIFF
--- a/main/src/addins/NUnit/Gui/NUnitOptionsPanel.cs
+++ b/main/src/addins/NUnit/Gui/NUnitOptionsPanel.cs
@@ -166,7 +166,7 @@ namespace MonoDevelop.NUnit
 		void OnCategoryEdited (object sender, EditedArgs args)
 		{
 			TreeIter iter;
-			if (!store.GetIter (out iter, new TreePath (args.Path)))
+			if (!store.GetIterFromString (out iter, args.Path))
 				return;
 			
 			string old = (string) store.GetValue (iter, 0);
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/RootWorkspace.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/RootWorkspace.cs
@@ -1572,7 +1572,7 @@ namespace MonoDevelop.Ide
 				FileService.NotifyFilesChanged (modified);
 		}
 		
-		void IDisposable.Dispose ()
+		public void Dispose ()
 		{
 			NotifyChanges ();
 		}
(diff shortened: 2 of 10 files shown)


main/src/addins/NUnit/Gui/NUnitOptionsPanel.cs
```diff
@@ -166,7 +166,7 @@ void OnRemoveCategory (object sender, EventArgs args)
 		void OnCategoryEdited (object sender, EditedArgs args)
 		{
 			TreeIter iter;
-			if (!store.GetIter (out iter, new TreePath (args.Path)))
+			if (!store.GetIterFromString (out iter, args.Path))
```
Therzok: This was a GLib Opaque which doesn't seem to have any unmanaged memory reclaimed on finalize.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/RootWorkspace.cs
```diff
@@ -1572,7 +1572,7 @@ public void NotifyChanges ()
 				FileService.NotifyFilesChanged (modified);
 		}
 		
-		void IDisposable.Dispose ()
+		public void Dispose ()
```
slluis: Let's keep this private. That's not something we want to be visible in the public api.

slluis:
Other than the comment, LGTM

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/RootWorkspace.cs
Therzok: But why is it marked as Disposable then? Dispose isn't called, so why not just move it on the Finalizer?

Therzok: Ah, nevermind, the finalizer might not work correctly on Mono (it's not precise gc-ing). why not just drop the Dispose method them? Makes no sense to hide it and have the type as IDisposable.

slluis: RootWorkspace implements IWorkspaceObject, which has IDisposable as base interface, so dispose has to be implemented.

Therzok: But we're in the `FileStatusTracker` class, not RootWorkspace. How is it related?

slluis: Ah, right, forget it then

PR #956, opened 2015-06-18
[IDE] Use NSMenu for Solution Explorer Pad

iainx:
Use NSMenu for the row button in the solution explorer pad instead of a Gtk.Menu

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/MDMenu.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/MDMenu.cs
@@ -42,8 +42,15 @@ namespace MonoDevelop.Components.Mac
 	{
 		static readonly string servicesID = "MonoDevelop.MacIntegration.MacIntegrationCommands.Services";
 
-		public MDMenu (CommandManager manager, CommandEntrySet ces, CommandSource commandSource, object initialCommandTarget)
+		EventHandler CloseHandler;
+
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/MDMenu.cs
```diff
@@ -164,6 +167,14 @@ void MenuNeedsUpdate (NSMenu menu)
 				UpdateCommands ();
 		}
 
+		[ExportAttribute ("menuDidClose:")]
```
Therzok: Nit: Remove Attribute.

iainx: It needs it to make it work, otherwise the objc runtime doesn't know that MenuDidClose is the method that responds to the menuDidClose: selector

Therzok: I meant `[Export ("menuDidClose:")]` :D

slluis:
Adding an optional parameter to a method is an api breaking change. You should add a new method overload.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/MDMenu.cs
iainx: Oh, right, yeah, that makes more sense.
I did it that way to keep in with the other ExportAttribute in the file. I can change it

slluis:
LGTM

PR #958, opened 2015-06-20
Add an extension method to Gtk.Window to have it hijack right-click

bratsche:
events on all Gtk.Entry widgets in its child tree and to manage
the context menu using MonoDevelop's ContextMenu component so we
have native context menus on OSX.

https://bugzilla.xamarin.com/show_bug.cgi?id=29751

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/GtkUtil.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/GtkUtil.cs
@@ -31,6 +31,7 @@ using System.Collections.Generic;
 using Gtk;
 using System.Runtime.InteropServices;
 using Mono.Addins;
+using MonoDevelop.Core;
 
 namespace MonoDevelop.Components
 {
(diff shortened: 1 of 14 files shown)


bratsche:
I only enabled this in one window so far, if it looks good then I'll use it in others as well.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/GtkUtil.cs
```diff
@@ -470,6 +471,155 @@ static Gdk.EventKey CreateKeyEvent (uint keyval, int keyCode, Gdk.ModifierType s
 			IntPtr ptr = GLib.Marshaller.StructureToPtrAlloc (nativeEvent); 
 			return new EventKeyWrapper (ptr);
 		}
+
+		public static IEnumerable<Gtk.Widget> FindAllChildWidgets (this Gtk.Container container)
```
slluis: This method doesn't need to be public.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/GtkUtil.cs
```diff
@@ -470,6 +471,155 @@ static Gdk.EventKey CreateKeyEvent (uint keyval, int keyCode, Gdk.ModifierType s
...
+		{
+			var widgets = new Stack<Widget> (new[] { container });
+
+			while (widgets.Any ()) {
+				var widget = widgets.Pop ();
+				yield return widget;
+
+				if (widget is Gtk.Bin) {
```
slluis: Why special case? a Bin is a Container

slluis:
LGTM

PR #959, opened 2015-06-22
[Assemby Browser] Fix some potential Null dereferences

iainx:
Some fixes for potential null dereferences discovered by Coverity.

The DOM related fixes may have been false positives.


mkrueger:
These things are only null, if the model isn't loaded as cecil model (which is the case in the assembly browser) - but it's ok to make coverty happy ...

iainx:
@mkrueger Thanks. I had a feeling that might have been the case.

PR #967, opened 2015-06-24
[Test Framework] Add ability to flash the AppResults

iainx:


manish:
I have noticed that it colors the whole widget instead of just drawing the outline. Could it be modified such that the flash method just highlights the widget instead of covering it?

new_project_and_monodevelop

iainx:
Sure, I'll do that tomorrow

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest.Results/GtkWidgetResult.cs
```diff
@@ -331,6 +331,43 @@ public override bool Toggle (bool active)
...
+			if (flashState) {
+				return;
+			}
+
+			Cairo.Context cr = Gdk.CairoHelper.Create (resultWidget.GdkWindow);
+			cr.SetSourceRGB (1.0, 0.0, 0.0);
+
+			Gdk.Rectangle allocation = resultWidget.Allocation;
+			Gdk.CairoHelper.Rectangle (cr, allocation);
+			cr.Fill ();
```
Therzok: I think it's s/Fill/Stroke and it's done.

PR #970, opened 2015-06-25
[AutoTest] Use string Type.FullName for ResultType instead of System.Type

manish:
After trying to explore the API, I realized that passing System.Type over remoting would randomly fail. This patch just  changes `AppResult. GetResultType` such that it just returns `System.Type.FullName` instead of `System.Type`


manish:
I tried writing a small tool to check if all the methods are working as expected. Most of them are as of now. 

mainwindow_and_xamarin_studio

PR #971, opened 2015-06-25
[UITest] Update custom user profile env var

decriptor:
MONODEVELOP_PROFILE is what is expected.
main/src/core/MonoDevelop.Core/MonoDevelop.Core/UserProfile.cs#L32

### DIFF
--- a/main/tests/UserInterfaceTests/TestService.cs
+++ b/main/tests/UserInterfaceTests/TestService.cs
@@ -27,6 +27,7 @@
 using System;
 using MonoDevelop.Components.AutoTest;
 using System.Collections.Generic;
+using MonoDevelop.Core;
 
 namespace UserInterfaceTests
 {
@@ -41,7 +42,7 @@ namespace UserInterfaceTests
 			//TODO: support for testing the installed app
 
 			Session.StartApplication (file: monoDevelopBinPath, environment: new Dictionary<string,string> {
-				{ "MONODEVELOP_TEST_PROFILE", profilePath ?? Util.CreateTmpDir ("profile") }
+				{ "MONODEVELOP_PROFILE", profilePath ?? Util.CreateTmpDir ("profile") }
 			});
 
 			Session.SetGlobalValue ("MonoDevelop.Core.Instrumentation.InstrumentationService.Enabled", true);


main/tests/UserInterfaceTests/TestService.cs
```diff
@@ -41,7 +42,7 @@ public static void StartSession (string monoDevelopBinPath = null, string profil
 			//TODO: support for testing the installed app
 
 			Session.StartApplication (file: monoDevelopBinPath, environment: new Dictionary<string,string> {
-				{ "MONODEVELOP_TEST_PROFILE", profilePath ?? Util.CreateTmpDir ("profile") }
+				{ "MONODEVELOP_PROFILE", profilePath ?? Util.CreateTmpDir ("profile") }
```
alanmcgovern: Our makefile sets MONODEVELOP_TEST_PROFILE. Are you sure this should be changed?

mhutch: Nothing reads MONODEVELOP_TEST_PROFILE: https://github.com/mono/monodevelop/search?utf8=%E2%9C%93&q=MONODEVELOP_TEST_PROFILE

Whereas MONODEVELOP_PROFILE actually does something: https://github.com/mono/monodevelop/search?utf8=%E2%9C%93&q=MONODEVELOP_PROFILE

Therzok: @mhutch I stumbled across this gem.

https://github.com/mono/monodevelop/blame/master/main/src/core/MonoDevelop.Startup/MonoDevelop.Startup.csproj#L25

manish: Is a new profile created for every test run?

mhutch: I imagine it would, since it it creates a temp directory.

decriptor: I'm curious. How many things did we "think" were working that were not actually working? Should I grep through the MD code base and change all of the MONODEVELOP_TEST_PROFILE over to MONODEVELOP_PROFILE?

Therzok: I'm not sure. I'd say that given that we use system configs for local builds, people might've got used to it.

Many people use `git clean -xfd` to ensure clean repository state, so that would nuke their local profile. That means any settings you set if you're using local builds to develop will be gone after cleaning the repo.

I, for one, think that this should be set by the user, not by the startup csproj or the makefile.

Thoughts?

decriptor: so far my experience has been that it still uses ~/Library/Preferences..

Therzok: If you _do_ change the other two occurrences to use MONODEVELOP_PROFILE, it _will_ change it to use a local test profile.

manish:
LGTM if it only changes TestService.cs

DavidKarlas:
This broke most UITests on Roslyn branch, MD that is tested doesn't start...

manish:
@kdubau I think we need to revert this commit till we find out a better solution.

Therzok:
Please merge to roslyn after reverting.

manish:
This needs to be reverted first.
https://github.com/mono/monodevelop/commit/adfc823d15583f26e9491dc8e9f5211ef4e9ccca

manish:
Reverted both of them

PR #975, opened 2015-07-09
[Build] Add `git submodule sync` as a pre-build command.

Therzok:


Therzok:
I haven't pushed to master directly because I'd like some input.

Given that we have branches that change submodules remotes, developers usually come asking about issues with submodules updating.

So, I opened this PR which invokes a git submodule sync before updating them. This should fix most issues people have and it _should_ also not last long. `git submodule sync` lasts 0.28s on my Mac machine and 5s on my Windows machine with msysgit v1. I haven't timed it with a more recent master under cygwin/msysgit v2.

alanmcgovern:
Makes sense to do it here

PR #977, opened 2015-07-14
[Ide] Disabling BringToFront for "Application Output" pad

DavidKarlas:
If BringToFront=true and LocalsPad(or any other Pad) shares same space as "Application Output" it's always hidden by "Application Output" when debugging starts which can be very annoying

At first I thought there is bug in Layout switching which forgets which pad was displayed/selected last and only later realised that "Application Output" is always throwing itself in front of LocalsPad or other pads...
So every time that I restart debugging I have to switch back to LocalsPad...

If user likes watching Application Output... no problem he can always switch to it... and it will stay in front in next debug session...(aka same behaviour as now unless he changes to some other pad)...

/cc @slluis

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/ProgressMonitors.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/ProgressMonitors.cs
@@ -77,7 +77,7 @@ namespace MonoDevelop.Ide.Gui
 		
 		public IProgressMonitor GetRunProgressMonitor ()
 		{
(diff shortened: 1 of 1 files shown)


Therzok:
This was a major annoyance to me for a long time. :+1: for this being fixed.

slluis:
Just FYI, in the initial default MD configuration the output pad is minimized at the bottom, to save vertical space. The BringToFront flag was set so that the pad would pop up when the app was started. This made sense since 1) console apps were using the output pad by default, and 2) the debugger was not enabled by default, so there were to debug pads to show. Nowadays we are using an external console for console apps and the debugger is the default execution mode, so there is less benefit in focusing the output pad by default. In any case, we have to be careful when changing the behavior of the IDE. What some users find annoying may be very useful to other users.

PR #978, opened 2015-07-14
[Mac] Guard against not-attached view in NSSavePanel

Therzok:
Bug 30925 - [OSX10.11] 'Open ...' on the landing page and 'File > Open'
crash Xamarin Studio


Therzok:
This is in c5sr4 and the 5.9 branch.

PR #979, opened 2015-07-14
Support packages.ProjectName.config file if present.

ddunkin:
If a file of the format packages.ProjectName.config is present, it should be used instead of packages.config.

Documentation: http://docs.nuget.org/release-notes/nuget-2.8#user-content-individual-packagesconfig-files-for-different-platforms.
Reference implementation: https://github.com/NuGet/NuGet.PackageManagement/blob/3.0-rtm/src/ProjectManagement/Projects/PackagesConfigNuGetProject.cs


mrward:
Thanks. Unfortunately this feature has already been implemented on a different [branch](https://github.com/mono/monodevelop/tree/nuget-packages-config-file-named-after-project).

PR #980, opened 2015-07-14
[Editor] Add a SelectUntilMatchingBrace command.

Therzok:


mkrueger:
Add that to the roslyn branch. Doesn't make sense to add features to master.

Therzok:
It doesn't conflict with master. And it's the same code for both. Why not just merge to master then merge to roslyn?

mkrueger:
It's on the wrong layer. All text editor commands should be done on the IDE layer. Commands written against the mono.texteditor will not work IF we replace the editor.

Therzok:
Closing PR as I filed a bug for this.

PR #981, opened 2015-07-15
[Components] Use the ContextMenu system to build the popup menu for MenuButton

iainx:
There are multiple systems for building menus in MonoDevelop
 - Using the standard Gtk.Menu functions like MenuButton uses
 - MonoDevelop.Components ContextMenu system where menus are described in an object model, and Gtk.Menu and NSMenus can be created from them
 - By using CommandEntrySets, another model that can then generate Gtk.Menu or NSMenu

This patch ports MenuButton to use the MonoDevelop.Components.ContextMenu system and deprecates the old Gtk.Menu methods. This should make it possible to switch to using NSMenu for these popup menus on the Mac


iainx:
(As an aside, I did have it using NSMenu instead of Gtk.Menu, but I wasn't able to get the popup to line up correctly in all cases so I thought I'd just push this anyway)

Therzok:
:+1:

slluis:
:+1:

PR #982, opened 2015-07-16
[Components] Prevent simultaneous command dispatches

iainx:
Add a check that a previously dispatched command has completed before dispatching a new one.

Most commands finish quickly, but those that open modal dialogs do not, and so it was possible to open multiple "Close without saving" dialogs by repeatedly pressing Ctrl-W on a dirty file or to cut/copy/paste while the dialog was open by Ctrl-W and then Ctrl-X

Fixes BXC #29519 - Multiple copies of the same modal dialog can be made visible at the same time

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -333,6 +333,12 @@ namespace MonoDevelop.Components.Commands
 					return null;
 			}
 
+			// If a modal dialog is running then the menus are disabled, even if the commands are not
+			// See MDMenuItem::IsGloballyDisabled
+			if (DesktopService.IsModalDialogRunning ()) {
+				return ev;
+			}
+
 			var gdkev = MonoDevelop.Components.Mac.GtkMacInterop.ConvertKeyEvent (ev);
 			if (gdkev != null) {
 				if (ProcessKeyEvent (gdkev))
@@ -1166,7 +1172,7 @@ namespace MonoDevelop.Components.Commands
 				return false;
 
 			commandId = CommandManager.ToCommandId (commandId);
-			
+
 			List<HandlerCallback> handlers = new List<HandlerCallback> ();
 			ActionCommand cmd = null;
 			try {


Therzok:
Does this affect paste command?

I don't like this fix.

The issue is most likely in this [block](https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs#L312-L342).

The modal dialog (close popup) cannot handle the close command, and it routes back to the IDE. But, changing that behaviour breaks other modal dialogs' logic.

iainx:
It does appear to only happen on Mac, so you could be right with regards to that block, but like you say, changing that would break other things.

It doesn't appear to affect holding down Cmd+V to repeatedly paste things, at least not on OS X. Let me try on Linux.

My thoughts are that it's safer to make command dispatch not be re-entrant but maybe there's somewhere that depends on that behaviour.

slluis:
I also don't like the fix. The bug must be elsewhere. While the message dialog is visible all the menu options are disabled (including the Close File option), so the shortcut should not work.

alanmcgovern:
cmd-enter can be pressed multiple times, which causes multiple "Do you want to build this project before you run it" dialogs to appear if you have the option to 'automatically build before launching' disabled

iainx:
The bug is that on the Mac the `OnNSKeyPress` method is called on keypress regardless of the menu option state and OnNSKeyPress feeds the keypress into the command dispatch system. @Therzok says he did that to fix another bug so removing the OnNSKeyPress will break some other expected behaviour. 

Adding the extra protection around command dispatch rather than just relying on the state of the menu UI seems to be the best compromise.

Other fun things you can do with this bug: edit a file, press cmd+w to get lots of dialogs on screen the  cmd+q, answer the quit dialog and the main window closes but all the modal cmd+w dialogs are left and the app doesn't actually quit

slluis:
If we already have logic that disables the menu options in this context, then the same logic can be used to discard the key press.

iainx:
The logic that disables the menu options is DesktopService.IsModalDialogRunning (), so I use that here.

Therzok:
May I squash and cherry-pick?

alanmcgovern:
Is this ready to merge?

PR #985, opened 2015-07-17
[MacPlatform] Don't use Carbon for application events

iainx:
This patch removes the use of Carbon for ApplicationEvents in MacPlatform

Requires https://github.com/mono/xwt/pull/532 - I'll update this PR if/when that gets merged

[Doesn't remove the use of Carbon in mdmonitor as it uses a symlink to the ApplicationEvents file)


Therzok:
@iainx Xwt PR is merged, what to do about this?

iainx:
The version of Xwt currently used by Monodevelop includes my patch, so I've rebased and this is good to go if desired

PR #987, opened 2015-07-20
[Ide] Attached SemanticModelAttribute with ExtensibleTreeView

manish:
This patch inspecting the TreeView values from the Test Framework


slluis:
How is this model interesting for UI automation? this is an internal implementation detail.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/ExtensibleTreeView.cs
```diff
@@ -156,6 +157,10 @@ public virtual void Initialize (NodeBuilder[] builders, TreePadOption[] options,
 			tree.Model = store;
 			tree.Selection.Mode = Gtk.SelectionMode.Multiple;
 
+			// Set up the list store so the test framework can work out the correct columns
+			var modelAttr = new SemanticModelAttribute ("store__NodeInfo", "store__DataObject", "store__Chain", "store__CheckVisible");
```
Therzok: Isn't it missing one of the bool members?

PR #990, opened 2015-07-28
Logging service extra

Therzok:


Therzok:
The main reason for this change is that usually logs sent by Subversion users do not have the native error code attached, so logs will now have it. Probably a line like "Exception Data:" can be prepended to the part where we log the extra data.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/LoggingService.cs
```diff
@@ -492,20 +493,35 @@ public static void LogFatalError (string messageFormat, params object[] args)
...
-		
+
+		static string FormatExceptionText (string message, Exception ex)
+		{
+			var exceptionText = new StringBuilder ();
+			exceptionText.Append (message);
+			if (ex != null) {
+				exceptionText.AppendLine ();
+				exceptionText.Append (ex);
+				foreach (DictionaryEntry item in ex.Data) {
```
Therzok: Should I prepend "Exception Data:" as a line here?

slluis: Yes, although only if ex.Data is not empty. Also, if ex is null it should return message, no need to create the StringBuilder in this case.

Therzok: Sure, I'll get on to these changes.

slluis:
:+1:

Therzok:
Merged manually.

PR #992, opened 2015-07-29
[NUnit] Do not clear test results when active config changed.

mrward:
Fixed [bug #31272](https://bugzilla.xamarin.com/show_bug.cgi?id=31272) - Test results disappear randomly after running a test.

Only clear the results in the Test Results window when a workspace item is closed or a new test run is started. Previously the Test Results window would be cleared when the active configuration
changed. The active configuration can be changed manually or if the SolutionItem's OnExecutionTargetsChanged method is called and the project does not use Debug as its default target. The Unit Tests window is still refreshed when the active configuration is changed.


Therzok:
:+1:

slluis:
:+1:

PR #993, opened 2015-07-29
Fix 30506

iainx:
Fixes BXC #30506


slluis:
Look good.

iainx:
Fixed the commits without the stetic files being changed

PR #994, opened 2015-07-29
[Ide] Set Gtk.LinkButton's URI handler

manish:
Needed to resolve bug #29481 - "Supported List" should be a hyperlink instead of a button under Target Framework for Mac projects


bratsche:
:+1:

PR #995, opened 2015-07-30
[IDE] Bypass Focus Document if the current document is already focused

iainx:
If we bypass the Focus Document command when the current document is already focused
then the keypress is passed on down the event chain

Would it be better to implement a more generic system with the command manager asking the focused widget if it wants to handle a keypress before passing it to command dispatch, or will just just make the whole system more complicated than it needs to be?

Fixes BXC #29887


slluis:
This is a tricky one. Even if the patch fixes the problem, it is a very specific fix, only for that use case. The real fix would be to give a chance to the widget to handle the key press, but this would make the command manager much more complex, and might have unwanted side effects. In my opinion, if the user redefines a key binding, it should be redefined to all extents, no matter if a widget handles it or not.

PR #996, opened 2015-07-30
[MacPlatform] Don't assume the search bar is the 1st item in overflow

iainx:
Search through the overflow menu items to find the SearchBar rather than assuming it is the first one. Fixes BXC #32002


slluis:
:+1:

Therzok:
This looks good to me. But we _should_ really strive to not have items in the overflow menu. I know it's impossible, but I'm working on something that could possibly improve the situation.

PR #997, opened 2015-07-30
[UITest] Added Git Tag Tests

manish:
* Updated NuGet Add Package button timeout to 30s
* Moved around a few methods into respective regions
* Added SemanticAttribute to Git Tag TreeView


manish:
@Therzok need review because I made some changes to `GitConfigurationDialog` file

PR #998, opened 2015-07-31
Fix 29887 take 2

iainx:
This is the more generic version of https://github.com/mono/monodevelop/pull/995 where the TextEditor has implemented the IKeyHandler interface, which the command manager then checks to see if the text editor is currently overriding the key press.

Fixes the issues where Escape wouldn't close the Insert Extracted Method overlay, or wouldn't close the autocompletion dialog if it was assigned to a command. (Bug 29887)

It also prevents essential keys from being overridden if they are assigned to a command such as alphanumeric keys or Tab, Return or Backspace.


Therzok:
This lgtm.

slluis:
It looks ok to me. @mkrueger can you review?

mkrueger:
I think that shadows a problem somewhere else. It used to work AFAIK. 
What about other components than the text editor ?

PR #706, opened 2015-01-10
[Po] Update Japanese Translation

matyapiro31:
Update Japanese Translation.

PR #999, opened 2015-08-03
Translate until ID 397

matyapiro31:
I was committed Japanese translation at #706 , but it was rejected as it includes incorrect or strange translation.
So I fixed and commit by small bit.


atsushieno:
I will review only if the PRs are made in reviewable size i.e. one single commit contains within certain lines that github can show the diffs inline.

matyapiro31:
The change is because of damn PoEdit...
This mr is failure.

PR #1003, opened 2015-08-17
Updated to run on FreeBSD 10.1

RussellHaley:
Updated the Makefile.am to run mozroots and execute mono using explicit syntax. Also updated the configure.sh to set SHELL env. Not sure how best to do that without breaking the script for Linux.

### DIFF
--- a/main/Makefile.am
+++ b/main/Makefile.am
@@ -20,7 +20,8 @@ clean: clean-local
 clean-local: sln_clean
 
 restore-packages:
-	nuget restore
+	mozroots --import --sync
+	mono /usr/home/rhaley/git/monodevelop/main/external/nuget-binary/NuGet.exe restore
 
 vcrevision:
 	touch vcrevision
--- a/scripts/configure.sh
+++ b/scripts/configure.sh
@@ -1,3 +1,5 @@
+env SHELL=/usr/local/bin/bash
+
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 if [ ! -f $DIR/configure.exe ]; then


main/Makefile.am
```diff
@@ -20,7 +20,8 @@ clean: clean-local
 clean-local: sln_clean
 
 restore-packages:
-	nuget restore
+	mozroots --import --sync
+	mono /usr/home/rhaley/git/monodevelop/main/external/nuget-binary/NuGet.exe restore
```
mhutch: This has a local path, it'll only work on your machine.

main/Makefile.am
```diff
@@ -20,7 +20,8 @@ clean: clean-local
 clean-local: sln_clean
 
 restore-packages:
-	nuget restore
+	mozroots --import --sync
```
mhutch: This should be done when installing the Mono package.

scripts/configure.sh
```diff
@@ -1,3 +1,5 @@
+env SHELL=/usr/local/bin/bash
```
mhutch: Most systems don't have bash installed to /usr/local

main/Makefile.am
RussellHaley: nuget restore wouldn't run until I ran mozroots, which caused the build to fail.

main/Makefile.am
RussellHaley: quite. I'll re-submit.

scripts/configure.sh
RussellHaley: I'll look at having this added to the platform specific Makefile (FreeBSD ports). Is there a cross platform/generic command for ensuring bash is the correct shell?

RussellHaley:
Great, thanks for the input Michael!

PR #1005, opened 2015-08-18
Fixed 'Bug 33065 - Dangerous: refactoring operations drop the null

mkrueger:
propagation operator'.
This adds an option + warning message to the c# projects.
Note that language version can't be checked because the c# binding
references the refactoring addin and the c# language configuration is
defined there.


manish:
build

mkrueger:
+ Removed the 'disable' option. This makes that a lot less complex to test.

PR #1006, opened 2015-08-21
[MacPlatform] Status icons can overlap with the build results

iainx:
When showing a status icon use RepositionStatusLayers instead of repeating the positioning code which didn't handle the case when there was a build result

Fixes BXC #33230


Therzok:
Thanks. :+1:

PR #630, opened 2014-08-17
New color scheme editor

sheff146:
An implementation of the new ColorSchemeEditor via Xwt through participation in Google Summer of Code program. The new editor supports Undo/Redo operations, style tree, color picking, quick style search, includes code samples for C#, HTML, CSS, JavaScript etc.

PR #1008, opened 2015-08-22
New color scheme editor

Therzok:
By hand port of #630


slluis:
Looks good to me, but it needs to go through QA before merging.

Therzok:
Some things are broken in there, such as missing icons (see my last commit) and it does need a QA. Some things were broken when I checked.

mkrueger:
We remove the color scheme editor during the textmate switch.

PR #1010, opened 2015-08-25
Fixed "Bug 33065 - Dangerous: refactoring operations drop the null

mkrueger:
propagation operator"
The default button was wrong because one button got removed. That
caused the dialog not to appear on windows.


slluis:
whitelist

PR #1011, opened 2015-08-25
[Ide] Group close all/other documents into one dialog.

Therzok:
Bug 9804 - Close all but this doesn't offer option to "Save all"
Bug 10134 - Show combined "save/discard" dialog when closing multiple
files

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
@@ -35,6 +35,7 @@ using MonoDevelop.Ide.Gui;
 using MonoDevelop.Ide.Navigation;
 using MonoDevelop.Core;
 using System.Linq;
+using MonoDevelop.Ide.Gui.Dialogs;
 
 namespace MonoDevelop.Ide.Commands
 {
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
```diff
@@ -55,10 +56,20 @@ protected override void Run ()
...
+					dlg.Modal = true;
+					if (MessageService.ShowCustomDialog (dlg) != (int)Gtk.ResponseType.Ok)
+						return;
+				}
+			
+			foreach (Document doc in docs) {
 				var w1 = (SdiWorkspaceWindow) doc.Window;
 				if (w1.TabControl == activeNotebook)
-					doc.Close ();
+					doc.Window.CloseWindow (true);
```
Therzok: doc.Close() still popups up a dirty dialog even though the `DirtyFilesDialog` does `DiscardChanges`, so I guess that's a bug? :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/DirtyFilesDialog.cs
```diff
@@ -18,9 +18,19 @@ internal class DirtyFilesDialog : Gtk.Dialog
...
+
+		public DirtyFilesDialog () : this (true, null)
 		{
+		}
+
+		public DirtyFilesDialog (bool closeWorkspace) : this (closeWorkspace, null)
+		{
+		}
+
+		public DirtyFilesDialog (bool closeWorkspace, IViewContent except) :  base (GettextCatalog.GetString ("Save Files"), IdeApp.Workbench.RootWindow, DialogFlags.Modal)
```
Therzok: I'm dissatisfied with this, but `Document` comparison does not work since `ActiveDocument` is getting wrapped and won't equal the one in the tab ever.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
```diff
@@ -55,10 +56,20 @@ protected override void Run ()
 			if (active == null)
 				return;
 			var activeNotebook = ((SdiWorkspaceWindow)active.Window).TabControl;
-			foreach (Document doc in IdeApp.Workbench.Documents.ToArray ()) {
+			var docs = IdeApp.Workbench.Documents.ToArray ();
+			var showDirtyDialog = docs.Any (doc => doc.IsDirty);
```
slluis: You should only check documents in the active notebook.

slluis:
A few comments:
* In all cases, you should check the dirty flags for the documents we are closing. That is, the documents for the active notebook (and ignoring the active document in the CloseAllButThis case.
* The dialog should only show the documents being closed (same as above)
* That dialog doesn't make sense if there is only one document open.
* Grouping the documents by project doesn't make sense in this context.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
```diff
@@ -49,33 +50,46 @@ public enum FileTabCommands
...
-			foreach (Document doc in IdeApp.Workbench.Documents.ToArray ()) {
-				var w1 = (SdiWorkspaceWindow) doc.Window;
-				if (w1.TabControl == activeNotebook)
+			var except = GetDocumentException ();
+
+			var docs = IdeApp.Workbench.Documents
+				.Where (doc => ((SdiWorkspaceWindow)doc.Window).TabControl == activeNotebook && (except == null || doc.Window.ViewContent != except))
+				.ToArray ();
+
+			var dirtyDialogShown = docs.Count (doc => doc.IsDirty) > 1;
```
Therzok: I hope you meant only one dirty dialog.

Therzok:
Up for review again. I refactored the code to use more common code in the command handlers. Also, fixed all the outstanding issues.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
```diff
@@ -49,33 +50,46 @@ public enum FileTabCommands
...
+			if (dirtyDialogShown)
+				using (var dlg = new DirtyFilesDialog (docs, except, closeWorkspace: false, groupByProject: false)) {
```
slluis: Why the 'except' parameter? that document is already excluded from the document list

Therzok: I refactored so it takes the docs parameters after I introduced the except variable. I'll fix this up! Thanks!

Therzok:
Okay, should be good now.

slluis:
LGTM

Therzok:
Cherry-picked to master in 1a665119b038242a37eeddf60a3f9cdb6eeb5598

PR #1012, opened 2015-08-26
Remove carbon

iainx:
Second pass at removing the use of Carbon for the ApplicationEvents, instead relying on the Cocoa application delegate methods instead.

Fixes the issue from the previous attempt that starting MonoDevelop from the command line or from inside itself would create a blank MonoDevelop solution in the build/bin directory due to Cocoa interpreting the MonoDevelop.exe from the command line mono-sgen MonoDevelop.exe as a filename


Therzok:
lgtm.

iainx:
Still seems to be failing the UI tests

iainx:
Still seems to be breaking the UI tests

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -326,82 +326,78 @@ void PatchGtkTheme ()
...
+					// OpenFiles may pump the mainloop, but can't do that from an AppleEvent, so use a brief timeout
 					GLib.Timeout.Add (10, delegate {
-						IdeApp.OpenFiles (e.Documents.Select (
-							doc => new FileOpenInformation (doc.Key, doc.Value, 1, OpenDocumentOptions.DefaultInternal))
+						// For mono based apps run from the commandline via `mono-sgen MonoDevelop.exe` Cocoa decides
+						// that MonoDevelop.exe is the first filename to be passed when the app starts up
+						// Therefore we strip any .exe filenames here so we don't attempt to create a new solution based on
+						// that name.
+
+						IdeApp.OpenFiles (e.Filenames.Where (filename => !filename.EndsWith (".exe")).Select (
```
slluis: XS can open .exe files as a solution, so you can run and debug them. So exe files should not be filtered.

iainx:
Ah, ok. I'll have a think how we can work around it then.

iainx:
@slluis filtering .exe files in the OpenFiles event from the AppDelegate doesn't break loading .exe files from the Open File dialog. The only way a .exe file is passed to the OpenFiles delegate is when running from the command line as make run or running from inside of MD itself.

Would you prefer that I just filter out MonoDevelop.exe here instead of *.exe?

iainx:
So this filters out /build/bin/MonoDevelop.exe but nothing else

slluis:
Is this problem specific to starting MD using mono-sgen? doesn't this happen when running from an app bundle?

alanmcgovern:
There are a few ways MD can be started on macos:
`make run-sgen` in a local build which turns into a `mono-sgen monodevelop.exe` call
`make run-boehm` in a local build which turns into a `mono-boehm monodevelop.exe` call.
`make run` which builds the native launcher and turns into a `./monodevelop` call (like an app bundle)
`open MonoDevelop.app` or double click the app bundle

If you need to filter exe's, maybe you could ignore any file which is the same as the entry executable ( Assembly.GetEntryAssembly ). That should give you the behaviour you need without hardcoding specific names or always ignoring all .exe files.

alanmcgovern:
Oh, i just remembered that debugging MD inside MD is a fifth way... :P Either way, that should work!

iainx:
@alanmcgovern yup, using the entry assembly location works. This was only an issue for when we run MD via make run-sgen/boehm or from debugging inside MD itself. The other methods were not affected

alanmcgovern:
sweet :)

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -326,82 +327,78 @@ void PatchGtkTheme ()
...
+				appDelegate.OpenFilesRequest += delegate (object sender, Xwt.Mac.OpenFilesEventArgs e) {
+					// OpenFiles may pump the mainloop, but can't do that from an AppleEvent, so use a brief timeout
 					GLib.Timeout.Add (10, delegate {
-						IdeApp.OpenFiles (e.Documents.Select (
-							doc => new FileOpenInformation (doc.Key, doc.Value, 1, OpenDocumentOptions.DefaultInternal))
+						// For mono based apps run from the commandline via `mono-sgen MonoDevelop.exe` Cocoa decides
+						// that MonoDevelop.exe is the first filename to be passed when the app starts up
+						// Therefore we strip any filenames that match the assembly that we started from.
+						var entryAsm = Assembly.GetEntryAssembly ();
+						Console.WriteLine ("Current ass {0}", entryAsm.Location);
```
Therzok: Some forgotten debug code.

*cough*.

Therzok:
Bump on this and #1065.

Therzok:
What are we missing here?

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -326,82 +327,77 @@ void PatchGtkTheme ()
 
 		void GlobalSetup ()
 		{
+			var appDelegate = (Xwt.Mac.AppDelegate)NSApplication.SharedApplication.Delegate;
```
alanmcgovern: Do we have to cast it to `Xwt.Mac.AppDelegate` or can we cast it to a plain `NSApplicationDelegate`?

iainx: the `Terminating`, `Unhidden`, `OpenUrl` and `OpenFilesRequest` events are `Xwt.Mac.AppDelegate` events, not `NSApplicationDelegate`

PR #1013, opened 2015-08-26
[Core] Fix a NRE exception when opening the unit test margin menu

iainx:
Clicking on the unit test margin was passing a null event to a method that was not expecting one.

### DIFF
--- a/main/src/core/Mono.Texteditor/Mono.TextEditor/Gui/GtkWorkarounds.cs
+++ b/main/src/core/Mono.Texteditor/Mono.TextEditor/Gui/GtkWorkarounds.cs
@@ -489,8 +489,15 @@ namespace Mono.TextEditor
 		public static void ShowContextMenu (Gtk.Menu menu, Gtk.Widget parent, Gdk.EventButton evt, Gdk.Rectangle caret)
 		{
 			int x, y;
+			if (evt == null) {
+				evt = (Gdk.EventButton) Global.CurrentEvent;
+			}
+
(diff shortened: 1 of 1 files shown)


main/src/core/Mono.Texteditor/Mono.TextEditor/Gui/GtkWorkarounds.cs
```diff
@@ -489,11 +489,20 @@ static void ShowContextMenuInternal (Gtk.Menu menu, Gtk.Widget parent, int ix, i
 		public static void ShowContextMenu (Gtk.Menu menu, Gtk.Widget parent, Gdk.EventButton evt, Gdk.Rectangle caret)
 		{
 			int x, y;
+			if (evt == null) {
+				evt = (Gdk.EventButton) Global.CurrentEvent;
+			}
+
 			var window = evt.Window;
```
Therzok: > Initializing coverity mode.

We're doing `evt.Window` here, but `evt` is possibly `null`, given the nullcheck on line 502.

iainx: Oh, the null check on 502 was left over from a previous attempt to fix.

iainx: Removed the null check

Therzok:
:+1:

PR #1016, opened 2015-08-28
[UITest] Adding 'Project Saved' message to array of expected messages…

kdubau:
… in WaitForPackageUpdate

Sometimes during UITests when packages are updated, the IDE message is
‘Project Saved.’ This needs to be added so UITests tests don’t hang at
this step 

@manish

### DIFF
--- a/main/tests/UserInterfaceTests/Ide.cs
+++ b/main/tests/UserInterfaceTests/Ide.cs
@@ -102,6 +102,7 @@ namespace UserInterfaceTests
 		public readonly static Action EmptyAction = delegate { };
 
 		static string[] waitForNuGetMessages = {
+			"Project Saved.",
 			"Package updates are available.",
 			"Packages are up to date.",
 			"No updates found but warnings were reported.",
--- a/main/tests/UserInterfaceTests/Ide.cs
+++ b/main/tests/UserInterfaceTests/Ide.cs
@@ -103,6 +103,7 @@ namespace UserInterfaceTests
 
 		static string[] waitForNuGetMessages = {
 			"Project Saved.",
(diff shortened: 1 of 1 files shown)


manish:
Also add `"Solution loaded."`

main/tests/UserInterfaceTests/Ide.cs
```diff
@@ -102,6 +102,8 @@ public static void RunAndWaitForTimer (Action action, string counter, int timeou
 		public readonly static Action EmptyAction = delegate { };
 
 		static string[] waitForNuGetMessages = {
+			"Project Saved.",
```
Therzok: Now would be a good time to make the 'saved' lowercase. cc @vancura

Therzok:
whitelist

main/tests/UserInterfaceTests/Ide.cs
manish: From what I know Xamarin Studio shows "Project saved." where "saved" is in lower case

main/tests/UserInterfaceTests/Ide.cs
```diff
@@ -34,8 +34,8 @@
 using MonoDevelop.Components.AutoTest;
 
 using NUnit.Framework;
-using System.Collections.Generic;
-
+using System.Collections.Generic;
+
 namespace UserInterfaceTests
```
manish: also use `git commit -p` to selectively choose which hunks to commit.

manish:
Squashed and pushed.
https://github.com/mono/monodevelop/commit/4dfcbddd566c9f79a9d1d9521b93e071ace97493

PR #1018, opened 2015-09-02
[Ide] Allow copy/pasting of read-only properties

Therzok:
This enables interaction with text in read-only properties in the property
pad.

Bug 15161 - Cannot copy disabled property values to clipboard


slluis:
LGTM

Therzok:
Merged manually with f3c2fb1be92440b96203310b365fa52c1a7a8e9e

PR #1022, opened 2015-09-03
[UITest] Remove forward slashes from Ide log path during test setup

kdubau:
The name of the test is used to create the log file path, if a forward slash is present, the path gets corrupted. This change removes forward slash from test name when creating the IDE log path during test setup.

### DIFF
--- a/main/tests/UserInterfaceTests/UITestBase.cs
+++ b/main/tests/UserInterfaceTests/UITestBase.cs
@@ -153,7 +153,7 @@ namespace UserInterfaceTests
 
 		void SetupIdeLogFolder ()
 		{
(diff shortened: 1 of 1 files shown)


main/tests/UserInterfaceTests/UITestBase.cs
```diff
@@ -153,7 +153,7 @@ void SetupScreenshotsFolder ()
 
 		void SetupIdeLogFolder ()
 		{
-			var currentXSIdeLog = Path.Combine (currentTestResultFolder, string.Format ("{0}.Ide.log", TestContext.CurrentContext.Test.FullName));
+			var currentXSIdeLog = Path.Combine (currentTestResultFolder, string.Format ("{0}.Ide.log", TestContext.CurrentContext.Test.FullName.Replace ('/','_')));
```
manish: Looks good, but also replace `\` with an underscore (for windows)

PR #1026, opened 2015-09-08
Reorder buttons for folder delete dialog

PiotrZierhoffer:
The order of buttons was different to the one in file removal dialog, which was confusing.


Therzok:
:+1:

slluis:
Cancel between delete and remove? that looks weird.

Therzok:
Oh, missed that. @slluis so we want, cancel, remove, delete?

PiotrZierhoffer:
Well, I went with the order from the more general (in my opinion) approach, in `ProjectFileNodeBuilder`. That is where the inconsistency arises. There is "Delete -> Cancel -> Remove from project" order there... So maybe it should be fixed as well?

Therzok:
:+1: from me

iainx:
Surely this is the Gtk+/OS X way, the more dangerous "extra" option to the left of the dialog

[Delete]                             [Cancel] [Remove from Project]

screen shot 2015-09-08 at 22 53 19

vs 

screen shot 2015-09-08 at 22 53 52

I think the [Delete] button should be moved further out.

Therzok:
QuestionMessage doesn't support it, sadly. We'd need it to support it.

PR #1027, opened 2015-09-09
Open up MonoDevelop APIs so I can implement a Protobuild add-in

hach-que:
This opens up a bunch of MonoDevelop APIs and adds various hooks that are required for me to implement a Protobuild add-in.

[Protobuild](http://protobuild.org/) is a cross-platform project generator; it works by generating MSBuild projects and solutions on disk when the Protobuild.exe executable is run.  This allows a common project format like this: https://github.com/mono/MonoGame/blob/develop/Build/Projects/MonoGame.Framework.definition to produce the appropriate C# projects for any platform (Windows, Linux, Mac, iOS, Android, etc.).

I've been writing an add-in that will allow MonoDevelop to load Protobuild modules directly instead of requiring the user to generate projects via the command-line first.  This requires additional APIs in MonoDevelop so that I can have an additional project layer on top of the MSBuild project, without losing functionality (there are a lot of places in MonoDevelop that check if the project inherits from an MSBuild project type to provide things like debugging).

Roughly:

- This adds a GetProjectForTypeSystem method so that the C# completion can get access to the real MSBuild project where types are defined.
- This adds Suspend and Resume methods to IViewContent.  This allows hot-reload of editor projects as demonstrated in this video: https://www.youtube.com/watch?v=TO7UxNKS5dQ
- Adds runtime registration and deregistration of file handlers in DisplayBindingService.  This allows Protobuild projects that are loaded by the IDE to handle additional file types.
- Makes a bunch of methods public and virtual.


Therzok:
Heya! I have a few comments before anything else goes:

> This adds Suspend and Resume methods to IViewContent. This allows hot-reload of editor projects as demonstrated in this video: https://www.youtube.com/watch?v=TO7UxNKS5dQ

This is an API break and might have to be delayed for 6.0, which is going to happen after `roslyn` branch is merged to master.

> Makes a bunch of methods public and virtual.

I'm not a big fan of those changes, but maybe we could work around having to do that?

Couldn't the add-in just regenerate the Protobuild data on solution opening (`IdeApp.Workspace.FirstWorkspaceItemOpened`)? There should be some sort of metadata that says - hey, your projects need to be generated.

I'm also not a big fan of `Protobuild.exe` being a special-case. While an extension point can be added to get a 'whitelist' of items, it's not something I'd do in the file selector dialog.

This PR needs a bit of fixing here and there and a bit of rethinking.

cc @slluis @mhutch @mkrueger

Therzok:
Opening up the New File Dialog, Node Builders and such shouldn't happen. You may copy the implementations inside your own add-in instead of changing the API here. It would limit what we can do without breaking compatibility and that's something we want to avoid.

Therzok:
Also, the MIT license requires you to keep the copyright on [copied-but-modified files](https://github.com/hach-que/Protobuild.IDE.MonoDevelop/blob/master/Protobuild.MonoDevelop/MonoDevelop.Protobuild/NewProtobuildModuleDialog.cs).

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/IViewContent.cs
```diff
@@ -49,6 +49,9 @@ public interface IViewContent : IBaseViewContent
 		void LoadNew (System.IO.Stream content, string mimeType);
         void Save (string fileName);
         void Save ();
+
+		void Suspend();
```
Therzok: This could become an `ISuspendableViewContent`, which would be casted using `as` for the suspend/resume calls.

Therzok: i.e. `public interface ISuspendableViewContent : IViewContent { ... }`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService.cs
```diff
@@ -230,13 +230,16 @@ static TypeSystemService ()
 
 				foreach (var content in projectContents.Values.ToArray ()) {
 					var files = new List<ProjectFile> ();
-					foreach (var file in e) {
+					foreach (var file in e) {
+					    if (content == null) {
```
Therzok: if the content is null, nothing happens anymore. So I guess we can move this at the beginning of the loop.

Therzok:
Also, if possible, this should go in in smaller batches which have atomic changes. The commit is too big and it would be easier to filter smaller changes and fix and merge them quickly locally rather than having to comment on everything.

i.e. I'd order commits such as some fix an issue you saw, some expose some API you need (one commit for one functionality), some extend existing code, etc.

hach-que:
> This is an API break and might have to be delayed for 6.0, which is going to happen after roslyn branch is merged to master.

For a bit of background on this, the reason I need Suspend / Resume is that switching tabs destroys the control holding the window handle that the embedded GL context is using.  I spent ages trying to find a way that works on Linux where this wasn't necessary, but unfortunately there isn't any solution other than destroying and re-creating a GL context when the window holding it disappears.

> This could become an `ISuspendableViewContent`, which would be casted using `as` for the suspend/resume calls.

This will work for me.

> Also, if possible, this should go in in smaller batches which have atomic changes. The commit is too big and it would be easier to filter smaller changes and fix and merge them quickly locally rather than having to comment on everything.

> i.e. I'd order commits such as some fix an issue you saw, some expose some API you need (one commit for one functionality), some extend existing code, etc.

The reason for the large commit is that for a long time I was developing the Protobuild add-in inline in the IDE code, and only recently separated it out into it's own add-in.  The reason for this is that I was expected to have to make far more major changes to MonoDevelop than I had to.

> Opening up the New Project Dialog, Node Builders and such shouldn't happen. You may copy the implementations inside your own add-in instead of changing the API here. It would limit what we can do without breaking compatibility and that's something we want to avoid.

I'm not sure this will work because the MonoDevelop code looks for specific implementations of classes.  If I can't derive solution / project I'm pretty sure debugging stops work among other things.  In particular, if I copied the folder and project file node builders, then they wouldn't get all of the behaviour and contextual menu items that are currently associated with those types.

The alternative to opening these up is to have Protobuild support in the upstream code (like MSBuild project support itself is), but I'm pretty sure you won't want to do this given that none of the upstream contributors are using it.

slluis:
> This adds a GetProjectForTypeSystem method so that the C# completion
> can get access to the real MSBuild project where types are defined.

You should be aware that we are working on a new project model and type system (based on Roslyn), which will be ready in the next months, and will be included in MD 6.0. At this point, we won't accept big changes on those systems. This needs to be reviewed by @mkrueger.

> This adds Suspend and Resume methods to IViewContent

We could add something like that, using the interface proposed above to avoid breaking the api.

> Adds runtime registration and deregistration of file handlers in DisplayBindingService.
> This allows Protobuild projects that are loaded by the IDE to handle additional file types

Looks good.

> Makes a bunch of methods public and virtual

I can't accept none of those changes without an explanation. We want to avoid making api public unless really necessary. Also, we try to avoid public virtual members in the api.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
```diff
@@ -1126,7 +1126,7 @@ protected override IList<string> GetCommonBuildActions ()
 			return BuildAction.DotNetCommonActions;
 		}
 
-		internal override void SetItemHandler (ISolutionItemHandler handler)
+		public override void SetItemHandler (ISolutionItemHandler handler)
```
hach-que: This is so that I can call `SetItemHandler(new MSBuildHandler(definition.Guid.ToString(), definition.Guid.ToString()));` and similar, which from memory appears to be required to get the items to behave correctly in the solution explorer.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Solution.cs
```diff
@@ -72,7 +72,7 @@ public Solution ()
 			configurations = new SolutionConfigurationCollection (this);
 		}
 		
-		public SolutionFolder RootFolder {
+		public virtual SolutionFolder RootFolder {
```
hach-que: This is so the Protobuild module can provide a different root folder for the solution (based on the module configuration).  It is overridden in a derived class (I also need to change the set behaviour which is why internal is removed below).

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
```diff
@@ -50,7 +50,7 @@
 
 namespace MonoDevelop.Ide.Gui.Pads.ProjectPad
 {
-	abstract class FolderNodeBuilder: TypeNodeBuilder
+	public abstract class FolderNodeBuilder: TypeNodeBuilder
```
hach-que: This is needed for https://github.com/hach-que/Protobuild.IDE.MonoDevelop/blob/b9f744ab29df37e45a4ff37d7e4342ec99d92989/Protobuild.MonoDevelop/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProtobuildDefinitionNodeBuilder.cs, which builds the node for a definition entry in the solution explorer (analogous to a C# project).

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
```diff
@@ -148,7 +148,7 @@ public override bool HasChildNodes (ITreeBuilder builder, object dataObject)
 		}
 	}
 	
-	abstract class FolderCommandHandler: NodeCommandHandler
+	public abstract class FolderCommandHandler: NodeCommandHandler
```
hach-que: This is needed for https://github.com/hach-que/Protobuild.IDE.MonoDevelop/blob/b9f744ab29df37e45a4ff37d7e4342ec99d92989/Protobuild.MonoDevelop/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProtobuildDefinitionNodeBuilder.cs#L359, which implements commands in addition to the default commands for project definitions.  If I didn't inherit this class, the default commands like adding new files wouldn't be available, and drag-and-drop behaviour would be missing.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
```diff
@@ -42,7 +42,7 @@
 
 namespace MonoDevelop.Ide.Gui.Pads.ProjectPad
 {
-	class ProjectFileNodeBuilder: TypeNodeBuilder
+	public class ProjectFileNodeBuilder: TypeNodeBuilder
```
hach-que: This is needed for https://github.com/hach-que/Protobuild.IDE.MonoDevelop/blob/b9f744ab29df37e45a4ff37d7e4342ec99d92989/Protobuild.MonoDevelop/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProtobuildProjectFileNodeBuilder.cs, which extends all file types in a project to have a "Platforms" option, which indicates what platforms the file should be included / excluded in build for.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
```diff
@@ -159,7 +159,7 @@ public override void BuildChildNodes (ITreeBuilder treeBuilder, object dataObjec
 		}
 	}
 	
-	class ProjectFileNodeCommandHandler: NodeCommandHandler
+	public class ProjectFileNodeCommandHandler: NodeCommandHandler
```
hach-que: Likewise, this is needed to create ProtobuildProjectFileNodeBuilder so that the behaviour is an extension of the current project file behaviour.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/SolutionNodeBuilder.cs
```diff
@@ -41,7 +41,7 @@
 
 namespace MonoDevelop.Ide.Gui.Pads.ProjectPad
 {
-	class SolutionNodeBuilder: TypeNodeBuilder
+	public class SolutionNodeBuilder: TypeNodeBuilder
```
hach-que: This is needed for https://github.com/hach-que/Protobuild.IDE.MonoDevelop/blob/b9f744ab29df37e45a4ff37d7e4342ec99d92989/Protobuild.MonoDevelop/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProtobuildModuleNodeBuilder.cs, so that Protobuild modules (analogous to C# solutions) have the same but extended behaviour.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/ProgressMonitors.cs
```diff
@@ -66,7 +66,7 @@ public IProgressMonitor GetRebuildProgressMonitor ()
 			return GetBuildProgressMonitor (GettextCatalog.GetString ("Rebuilding..."));
 		}
 		
-		private IProgressMonitor GetBuildProgressMonitor (string statusText)
+		public IProgressMonitor GetBuildProgressMonitor (string statusText)
```
hach-que: I use this here: https://github.com/hach-que/Protobuild.IDE.MonoDevelop/blob/b9f744ab29df37e45a4ff37d7e4342ec99d92989/Protobuild.MonoDevelop/MonoDevelop.Ide/ProtobuildStartup.cs#L59 to show progress of project generation when the user is loading a Protobuild module.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/FinalProjectConfigurationPage.cs
```diff
@@ -31,7 +31,7 @@
 
 namespace MonoDevelop.Ide.Projects
 {
-	class FinalProjectConfigurationPage
+	public class FinalProjectConfigurationPage
```
hach-que: I use this on https://github.com/hach-que/Protobuild.IDE.MonoDevelop/blob/b9f744ab29df37e45a4ff37d7e4342ec99d92989/Protobuild.MonoDevelop/MonoDevelop.Protobuild/FinalProtobuildModuleConfigurationPage.cs, where I add a ProtobuildSolutionTemplate property which is used to track which Protobuild template should be used to create the new solution.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkProjectFolderPreviewWidget.cs
```diff
@@ -35,7 +35,7 @@
 namespace MonoDevelop.Ide.Projects
 {
 	[System.ComponentModel.ToolboxItem (true)]
-	partial class GtkProjectFolderPreviewWidget : Gtk.Bin
+	public partial class GtkProjectFolderPreviewWidget : Gtk.Bin
```
hach-que: This is used on the Protobuild new module configuration page because I want identical behaviour to the new project wizard.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkTemplateCategoryCellRenderer.cs
```diff
@@ -31,7 +31,7 @@
 
 namespace MonoDevelop.Ide.Projects
 {
-	class GtkTemplateCategoryCellRenderer : CellRendererText
+	public class GtkTemplateCategoryCellRenderer : CellRendererText
```
hach-que: Needed for the new module wizard here https://github.com/hach-que/Protobuild.IDE.MonoDevelop/blob/b9f744ab29df37e45a4ff37d7e4342ec99d92989/Protobuild.MonoDevelop/MonoDevelop.Protobuild/GtkNewProtobuildModuleDialogBackend.UI.cs#L294.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkTemplateCellRenderer.cs
```diff
@@ -36,7 +36,7 @@
 
 namespace MonoDevelop.Ide.Projects
 {
-	class GtkTemplateCellRenderer : CellRendererText
+	public class GtkTemplateCellRenderer : CellRendererText
```
hach-que: Needed for the new module wizard here https://github.com/hach-que/Protobuild.IDE.MonoDevelop/blob/b9f744ab29df37e45a4ff37d7e4342ec99d92989/Protobuild.MonoDevelop/MonoDevelop.Protobuild/GtkNewProtobuildModuleDialogBackend.UI.cs#L311.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/TemplateCategoryPath.cs
```diff
@@ -28,7 +28,7 @@
 
 namespace MonoDevelop.Ide.Templates
 {
-	class TemplateCategoryPath
+	public class TemplateCategoryPath
```
hach-que: Needed so I can create the new module wizard: https://github.com/hach-que/Protobuild.IDE.MonoDevelop/blob/b9f744ab29df37e45a4ff37d7e4342ec99d92989/Protobuild.MonoDevelop/MonoDevelop.Protobuild/NewProtobuildModuleDialog.cs#L340

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/TemplateImageProvider.cs
```diff
@@ -31,7 +31,7 @@
 
 namespace MonoDevelop.Ide.Templates
 {
-	class TemplateImageProvider : IDisposable
+	public class TemplateImageProvider : IDisposable
```
hach-que: Needed for the new module wizard...

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/TemplateWizardProvider.cs
```diff
@@ -33,7 +33,7 @@
 
 namespace MonoDevelop.Ide.Templates
 {
-	class TemplateWizardProvider
+	public class TemplateWizardProvider
```
hach-que: Needed for the new module wizard...

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/TemplateWizardProvider.cs
```diff
@@ -129,7 +129,7 @@ protected virtual TemplateWizard GetWizard (string id)
 			return IdeApp.Services.TemplatingService.GetWizard (id);
 		}
 
-		void Reset ()
+	    public void Reset ()
```
hach-que: Needed for the new module wizard...

main/src/core/MonoDevelop.Ide/gtk-gui/MonoDevelop.Ide.Projects.GtkProjectFolderPreviewWidget.cs
```diff
@@ -1,76 +1,76 @@
...
-			}
-			this.Hide ();
-		}
-	}
-}
+
+// This file has been generated by the GUI designer. Do not modify.
+namespace MonoDevelop.Ide.Projects
+{
+	public partial class GtkProjectFolderPreviewWidget
```
hach-que: Needed for the new module wizard...

hach-que:
I've added inline comments on the public / virtual changes on why they're needed and where they're used.  Some of them I couldn't find references to so when I get a chance I'll try reverting them in the commit and see if everything still builds.

hach-que:
> You should be aware that we are working on a new project model and type system (based on Roslyn), which will be ready in the next months, and will be included in MD 6.0. At this point, we won't accept big changes on those systems. This needs to be reviewed by @mkrueger.

I don't really need to make changes to those systems; I just need a hook point to say "the project that you see in the solution explorer may not be the one you need to use to evaluate the type system".  A level of indirection where I can hand back a different project than the one in the solution explorer is all I need to make this work (since in terms of the build and type systems I still want MonoDevelop to be using the actual C# projects and solutions, while being able to present the user with a solution view that's more useful for working with cross-platform projects).

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
slluis: You can use ProjectExtensionUtil.InstallHandler.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
slluis: You can use a NodeBuilderExtension for the ProtobuildProjectFile class. There is no need to subclass TypeNodeBuilder.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
slluis: Your NodeBuilderExtension subclass for ProtobuildProjectFile can provide its own command handler. No need to subclass ProjectFileNodeCommandHandler.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/SolutionNodeBuilder.cs
slluis: Use a NodeBuilderExtension

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/FinalProjectConfigurationPage.cs
slluis: If you need a specific extensibility feature in the project creation dialog we can try to add it, but we are not going to make any of those classes public.

hach-que:
I'm going to start breaking out this PR into smaller PRs.  I'll base those PRs on the roslyn branch so that this stuff will be available in 6.0 (and to allow for those API breaks like `IViewContent`).

Therzok:
Sounds good, thanks!

PR #1030, opened 2015-09-10
[UITest] Fix PollStatusMessage for proper error message

manish:


main/tests/UserInterfaceTests/Ide.cs
```diff
@@ -141,13 +141,14 @@ public static void WaitForNoStatusMessage (string[] statusMessage, int timeoutIn
 		static void PollStatusMessage (string[] statusMessage, int timeoutInSecs, int pollStepInSecs, bool waitForMessage = true)
 		{
 			Ide.WaitUntil (() => {
+				string actualStatusMessage = null;
```
kdubau: `string actualStatusMessage = string.Empty;` ? @manish

manish: I don't think it would be a problem here. `string.Format` doesn't break on `null`.

Infact it would tell that `Workbench.GetStatusMessage` actually failed instead of getting an empty string from status bar.

kdubau: Sounds good.

PR #1038, opened 2015-09-15
Revert "[FSharp] Bump fsharpbinding submodule to fix #31456"

slluis:
Reverts mono/monodevelop#1037

Looks like this commit doesn't fix the bug.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -34,7 +34,7 @@
 [submodule "main/external/fsharpbinding"]
 	path = main/external/fsharpbinding
 	url = git://github.com/fsharp/fsharpbinding.git
-	branch = 5.9-c5sr4
+	branch = master
 [submodule "main/external/nuget-binary"]
 	path = main/external/nuget-binary
 	url = git://github.com/mono/nuget-binary.git


7sharp9:
It does fix the bug in the F# addin iOS template, the problem is the
different md-addins for this branch also has the iOS templates with the non
fixed template which is overriding this.  The templates were moved from the
F# addin to md-addins which is what's confounded this.

On Tuesday, 15 September 2015, Lluis Sanchez <notifications@github.com>
wrote:

> Merged #1038 <https://github.com/mono/monodevelop/pull/1038>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mono/monodevelop/pull/1038#event-409755503>.
>


-- 
Use F#

slluis:
@7sharp9 The templates were removed from fsharpbinding in the 5.9 branch of fsharpbinding, and moved to md-addins master. The problem is that the fsharpbinding branch was not properly bumped in  monodevelop-5.9-branch so we ended having the templates both in md-addins and fsharpbinding. XS seems to be loading the templates from md-addins, so the fix has to go there.

PR #1039, opened 2015-09-15
[Mac] Dynamically resize the config/platform selector

Therzok:
Bug 33769 - InvalidCast from the main bar
Bug 31116 - Mac Toolbar: Responsive treatment
Bug 30550 - Xamarin Studio crashes when resizing

cc @vancura @slluis @iainx

### DIFF
--- a/main/src/addins/MacPlatform/MainToolbar/MainToolbar.cs
+++ b/main/src/addins/MacPlatform/MainToolbar/MainToolbar.cs
@@ -114,7 +114,6 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 			// Add spacings.
 			nfloat spacing = (allItems.Length - 1) * 16;
 			e.AllItemsWidth += spacing;
-			e.WindowWidth -= spacing;
 
 			return e;
 		}
--- a/main/src/addins/MacPlatform/MainToolbar/MainToolbar.cs
+++ b/main/src/addins/MacPlatform/MainToolbar/MainToolbar.cs
@@ -66,6 +66,10 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 		}
 
 		int selectorIdx;
+		SelectorView selector {
+			get { return (SelectorView)widget.Items[selectorIdx].View; }
+		}
+
 		SelectorView.PathSelectorView selectorView {
 			get { return (SelectorView.PathSelectorView)widget.Items[selectorIdx].View.Subviews [0]; }
 		}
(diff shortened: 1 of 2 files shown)


main/src/addins/MacPlatform/MainToolbar/MainToolbar.cs
```diff
@@ -181,15 +224,9 @@ NSToolbarItem CreateSearchBarToolbarItem ()
...
-			var menuBar = new SearchBar {
-				Frame = new CGRect (0, 0, 180, bar.FittingSize.Height),
-			};
 			var item = new NSToolbarItem (SearchBarId) {
 				View = bar,
-				MenuFormRepresentation = new NSMenuItem {
-					View = menuBar,
-				},
-				MinSize = new CGSize (180, bar.FittingSize.Height),
+				MinSize = new CGSize (150, bar.FittingSize.Height),
```
Therzok: This size had to change so we could fit everything on screen _even with_ the new resizing..

main/src/addins/MacPlatform/MainToolbar/MainToolbar.cs
```diff
@@ -220,7 +257,7 @@ NSToolbarItem CreateStatusBarToolbarItem ()
 
 				double maxSize = Math.Round (bar.Window.Frame.Width * 0.30f);
 				double minSize = Math.Round (bar.Window.Frame.Width * 0.25f);
-				item.MinSize = new CGSize ((nfloat)Math.Max (280, minSize), 22);
+				item.MinSize = new CGSize ((nfloat)Math.Max (220, minSize), 22);
```
Therzok: This size had to change so we could fit everything on screen even with the new resizing..

Therzok:
This gets rid of the overflow menu at minimum size. There's also a quirk that happens in some cases that I didn't get to track down (it's really hard to debug this thing) - resizing too fast will leave the platform selector with no text, but I have no clue how/why that happens.

Therzok:
Quirks are gone with 51781b7

iainx:
The status bar icons don't always update

screen shot 2015-09-16 at 12 27 04 pm

iainx:
It still seems a bit glitchy, with the overflow menu still being able to be triggered by quickly reducing the size of the window, and it only corrects itself when releasing mouse pointer. Partly it seems there's a delay between when the selector view's space is changed and the view redraws itself.

Therzok:
@iainx that's why I added the endliveresize, cocoa is not being really helpful.

Therzok:
@iainx stuff got fixed. We don't care if the item goes in the overflow menu for a bit. Nothing's going to change while you resize the window. You can't even do the search command, which would be the only fail case.

Therzok:
Bugs 33955 too.

PR #1043, opened 2015-09-16
[AutoTest] Adding logic in StartApplication for running windows UI Te…

kdubau:
…sts on wrench

When you build on wrench, the executable is XamarinStudio.exe. The logic still looks for MonoDevelop.exe first, but if it's not found, will use XamarinStudio.exe instead.


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestClientSession.cs
```diff
@@ -65,6 +65,11 @@ public void StartApplication (string file = null, string args = null, IDictionar
 			if (file == null) {
 				var binDir = Path.GetDirectoryName (typeof(AutoTestClientSession).Assembly.Location);
 				file = Path.Combine (binDir, "MonoDevelop.exe");
+				if (!File.Exists (file)) {
+					file = Path.Combine (binDir, "XamarinStudio.exe");
+				}
+			} else if (!File.Exists (file)) {
+				file = file.Replace ("MonoDevelop.exe", "XamarinStudio.exe");
```
manish: maybe also add

if
```
if (!File.Exists (file)) {
    throw new FileNotFoundException (file);
}
```

manish:
:100:

PR #1044, opened 2015-09-17
[UITest] Change LogMessageValidator to allow access to the log on win…

kdubau:
…dows

On windows, the IDE holds a lock on the IDE's log file which was causing
unauthorized exception when we try to read it. With this change we can
read it on both windows and mac.


manish:
:+1:

PR #1045, opened 2015-09-17
Fixup Mac toolbar crashers

Therzok:


alanmcgovern:
Confirmed to fix multiple different crashes in master so merging it in to c6 too

PR #1046, opened 2015-09-17
[DesignerSupport] Update PropertyGrid on ProjectFile property changes.

Therzok:
Refresh the propertygrid on projectfile property changes.

Bug 33629 - Property grid should refresh when build action changed


slluis:
Looks good to me. Just make sure the provider is always disposed.

Therzok:
Yep, they are.

PR #1047, opened 2015-09-17
[PropertyPad] Ignore autohide boxes from the command target visitor

Therzok:
Bug 33626 - [Regression] Auto-hidden property pad can't be used with
solution tree


Therzok:
I'm going to merge this as it fixes the said bug with minimum impact. The other PR is more intrusive and while it seems like a general fix, given the fact that the PropertyPadVisitor is the only consumer of target visiting, this is 100% general.

PR #1048, opened 2015-09-17
[Ide] Delegate the DockFrameTopLevel command to its frame.

Therzok:


Therzok:
Same as #1047, but approached differently.

Therzok:
Also fixes bug 25991.

slluis:
This looks weird to me. Why redirect the command route to the pad content? if the pad content has the focus, it will be already part of the route. What is the use case?

Therzok:
It is not redirecting to the pad content. The one which gets the focus is a RadioButton inside AutoHideBox (I still don't know why a RadioButton).

Redirecting to the pad content allows the hierarchy to switch to the pad itself when using AutoHide pads. This means at once the DockItemContainer from any widget inside the pad, we end up redirecting to the actual pad _at least at some point_.

Therzok:
Should we just merge #1047 instead for now? Fixes the bug, we can revisit this in the new dock abstraction.

Therzok:
Possibly since it's the _window_ which gets the focus on OS X (auto-hide pads are windows inside XS), the first element that gets the focus could be a radio button inside the pad toolbar.

We could possibly force the focus to not go to the toolbar by default.

Therzok:
I tried forcing the toolbars not getting focus and their items too, but didn't work out. RadioButtons set with `Default = false` still get the focus. This is a creepy situation where I don't understand why the focus chain is still pointing to those even though they're forced not to be default.

PR #1049, opened 2015-09-18
Stub launcher

Therzok:
cc @alanmcgovern


slluis:
What does this fix?

alanmcgovern:
It's so a normal 'make run' will use the same launcher as our app bundle. This should let local devs trigger the same bugs that previously we could only trigger with an app bundle. I'll test that theory in a few minutes and see if yesterday's crashed shows up when we use this pr

alanmcgovern:
I sent an email asking for this yesterday :)

alanmcgovern:
This this result in `make run` triggering the same issues that the app bundle triggers. :+1: for reproducibility of bugs!

Therzok:
Merged manually.

PR #1050, opened 2015-09-18
[AutoTest] handle unauthorized exception in OnCleanUp ()

kdubau:
When running ui tests on windows via wrench, some git files can't be cleaned up until mdtool exits. Handle the unauthorized exception to avoid test failure. Wrench will clean up the rest


manish:
:100:

PR #1054, opened 2015-09-21
[IDE] Display the main IDE window before the Yes/No addin load error …

iainx:
…dialog

If we display an error dialog before the main window, then Cocoa creates an extra empty Application menu that turns into a second Apple menu when the IDE starts. This patch moves the IDE main window creation to before the Yes/No addin load error dialog.

Fixes BXC #33699


Therzok:
:+1:

slluis:
:+1:

PR #1055, opened 2015-09-22
[IDE] Force the solutionpad treeview to update its offsets

iainx:
If the treeview is changed when unrealised (autohidden), the treeview doesn't update its internal vertical offset. This can lead to items being offset outside the visible area and therefore becoming unreachable. The only way to force the treeview is to recalculate this offset by setting the Vadjustment.Value, but it ignores values the same as the current value. Therefore we simply set it to something slightly different then back again.

See also MonoDevelop.Ide.Gui.Components.PadTreeView for the same fix.

Fixes BXC #31712

This sounds like something that might be able to be fixed in Gtk+ itself? @bratsche


slluis:
@bratsche it would be great if this could be fixed in GTK.

PR #1057, opened 2015-09-22
[UITest] Add NUnit Timeout attribute to UI Tests

kdubau:


main/tests/UserInterfaceTests/VersionControlTests/Git/GitTests.cs
```diff
@@ -30,11 +30,11 @@
 
 namespace UserInterfaceTests
 {
-	[TestFixture]
+	[TestFixture, Timeout(60000)]
 	[Category ("Git")]
 	public class GitTests : VCSBase
 	{
-		[Test]
+		[Test, Timeout(90000)]
```
manish: TestGitClone is 90s and GitTests is 60s.

kdubau: It's intentional. The clone test takes a bit longer than the others.

manish: but time taken by a fixture includes all the tests.

So even if GitClone passes (say 70s), GitTests will fail as it took more than 60s

kdubau: The timeout on TestFixture is just the default for any subordinate test cases. According to the [docs] (http://www.nunit.org/index.php?p=timeout&r=2.5).
> The attribute may also be specified on a fixture or assembly, in which case it indicates the default timeout for any subordinate test cases.

manish: :+1:

PR #1058, opened 2015-09-22
[AutoTest] Implement TakeScreenshot for Windows

kdubau:
cc: @manish @Therzok 

Will this be okay on Linux?


kdubau:
I left the types fully qualified so I didn't have to add any using statements as I wasn't sure what the impact would be. Let me know if I should change that.

Therzok:
:+1: this is fine by me.

PR #1059, opened 2015-09-23
[IDE] Reset the search results when changing solution

iainx:
Fixes BXC #33281


slluis:
This is not the right event to subscribe. CurrentSelectedSolutionChanged is raised when the solution selected in the solution pad changes (when you have more than one solution open). In this case it would make sense to use IdeApp.Workspace.LastWorkspaceItemClosed.

PR #1060, opened 2015-09-23
[Components] Only add the Search in Solution category if a solution i…

iainx:
…s open.

Don't add the Search in Solution category to the Search popup window if no solution is open.

Fixes BXC #33199


slluis:
The roslyn branch already has a fix for this: https://github.com/mono/monodevelop/blob/roslyn/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/SearchInSolutionSearchCategory.cs#L49

We can use the same fix.

PR #1063, opened 2015-09-23
[AutoTest] Add code to set the active runtime and configuration in SelectorView

manish:


Therzok:
As this doesn't affect production code, I think it can go as soon as people give their :+1: .

So :+1: from me.

PR #1064, opened 2015-09-24
[UITest] Implement test log file and ReproStep method

kdubau:
This creates a log file specific for the current test in the TestResults folder. It is registered with the LoggingService so any log entry will go to the log (usually errors which will be useful). 

The ReproStep method allows entries to the log that we can later parse to create bug report repro steps. The log looks something like this:
```
INFO[2015-09-24 15:09:13Z]: @Repro-Step-01: Create a new project
INFO[2015-09-24 15:09:13Z]: @Repro-Info-01: CategoryRoot=Other, Category=.NET, TemplateKindRoot=General, TemplateKind=Console Project
INFO[2015-09-24 15:09:13Z]: @Repro-Info-01: ProjectName=ConsoleProject, SolutionName=ConsoleProject, SolutionLocation=/Users/kylewhite/Workspace/monodevelop/main/build/tests/wrveibzd.8xy, ProjectInSolution=True, AddProjectToExistingSolution=False
INFO[2015-09-24 15:09:17Z]: @Repro-Step-02: Build solution
```

@manish What do you think? We would need to retroactively go to all the helper methods and add ReproStep calls where necessary, I only did it for `CreateBuildProject`.

### DIFF
--- a/main/tests/UserInterfaceTests/UITestBase.cs
+++ b/main/tests/UserInterfaceTests/UITestBase.cs
@@ -31,6 +31,8 @@ using System;
 using System.Collections.Generic;
 using Newtonsoft.Json;
 using System.Linq;
+using MonoDevelop.Core.Logging;
+using MonoDevelop.Core;
 
 namespace UserInterfaceTests
 {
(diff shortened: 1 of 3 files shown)


manish:
LGTM. Please create a trello card so that we can track all the places where we need to add this.

Does this output you posted, comes from console output or from actual Ide.log file?

manish:
:ship:

main/tests/UserInterfaceTests/UITestBase.cs
```diff
@@ -95,6 +99,7 @@ public virtual void Teardown ()
 						Assert.Inconclusive ("Xamarin Update is blocking the application focus");
 				}
 				ValidateIdeLogMessages ();
```
manish: You need to add `LoggingService.RemoveLogger (Logger.Name);` here

PR #1066, opened 2015-09-25
[UITest] Added WaitForPackageUpdateOrSaved for NuGetAdded or Project Saved

manish:
There are times when we are not sure if "Packages Added" or
"Project Saved" would end up showing on the status bar as there
is no defined order. This method can be used in tests to make
sure that out tests don't fail just because we can't predict
the order of status message

Also check for Xamarin Update open in a different line checking
first if it is not null to avoid NRE

Also update WaitForTextArea in git tests to give it 20 seconds
instead of standard 5 seconds for the project to be opened


kdubau:
I'm fine with it

PR #1067, opened 2015-09-25
Cycle6 vcs fixes

Therzok:


slluis:
Is this still valid?

Therzok:
Only the subversion only isn't cherry-picked into cycle6.

PR #1068, opened 2015-09-25
[UITest] Add BuildTimeout in ProjectDetails as some templates take more time to build than others

manish:
Android Wear takes way more time, so hardcoding time for all templates of one Category is not a sensible way to go. The timeout should be associated with each template


kdubau:
:+1:

PR #1070, opened 2015-09-25
Fix windows gitconfig uitests

kdubau:


manish:
:shipit:

main/tests/UserInterfaceTests/Util.cs
```diff
@@ -39,6 +39,11 @@ public static void PrintData (this object data)
 				TestService.Session.DebugObject.Debug (data.ToString ());
 		}
 
+		public static string ToPathSafeString (this string str)
+		{
+			return str.Replace (@"\", "").Replace ("/", "");
```
Therzok: Shouldn't this replaced with dashes?

Therzok: You'd end up with `originmaster` strings in the worst case scenario.

kdubau: good Idea I'll update it to dashes

PR #1071, opened 2015-09-26
[UITest] Enhanced ToPathSafeString utility method

kdubau:
It now removes most invalid characters rather than only slashes. 

cc: @Therzok, since you pointed out it should be dashes

### DIFF
--- a/main/tests/UserInterfaceTests/Util.cs
+++ b/main/tests/UserInterfaceTests/Util.cs
@@ -28,6 +28,7 @@ using System;
 using System.IO;
 using MonoDevelop.Core;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


Therzok:
Why not use a tr(ansliteration) instead of a regex?

Using the below:
`Path.GetInvalidFilenameChars().Concat(Path.GetInvalidPathChars())`

We find out invalid path/filename chars. So we can foreach it and replace every char we find in there with a dash using `string.Replace`.

kdubau:
Yes you are right, should be using the `Path` calls. Updated the commit.

Therzok:
:+1: lgtm now. :D

PR #1072, opened 2015-09-28
[UITest] Updated locator for SelectPane

kdubau:
The locator for SelectPane in the OptionsController only worked on Mac, with this change it will work on both platforms.

CC: @manish


manish:
:shipit:

PR #1073, opened 2015-09-28
GtkWpfWidget implementation

Therzok:


Therzok:
cc @bratsche @slluis

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
```diff
@@ -201,6 +201,11 @@ public Document GetDocument (string name)
 					return false;
 				if (toplevel == RootWindow)
 					return true;
+				#if WIN32
+				var wpfWindow = System.Windows.Application.Current.Windows.OfType<System.Windows.Window>().SingleOrDefault (x => x.IsActive);
```
Therzok: We can't properly detect whether the Window is parented to a GtkWpfWindow. Possibly this needs to be changed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Windows/GtkWPFWidget.cs
```diff
@@ -0,0 +1,130 @@
...
+		{
+			base.OnShown ();
+
+			wpfWindow.Show ();
+			AttachWindow ();
+		}
+
+		void AttachWindow ()
+		{
+			IntPtr gtkWindowPtr = GtkWin32Interop.HWndGet (Ide.IdeApp.Workbench.RootWindow.GdkWindow);
```
Therzok: We're always attaching to the RootWindow. When we need to attach to an arbitrary window, we need to fix this.

alanmcgovern:
/cc @garuma 

Might be worth investigating to kill the Xwt based renderer in the designer. Maintaining only Wpf and XamMac would be nice :)

Therzok:
Proof from https://github.com/mono/monodevelop/tree/windowsToolbar

garuma:
Is it something we could move to Xwt instead through the wrap native widget facility? That would indeed make life easier for us to kill the xwt-based renderer.

Let's see what @bratsche thinks.

alanmcgovern:
This is analogous to `GtkNSView`, so it should be possible to provide as much XWT support for `GtkWpfWidget` as already exists for `GtkNSView`. We (or @bratsche ) can look into it post c6

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
```diff
@@ -1028,7 +1028,8 @@ public void ResetToolbars ()
 		protected override bool OnConfigureEvent (Gdk.EventConfigure evnt)
 		{
 			SetActiveWidget (Focus);
-			return base.OnConfigureEvent (evnt);
+			base.OnConfigureEvent (evnt);
+			return false;
```
Therzok: This was needed for window changes events to be propagated in the event we hook in the `GtkWPFWindow`.

bratsche:
Would it make sense/be possible for GtkWPFWidget to accept a `Control` instead of a `Window`, and then programmatically create a window and set the state (such as disabling window decorations) on it and add the control to it?

Therzok:
That could work, I can work on getting that.

PR #1074, opened 2015-09-28
Add ReproStep support for NuGet tests and other fixes

manish:


kdubau:
I think we should consistently use the imperative mood when writing Repro steps. ie. we should use 
> Update NuGet package "package name."

not
> Updating NuGet package "package name."

Other than that looks good. There was only one test (`AddPackagesTest`) that did not have and expected/validation step. Should we add and explicit "Ensure package is added to solution without error" step?

manish:
> Update NuGet package "package name."

not

> Updating NuGet package "package name."

If this generates reproduction steps, then it should sound like instructions making it clear that the user has to do it instead of software doing it for them,

kdubau:
Exactly my point..

On Tue, Sep 29, 2015 at 11:08 AM, Manish Sinha <notifications@github.com>
wrote:

> Update NuGet package "package name."
>
> not
>
> Updating NuGet package "package name."
>
> If this generates reproduction steps, then it should sound like
> instructions making it clear that the user has to do it instead of software
> doing it for them,
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mono/monodevelop/pull/1074#issuecomment-144089198>.
>



-- 
Thanks,
Kyle

manish:
@kdubau made the changes

kdubau:
:100:

PR #1075, opened 2015-09-28
[Projects] Portable Class Libraries should not be runnable by default.

Therzok:


Therzok:
All things considered, we _should_ probably do a check for custom commands and add another type check here, instead of overriding. https://github.com/mono/monodevelop/blob/pclExecutable/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs#L970

slluis:
Why is this necessary at all? Libraries are not executable by default.

Therzok:
No, this one ends up as 'true'. Because only `DotNetAssemblyProject` is counted as a library project.

https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs#L970-L971

slluis:
Let's add the additional type check. All this is not necessary in the roslyn branch.

Therzok:
Closed in favor of https://github.com/mono/monodevelop/commit/a3ebe7380cbdfc4656d984c0ce0ed8428f80a83e

PR #1078, opened 2015-09-30
GtkWpfWidget implementation

Therzok:


bratsche:
:+1:

I like it better now where it accepts a Control instead of a Window.

PR #1079, opened 2015-09-30
[IDE] Hide the updater with a hidden property key

iainx:
Setting the property MonoDevelop.Ide.AddinUpdater.HideFromUI will hide the Check For Updates menu

Fixes BXC #14627

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Updater/UpdateCheckHandler.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Updater/UpdateCheckHandler.cs
@@ -26,6 +26,7 @@
 
 
 using MonoDevelop.Components.Commands;
+using MonoDevelop.Core;
 using MonoDevelop.Ide.Updater;
 
 namespace MonoDevelop.Ide.Updater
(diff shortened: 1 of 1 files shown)


Therzok:
I think we can safely merge this.

slluis:
The updater also would need to check this flag, since it gets updates by default. Honestly, I don't think it is worth spending time on this feature, the use case is not very clear.

manish:
@slluis 

If this won't be merged, would it be possible to use an ENV VAR to disable it for Xamarin Studio automated tests. Updater sometimes steals focus and leads to failures of test.

So the update window not popping up when a specific env var is set would be a great help and we would be able to cut down significant amount of time from our tests (we have a workaround which takes around 10 seconds per test)

Therzok:
Sorry for the miscommunication on Slack. The fix is in the proprietary download service. We can add that there.

manish:
Marius, I see it is for a different thing, but I was hoping if we can have. We can cut down around 20 minutes in the UI test suite on wrench if we can be sure if update window wont open

slluis:
@manish you don't need an env var for this. When starting the test service, MD could just set MonoDevelop.Ide.Updater.UpdateService.AutoCheckForUpdates = false.

slluis:
Is this still required?

manish:
I don't think this is required.

dwhathaway:
@slluis, @manish Can we re-visit this feature?  This was requested for customers who want/need their teams to all be on the same version of the products or have company policies that prohibit end users from performing their own software updates.  The desire is to have a way to completely shut off any checks for updates from within the IDE.

This has come back up again with a potential customer, and we would like to be able to offer this as a solution.

PR #1083, opened 2015-10-05
FileLogger: use the testname as the logger, WaitForIdeIdle keeps blocked on certain messages

manish:
**[UITest] For UITest FileLogger use the testname as the logger**

This change is required because there are times when the test fails where XS crashes and doesn't get time for the logger to be cleaned up. Having different logger names for different tests would give other tests a chance to continue

---

**[UITest] WaitForIdeIdle does not unblock on specific messages. Extra messages can be added**

There are situations where the operation takes more than poll time and the method unblocks. Examples of such messages as "Solution loading..." or "Updatinng packages in solution..." etc. The method now also takes an optional argument to add more messages to block on if needed

------

**[UITest] Fix NuGet and Git tests by fixing timeouts and using gtk-sharp repo to clone**

### DIFF
--- a/main/tests/UserInterfaceTests/DialogTests/NuGetDialogTests.cs
+++ b/main/tests/UserInterfaceTests/DialogTests/NuGetDialogTests.cs
@@ -130,6 +130,7 @@ namespace UserInterfaceTests
 		}
 
 		[Test]
+		[Timeout (90000)]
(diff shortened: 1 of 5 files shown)


manish:
Yes, I have noticed that when XS crashes and there is a timeout in connecting, then the Logger is not disposed.

kdubau:
:100:

PR #1088, opened 2015-10-08
Bring back pad zooming

slluis:


Therzok:
Should this be closed?

PR #1094, opened 2015-10-12
[IDE] Switch tabs when selecting from the drop down

iainx:
Fix a typo that was setting the instance variable rather than the property when selecting a new tab

Fixes BXC #34640


slluis:
Can you create a PR against C6? We'll include it in SR1.

Therzok:
https://github.com/mono/monodevelop/pull/1095

PR #1097, opened 2015-10-14
[UITest] When test can't find a package, assert inconclusive

kdubau:


manish:
LGTM

kdubau:
@manish the Assert will end the test correct? Or should we be Asserting _and then_ throwing as well?

manish:
Asserting would end the test and call TearDown

kdubau:
:+1:

PR #1107, opened 2015-10-21
[AddIns] Disable deployment/deployment.linux/cbinding by default.

Therzok:
@slluis do we want ifdef'd code on the EnabledByDefault value?

```csharp
EnabledByDefault =
#if MAC || WIN32
    false
#else
    true
#endif
```

?


slluis:
Yep, let's do it conditionally.

Therzok:
Updated.

main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
```diff
@@ -340,7 +340,7 @@
 	<Command id = "MonoDevelop.Ide.Commands.ProjectCommands.ExportSolution"
 			defaultHandler = "MonoDevelop.Ide.Commands.ExportSolutionHandler"
 			_description = "Convert selected solution to another format"
-			_label = "_Export..." />
+			_label = "Convert solution format..." />
```
slluis: Let's keep this command as Export

Therzok: The document wanted this command gone. I thought renaming it to be more explicit was the way to go.

Cc @kallisto

PR #1108, opened 2015-10-21
Test] Fix the order in which TearDown operations are called

manish:


main/tests/UserInterfaceTests/UITestBase.cs
```diff
@@ -93,36 +93,39 @@ public virtual void SetUp ()
...
 						TakeScreenShot (string.Format ("{0}-Test-Failed", TestContext.CurrentContext.Test.Name));
 					} catch (Exception e) {
 						Session.DebugObject.Debug ("Final Screenshot failed");
+						Session.DebugObject.Debug (e.ToString ());
 					}
 				}
 
 				File.WriteAllText (Path.Combine (currentTestResultFolder, "MemoryUsage.json"),
-					JsonConvert.SerializeObject (Session.MemoryStats, Formatting.Indented));
+									   JsonConvert.SerializeObject (Session.MemoryStats, Formatting.Indented));
```
kdubau: what's with this weird extra space? other than that looks good.

PR #1112, opened 2015-10-26
[UITest] Added Cycle6 category to tests so that they can be ignored when running

manish:


kdubau:
looks good, we'll just need to add the minus prefix to the lane now

PR #1115, opened 2015-10-28
[MacPlatform] Fix the modal window detection for El Capitan

iainx:
When two windows share a fullscreen in El Capitan a special window is added that was causing us to flag it as modal which would disable the windows


Therzok:
Merged manually because I had to append "Bug 34899 - Most Menu Bar Buttons Disabled when in OSX Split Screen " to the commit message.

PR #1118, opened 2015-10-30
Roslyn msbuild generator support

sgmunn:


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -279,6 +279,96 @@ protected override Task OnLoad (ProgressMonitor monitor)
...
+		/// <summary>
+		/// Gets the source files that are included in the project, including any that are added by `CoreCompileDependsOn`
+		/// </summary>
+		public Task<ProjectFile[]> GetSourceFilesAsync (ConfigurationSelector configuration)
+		{
+			if (sourceProject == null)
+				return Task.FromResult (new ProjectFile [0]);
+
+			return BindTask<ProjectFile []> (async cancelToken => {
+				using (var monitor = new ProgressMonitor ()) {
```
slluis: Don't create monitors like this. GetSourceFilesAsync should take a monitor as argument.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -279,6 +279,96 @@ protected override Task OnLoad (ProgressMonitor monitor)
...
+			var results = new List<ProjectFile> (this.Files);
+
+			var dependsList = coreCompileDependsOn.Split (new [] { ";" }, StringSplitOptions.RemoveEmptyEntries);
+			foreach (var dependTarget in dependsList) {
+				try {
+					// evaluate the Compile targets
+					var ctx = new TargetEvaluationContext ();
+					ctx.ItemsToEvaluate.Add ("Compile");
+
+					var evalResult = await this.RunTarget (monitor, dependTarget, configuration, ctx);
```
slluis: This is a slow operation and the result should be cached.

sgmunn:
Doh, I knew I needed to implement caching but with the day off in between I totally forgot about it.

sgmunn:
ok, @slluis I just need to know where to get the progressMonitor from in MonoDevelopWorkspace to pass to GetSourceFiles. I also need to know how that would work with BindTask to ensure that RunTarget is cancelled when / if the Project is Disposed.

sgmunn:
@slluis One thing I'm not 100% sure of and that is how the android add-in will add the Generator to each file that needs it. What I think will happen is that the add-in will set the Generator property when the file is added that has a build action of AndroidResource and from what I understand, this will trigger the custom tool service to process the Generator - on the assumption that `FilePropertyChangedInProject` will trigger the tool when we set `Generator` for an item.

sgmunn:
@slluis how does this look now?

Therzok:
:trollface: unmergeable to me.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -279,6 +279,157 @@ protected override Task OnLoad (ProgressMonitor monitor)
...
+		/// <summary>
+		/// Returns a list containing FileInfo for all the source files in the project
+		/// </summary>
+		async Task<List<FileInfo>> GetProjectFileTimestamps (ProgressMonitor monitor, ConfigurationSelector configuration)
+		{
+			var infoList = new List<FileInfo> ();
+			var projectFiles = await this.GetSourceFilesAsync (monitor, configuration);
+
+			foreach (var projectFile in projectFiles) {
+				if (File.Exists (projectFile.FilePath)) {
```
Therzok: Can we hit the filesystem just once using [`new FileInfo(path)`](https://msdn.microsoft.com/en-us/library/system.io.fileinfo.fileinfo(v=vs.110).aspx) then using the .Exists property?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -279,6 +279,157 @@ protected override Task OnLoad (ProgressMonitor monitor)
...
+		/// Sends a file change notification via FileService for any file that has changed since the timestamps in beforeFileInfo
+		/// </summary>
+		async Task SendFileChangeNotifications (ProgressMonitor monitor, ConfigurationSelector configuration, List<FileInfo> beforeFileInfo)
+		{
+			var changedFiles = new List<FileInfo> ();
+			var projectFiles = await this.GetSourceFilesAsync (monitor, configuration);
+
+			foreach (var projectFile in projectFiles.Reverse ()) {
+				Console.WriteLine (projectFile.FilePath.FileName);
+				if (File.Exists (projectFile.FilePath)) {
```
Therzok: Same as above.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -279,6 +279,157 @@ protected override Task OnLoad (ProgressMonitor monitor)
...
+		/// <summary>
+		/// Sends a file change notification via FileService for any file that has changed since the timestamps in beforeFileInfo
+		/// </summary>
+		async Task SendFileChangeNotifications (ProgressMonitor monitor, ConfigurationSelector configuration, List<FileInfo> beforeFileInfo)
+		{
+			var changedFiles = new List<FileInfo> ();
+			var projectFiles = await this.GetSourceFilesAsync (monitor, configuration);
+
+			foreach (var projectFile in projectFiles.Reverse ()) {
+				Console.WriteLine (projectFile.FilePath.FileName);
```
Therzok: Debug text.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -279,6 +279,157 @@ protected override Task OnLoad (ProgressMonitor monitor)
...
+				foreach (var dependTarget in dependsList) {
+					try {
+						// evaluate the Compile targets
+						var ctx = new TargetEvaluationContext ();
+						ctx.ItemsToEvaluate.Add ("Compile");
+
+						var evalResult = await this.RunTarget (monitor, dependTarget, configuration, ctx);
+						if (evalResult != null && !evalResult.BuildResult.HasErrors) {
+							var evalItems = evalResult.Items.Select (i =>
+																		new ProjectFile (Path.Combine (sourceProject.BaseDirectory, i.Include), "Compile") { Project = this }
```
Therzok: Style: these are way off in terms of whitespace.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -279,6 +279,157 @@ protected override Task OnLoad (ProgressMonitor monitor)
...
+			return infoList;
+		}
+
+		/// <summary>
+		/// Sends a file change notification via FileService for any file that has changed since the timestamps in beforeFileInfo
+		/// </summary>
+		async Task SendFileChangeNotifications (ProgressMonitor monitor, ConfigurationSelector configuration, List<FileInfo> beforeFileInfo)
+		{
+			var changedFiles = new List<FileInfo> ();
+			var projectFiles = await this.GetSourceFilesAsync (monitor, configuration);
```
slluis: There is no need to get the file list again. The files you want to check are already in the beforeFileInfo list.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -279,6 +279,157 @@ protected override Task OnLoad (ProgressMonitor monitor)
...
+							changedFiles.Add (info);
+						}
+					} else {
+						changedFiles.Add (info);
+					}
+				}
+			}
+
+			foreach (var changedFile in changedFiles) {
+				FileService.NotifyFileChanged (changedFile.FullName);
```
slluis: Please use the overload that takes a list of paths. It allows notifying many file changes in a single event.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -279,6 +279,157 @@ protected override Task OnLoad (ProgressMonitor monitor)
...
+					} catch (Exception ex) {
+						LoggingService.LogInternalError (string.Format ("Error running target {0}", dependTarget), ex);
+					}
+
+					evaluatedCompileItemsWaitHandle.Set ();
+				}
+			} else {
+				// wait for evaluation
+				await Task.Run (() => {
+					evaluatedCompileItemsWaitHandle.WaitOne ();
```
slluis: This is a waste of resources. A more efficient way of doing this: create a TaskCompletionSource<ProjectFile[]> when performEvaluation=true, and keep it in the instance. Any additional thread that enters the method while evaluating will await on that completion source. When evaluation completes, set the result to the completion source. All tasks awaiting will get the result.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
sgmunn: If you mean the huge amount on line 408, I'm fighting XS with this one - every time it formats it sticks it way out there. it irks me too.

sgmunn:
Squashed and new PR - https://github.com/mono/monodevelop/pull/1126

PR #1121, opened 2015-10-31
Build fix for main/build/MacOSX/render.exe as gmcs no longer exists

sushihangover:
Building on OS-X with Mono 4.2.x that no longer sym-links a `gmcs` causes:

When bundling app on OS-X (with Mono 4.2.x that no longer sym-links a `gmcs`) results in:

```
pushd main/build/MacOSX
./make-dmg-bundle.sh
Building bundle for MonoDevelop 5.11.0.384...
Creating disk image (122MB)...
Attaching to disk image...
Populating image...
Cannot open assembly 'render.exe': No such file or directory.
mv: rename dmg-bg-with-version.png to MonoDevelop.mounted/.background/dmg-bg.png: No such file or directory
Detaching from disk image...
Creating distributable image...
Built disk image MonoDevelop-5.11.0.384.dmg
```

The root-cause of error via 'make render.exe'
```
pushd main/build/MacOSX
make render.exe
gmcs -r:System.Drawing render.cs
make: gmcs: No such file or directory
make: *** [render.exe] Error 1
```

### DIFF
--- a/main/build/MacOSX/Makefile.am
+++ b/main/build/MacOSX/Makefile.am
@@ -19,7 +19,7 @@ EXTRA_DIST = dmg-bg.png DS_Store Info.plist.in make-dmg-bundle.sh render.cs
 all: monostub monostub-test
 
 render.exe: render.cs
-	gmcs -r:System.Drawing render.cs
+	mcs -r:System.Drawing render.cs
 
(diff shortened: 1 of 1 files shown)


Therzok:
Thanks!

PR #1122, opened 2015-11-02
[NUnit] Add rerun command to the test pad.

Therzok:


mrward:
Code changes look good to me.

Only minor thing is that we have "Re-run test" in the context menu but "Rerun Tests" for the button in the Test Results pad. So I guess we should use Rerun in the new context menu label.

Therzok:
Sure thing. I can relabel it. I think we need to redesign the NUnit service to support not-reloading the whole test results when triggering a rerun.

main/src/addins/NUnit/Gui/TestResultsPad.cs
```diff
@@ -604,9 +604,19 @@ protected void OnShowTestCode ()
 			if (loc != null)
 				IdeApp.Workbench.OpenDocument (loc.FileName, loc.Line, loc.Column);
 		}
+
+		[CommandHandler (TestCommands.RerunTest)]
+		protected void OnRerunTest ()
+		{
+			UnitTest test = GetSelectedTest ();
```
DavidKarlas: Is SelectedTest==Last ran test?

Therzok: No, it's the test you select in the results pad.

DavidKarlas:
We got button for re-running already in result page...
Shouldn't logic/code be unified?

Therzok:
@DavidKarlas That reruns with the _root_ unit test. Possibly the code can be deduplicated, yes.

PR #1126, opened 2015-11-03
Roslyn msbuild generator support

sgmunn:


sgmunn:
this supersedes https://github.com/mono/monodevelop/pull/1118

sgmunn:
@Therzok @slluis -  new PR, all squashed up.

sgmunn:
@slluis I know this isn't a high priority atm, given I won;t need this until C8 I think, but it would be nice to know if we can merge this in before then.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
```diff
@@ -378,7 +378,6 @@ public void Save ()
 							FileService.NotifyFileChanged (fileName);
 						}
 						Window.ViewContent.Save (fileName);
-						FileService.NotifyFileChanged (fileName);
```
slluis: Why remove this?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -957,6 +968,38 @@ void OnFileRenamed (object sender, MonoDevelop.Projects.ProjectFileRenamedEventA
 			}
 		}
 
+		/// <summary>
+		/// Processes any genrator targets that the file might have associated with it
+		/// </summary>
+		async Task HandleGenerator (MonoDevelop.Projects.ProjectFile file, MonoDevelop.Projects.Project project)
```
slluis: The HandleGenerator method and the calls to this method should not be here, for two reasons:
1) It is not the responsibility of the type system. The type system parses code, it is not its job to generate code.
2) Roslyn only works for C#, but generators should work for other languages, such as F#

This should be handled in CustomToolService.

main/tests/test-projects/project-with-corecompiledepends/project.csproj
```diff
@@ -0,0 +1,100 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{7D68563A-E933-4AC0-A333-2F909720565B}</ProjectGuid>
+    <ProjectTypeGuids>{EFBA0AD7-5A72-4C68-AF49-83D382785DCF};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
```
slluis: We can't have tests on the monodevelop repo that depend on Xamarin products, since they need to run with a standalone MonoDevelop build (e.g. on Linux). You'll have to create a simple generator target to test this.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
sgmunn: The file service was sending multiple file change notifications when the document was saved. This would cause us to run the generator target twice on file save.

The view was changed to be responsible for sending the change notification instead of the document
https://github.com/mono/monodevelop/commit/01effaa665f6b498bf42057278615812188f7719

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
sgmunn: 1. I understand, I should have thought of that.
2. I don't follow how this is language specific - aren't we just running targets in the project, shouldn't that be supported in F#?

main/tests/test-projects/project-with-corecompiledepends/project.csproj
sgmunn: That's a copy+paste bug when I just copied the project I was working with. I would expect this to work with just a console project.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
slluis: ok

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
sgmunn: With 1) above, if we move this to a custom tool, we would then need to run custom tools on file removed and file added. I believe this is only done on file changed atm. Is that ok?

slluis: Yeah, we may need to add add/remove support to CustomToolService

sgmunn:
Something like this?

slluis:
LGTM

PR #1127, opened 2015-11-04
Removing obsolete solution pad tree popup button

vancura:
After a discussion in Boston last month we decided to remove the solution pad tree popup button, as it's not needed anymore. Here's the removal.


DavidKarlas:
Out of curiosity... What changed to go from useful to obsolete?

vancura:
It duplicated right click behavior and it was very non-standard (I don't recall any other case if a similar button in any application I've ever seen except MS DOS 3D Studio, do you know about any?)

vancura:
Also, did you use it? How often?

DavidKarlas:
Sometimes, not sure why... I guess my question is, why it was added/useful? :)

vancura:
It was in md since ever. No idea who has designed it.

vancura:
I guess it made much more sense when Macs had only one mouse button:)

vancura:
@sevoku is right, this shouldn't be based on dark skin. Once @slluis approves the PR (if he approves), I'll cherry pick it to master.

slluis:
Let's do this in dark-skin only for now

vancura:
OK.

Therzok:
:+1:

PR #1131, opened 2015-11-04
Add support for Run/Debug/Stop apps and checking Output windows text

manish:
Also fixes WaitForNoStatusMessage by actually getting the empty message from GetStatusMessage


Therzok:
Why are we no longer asserting that the icon changes?

manish:
I thought there were better methods to actually check if the debugger is running.

PR #1150, opened 2015-11-25
[roslyn branch] Add Suspend / Resume methods for IViewContent

hach-que:
This adds Suspend and Resume methods for IViewContent.  They are invoked when a tab is switched away from and when a tab is switched to.

For the Protobuild plugin, I embed another window within the tab.  When the tab is switched away from, it destroys the GTK widgets underneath the tab, and thus destroys the window handle.  I need to know when the tab is switched back to so I can reconstruct the window handle and pass it to the other application.


slluis:
I'm working on a redesign of IViewContent, which will be converted to a class, and it already has something like the Suspend / Resume methods. I'm planning to push it next week.

PR #1151, opened 2015-11-25
[roslyn branch] Allow display bindings to be registered at runtime

hach-que:
This allows display bindings to be registered at runtime.  This is useful when an extension (like the Protobuild add-in I'm developing) needs to add and remove display bindings as it becomes aware of more supported extensions.

In the Protobuild add-in, we support having projects in the solution provide IDE editors, and we load the compiled assemblies into an AppDomain or GtkPlug/Socket/separate process to allow developers to provide additional editors for the content of their own project.


slluis:
whitelist

slluis:
Looks good.

PR #1152, opened 2015-11-25
[roslyn branch] Allow derived types in Initialize

hach-que:
When deriving from a class that calls Initialize in it's constructor
(like Solution), the initialization does not occur because the derived
type does not match the type that calls Initialize.  However, if that
derived class then relies on being initialized later in the constructor,
an exception is thrown.

Instead, permit derived types to initialize in the base constructor (in
the use case this is for, we actually want the derived type in the add-
in to behave as the base type for almost every scenario).

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
@@ -53,7 +53,7 @@ namespace MonoDevelop.Projects
 
 		internal protected void Initialize<T> (T instance)
 		{
-			if (instance.GetType () != typeof(T))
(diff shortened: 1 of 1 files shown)


slluis:
The solution to the problem is to call Initialize() in the constructor of the derived class. This check exists because Initialize needs to be called when all the constructors of the class hierarchy have been invoked. Otherwise Initialize might end invoking a virtual method implemented in a class that has not yet been constructed.

hach-que:
I did, the problem is that Solution hits a line of code that requires
initialize to have been called before it even exits the constructor; the
derived class constructor doesn't even run before the exception is raised.

slluis:
Then that code should be moved to the OnInitialize method.

hach-que:
Okay, I'll modify Solution so that code is moved instead (and submit a separate PR).

PR #1154, opened 2015-11-25
[roslyn branch] Make OnRunTarget public for Project class

hach-que:
As per #1153, we need to redirect
operations performed on Protobuild objects (modules / definitions) to
the underlying MSBuild solutions and projects.  Although we can override
OnRunTarget in the Protobuild classes, we can't actually call
OnRunTarget on the MSBuild classes (that are being proxied / shadowed)
because it's internal protected.  In order to pass the OnRunTarget
operation through to the actual project, we need this method to be
public.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -854,7 +854,8 @@ namespace MonoDevelop.Projects
 		/// build or clean. The default implementation delegates the execution to the more specific OnBuild
 		/// and OnClean methods, or to the item handler for other targets.
 		/// </remarks>
-		internal protected virtual Task<TargetEvaluationResult> OnRunTarget (ProgressMonitor monitor, string target, ConfigurationSelector configuration, TargetEvaluationContext context)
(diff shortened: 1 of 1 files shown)


slluis:
I can't make this method public, since it is intended for internal use in the Project class. Why not just call RunTarget?

hach-que:
The problem is that RunTarget is a few levels higher than where the override of OnRunTarget happens; I don't believe it's possible to override RunTarget either.  Basically I'm concerned about side-effects of having the logic between "RunTarget" and the internal "OnRunTarget" occurring twice (once for the Protobuild definition that derives Project, and a second time for the MSBuild project).

slluis:
You can override OnRunTarget in a project subclass or in a ProjectExtension. If you do it in a ProjectExtension, the call from RunTarget is almost direct: https://github.com/mono/monodevelop/blob/roslyn/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs#L824

hach-que:
Checking now it seems like the code inside `DoRunTarget` won't interfere as much as I thought it might; I'll try just calling RunTarget instead and I'll see if that works for my purposes.

slluis:
Ok.

PR #1157, opened 2015-11-26
[roslyn branch] Add GetDefaultConfiguration behaviour

hach-que:
In Protobuild, we use project configurations as targetable platforms, instead of Debug / Release.  This is because there's no way for an add-in to add more drop-down menus to the toolbar.

By default MonoDevelop picks the first entry in the configuration in the list when opening a solution, which for Protobuild projects will be Android (because it's alphabetically sorted).  However, we actually want to default the configuration to the current host platform.

This adds a virtual method that allows MonoDevelop to query the current solutions to see if they have a preferred default configuration.


hach-que:
(btw it seems like the build server doesn't have the PCL assemblies installed, which is making this build fail)

directhex:
referenceassemblies-pcl package added to build script.

hach-que:
How do I get the build to re-run?

directhex:
build

hach-que:
Going back through the PRs, it doesn't look like the roslyn branch is passing for any of them (even the official PRs).

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Solution.cs
```diff
@@ -112,6 +112,11 @@ protected override void OnExtensionChainInitialized ()
 				NotifyItemReady ();
 		}
 
+		public virtual string GetDefaultConfiguration()
```
slluis: Please don't add public virtual methods to the Solution class. Instead, add a non-virtual GetDefaultConfiguration() method that calls a protected virtual OnGetDefaultConfiguration().

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/MainToolbarController.cs
```diff
@@ -318,6 +318,15 @@ void SelectActiveConfiguration ()
 				var confs = ToolbarView.ConfigurationModel.ToList ();
 				if (confs.Count > 0) {
 					string defaultConfig = ToolbarView.ActiveConfiguration != null ? ToolbarView.ActiveConfiguration.OriginalId : confs[0].OriginalId;
+
```
slluis: Why is this necessary? isn't GetBestDefaultConfiguration already called when loading a solution?

hach-que:
I'll have to close this PR and re-open it to change the target branch (since roslyn got merged into master).

PR #1160, opened 2015-11-27
[roslyn branch] Add GetRealProject() proxy method

hach-que:
In Protobuild, we want auto-completion and text highlighting to work in C# documents when they're opened.  But the types of projects in the solution are `ProtobuildStandardDefinition`, not `CSharpProject` or `DotNetProject`.

To support this, we need to have a `GetRealProject` method, which the type systems and text highlighters can use to get the actual underlying project.  The value of this method will change whenever the active configuration changes.

This doesn't cover everywhere I need to intercept this value; when I open a Protobuild project in MonoDevelop it still raises errors in text files about System.Object not being loaded.  If someone could point me any other locations I might missing to make this work, that'd be appreciated.


hach-que:
Looks like the build is failing because the build server needs Mono 4.2 or greater:

```
CSharp/CodeRefactorings/Uncategorized/UseStringFormatAction.cs(187,26): error CS1056: Unexpected character `$'
```

hach-que:
Resubmitting to target master branch.

PR #1162, opened 2015-11-27
[NUnit] Use ContextMenu for NUnit gutter menus

iainx:
Use the ContextMenu for the NUnit gutter menus so that the correct platform menus are used

Fixes BXC #34997


Therzok:
:+1:

PR #1163, opened 2015-12-03
[TextEditor] Fix memory leak.

mrward:
The TextEditorOptions StaticZoomChanged static event handler was
keeping each TextEditor instance alive even after they were closed.
This was also causing the project to be kept alive since the
TextEditor's PolicyBag references the project. The TextEditorFactory
creates a new TextEditor and then adds an anonymous delegate to the
TextEditor's ZoomLevelChanged event. This delegate references the
original TextEditor. The ZoomLevelChanged event maps to the
TextEditorOptions ZoomLevelChanged event which adds the delegate
to the static event. The static event then references the original
TextEditor and keeps it alive until the IDE is closed.

Making the ZoomLevelChanged event not directly reference the static
event fixes the problem since there is only an instance event handler
pointing to the TextEditor instead of a static event handler. The
other static event has to remain static so that changing the zoom in
one text editor is reflected in the other open text editors.

### DIFF
--- a/main/src/core/Mono.Texteditor/Mono.TextEditor/TextEditorOptions.cs
+++ b/main/src/core/Mono.Texteditor/Mono.TextEditor/TextEditorOptions.cs
@@ -95,7 +95,7 @@ namespace Mono.TextEditor
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


slluis:
@mkrueger hey, can you review?

PR #1166, opened 2015-12-04
Fixed bug in referenced assembly resolving

tomkcook:
The old implementation of this method attempted to resolve all libraries referenced by the template, even ones that had already been loaded by the runtime, and would sometimes return the wrong assembly (it would essentially always return whatever was last in the list).

This implementation should only load assemblies that can't be found through some other mechanism.


mrward:
It looks like the change now means the assemblyFiles list is no longer used. This list is taken from the TemplateSettings and also includes the host's StandardAssemblyReferences. So with this change these assemblies would not be used when resolving.

Probably need to review the problem Tom is having in more detail:

http://lists.ximian.com/pipermail/monodevelop-list/2015-December/016420.html
https://github.com/tomkcook/T4CSTest

tomkcook:
Okay; I'm going to open a new pull request to try to fix this properly, as I didn't realize my fork was several thousand commits behind when I started this one and it makes building the repository a pain.

tomkcook:
See #1168.

PR #1168, opened 2015-12-04
Added host.ResolveAssemblyReference() as a fallback for resolving assemblies not found in the directly-referenced assembly list

tomkcook:
This is a continuation of the discussion at #1166.  The problem with the code in master is that it will only find assemblies directly referenced by the template, not assemblies that those assemblies themselves reference.  This change adds `host.ResolveAssemblyReference()` as a fallback step after checking the reference assemblies listed in `assemblyFiles`, so that these 'second-step' dependencies can be found in the same directory as directly-referenced assemblies.

@slluis @mhutch @mrward


tomkcook:
Does this address the concerns raised with #1166 ?

mhutch:
Looks good to me.

Thank you!

PR #1169, opened 2015-12-04
[Core] Allow # and % characters in file names.

mrward:
Fixed bug #36432 - Cannot create new project c# console
https://bugzilla.xamarin.com/show_bug.cgi?id=36432

From the New Project dialog it was not possible to create a
project inside a directory that contained a '#' or a '%' even though
these are valid.


slluis:
This reverts https://github.com/mono/monodevelop/commit/300a9ddfcfa5a7e93ce9af74a4d4eb536da02d8d#diff-1031f5506e903a9f9111c9fb2d78b333R72

@Therzok why was that added?

Therzok:
@slluis project dialog shouldn't have allowed those items as valid inputs, nor should renaming a file by having those characters rename the class to have those. It was a way of sanitizing input for solution pad and file/project creation.

PR #1171, opened 2015-12-07
Bug 34569 - [Regression] Project references don't handle Condition attributes

DavidKarlas:
I'm pretty sure this is OK fix, but would prefer quick review by you @slluis 

Thank you


slluis:
Can you please add unit tests?

DavidKarlas:
Ok

DavidKarlas:
Updated with unit tests

slluis:
Thanks!

PR #1172, opened 2015-12-08
[IDE] Separate the notification model from the status bar

iainx:
Previously to place a message in the statusbar, it had to be accessed directly. This commit introduces the NotificationService which is an attempt to separate out the message data from the status bar.

The NotificationService contains NotificationContexts. These contexts are the data in the model which the Statusbar displays. Sitting in between is the StatusbarContextHandler which marshals the data from the NotificationService and the statusbar listens for changes.

The NotificationService handles multiple contexts correctly (Bug 30449), and the StatusbarContextHandler will switch between the contexts every 5 seconds.

This is an initial commit to discuss, it has only been tested on the OSX version of MD.
If the idea is ok, I'll check the Windows/Linux ports


iainx:
Do you have any comments, @slluis & @Therzok 

Splitting this off could give us a way to show all status messages the user has had, or do things like combine multiple progresses into one, and maybe change how transient messages work.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/StatusArea.cs
```diff
@@ -138,7 +134,33 @@ public StatusArea ()
...
+				case NotificationContextProgressChangedArgs.ProgressChangedType.Finish:
+					EndProgress ();
+					break;
+
+				case NotificationContextProgressChangedArgs.ProgressChangedType.Fraction:
+					SetProgressFraction (e.Work);
+					break;
+
+				case NotificationContextProgressChangedArgs.ProgressChangedType.Pulse:
+					// Nothing
```
Therzok: We need to set autopulse here. That makes the searchbar go round and round.

Therzok:
:+1: Looks good to me.

slluis:
I like the idea in general. Maybe the service could be named StatusService? I think it makes more sense. I will add some comments on the code.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
```diff
@@ -256,10 +256,10 @@ public void GrabDesktopFocus ()
 		public ProgressMonitorManager ProgressMonitors {
 			get { return monitors; }
 		}
-		
-		public StatusBar StatusBar {
+
+		public IStatusBar StatusBar {
 			get {
-				return workbench.StatusBar.MainContext;
+				return workbench.StatusBar;
```
slluis: Since IStatusBar now only has the ShowStatusIcon method, I don't think it is necessary to make it accessible here. We can move ShowStatusIcon to the status service class which would internally call workbench.StatusBar.ShowStatusIcon.

slluis:
What about context stacking? it doesn't seem like this implementation does it.

slluis:
BTW, we need unit tests for this.

iainx:
@slluis What do you mean by context stacking? This patch lets multiple contexts exist simultaneously and the status bar implementation flips through them every few seconds.

Currently it's not used in many places in the code because it never worked before, but the ability is there.

slluis:
For example, if I'm running a long operation such as building a solution, and then I run a fast operation like a search, the status area will switch to the most recent operation, and switch back to the previous operation when done.

iainx:
Right, in that case the different operations should create a new Context and use that to set their messages/progress and destroy them when they're finished because otherwise we don't really know when a message should be removed.

As an example, if I run a search and the status area says "Wrapped round to the beginning" and then run a build, once the build is finished should the status area say "Wrapped round to the beginning" again? It seems odd to me that such transient messages should be allowed to remain until something else has replaced them.

slluis:
> As an example, if I run a search and the status area says "Wrapped round to the beginning" and then run a build, once the build is finished should the status area say "Wrapped round to the beginning" again? It seems odd to me that such transient messages should be allowed to remain until something else has replaced them

That's not the case I'm describing. If I build a big solution the status area will show the build progress for several minutes. While building, I do a search on the solution. This may take for example 5 seconds. When I start the search operation I would expect to see the progress of the search in the status area, and when done, it would go back showing the progress of the build.

iainx:
Ok in that case then yes, the correct solution is for the code that does the build to call `StatusService.CreateContext ()` to get its own context, use that context to set the progress/any build messages and `Dispose` it once the build is finished.

When the search starts, it also calls `StatusService.CreateContext ()` to get a context that it owns, uses that to set progress/messages and calls `Dispose` once the search is finished.

That way while both the contexts are valid the status bar will flip between the different messages every few seconds and once one is disposed will just show the other message.

The MainContext IMO shouldn't exist, but because the context thing never worked before now, it isn't used in the code hardly at all, with the code using MainContext instead so that was what I ported for minimal code disruption.

slluis:
I'm not discussing what the code needs to do, but what is shown in the status area. Context stacking means that the status area shows the status of the operation at the top of the stack. This works well for the scenario I described in my last comment: the progress of the search would be shown while the search is running. It will be confusing if the status area suddenly flips to the build progress in the middle of the search.

iainx:
Ok, I can remove the flipping, it'll make the code easier :)

Therzok:
Can we get this into vNext?

slluis:
I'll review again, but the conflicts need to be resolved.

iainx:
Obsoleted by #3100

PR #1178, opened 2015-12-11
[Ide] When sorting SearchResultWidget by FileName also take "line" number into account

DavidKarlas:
Now it looks all nice and sorted instead of random:
image


DavidKarlas:
I'm pretty sure this is OK commit, just want quick review

mkrueger:
Yeah that'll work nicely - thanks

PR #1179, opened 2015-12-11
[Components] Prevent simultaneous command dispatches

iainx:
Disable command processing on OS X when a modal dialog is running

Fixes BXC #29519 - Multiple copies of the same modal dialog can be made visible at the same time


slluis:
What about Edit commands and Window management commands?

iainx:
Edit commands work as expected: If you open a new project dialog, all the menus are disabled, but cut/copy/paste still function in entry boxes.

I'm not sure what window management commands you mean: Cmd-M for minimise doesn't work when there's a modal dialog open with or without this patch

PR #1181, opened 2015-12-14
[IDE] Use the unsorted path to edit the model

iainx:
The path from the various editing args is relative to the sorted model, so we need to convert it to the unsorted model before accessing the data

Fixes BXC #36056


mrward:
This fixes the problem when editing the priority column when it is sorted. For some reason there is still a problem with the check box and the description column. I have updated the bug report.

iainx:
That's odd, the same fix has been applied to all three edit methods

iainx:
Hmm, but it's working ok for me. :/

mrward:
Looks like we have a winner now. Works good. :+1: :shipit:

PR #1183, opened 2015-12-15
[SourceEditor] Don't delete the history while changing the test

iainx:
When changing the history, we would completely delete the old entry, making it almost impossible to get a search term added to the Recent Searches menu

Fixes BXC #35928

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SearchAndReplaceWidget.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SearchAndReplaceWidget.cs
@@ -212,8 +212,17 @@ namespace MonoDevelop.SourceEditor
 				if (oldPattern != SearchAndReplaceOptions.SearchPattern)
 					UpdateSearchEntry ();
 				var history = GetHistory (seachHistoryProperty);
+
+				// Don't do anything to the history if we have a blank search
+				if (string.IsNullOrWhiteSpace (SearchPattern)) {
+					return;
+				}
+
 				if (history.Count > 0 && history [0] == oldPattern) {
-					ChangeHistory (seachHistoryProperty, SearchAndReplaceOptions.SearchPattern);
+					// Only update the current history item if we're adding to the search string
+					if (!oldPattern.StartsWith (SearchPattern)) {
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SearchAndReplaceWidget.cs
```diff
@@ -212,8 +212,17 @@ public SearchAndReplaceWidget (MonoTextEditor textEditor, Widget frame)
...
+
+				// Don't do anything to the history if we have a blank search
+				if (string.IsNullOrWhiteSpace (SearchPattern)) {
+					return;
+				}
+
 				if (history.Count > 0 && history [0] == oldPattern) {
-					ChangeHistory (seachHistoryProperty, SearchAndReplaceOptions.SearchPattern);
+					// Only update the current history item if we're adding to the search string
+					if (!oldPattern.StartsWith (SearchPattern)) {
```
Therzok: Better do a substring. What if we delete items at the beginning of the text?

slluis:
The history should be changed when clicking on Search, not every time the text entry changes.

slluis:
Discard my comment, for some reason I was thinking about the find in files dialog

iainx:
Yeah, I probably should have been more specific in the commit message, sorry

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SearchAndReplaceWidget.cs
iainx: My thinking is that we just want to protect against straight backspacing, other edits are more flexible.

Therzok: Yeah, well, deleting from the beginning should be the _same_ as backspacing.

iainx: Switching it to Contains makes things a lot more complex. I don't think there's a perfect solution to this problem. Keeping it simple with BeginsWith is better than what we have currently

PR #1187, opened 2015-12-16
Revert "[mdtool] Fix tools that depend on PlatformService"

alanmcgovern:
Reverts mono/monodevelop#1186


alanmcgovern:
@slluis @mhutch Actually it's probably simpler to revert this than change it's behaviour. If we need PlatfomService to work for some mdtools they should initialize it themselves or we should initialize it in a way which doesn't break DesktopService for tools which try to init it properly. Specifically this needs to be able to run: https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/DesktopService.cs#L80-L82

PR #1189, opened 2015-12-17
[Ide] Load projects into Roslyn only if they are C# or VB.net projects

DavidKarlas:
Also removed unused method
Fixes Bug 36963 - Error while getting C# recommendations

@slluis is this correct way of checking this, I can't check "project is CSharpProject", because we are inside MD.Ide.csproj project...?
@mkrueger this change makes sense, right?

Problem was, that also shared project was loaded inside Roslyn... which doesn't have references, files weren't updated... All kind of weird stuff...


slluis:
It is correct to check the project type like this, however the you should move that logic to a method like SupportsRoslyn(project).

DavidKarlas:
I updated commit with suggested change for SupportsRoslyn method, and since Mike won't be online for long time and I'm highly confident this change is good, I pushed fix to master.

PR #1192, opened 2015-12-20
Fix: Missing constructor prevents TextTemplatingSession value deseria…

ChristopherVB:
…lization in hosted AppDomain


mrward:
Adding this constructor to TextTemplatingSession does not match the [Microsoft's TextTemplatingSession documentation for the class](https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.texttemplating.texttemplatingsession.aspx). My guess is that the correct way to fix this, and to maintain the same class API, is to implement the ISerialization interface on the TextTemplatingSession class.

slluis:
The constructor can be private, the serialization system will call it anyways.

mrward:
I cherry-picked this commit to master and made the constructor private so the class API matches the documentation. I also added a unit test and implemented the ISerializable interface since otherwise the Id property did not get serialized across the app domain.

PR #1194, opened 2015-12-24
[Projects] Add imported configurations to master project

iainx:
When dealing with imported projects we use the properties to evaluate the master project and then we discard them. When the imported projects include configuration property groups they need to be added to the master project without being evaluated so the IDE can find them


slluis:
I have a branch that tries to solve this problem (https://github.com/mono/monodevelop/tree/roslyn-condprops), although in a different way. It uses a "conditioned properties" collection to store values of properties referenced in conditions while parsing the project. The original msbuild api also has this collection, and it is used by VS to generate the configurations. I haven't merged that branch yet because it has some issues. I'll finish it, and let's see if that fixes the project loading issue with your project.

iainx:
Ok, thanks.

slluis:
Alternative fix merged.

PR #1195, opened 2015-12-27
[Projects] Only add references if they can be built with the current configuration

iainx:
Only add a reference if it is buildable with the current configuration

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionItem.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionItem.cs
@@ -664,10 +664,16 @@ namespace MonoDevelop.Projects
 			if (!visited.Add(item))
 				return;
 
-			referenced.Add (item);
+			// Only add a reference if it can be built for the current configuration
+			foreach (var config in item.Configurations) {
+				if (configuration.ToString () == config.Id) {
+					referenced.Add (item);
 
-			foreach (var ritem in item.GetReferencedItems (configuration))
-				GetBuildableReferencedItems (visited, referenced, ritem, configuration);
+					foreach (var ritem in item.GetReferencedItems (configuration)) {
+						GetBuildableReferencedItems (visited, referenced, ritem, configuration);
+					}
+				}
+			}
 		}
 
 		internal bool ContainsReferences (HashSet<SolutionItem> items, ConfigurationSelector conf)


slluis:
What is this needed for?

slluis:
There is a unit test that contradicts this change: https://github.com/mono/monodevelop/blob/ee36637bfdc894f89b33dc32ebdbb01c4480b8b3/main/tests/UnitTests/MonoDevelop.Projects/ProjectTests.cs#L740

iainx:
I have a solution file that amongst other things has a c# project (ProjectA) that depends on a c++ vcxproj (ProjectB). ProjectB only has build configurations for a windows build, not for the mac build. 

The solution file lists ProjectB as a requirement for ProjectA, but as solution files don't appear to have any conditions as far as I could see, when building the solution using the Mac configuration XS attempts to build ProjectB anyway, and fails the build because it doesn't know how to.

The solution builds correctly with xbuild so I made this change so that XS wouldn't try to build ProjectB if it didn't have the appropriate configuration.

https://github.com/iainx/OpenLiveWriter/blob/Makefile/src/managed/writer.sln is the solution file

These are the relevant dependency lines: OpenLiveWriter is ProjectA and OpenLiveWriter.Ribbon is ProjectB

```Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OpenLiveWriter", "OpenLiveWriter\OpenLiveWriter.csproj", "{4C45353C-8B67-4E88-BCCC-94FECE0FEA46}"
	ProjectSection(ProjectDependencies) = postProject
		{195A60BF-7A4D-42E6-B5F4-FEBC679E19F0} = {195A60BF-7A4D-42E6-B5F4-FEBC679E19F0}
	EndProjectSection
EndProject
Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "OpenLiveWriter.Ribbon", "..\unmanaged\OpenLiveWriter.Ribbon\OpenLiveWriter.Ribbon.vcxproj", "{195A60BF-7A4D-42E6-B5F4-FEBC679E19F0}"
EndProject```

Further down the solution file you can see that OpenLiveWriter.Ribbon only has `Debug|Win32` and `Release|Win32` and no `DebugMac`/`ReleaseMac` configurations defined

mhutch:
Comparing the project configuration IDs look like a red flag. When building in the solution environment, it should really respect the solution configuration mappings.

PR #1196, opened 2015-12-29
Add GetDefaultConfiguration behaviour

hach-que:
Follow up from #1157 targeting the master branch instead.  Still have to update based on feedback in the previous PR.


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/MainToolbarController.cs
```diff
@@ -315,6 +315,15 @@ void SelectActiveConfiguration ()
 				var confs = ToolbarView.ConfigurationModel.ToList ();
 				if (confs.Count > 0) {
 					string defaultConfig = ToolbarView.ActiveConfiguration != null ? ToolbarView.ActiveConfiguration.OriginalId : confs[0].OriginalId;
+
+					foreach (Solution sol in IdeApp.Workspace.GetAllSolutions()) {
```
hach-que: This second call is necessary (in addition to `GetBestDefaultConfiguration`) because this is called when the solution is first loaded, while `GetBestDefaultConfiguration` is not.  They trigger under different scenarios, so adding the logic in both places is required.

slluis: AFAICT GetBestDefaultConfiguration is always called, event when the solution is first loaded. In which case it isn't called?

PR #1197, opened 2015-12-29
Add GetRealProject() proxy method

hach-que:
Follow up from #1160.

In Protobuild, we want auto-completion and text highlighting to work in C# documents when they're opened.  But the types of projects in the solution are `ProtobuildStandardDefinition`, not `CSharpProject` or `DotNetProject`.

To support this, we need to have a `GetRealProject` method, which the type systems and text highlighters can use to get the actual underlying project.  The value of this method will change whenever the active configuration changes.

This doesn't cover everywhere I need to intercept this value; when I open a Protobuild project in MonoDevelop it still raises errors in text files about System.Object not being loaded.  If someone could point me any other locations I might missing to make this work, that'd be appreciated.


slluis:
Can ProtobuildStandardDefinition be a subclass of DotNetProject?

hach-que:
No, because we need to swap out the DotNetProject instance when the current platform changes.

PR #1198, opened 2015-12-29
Bug 37248 - Roslyn: typing "<" after IEnumerable deletes the text in front of it.

DavidKarlas:
Fan fact: not roslyn specific bug
Basically fix is to replace text behind cursor only in case of Enter and Tab keys...


Therzok:
:+1:

PR #1201, opened 2015-12-30
Add packages restore for mono-addins in winbuild.bat

mfilippov:
Without it build on broken with error reference to SharpZipLib.

### DIFF
--- a/main/winbuild.bat
+++ b/main/winbuild.bat
@@ -1,3 +1,4 @@
 ".nuget\NuGet.exe" restore Main.sln
 "external\RefactoringEssentials\.nuget\NuGet.exe" restore external\RefactoringEssentials\RefactoringEssentials.sln
+".nuget\NuGet.exe" restore external\mono-addins\Mono.Addins.sln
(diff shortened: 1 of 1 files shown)


mrward:
Currently on Windows the [MonoLibraries.msi](https://files.xamarin.com/~jeremie/MonoLibraries.msi) is used to install SharpZipLib into the GAC. This is the assembly that Mono.Addins.Setup builds against when compiling MonoDevelop on Windows, which is a different version to that used in the packages.config file. If you install the MonoLibraries.msi then the build will work without the change you have made.

So I think restoring SharpZipLib from NuGet will cause different versions of this .dll to be used by different parts of MonoDevelop, unless we publish the version currently being used by MonoDevelop to NuGet and use that in Mono.Addins.Setup.

mfilippov:
@mrward I thought in external dependencies we should use version from dependency package.config same as RefactoringEssentials.

mrward:
@mfilippov - Yes, most of the time :) SharpZipLib is currently a special case.

mfilippov:
@mrward ok

PR #1202, opened 2016-01-01
Ukr

lxlalexlxl:
Ukrainian translation update


slluis:
This PR does not apply cleanly

PR #1206, opened 2016-01-07
Bug 36857 - [Roslyn] False positive errors referencing PCL project

DavidKarlas:
Fix is made from two parts...
1st part(TargetRuntime.cs) is fixing method FindFacadeAssembliesForPCL for MonoDroid frameworks since instead of having Facades in every framework folder it only has in v1.0 folder which is hardcoded in [.targets file](https://github.com/xamarin/monodroid/blob/200ed2f87bdc0933338c623aa32b4b868859870f/tools/msbuild/Xamarin.Android.Common.targets#L547)...
2nd part(MonoDevelopWorkspace.cs) is fixing logic to also check if project reference includes Portable projects...

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
@@ -878,7 +878,7 @@ namespace MonoDevelop.Projects
 				RemoteProjectBuilder builder = await GetProjectBuilder ();
 				var configs = GetConfigurations (configuration, false);
 
-				string[] refs;
+				string [] refs;
 				using (Counters.ResolveMSBuildReferencesTimer.BeginTiming (GetProjectEventMetadata (configuration)))
 					refs = await builder.ResolveAssemblyReferences (configs, CancellationToken.None);
 				foreach (var r in refs)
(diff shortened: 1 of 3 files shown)


DavidKarlas:
Proper fix would be that MSBuild ResolveReferences target returns also Facades dlls, but after 2 days I gave up on this quest and should be revisited after we switch from XBuild to MSBuild... imho...

slluis:
I'd like to have a proper fix for this. If the ResolveAssemblyReferences doesn't return the facade assemblies, maybe the targets for portable libraries use some other target for getting the facades. It might be worth checking if MSBuild returns the facade assemblies on Windows.

In any case, it is not the job of the type system to get the facade assemblies. The project model should do it. There is actually a comment about that in the code. So OnGetReferencedAssemblies should be overridden for portable libraries and should include the facade assemblies in the list. Also, the special case you added for Android should be handled by the Android flavor.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
```diff
@@ -914,6 +914,31 @@ public IEnumerable<DotNetProject> GetReferencedAssemblyProjects (ConfigurationSe
 				if (sa != null)
 					result.Add (sa.Location);
 			}
+			var addFacadeAssemblies = false;
+			foreach (var r in OnGetReferencedAssemblyProjects (configuration)) {
```
slluis: This should call GetReferencedAssemblyProjects, not OnGetReferencedAssemblyProjects

DavidKarlas:
Updated with OnGetReferencedAssemblyProjects->GetReferencedAssemblyProjects

PR #1207, opened 2016-01-08
[Ide] Store policies using the ID not the name.

Therzok:
Bug 10612 - Policies stored in file matching localized name, not ID


slluis:
Won't this break for existing policy files?

Therzok:
Actually, the only problem I see with this is that, given a policy that has already been localized, reserializing it would introduce a one-time duplicate.

Therzok:
We just iterate the policies folder for .mdpolicy files then load them, so that won't break anything existing.

Therzok:
But, on a further look on 'localizable' names, I think the only sample which exhibits this behaviour is XmlFormatting. Possibly we should remove the GettextCatalog.GetString calls there.

PR #1208, opened 2016-01-09
[Project] Don't write out imported configurations into the csproj file

iainx:
When a csproj file was being saved extra config lines were being written for imported configurations that didn't need to be there.

For example, with OpenLiveWriter the following would be added to each csproj file

<PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
   <IntermediateOutputPath>..\obj\Debug\OpenLiveWriter.Api</IntermediateOutputPath>
   <OutputPath>..\bin\Debug\i386\Writer</OutputPath>
</PropertyGroup>

for each configuration that was imported. These extra lines broke xbuild.

This patch prevents imported configurations being written to the csproj file.


slluis:
Alternative fix merged.

PR #1210, opened 2016-01-13
Bug 36857 - [Roslyn] False positive errors referencing PCL project

DavidKarlas:
This is different approach to https://github.com/mono/monodevelop/pull/1206 it uses "ResolveReferences".
Problems with this approach:
* If Portable project is not built(.dll is not in bin/Debug/projectName.dll), Facades are not resolved(Visual Studio has same problem) 
* Had to add `.Where (i => !i.Metadata.ContainsKey ("Project"))` to exclude now included project references.
* Probably slower since it executes more targets...

I would still prefer #1206.


slluis:
About point #1, do you mean that the editor will show unknown type errors for the system types until the project is built?

DavidKarlas:
Yes because the way MSBuild and XBuild discover if Facades are needed is to resolve PortableProject output filepath of .dll, open that .dll and check if it references System.Runtime.dll:
https://github.com/Microsoft/msbuild/blob/master/src/XMakeTasks/AssemblyDependency/ResolveAssemblyReference.cs#L2373
https://github.com/mono/mono/blob/master/mcs/class/Microsoft.Build.Tasks/Microsoft.Build.Tasks/ResolveAssemblyReference.cs#L362

PR #1212, opened 2016-01-13
[Ide] Fix two deadlocks in Project class.

Therzok:


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -463,7 +463,7 @@ public Task<ProjectFile[]> GetSourceFilesAsync (ConfigurationSelector configurat
 
 				if (string.IsNullOrEmpty (coreCompileDependsOn)) {
 					evaluatedCompileItemsTask.SetResult (new ProjectFile [0]);
-					return evaluatedCompileItemsTask.Task.Result;
+					return await evaluatedCompileItemsTask.Task;
```
slluis: Is this really a deadlock?

Therzok: I don't think it is, I can get rid of that.

Therzok: Just saw the SetResult above.

slluis: Just wondering if I missed something. Also, the first change is not a deadlock, although I'm ok with the change.

Therzok:
Manually pushed to master 1a6dd5dc0652190afeda1684dd3a6d7a090c4043

PR #1214, opened 2016-01-13
[Core] Introducing DotNetProject.GetReferencedAssemblyProject

DavidKarlas:
which returns only projects which output is actually referenced


alanmcgovern:
This has caused at least one regression: https://bugzilla.xamarin.com/show_bug.cgi?id=37690

Reverting the commit fixed that for me

PR #1217, opened 2016-01-18
 [source-editor] Set encoding of view to encoding used for writing.

lukaszunity:
Fixes issue with text view showing garbage characters (file loaded with wrong encoding) when using "Save As" and saving to a different encoding than the current view is using.

Repro steps on Windows:
1. Open UTF-8 encoded file.
2. Use "Save As" to save with new filename and select UTF-16 encoding (Default for me)
3. File is reloaded as UTF8 instead of UTF-16.


Therzok:
LGTM.

slluis:
Thanks!

PR #1224, opened 2016-01-19
[Debugger] Fix some leaks caused by the permanent busy dialog.

Therzok:
cc @DavidKarlas @jstedfast 

This is untested, as I don't know how to trigger the busy dialog.


DavidKarlas:
finnally managed to test this...

PR #1228, opened 2016-01-19
[Toolbox] Assume ItemToolboxNode.Name uses Pango markup formatting.

garuma:
@slluis I assume this change, even though not technically an API break, is a breach of contract so I would like to know what you think about it.

The rationale for this change is to be able to create toolbox item with a bit of style for conveying more information than simple text. My use case is to inform from which assembly/location a custom control is originating (see http://screencast.com/t/71JK7a21kD).

If you have a better idea on how I could achieve this better I'm open to suggestions.

NB: I noticed that toolbox is already having escaping issue because the tooltip that is created for those items already uses `SetMarkup` instead of `SetText` and that causes some text (e.g. clipboard ring of an XML file) to display incorrectly or not at all.


slluis:
A better solution would be to add a Source property or something like that, and let the toolbox render it

garuma:
I was going for the less disruptive approach in terms of API but if you think it's fine to break it I can redo a PR that does that.

slluis:
Well, adding a property to a class is not an api break

PR #1230, opened 2016-01-20
[AutoTest] Added a method for getting TaskService errors

kdubau:
cc: @manish


manish:
This can be done even without any enhancements in AutoTest. Look at the warnings/error counts in status bar.

manish:
Look at QualityAssurance repo, file: `XamarinStudioUtility.cs`

`XQA.XS.Controllers.XamarinStudio.IsBuildSuccessful` checks for errors. Also add checks for warnings.

kdubau:
It's a regex on the status message.. That should probably be updated to use `Session.ErrorCount()` - anyway this is more explicit and will give us the exact error, line, file, etc. without reinventing anything.

manish:
Yeah, the concept of UITest should be that the tests should be based on what is visible on the screen. With using `Session.ErrorCount()`, we are moving more and more towards non-UI tests

kdubau:
Disagree. The use case for this PR is to provide as much information as possible in the event of a failed assertion during a UITest. Feel free to ping me on slack to discuss further.

PR #1232, opened 2016-01-21
Async renaming

DavidKarlas:
This change should be reviewed by looking at commits separately...

* [Make all Renaming operations use Tasks/async/await instead of .Result](https://github.com/mono/monodevelop/commit/b00fa0bad2cea82dfd8991f17035deae1ed36eed)
 * This one is pretty straight forward, just introduce async Task to Rename method and propagate Task all the way back to events
* [[Refactoring] Changed GetSymbolInfoAsync to run on thread pool](https://github.com/mono/monodevelop/commit/4fea349627a50fc1c6efe98bcc3b602cbe68d6c9)
 * @mkrueger What do you think about this commit? I think this is best way to handle getting symbol info...
* [[Ide] Introduced MessageService.ExecuteTaskAndShowWaitDialog](https://github.com/mono/monodevelop/commit/db3dc0e6edfd56a9653e4690d38dc381bac90c0a)
 * @slluis and anyone else interested, can you review this commit and make comments about this new method, I think it's nice... Would appreciate comments...

Commits need to be merged simultaneously... because otherwise there is awkward delay between using Rename command and rename actually starting...(in past there was beachball instead of nice cancel dialog)

SS of dialog:
image

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Refactoring/CSharpFeaturesTextEditorExtension.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Refactoring/CSharpFeaturesTextEditorExtension.cs
@@ -89,9 +89,9 @@ namespace MonoDevelop.CSharp.Refactoring
 		}
 
 		[CommandHandler (EditCommands.Rename)]
-		public void RenameCommand ()
+		public async void RenameCommand ()
 		{
-			new RenameHandler ().Run (Editor, DocumentContext);
+			await new RenameHandler ().Run (Editor, DocumentContext);
 		}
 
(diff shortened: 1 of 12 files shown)


slluis:
It looks good to me, with some changes:
* Don't use "Waiting" in message dialogs, since it conveys that the IDE is waiting for some external event to happen, which is not the case. Instead of "Waiting for rename operation to resolve symbol", just show "Resolving symbol...", and instead of "Waiting for rename operation to find all references" show "Looking for all references".
* Maybe waiting 1s before showing the dialog is too much, unless there is some other visual feedback about the rename operation. But we can try it and see how it works.

mkrueger:
I think it's ok - having a popup is needed for now I guess.

PR #1236, opened 2016-01-27
[Ide] Second translation korean

powerumc:
It's my second translation work and fix.

- tranlsated all fuzzy.
- remove invalid translation.
- add translation, a few empty statments.


mhutch:
approve

PR #1237, opened 2016-01-27
Remove outdated docs and ChangeLogs

mhutch:
They fill search results with irrelevant information.

The docs are almost 10 years out of date, and almost
none of them apply to the current code.

The ChangeLogs are all in the git history, and haven't
been used for 5 years.


Therzok:
I agree, important docs are now on http://www.monodevelop.com/documentation/.

Therzok:
List of files removed: https://gist.github.com/Therzok/8d77192adf2dd565d1ed

But based on https://github.com/mono/monodevelop/blob/master/configure#L12, apparently profiles/ChangeLog is used to detect autoconf root or something.

ArsenShnurkov:
removal of ChangeLogs bring the following error:
https://bugzilla.xamarin.com/show_bug.cgi?id=40773

please fix that regression

knocte:
@ArsenShnurkov can you provide a PR please?

ArsenShnurkov:
i just used tocuh and forgot about it - https://github.com/ArsenShnurkov/shnurise-tarballs-automation/blob/master/dev-util/monodevelop/monodevelop-6.1_p2016062206.eroll#L32

PR #1238, opened 2016-01-27
Fix WPF toolbar for high-dpi displays

bratsche:
We need to convert the gtk widget's coordinates into the same units that
WPF is using.

Here's what it currently looks like:

highdpitest1

Here's what it looks like when I scale the coordinates:

screen shot 2016-01-27 at 1 37 43 pm


Therzok:
I see 2 issues. The Menu bar is now twice as tall and the text scaling in the statusbar/searchbar is huge. Any idea?

bratsche:
I think the issue is that when gtk is using a foreign window it's not handling DPI scaling.

PR #1240, opened 2016-01-28
[mono-debugger] Added tooltip for exceptions list.

lukaszunity:
Added hint for exceptions list in breakpoint dialog, explaining that is generated from the currently select project in solution pad.


DavidKarlas:
I see where need for such label is coming from, until recently(17 days ago): https://github.com/mono/monodevelop/commit/bec341a6738179598208ee2223322150586ed120 user was not able to add exception catchpoint if it was not in list... After that commit user can add any exception but it still shows warning icon if it's not found + code completion is not working... Also in case of project not selected it falls back to System.dll + corlib.dll exceptions list(before that commit that could fail).

So instead of accepting this PR, I extended warning text with better explanation...

Sounds like OK solution?

Therzok:
Superseded by https://github.com/mono/monodevelop/commit/3f1a581b9b316bf6c5dbb147789476a2b02cc4a1 I guess.

lukaszunity:
Looks good :+1:

PR #1242, opened 2016-01-29
Fix remaining MONODEVELOP_TEST_PROFILE usage

mhutch:


alanmcgovern:
This has made `make run` kinda unusable. If we really want this can you add an opt-out? Alternatively, if we don't really want this how about we just remove the flag and let the user manually export it if they want to test with a clean profile?

DavidKarlas:
I would also prefer `let the user manually export it if they want to test with a clean profile`

alanmcgovern:
https://github.com/mono/monodevelop/commit/a280fb364c71a5d087c0c2a5c773bd8689ee444d

I didn't remove it from the csproj files so it's still exported if you debug mono develop inside monodevelop. It should probably be removed from there too.

PR #1243, opened 2016-01-29
[NUnit+Core] Debug.WriteLine and tracepoints now work when debugging unit test

DavidKarlas:
Also reduced number of CreateExternalProcessObject overloads(removed unused)

Before this change it was impossible(afaik) to use tracepoints or Debug.WriteLine in unit tests...


Therzok:
Lgtm

DavidKarlas:
Added two more commits so it works also for MonoDevelop unit tests...

slluis:
LGTM

PR #1244, opened 2016-01-31
[Core] Move optional hintPath parameter from GAC to File reference method

DavidKarlas:
@slluis this change make sense, right?
both monodevelop and md-addins compile fine with this change(apparently nobody is using hintPath, I will use it soon :) )...

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectReference.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectReference.cs
@@ -152,14 +152,14 @@ namespace MonoDevelop.Projects
 			return new ProjectReference (asm);
 		}
 
-		public static ProjectReference CreateAssemblyReference (string assemblyName, string hintPath = null)
+		public static ProjectReference CreateAssemblyReference (string assemblyName)
 		{
-			return new ProjectReference (ReferenceType.Package, assemblyName, hintPath);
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectReference.cs
```diff
@@ -152,14 +152,14 @@ public static ProjectReference CreateAssemblyReference (SystemAssembly asm)
...
 
-		public static ProjectReference CreateAssemblyReference (string assemblyName, string hintPath = null)
+		public static ProjectReference CreateAssemblyReference (string assemblyName)
 		{
-			return new ProjectReference (ReferenceType.Package, assemblyName, hintPath);
+			return new ProjectReference (ReferenceType.Package, assemblyName, null);
 		}
 
-		public static ProjectReference CreateAssemblyFileReference (FilePath path)
+		public static ProjectReference CreateAssemblyFileReference (FilePath path, string hintPath = null)
```
slluis: Why add the hintpath argument here? the file path is the hint path

DavidKarlas:
Oh, OK I didn't realise hintPath is set automatically inside if it's null:
https://github.com/DavidKarlas/monodevelop/blob/hintPath/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectReference.cs#L179-L182

PR #1245, opened 2016-02-01
[Ide] Allow other project types to be supported by the type system.

mrward:
Currently only projects that have a type guid for C# or VB.NET are added to the type system. This is currently hard coded.

Now a project can indicate that it supports the type system. This allows other projects, such as DNX projects, which have a different type guid, to be supported by the type system and code completion will then work for C# files in these projects.

This is a rework of commit: a49d6220af5f369dcc243dde21c018b20755a846


slluis:
It seems like SupportsRoslyn() should be defined in DotNetProject, not Project.
The SupportsRoslyn() override should be done by CSharpProject and VBProject.

mrward:
Moved the logic from Project to DotNetProject. The CSharpProject and the VBProject now indicate they support Roslyn. Also changed SupportsRoslyn to be a property instead of a method.

PR #1248, opened 2016-02-04
[Core] Make WebRequestHelper.GetResponseAsync actually async

iainx:
Instead of just passing the synchronous GetResponse to Task.Factory.StartNew, use the async version of all the functions instead

Fixes BXC #37284


slluis:
LGTM /cc @mrward

mrward:
LGTM too. :shipit:

PR #1250, opened 2016-02-06
[Ide] Removed 600ms delay from StartReparseThread + few code cleanups

DavidKarlas:
Reasoning here is this, user experience with 600ms delay is suffering mostly noticeable is SemanticHighlighting but also other stuff that is executed based on DocumentChanged with 600ms delay which is not ideal...
Second reasoning why this is acceptable is... It's parser job to cancel parsing based on CancellationToken and not abuse ThreadPool/CPU... If this change is problem... Parser should be reviewed to cancel faster...

@mkrueger Can you review this?
Changes in CSharpBinding/MonoDevelop.CSharp.Parser/TypeSystemProvider.cs:

* It's my understanding C# files are not Serialisable anymore, correct? Hence removed logic excluding them from `ParsedDocumentFlags.NotSerializable` based on IsCompileBuildAction... Instead making them `ParsedDocumentFlags.NotSerializable` always...
* I wrapped `CSharpSyntaxTree.ParseText` into try catch(cancelation) because I added cancellationToken...
* Generating `compilerArguments` only if needed
* I assume `result.LastWriteTimeUtc` was set for serialization reasons?(hence I removed logic)

Changes in Document.cs:

* At begining of `EnsureAnalysisDocumentIsOpen` I swapped
```
if (analysisDocument != null)
    return SpecializedTasks.EmptyTask;
```
and
```
if (editor == null) {
    UnsubscibeAnalysisdocument ();
    return SpecializedTasks.EmptyTask;
}
```
because otherwise UnsubscibeAnalysisdocument had no effect...

Rest of changes are just refactoring

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Parser/TypeSystemProvider.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Parser/TypeSystemProvider.cs
@@ -43,15 +43,8 @@ namespace MonoDevelop.CSharp.Parser
 			var fileName = options.FileName;
 			var project = options.Project;
 			var result = new CSharpParsedDocument (fileName);
+			result.Flags |= ParsedDocumentFlags.NonSerializable;
 
-			if (project != null) {
-				
-				var projectFile = project.Files.GetFile (fileName);
-				if (projectFile != null && !TypeSystemParserNode.IsCompileBuildAction (projectFile.BuildAction))
-					result.Flags |= ParsedDocumentFlags.NonSerializable;
-			}
-
-			var compilerArguments = GetCompilerArguments (project);
 			SyntaxTree unit = null;
 
 			if (project != null) {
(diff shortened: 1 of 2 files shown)


mkrueger:
Reparse delay was for NR5 limitations - however I think we can still need it for F# - try to ask the F# devs about that. Not every parser is incremental & cancelable.

DavidKarlas:
I think more important thing is cancelable, incremental is just bonus...
@nosami @7sharp9 Will F# parser cancel fast enough, so it doesn't process multiple old text changes at once and kill CPU cores?

7sharp9:
The F# compiler is cancellable on any change that makes the current file dirty.

@mkrueger To make the caching of F# partial parses better it would be nice if the current documents version number was available, this would allow a more finer grained cache in the F# compiler.

mkrueger:
You can always track an editor version number. The editor itself has not really a model of version numbers. It uses a text change tracking system - maybe you can use that instead ?

mkrueger:
Ok we'll try it - from the looks it shouldn't break IDE experience. But test F# @7sharp9

nosami:
This PR destroys the F# plugin. I can't open an editor, never mind start to type.

Please revert until I get chance to look into it.

alanmcgovern:
This may have triggered, or helped uncover, https://bugzilla.xamarin.com/show_bug.cgi?id=38524

DavidKarlas:
I reverted this commit...

* If I understand correctly F# FSI parser doesn't support cancelation token here: https://github.com/fsharp/xamarin-monodevelop-fsharp-addin/blob/047edac/MonoDevelop.FSharpBinding/FSharpParser.fs#L144
Hence every keypress starts new parsing which is not canceled on next keypress and it kills CPU...
* About bug Alan found with deadlock.. Problem is that CreateProject("C#") needs to execute some code on UIThread and I moved EnsureAnalysisDocumentIsOpen code off UI Thread some other code called this method from UI thread and deadlock appeared...

nosami:
I also noticed another issue. Holding down the backspace key was very jittery in C# files.

PR #1253, opened 2016-02-08
[Ide] Rename "OK" button to "Save" in options dialogs

hbons:


slluis:
Why Save? I don't think I've ever seen a dialog with a Save button.

PR #1254, opened 2016-02-08
[NUnit] Removes markup from test names

nosami:
This was needed to fix the rendering of F# test cases. Somewhere deep
inside nunit.core, the test name was being converted from something
like
    `TestName("param1", "param2")`
to
    `TestName<String>("param1", "param2")`

This was causing havoc in the Test pad.


slluis:
This doesn't look right. Don't we want to show that markup in the tests pad?

nosami:
No. It looks like a generic method signature, but it isn't. 

The test name in my example is TestName and the TestCase is `("param1", "param2")` and returns unit or void.

nosami:
Test cases in F# now render the same as the equivalent C# test.

PR #1255, opened 2016-02-08
Remove CBinding

mhutch:
It doesn't work any more - the build system hasn't been updated to
MSBuild. New projects don't even have any build configurations.

Removing since it's still shipping, though disabled, and conflicts with
the forked and updated version at https://github.com/mhutch/cbinding


decriptor:
With things like the android ndk would it be better to refresh the c binding addin?

Therzok:
See the CBinding that's under michael's account. It supports clang and cmake and lots of other nice things the MonoDevelop one doesn't.

decriptor:
Then drop this one and bring in the one under michael's account?

alanmcgovern:
Maintenance might be easier if it were left in its current location and shipped out of band as an external plugin for now :)

mhutch:
If we remove this one, I can ship my fork on addins.monodevelop.com (once I actually fix the build system).

I see no reason to bundle it. It's been completely unmaintained (except API tracking) for ~5 years. Bundling it will make maintenance _harder_.

decriptor:
Cool, that makes sense to me.

Will it just be an addin that ends up just tracking API changes too?

mhutch:
Ideally it'll get developed more actively, get bugfixes, and bitrot fixes. This will depend on whether anyone steps up to do that... but that didn't happen when it was part of the MD source tree, so it certainly can't get any worse as a standalone addin. Hopefully the standalone addin will be more accessible that setting up a full MD build environment :)

Last year we had two GSoC students working on the standalone addin. I was able to merge their changes much more readily because I wasn't as concerned about stability as I would be for a built-in addin.

decriptor:
I agree. Makes much more sense as something that is separate. I know there is an org/set of repos that have addins. Is it worth pushing more addins there and promoting them?

mhutch:
@lluis bump

PR #1258, opened 2016-02-11
Moved ConfigurationChanged and UpdatePathText inside ActiveConfiguration.set

manish:
Not sure if this the right way to do things, but this is the only way in which UITest can actually set configurations.

Without this patch, only the UI is updated since the UITest was just setting `ActiveConfiguration` property, but this change was not propagated all along.

I used `ConfigurationChanged (this, EventArgs.Empty)` since I noticed that `(object o, EventArgs e)` were not being used anywhere


Therzok:
LGTM.

PR #1262, opened 2016-02-12
[IDE] Don't let pad label overlap icon

iainx:
Stop the pad label overlapping the icon when there's not enough space allocated for the whole tab.

Returns to the pre-dark skin behaviour of cropping the label.

Fixes BXC#38011


sevoku:
LGTM, but it doesn't solve the bug actually, since the text overlaps the buttons and the tab has no min size.

iainx:
No, I agree it doesn't fix the problem, I just wanted to get it back to how it always was before for the C7 release. I hadn't realised you had something better

sevoku:
Hhm, it takes too long to fix it correctly. I merge this PR to have the old behavior.

PR #1266, opened 2016-02-17
[MacToolbar] Map the Cocoa coords -> Gdk correctly so that search pop…

iainx:
…up appears in correct place

Fixes BXC #37191


main/src/addins/VersionControl/MonoDevelop.VersionControl.Subversion/gtk-gui/objects.xml
```diff
@@ -1,304 +1 @@
-<objects attr-sync="on">
-  <object type="MonoDevelop.VersionControl.CommitMessageStylePanelWidget" palette-category="General" allow-children="false" base-type="Gtk.Bin">
```
Therzok: Why was this removed?

PR #1267, opened 2016-02-18
[Components] Protect against NaN in the Dock layout file

iainx:
The EditingLayout.xml in bug 32385 contains NaN for certain widths which caused 1 warning and 1 assertion.
Protect against both by checking for NaN before proceeding, and limiting widths to >= 1

Fixes BXC #32385


alanmcgovern:
Does this also fix https://bugzilla.xamarin.com/show_bug.cgi?id=15641 ?

Therzok:
@alanmcgovern I think it does not. That looks like the real issue that happens at runtime, not from a loaded xml file.

iainx:
Should we back port this to a hypothetical future C6 release?

PR #1269, opened 2016-02-19
[IDE] Save layouts when they're changed

iainx:
Save layouts after a short delay after they've been changed, rather than waiting until the workspace window is closed

Fixes BXC #25609
Fixes BXC #33009


Therzok:
LGTM.

But: 25609 is talking about open files, and those are saved to userprefs. So 25609 isn't fixed by this.

decriptor:
:+1:  This'll be nice to have!

iainx:
Changed the commit message to not reference 25609.

slluis:
I think this implementation is abusing events a bit. DockGroup could be more simple: in DockObject.OnAllocationChanged you can just call parentGroup?.OnAllocationChanged(), no need to subscribe child events.

DockNotebook events can also be simplified. We don't really need to know when something is added/removed, a global event that is raised when something changes is enough (no need to chain notebook or page events).

iainx:
Ok, I've rebased and removed the majority of the event stuff

PR #1274, opened 2016-02-23
[IDE] Only run the updater automatically if it isn't turned off

iainx:
The daily check for updates was ignoring if the automatic updater checkbox was checked or not


Therzok:
@grendello will be happy :)

PR #1278, opened 2016-02-29
[MacToolbar] Have the search shortcut in the search bar

iainx:
Fixes BXC #38945`

### DIFF
--- a/main/src/addins/MacPlatform/MainToolbar/SearchBar.cs
+++ b/main/src/addins/MacPlatform/MainToolbar/SearchBar.cs
@@ -155,6 +155,18 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 			}
 		}
 
+		static readonly string firstResponderPlaceholder = "Search";
+		string placeholderText = "Search";
+		public string PlaceholderText {
+			get {
+				return placeholderText;
+			}
+			set {
+				placeholderText = value ?? "Search";
+				PlaceholderAttributedString = MakePlaceholderString (placeholderText);
+			}
+		}
+
 		public SearchBar ()
 		{
 			Cell = new DarkSkinSearchFieldCell ();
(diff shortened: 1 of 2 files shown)


main/src/addins/MacPlatform/MainToolbar/SearchBar.cs
```diff
@@ -155,6 +155,8 @@ public override void SelectWithFrame (CGRect aRect, NSView inView, NSText editor
 			}
 		}
 
+		static readonly string firstResponderPlaceholder = "Search";
+		static readonly string placeholder = "Press '⌘.' to search";
```
Therzok: This is not the right fix.

We have [this](https://github.com/iainx/monodevelop/blob/fix-38945/main/src/addins/MacPlatform/MainToolbar/MainToolbar.cs#L318-L321)

Therzok: This comes from the keybinding service and handles setting the text to the correct one. If we want to sanitize it to show a nice ⌘ symbol, sure.

iainx: Oh, how in my refactoring of the toolbar did I never notice that...

iainx:
Fixed

main/src/addins/MacPlatform/MainToolbar/SearchBar.cs
```diff
@@ -155,6 +155,18 @@ public override void SelectWithFrame (CGRect aRect, NSView inView, NSText editor
...
 		}
 
+		static readonly string firstResponderPlaceholder = "Search";
+		string placeholderText = "Search";
+		public string PlaceholderText {
+			get {
+				return placeholderText;
+			}
+			set {
+				placeholderText = value ?? "Search";
```
slluis: This "Search" text should be localizable

Therzok: It is. `MakePlaceholderString` calls `GettextCatalog.GetString`.

slluis: The GetString call in MakePlaceholderString is useless if the string is not marked as translatable, and it won't be marked as such unless there is an explicit GettextCatalog.GetString("Search") call.

iainx: I've fixed this up

slluis: It's not fixed, you need to add a GettextCatalog.GetString("Search") call.

Therzok: https://github.com/mono/monodevelop/pull/1287/files

PR #1279, opened 2016-02-29
[MacToolbar] Have the search shortcut in the search bar - [c7]

iainx:
Fixes BXC #38945`

### DIFF
--- a/main/src/addins/MacPlatform/MainToolbar/SearchBar.cs
+++ b/main/src/addins/MacPlatform/MainToolbar/SearchBar.cs
@@ -155,6 +155,23 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 			}
 		}
 
+		static string FirstResponderPlaceholder {
+			get {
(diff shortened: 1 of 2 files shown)


main/src/addins/MacPlatform/MainToolbar/SearchBar.cs
```diff
@@ -155,6 +155,8 @@ public override void SelectWithFrame (CGRect aRect, NSView inView, NSText editor
 			}
 		}
 
+		static readonly string firstResponderPlaceholder = "Search";
+		static readonly string placeholder = "Press '⌘.' to search";
```
Therzok: As noted in the other PR, this is not the correct fix.

We have [this](https://github.com/iainx/monodevelop/blob/fix-38945/main/src/addins/MacPlatform/MainToolbar/MainToolbar.cs#L318-L321).

iainx:
Fixed :)

PR #1281, opened 2016-02-29
[MacToolbar] Ellipsize text if the string is very long [c7]

iainx:
If the runtime or configuration names are > 50 chars long ellipsize them so that the selector doesn't switch immediately to minimode

Fixes BXC 39018


Therzok:
:+1:

PR #1284, opened 2016-03-01
[NUnit] - allow the tests to run without being in debug mode

nosami:
testContext.ExecutionContext.ExecutionHandler was null when not being ran under the debugger.


slluis:
@nosami when doing changes in code somebody else maintains, please wait for a review before merging.

PR #1290, opened 2016-03-08
Bug 39075 - XS is now writing empty properties to my debug and release property groups in android projects

DavidKarlas:
Based on debugging I came to conclusion to add this code which fixes that bug...
Since I don't have enough understanding in this part of XS I can't be certain I didn't introduce new bug, remove some feature or anything like that...
@slluis can you review this?


DavidKarlas:
This was wrong approach to solving this...

PR #1291, opened 2016-03-08
[Core] Do not show warning dialog for known unsupported projects.

mrward:
Fixed bug #36960 - Adding/Opening unsupported project shows 'Could
not load...' warning pop up.
https://bugzilla.xamarin.com/show_bug.cgi?id=36960

Do not show a warning dialog when opening a solution containing
unsupported but known project types. A warning dialog is now only
shown for unknown project types or known unsupported project types
that cannot have their files loaded into the solution window.


Therzok:
:+1:

slluis:
LGTM

PR #1294, opened 2016-03-09
[ide] Show missing GTK# installation error in native dialog on Windows.

lukaszunity:
If GTK# is not installed on Windows and you start MonoDevelop, then you get only get an error in the log file and MonoDevelop does not start.

This change shows a native Windows dialog if MonoDevelop fails to start.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
@@ -597,6 +597,26 @@ namespace MonoDevelop.Ide
 						"may not be properly installed in the GAC.",
 						BrandingService.ApplicationName
 					), ex);
+
+				if (Platform.IsWindows)
+				{
+					string url = "http://monodevelop.com/Download";
+					string caption = "Fatal Error";
+					string message =
+						"{0} failed to start. Some of the assemblies required to run {0} (for example GTK#) " +
+						"may not be properly installed in the GAC.\n\r\n\r" +
+						"Please click OK to open the download page, where " +
+						"you can download the necessary dependencies for {0} to run.";
+
+					if (DisplayWindowsOkCancelMessage(
+						string.Format(message, BrandingService.ApplicationName, url), caption)
+					)
+					{
+						Process.Start(url);
+					}
+				}
+
+
 			} finally {
 				Runtime.Shutdown ();
 			}


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
```diff
@@ -597,6 +597,26 @@ public static int Main (string[] args, IdeCustomizer customizer = null)
 						"may not be properly installed in the GAC.",
 						BrandingService.ApplicationName
 					), ex);
+
+				if (Platform.IsWindows)
```
Therzok: Shouldn't https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs#L334-L377 suffice?

lukaszunity:
It does not look like that check is enough. I just uninstalled GTK# and get an exception thrown before entering `IdeStartup.Run`.

image

mrward:
With GTK# not installed the gtk check is not reached. When the IdeApp's constructor is initialized it tries to use the FontService and throws an exception.

   ```
   "System.TypeInitializationException: The type initializer for 'MonoDevelop.Ide.IdeApp' threw an exception. ---> System.TypeInitializationException: The type initializer for 'MonoDevelop.Ide.Fonts.FontService' threw an exception. ---> System.DllNotFoundException: Unable to load DLL 'libpango-1.0-0.dll': The specified module could not be found. (Exception from HRESULT: 0x8007007E)\r\n   at Pango.FontDescription.pango_font_description_new()\r\n   at Pango.FontDescription..ctor() in c:\\MinGW\\msys\\1.0\\home\\builder\\gtk-sharp\\pango\\generated\\FontDescription.cs:line 291\r\n   at MonoDevelop.Ide.Fonts.FontService..cctor() in src\\core\\MonoDevelop.Ide\\MonoDevelop.Ide.Fonts\\FontService.cs:line 43\r\n   --- End of inner exception stack trace ---\r\n   at MonoDevelop.Ide.Fonts.FontService.GetFontProperty(String name)\r\n   at MonoDevelop.Ide.IdePreferences..ctor() in src\\core\\MonoDevelop.Ide\\MonoDevelop.Ide\\IdePreferences.cs:line 129\r\n   at
 MonoDevelop.Ide.IdeApp..cctor() in src\\core\\MonoDevelop.Ide\\MonoDevelop.Ide\\Ide.cs:line 110\r\n   --- End of inner exception stack trace ---"
```

```
   at MonoDevelop.Ide.IdeApp.set_Customizer(IdeCustomizer value)
   at MonoDevelop.Ide.IdeStartup.Run(MonoDevelopOptions options) in src\core\MonoDevelop.Ide\MonoDevelop.Ide\IdeStartup.cs:line 85
   at MonoDevelop.Ide.IdeStartup.Main(String[] args, IdeCustomizer customizer) in src\core\MonoDevelop.Ide\MonoDevelop.Ide\IdeStartup.cs:line 593
   ```

Not sure if it is possible to move the gtk check above the code that configures the IdeApp. Maybe instead the FontService should be lazily used so it is not called in the IdeApp's constructor so the gtk# check can be run afterwards but before the FontService is used.

lukaszunity:
I think it would be nice for the users if a dialog appeared when the exception is caught, regardless of the GTK# check later. Instead of failing silently with an error written to the log.

mrward:
If the code that sets the IdeApp's Customizer is moved after the GTK# check then everything seems to work. With GTK# not installed you get a dialog which will take you to a web page with the GTK# installer can be downloaded. It does not silently fail after this change is made.

lukaszunity:
Alright. Should I make the changes to this PR or will you fix it directly?

mrward:
I can make the change.

lukaszunity:
Thanks!

PR #1295, opened 2016-03-10
Fix files open race

lukaszunity:
Make sure solution is fully loaded before loading other files. Fixes race condition with file opening from launch arguments.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
@@ -46,6 +46,7 @@ using MonoDevelop.Components.AutoTest;
 using MonoDevelop.Ide.TypeSystem;
 using MonoDevelop.Ide.Extensions;
 using MonoDevelop.Ide.Templates;
+using System.Threading.Tasks;
 
 namespace MonoDevelop.Ide
 {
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
```diff
@@ -338,7 +341,14 @@ public static void OpenFiles (IEnumerable<FileOpenInformation> files)
...
 				}
 			}
-			
+
+			// Wait for solution and it's open files to load, so we are sure
+			// that the files we open afterwards are actually opened in tabs
+			// after the solution's saved open files and that the last file
+			// in the filteredFiles gets focus, if specified as an option.
+			if (filteredFiles.Count > 0 && openTaskOperation != null)
+				openTaskOperation.Wait ();
```
slluis: This will block the UI thread, a continuation should be used here. Or maybe OpenFiles can be converted to async.

lukaszunity: Don't the OpenDocument calls also block the UI thread?

slluis: OpenDocument is now async. But in any case, the problem is calling Task.Wait, which should never be done in the UI thread, since it can cause deadlocks.

lukaszunity: I see, so are you suggesting to open all the remaining files inside a `ContinueWith` call on the `Task`?

alanmcgovern: We could also just make it an `async` method and use an `await` on `openTaskOperation` if we want that form of delayed execution. It's simpler and more readable than continuations

slluis:
Closing since there is a more complete fix in PR #2415.

PR #1298, opened 2016-03-16
Bug 39557 - [XS 6.0] "Could not load assembly Perhaps it doesn't exist in the Mono for Android profile?"

DavidKarlas:
When attempting to build immediately after restoring NuGet packages.

@slluis This makes sense, right?
Problem was that DefaultLocalCopy returned False, because File.Exists("..\Packages\NLua.Android\file.dll") returned false. Which resulted in localCopy field being set and even after NuGets were restored it didn't change to True... DefaultLocalCopy did change but localCopy was already set, hence not copying .dlls and bug... With this change we set localCopy only if it's actually written in .csproj file otherwise LocalCopy property returns DefaultLocalCopy...


slluis:
The change looks correct to me. Just make sure the unit tests pass before merging.

DavidKarlas:
ran unit tests & pushed directly

PR #1307, opened 2016-03-21
[IDE] Disable maximize/minimize on Preferences dialog

iainx:
Fixes BXC #38503


slluis:
Iain, this should be encapsulated in a method in the GtkWorkarounds class.

iainx:
I've done that in https://github.com/mono/monodevelop/pull/1310

PR #1309, opened 2016-03-21
[NuGet] insert .props at top of project file

mrward:
Adds support to insert an Import for .props in a NuGet package at the start of the project.

Added a fix for an extra newline being inserted when MSBuildProject's AddNewImport is called passing a beforeObject which is the first item in the project. This particular [commit](https://github.com/mono/monodevelop/commit/e834096c420d5c3938402f3333d0608932ee4123) could do with being reviewed.


slluis:
LGTM

PR #1310, opened 2016-03-21
[IDE] Disable minimize/maximise for dialogs on OS X

iainx:
Improved the method of disabling minimise/maximise for dialogs, and apply it to all IdeDialog derived classes


main/src/core/MonoDevelop.Ide/MonoDevelop.Components/GtkWorkarounds.cs
```diff
@@ -1279,6 +1284,42 @@ public static void SetTransparentBgHint (this Widget widget, bool enable)
 		{
 			SetData (widget, "transparent-bg-hint", enable);
 		}
+
+		static void OnMappedDisableButtons (object sender, EventArgs args)
+		{
+			var window = sender as Gtk.Window;
```
Therzok: How would `window` not be Gtk.Window here? `(Gtk.Window)sender;` cast is more appropriate.

PR #1317, opened 2016-03-27
Treeview speedups

Therzok:
To be taken with a grain of salt for now. It's orthogonal with https://github.com/mono/gtk-sharp/commit/f4d4443e4eb4ee130bbf786bd1718fafab44a524 and https://github.com/mono/gtk-sharp/commit/decbc5ac25d389d474f06e57dbaa1e99ef96d26c, but combined they improve the situation a lot.

I have no idea if this causes a slowdown with the current gtk-sharp, but it shouldn't do so.

Those patches above brought down building the TreeBuilder of libgit2sharp with one open file from 2 seconds down to 1.2s. Combined with https://github.com/mono/monodevelop/commit/18bb88f6c827e470837f239cd695d6af9ec769c8, it  _further_ optimized it down to 300ms.

That said, I thought it would be a good idea to replace occurrences of AddChild with AddChildren, as AddChildren is guarded against multiple resorting. This should reduce a few cycles on sorting, after the above commits in gtk-sharp are merged.

Might be a good idea to look at tweaking that a bit more.


Therzok:
cc @slluis

slluis:
LGTM

PR #1318, opened 2016-03-27
[Ide] Asynchronize OnTheFly formatting.

Therzok:
This patch asynchronizes all the way up for `OnTheFlyFormatter`'s `Format` and `FormatStatmentAt` methods.

The reason for this patch is to fix editor choking on formatting a block and keypresses.


Therzok:
Don't merge yet, this seems to have broken completion popup on first keypress.

Therzok:
Fixed.

PR #1319, opened 2016-03-28
[MacPlatform] Use the OS X theme colour to decide the dock icon colour

iainx:
Use the OS X theme to decide the colour of the icon in the dark, not the XS theme

@vancura, in case you think this is better suited?


vancura:
LGTM! I'll provide new icons ASAP.

vancura:
Reverted in https://github.com/mono/monodevelop/commit/a6576eca9eb2a8935359ef702b21580419fd31e3

PR #1323, opened 2016-04-01
[Ide] Fix editor traversal in command target chain

Therzok:
This commit fixes the traversal of the editor in the command target
chain, to take into account the traversal of the editor extensions too.

The SourceEditorView does not route into the TextEditor, which is in
fact the real iterator of the text editor extensions in the chain.

Bug 39660 - Upon changing "Build Action" property grid gets reset


Therzok:
cc @mkrueger @slluis

Therzok:
Another solution would be to make it a `IMulticastCommandRouter` and send both for traversal.

Therzok:
Scrap that, the Editor itself will always route to the `SourceEditorView` and the extension chain.

PR #1325, opened 2016-04-04
Bug 39716 - When a reference has the Version attribute, Xamarin Studio displays a red cross

DavidKarlas:


main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/AssemblyContext.cs
```diff
@@ -339,13 +343,21 @@ public string FindInstalledAssembly (string fullname, string package, TargetFram
 			return bestMatch;
 		}
 
-		static SystemAssembly BestFrameworkAssembly (IEnumerable<SystemAssembly> assemblies, TargetFramework fx)
+		public Version ParseVersion (string version)
```
slluis: Should be private.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/AssemblyContext.cs
```diff
@@ -323,14 +325,16 @@ public string FindInstalledAssembly (string fullname, string package, TargetFram
...
+				var fxAsm = BestFrameworkAssembly (asms, fx, version);
 				if (fxAsm != null)
 					return fxAsm.FullName;
 			}
 
 			string bestMatch = null;
 			foreach (SystemAssembly asm in asms) {
-				if (fx.CanReferenceAssembliesTargetingFramework (asm.Package.TargetFramework)) {
+				if (fx.CanReferenceAssembliesTargetingFramework (asm.Package.TargetFramework) &&
+				    (version == null || version == ParseVersion(asm.Version))) {
```
slluis: Maybe it would be better to convert 'version' to string once instead of doing ParseVersion(asm.Version) many times.

slluis:
It is not clear to me which problem is this patch trying to fix, and how that solves bug 39716. Can you please explain a bit?

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/AssemblyContext.cs
DavidKarlas: not sure how this turned into public :(

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/AssemblyContext.cs
DavidKarlas: idea is... it's safer to compare Version, since strings `4.4` is not same as `4.4.0.0`...

DavidKarlas:
Bug is actually fixed by changing NormalizeAsmName to ParseAssemblyName so assembly name is correctly extracted... but I also fixed that if version is not matching to what framework has to show it as invalid, hence comparing version...

PR #1327, opened 2016-04-04
[Debugger] Set the debug layout after executing the process.

Therzok:
Bug 3770 - Application output open while editing execution parameters


slluis:
How does that prevent the application output from showing up in the current layout?

Therzok:
The approach is wrong. I thought I fixed it, but it was actually using a different project that used an external console. The Console is created within the execution system and forced to front by default, not counting whether it has a prompt.

PR #1337, opened 2016-04-05
Command update checks

Therzok:


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -2245,24 +2245,24 @@ public void Run (object cmdTarget, CommandInfo info)
 			if (customHandlerChain != null) {
 				info.UpdateHandlerData = Method;
 
-				DateTime t = DateTime.Now;
```
slluis: We should actually use Stopwatch

Therzok: I'm probably pedantic here, but DateTime.UtcNow is faster than Stopwatch based on my profiling, and we don't need nanosecond resolution. I didn't see it necessary to switch to stopwatch for that.

slluis: Is it so faster so that it is worth losing accuracy?

Therzok:
Manually merged to master with `Stopwatch` instead of DateTime.UtcNow.

PR #1344, opened 2016-04-06
[MacToolbar] Don't cut off the searchbar when in fullscreen mode

iainx:
Listen to the superview's frame changing and adjust the layout accordingly.

Fixes BXC #40099


Therzok:
Lgtm.

PR #1346, opened 2016-04-06
[IDE] Use a lock when scheduling the workspace saving.

iainx:
The code which scheduled the workspace saving had a potential race condition due to the timer executing the delegate on a background thread. Use an async lock so it won't block the UI thread

Fixes BXC #40140

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
@@ -1305,21 +1305,29 @@ namespace MonoDevelop.Ide.Gui
 		}
 
 		System.Timers.Timer tabsChangedTimer = null;
-		void WorkbenchTabsChanged (object sender, EventArgs ev)
-		{
-			if (tabsChangedTimer != null) {
-				tabsChangedTimer.Stop ();
-				tabsChangedTimer.Dispose ();
-			}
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
```diff
@@ -1305,21 +1305,29 @@ public void ReparseOpenDocuments ()
...
-				await IdeApp.Workspace.SaveAsync ();
-				tabsChangedTimer.Stop ();
-				tabsChangedTimer.Dispose ();
-				tabsChangedTimer = null;
-			};
-			tabsChangedTimer.Start ();
+				tabsChangedTimer = new System.Timers.Timer (10000);
+				tabsChangedTimer.AutoReset = false;
+				tabsChangedTimer.Elapsed += async (s, e) => {
+					using (tabsChangedLock.Enter ()) {
```
Therzok: Please use a local function and unregister the event registration when disposing.

alanmcgovern: An inline function means you don't have to unregister. Wouldn't it be easier to just leave it here?

alanmcgovern: Weelll, either way you don't have to unregister because the timer is being disposed/nulled anyway

PR #1347, opened 2016-04-06
[IDE] Use a lock when scheduling the workspace saving. [c7]

iainx:
The code which scheduled the workspace saving had a potential race condition due to the timer executing the delegate on a background thread. Use an async lock so it won't block the UI thread

Fixes BXC #40140

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
@@ -1302,6 +1302,35 @@ namespace MonoDevelop.Ide.Gui
 					doc.ReparseDocument ();
 			}
 		}
+
+		System.Timers.Timer tabsChangedTimer = null;
+
+		void DisposeTimerAndSave (object o, EventArgs e)
+		{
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
```diff
@@ -1302,6 +1302,35 @@ public void ReparseOpenDocuments ()
...
+				tabsChangedTimer.Dispose ();
+				tabsChangedTimer = null;
+
+				IdeApp.Workspace.SavePreferences ();
+			});
+		}
+
+		void WorkbenchTabsChanged (object sender, EventArgs ev)
+		{
+			if (tabsChangedTimer != null) {
```
alanmcgovern: you could replace all of this with:

Task tabsChangedTask;
async void WorkbenchTabsChanged (object sender, EventArgs ev)
{
    if (tabsChangedTask != null)
         return;
    tabsChangedTask = Task.Delay (10000);
    await tabsChangedTask;
    tabsChangedTask = null;
    IdeApp.Workspace.SavePreferences ();
}

PR #1348, opened 2016-04-06
[IDE] Fix race conditions when scheduling preferences save

iainx:
There was a potential race condition when scheduling the preferences save after the workbench tabs had changed because the timer uses threads. Use an AsyncCriticalSection to avoid blocking the UI thread

Fixes BXC #40140

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
@@ -1305,20 +1305,31 @@ namespace MonoDevelop.Ide.Gui
 		}
 
 		System.Timers.Timer tabsChangedTimer = null;
+
+		void DisposeTimerAndSave (object o, EventArgs e)
+		{
+			Runtime.RunInMainThread (() => {
+				tabsChangedTimer.Stop ();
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
```diff
@@ -1305,21 +1305,32 @@ public void ReparseOpenDocuments ()
...
+		void DisposeTimerAndSave (object o, EventArgs e)
 		{
-			if (tabsChangedTimer != null) {
+			using (tabsChangedLock.Enter ()) {
 				tabsChangedTimer.Stop ();
+				tabsChangedTimer.Elapsed -= DisposeTimerAndSave;
 				tabsChangedTimer.Dispose ();
+				tabsChangedTimer = null;
 			}
+			IdeApp.Workspace.SavePreferences ();
```
slluis: Let's simplify a bit. SavePreferences() must be called from the UI thread, in which case you can also do the tabsChangedTimer disposal in the UI thread, in which case you don't need a lock.

slluis:
Looks good now!
BTW, if you push additional changes to a PR, please add a comment, since new commits don't generate any mail notification.

PR #1350, opened 2016-04-06
[Ide] TreeView now takes into account different font sizes

Therzok:
Bug 37814 - Unnecessary tooltip in solution pad

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/ExtensibleTreeView.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/ExtensibleTreeView.cs
@@ -2419,8 +2419,7 @@ namespace MonoDevelop.Ide.Gui.Components
 				return icon.WithSize (size);
 			}
 
-
-			void SetupLayout (Gtk.Widget widget)
+			void SetupScaledFont ()
 			{
 				if (scaledFont == null) {
 					if (scaledFont != null)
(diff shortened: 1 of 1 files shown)


Therzok:
@slluis The previous behaviour would calculate a row's width based on the font size at the construction of the widget - which is later changed by setting it as per Ide Theme. With this change, the layout now takes into account 'zooming' which changes font size and fixes the width calculation on Mac, where the font size is 11 (in the layout - as opposed to the Widget's context with 12).

slluis:
Hmm, changing the tree font every time a cell is rendered doesn't look right to me. The problem seems to be that the cell size is not properly calculated, in which case it should be fixed in the GetSize() override.

Therzok:
Does this look better?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/ExtensibleTreeView.cs
```diff
@@ -2494,6 +2499,19 @@ public Gdk.Rectangle GetStatusIconArea (Gtk.Widget widget, Gdk.Rectangle cell_ar
 			public override void GetSize (Gtk.Widget widget, ref Gdk.Rectangle cell_area, out int x_offset, out int y_offset, out int width, out int height)
 			{
 				base.GetSize (widget, ref cell_area, out x_offset, out y_offset, out width, out height);
```
slluis: We have all the information to calculate the size, so calling base.GetSize() shouldn't be necessary. All we need to do is use the layout to calculate the size of the text and add the margins.

Therzok:
Superseded by https://github.com/mono/monodevelop/pull/1354

PR #1353, opened 2016-04-07
[Ide]Improve checks for if ToolTipWindow should be shown

Sl0vi:
The tooltip window often appears when the intellisense list is empty or even
after it has been closed.

These changes greatly reduce the likelyhood that this will happen, altho there
is still a tiny chance that it could.

Some screenshots of this happening:

Tooltip showing when intellisense completion list is not:
tooltip-no-intellisense

Occasionally it shows as a white box:
white-tooltip-invalid-syntax

This happens so often that it makes MonoDevelop almost unbearable to use. 

I can still get this to happen occasionally after my changes, but I have to deliberately try to provoke the issue with some well timed keyboard mashing.


Sl0vi:
This CLA makes no sense. I am not self employed, but my employer has no right to or even a right to know about contributions I make in my own time.

Why is there no I am employed AND have sole rights to my contributions option?

Therzok:
This looks good to me.

Therzok:
Manually merged in 5f388130b4a140c1b5aa9051bdd33cbf41481a4b.

PR #1354, opened 2016-04-07
[Ide] TreeView now takes into account different font sizes

Therzok:
Bug 37814 - Unnecessary tooltip in solution pad


slluis:
It looks good now. Thanks!

PR #1357, opened 2016-04-11
[CSharpBinding] Navigation text editor extension no longer does

mkrueger:
request links on the UI thread.

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp/CSharpNavigationTextEditorExtension.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp/CSharpNavigationTextEditorExtension.cs
@@ -43,18 +43,18 @@ namespace MonoDevelop.CSharp
 	{
 		static List<NavigationSegment> emptyList = new List<NavigationSegment> ();
 
-		protected override Task<IEnumerable<NavigationSegment>> RequestLinksAsync (int offset, int length, CancellationToken token)
+		protected override async Task<IEnumerable<NavigationSegment>> RequestLinksAsync (int offset, int length, CancellationToken token)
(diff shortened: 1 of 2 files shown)


main/src/addins/CSharpBinding/MonoDevelop.CSharp/CSharpNavigationTextEditorExtension.cs
```diff
@@ -43,21 +43,23 @@ class CSharpNavigationTextEditorExtension : AbstractNavigationExtension
...
-				visitor.Visit (await model.SyntaxTree.GetRootAsync (token).ConfigureAwait (false));
-				return visitor.result;
-			} catch (OperationCanceledException) {
-				return emptyList;
-			}
+				return TaskUtil.EmptyEnumerable<NavigationSegment> ();
+			return Task.Run (delegate {
+				try {
+					var visitor = new NavigationVisitor (DocumentContext, model, new TextSpan (offset, length), token);
+					visitor.Visit (model.SyntaxTree.GetRootAsync (token).Result);
```
alanmcgovern: You should probably still use an `await` with `.ConfigureAwait (false)` here. We only have a catch for `OperationCancelledException` but if you touch the `Result` property you'll get an `AggregateException` wrapper.

alanmcgovern: And we should probably make master do the same if this commit was cherry-picked from there :)

slluis: Makes sense.

PR #1359, opened 2016-04-12
[Papercut][Ide] Focus Search result pad on search dialog closed

sevoku:
When the "Find in Files" dialog is being closed, the Search Result Pad disappears (if it is not docked).
Fix: Search dialog retrieves the result pad from the monitor and focuses it when the Close button has been activated.

(fixes bug #37577)


Therzok:
LGTM!

slluis:
:+1:

PR #1366, opened 2016-04-13
[Refactoring] Reduced the amount of thread pool threads required for

mkrueger:
background code analyzation.


slluis:
Let's try it

PR #1368, opened 2016-04-13
Fixed 'Bug 40316 - [Cycle 7]User is not able to Search dll in Assembly

mkrueger:
Browser.'

widget.CecilLoader.LoadAssembly isn.t thread safe so it needs to run
on the main thread.


slluis:
Wouldn't be enough to add a lock to avoid concurrent use of CecilLoader or LoadAssembly?

mkrueger:
I'll try a lock .

PR #1381, opened 2016-04-14
[Core] Fix ProcessWrapper

VincentDondain:
Fixed bug #40379.

Avoid "System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'handle'" when calling HasExited.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/ProcessWrapper.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/ProcessWrapper.cs
@@ -96,10 +96,10 @@ namespace MonoDevelop.Core.Execution
 				if (endEventErr != null)
 					endEventErr.WaitOne ();
 
-				if (HasExited)
-					operation.ExitCode = ExitCode;
-
 				try {
+					if (HasExited)
+						operation.ExitCode = ExitCode;
+
 					OnExited (this, EventArgs.Empty);
 				} catch {
 					// Ignore


Therzok:
👍

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/ProcessWrapper.cs
```diff
@@ -96,10 +96,10 @@ private void CaptureOutput ()
...
 					endEventErr.WaitOne ();
 
-				if (HasExited)
-					operation.ExitCode = ExitCode;
-
 				try {
+					if (HasExited)
+						operation.ExitCode = ExitCode;
+
 					OnExited (this, EventArgs.Empty);
```
alanmcgovern: Is it OK that we'd then silently stop emitting the `OnExited` event?

slluis: I don't think so. OnExited should be called in any case.

VincentDondain: That's a good question but not related to that particular fix.

VincentDondain:
@slluis can we merge that? It's a high critical C7 bug of death.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/ProcessWrapper.cs
alanmcgovern: Just wrap that `operation.ExitCode` thing in it's own try/catch if we do rely on OnExited to always be invoked. It's pretty relevant to this patch.

VincentDondain: Oh didn't see lluis' comment and misunderstood your comment @alanmcgovern. Makes perfect sense let me fix that.

VincentDondain:
Fixed in f66a5502cd6399007aec863cfe2ca5132c15e55e

PR #1383, opened 2016-04-15
Recent open concurrency

Therzok:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentFileStorage.cs
```diff
@@ -207,22 +209,51 @@ static bool RemoveMatches<T> (List<T> list, Func<T,bool> predicate)
...
+					}
+
+					modified = true;
+				}
+
+				if (modified) {
+					fs.Position = 0;
+					fs.SetLength (0);
+					WriteStore (fs, list);
+					OnRecentFilesChanged (list);
```
Therzok: This makes recent file changed event to happen as late as possible, but it shouldn't be a problem. We keep both multiple-instance concurrency via `AcquireFileExclusive` lock, and we batch as many modifications as possible in a 1 second window.

slluis: If you are commenting code, do it in the code. Comments in PRs and commits will be lost.

slluis:
There is also a problem in AcquireFileExclusive. It does a Thread.Sleep of 1s. It should be converted to a Task.Delay. AcquireFileExclusive should be an async method.

Therzok:
Updated, had to async all the way up because of `GetRecentItemsInGroup`.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileCommands.cs
```diff
@@ -267,9 +267,9 @@ protected override void Update (CommandInfo info)
 	// MonoDevelop.Ide.Commands.FileCommands.RecentFileList
 	public class RecentFileListHandler : CommandHandler
 	{
-		protected override void Update (CommandArrayInfo info)
+		protected override async void Update (CommandArrayInfo info)
```
slluis: Command update handlers can't be async.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
```diff
@@ -229,15 +229,23 @@ int Run (MonoDevelopOptions options)
...
-					openedProject = DesktopService.RecentFiles.GetProjects ().FirstOrDefault ();
-					if (openedProject != null)
-						IdeApp.Workspace.OpenWorkspaceItem (openedProject.FileName).ContinueWith (t => IdeApp.OpenFiles (startupInfo.RequestedFileList), TaskScheduler.FromCurrentSynchronizationContext ());
-				}
-				if (openedProject == null)
-					IdeApp.OpenFiles (startupInfo.RequestedFileList);
-				
+				Task.Run (async () => {
+					RecentFile openedProject = null;
+					if (IdeApp.Preferences.LoadPrevSolutionOnStartup && !startupInfo.HasSolutionFile && !IdeApp.Workspace.WorkspaceItemIsOpening && !IdeApp.Workspace.IsOpen) {
```
slluis: This api may not be thread safe

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileCommands.cs
Therzok: The solution I guess would be to cache the list inside RecentFiles and mirror whatever changes we write to the file in the cache, right? We can keep multiple process synchronization by refreshing the cache on read via file synch lock.

Therzok:
Superseded by #1424

PR #1384, opened 2016-04-15
[Ide] Fix concurrency on defaultmonitorpad accessing.

Therzok:
Bug 40373 - Null reference exception in ScrollView

cc @alanmcgovern

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs
@@ -487,6 +487,9 @@ namespace MonoDevelop.Ide.Gui.Components
 		void addQueuedUpdate (QueuedUpdate update)
 		{
 			lock (updates) {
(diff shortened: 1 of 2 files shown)


Therzok:
I assume the NRE was coming from a monitor pad being reused after it was being disposed, as a side-effect of properly disposing of pads in https://github.com/mono/monodevelop/commit/abf8bcaa7046978c7eb65e9f423665d314decdcf

slluis:
GetPadForMonitor should only be called from the UI thread, in which case the lock should not be necessary.

Therzok:
Fixed review comment and added safeguard against addins manually destroying pads. The progress monitor completion was happening after its underlying pad was destroyed.

Therzok:
After further investigation, the problem is that the pad outlives the monitor in the case of the resgen pad, because the pad is manually destroyed before the monitor is disposed. That causes the monitor to try and queue the completed message. 65ba2fd safeguards against this case now, that the pad is no longer getting leaked.

Therzok:
Bump.

PR #1388, opened 2016-04-15
Bug 40344 - [Cycle7] When user creates a project with same name then XS create two Projects

DavidKarlas:
Btw, this is not regression compared to C6(XS 5.10) I can reproduce same thing there.
https://bugzilla.xamarin.com/show_bug.cgi?id=40344

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -874,7 +874,7 @@ namespace MonoDevelop.Projects
 		protected override void OnDispose ()
 		{
 			foreach (ProjectConfiguration c in Configurations)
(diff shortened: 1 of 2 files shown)


DavidKarlas:
Apperently I wrongly read reproduction instructions to reproduce bug... I'm going to fix original bug and add commits...

DavidKarlas:
So problem is that https://github.com/mono/monodevelop/blob/cycle7/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/SlnFileFormat.cs#L114 is reading .sln file and making modifications to it instead of overriding...

@slluis does asking + deleting .sln file in project creation menu sounds OK? Or should something be chancged in SlnFileFormat.cs file?

slluis:
I feel something is not right in SlnFileFormat. It shouldn't read the existing sln when creating a new one.

PR #1391, opened 2016-04-15
[Ide] Fix NRE when determining if the LogView should auto scroll.

sgmunn:


slluis:
PR #1384 intends to fix the same issue.

slluis:
BTW #1384 should be fixing the root of the problem. I prefer that solution if it really works.

sgmunn:
Good to know, thanks

Therzok:
The changes done that caused this issue to surface are not part of Cycle7. Closing this PR as the other one has been merged.

PR #1397, opened 2016-04-19
[Refactoring] Fixed extract method bug.

mkrueger:
Declaring references need to be checked for the filename not just for
the position -> partial classes could span accross multiple files.


slluis:
Should we have unit tests for this?

PR #1398, opened 2016-04-19
Fixed 'Bug 40504 - Once in a while the text editor stops working'

mkrueger:
It's not a real fix - but without the cause it's impossible to tell
why that happened. However the editor should now recover and no longer
stop working.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
@@ -789,8 +789,16 @@ namespace MonoDevelop.Ide.Gui
 
 		Task EnsureAnalysisDocumentIsOpen ()
 		{
-			if (analysisDocument != null)
-				return SpecializedTasks.EmptyTask;
+			if (analysisDocument != null) {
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
```diff
@@ -789,8 +789,16 @@ void CancelEnsureAnalysisDocumentIsOpen ()
...
 		Task EnsureAnalysisDocumentIsOpen ()
 		{
-			if (analysisDocument != null)
-				return SpecializedTasks.EmptyTask;
+			if (analysisDocument != null) {
+				Microsoft.CodeAnalysis.Document doc;
+				try {
+					 doc = RoslynWorkspace.CurrentSolution.GetDocument (analysisDocument);
+				} catch (Exception) {
+					doc = null;
```
slluis: The exception should be logged

mkrueger: It's logged in the MonoDevelopSourceTextContainer - no need to double log it.

PR #1407, opened 2016-04-19
[Ide] Signal that the last task was removed to progressmonitor users.

Therzok:
This adds a LastTaskRemoved entrypoint where subscribers can decide
whether to end animations relating to the current progress of a monitor.

The current ProgressMonitor behaviour is to replace the rootTask if none
exists.

In the case of the Rebuild command, a Clean Task would be attached, then
finished, afterwards a Build Task would be attached, then finished.

The Clean Task would end up with a Progress > 0, so the animation would
progress in the statusbar.

The newly placed Build Task would end up with a Progress = 0, in a context
where a progress is started, but never ended. Therefore the progressbar
would animate backwards to 0 and continue forward with the Build Task's
result.

Bug 39314 - Progress bar goes backwards when building solution

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core/ProgressMonitor.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core/ProgressMonitor.cs
@@ -237,8 +237,13 @@ namespace MonoDevelop.Core
 				openStepWork = -1;
 				var t = currentTask;
 				currentTask = t.ParentTask;
-				if (currentTask == null)
+				if (currentTask == null) {
 					rootTask = null;
+					if (context != null)
+						context.Post ((o) => OnLastTaskRemoved (), null);
+					else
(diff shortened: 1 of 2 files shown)


Therzok:
Bumping this, as there is no way to know ahead of time how many steps a progress monitor will have. In the case of rebuilding, we cannot do that as it will cause the progress monitor to not show any progress, as the total progress is smaller than the sum of the child tasks unless we do insane heuristics.

Switching the main task is a valid use case in our implementation of the progress monitors, and the code itself actually lets that happen. I'd say we go with this solution.

slluis:
We do know how many steps the progress will have. A rebuild operation has two main steps: clean and build. Each of those operations can also be split in steps. We know how many projects are we going to clean and build. So I'm not sure why you say there is no way to know ahead of time how many steps a progress monitor will have. This change is not acceptable.

PR #1408, opened 2016-04-19
[Ide] FindInFilesDialog will now focus the pad in different scenarios

Therzok:
Added focusing the Result Pad on clicking the close button in the
window titlebar.

Removed focusing of the result pad when no results are returned.

cc @slluis @VincentDondain @sevoku


Therzok:
I opened a PR as I'm not sure about the behaviour on clicking the titlebar close button.

PR #1416, opened 2016-04-22
[Core] Fix task failure not reported for Runtime.RunInMainThread

mrward:
If the Runtime.RunInMainThread (Func<Task> func) is called with a func
that uses await then the Task returned from RunInMainThread would not
report any failure of the task that it ran. When an await is used the
func itself does not throw an exception and instead returns a faulted
Task. The existing RunInMainThread then set a successful result on the
task it returns even though the task it ran failed.

Simple example below that reproduces the problem:

    Task task = Runtime.RunInMainThread (() => ThrowException ());
    task.Wait ();

    async Task ThrowException ()
    {
        await Task.Delay (1);
        throw new Exception ("error");
    }

If there is no await in ThrowException then the original code in
RunInMainThread worked.


main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
```diff
@@ -407,7 +407,11 @@ public static Task RunInMainThread (Func<Task> func)
 						var t = func ();
 						t.ContinueWith (ta => {
 							try {
-								ts.SetResult (0);
+								if (ta.IsFaulted) {
+									ts.SetException (ta.Exception);
+								} else {
+									ts.SetResult (0);
+								}
```
alanmcgovern: We could get nicer exception propagation (i.e. get the real exception and not an AggregateException) if we used something like:

```
MainSynchronizationContext.Post (async () => {
  try {
        await func ();
        ts.SetResult (0);
    } catch (Exception ex) {
        ts.SetException (ex);
    }
});

```

main/tests/UnitTests/MonoDevelop.Core/RuntimeTests.cs
```diff
@@ -0,0 +1,113 @@
...
+		/// The func throws the exception so it is properly handled.
+		/// </summary>
+		[Test]
+		public async Task RunInMainThreadWithFuncTask ()
+		{
+			await Task.Run (() => {
+				var expectedException = new Exception ("Error");
+
+				Task task = Runtime.RunInMainThread (() => ThrowException (expectedException));
+				AggregateException ex = Assert.Throws<AggregateException> (() => task.Wait ());
```
alanmcgovern: If we did what i suggested in my previous comment, we *should* be able to await the `task` object here and get the actual `Exception` object nicely unwrapped here.

i.e. if expectedException was an `InvalidOperationException` then awaiting `task` would throw an InvalidOperationException :)

slluis:
@mrward see Alan's comments.

mrward:
OK I have updated both the RunInMainThread methods to use async/await and also updated the tests to verify that the underlying/real exception can be awaited.

Therzok:
@mrward conflicts.

mrward:
Conflicts fixed.

This fix should probably also go into C7 SR1. I will open another pull request for C7 if this change is merged in to master.

PR #1419, opened 2016-04-26
[WIP] Use static treeviewcolumn data funcs

Therzok:
This prevents a memory leak with the data func, so it does not keep everything it references in memory.


slluis:
Why is SetCellDataFunc causing a leak?

Therzok:
The problem is in gtk# due to refcounting issues. As @davidkarlas 's patch is way more intrusive, this is one way to fix this. Also leads to cleaner code in the cases I changed.

https://github.com/mono/gtk-sharp/pull/136

This causes crashes for stuff which is de-ref'd just to fix leaks.

PR #1424, opened 2016-04-30
[Ide] Reduce concurrency on RecentFiles modifications.

Therzok:


Therzok:
Setting as WIP as there are a few changes I want to make:
* Better docs
* Applying changes to `cachedItemList` instantly so we can return whether the list was modified or not.
* Fix the saving to _not_ happen on the UI thread.
* Slightly optimize the `localModifyList` construction to ref-change + new empty list.

alanmcgovern:
We should try not to leak exceptions too.  There are a couple of places where we don't await tasks in the new impl and those tasks may end up holding an exception, for example if `AcquireFileExclusive` the continuation in the constructor will error out and show up as unhandled in the logs.

Therzok:
@alanmcgovern Can you take a look at this version?

Therzok:
Signed off by @alanmcgovern , merging!

PR #1430, opened 2016-05-04
[C7][Ide] Run command update handlers in key bindings option panel

sevoku:
This fixes the branding icon (and maybe some others) in the key bindings option panel.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
+++ b/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
@@ -769,7 +769,6 @@
 	<Command id = "MonoDevelop.Ide.Commands.HelpCommands.About"
 			defaultHandler = "MonoDevelop.Ide.Commands.AboutHandler"
 			_label = "_About"
-			icon = "md-about"
 			_description = "Show about dialog" />
 	<Command id  = "MonoDevelop.Ide.Updater.UpdateCommands.CheckForUpdates"
 			defaultHandler = "MonoDevelop.Ide.Updater.CheckForUpdatesHandler"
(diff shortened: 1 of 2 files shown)


slluis:
I don't like the idea of calling GetCommandInfo on every command, it may be a slow operation. What is this trying to fix?

sevoku:
Well command handlers can update Labels and Icons. This is definitely used for branding (About command) and without this fix we have a MD Logo instead of the Xamarin logo there. Maybe some other commands are affected too (wrong icons).

sevoku:
An other option is to hide the command from the Key Bindings panel (which actually makes more sense in that specific case), but I'm not sure about other icons.

mhutch:
Could we do it asynchronously?

sevoku:
The problem is that `CommandService.GetCommandInfo` needs to run on the main ui thread. We could do something like:
```c#
var iters = new Dictionary<Command, TreeIter> ();
[... populate tree ...]
System.Threading.Tasks.Task.Run (() => {
	foreach (var cmd in iters) {
		CommandInfo info = null;
		Runtime.RunInMainThread (() => {
			info = IdeApp.CommandService.GetCommandInfo (cmd.Key.Id);
			keyStore.SetValue (cmd.Value, iconCol, info.Icon);
		});
	}
});
```
which would save us 100-150ms

slluis:
If the problem is the icon in the About command, we can just remove it from the command definition.

sevoku:
It looks like we could just remove the icon from the default command info in https://github.com/mono/monodevelop/blob/cycle7/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml#L791, so it would remain in the menus only.

sevoku:
@slluis sorry I missed your :+1: :)
This is already included in the `main-menu-redesign` (09e9871fdac14cae9e86a2f8c264206c7f5589c5) branch.

PR #1431, opened 2016-05-04
Fixed 'Bug 40770 - Search Result links don't display properly if code

mkrueger:
is in a collapsed region in XS 6.1'

It's caused by a race - we made the loading async this seems to be a
side effect of that.

### DIFF
--- a/main/src/core/Mono.Texteditor/Mono.TextEditor/Document/TextDocument.cs
+++ b/main/src/core/Mono.Texteditor/Mono.TextEditor/Document/TextDocument.cs
@@ -35,7 +35,6 @@ using System.ComponentModel;
 using ICSharpCode.NRefactory.Editor;
 using System.Threading.Tasks;
 using System.Threading;
-using Gtk;
 
 namespace Mono.TextEditor
 {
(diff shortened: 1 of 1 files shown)


main/src/core/Mono.Texteditor/Mono.TextEditor/Document/TextDocument.cs
```diff
@@ -1308,13 +1295,16 @@ internal void InformFoldTreeUpdated ()
...
+
 		public IEnumerable<FoldSegment> FoldedSegments {
 			get {
 				return foldedSegments;
 			}
 		}
+
 		internal void InformFoldChanged (FoldSegmentEventArgs args)
 		{
+			foldSegmentTask?.Wait ();
```
slluis: Please don't use Wait in the UI thread.

mkrueger: I fixed it - but it was a guaranteed to finished task. However not an O(1) one :/ so Wait is maybe even then not the best idea to do :(. - anyways fixed.

PR #1434, opened 2016-05-04
Search cancellation

Therzok:
cc @mkrueger @slluis 

This is pretty annoying as we can do Find references of methods like Dispose or ToString and the task takes a _LOT_ of CPU and Memory to finish.

This way, we can cancel an active search and get rid of wait times / having to close XS.

Unified cancellation reporting for searchprogressmonitor and fixed an exception that happened in Go to base command.

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Navigation/FindDerivedSymbolsHandler.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Navigation/FindDerivedSymbolsHandler.cs
@@ -87,8 +87,7 @@ namespace MonoDevelop.CSharp.Refactoring
 						}
 						foreach (var foundSymbol in result) {
 							foreach (var loc in foundSymbol.Locations) {
-								if (monitor.CancellationToken.IsCancellationRequested)
-									monitor.CancellationToken.ThrowIfCancellationRequested ();
+								monitor.CancellationToken.ThrowIfCancellationRequested ();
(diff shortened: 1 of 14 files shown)


main/src/addins/CSharpBinding/MonoDevelop.CSharp.Navigation/FindDerivedSymbolsHandler.cs
```diff
@@ -69,24 +69,30 @@ public static void FindDerivedSymbols (ISymbol symbol)
...
 						}
-					} else {
-						result = await SymbolFinder.FindOverridesAsync (symbol, TypeSystemService.Workspace.CurrentSolution).ConfigureAwait (false);
-					}
-					foreach (var foundSymbol in result) {
-						foreach (var loc in foundSymbol.Locations)
-							monitor.ReportResult (new MemberReference (foundSymbol, loc.SourceTree.FilePath, loc.SourceSpan.Start, loc.SourceSpan.Length));
+						foreach (var foundSymbol in result) {
+							foreach (var loc in foundSymbol.Locations) {
+								if (monitor.CancellationToken.IsCancellationRequested)
```
slluis: This check is redundant if you are calling ThrowIfCancellationRequested()

Therzok: :+1:

slluis:
LGTM

PR #1435, opened 2016-05-05
[Ide] Add support for IPropertyEditors which change ReadOnly state.

Therzok:
Since the base interface for IPropertyEditor does not have a ReadOnly
tag we can use, a new interface IPropertyEditorWithReadOnly has been
added to support this case.

Trigger IsReadOnly update on editor updates. This is currently done for
the widgets which had IsReadOnly support - BooleanEditorCell and
PropertyTextEditor.

Bug 40753 - On Demand Resource changes doesn't get reflected after
changing the Build action, until property window not get reopened.


Therzok:
Superseded by #1445.

PR #1438, opened 2016-05-06
Reset the tree selection func when a node is activated.

bratsche:
When we double-click a project we're getting a sequence of button
events ['press', 'release', 'press'] but we're not getting the
second release event that we're expecting. ContextMenuTreeView is
expecting this in order to reset the selection func, so instead
this works around it by resetting it in the activation event.

https://bugzilla.xamarin.com/show_bug.cgi?id=40469


slluis:
I manually merged a modified version of this fix.

PR #1439, opened 2016-05-08
[F#] Add F# projects to main sln

nosami:


knocte:
If you do this, I guess you also need to change configure.ac to require the F# compiler.

slluis:
Yes, we need a check in configure.ac.

knocte:
> Yes, we need a check in configure.ac.

But IMNSHO we would need a conditional check (i.e. it should still be possible to build MD without an F# compiler). That is, a conditional check like the one I developed in this unfinished patch: https://bugzilla.xamarin.com/show_bug.cgi?id=21878#c4

nosami:
Added configure check.

knocte:
That check is not conditional.

nosami:
@knocte - I started working on making `./configure && make` work for linux (with F# as a conditional adidin), but then found out that our new translation infrastructure requires that the projects are built as part of the main sln - so no way to make it conditional

knocte:
New translation infrastructure? You're still using gettext AFAIU, and for that you just need the proper way to loop through the .fs files in the git submodule.

nosami:
We're moving away from gettext

knocte:
If you commit this PR as is, then the separation of the F# binding as a submodule doesn't make sense anymore. It would be better to bring the sources to this repo.

Therzok:
@knocte The problem is that the Gettext addin which generates the strings to be translated through regexes, does not go into projects which aren't included in the solution.

So we decided to go the easy way of including it inside Main.sln.

As a side-idea, we could export an msbuild property that decides whether or not we build the F# binding (when we run xbuild, we append /p:NoFSharp=yes)

Another approach would be for the F# binding to have its own `.po` file and use `msgcat` on that file and the one in Main.sln, but that can get messy.

knocte:
I like the msbuild solution, but shouldn't it be opt-in instead (`/p:FSharpEnabled=yes`), so that it gets enabled only if `fsharpc` is found.

Therzok:
@knocte I think it's better to plan ahead and expect more addins in F# inside the MD source tree.

nosami:
@slluis what do you think? Should building the F# addin be optional on linux?

PR #1442, opened 2016-05-09
[Ide][Feature] Adds feature in MonoDevelop to Pin opened files (tabs)

netonjm:
Pinned files are a group of elements aligned at left... useful to use in frequently used files.. similar behaviour like VS2013-15


sevoku:
Looks great! But it has some small issues:
* if the tabs don't fit, it should scroll unpinned tabs only.
  Atm it looks like this:
  bildschirmfoto 2016-05-10 um 11 04 18
* we need a TogglePinned command so the user can assign a key binding to it and it should be visible in the tab context menu.
* the pinned/unpinned icons should be 1-2px bigger (I mean the drawing inside the images, which is 11x11@2x atm)

alanmcgovern:
If this feature is eventually going to ship but we're not sure when, can we merge it once it's passed review and add a feature toggle so it's only enabled if the file `~/enable-pin-tabs` exists.

Any other feature toggle mechanism would be fine by me too :)

Therzok:
:+1: sounds great to me Alan!

PR #1455, opened 2016-05-17
Fixed 'Bug 40915 - Assembly Reference Aliases are not support'

mkrueger:


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/AssemblyReference.cs
```diff
@@ -51,6 +53,12 @@ public override int GetHashCode ()
...
 			}
 		}
-	}
-}
 
+		public IEnumerable<string> GetAliases ()
+		{
+			if (Aliases == null)
+				return Enumerable.Empty<string> ();
+			return Aliases.Split (',', ';');
```
slluis: Can you please rename that method to GetAliasesList()? It is confusing to have an Aliases property and a GetAliases method.
Aliases can't be null. If it has no value it will be empty, in which case the split will always return and enumeration of one empty string.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/AssemblyReference.cs
```diff
@@ -51,6 +53,13 @@ public override int GetHashCode ()
...
 		}
-	}
-}
 
+		/// <summary>
+		/// Returns an enumerable collection of aliases. 
+		/// </summary>
+		public IEnumerable<string> EnumerateAliases ()
+		{
+			return Aliases.Split (',', ';').Where (a => !string.IsNullOrEmpty (a));
```
slluis: Better use the StringSplitOptions.RemoveEmptyEntries flag.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -555,6 +555,8 @@ IEnumerable<DocumentInfo> GenerateProjections (MonoDevelop.Projects.ProjectFile
 					var metadataReference = MetadataReferenceCache.LoadReference (projectId, fileName);
 					if (metadataReference == null)
 						continue;
+					if (file.Aliases != null)
```
slluis: Aliases can't be null. It will be empty if there are no aliases.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/AssemblyReference.cs
slluis: @mkrueger Can you fix ^ ?

PR #1458, opened 2016-05-17
[C7][Toolbox] Populate the index on a background thread to allow monitor …

Therzok:
…updating.

Also do not show the dialog if the user cancelled the operation.

Bug 41113 - On clicking '+' button of ToolBox window, XS get hanged for few seconds.


slluis:
I'm not sure this can be done in a background thread. Getting the component list may involve loading widget toolkit stuff, which needs to be loaded in the UI thread. It needs to be verified.

PR #1460, opened 2016-05-20
[Ide] Reduce the amount of tasks created during loading solutions.

mkrueger:


slluis:
We need to test this in master before merging.

PR #1462, opened 2016-05-22
[MacPlatform] Searchlight

iainx:
A Spotlight like overlay search system for the Mac platform to replace the search bar in the toolbar.


iainx:
A video of it in action: http://screencast.com/t/kQXd2E0S4c9a

iainx:
Fully Cocoa native with keyboard controls.

Missing things:
- Doesn't have the popup menu that the toolbar search entry has.
- Doesn't have the Top Item category implemented yet.
- Doesn't display the tooltip when selecting an item.

I thought with the extra space we could design a nicer UI for these features, maybe like the Spotlight UI? @vancura any ideas?

vancura:
We need some design polish. I'll take a look at it.

mhutch:
I like it!

My one concern is that it and its keybinding would become much less discoverable. Maybe we could have a magnifying glass button in the toolbar, and display the shortcut in the placeholder text?

vancura:
@mhutch: Exactly that's the plan :)

iainx:
@mhutch should I file a spec for this feature on the trello spec board, so we don't forget/lose it?

slluis:
@iainx A revamp of the search and navigation system is already planned.

slluis:
Global search improvements are in progress.

sssw2:
😃

PR #1465, opened 2016-05-23
Needed for UI Automation: Added Widget.Name and SemanticModelAttribute to TreeStore/ListStore

manish:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/DefaultMonitorPad.cs
```diff
@@ -66,7 +66,7 @@ public DefaultMonitorPad (string typeTag, string icon, int instanceNum)
 			
 			this.icon = icon;
 
-			logView = new LogView ();
+			logView = new LogView { Name = "applicationOutput" };
```
kdubau: This will apply to all MonitorPads, thats not what you want. You can probably set the name to the `typeTag` local var.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/DefaultMonitorPad.cs
```diff
@@ -66,7 +66,7 @@ public DefaultMonitorPad (string typeTag, string icon, int instanceNum)
 			
 			this.icon = icon;
 
-			logView = new LogView ();
+			logView = new LogView { Name = typeTag };
```
Therzok: `typeTag` isn't enough. You need to make use of `instanceNum` too.

Therzok:
Merge at will after `instanceNum` is being used to be able to query multiple pads based on pinning status.

PR #1466, opened 2016-05-23
Fixed 'Bug 41245 - Attribute code completion not showing all

mkrueger:
constructors and showing too many things'


slluis:
@mkrueger hey, can you fix the conflicts?

slluis:
@mkrueger hmm, why remove unit tests?

mkrueger:
Because of the diff - it wasn't really readable in the editor - the unit test is in master.

slluis:
I don't understand what you mean with the diff not being readable.

mkrueger:
Hm, not sure that this was the case here - doesn't look like this is complex.
I had one case where the merge diff wasn't readable/junk and there it was easier to remove just the case alltogether.

The case is in master and it's not removed there - so not 100% sure about this commit here - it might be the unreadable case. But anyways the completion is case is checked in master and the engine is distinct enough so that it doesn't matter.

PR #1470, opened 2016-05-27
[Ide][Mac] Use system font.

bratsche:


slluis:
What will happen if the installed GTK doesn't have your patches?

sevoku:
It will run with the default Sans font then (instead of Helvetica Neue or the new SF fonts)

bratsche:
Yes, that's right. It will fall back to Sans in that case.

slluis:
Ok, let's wait for the Mono patch to be merged, and bump the Mono dependency.

bratsche:
bockbuild branch has been merged

PR #1472, opened 2016-05-27
Add documentation extension point

VincentDondain:
`IErrorDocumentationProvider` has one method: `GetDocumentationLink` which gives the address of the documentation for a given error code within the error's description.

Related to https://github.com/xamarin/md-addins/pull/1080.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Extensions/IErrorDocumentationProvider.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Extensions/IErrorDocumentationProvider.cs
@@ -1,0 +1,34 @@
+﻿//
+// IErrorDocumentationProvider.cs
+//
+// Author:
+//       Vincent Dondain <vincent.dondain@xamarin.com>
+//
+// Copyright (c) 2016 Xamarin
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+namespace MonoDevelop.Ide.Extensions
+{
+	public interface IErrorDocumentationProvider
+	{
+		string GetDocumentationLink (string errorDescription);
+	}
+}
+
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/ErrorListPad.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/ErrorListPad.cs
@@ -539,21 +539,24 @@ namespace MonoDevelop.Ide.Gui.Pads
 		void OnShowReference (object o, EventArgs args)
 		{
 			string reference = null;
-			if (GetSelectedErrorReference (out reference)) {
-				Process.Start ("http://google.com/search?q=" + System.Web.HttpUtility.UrlEncode (reference));
-				return;
-			}
+			if (GetSelectedErrorReference (out reference) && reference != null)
+				DesktopService.ShowUrl (reference);
 		}
 
 		bool GetSelectedErrorReference (out string reference)
 		{
+			string webRequest = "http://google.com/search?q=";
 			TaskListEntry task = SelectedTask;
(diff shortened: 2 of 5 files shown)


VincentDondain:
@Therzok @mrward feel free to comment it's my first extension point (:

VincentDondain:
@mrward According to what we discussed on slack I updated the `ErrorListPad` to also redirect to the documentation like my extension point is providing.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Tasks/TaskListEntry.cs
```diff
@@ -235,15 +235,37 @@ public TaskListEntry (BuildError error, object owner)
...
 		{
 			if (!file.IsNullOrEmpty) {
 				if (System.IO.File.Exists (file)) {
 					var project = WorkspaceObject as Project;
 					IdeApp.Workbench.OpenDocument (file, project, Math.Max (1, line), Math.Max (1, column));
 				} else {
-					var pad = IdeApp.Workbench.GetPad<ErrorListPad> ()?.Content as ErrorListPad;
-					pad?.FocusOutputView ();
+					if (HasDocumentationLink ()) {
+						System.Diagnostics.Process.Start (DocumentationLink);
```
Therzok: `MonoDevelop.Ide.DesktopService.OpenUrl` here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/ErrorListPad.cs
```diff
@@ -539,21 +539,24 @@ void OnTaskCopied (object o, EventArgs args)
 		void OnShowReference (object o, EventArgs args)
 		{
 			string reference = null;
-			if (GetSelectedErrorReference (out reference)) {
-				Process.Start ("http://google.com/search?q=" + System.Web.HttpUtility.UrlEncode (reference));
-				return;
-			}
+			if (GetSelectedErrorReference (out reference) && reference != null)
+				Process.Start (reference);
```
Therzok: Possibly change this to `MonoDevelop.Ide.DesktopService.OpenUrl`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Tasks/TaskListEntry.cs
```diff
@@ -235,15 +235,37 @@ public TaskListEntry (BuildError error, object owner)
...
+		}
+
+		public bool HasDocumentationLink ()
+		{
+			bool documentationProvided = false;
+			var extensions = Mono.Addins.AddinManager.GetExtensionObjects<Extensions.IErrorDocumentationProvider> ("/MonoDevelop/Ide/IErrorDocumentationProvider", false);
+			for (var i = 0; !documentationProvided && i < extensions.Length; i++) {
+				var link = extensions [i].GetDocumentationLink (description);
+				if (!string.IsNullOrEmpty (link)) {
+					DocumentationLink = link;
```
Therzok: We need to settle on a standard, choose the first or the last. :) I'm not sure what the current one is, most of the code I've seen always chose the first extension which could handle a situation.

VincentDondain: Will be the first in this case, as soon as we get a non-null link we exit the loop (`documentationProvided = true`).

Therzok: I missed the condition in the for loop. The code is pretty verbose in this way, maybe we should just get rid of `documentationProvided` and do `return true` and `return false` to exit the loops? Code would look clearer this way.

Anyway, this is bikeshedding and preference only.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/ErrorListPad.cs
VincentDondain: There isn't such thing as `MonoDevelop.Ide.DesktopService.OpenUrl`. Are you talking about Xwt `Desktop.OpenUrl`? `MacPlatform.OpenURL`?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Tasks/TaskListEntry.cs
VincentDondain: I'm not a fan of multiple returns in general but I guess in this case it's okay indeed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/ErrorListPad.cs
Therzok: I meant `DesktopService.ShowUrl`, sorry.

VincentDondain: Ah yup, got it :P

VincentDondain:
Comments addressed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Extensions/IErrorDocumentationProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+namespace MonoDevelop.Ide.Extensions
+{
+	public interface IErrorDocumentationProvider
```
slluis: Please use an abstract class instead of an interface.

slluis:
The extension point should should allow specifying a regular expression in the extension node, which would be used to match the error code. It could also allow specifying an url with a placeholder for the error code. The goal is to be able to generate an error link from the extension data, and avoid loading the add-in assembly if not necessary.

PR #1475, opened 2016-05-30
Improve the speed of Find in files

iainx:
A couple of patches to improve the Find In Files speed. tl;dr these patches cut about 10seconds from the time taken to search through files in Main.sln.

Some speed results from runs of 10 goes on Main.sln (18,000 files)

Average time to search every file for the word "Updater"
Master branch: 21.85 seconds (Range 20.36 -> 23.3)
With this patch applied: 11.63 seconds (Range 9.54 -> 14.34)

We can look closer at this because there are two phases to the search, generating the list of files and readers for them, and doing the search on those files.

Master (average over 10 runs):
 - List files: 18.98 seconds (range: 17.5 -> 20.43)
 - Search files: 2.85 seconds

With patch applied:
 - List files: 8.81 seconds (range: 6.75 -> 11.46)
 - Search files: 2.85 seconds

Which makes sense as we haven't touch the searching code.
With the patch attached, we can split the List files section into another two phases: getting a list of the files to search, and creating readers for those files, and the times for that are

 - creating list of files: 3.23 seconds (range 3.04 -> 3.39)
 - creating readers for the files: 5.5 seconds (range 3.57 -> 8.07)

it's clear that the creating readers section is pretty time consuming, ~~and I'm wondering if the wild fluctuation in the range is maybe because of the context switch that occurs.~~ [edit: no, checked, the context switch only takes .2seconds] 

Creating the readers for each file involves loading the whole file into memory, then checking the BOM on it, before creating a StreamReader. I tried porting this to use a memory mapped file, but it was slower for some reason. This is certainly an area for further investigation

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindReplace.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindReplace.cs
@@ -112,17 +112,48 @@ namespace MonoDevelop.Ide.FindInFiles
 
 
 				var contents = new List<FileSearchResult>();
+				var filenames = new List<string> ();
 				foreach (var provider in scope.GetFiles (monitor, filter)) {
 					if (token.IsCancellationRequested)
 						return Enumerable.Empty<SearchResult> ();
 					try {
 						searchedFilesCount++;
-						contents.Add(new FileSearchResult (provider, provider.ReadString (), new List<SearchResult> ()));
+						contents.Add(new FileSearchResult (provider, /*provider.ReadString ()*/null, new List<SearchResult> ()));
+
+						filenames.Add (Path.GetFullPath (provider.FileName));
+
 						if (searchedFilesCount % step == 0)
 							monitor.Step (2); 
 					} catch (FileNotFoundException) {
 						MessageService.ShowError (string.Format (GettextCatalog.GetString ("File {0} not found.")), provider.FileName);
 					}
+				}
+
+				TextReader [] readers = null;
+				var task = GetSearchDocumentsReaders (filenames);
+				if (task.Wait (1000))
+					readers = task.Result;
+
+				int idx = 0;
+				if (readers != null) {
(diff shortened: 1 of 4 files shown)


Therzok:
:+1: looks good

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindReplace.cs
```diff
@@ -181,6 +209,11 @@ public IEnumerable<SearchResult> FindAll (Scope scope, ProgressMonitor monitor,
 			}
 		}
 
+		async Task<TextReader[]> GetSearchDocumentsReaders (List<string> filenames)
```
slluis: Unused method?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
```diff
@@ -179,6 +180,31 @@ public Document GetDocument (string name)
...
+			TextReader [] results = new TextReader [filenames.Count];
+
+			int idx = 0;
+			lock (documentsByFilePath) {
+				foreach (var f in filenames) {
+					var fullPath = (FilePath)FileService.GetFullPath (f);
+					//var r = documentsByFilePath [fullPath];
+
+					Document r;
+					bool present = documentsByFilePath.TryGetValue (fullPath, out r);
```
slluis: I don't think you need to keep track of documents on a dictionary for this. There aren't that many open files usually.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindReplace.cs
Therzok: Yep. The call got removed in 7af90e970dfaf83ed33245e9e713679e8a87f844.

iainx:
@slluis Ok, I removed the dictionary and the unused method

PR #1478, opened 2016-05-31
[Ide] Implement cancellation via status bar.

Therzok:
This adds a basis for searchbar cancellation via the status bar. The
CancellationTokenSource is passed to the statusbar implementation to be
used to cancel the current operation.

The implementation is only done for Mac.

Bug 41335 - StatusBar cancellation

cc @slluis on API changes.
cc @iainx on Mac changes.
cc @vancura on UI changes.


Therzok:
There currently is a problem where we have the following situation:
* A monitor does not set its own cancellationtokensource.
* One is created by default for it [here](https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Core/ProgressMonitor.cs#L542-L543) - one example is loading a solution
* The cancel button appears but does nothing since the operation doesn't handle cancellationtokensource cancellation.

Maybe we should default it to not construct a source by default and add nullchecks on where we work on it?

iainx:
Mac bits look good to me 👍

vancura:
Design bit looks good to me 👍

slluis:
The button should have some kind of visual feedback when hovering.

slluis:
@Therzok I don't understand the issue you are explaining. Can you elaborate?

Therzok:
@slluis We decided it should be opt-in, not on by default. The issue is that not all operations have a cancellation handler.

Therzok:
TODO: hover effect.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/ProgressMonitors.cs
```diff
@@ -124,9 +124,9 @@ protected override OperationConsole OnCreateConsole (CreateConsoleOptions option
 		/******************************/
 		
 		
-		public ProgressMonitor GetStatusProgressMonitor (string title, IconId icon, bool showErrorDialogs, bool showTaskTitle = true, bool lockGui = false, Pad statusSourcePad = null)
+		public ProgressMonitor GetStatusProgressMonitor (string title, IconId icon, bool showErrorDialogs, bool showTaskTitle = true, bool lockGui = false, Pad statusSourcePad = null, bool showCancelButton = false)
```
slluis: You can't just add an optional parameter. This is an API break.

Therzok: Will fix.

slluis: For now you can just add an overload that requires all parameters.

Therzok: That was my plan, thanks.

Therzok:
Updated.

PR #1479, opened 2016-05-31
[Ide] Fixup rebuild progress monitor behaviour

Therzok:
We need to have one statusbar progress per action we're doing, so we
will have to attach the same error pad to a different status bar for
build and clean.

The idea of the code is to have one statusbar monitor alive at a time,
while the error one is alive for the whole operation, and the
aggregated monitors are disposed at the end just so they're not
marked as finalized but not Disposed by profiler tools.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
@@ -1180,7 +1180,11 @@ namespace MonoDevelop.Ide
 			try {
 				OnStartClean (monitor, tt);
 
+				monitor.BeginTask (GettextCatalog.GetString ("Rebuilding..."), 2);
+				monitor.BeginStep (GettextCatalog.GetString ("Rebuilding... (Clean)"));
+
 				var res = await CleanAsync (entry, monitor, tt, true, operationContext);
+				monitor.EndStep ();
 				if (res.HasErrors) {
 					tt.End ();
 					monitor.Dispose ();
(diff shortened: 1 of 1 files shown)


Therzok:
@slluis maybe we should make it illegal to switch the root task of a monitor that has a statusprogressmonitor so we don't have this bug in the future.

slluis:
Why do we need to have one statusbar progress per action? We can use a single status bar progress for clean + build.

Therzok:
@slluis If we keep the same statusbar, we progress will go backwards, since we don't reset the ongoing progress anywhere. This is what we currently have.

The rebuild action is a special case, where we try to switch the main task of the progress monitor, therefore 'resetting' the progress. The implementations of the statusbar cannot decide when this happens - when the task's progress has ended so a new one will start from 0 progress onwards - there are no BeginProgress/EndProgress calls from the beginning and the end of the StatusProgressMonitor's lifetime. That's what decides the animation to be used by SetProgressFraction.

And due to threading issues with CALayers, we can't make that synchronization happen in the UI code - instantly setting a bounds might have it reset by the animation that is in progress, even though we remove it.

We also can't know how many projects we will build from where the statusprogressmonitor is created - without breaking IBuildTarget API - aka one big progress for both Clean and Build, Clean would stop half-way.

I also tried making one big task that aggregated the Clean/Build tasks with 2 steps, but it wouldn't do any animation since it didn't consider each step split into two.

slluis:
What I mean is:
```
monitor.BeginTask(2);
monitor.BeginStep(1);
project.Clean(monitor)
monitor.EndStep();
monitor.BeginStep(1);
project.Build(monitor)
monitor.EndTask();
```

Therzok:
@slluis Fixed

PR #1481, opened 2016-05-31
Make ErrorDocumentationProvider a custom extension point

VincentDondain:
It's now easier for add-ins to implement link to the documentation for specific error codes thanks to the custom extension point logic.

Here's an exemple of the only thing add-ins need to do:

```
<Extension path = "/MonoDevelop/Ide/ErrorDocumentationProvider">
        <ErrorDocumentationProvider id = "MTErrorDocumentation"
                                    regex = "MT\d{4}"
                                    url = "https://developer.xamarin.com/guides/ios/troubleshooting/mtouch-errors/#{0}"/>
</Extension>
```


VincentDondain:
@slluis that follows your custom extension point advice in https://github.com/mono/monodevelop/pull/1472#issuecomment-222564814.
Let me know if I'm missing anything.

slluis:
It looks good!

PR #1484, opened 2016-06-01
[Ide][Mac] Fix unfocused selected tree row background

sevoku:
This fix depends on mono/xamarin-gtk-theme@c202097fba9982925fff4d5e67d0ca0feadd8ecb and requires an updated mono build.

The problem with this is, that without the updated mono build the trees will always have the inactive selection color on Mac:
pasted image at 2016_05_31 14_59

~~**Please don't merge now**, I'll update the status as soon as the new mono builds include the patch.~~
(fixes bug #40877)

### DIFF
--- a/main/src/core/MonoDevelop.Ide/gtkrc.mac
+++ b/main/src/core/MonoDevelop.Ide/gtkrc.mac
@@ -241,9 +241,7 @@ style "tooltip"
 
 style "treeview" = "default"
 {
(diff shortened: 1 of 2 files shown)


sevoku:
@slluis sorry I forgot about this PR. Its ok to merge now. I've tested it with mono 4.6.1 so it would work for `cycle8` too, should I create a c8 PR, or can we skip it?

PR #1487, opened 2016-06-02
Localization set

Therzok:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/LocalizationService.cs
```diff
@@ -0,0 +1,106 @@
...
+using MonoDevelop.Core;
+
+namespace MonoDevelop.Ide
+{
+	public static class LocalizationService
+	{
+		const string path = "/MonoDevelop/Ide/LocaleSet";
+		static readonly List<LocaleSet[]> locales = new List<LocaleSet[]> ();
+
+		public static void Initialize ()
```
slluis: This should be internal.

Therzok: Ack.

PR #1491, opened 2016-06-06
Allow the user to restart MD from the preferences dialog

sevoku:
Adds a "Restart" button to the style preferences panel, which allows the user to restart MD after a language or theme change. The restart logic is im implemented in `DesktopService.RestartIde ()` and can be overriden by the PlatformService.
The default implementation starts MonoDevelop in a new process after the Ide has been closed successfully.

**MAC**: in order to keep the App lifecycle intact (without messing up the icons in the dock) MD needs to be terminated. I've added a new `--start-app-bundle`argument to *monostub*. If its passed to i.e. *mdtool* (wrapped), *monostub* will wait for MD to terminate and simply launch the App bundle again.
But I'm not sure whether `monostub` is the right place for this. Maybe a separate binary (i.e. *mdlauncher*) - which could cover more cases (start/restart/restart-wait/restart-terminate) - would make more sense?


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/DesktopService.cs
```diff
@@ -370,5 +370,15 @@ internal static void PlaceWindow (Gtk.Window window, int x, int y, int width, in
 		{
 			PlatformService.PlaceWindow (window, x, y, width, height);
 		}
+
+		/// <summary>
+		/// Restarts MonoDevelop
+		/// </summary>
+		/// <returns> false if the user cancels exiting. </returns>
+		/// <param name="reopenWorkspace"> true to reopen current workspace. </param>
+		public static bool RestartIde (bool reopenWorkspace)
```
slluis: This should be internal, since the public method is in the Ide class.

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -893,5 +894,37 @@ public override void OpenFolder (FilePath folderPath, FilePath[] selectFiles)
...
+		public override bool RestartIde (bool reopenWorkspace)
+		{
+			FilePath bundlePath = NSBundle.MainBundle.BundlePath;
+
+			if (bundlePath.Extension != ".app")
+				return base.RestartIde (reopenWorkspace);
+
+			var reopen = reopenWorkspace && IdeApp.Workspace != null && IdeApp.Workspace.Items.Count > 0;
+
+			if (IdeApp.Exit ()) {
```
slluis: Please move the IdeApp.Exit() call to IdeApp.Restart().

sevoku:
@slluis done :), and I've added some error logging, cause the user won't see anything if a restart failes (Main loop already ended).

Any thoughts about `monostub` vs separate binary?

slluis:
I think using monostub is the easiest solution for now. Feel free to commit once the conflicts have been resolved.

PR #1492, opened 2016-06-06
When ActiveRuntime is changed, fire RuntimeChanged event

manish:
This is a corresponding PR to https://github.com/mono/monodevelop/pull/1258 

This PR would help us select the device via test frameworks so that we can automate deploy to device


main/src/addins/MacPlatform/MainToolbar/SelectorView.cs
```diff
@@ -482,6 +482,8 @@ void OnSizeChanged ()
 					activeRuntime = value;
 					using (var mutableModel = value.GetMutableModel ()) {
 						state |= CellState.RuntimeShown;
+						if (RuntimeChanged != null)
+							RuntimeChanged (this, new HandledEventArgs ());
```
alanmcgovern: It might make sense to fire this event after all the state has been updated. Maybe move it after the `UpdatePathText` invocation

PR #1494, opened 2016-06-09
[Addins] Add support for checking whether addins are enabled

Therzok:


main/src/core/MonoDevelop.Core/MonoDevelop.Core.AddIns/AddinEnabledCondition.cs
```diff
@@ -0,0 +1,52 @@
...
+using System;
+using Mono.Addins;
+
+namespace MonoDevelop.Core.AddIns
+{
+	class AddinEnabledCondition : ConditionType
+	{
+		public AddinEnabledCondition ()
+		{
+			AddinManager.AddinLoaded += (o, e) => NotifyChanged ();
```
Therzok: I'm pretty sure we're not leaking this, since the ConditionType will be alive anyway.

slluis:
The correct solution for this would be to use an optional add-in module: https://github.com/mono/mono-addins/wiki/Optional-Modules.

However, if module dependencies are specified in the xml manifest, we'll have to hardcode the add-in version, which we can't do. Maybe I can find a solution for this in Mono.Addins.

Therzok:
I had no idea that existed, it indeed does look like the better solution.

Therzok:
Implemented version-less check in https://github.com/mono/mono-addins/pull/38

Therzok:
Fixed via https://github.com/mono/mono-addins/commit/1919a2edc2cd4d85120062fec3b4b5f55cf6773e

PR #1497, opened 2016-06-10
Bug 41604 - File not found exception logged every second into IDE log when not project not compiled

DavidKarlas:
Reason this appeared only now and only to Matt is because he installed my adding which caused this here: https://github.com/DavidKarlas/VSCodeDebugger/blob/master/VSCodeDebugger/VSCodeDebuggerEngine.cs#L23

Since I will fix this in AddIn this weekend, maybe we can reject this commit for C7SR1(I already committed this check to master...)
I opened this PR against C7SR1 because bug is specified as such...


mrward:
I did not know it was due to the VSCodeDebugger addin. I assigned the bug to the C7SR1 milestone since I thought it was affecting everyone but if it is not then there is no real need for it to be included there.

PR #1498, opened 2016-06-13
Shared asset files need to be added first for F#

nosami:
If the shared asset files aren't included first, then the consuming project won't compile. 

I'm working on the assumption that it won't make any difference for C# projects, so can treat them the same.
/cc @slluis


slluis:
Why does it matter? projects are built by msbuild, it doesn't matter how files are ordered in the project model.

nosami:
It matters because we read `project.Files` to pass arguments to the F# compiler for IDE usage - https://github.com/mono/monodevelop/blob/master/main/external/fsharpbinding/MonoDevelop.FSharpBinding/Services/CompilerArguments.fs#L295-L297

slluis:
don't you also need to handle OnFileAddedToProject?

nosami:
Yeah, looks like it. Added to OnFileRenamedInProject too.

nosami:
Thinking about this some more, I don't think this is going to work for OnFileAddedToProject - I would need to make sure that the file in the shared project is added to the correct position in the consuming project. I think the best course of action is to do this from the F# binding

PR #1506, opened 2016-06-14
Fix App Config is not read in Unit Tests

tstivers1990:
Set ShadowCopyFiles to false in order to fix a bug that prevents app.config settings from being loaded when running tests.

https://bugzilla.xamarin.com/show_bug.cgi?id=41541


slluis:
approve

tstivers1990:
This didn't actually fix the bug. Closing this and creating a new pull request with the correct bug fix.

tstivers1990:
See #1508 for the correct fix.

PR #1508, opened 2016-06-14
[NUnit] Fix 41541 - App Config is not read in Unit Tests

tstivers1990:
The BasePath property on the created TestPackage was being set to the directory that contains the NUnitRunner.exe binary. It is supposed to be set to the directory that contains the test assembly.


LiohAu:
Is there a way to retrieve the new add-in version without waiting for cycle7-sr1 ?

PR #1509, opened 2016-06-15
[Ide]Stop tooltip from appearing after completion list has closed 

Sl0vi:
I believe I've found the cause of the bug I tried to fix in pull request #1353 

Commit 176ab0480dd24fa2e3623ea5b85c228dd4036ab1 added a hack that uses a timeout to change the opacity of the tooltip window.

This timeout adds a small timing window where the opacity on the tooltip window could get updated after the tooltip window has been hidden.

After adding a check for if the tooltip window is set to visible before changing the opacity inside the timeout handler, I am no longer able to reproduce the bug.

The extra checks to the completion list window in my last pull request do not appear to be necessary after this fix and can be reverted.


slluis:
whitelist

slluis:
It looks good to me. Thanks!

PR #1511, opened 2016-06-15
Bug 41572 - InitializeComponent() The Call is Ambiguous between

DavidKarlas:
Since this is pretty visible(few users already reported), I suggest we include this fix in SR0


DavidKarlas:
For this commit to work correctly, https://github.com/xamarin/md-addins/pull/1102 needs to be merged

DavidKarlas:
Pushed manually, since md-addins bumped changed branch...

PR #1515, opened 2016-06-16
[Fix] 41750 - TestCase detection broken

tstivers1990:


slluis:
whitelist

slluis:
Thanks!

tstivers1990:
No problem. I'm not 100% certain if the fix actually works or not. I'll check it in a few hours when I get my new computer which will actually be capable of compiling MonoDevelop. I'll respond here to let you know whether or not the fix was successful.

I'm pretty certain this change fixes it though.

tstivers1990:
Managed to test this fix today, and confirmed it did actually fix the bug.

PR #1517, opened 2016-06-16
Bug 41738 - The 'TextTemplatingFileGenerator' code generator crashed

DavidKarlas:
https://bugzilla.xamarin.com/show_bug.cgi?id=41738

Changes by files:
ProjectFileTemplatingHost.cs:
Replacing properties like $(SolutionDir) with correct value when resolving assembly path(https://msdn.microsoft.com/en-us/library/bb126478.aspx#Anchor_3)
MonoDevelopTemplatingHost.cs:
Apperently it's legal to say "System.dll" in T4, which GAC lookup doesn't like
TemplatingEngine.cs:
/noconfig is needed so mcs.exe(or csc.exe) doesn't report error CS1703(An assembly `System' with the same identity has already been imported. Consider removing one of the references) for System.dll
TemplateGenerator.cs:
By adding reference by FullName we let framework resolve full path, which allows deduplication of System.dll reference in case when user also specifies <#@ assembly name="System.dll" #>

@mhutch can you review this?

### DIFF
--- a/main/src/addins/TextTemplating/Mono.TextTemplating.Tests/DummyTemplateGenerator.cs
+++ b/main/src/addins/TextTemplating/Mono.TextTemplating.Tests/DummyTemplateGenerator.cs
@@ -1,0 +1,44 @@
+﻿//
+// DummyTemplateGenerator.cs
+//
+// Author:
+//       David Karlaš <david.karlas@xamarin.com>
+//
+// Copyright (c) 2016 Xamarin, Inc (http://www.xamarin.com)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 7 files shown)


mrward:
Is it possible to write a unit test for this bug?

DavidKarlas:
Added some basic unit test references resolving

mrward:
@DavidKarlas This change looks OK to me. :shipit:

mrward:
This has broken the T4 unit tests on Windows.

The test is failing since on Windows when the assemblyReference is a full path to an assembly file trying to create an AssemblyName will fail with `The given assembly name or codebase was invalid.`. This can be fixed by modifying the DummyTemplateGenerator:

                protected override string ResolveAssemblyReference (string assemblyReference)
                {
    +                       if (System.IO.Path.IsPathRooted (assemblyReference))
    +                               return assemblyReference;
    +
                        var assemblyName = new AssemblyName (assemblyReference).Name;
                        if (assemblyName == typeof (Uri).Assembly.GetName ().Name)

However the template generation then fails since there are two compiler errors on Windows:

  - warning CS2023: Ignoring /noconfig option because it was specified in a response file
  - error CS1703: An assembly with the same identity 'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' has already been imported. Try removing one of the duplicate references.

mrward:
OK I have modified the DummyTemplateGenerator some more to avoid the duplicate System assembly and now the tests pass on Windows.

PR #1522, opened 2016-06-21
[StatusBar] Implement cancellation tooltip.

Therzok:
Bug 41720 - Add tooltip to searchbar cancellation


slluis:
I feel it is not necessary to have a custom cancel tooltip for each operation. A generic "Cancel operation" might be enough.

/cc @vancura

vancura:
Exactly. Generic tooltip is enough.

Therzok:
Updated.

PR #1524, opened 2016-06-23
WIP: Clean up MonoDevelop.UnitTesting

tstivers1990:
This is a work in progress, please do not merge it yet. Tag it with do not merge if you can.

I'm looking for comments here. Mostly whether or not what I'm doing is against a guideline, or I'm otherwise wasting my time. I've read the MonoDevelop and Mono guidelines, they don't mention the things I'm doing.

I also intend to clean up some of the classes. Specifically, I notice a few classes could be made abstract as they are never instantiated directly, but always derived from, and they have a few blank methods that could be made abstract rather than cluttering the class up with blank methods. I feel this will make the code a lot more clear, and less confusing.


tstivers1990:
blacklist

main/src/addins/MonoDevelop.UnitTesting/Services/TestSessionEventArgs.cs
```diff
@@ -26,18 +26,14 @@
...
 
-
 using System;
-using System.Collections;
+using MonoDevelop.Core;
 
 namespace MonoDevelop.UnitTesting
 {
-	public class GeneralTestOptions: ICloneable
+	public class TestSessionEventArgs: EventArgs
```
slluis: What happened here?

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
```diff
@@ -36,7 +36,7 @@
 
 namespace MonoDevelop.UnitTesting
 {
-	public class UnitTestGroup: UnitTest
+	public abstract class UnitTestGroup: UnitTest
```
slluis: Don't make this class abstract. That's an API break.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
```diff
@@ -132,50 +129,37 @@ protected virtual void OnCreateTests ()
...
+		protected override bool OnCanRun (Core.Execution.IExecutionHandler executionContext)
 		{
 			foreach (UnitTest t in Tests)
 				if (!t.CanRun (executionContext))
 					return false;
 			return true;
 		}
-
-		
-		protected virtual void OnBeginTest (TestContext testContext)
```
slluis: Don't remove public api.

main/src/addins/MonoDevelop.UnitTesting/Services/TestSessionEventArgs.cs
tstivers1990: Not sure. Git nonsense I guess.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
tstivers1990: How does it break the API?

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
tstivers1990: Nothing is using it.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
slluis: Making a class abstract is an API break. If there is code that creates an instance of that class, it will break.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
slluis: Nothing is using it in the monodevelop repository, but this is public API, third party test runners may use it.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
tstivers1990: I know of no code that directly creates an instance of this class. The only pieces of code that I'm aware of that use this class are MonoDevelop.UnitTesting.NUnit, which is part of MonoDevelop, and wasn't broken by the change. And the xUnit runner, which is currently broken and completely un-functional, and needs to be rewritten. So unless there's something else I'm not aware of that's directly creating instances of this class, I think the benefits of changing this class to abstract for clarity outweigh the potential drawbacks.

This also prevents people from directly creating instances of this class in the future, which I'm fairly certain wasn't something that was intended to be done. From what I can understand, based on reading the code and reading code that uses this namespace, it was intended for this class to always be inherited from, and its members overridden. There are language constructs that allow us to enforce this behavior. So why not use them?

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
tstivers1990: I am only aware of two test runners. NUnit and xUnit. NUnit is part of MonoDevelop, so breaking it isn't as big an issue. xUnit is currently broken and completely un-functional, and needs to be rewritten heavily anyway. So breaking it isn't a big deal either. Perhaps there are other test runners as well, but didn't 5.* to 6.0 break addin compatibility? Why not improve the API while those breaking changes are still early, rather than let them stick around and manifest and cause headaches for years to come?

slluis: We'll allow breaking the public API on specific major releases (such as 6.0), but not at every release. Right now the public API is frozen, for good or for bad. There will be chances in the future to fix it, if necessary.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
slluis: API stability is much more important right now than API clarity. We can't evaluate the cost and impact of every single API change proposal. We can't force add-in developers to rebuild their add-ins for every MD release. That's why any PR that breaks the public API will be rejected.

main/src/addins/MonoDevelop.UnitTesting/Services/TestSessionEventArgs.cs
Therzok: Likewise, `GeneralTestOptions` was a public class. Please add it back and mark it as Obsolete, with a `[Obsolete("Will be removed in 7.0")]` attribute, rather than completely removing it.

PR #1527, opened 2016-06-27
Initial work to make the Mac toolbar accessible

iainx:
Requires Xamarin.Mac 2.9.2+ (master)

Adds initial work to make the Cocoa parts of the toolbar accessible.
- Run button
- Status bar
- Search bar
- The Debug button bar is partly accessible, and the runtime/configuration selector is not. I need to investigate some more about how accessibility works with them.
- The Button bar needed some new API to make it nicer for accessibility, but I've left in the old API so it won't break. If people use the old API then it just won't be accessible.

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/BaseView.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/BaseView.cs
@@ -5,14 +5,22 @@ namespace MonoDevelop.VersionControl
 	public abstract class BaseView : ViewContent
 	{
 		readonly string name;
+		readonly string accessibilityDescription;
 
-		protected BaseView (string name)
+		protected BaseView (string name, string description = "")
 		{
 			ContentName = this.name = name;
+			accessibilityDescription = description;
 		}
 
 		public override string TabPageLabel {
 			get { return name; }
+		}
+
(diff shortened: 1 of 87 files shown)


iainx:
@slluis this is the toolbar now as fully accessible as it can be for the moment, until we get the Gtk parts accessible. Would it be possible to merge it now, so that it doesn't bit rot over time until all the Gtk bits are ready?

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/BaseView.cs
```diff
@@ -5,14 +5,22 @@ namespace MonoDevelop.VersionControl
 	public abstract class BaseView : ViewContent
 	{
 		readonly string name;
+		readonly string accessibilityDescription;
 
-		protected BaseView (string name)
+		protected BaseView (string name, string description = "")
```
slluis: This is an api break. Please add an overload.

iainx: Fixed in commit e3a153e0b79c6000e746713bd17d94f8b174dd80

iainx:
Closing this branch as it is using my fork of Monodevelop and there is an a11y branch on the main fork now.

PR #1528, opened 2016-06-27
[Refactoring] Fix analysis tasks not cleared on solution being closed

mrward:
After analyzing the source the diagnostic errors and warnings are
added to TaskService and displayed in the Errors window. On closing
the solution and then opening another solution the status bar would
still show the warning and error icons for these diagnostics tasks.

The tasks created for the analysis errors and warnings are now
associated with the project or solution so they are removed when
the solution is closed.


Therzok:
Fixes  part of https://bugzilla.xamarin.com/show_bug.cgi?id=42210

PR #1529, opened 2016-06-28
Add support for building projects with MSBuild 14.1 (OSS)

radical:


main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/MonoTargetRuntime.cs
```diff
@@ -174,12 +174,26 @@ internal protected override IEnumerable<string> GetGacDirectories ()
...
+					return path;
+				}
+
+				// ToolsVersion >= 14.1 is supported only by msbuild, so, just
+				// return null here
+				return null;
+			}
+
+			//fall back to xbuild. it only has one version, which lives in Mono's 4.5 directory.
+			var xbpath = Path.Combine (monoDir, "4.5");
```
Therzok: I think this should use `toolsVersion`

Therzok: Nevermind, just read the comment.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdePreferences.cs
```diff
@@ -77,6 +77,7 @@ internal IdePreferences ()
 
 		public readonly ConfigurationProperty<bool> BuildBeforeExecuting = ConfigurationProperty.Create ("MonoDevelop.Ide.BuildBeforeExecuting", true);
 		public readonly ConfigurationProperty<bool> BuildBeforeRunningTests = ConfigurationProperty.Create ("BuildBeforeRunningTests", true);
+		public readonly ConfigurationProperty<bool> BuildWithMSBuild = ConfigurationProperty.Create ("MonoDevelop.Ide.BuildWithMSBuild", PropertyService.Get<bool> ("MonoDevelop.Ide.BuildWithMSBuild"));
```
slluis: This property should be defined in RuntimePrefrences (https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs#L495) and all code should get the value of MonoDevelop.Ide.BuildWithMSBuild through that property.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -930,6 +927,8 @@ public static bool TrySplitResourceName (string fname, out string only_filename,
 
 					MonoDevelop.Core.Execution.RemotingService.RegisterRemotingChannel ();
 					var pinfo = new ProcessStartInfo (exe) {
+						Arguments = binDir,
```
slluis: Let's use the standard input to provide the bin dir. It is already being used to provide the parent process id.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdePreferences.cs
radical: Do I still need the `PropertyService.Get` call for the default value then? Or is that handled automatically and I can just use `false` here?

radical: Looking at the code, it pokes the value in via `PropertyService`, the first time it changes from the default value. I am updating the code to just use `false` as the default value.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/MonoTargetRuntime.cs
akoeplinger: That's not accurate. XBuild 14.0 lives in /Library/Frameworks/Mono.framework/Versions/4.4.1/lib/mono/xbuild/14.0/bin/xbuild.exe for example.

PR #1530, opened 2016-06-28
[VCS] Optimize status querying

Therzok:
This refactors the code so both code paths use the same logic.

The problem of the old code was that for directory statuses, we'd query
the whole tree for conflicts and iterate all the index entries when
setting the info, thus doing an index refresh each time.

In comparison old and new case, we have these relative values:
[old] 128 OnGetDirectoryVersioninfoCalls with 222000ms
[new] 48 OnGetDirectoryVersionInfo calls with 50000ms

This optimizes by approximately 40%.


Therzok:
Another note: The old code path stayed 13000ms/222000ms in `RetrieveStatus` from LibGit2Sharp, while the new variant uses 32506ms/50000ms. This is probably due to the conflicts grabbing via index iteration from the earlier variant, thus having all the results already cached.

PR #1533, opened 2016-06-28
[Ide] Improve handling of unsaved files in floating windows.

mrward:
Fixed bug #31345 - Save file closed and confirmation popup is
appearing in Notebook window.
https://bugzilla.xamarin.com/show_bug.cgi?id=31345

Saved files are now no longer closed in a floating window before
unsaved files are saved. This allows the user to cancel the save and
all the files that were previously open will remain open in the
floating window.

When multiple unsaved files are open and the floating window is closed
only a single save file dialog will now be displayed with all the
files listed, each with a check box allowing a file to be saved or
not. Previously a save file dialog would appear for each unsaved file.
If there is only one unsaved file then the standard save file dialog is
used, the same as if you were closing a single unsaved file, that does
not have the check boxes.


Therzok:
LGTM! :+1:

slluis:
@mrward can you make a PR for cycle8? thanks

PR #1534, opened 2016-06-29
[Ide] Optimize RecentFiles updates.

Therzok:
This modifies RecentFiles updates so that we do the following:
* Split Projects/Files into different events. This way we can guarantee
that for the WelcomePage, for example, we can subscribe only to project
changes that will cause a refresh of the items in the list.

* Don't trigger updates of recent items if nothing has changed
between the old and the new version.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentOpen.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentOpen.cs
@@ -54,11 +54,47 @@ namespace MonoDevelop.Ide.Desktop
 		{
 			recentFiles = new RecentFileStorage (storageFile);
 			recentFiles.RemoveMissingFiles (fileGroup);
+
+			recentFiles.RecentFilesChanged += OnRecentFilesChanged;
 		}
-		
+
+		void OnRecentFilesChanged (object sender, RecentItemsChangedEventArgs args)
+		{
+			changed?.Invoke (sender, args);
+			if (args.Groups.Any (i => i.Contains (projGroup)))
+				projectsChanged?.Invoke (sender, args);
+			if (args.Groups.Any (i => i.Contains (fileGroup)))
+				filesChanged?.Invoke (sender, args);
+		}
+
+		event EventHandler changed;
 		public override event EventHandler Changed {
-			add { recentFiles.RecentFilesChanged += value; }
-			remove { recentFiles.RecentFilesChanged -= value; }
+			add {
+				changed += value;
+			}
+			remove {
+				changed -= value;
+			}
+		}
+
+		event EventHandler projectsChanged;
+		public override event EventHandler ProjectsChanged {
+			add {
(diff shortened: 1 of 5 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentOpen.cs
```diff
@@ -55,10 +55,60 @@ public FdoRecentFiles (string storageFile)
...
 			recentFiles.RemoveMissingFiles (fileGroup);
 		}
+
+		Dictionary<EventHandler, EventHandler<RecentItemsChangedEventArgs>> changedDict = new Dictionary<EventHandler, EventHandler<RecentItemsChangedEventArgs>> ();
 		
 		public override event EventHandler Changed {
-			add { recentFiles.RecentFilesChanged += value; }
-			remove { recentFiles.RecentFilesChanged -= value; }
+			add {
+				EventHandler<RecentItemsChangedEventArgs> ev = (sender, e) => {
```
Therzok: I don't like this, but we don't have covariance on eventhandler subscription inside an `add/remove` block.

slluis: A more simple solution is to subscribe recentFiles.RecentFilesChanged just once in the FdoRecentFiles constructor, and in the event handler, raise the Changed, ProjectsChanged or FilesChanged depending on what you get.

Therzok:
Code is better now.

slluis:
LGTM

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentOpen.cs
```diff
@@ -54,11 +54,47 @@ public FdoRecentFiles (string storageFile)
...
 		}
-		
+
+		void OnRecentFilesChanged (object sender, RecentItemsChangedEventArgs args)
+		{
+			changed?.Invoke (sender, args);
+			if (args.Groups.Any (i => i.Contains (projGroup)))
+				projectsChanged?.Invoke (sender, args);
+			if (args.Groups.Any (i => i.Contains (fileGroup)))
+				filesChanged?.Invoke (sender, args);
```
alanmcgovern: What's the point in having 3 distinct events if you pass the same information into each event? The person using the event still has to filter everything, so it's just more confusing than it needs to be. Right?

Therzok: The event is simple `EventArgs`. Unless they force cast, they get no data. I can just use `EventArgs.Empty` instead.

Therzok: The idea is that the WelcomePage can't randomly insert elements anyway, it just clears all of them and reinserts them.

PR #1535, opened 2016-07-01
[Ide] Allow existing project files to be overwritten when adding files

mrward:
Fixed bug #20204 - Adding existing items to a folder does not offer
overwrite option
https://bugzilla.xamarin.com/show_bug.cgi?id=20204

Adding existing files to a folder in a project would not allow the
existing files to be overwritten. The existing project files would
need to be deleted first. Now when adding an existing file to a folder
in a project where they already exist the user is prompted to
overwrite the existing file.


slluis:
@mrward we should let the user decide whether to overwrite a file that already exists or not. The dialog should be a question dialog with options to Overwrite, Skip, Cancel and Apply to all checkbox.

mrward:
OK I have changed the code so the dialog has a overwrite, skip, cancel and apply to all check box when multiple files are being added/moved. If a single file is being added the dialog is the same as it was before with just the cancel and overwrite button.

slluis:
Great, it looks good to me now.

slluis:
Can you please make a PR for C8?

PR #1536, opened 2016-07-04
[Ide] Fix failure to drag and drop file

mrward:
Fixed bug #38073 - Failure when dragging file from Project A to
Project B then back to Project A
https://bugzilla.xamarin.com/show_bug.cgi?id=38073

After adding a file to a project and then trying to drag that file
to another folder the file would not be dragged. The problem was that
selection was disabled by the ContextMenuTreeView and not reset
since the button release event is sometimes not fired.


mrward:
This is a small change to the [commit](https://github.com/mono/monodevelop/commit/d0bc3fc0b290ff213ce673836a1d829a9dfdc9eb) to fix bug 18653. @DavidKarlas

slluis:
@DavidKarlas can you review?

DavidKarlas:
Wondering why release event doesn't fire always :/

PR #1542, opened 2016-07-06
Remove extraneous mime types added in d891cf0

jleclanche:
The added mime types would cause monodevelop to have priority when
opening text or html files over more appropriate applications such
as web browsers or text editors.

Additionally, this would cause opening some text-based formats to open
monodevelop without actually opening the file because of a bug in the
file opening code.


jleclanche:
I'm not signing a CLA I can't read without signing up for it first...

Consider this change public domain / cc0, please copy it and retain authorship.

slluis:
You don't really need to signup anything, you'll just be asked to authorize xamarin to get your email address from your github account, that's all.

jleclanche:
It's asking me for my email, full name, address and phone number without giving me an option to even read the agreement beforehand.

I also now have an account on cla.xamarin which I have no way to delete. Not great.

PR #1545, opened 2016-07-06
Fixed 'Bug 41727 - Documentoutline Pad - When an item on document

sgmunn:
outline is clicked the caret is not moved to item on text editor'


slluis:
LGTM

PR #1557, opened 2016-07-11
MonoTargetRuntime: Use correct xbuild dirs for toolsVersion 12.0/14.0 or

radical:
.. fallback to `$prefix/lib/mono/4.5` if toolsVersion==4.0 .

This restores the older behavior from before the msbuild switch was
added.


radical:
cc @akoeplinger , @mhutch

PR #1558, opened 2016-07-11
[Ide] Use the project that is passed from Workspace FileXXX events instead of the ProjectFile.Project

sgmunn:
When files are removed from a project, ProjectFile.Project is null. We
can use the value of project from the event args instead.

Fixes bug #42345


sgmunn:
@slluis something like this? for https://bugzilla.xamarin.com/show_bug.cgi?id=42345

slluis:
Yes, it looks good

PR #1559, opened 2016-07-11
[Ide] Use the project that is passed from Workspace FileXXX events instead of the ProjectFile.Project

sgmunn:
When files are removed from a project, ProjectFile.Project is null. We
can use the value of project from the event args instead.

Fixes bug #42345

### DIFF
--- a/main/src/addins/TextTemplating/MonoDevelop.TextTemplating/GenerateCommandHandler.cs
+++ b/main/src/addins/TextTemplating/MonoDevelop.TextTemplating/GenerateCommandHandler.cs
@@ -41,13 +41,13 @@ namespace MonoDevelop.TextTemplating
 
 	class GenerateCommandHandler : CommandHandler
 	{
-		protected override void Run ()
+		protected async override void Run ()
 		{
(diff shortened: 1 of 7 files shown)


slluis:
I forgot one thing. You should be able to register a SingleProjectFileCustomTool subclass using CustomToolExtensionNode, in which case a wrapper would not be needed.

sgmunn:
<ExtensionPoint path = "/MonoDevelop/Ide/CustomTools" name = "Custom code generation tools">
	<Description>Extensions to the text editor. Classes must be a subclass of TextEditorExtension.</Description>
	<ExtensionNode name="Tool" type="MonoDevelop.Ide.Extensions.CustomToolExtensionNode"/>
</ExtensionPoint>

Does that description look like a copy+paste error to you?

-------------

So, `CustomToolExtensionNode` defines an extension node and has a `Tool` property of type `ISingleFileCustomTool`. Would your expectation be that in CustomToolExtensionNode.Tool we get the instance from GetInstance and if it is a ISingleFileCustomTool we then wrap it and if it is a SingleProjectFileCustomTool the we return it as is?

PR #1560, opened 2016-07-11
MSBuild(oss): Fix the msbuild assemblies' HintPaths for 14.1 builder

radical:
Don't hard code
`/Library/Frameworks/Mono.frameworks/Versions/Current/...`, instead
construct the path from `$(MSBuildToolsPath)`


radical:
PR #1561 should fix the failing test

PR #1563, opened 2016-07-12
Traffic decrease

Therzok:
https://github.com/mono/monodevelop/commit/ba4c744b29c7ee493de9100af432ff029eeaf248 and https://github.com/mono/monodevelop/commit/eecb9bfe7adf4152db08214f7c3c8233fa9b9ce6 are general purpose commits. The commit most important commit is https://github.com/mono/monodevelop/commit/50ebb30a842d05d765ca0de052246fd55038aa92.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/IMSBuildPropertySet.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/IMSBuildPropertySet.cs
@@ -231,7 +231,7 @@ namespace MonoDevelop.Projects.MSBuild
 				if (i != -1) {
 					int fi = val.IndexOf ('\\');
 					if (fi != -1 && fi < i) i = fi;
-					sb.Append (val.Substring (0,i));
+					sb.Append (val, 0,i));
 				} else
 					i = 0;
 				for (int n = i; n < val.Length; n++) {
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/IMSBuildPropertySet.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/IMSBuildPropertySet.cs
@@ -231,7 +231,7 @@ namespace MonoDevelop.Projects.MSBuild
 				if (i != -1) {
 					int fi = val.IndexOf ('\\');
 					if (fi != -1 && fi < i) i = fi;
-					sb.Append (val, 0,i));
+					sb.Append (val, 0,i);
 				} else
 					i = 0;
 				for (int n = i; n < val.Length; n++) {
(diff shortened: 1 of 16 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/IMSBuildPropertySet.cs
```diff
@@ -258,14 +258,32 @@ static string EncodeString (string val)
...
+			int start = 0;
+			int end = val.Length;
+
+			for (; start < val.Length; ++start) {
+				if (val [start] == ' ' || val [start] == '\t')
+					continue;
+				break;
+			}
+			for (int i = end - 1; i >= start; --i) {
+				if (val [i] == ' ' && val [i] != '\t') {
```
slluis: This condition should be `val [start] == ' ' || val [start] == '\t'`

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/IMSBuildPropertySet.cs
```diff
@@ -258,14 +258,32 @@ static string EncodeString (string val)
...
+				return string.Empty;
+			
+			if (val [start] == '\"') {
+				start++;
+				end--;
+			}
+			if (val [start] == '@') {
 				StringBuilder sb = new StringBuilder (val.Length);
-				for (int n = 1; n < val.Length; n++) {
+				for (int n = start + 1; n < val.Length; n++) {
```
slluis: End condition should use `end` instead of `val.Length`

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/IMSBuildPropertySet.cs
Therzok: Good catch.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/IMSBuildPropertySet.cs
```diff
@@ -278,7 +296,7 @@ static string DecodeString (string val)
 				return sb.ToString ();
 			}
 			else
-				return val;
+				return val.Substring (start, end);
```
slluis: Second argument is a length, so it should be `end - start`

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/IMSBuildPropertySet.cs
Therzok: Also, I think you mean `val[i]`, since we're iterating from the end, we don't want to check the start again.

Therzok:
Fixed.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/IMSBuildPropertySet.cs
```diff
@@ -258,14 +258,32 @@ static string EncodeString (string val)
...
+			int start = 0;
+			int end = val.Length;
+
+			for (; start < val.Length; ++start) {
+				if (val [start] == ' ' || val [start] == '\t')
+					continue;
+				break;
+			}
+			for (int i = end - 1; i >= start; --i) {
+				if (val [i] == ' ' && val [i] == '\t') {
```
Therzok: Fixed.

Therzok:
Should be okay now.

PR #1575, opened 2016-07-18
Added an AppData file for Linux app stores

Mailaender:
This is displayed in GNOME and KDE software installers.
See https://people.freedesktop.org/~hughsient/appdata/


alanmcgovern:
approve

PR #1579, opened 2016-07-22
Support for loading and building PCL5/netstandard projects

mhutch:
Requires Mono 4.6 as of a couple days ago


knocte:
If this requires Mono 4.6, shouldn't this include a change in configure.ac?

knocte:
@mhutch ping?

mhutch:
@knocte no. It requires Mono 4.6 for the netstandard projects to be able to build correctly, but that's no different from e.g. PCL projects which require the PCL profiles to be installed.

PR #1581, opened 2016-07-25
[NuGet] Load package images on the UI thread

sevoku:
This is not really a bug, but Xwt images should always be loaded in the UI thread. It didn't break anything in the NuGet dialog, but has the potential to cause failures in the Xwt backend code (i.e. if the current Xwt toolkit is not Gtk for some reason).

### DIFF
--- a/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/ImageLoader.cs
+++ b/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/ImageLoader.cs
@@ -99,9 +99,8 @@ namespace MonoDevelop.PackageManagement
 		{
 			try {
 				Stream stream = GetResponseStream (uri);
-				Image image = Image.FromStream (stream);
-
-				return new ImageLoadedEventArgs (image, uri, state);
+				var loader = Runtime.RunInMainThread (() => Image.FromStream (stream));
+				return new ImageLoadedEventArgs (loader.Result, uri, state);
 			} catch (Exception ex) {
 				return new ImageLoadedEventArgs (ex, uri, state);
 			}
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/ImageLoader.cs
```diff
@@ -99,23 +99,29 @@ ImageLoadedEventArgs LoadImage (Uri uri, object state)
...
 		static Stream GetResponseStream (Uri uri)
 		{
+			WebResponse response = null;
 			if (uri.IsFile) {
 				var request = WebRequest.Create (uri);
+				response = request.GetResponse ();
 				return request.GetResponse ().GetResponseStream ();
+			} else {
+				var httpClient = new HttpClient (uri);
+				response = httpClient.GetResponse ();
```
alanmcgovern: Is this right? We still have a return statement right after this line

alanmcgovern: I mean the branch above the one I commented on :p

mrward: Yep you are correct. I missed that.

@sevoku - Can you fix the `if (uri.IsFile)` block? Should just be a case of removing the return line there so the memory stream is used.

mrward: @sevoku - I have removed that extra return statement on master. So there is nothing for you to do now.

sevoku: Oh, sorry, I missed that. Thanks guys!

PR #1582, opened 2016-07-26
Fixed 'Bug 42795 - View >> Focus Document is not functioning.'

mkrueger:


kdubau:
Was this also committed to `cycle7` ?

PR #1583, opened 2016-07-26
Improve command-line builds

mhutch:


slluis:
If we change the default value of enable-release, then we need to make sure the build bots do explicitly enable it, somehow.

mhutch:
@slluis as it was it was broken, it was impossible to configure a debug build with the top level configure script.

slluis:
@mhutch I'm just saying that if we commit as is, we won't be producing release builds on Wrench anymore. So we either find a way to enable release builds on Wrench, or we make the root configure script to play nice with --enable-release.

mhutch:
My feeling is that release builds should be op-in, but it's up to you.

borgdylan:
Current command-line builds on Linux are painful! All are related to loading gtk-sharp/cairo-sharp.

PR #1599, opened 2016-08-15
[Cycle8][Core] Fix C# formatting policy not saved for solution

mrward:
Fixed bug #43246 - Custom code formatting (Indentation) settings are
not stored and reset after the solution restart.
https://bugzilla.xamarin.com/show_bug.cgi?id=43246

When a single C# formatting option was changed so it matched the
inherited policy then the solution was not saved since the policy
information in the solution was not being updated correctly.
When a setting  matches the inherited policy it is removed from the
properties that are to be saved with the solution file. However if
that property was originally stored in the solution file it would not
be removed when the solution was saved. A simple way to reproduce this
is to create a new solution, open the Solution Options dialog, select
Source Code - Code Formatting - C# source code, C# format, Edit,
then change Indent case sections in the Indentation category, close
and re-open the solution. If this repeated a few times then eventually
the option will not have the correct value. At this point changing
the option will not update the solution file at all.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Serialization/DataItem.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Serialization/DataItem.cs
@@ -28,6 +28,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace MonoDevelop.Core.Serialization
 {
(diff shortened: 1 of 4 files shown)


mrward:
This change breaks the existing test: SolutionTests.KeepUnknownCustomData

mrward:
Updated the pull request with a different fix. The builds are now green. This could do with being reviewed by someone who knows the serialization and policy service code.

sgmunn:
@Therzok @DavidKarlas can either one of you see a problem with this PR, can we merge it?

Therzok:
:+1: from me

PR #1602, opened 2016-08-29
[Core] Changed processing of remote process messages/responses to sequential execution

DavidKarlas:
So code processing messages doesn't need to handle multithreading
Discovered this problem while looking into "Bug 43522 - MonoDevelop.UnitTesting.NUnit.NUnitAssemblyTestSuite.RunUnitTest throws error on large test suite" when running a lot of unit tests at once, list containing results of unit tests got null value inside because List<T> is not thread safe... Throwing all kind of random errors.
I suspect this might also fix that problem.


slluis:
I would rather fix the problem in the NUnit code. It should not assume that callbacks are all executed sequentially. Even if this change will fix this specific issue, it is not a general fix.

PR #1608, opened 2016-09-05
Japanese translation update

plonk:


slluis:
Hi, unfortunately we won't be able to take contributions to the Japanese translation anymore. This translation is now being done at Microsoft, and the localization process doesn't have the possibility of merging external contributions. If you find any issue in the translation, please [file a bug](http://xamar.in/r/file_studio_bug) and the localization team will take care of it. Thanks!

PR #1610, opened 2016-09-05
[Cycle8] Synchronizing package icons with our other products

vancura:


vancura:
Closing since this PR is too old, I'll create a new one based on `master` in a while.

PR #1612, opened 2016-09-05
[Stub] Implement support for hijacking language if none is set in XS.

Therzok:
This issue is new in Sierra, where the preferred user interface
language contains regional information (which it should not).

In El Capitan, having Language set to English and Region set to Romania
would set [NSLocale currentLocale] to en_RO, while [NSLocale
preferredLanguage[0]] to en.

In Sierra, both now happen to be en_RO, which causes issues if you have
multiple preferred languages and as a result, causes bug
https://bugzilla.xamarin.com/show_bug.cgi?id=43806

### DIFF
--- a/main/build/MacOSX/monostub.m
+++ b/main/build/MacOSX/monostub.m
@@ -14,6 +14,8 @@
 
 #include "monostub-utils.h"
 
+#import <Foundation/Foundation.h>
+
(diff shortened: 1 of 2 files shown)


slluis:
@iainx hey, can you peer review this PR?

PR #1618, opened 2016-09-08
MSBuild parser optimizations

Therzok:


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -817,8 +817,8 @@ internal static bool FromMSBuildPath (string basePath, string relPath, out strin
 				if (fpath == null) {
 					// Part of the path does not exist. Can't do any more checking.
 					part = Path.GetFullPath (part);
-					for (; n < names.Length; n++)
-						part += "/" + names[n];
+					if (n < names.Length)
+						part += "/" + string.Join ("/", names, n, names.Length - n);
 					resultPath = part;
```
jstedfast: I wonder if it'd be better to build the `part` string by allocating a `char[]` and then manually copying the chars into it and then doing `part = new string (charArray)`.

Therzok: Good idea, will test it.

Therzok: Didn't improve it much, verbosity would be useless there for the impact it has.

Therzok:
These are the allocation statistics for this profiling session. We reduce the global allocations by 300mb on full Main.sln solution load (this is including roslyn finishing its work)

old-solution-load
old-solution-load-new

Therzok:
:eyes: please @slluis

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
jstedfast: nod, I agree that if it isn't much savings, better to keep code that is more readable

Therzok:
Numbers went down a bit more. Down to 3.5GB of allocations now.

Therzok:
I just noticed, we regressed the optimization from here:
https://github.com/mono/monodevelop/commit/ee93eadde4f7336266ef3592f1a2f3cb8648232f
via this commit:
https://github.com/mono/monodevelop/commit/8b17d75ff2dc26c5aa1a6246f6a2ce7d4f65e3be#diff-977a7fdc4f12ac06a9e04465483843d9R380

We need to find a better way to reuse sbs. A queue of stringbuilders? :p

Therzok:
Brought down to 3.2GB of allocations. Not sure I can go down to 2.3gb again, as the change I initially did assumed we'd never have recursive evaluations.

old-solution-load-new2

slluis:
LGTM

Therzok:
Performance with a stack of sbs is ~ the same. Merging.

PR #1621, opened 2016-09-12
[UITest] Redirect stderr when running from autotest

manish:


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestClientSession.cs
```diff
@@ -84,19 +85,30 @@ public int StartApplication (string file = null, string args = null, IDictionary
...
+			
 			pi.EnvironmentVariables ["MONO_AUTOTEST_CLIENT"] = sref;
 			if (environment != null)
 				foreach (var e in environment)
 					pi.EnvironmentVariables [e.Key] = e.Value;
 
-			process = Process.Start (pi);
+			process = new Process ();
+			process.StartInfo = pi;
+			process.OutputDataReceived += (sender, e) => {
```
Therzok: Don't we want `ErrorDataReceived` here?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestClientSession.cs
```diff
@@ -84,19 +85,30 @@ public int StartApplication (string file = null, string args = null, IDictionary
...
+				lock (output) {
+					output.Append (e?.Data);
+				}
```
alanmcgovern: You should only append e.Data if it's not null. This will probably cause null references to things reading the string data

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestClientSession.cs
```diff
@@ -84,19 +85,30 @@ public int StartApplication (string file = null, string args = null, IDictionary
...
+			};
+			if (process.Start ())
+				throw new Exception ("Process start failed");
```
alanmcgovern: You missed a ! Here

iainx:
Go ahead and merge once the comments have been dealt with

manish:
Looks like I screwed up this PR big time.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestClientSession.cs
```diff
@@ -84,19 +85,31 @@ public int StartApplication (string file = null, string args = null, IDictionary
...
 				foreach (var e in environment)
 					pi.EnvironmentVariables [e.Key] = e.Value;
 
-			process = Process.Start (pi);
+			process = new Process ();
+			process.StartInfo = pi;
+			process.ErrorDataReceived += (sender, e) => {
+				lock (errors) {
+					if (e != null && e.Data != null)
+						errors.Append (e.Data);
```
alanmcgovern: the data is line buffered so you should use `AppendLine` otherwise you'll end up with a single unreadably long string

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestClientSession.cs
```diff
@@ -84,19 +85,31 @@ public int StartApplication (string file = null, string args = null, IDictionary
...
 			pi.EnvironmentVariables ["MONO_AUTOTEST_CLIENT"] = sref;
 			if (environment != null)
 				foreach (var e in environment)
 					pi.EnvironmentVariables [e.Key] = e.Value;
 
-			process = Process.Start (pi);
+			process = new Process ();
+			process.StartInfo = pi;
+			process.ErrorDataReceived += (sender, e) => {
+				lock (errors) {
```
Therzok: There is no need to lock here.

alanmcgovern:
Actually why not just reuse ProcessWrapper, or one of the other XS Process helpers, which gives a much simpler API to interact with. This patch doesn't wait for stderr to flush either so more data can be written to the builder after the method has returned

PR #1623, opened 2016-09-13
[msbuild] Add support for building with msbuild 15.0 instead of 14.1

radical:
OSS MSBuild bundled with the mono OSX installer is being updated from
14.1 to 15.0 (toolsVersion), and this updates our support for it.

Any new XS builds installed on OSX with this code will have 15.0
installed and not 14.1 . So, we just disable support for building
projects with 14.1 .

The app.config needs to be updated though because in our earlier build
the element name for extension paths was
`msbuildExtensionsPathSearchPaths`, but that was changed when it was
merged upstream to `projectImportSearchPaths`.

Also, the references in the project file are updated such that the
project can build if either 14.1 or 15.0 msbuild is available. And
at runtime, it will switch to 15.0 assemblies. The check for 14.1 can be
removed with the next release.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
@@ -873,7 +873,7 @@ namespace MonoDevelop.Projects.MSBuild
 
 		static string GetNewestInstalledToolsVersion (TargetRuntime runtime, out string binDir)
 		{
-			var supportedToolsVersions = new [] { "14.1", "14.0", "12.0", "4.0" };
+			var supportedToolsVersions = new [] { "15.0", "14.0", "12.0", "4.0" };
 
 			foreach (var toolsVersion in supportedToolsVersions) {
 				binDir = runtime.GetMSBuildBinPath (toolsVersion);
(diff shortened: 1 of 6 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -873,7 +873,7 @@ public static bool TrySplitResourceName (string fname, out string only_filename,
 
 		static string GetNewestInstalledToolsVersion (TargetRuntime runtime, out string binDir)
 		{
-			var supportedToolsVersions = new [] { "14.1", "14.0", "12.0", "4.0" };
+			var supportedToolsVersions = new [] { "15.0", "14.0", "12.0", "4.0" };
```
slluis: Why not just add 15.0 to the list instead of replacing 14.1?

main/src/core/MonoDevelop.Projects.Formats.MSBuild/app.v15.0.config
```diff
@@ -10,38 +10,38 @@
 		<assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
 			<dependentAssembly>
 				<assemblyIdentity name="Microsoft.Build.Framework" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
-				<bindingRedirect oldVersion="0.0.0.0-100.0.0.0" newVersion="14.1.0.0" />
+				<bindingRedirect oldVersion="0.0.0.0-100.0.0.0" newVersion="15.1.0.0" />
```
slluis: shouldn't this be 15.0.0.0?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
radical: - 14.1 will be replaced by 15.0 once this and another related patch is merged.
- 14.1 and 15.0 require slightly different app.config files, as mentioned in the commit message
- The one scenario where supporting it might be useful, is if a dev has C7 with 14.1 installed and they are building XS from source, which would go away once we release with 15.0 support (shortly). With a new release, they would have only 15.0 installed and then this scenario won't be useful any longer.
- It required some temporary hacks to support 14.1 and they would have to be removed after that. So, it seemed an overkill and I opted to just disable building with MSBuild if user has only 14.1 installed.

Note: the project will *build* even if only 14.1 is available. And at runtime, it loads the 15.0 assemblies instead.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/app.v15.0.config
radical: No, the assembly version is 15.1.0.0 . (https://github.com/Microsoft/msbuild/pull/591)

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
mhutch: Why not rename the csproj and assembly info to match?

radical: Should it be renamed to `dotnet.v15.0` or `dotnet.mac_oss` ? It is built only on OSX currently.

mhutch: dotnet.v15.0

it should work with any v15

PR #1624, opened 2016-09-14
Extract Win32 debugger into Mono.Debugging-like API to get rid of dependency on MonoDevelop

nerzhulart:
Hi! I'm Jetbrains Rider developer and in Rider we're using your Mono.Debugging.* libraries to implement debugger. In this change I've cut the dependency on MonoDevelop to be able to use CorDebugger as a lib. As well I have a bunch of fixes and improvements ontop of this change and I'm going to pull all of them into MonoDevelop main repo. It's the first of them.


DavidKarlas:
Can you revert 1st line change in Main.sln?(this might cause problems to someone, I would prefer to keep same BOM or whatever that is)
Can you revert formatting change in MtaThread.cs?(not really that important)
Can you remove signing related changes, I have to admit... I'm not very familiar with signing... But I know I don't like it because it can cause that GAC assembly has priority over local .dll...

DavidKarlas:
Btw... Are you interested in moving this to debugger-libs repo?

nerzhulart:
> Can you revert 1st line change in Main.sln?(this might cause problems to someone, I would prefer to keep same BOM or whatever that is)
Sorry, didn't noticed that BOM was changed, will revert
> Can you revert formatting change in MtaThread.cs?(not really that important)
Ok, will do. And don't you mind if I fix indenting in debugging code, because mixed tabs/spaces is one of the biggest evil in this code?
> Can you remove signing related changes, I have to admit... I'm not very familiar with signing... But I know I don't like it because it can cause that GAC assembly has priority over local .dll...
All Rider assemblies are signed and so it doesn't allow to load unsigned dlls, that's why I've added signing to these projects. Otherwise we have to have different projects for Mono.Debugging.* : without signing for your repo and with signing for us, but it will be good not to do like this. Also Mono.Debugger.dll was already signed, I've taken .snk from this project.

nerzhulart:
> Btw... Are you interested in moving this to debugger-libs repo?

Good idea, and it was the first one when I started to work with this repo. But it requires some tricky actions to prevent git history loosing. If you don't mind I can do it a bit later because now I need to integrate all of our changes

nerzhulart:
Hey, guys! Any new thoughts on assebmly signing? Now you have Mono.Debugging.Soft and Mono.Debugger.Soft that are signed, and Win32 implementation and base API Mono.Debugging is unsigned? Maybe we should unify the signing of these projects?
image
image

nerzhulart:
Also I've fixed the formatting and BOM

alanmcgovern:
Did you see https://github.com/mono/monodevelop/pull/1624#pullrequestreview-341169 as a solution for your signing needs?

nerzhulart:
Yep, I've seen. But I'm confusing by the fact I mentioned above: the half of the assemblies are already signed in your repo.

DavidKarlas:
I created internal CI build lane which will produce .pkg and .msi so I can test this change related to signing change(I'm 99% it's harmless) and then I will 👍

DavidKarlas:
LGTM, I tested changes on Mac and Windows...

slluis:
LGTM

nerzhulart:
Thanks!

PR #1625, opened 2016-09-16
[msbuild] Add support for building with msbuild 15.0 instead of 14.1

radical:
OSS MSBuild bundled with the mono OSX installer is being updated from
14.1 to 15.0 (toolsVersion), and this updates our support for it.

Any new XS builds installed on OSX with this code will have 15.0
installed and not 14.1 . So, we just disable support for building
projects with 14.1 .

The app.config needs to be updated though because in our earlier build
the element name for extension paths was
`msbuildExtensionsPathSearchPaths`, but that was changed when it was
merged upstream to `projectImportSearchPaths`.

Also, the references in the project file are updated such that the
project can build if either 14.1 or 15.0 msbuild is available. And
at runtime, it will switch to 15.0 assemblies. The check for 14.1 can be
removed with the next release.


slluis:
We are not adding new features to cycle 8

PR #1632, opened 2016-09-20
[Templates] Switch PCL profile to 111

Therzok:
Bug 44293 - Incorrect target framework profile for new PCL in XS compared to VS


Therzok:
Merged manually.

PR #1635, opened 2016-09-22
[Core] FilePath no longer creates extra strings on IsChildPathof.

Therzok:
Rewrote the algorithm so it creates fewer strings. The only do the check when we get a match on the first path, so it shouldn't slow down the code a lot.

On 10,000,000 iterations, the benchmark results are as given:

New with slash: 1331ms
Old with slash: 1311ms

New without slash: 1397ms
Old without slash: 3088ms

This fixes memory pressure, slightly makes the slash variant slower, but
greatly improved the variant without slash.


Therzok:
The tests were done on test strings:
`asdf`, `asdf\\`
and
`asdf\\test`

Therzok:
Linux failure unrelated, looks like a bug in gtk# in mono 4.4.

PR #1636, opened 2016-09-23
Editor optimizations

Therzok:
Take the second commit with a grain of salt, as I suck at naming. Any suggestion on what to replace 'SomethingSomething' with?

cc @DavidKarlas this fixes the issue you've been having.
cc @slluis and @mkrueger for sanity reviews.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/SdiWorkspaceWindow.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/SdiWorkspaceWindow.cs
@@ -84,7 +84,6 @@ namespace MonoDevelop.Ide.Gui
 		{
 			this.tabControl = tabControl;
 			this.tab = tabLabel;
-			this.tabPage = content.Control;
 			SetTitleEvent(null, null);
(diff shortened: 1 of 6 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/SdiWorkspaceWindow.cs
```diff
@@ -110,8 +108,6 @@ public SdiWorkspaceWindow (DefaultWorkbench workbench, ViewContent content, Dock
 
 			// The previous WorkbenchWindow property assignement may end with a call to AttachViewContent,
 			// which will add the content control to the subview notebook. In that case, we don't need to add it to box
-			if (subViewNotebook == null)
```
Therzok: I'm not sure if this can/will break anything.

DavidKarlas:
Maybe "SomethingSomething" could be replaced with "realize", "realized"... Similar to what GTK has...

Therzok:
Renamed to use Realized instead of SomethingSomething.

slluis:
LGTM

PR #1637, opened 2016-09-25
[7.0][Perf] Improve enumerators in custom collections.

Therzok:
This makes it so there is no Enumerator created, as the compiler is smart enough to detect that if there is a strongly typed GetEnumerator() method with a return value that structurally matches IEnumerator, it'll prefer that.
In this case, we're talking about structs. This will save us IEnumerator allocations all over the place.


Therzok:
I have no data for this, but it's safe to assume that this will improve allocations caused by using `foreach`

Therzok:
This might be an API/ABI break.

slluis:
It is ok to change the API if it doesn't cause an ABI break. If the ABI breaks, the changes can't go in. This needs to be checked.

Therzok:
It's an ABI break. :(

Therzok:
I'll split this PR into public and internal API.

Therzok:
Merged to vNext.

PR #1638, opened 2016-09-26
[Core] Optimize ProcessWrapper threads usage.

Therzok:
This moves the ProcessWrapper to use tasks instead of threads. The net benefit of this is that we use 60 less threads for loading Main.sln, as the threads are now assigned to threadpool threads.
Reduces total time spent in CaptureError from 60s to 0.8ms and CaptureOutput from 64s to 33ms.


Therzok:
Maybe we should go further and only create one task and do `Task.WhenAny` internally?

alanmcgovern:
It seems fine to me. I don't think there's any real benefit to trying to reduce the tasks further.

slluis:
👍

PR #1644, opened 2016-10-06
[build] Use nuget.exe from nuget-binary's submodule if nuget not found

knocte:
Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=44889


slluis:
Thanks!

PR #1655, opened 2016-10-15
[MSBuild, Perf] Move Token to be a struct

Therzok:
This commit is pretty much self-explanatory. this reduces allocations for
the Token struct, making the code behave better.

Tokens are short-lived, usually not lasting more than two stackframes. The
idea is that a Token is at most 16bytes in size (on 64bit), it is
immutable and it represents a single value.

This change overall causes a huge change in allocations:
* MonoDevelop.Projects.MSBuild.Conditions allocations go down by 120MB on Main.sln
* 144MB of Tokens are no longer created on the heap (7 million obj)
* ConditionTokenizer allocates 12MB more now, as it has two token members.
* ConditionFactoryExpression allocates 12MB more now as it has a token member.

This should also increase performance.


Therzok:
Gathering CPU stats.

Therzok:
Perf wise, the numbers are the same.

PR #1657, opened 2016-10-17
Ensure F# projects are output tracked. Fixes #45556

nosami:
Looks like this hasn't been working for at least 18 months


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService_WorkspaceHandling.cs
```diff
@@ -344,7 +344,7 @@ static void OnSolutionItemRemoved (object sender, MonoDevelop.Projects.SolutionI
 		static void IntitializeTrackedProjectHandling ()
 		{
 			AddinManager.AddExtensionNodeHandler ("/MonoDevelop/TypeSystem/OutputTracking", delegate (object sender, ExtensionNodeEventArgs args) {
-				var projectType = ((TypeSystemOutputTrackingNode)args.ExtensionNode).ProjectType;
+				var projectType = ((TypeSystemOutputTrackingNode)args.ExtensionNode).LanguageName;
```
nosami: I didn't find any extension using OutputTracking except for the F# language binding, so this seems like a safe change to make.

PR #1659, opened 2016-10-19
[Tests] Initial support for benchmarking unit tests.

Therzok:
They don't work yet, they show up as passed, but assemblies aren't loaded because they're shadow copied.

I'll add documentation on how to write stuff, but this is an initial glance.

Each benchmark would need its own class with one `Bench` method.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild.Conditions/ConditionParser.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild.Conditions/ConditionParser.cs
@@ -30,6 +30,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics.Contracts;
 using System.Text;
 using Microsoft.Build.BuildEngine;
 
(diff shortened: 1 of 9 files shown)


slluis:
Which output does this generate?

Therzok:
Default output which is in the format of:
```
 Method |         Median |      StdDev |  Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
------- |--------------- |------------ |------- |------ |------ |------------------- |
  Test1 | 20,892.8483 us | 289.0976 us | 143.00 |     - |     - |         655,768.24 |
  Test2 |    272.9095 us |   2.3111 us |      - |     - |     - |              26.89 |
```

We can configre the output to have the data in whatever format we like:
https://github.com/PerfDotNet/BenchmarkDotNet/tree/master/src/BenchmarkDotNet.Core/Exporters

Therzok:
Also waiting on the cross platform GC stats that don't use Windows ETW.

Therzok:
Closing for now but leaving the branch alive. Might be useful, but we now have a profiler-integrated infrastructure.

PR #1660, opened 2016-10-24
Use portable flavor for F#

nosami:
Adds DotnetProjectFlavor for F# portable libraries. XS expects this to be used in a few places such as the `.NET Portable Subset` node under references and the new `Service Capabilities` node.


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.Extensions/SolutionItemExtensionNode.cs
```diff
@@ -55,6 +55,9 @@ public class SolutionItemExtensionNode: ProjectModelExtensionNode
 		[NodeAttribute ("alias", Description = "Friendly id of the extension")]
 		public string TypeAlias { get; set; }
 
+		[NodeAttribute ("language", Description = "Language of the extension")]
+		public string LanguageName { get; set; }
```
nosami: Needed to add LanguageName here to differentiate between C#PortableLibrary and F#PortableLibrary which share the same flavor guid.

Therzok:
Was this intended?

PR #1661, opened 2016-10-25
Remove all use of GNOME-specific libraries from GnomePlatform

directhex:
GnomePlatform relies on some *super* deprecated libraries, from old versions of GNOME 2. All of the things we use GNOME# for can be done better using more modern replacements, e.g. GIO (part of modern GLib). It'd be nice to have some Gtk+ 2.14+ methods, but that's another story for another day.

For now, this greatly simplifies the dependency chain on Linux, and makes it easier to handle some FlatPak cases in future work.

### DIFF
--- a/main/src/addins/GnomePlatform/Gio.cs
+++ b/main/src/addins/GnomePlatform/Gio.cs
@@ -56,13 +56,21 @@ namespace MonoDevelop.Platform {
 		[DllImport (gio, CallingConvention = CallingConvention.Cdecl)]
 		static extern IntPtr g_file_info_get_content_type (IntPtr handle);
 		[DllImport (gio, CallingConvention = CallingConvention.Cdecl)]
+		static extern IntPtr g_file_info_get_icon (IntPtr fileinfo);
+		[DllImport (gio, CallingConvention = CallingConvention.Cdecl)]
+		static extern IntPtr g_file_new_for_path (IntPtr path);
+		[DllImport (gio, CallingConvention = CallingConvention.Cdecl)]
 		static extern IntPtr g_file_new_for_uri (IntPtr uri);
 		[DllImport (gio, CallingConvention = CallingConvention.Cdecl)]
 		static extern IntPtr g_file_query_info (IntPtr handle, IntPtr attrs, int flags, IntPtr cancellable, out IntPtr error);
+		[DllImport (gio, CallingConvention = CallingConvention.Cdecl)]
+		static extern IntPtr g_icon_to_string (IntPtr icon);
 		[DllImport(gio, CallingConvention = CallingConvention.Cdecl)]
 		static extern IntPtr g_settings_get_string(IntPtr gsettings, IntPtr key);
 		[DllImport(gio, CallingConvention = CallingConvention.Cdecl)]
 		static extern IntPtr g_settings_new(IntPtr schema);
+		[DllImport (glib, CallingConvention = CallingConvention.Cdecl)]
+		static extern void g_error_free (IntPtr raw);
 		[DllImport (glib, CallingConvention = CallingConvention.Cdecl)]
 		static extern void g_list_free (IntPtr raw);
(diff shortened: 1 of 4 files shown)


main/src/addins/GnomePlatform/Gio.cs
```diff
@@ -142,6 +148,29 @@ public static string GetGSettingsString(string schema, string key)
...
+			IntPtr native = GLib.Marshaller.StringToPtrGStrdup (filename);
+			IntPtr gfile = g_file_new_for_path (native);
+			GLib.Marshaller.Free (native);
+			IntPtr native_attrs = GLib.Marshaller.StringToPtrGStrdup ("standard::icon");
+			IntPtr error;
+			IntPtr info = g_file_query_info (gfile, native_attrs, 0, IntPtr.Zero, out error);
+			if (error != IntPtr.Zero)
+				return null;
+			IntPtr icon = g_file_info_get_icon (info);
+			string[] iconid = GLib.Marshaller.Utf8PtrToString (g_icon_to_string (icon)).Split (' ');
```
iainx: Does the return of g_icon_to_string need to be freed here?

directhex: .................................................................yes.

Therzok: Use PtrToStringGFree.

main/src/addins/GnomePlatform/Gio.cs
```diff
@@ -124,6 +134,43 @@ public static System.Collections.Generic.IList<DesktopApplication> GetAllForType
...
+		{
+			if (String.IsNullOrEmpty (filename))
+				return null;
+			IntPtr native = GLib.Marshaller.StringToPtrGStrdup (filename);
+			IntPtr gfile = g_file_new_for_path (native);
+			GLib.Marshaller.Free (native);
+			IntPtr native_attrs = GLib.Marshaller.StringToPtrGStrdup ("standard::icon");
+			IntPtr error;
+			IntPtr info = g_file_query_info (gfile, native_attrs, 0, IntPtr.Zero, out error);
+			if (error != IntPtr.Zero)
```
Therzok: Free the error if it's not IntPtr.Zero, via `g_error_free`

main/src/addins/GnomePlatform/Gio.cs
```diff
@@ -124,6 +134,43 @@ public static System.Collections.Generic.IList<DesktopApplication> GetAllForType
...
+				return null;
+			IntPtr icon = g_file_info_get_icon (info);
```
Therzok: https://developer.gnome.org/gio/stable/GFileInfo.html#g-file-info-get-icon

We don't own the icon.

main/src/addins/GnomePlatform/Gio.cs
```diff
@@ -124,6 +134,43 @@ public static System.Collections.Generic.IList<DesktopApplication> GetAllForType
...
+			string[] iconid = GLib.Marshaller.Utf8PtrToString (g_icon_to_string (icon)).Split (' ');
+			g_object_unref (icon);
```
Therzok: https://developer.gnome.org/gio/stable/GFileInfo.html#g-file-info-get-icon

We don't own the icon.

main/src/addins/GnomePlatform/Gio.cs
```diff
@@ -142,6 +150,32 @@ public static string GetGSettingsString(string schema, string key)
...
 		}
 
+		public static string GetIconIdForFile (string filename)
+		{
+			if (String.IsNullOrEmpty (filename))
+				return null;
+			IntPtr native = GLib.Marshaller.StringToPtrGStrdup (filename);
+			IntPtr gfile = g_file_new_for_path (native);
+			GLib.Marshaller.Free (native);
+			IntPtr native_attrs = GLib.Marshaller.StringToPtrGStrdup ("standard::icon");
```
iainx: native_attrs needs freed

PR #1665, opened 2016-11-02
MSBuild 15 detection - Latest Mono

mrward:
MSBuild 15 in the latest Mono 4.9.0.671 no longer ships as MSBuild.exe but is instead msbuild.dll. So the detection logic was failing to find the MSBuild bin directory. This breaks building the monodevelop code from source and using MSBuild to build projects inside the IDE.


mrward:
@radical - Yeah sorry my fault. I thought it was msbuild.dll but it is definitely MSBuild.dll so this needs to change.

mrward:
OK I changed the case of MSBuild.dll and committed the updated commits to master. Closing this pull request.

PR #1666, opened 2016-11-03
[MSBuild] Guard against empty lines in a solution file.

Therzok:
Skip lines which only contain whitespace characters, as they should not be parsed.
Bug 11494 - Exception on opening solution file after Perforce converts line endings to \r\r\n

cc @mhutch


mhutch:
lgtm

PR #1667, opened 2016-11-03
[Ide][Bug 46332] Fix preselection of grouped templates in new project dialog

sevoku:
SelectTemplate(templateId) does not take into account
that a template can be part of a group, where the parent
category references only the first template.

This patch adds an additional check for template group
members. If a group contains the requested template,
it selects the group in the dialog and the target language
accordingly.

(fixes bug #46332)

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
@@ -346,7 +346,26 @@ namespace MonoDevelop.Ide.Projects
 
 		void SelectTemplate (string templateId)
 		{
-			SelectTemplate (template => template.Id == templateId);
+			SolutionTemplate matchedInGroup = null;
+			SelectTemplate (template => {
+				if (template.HasGroupId) {
+					var inGroup = template.GetTemplate ((t) => t.Id == templateId);
+					// check if the requested template is part of the current group
+					// becasue it may be not referenced by a category directly.
+					// in this case we match/select the group and change the selected
+					// language if required.
+					if (inGroup?.Id == templateId) {
+						matchedInGroup = inGroup;
+						return true;
+					}
+				}
+				return template.Id == templateId;
+			});
+
+			// make sure that the requested language has been selected
+			// if the requested template is part of a group
+			if (matchedInGroup != null)
+				SelectedLanguage = matchedInGroup.Language;
 		}
 
 		void SelectFirstAvailableTemplate ()


mrward:
Do you have an example of what the problem was before this fix was added? As in is there a template where I can see the problem? Just trying to understand what the problem is.

Normally I would not expect the same template id to be not part of the same group. Or maybe I am not understanding something here :)

mrward:
OK now I have thought about it some more this makes more sense now.

So you want to pre-release a template with a particular id and because it is in a group you may not find the id because it would be part of a group and not necessarily the first solution template you find that is added as the parent to the project dialog.

Does this template only have one language? Just wondering why you are selecting the language here. The language selection is the previously used language not the template's language. If a template is grouped it normally means it has multiple languages. Not sure I understand the language selection part.

sevoku:
@mrward NUnit Library is a good example, there are 3 templates: `MonoDevelop.CSharp.NUnitLibraryProject`, `MonoDevelop.VBNet.NUnitLibraryProject` and `MonoDevelop.FSharp.NUnitLibraryProject`. The first loaded is `MonoDevelop.FSharp.NUnitLibraryProject` (depends on which addin is loaded first) and lands correctly in the appropriate section of the dialog. Now when the other 2 are being loaded, they'll be grouped together, but the section tree contains only `MonoDevelop.FSharp.NUnitLibraryProject` (which refs the others internally as a group).

Now if you want to create a new NUnit Lib in C# you'd pass `MonoDevelop.CSharp.NUnitLibraryProject` as a parameter, but the section loop would never hit it, because it selects the group templates. So my solution is to let it select the same group template (`MonoDevelop.FSharp.NUnitLibraryProject` in the example) and switch the language to the desired one which would make the dialog select the requested `MonoDevelop.CSharp.NUnitLibraryProject` template.

This is of course only possible because we group the templates by language only and its more like a hackaround, but I think a better solution could only be to redesign how the controller handles templates in general. We should target this at some point, but this should be enough as a quick fix for now.

sevoku:
Sorry I missed that part:

> Does this template only have one language? Just wondering why you are selecting the language here. The language selection is the previously used language not the template's language. If a template is grouped it normally means it has multiple languages. Not sure I understand the language selection part.

A template has only one language, but a group consists for several templates for different languages. So I select the group which is not necessarely the requested template (e.g. `FSharp.NUnitLibraryProject` instead of `CSharp.NUnitLibraryProject` because FSharp is the first template in the group) and the language.
The Dialog picks up the language from the controller on initalization here: https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs#L86, so it works in both ways (get users selection and tell the dialog which language to pre-select)

PR #1672, opened 2016-11-10
Bug 45274 - Cannot change the code formatting policy

DavidKarlas:
Problem was that Policies have 3 ItemData values with name “TextStylePolicy”(one for plain text, one for C# and one for XML).
Since commit 42b7775af9 when UpdateFromItem was called it overwrote same “TextStylePolicy” 3x, instead of creating 3 separate TextStylePolicy each for one language
With this change.. Once TextStylePolicy is updated it’s removed from list so on next loop, next TextStylePolicy is updated

@slluis can you please review this?

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Serialization/DataItem.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Serialization/DataItem.cs
@@ -70,6 +70,7 @@ namespace MonoDevelop.Core.Serialization
 
 		internal void UpdateFromItem (DataItem item, HashSet<DataItem> removedItems)
 		{
+			var items = new List<DataNode> ();
 			foreach (var d in item.ItemData) {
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Core.Serialization/DataItem.cs
```diff
@@ -70,6 +70,7 @@ public DataNode Extract (string name)
 
 		internal void UpdateFromItem (DataItem item, HashSet<DataItem> removedItems)
 		{
+			var items = new DataCollection ();
```
Therzok: We're doing a costly operation on adding items to this temporary collection. We should only care about the final collection. Can we use a `List<DataNode>` here?

sgmunn:
@DavidKarlas can you PR to cycle9 as well please

PR #1677, opened 2016-11-16
[Core] Handle NRE when determining if a file is compilable

sgmunn:
Fixes bug #46960
File MyClass.cs could not be written. error when adding new shared library to solution
https://bugzilla.xamarin.com/show_bug.cgi?id=46960


sgmunn:
Verified in master.

PR #1680, opened 2016-11-17
[Core] Set ItemId of .Net Core project to .sln value or in case of Reload to old ItemId

DavidKarlas:
Can @mrward or @slluis can you do review?
I'm hoping to get this in before vNext preview 2. Since XS is having problems with changing ItemId on project.


mrward:
Looks good to me. Just wondering if we should maybe make the SolutionFolderItem's itemId internal and use that instead of adding a new defaultItemId field.

DavidKarlas:
I will merge it as is for now...

Lluis if think aproach with setting `itemId` instead of adding new `defaultItemId` field I can refactor it...

PR #1681, opened 2016-11-18
Set LANGUAGE from monostub

bratsche:
This fixes an issue on OSX Sierra where having a rtl language second in your list would cause gtk's direction to be set to rtl.

https://forums.xamarin.com/discussion/82412/xamarin-studio-has-messed-up-and-now-shows-all-its-own-text-ui-from-right-to-left-cant-fix-it

Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=41969


alanmcgovern:
Should we backport this to C8 too? That's going to be in stable for the next few months

sgmunn:
yes, good idea.

PR #1690, opened 2016-11-26
Gettext Editor + fixing translations support

FallenPhoenix:
See commits de992a6 - d71d968.

[de992a6] First, I fixed line counting in Gettext Editor additionaly to commit you cherry picked before.
How it worked before: scanner extracted line 1 as 0 and when user clicked reference, editor turned it back to 1: int.Parse(line) + 1. So conversions sequence was 1 → 0 → 1. As I changed start line index to 1, I now also deleted that increment to make it 1 → 1 → 1 (not 1 → 1 → 2).

[b199de7] IKVM.Reflection project has method named "GetString", but it not for Gettext. So it only occurs garbage messages in po-files.

[368b670] "Run Configuration" string for project settings has wrong format style. I fixed it to make "Run Configuration: {0}" translatable string.

[d71d968] Some entries of main menu was not translatable.


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.OptionPanels/RunConfigurationsPanel.cs
```diff
@@ -164,7 +164,7 @@ public RunConfigurationOptionsDialogSection (RunConfigInfo configInfo): base (ty
 		{
 			this.RunConfiguration = configInfo.EditedConfig;
 			Label = configInfo.EditedConfig.Name;
-			HeaderLabel = GettextCatalog.GetString ("Run Configuration: " + configInfo.EditedConfig.Name);
+			HeaderLabel = string.Format(GettextCatalog.GetString ("Run Configuration: {0}"), configInfo.EditedConfig.Name);
```
Therzok: :+1:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/ViewCommands.cs
```diff
@@ -141,10 +141,15 @@ public class LayoutListHandler : CommandHandler
 		static LayoutListHandler ()
 		{
 			NameMapping = new Dictionary<string, string> ();
-			NameMapping ["Solution"] = "Code";
-			NameMapping ["Visual Design"] = "Design";
-			NameMapping ["Debug"] = "Debug";
-			NameMapping ["Unit Testing"] = "Test";
+			NameMapping ["Solution"] = TranslateString ("Code");
```
Therzok: These should be calling `GettextCatalog.GetString`

Therzok:
I'll manually merge this PR and cherry-pick everything _except_ the translation regen.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/ViewCommands.cs
FallenPhoenix: GettextCatalog.GetString is method for actually extraction of translation. Not sure about TranslateString, but as I understand its a dummy method which only marks string for RegexScanner, when real getting translated string can be later. For example:
```
var str = TranslateString("original string");
// some code
GettextCatalog.GetString(str);
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.OptionPanels/RunConfigurationsPanel.cs
mhutch: FWIW, GettextCatalog.GetString already calls `string.Format` so you can do

    GettextCatalog.GetString ("Run Configuration: {0}", configInfo.EditedConfig.Name);

Therzok:
Dupe of #1733

PR #1693, opened 2016-11-28
OnBeforeSaved event added to hookup addins

alexsorokoletov:
Hi! 
This pull request adds new `BeforeSaved` event to the `DocumentContext` class in the IDE so that any add-in can use this event to format/cleanup code before file is saved.

This will be very useful for add-ins like [SortRemoveUsings](https://github.com/alexsorokoletov/XamarinStudio.SortRemoveUsings). Right now add-in works with Saved event and when it removes/sorts usings, file becomes modified so user has to save file twice.
With new `BeforeSaved` event extension will be able to sort/remove usings before file saved and will work better from a user prospective.

Maybe there is even better way for that (Ideally would be cool to run add-in on every Cmd+S/Ctrl+S call, why not?).

I've discussed this event with @mhutch during MVP Summit Hackathon.

Please let me know what could be improved in this pull request.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/DocumentContext.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/DocumentContext.cs
@@ -162,6 +162,15 @@ namespace MonoDevelop.Ide.Editor
 				handler (this, e);
 		}
 
+		public event EventHandler BeforeSaved;
+
+		protected virtual void OnBeforeSaved (EventArgs e)
+		{
+			var handler = BeforeSaved;
+			if (handler != null)
+				handler (this, e);
+		}
+
 		internal virtual Task<IReadOnlyList<Editor.Projection.Projection>> GetPartialProjectionsAsync (CancellationToken cancellationToken = default(CancellationToken))
 		{
 			return null;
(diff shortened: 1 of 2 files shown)


slluis:
I feel adding an overridable Save method in TextEditorExtension would be a better solution.

alexsorokoletov:
@slluis thank you for idea. Does it mean I will have to create new kind of a C# text editor extension and replace default one with it? Is that how you imagine it?

slluis:
You would create a new TextEditorExtension subclass and override the Save method. The extension would be registered in the editor extension chain.

alexsorokoletov:
Thank you, @slluis! 
Do you have in mind any examples that would help me on this path?

Therzok:
@slluis The TextEditorExtension does not seem to have an overridable save method.

https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/TextEditorExtension.cs#L38

alexsorokoletov:
Interesting that @xamarin-cla-bot said the NDA is signed. My employer did not sign the NDA so I cannot continue with this contribution.

Sorry for that, dear MonoDevelop team. Maybe you will go ahead and add the API yourself or someone else could contribute similar functionality my extension could take use of.

PR #1696, opened 2016-12-02
[UnitTesting] Run all the selected unit tests.

rolfbjarne:
Run all the selected unit tests, not just one of them.

Partially fixes #19843.

https://bugzilla.xamarin.com/show_bug.cgi?id=19843


main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestService.cs
```diff
@@ -128,9 +126,7 @@ public static AsyncOperation RunTest (UnitTest test, MonoDevelop.Projects.Execut
 						return;
 
 					await RefreshTests (cs.Token);
-					test = SearchTest (testName);
```
slluis: This code can't be removed. This search is done here because the RefreshTests method may cause the 'test' instance to be recreated, so it may not be valid anymore after the call.

rolfbjarne: The problem I ran into was that SearchTest didn't find the test, because it's not in the tree of tests (it's a `UnitTestSelection` test which is created on-the-fly containing the selected tests only).

Is there any other way around this?

slluis: A better approach would be to add support for executing multiple tests in UnitTestService. Add a RunTests method that takes a list of tests to run. It would need to check buildOwnerObject and determine which projects needs to be built (tests could belong to different projects). After building, it could internally create a UnitTestSelection grouping all the tests and start a session to run it.

rolfbjarne: I tried something like that first, but it turned out there is more code that thinks there will be only one test: the Test Results pad would have to updated as well, since it would only show the last test that was executed (and it didn't look like a simple fix to change it to support more than one root test).

IOW it turned into a significant refactoring, which is why I went with this approach, which was much simpler and fairly self-contained.

I can have a look at it again though.

slluis: What I propose is to create the UnitTestSelection object after building the project(s). In this way there is still only one root test (like in your approach), composed by the selected tests, that have been properly updated after the build.

rolfbjarne: Good idea, that could work. Let me try.

rolfbjarne:
OK, I've reworked it as you suggested now.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestService.cs
```diff
@@ -108,29 +108,43 @@ public static AsyncOperation RunTest (UnitTest test, MonoDevelop.Projects.Execut
...
+					}
+
+					var test_names = new HashSet<string> (tests.Select ((v) => v.FullName));
 
 					await RefreshTests (cs.Token);
-					test = SearchTest (testName);
-					if (test != null)
-						await RunTest (test, context, false, checkCurrentRunOperation, cs);
+
+					tests = test_names.Select ((fullName) => SearchTest (fullName)).Where ((t) => t != null);
```
slluis: It would be good to do a .ToList() here, since this enumeration is navigated several times.

slluis:
It looks good. Thanks!

PR #1697, opened 2016-12-02
Bug 48234 - Unable to determine which `Application Output` pad belongs to which project

DavidKarlas:
@slluis can you do quick check if I did something wrong?
I added SetName to OperationConsole. Logic here is... Console has beside `TextWriter Out`, `TextWriter In`... also string Name(aka. Window Title)...

Once debugger finds out process name(this is same as name of process in ThreadsPad) and calls SetName on Console... This is better then just Project name, because in case if we attach or debug .exe directly without .csproj...

Then Console takes care of setting new title on parent Pad...


slluis:
I think it would be easier to just add a targetName argument to ProgressMonitorManager.GetRunProgressMonitor(). Some users of that monitor never access the underlying console object.

DavidKarlas:
Fixed it differently

PR #1699, opened 2016-12-05
Bug 48234 - Unable to determine which `Application Output` pad belongs to which project

DavidKarlas:
This is different aproach from https://github.com/mono/monodevelop/pull/1697
It sets suffix when Application Output is created and not from debugger when process is started/connected to.

Id of pad needs to be suffixed with title otherwise on reuse title can be wrong. Only downside of this is that we can end with many open Application Output pads because we don't reuse pads with different titles.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/ProgressMonitors.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/ProgressMonitors.cs
@@ -206,9 +206,11 @@ namespace MonoDevelop.Ide.Gui
 						if (mon.TypeTag == id) {
 							if (mon.InstanceNum > instanceCount)
 								instanceCount = mon.InstanceNum;
-							if (mon.Title == originalTitle)
+							if (mon.Title == originalTitle && !mon.AllowReuse)
 								usedTitleIds.Add (mon.TitleInstanceNum);
-							if (pad == null && mon.AllowReuse) {
+							if (mon.AllowReuse &&
+							   (pad == null ||
+							    mon.Title == originalTitle)) {//Prefer reusing output with same title(project)
 								pad = mpad;
 							}
 						}
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/ProgressMonitors.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/ProgressMonitors.cs
@@ -146,7 +146,6 @@ namespace MonoDevelop.Ide.Gui
 		{
 			if (!string.IsNullOrEmpty (titleSuffix)) {
 				title += " - " + titleSuffix;
-				id += titleSuffix;//We need different Id to prevent sufixes like "Application Output - MyProject (2)" where (2) is index 2 of AppOutput Id
 			}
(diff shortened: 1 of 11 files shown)


slluis:
We have to reuse pads with different titles. For example, if I a solution has projects A and B, if I start A, stop, then start B, it should reuse the same output pad.

DavidKarlas:
@slluis I pushed new commit that removes need for different Ids by doing checking if pads have same title and finding smallest free " (#)" suffix

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Pad.cs
```diff
@@ -63,6 +63,7 @@ internal Pad (DefaultWorkbench workbench, PadCodon content)
 		
 		public string Title {
 			get { return window.Title; }
+			set { window.Title = value; }
```
slluis: Thist setter should not be public. The implementation of a pad has api for changing the title.

DavidKarlas:
@slluis I updated commit, can you check if it's OK now?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/ProgressMonitors.cs
```diff
@@ -78,9 +78,9 @@ private ProgressMonitor GetBuildProgressMonitor (string statusText)
 			return mon;
 		}
 		
-		public OutputProgressMonitor GetRunProgressMonitor ()
+		public OutputProgressMonitor GetRunProgressMonitor (string titleSuffix = null)
```
slluis: This is an abi break. Please add a new overload.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/ProgressMonitors.cs
```diff
@@ -142,8 +142,11 @@ public OutputProgressMonitor GetOutputProgressMonitor (string title, IconId icon
 			return GetOutputProgressMonitor (null, title, icon, bringToFront, allowMonitorReuse, visible);
 		}
 		
-		public OutputProgressMonitor GetOutputProgressMonitor (string id, string title, IconId icon, bool bringToFront, bool allowMonitorReuse, bool visible = true)
+		public OutputProgressMonitor GetOutputProgressMonitor (string id, string title, IconId icon, bool bringToFront, bool allowMonitorReuse, bool visible = true, string titleSuffix = null)
```
slluis: another abi break

DavidKarlas:
@slluis I fixed ABI breaking change, see any other issue?

slluis:
The behavior is not yet correct. Let's say you have a solution with projects A and B. Issues I found:

* Run project A. The title of the output channel is changed to A. Wait for it to end. Now run A again. The title of the pad changes to A (1). It should not change.
* Run project A. While A is running, start project B. Pads with titles A and B are created. That's correct. Now start B again. Instead of reusing the pad with title B that already exists, it uses the pad that had title A and changes it to B (2).

DavidKarlas:
@slluis any more comment on this PR? I fixed both issues.

slluis:
It looks good now. Please merge to master too.

PR #1700, opened 2016-12-05
[Mac] Only notify ActiveRuntime changed once on menuitem selection.

Therzok:


Therzok:
cc @sgmunn this fixes the bug for me

PR #1711, opened 2016-12-12
List recently used templates in the New Project dialog

sevoku:
This PR adds a "Recently used" templates list to the New Project dialog:

Recently used Templates

* the recent templates list is stored in `[LocalConfigDir]/RecentlyUsedTemplates.xml` using [RecentFileStorage](https://github.com/mono/monodevelop/blob/recent-templates/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentFileStorage.cs)
* a template will be added to the list after each successful project creation
* the list is language specific, creating the same project type with different languages will add the template for each language once
* the original category and language is shown for each template in the list


slluis:
@sevoku hey, can you fix the conflicts?

mrward:
This is already in vNext.

PR #1721, opened 2016-12-13
Bug 49500 - Null reference when installing a NuGet package

DavidKarlas:
This bug happened with 512768ca3 and 7356c217 aka https://github.com/mono/monodevelop/pull/1636
Reason for this bug was that TextEditor was Disposed only when underlying GTK editor was destroyed which resulted in leaking extensions(which kept event handlers causing exception). Ever since commits mentioned above GTK editor is not created until tab is selected(visible) hence GTK editor was never destroyed(hence no disposing)…


slluis:
Don't forget to fix in master...

PR #1728, opened 2016-12-22
CorDebug fixes

nerzhulart:
* MetadataType.GetPropertyImpl() implementation
* Properly getting module for native types in RuntimeInvoke, because there were exceptions
* try{} block removed, null check for 'method' removed, because it's unnecessary
* try-catch in OverloadResolve()


DavidKarlas:
Looks like there are some conflicts, can you rebase?

nerzhulart:
@DavidKarlas Fixed!

PR #1733, opened 2016-12-25
Translation support and some fixes (updated request #1690)

FallenPhoenix:
See starting from commit ba1733f.
I did remade something from previous pull request and added something.

There some trouble with Russian language and formatting tags, I didn't found yet how to fix it good. But at least avoiding using tags in translation works fine.

Note for 118b252: There was unneded word wrapping in tab labels because of wrong calculation.


slluis:
It looks like this PR was created from the wrong branch.

PR #1734, opened 2016-12-26
[CorDebug] More extensive search of pdb files

nerzhulart:
More extensive search of pdb files that allows to load pdb specified in .dll/.exe debug info. (E.g. pdb file can be located at different location than dll/exe)


nerzhulart:
Could you please merge this ASAP? I have another important change dependent on this.

PR #1746, opened 2016-12-31
Please, add the issues tab!

AndersBillLinden:
Mailing lists are a pain for many reasons.
The same goes for creating accounts on bugzillas/bugtrackers here and where.
Please, let developers pick their preferred information channel and add the "Issues" tab!
It cant hurt, can it?


slluis:
I'm sorry, but we track bugs for the whole Mono organization in Bugzilla. Our whole development process depends on having bugs handled in a single place.

PR #1750, opened 2017-01-06
Make framework check more robust. Fixes #49518

nosami:
https://bugzilla.xamarin.com/show_bug.cgi?id=49518


Therzok:
Why is `framework.Assemblies` empty though?

slluis:
Good question, why is this happening?

nosami:
Not sure. Will investigate

nosami:
Took me a ridiculously long time to work out the cause of this error.. but this was it https://github.com/mono/monodevelop/commit/dc6a7646e17aea9378a4fc44c8b668f6de7de8a4

PR #1752, opened 2017-01-07
Handle either Gtk.Application.Quit() or NSApplication.Terminate()

bratsche:
After the GLib mainloop exits, call NSApplication.Terminate() to make sure Cocoa terminates correctly and shuts down threads like those used for animations and such.

If NSApplication.Terminate() is called first, catch that call and cancel termination, instead calling Gtk.Application.Quit() so that we shutdown the GLib mainloop. This will in turn call NSApp.Terminate() a second time, which will *not* cancel termination and will perform
Cocoa cleanup as described above.

https://bugzilla.xamarin.com/show_bug.cgi?id=14789

### DIFF
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -153,6 +153,22 @@ namespace MonoDevelop.MacIntegration
 			GlobalSetup ();
 			timer.EndTiming ();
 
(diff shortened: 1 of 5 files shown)


alanmcgovern:
lgtm

alanmcgovern:
Just wondering if we can get another build on wrench before we merge it. Why did windows and mac fail?

PR #1767, opened 2017-01-12
[Ide] Run editor commands only if the editor is focused

sevoku:
Allows editor widgets (search and goto line) to handle all keys.

(fix bug #49816)

(cherry picked from commit 1774b260b1c37c16b47e6cec07e8609782d1777c)


slluis:
I'm merging, but this is a hack.

sevoku:
I wouldn't call it a hack, cause the commands are supposed to work inside the editor only. We do already check it though the handler chain (if the editor is not focused, it won't receive the command at all), but it doesn't cover internal editor widgets. Sure its a workaround, but I don't see a reason to list *default* editing bidnings in the bindings panel separately (and we have no design for it). This will be also important for a new rename dialog.

Therzok:
> internal editor widgets

Seems like a command routing issue?

PR #1769, opened 2017-01-12
Fixed Mono Build on Windows.

simonegli8:
Those projects did not build on Windows, so they were lacking in the
Bin\MSBuild folder of Xamarin Studio, and therefore Mono xbuild was
broken.


simonegli8:
Signed the CLA.

PR #1772, opened 2017-01-12
Master translation fixes

Therzok:
https://github.com/mono/monodevelop/pull/1733, but on top of master and without the last commit which affects the tabstrip


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageFeedItem.cs
```diff
@@ -210,7 +210,7 @@ void SetDate (string dateString)
 						subtitle = GettextCatalog.GetString ("Yesterday");
 					}
 					else if (days < 7) {
-						subtitle = GettextCatalog.GetString ("{0} days ago", days);
+						subtitle = GettextCatalog.GetPluralString ("{0} day ago", "{0} days ago", days, days);
```
Therzok: This is not necessary, but I'm ok with it /cc @FallenPhoenix 

We already check for days == 1 above

Therzok:
LGTM.

PR #1774, opened 2017-01-12
[DotNetCore] Remove xmlns attribute for Sdk style projects

mrward:
If a project has an Sdk attribute then on saving the xmlns attribute will not be added. If it exists on reading then on saving it will be removed.

There are few other places in the code, shown below, where the MSBuildProject.Schema constant, which holds the MSBuild xml namespace, is used instead of the MSBuildProject's xml namespace. MonoDevelop project extensions and external properties are currently handled. Not sure if we need to change LinuxDeployData and MakefileData to use the project's namespace instead. The others I suspect we will want to change to be consistent even if .NET Core projects are possibly not using them.

- [x] [Patching MSBuildImports](https://github.com/mono/monodevelop/blob/d6f74f3c0029a1025691ce4b9fe722999e7b120c/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildImport.cs#L102)
- [x] [MSBuildProject.FormatElement - unused?](https://github.com/mono/monodevelop/blob/d6f74f3c0029a1025691ce4b9fe722999e7b120c/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs#L1058)
- [x] [MSBuildProperty writing XML values](https://github.com/mono/monodevelop/blob/d6f74f3c0029a1025691ce4b9fe722999e7b120c/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProperty.cs#L200)
- [x] [ProjectConfiguration environment variables](https://github.com/mono/monodevelop/blob/d6f74f3c0029a1025691ce4b9fe722999e7b120c/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectConfiguration.cs#L93)
- [x] [LinuxDeployData](https://github.com/mono/monodevelop/blob/d6f74f3c0029a1025691ce4b9fe722999e7b120c/main/src/addins/Deployment/MonoDevelop.Deployment.Linux/MonoDevelop.Deployment.Linux/LinuxDeployData.cs#L69)
- [x] [MakefileData](https://github.com/mono/monodevelop/blob/d6f74f3c0029a1025691ce4b9fe722999e7b120c/main/src/addins/MonoDevelop.Autotools/MakefileData.cs#L85)

### DIFF
--- a/main/src/addins/Deployment/MonoDevelop.Deployment.Linux/MonoDevelop.Deployment.Linux/LinuxDeployData.cs
+++ b/main/src/addins/Deployment/MonoDevelop.Deployment.Linux/MonoDevelop.Deployment.Linux/LinuxDeployData.cs
@@ -66,7 +66,7 @@ namespace MonoDevelop.Deployment.Linux
 			var data = ser.Serialize (this);
 
 			XmlDocument doc = new XmlDocument ();
-			var writer = new XmlConfigurationWriter { Namespace = MSBuildProject.Schema };
+			var writer = new XmlConfigurationWriter { Namespace = entry.MSBuildProject.Namespace };
 			var elem = writer.Write (doc, data);
 
 			entry.MSBuildProject.SetMonoDevelopProjectExtension ("Deployment.LinuxDeployData", elem);
--- a/main/src/addins/MonoDevelop.Autotools/MakefileData.cs
+++ b/main/src/addins/MonoDevelop.Autotools/MakefileData.cs
@@ -81,8 +81,18 @@ namespace MonoDevelop.Autotools
 
 		public XmlElement Write ()
 		{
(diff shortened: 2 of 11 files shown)


slluis:
I think all the above cases have to be changed to use the project's namespace.

slluis:
Other than that, it looks ok to me.

mrward:
This pull request has been updated so those other cases are now using the project's namespace.

In order to allow non-Core addins to be able to use the project's namespace the MSBuildObject's Namespace property has been changed from internal to public.

Also modified the MSBuildProject's Namespace property so it returns an empty string instead of null for Sdk style projects. This makes it easier to use with certain XML APIs which expect an empty string instead of a null for the XML namespace.

PR #1775, opened 2017-01-12
Fixed way of choosing best sequence point for breakpoints in CorDebuggerSession

ArsenyChernyaev:
Changed insertion of breakpoints in CorDebuggerSession, now they are set to the outermost method, instead of how it was before:
bp1
In this example breakpoint on line 9 is set to the begin of lambda block

Now it looks like this:
bp2


DavidKarlas:
LGTM, waiting for CLA

PR #1780, opened 2017-01-13
[Ide] Fix searchbar unable to be typed in

Therzok:
This fixes the assumption that only one commandmanager would exist. MenuButtonEntry would create its own command manager, that would then intercept all native cocoa keypresses.

Moved native cocoa key registration only to commandmanagers with
TopLevelWindows.

Special shout out to bug 51469 for giving us a repro to fix this.

Bug 40414 - [C7] Quick Search does not work at all.


Therzok:
Don't know what cherry-pick did, why I have whitespace changes and `PerformDefaultNSAppAction` included. It's not used, so harmless

PR #1783, opened 2017-01-13
[DotNetCore] Support custom attributes on MSBuildItem

mrward:
MSBuildItem now has an `void AddKnownAttributes (params string[] names)` method where custom known attributes can be added. When the MSBuildItem is then written the associated properties will be written as attributes instead of child elements.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildItem.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildItem.cs
@@ -41,6 +41,7 @@ namespace MonoDevelop.Projects.MSBuild
 		string exclude;
 		string remove;
 		string update;
+		List<string> customKnownAttributes;
 
 		public MSBuildItem ()
 		{
@@ -73,6 +74,16 @@ namespace MonoDevelop.Projects.MSBuild
 			return knownAttributes;
 		}
 
+		public void AddKnownAttributes (params string[] names)
+		{
+			AssertCanModify ();
+
+			if (customKnownAttributes == null)
+				customKnownAttributes = new List<string> ();
+
+			customKnownAttributes.AddRange (names);
+		}
+
 		internal override void ReadAttribute (string name, string value)
 		{
 			if (name == "Include")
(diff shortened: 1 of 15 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildItem.cs
```diff
@@ -145,7 +186,7 @@ internal override void Write (XmlWriter writer, WriteContext context)
 						lastAttr = null;
 				} while (propIndex < props.Count || knownIndex < knownAtts.Length);
 			} else {
-				var knownAtts = attributeOrder ?? GetKnownAttributes ();
+				var knownAtts = GetKnownAttributesForWrite ().ToArray ();
```
slluis: This ToArray() call can be easily avoided. We can save an allocation.

slluis:
Shouldn't we handle this at project level? For example, have a method like:

    class MSBuildProject {
    ...
    public void AddKnownItemAttribute (string itemName params string[] attributes);
    ...
    }

I think we can assume that all items of the same type will have the same attributes. This would avoid having to create a list for every item that uses custom attributes.

mrward:
Thanks for the review and feedback. I have updated the code so there is now a AddKnownItemAttribute method on the MSBuildProject class. Also removed the ToArray calls in MSBuildItem.

Testing this code the only issues I can see are:

  1. Saving a project that has an existing PackageReference element that has a Version child element does not remove the end PackageReference element. It does change the Version element to use be a Version attribute. So you end up with:

         <PackageReference Include="AutoMapper" Version="5.1.1">
         </PackageReference>
  2. No code handles attributeOrder when it is not null. This is set if the attribute order does not match the known attributes. Most of the time attributeOrder is null. Saving an existing PackageReference element with a Version child element when attributeOrder is not null does not remove the Version child element. I can repro this by doing something like the following with the Label attribute:

         <PackageReference Label="1" Include="AutoMapper">
           <Version>5.1.1</Version>
         </PackageReference>

In both the above cases the project is being saved due to a change unrelated to the PackageReference element, say removing another NuGet package from the project. Updating the NuGet package will remove the existing PackageReference and add a new one.

mhutch:
Shouldn't we handle arbitrary metadata as attributes?

slluis:
This is for setting the preference that some specific metadata for an item should be written as attribute. Arbitrary metadata specified as attribute will be loaded and saved as attribute.

mhutch:
Ah, ok, just wanted to be sure it would be roundtripped :)

slluis:
LGTM

PR #1784, opened 2017-01-13
Breakpoints.Clear() deactivates all the breakpoints

nerzhulart:
Breakpoints.Clear() deactivates all the breakpoints. Otherwise COM exception may be thrown on Detach() or Terminate()


slluis:
Do we really want to remove all breakpoints when the session ends? That doesn't sound like a good idea.

nerzhulart:
@slluis I've got what you're talking about. Will fix ASAP

nerzhulart:
@slluis Fixed https://github.com/mono/monodevelop/pull/1789

slluis:
Thanks!

PR #1785, opened 2017-01-16
[Ide] Fix Solution pad tree corruption

sevoku:
The Gtk.TreeView sort implementation expects the compare function to return consistent results
for both nodes being compared. Hence the result of `NodeBuilder.CompareObjects (nodeB, nodeA)` must always be the opposite of `NodeBuilder.CompareObjects (nodeA, nodeB)`.

This patch is a simple workaround, but its really ugly to get the node types and to compare their names and its also very problematic for external AddIns. We may need a better sorting implementation with some kind of grouping or priorities for example, but I'm not sure how it may look like without badly breaking the API.

(fixes bug #46681)


sevoku:
We have a better fix: https://github.com/mono/monodevelop/pull/1795

PR #1796, opened 2017-01-18
Fixed ObjectDisposed exception and MS.CodeAnalysis.Solution leak in c…

DavidKarlas:
…ase if solution was closed soon after opening

ObjectDisposed exception was caused by the fact that even after project was unloaded TypeSystem was still loading causing ObjectDisposed exception(inside `p.GetSourceFilesAsync`). To prevent this TryLoadSolution now also listens to `src.Token` which is cancelled in `Dispose()`. Another `if (token.IsCancellationRequested)` was added just before calling `p.GetSourceFilesAsync` to avoid project being disposed before call being made.

Cause of memory leak was that `workspace` was added to `workspaces` after `TryLoadSolution` call which meant in case if `TypeSystemService.Unload` was called before `TryLoadSolution` finished that `Unload` didn’t `Dispose` workspace. Fix for this is to add `workspace` to `workspaces` before `TryLoadSolution` call, to ensure `monodevelopSolution` is set at all times(so `GetWorkspace` always work) it’s set via constructor. For better readability it was made `readonly`.

HandleActiveConfigurationChanged is now registered only if `solution != null` so check was removed inside method.

I also noticed TypeSystemService.Load is disposing `Counters.ParserService.WorkspaceItemLoaded.BeginTiming` too soon, so I fixed that.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService_WorkspaceHandling.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService_WorkspaceHandling.cs
@@ -150,11 +150,12 @@ namespace MonoDevelop.Ide.TypeSystem
 			}
 		}
 
-		internal static Task<List<MonoDevelopWorkspace>> Load (WorkspaceItem item, ProgressMonitor progressMonitor, CancellationToken cancellationToken = default(CancellationToken))
+		internal static async Task<List<MonoDevelopWorkspace>> Load (WorkspaceItem item, ProgressMonitor progressMonitor, CancellationToken cancellationToken = default(CancellationToken))
 		{
 			using (Counters.ParserService.WorkspaceItemLoaded.BeginTiming ()) {
 				var wsList = new List<MonoDevelopWorkspace> ();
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService_WorkspaceHandling.cs
```diff
@@ -150,11 +150,12 @@ public static void NotifyFileChange (string fileName, string text)
 			}
 		}
 
-		internal static Task<List<MonoDevelopWorkspace>> Load (WorkspaceItem item, ProgressMonitor progressMonitor, CancellationToken cancellationToken = default(CancellationToken))
+		internal static async Task<List<MonoDevelopWorkspace>> Load (WorkspaceItem item, ProgressMonitor progressMonitor, CancellationToken cancellationToken = default(CancellationToken))
```
Therzok: Why is this async if it still uses `ContinueWith`?

DavidKarlas: I added async+await just for sake of `using (Counters.ParserService.WorkspaceItemLoaded.BeginTiming ()) { ` working correctly... Didn't notice ContinueWith there, hmm

PR #1797, opened 2017-01-19
Add tab navigation buttons to Touch Bar

michaelsavich:
In addition to the above, for both sets of navigations buttons NSSegmentedControl is now used instead of NSButton to make for a more logical appearance.

### DIFF
--- a/main/src/addins/MacPlatform/MainToolbar/AwesomeBar.cs
+++ b/main/src/addins/MacPlatform/MainToolbar/AwesomeBar.cs
@@ -34,6 +34,14 @@ using MonoDevelop.Core;
 using Xwt.Mac;
 using MonoDevelop.Ide;
 
+public enum TouchBarType
+{
+	WelcomePage,
+	TextEditor,
+	Debugger,
+	Preferences
+}
+
 namespace MonoDevelop.MacIntegration.MainToolbar
 {
 	public class AwesomeBar : NSView, INSTouchBarDelegate
@@ -42,6 +50,12 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 		private static NSSegmentedControl navSegments = null;
 		private static NSSegmentedControl tabNavSegments = null;
 		//End variables decl… *sigh*
+
+		internal TouchBarType barType = TouchBarType.TextEditor;
+		internal NSTouchBar touchbar = null;
+
+		//touch bar items that need to be dynamically updated
+		private NSButton touchBarRunButton;
 
 		internal RunButton RunButton { get; set; }
 		internal SelectorView SelectorView { get; set; }
@@ -76,6 +90,7 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 		NSTouchBar MakeTouchBar ()
 		{
 			var aTouchbar = new NSTouchBar ();
+			this.touchbar = aTouchbar;
 			aTouchbar.Delegate = this;
 			aTouchbar.DefaultItemIdentifiers = GetItemIdentifiers ();
 			return aTouchbar;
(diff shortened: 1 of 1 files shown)


michaelsavich:
…if iainx reads this, part of the reason I filed this pull request so soon after the last is I'm not sure how else to contact you. I wanted to talk about the next pull request I'm working on. Currently it seems the Touch Bar rebuilds itself like, every frame or so. The problem is that winds up canceling touches that are in progress at the time, which leads to dropped inputs. I'm going to have to rewrite that section of your code so that the touch bar only changes when it needs to. Currently my plan is to have several "bars", one for the welcome page that lets you select recent projects, a customizable one for the text editor, and a debugging bar that shows the buttons from the button bar among other things. If you see this, feel free to contact me at savichmichael@icloud.com if you want to hash out details. I mean, obviously you'll probably have to review the changes, but I don't want to act like I just barged in and changed some of your code.

michaelsavich:
Oh wait it looks like committing to my repo added some more commits to the extant pull request? It's not quite ready if you include that last commit. Closing for now; will open a new pull request when I'm done with this latest batch of changes.

Therzok:
@michaelsavich Why do you say that the touchbar rebuilds itself every frame? The UpdateTouchBar method is only called once when the AwesomeBar is constructed. The only delegate method the touchbar exposes is `touchBar:makeItemForIdentifier`, which doesn't do any heavy lifting.

michaelsavich:
I added a debug line to UpdateTouchBar() and as you can see it gets called quite a lot. Probably not every frame, but it seems intentional because the multifunction run/build/stop button on the touchbar only changes its icons when a new one is made.

Therzok:
Ah, it's on the 700ms  timer. Got it. @iainx why do we do this? We shouldn't be updating the whoe touchbar on the 700ms timer, only the run button and button bars.

michaelsavich:
@Thezrok Oh, I just realized this is through GitHub, so the images of the logs I attached didn’t show up, though you seem to have figured it out anyway. 

I’ve hit a major problem though, in my latest builds, the touch bar is now customizable, but customizing the touch bar seems have a 50% chance of causing the entire window system to become unresponsive? Like, the MonoDevelop window can’t be interacted with, and selecting other apps causes MonoDevelop to regain focus after a quarter of a second, until eventually not even that works and the whole system appears unresponsive. The dock and menu bar still worked though, so a system restart wasn’t necessary, as quitting MonoDevelop fixes the problem. I ran Instruments on it and the problem seems to be centered around syscalls to psynch_cvwait or something like that. I know almost nothing about concurrency implementation, so I’m at a bit of a loss here. One of the last calls before the aforementioned is libmonosgen's inner_start_thread, which leads into JIT code, which I don’t know much about either and I’m not sure how to symbolicate. I’m hoping somebody here might know how to go about debugging this?

Therzok:
Yeah, the whole update starts on the Gtk's UI thread, so if it's stuck in that (those things take a lot of time to get back to the initial stacktrace), the whole UI will be unresponsive. The dock and menu are pure Cocoa.

We probably need to render that thing on a background thread.

michaelsavich:
By thing, do you mean the customization UI? Right now, it's handled automagically by AppKit, I set NSApplication.sharedApplication.isAutomaticCustomizeTouchBarMenuItemEnabled to true and it does the rest of the work. I could try using the explicit call to toggleTouchBarCustomizationPalette(_:) in an NSOperationQueue but I'm not sure that would make it more asynchronous than the current solution…

main/src/addins/MacPlatform/MainToolbar/AwesomeBar.cs
```diff
@@ -70,45 +89,86 @@ public AwesomeBar ()
...
+				}
+			}
+		Rebuild: //switch current bar
 			NSApplication.SharedApplication.SetTouchBar (MakeTouchBar ());
+		Update: //operations that change bar items but not the bar itself, e.g. validation goes here
+			NSImage runImg = null;
+			string runLabel = null;
+			switch (RunButton.Icon) {
+			case Components.MainToolbar.OperationIcon.Build:
+				runImg = MultiResImage.CreateMultiResImage ("build", "");
```
Therzok: This is not ok. We're creating the image on a 700ms timer which is costly (from streams). These should be cached.

michaelsavich: Ah, sorry! I was refactoring the code that was already there and didn't give that line much thought. I'll assign each image to private fields during the constructor method.

Therzok: By this I mean these things can lockup the UI :)

Therzok:
@michaelsavich No, I meant that we're probably doing too much work in MakeTouchBar and UpdateTouchBar (which are all called on a 700ms update timer), which therefore causes a lot of hiccups.

We probably want to move away from this update model into a better, eventbased model (we know when we switch from welcomepage to solution to document to whatever).

In short, this is a shortcoming of the implementation of how the touchbar is updated. The run button drawing will still need to be on 700ms timer, so are the debug ones, but not all.

michaelsavich:
@Therzok So, you mean adding lines to methods in other files/classes that broadcast the types of state changes we need to observe? Or does MonoDevelop already have such a system in place?

Therzok:
@michaelsavich that's a problem I keep bringing up with my team. MonoDevelop does not have a system in place, I think it's time to bite the bullet and go for a better command system approach, rather than evaluating the status of a command on a 700ms timer.

In case the touchbar doesn't work at all (lockups), we really need to do better.

michaelsavich:
Yeah, though I think more important than speed, without such a system stuff can be really fragile. There's no way to tell if a property is being used to determine state by some class on the other side of the codebase. 

By the way, have you seen the animation for customizing the touchbar? I ask because the animation is really weird and hides all windows among other things, so I'm not certain that MonoDevelop's performance is the problem, it could be due to assumptions made elsewhere?

michaelsavich:
I moved the image loading code to the constructor, but the bug still persists. I doubt it's a straight performance issue, something is getting stuck. 

I confirmed that the bug can be reproduced on any Mac with the touch bar simulator in Xcode, so if anybody wants to test it you can build my repository and try customizing the touch bar-- just make sure you are not on the welcome page, as that is a separate bar. The bug doesn't occur every time, but if you add or remove an item in the customization menu before dismissing it, you'll probably run into the bug in 1-3 tries.

Therzok:
@michaelsavich I do not have a touchbar-ready macbook. iainx is also using the simulator for testing this.

michaelsavich:
@Therzok I figured odds are that would be the case, which is why I made sure the issue was reproducible on devices without touchbars before writing about it above. Anyway, I'm not giving up on this bug per se but my next steps might involve building Mono with Console.WriteLine() calls in it, so I'm noting these steps for posterity… in the meantime I'll ifdef out the customization options, re-add the debug/button bar functionality that is missing in my builds and open a pull request. College starts back up on Monday, so in case I don't have much free time… well, I'm trying to get as much done as possible.

iainx:
Yeah, as @Therzok says the command stuff refreshes every 700ms and needs fixed. Not sure what to do with it, we'll need to discuss it more with the team.

If you want to get in touch with the mlnodevelop team we've got some mailing lists that are probably the best place to discuss things like this: http://www.mono-project.com/community/help/mailing-lists/
monodevelop-devel-list is probably most appropriate for this discussion.

Thanks for working on this.

PR #1799, opened 2017-01-19
[MSBuild] Send “Process Stop” message to RemoteProcess so MSBuild pro…

DavidKarlas:
…cess terminates itself

Dispose did call `doneEvent.Set();` and for which https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/Main.cs#L47 was waiting but problem was that https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/RemoteProcessServer.cs#L151 is not background thread, hence process didn't quit, Sending "Process" "Stop" is probably cleanest way to stop process(https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/RemoteProcessServer.cs#L168).


slluis:
I committed a more complete fix.

PR #1801, opened 2017-01-22
Multiple-bar touchbar interface 

michaelsavich:
Hopefully this pull request isn't too big, but it took me this long to get around to reimplementing all of the features present in the base fork. The touchbar now presents different bars depending on the context:
*Welcome Page bar, which displays recent solutions from the Welcome Page allowing for easy access
*Text Editor bar, which is basically the main bar, it is intended to be user-customizable, but a bug means that functionality is currently disabled
*Debug bar, which presents the debug controls
*There are references to a Preferences bar, currently unimplemented, which would allow the user to scrub through tabs in the preferences window

Basically this is a first pass at figuring out how the touchbar features would work, both in terms of UI and structuring in code. This feature isn't done, and I'd like to stress nobody should be afraid to improve upon what I've written here.

### DIFF
--- a/main/src/addins/MacPlatform/MainToolbar/AwesomeBar.cs
+++ b/main/src/addins/MacPlatform/MainToolbar/AwesomeBar.cs
@@ -66,7 +66,8 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 		internal NSTouchBar Touchbar = null;
 
 		//touch bar items that need to be dynamically updated
-		private NSButton TouchBarRunButton;
+		private NSButton touchBarRunButton;
+		private bool rebuildTouchbar = false;
 
 		internal RunButton RunButton { get; set; }
 		internal SelectorView SelectorView { get; set; }
@@ -124,7 +125,7 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 		public void UpdateTouchBar ()
 		{
 			if (Touchbar == null) {goto Rebuild;} //initialize on launch
-
+			if (rebuildTouchbar)  {goto Rebuild;}
 			if (MonoDevelop.Ide.WelcomePage.WelcomePageService.WelcomePageVisible) {
 				if (BarType != TouchBarType.WelcomePage) {
 					BarType = TouchBarType.WelcomePage;
@@ -158,8 +159,8 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 				break;
 			}
 			if (runImg != null) {
-				if (TouchBarRunButton != null) {
-					TouchBarRunButton.Image = runImg;
+				if (touchBarRunButton != null) {
+					touchBarRunButton.Image = runImg;
 				}
 			}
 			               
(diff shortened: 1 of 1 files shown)


DavidKarlas:
In case of "Debug bar, which presents the debug controls" I would expect at least StepIn, StepOver and StepOut actions...

But more in general I don't like how everything is hardcoded for example `public static class Id //unique IDs for TouchBar items` I would prefer if this Ids were generated from Command Id, with maybe some global prefix.

What should be goal/design of this whole thing is...
We should have hardcoded scenarios like Preferences window and Welcome page, where UI/TouchBar is very specific...

But for other cases like coding, debugging, unit testing... We should respect layouts:
image
And have some defaults that we think are best...

This is important part:
But give user option in Preferences to modify that like keyboard bindings per layout and rearrange and add/remove any command user wants... In case we don't have icon for that command we should use text name of command...

michaelsavich:
@DavidKarlas Thanks for the feedback. This is very much a "first pass" at the UI so your ideas are genuinely useful. 

Generating the IDs in a more global way seems like a good idea. Probably make a function that returns an ID for a Command.

It's interesting you brought up unit testing… I wonder what controls would be useful in that context? It's definitely worth brainstorming.

The step in-out-over buttons are on the to-do list, but do present a problem in that MD must be in focus to use them, making them marginally less useful. There is a private API Xcode uses to work around this, and since MD isn't in the App Store, we could use it... we just have to be careful to code it so if the API breaks MD isn't negatively affected.

Unfortunately Apple's customization UI enforces a finite number of Touch Bar items, so we'd still have to hardcode the subset of commands to expose as Touch Bar controls. It's possible to create our own customization interface but it'd take a lot of work. Apple's customization UI is actually wired up in this build but it's sealed off with an ifdef because it seems to cause the whole app to freeze, some bug in Xamarin.mac I think.

DavidKarlas:
Re-Run last unit test would be good start, maybe navigate to failed test source code... Thats why we need to give users ability to set any command in Preferences... They know what is best for them...
I wasn't thinking specifically about unit testing... I'm just saying it should be generic on per Layout basis...

About MD needs being in focus, thats not issue, even today MD needs to be in focus to listen to keyboard shortcuts for stepping and stuff, thats not really issue imo.

michaelsavich:
Yeah, but just to make sure we're clear, there are a finite number of Touch Bar controls we can reasonably display in the customization UI. Think 20 as close to the upper limit. As such can't let the users map to any command without implementing a custom customization UI… and since there is no API to detect the presence of a Touch Bar we'd have to make that interface visible on every Mac, even ones without Touch Bars.

DavidKarlas:
We can always limit how many commands user can add to not crash TouchBar :)
But in general, assume user will remove commands user doesn't need to make TouchBar useful...

About showing it in Preferences... If there is really no way of detecting if MacBook has TouchBar or not, always displaying it in Preferences is small price to pay...

michaelsavich:
It's not that adding to many items would cause it to crash, it's just that the UI would require a lot of scrolling. If you haven't seen the touch bar customization menu, it looks like this: http://pgj.cc/OnWjau

That being said, what you're suggesting isn't impossible. Basically we could have a menu in preferences that lets you set which items are available for customization, then the user can invoke Apple's UI and add the items they want.  It'd be a bit clunky, and I'm not certain it's the best solution…

Side note, I assume the lack of an API to check for TouchBar is because Apple is entertaining the idea of having Bluetooth keyboards with bars on them, and they don't want apps checking for a touch bar on launch, changing the UI & state, and then nothing works if the user then connects a keyboard with a touch bar.

DavidKarlas:
That editor looks nice, but I'm not sure if it can handle ~100 commands that we have... We probably also want to give user ability to add commands that are introduced by AddIns... So, we need to give option to add any command in case command doesn't have Icon for touchbar... display Text(name of command)

For same reason, we should probably always show Preferences editor for TouchBar, that is not real problem...

michaelsavich:
I didn't design the customization interface, I have no control over its appearance, that's on Apple. Like I said, we can make our own UI, but that would involve creating a drag and drop interface so that the user can visualize things like spacing. And that would require reimplementing a LOT of stuff. And users would hate the weirdness of the UI. I just don't think supporting every command is reasonable.

TomSoderling:
Have you considered making this into a VS for Mac Extension? https://docs.microsoft.com/en-us/visualstudio/mac/extending-visual-studio-mac

michaelsavich:
Sorry it took me so long to respond. I don't think an extension would work here. There aren't any extension points for the Touchbar, and from my experience with AppKit, I seriously doubt there is a sane way for extensions to add Touchbar support.

Dids:
Any updates on this?

PR #1802, opened 2017-01-23
[CommandManager] Throttle undo/redo on Mac

bratsche:
Not an ideal solution, but improves user experience noticeably.
This lets the text editor update and render before more undo
commands are fired.

https://bugzilla.xamarin.com/show_bug.cgi?id=34828

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -428,6 +428,11 @@ namespace MonoDevelop.Components.Commands
 			}
 		}
 
+#if MAC
+		const uint THROTTLE_TIME_SPREAD = 75;
+		static uint throttleLastEventTime = 0;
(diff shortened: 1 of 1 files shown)


DavidKarlas:
Maybe we should just remove calling `MDMenu.FlashMenu()` for Undo and Redo commands("Main Menu->Edit->Undo/Redo") because Mac is doing throttling by putting **UIThread into sleep** for example see flame graph below of CPU profiling... It's ~80% of time sleeping when holding Cmd+Z instead of rendering text editor content, hence terrible UX...

image

bratsche:
@DavidKarlas That makes sense. I just added another commit which solves the problem you described by resetting the throttle time when the key is released. So if you hit Cmd-Z twice really fast then it won't be throttled.

slluis:
Can we handle this is the Undo and Redo command handlers instead of the command manager?

bratsche:
@slluis I opened another PR that I think is better than this one, so closing this one. The new one is at https://github.com/mono/monodevelop/pull/1837

PR #1803, opened 2017-01-24
Always set CommandInfo.DataItem + parametrized connected services command

sevoku:
Usually a command is dispatched from a menu item in the context of the current active IDE element (editor, curretly selected project, etc.). Only array commands (recent files, windows, etc.) have an additional `DataItem` which gets passed to the command handler.

An alternative method to dispatch commands is `CommandService.DispatchCommand` which accepts an optional `dataItem` object as a parameter to any command, but it won't be received by the handler, because we assign `CommandInfo.DataItem` only for array commands.

With a915982 `CommandInfo.DataItem` will be always set, allowing all command handlers to receive parameters from the command service.

The first command making use of the `dataItem` parameter is the connected services gallery, which can be opened in the context of a specific project and receive a special service id, to open a  service page directly.

### DIFF
--- a/main/src/addins/MonoDevelop.ConnectedServices/CommandHandlers/OpenServicesGalleryCommandHandler.cs
+++ b/main/src/addins/MonoDevelop.ConnectedServices/CommandHandlers/OpenServicesGalleryCommandHandler.cs
@@ -14,19 +14,21 @@ namespace MonoDevelop.ConnectedServices.CommandHandlers
 		protected override void Update (CommandInfo info)
 		{
 			base.Update (info);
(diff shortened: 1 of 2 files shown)


slluis:
This is abusing the command system. If the welcome page needs to execute an action on the connected services add-in, the add-in should provide an API for that.

sevoku:
Oh, ok. I thought that this is the way how its meant to be used, since its all public API (the dataItem property may be slightly confusing here). The ConnectedServices Addin doesn't have any public API to control its behaviour at all, lets fix it there then.

PR #1806, opened 2017-01-24
[Ide] Ensure document actions are run on UI thread.

mrward:
Without the main UI thread's task scheduler being passed to the
ContinueWith the action was run on a thread pool thread instead of
the UI thread. This was causing the FileService.FileChanged event to
fire on a non-UI thread. It was also causing a crash if the project
file is edited in the text editor since the reload was being done on
a non-UI thread.


alanmcgovern:
No more crashing! 👍

slluis:
Thanks @mrward!

PR #1809, opened 2017-01-25
Toolbar requests

Therzok:
Implemented abstraction and mac backend for the toolbar selector requests.

### DIFF
--- a/main/src/addins/MacPlatform/MainToolbar/SelectorView.cs
+++ b/main/src/addins/MacPlatform/MainToolbar/SelectorView.cs
@@ -262,6 +262,9 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 						Enabled = mutableModel.Enabled,
 						Hidden = !mutableModel.Visible,
 					};
+					if (!string.IsNullOrEmpty (runtime.Image)) {
+						menuItem.Image = ImageService.GetIcon (runtime.Image).ToNSImage ();
+					}
 					if (ActiveRuntime == runtime || (ActiveRuntime?.Children.Contains (runtime) ?? false)) {
 						menuItem.State = NSCellStateValue.On;
 					}
--- a/main/src/addins/MacPlatform/MainToolbar/SelectorView.cs
+++ b/main/src/addins/MacPlatform/MainToolbar/SelectorView.cs
@@ -303,6 +303,10 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 			NSImage projectImageDisabled = MultiResImage.CreateMultiResImage ("project", "disabled");
 			NSImage deviceImage = MultiResImage.CreateMultiResImage ("device", "");
 			NSImage deviceImageDisabled = MultiResImage.CreateMultiResImage ("device", "disabled");
+
+			string lastDeviceIconId;
+			NSImage lastDeviceImage;
+			NSImage lastDeviceImageDisabled;
 			public PathSelectorView (CGRect frameRect) : base (frameRect)
 			{
 				Cells = new [] {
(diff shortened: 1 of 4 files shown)


Therzok:
Will implement Windows/Linux when I fix my PC.

slluis:
Let's wait a bit for this. The device selector is being redesigned.

Therzok:
Sure, just wanted the backend in place.

VincentDondain:
@Therzok please have the toolbar use the selected `ExecutionTarget`'s icon and not always the default iPhone-like icon (:

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/ExecutionTarget.cs
```diff
@@ -72,6 +72,28 @@ protected ExecutionTarget ()
...
 
 		/// <summary>
+		/// Gets or sets a value indicating whether this <see cref="T:MonoDevelop.Core.Execution.ExecutionTarget"/> is greyed.
+		/// </summary>
+		/// <value><c>true</c> if greyed; otherwise, <c>false</c>.</value>
+		/// <remarks>
+		/// This is introduced to be able to gray out execution targets for whatever reason makes sense for the project. 
+		/// For example, the ios add-in uses this to indicate which simulators can't be interacted with.
+		/// </remarks>
+		public bool Greyed { get; set; }
```
slluis: I would prefer this to be `Enabled { get; set; } = true`. "Greyed" is too specific to a visual representation.

Therzok: There is already `Enabled` which hides the execution target. We can rename the APIs in vNext.

Therzok: Will switch this to Enabled and remove the prop alltogether /cc @VincentDondain .

Enabled=false is used in android/ios addins as informational items - i.e. "Connect a provisioned Apple device via USB". The approach of hiding items via Enabled is wrong.

VincentDondain:
@Therzok you'd also need to provide a way to set right hand side labels I guess. See comment: https://github.com/xamarin/icons/issues/199#issuecomment-277233362

Therzok:
@VincentDondain that's a bit more complicated to do, it would require custom NSViews inside NSMenuItem (otherwise, we could just format the text to have spacing) and an NSAttributedString as text.

Problem is, how much space should there be between the left side and right side of the label?

VincentDondain:
Yea we should go for the custom NSView option.

That's not the most important thing to ship this PR though, addressing https://github.com/mono/monodevelop/pull/1809#issuecomment-276517894 comes first.

Therzok:
It's not possible right now to get the icon in the selector. The selector doesn't like ImageService icons, so we need to create a MultiResImage from an assembly resource.

sevoku:
@Therzok AFAIK Cocoa needs template icons in menus and additionally it doesn't support custom drawing. Since the `ImageService` icons are custom drawn  (`ThemedImage`), they need to be converted to bitmaps and have `Template` enabled.
This is what I did to support ImageService icons in our App menu: https://github.com/mono/monodevelop/blob/4338ed99191778611e605cfa192ce70a02b7ce5b/main/src/addins/MacPlatform/MacMenu/MDMenuItem.cs#L199-L206
We do the same thing in many other places where Cocoa and ImageService are involved. Maybe it would help here too.

Therzok:
@VincentDondain implemented

main/src/addins/MacPlatform/MainToolbar/SelectorView.cs
```diff
@@ -501,21 +511,50 @@ void UpdatePathText (int idx, string text)
...
+					return enabled ? deviceImage : deviceImageDisabled;
+				if (ActiveRuntime.Image == lastDeviceIconId)
+					return enabled ? lastDeviceImage : lastDeviceImageDisabled;
+
+				lastDeviceIconId = ActiveRuntime.Image;
+				var scale = GtkWorkarounds.GetScaleFactor (Ide.IdeApp.Workbench.RootWindow);
+				Xwt.Drawing.Image baseIcon = ImageService.GetIcon (ActiveRuntime.Image, Gtk.IconSize.Menu);
+
+				string [] styles, disabledStyles;
+				if (NSUserDefaults.StandardUserDefaults.StringForKey ("AppleInterfaceStyle") == "Dark") {
```
slluis: Why this check? ImageService should already be returning the icons with the correct styles.

Therzok: @sevoku would konw better, this code is borrowed from https://github.com/mono/monodevelop/blob/4338ed99191778611e605cfa192ce70a02b7ce5b/main/src/addins/MacPlatform/MacMenu/MDMenuItem.cs#L194-L206

sevoku: I'm not sure whether its needed here. In the AppMenu it is required because it doesn't follow the Ide theme setting (its theme is defined by OSX system settings). In this case the menu should follow the NSAppearance of its parent and there should be no need to modify the "dark" style setting. But this needs to be verified.

Therzok: On it to check.

Therzok: Seems to be needed, otherwise I get the wrong themed icon in the selectorview.

PR #1810, opened 2017-01-25
Solution file for Win32 debugger to build it independently

nerzhulart:


DavidKarlas:
Can you move Win32 debugger to debugger-libs repo and include into that solution? Would that work for you?

nerzhulart:
@DavidKarlas It would be very nice (and to move dotnetcore debugger libs as well). But it needs some amount of time to move the sources preserving history. I'm going to do this, but a bit later.

DavidKarlas:
OK

PR #1811, opened 2017-01-25
OutlineTreeView was null and causing a NRE

decriptor:
screen shot 2017-01-25 at 10 44 42 am

/cc @mkrueger


main/src/addins/CSharpBinding/MonoDevelop.CSharp.ClassOutline/CSharpOutlineTextEditorExtension.cs
```diff
@@ -301,7 +301,7 @@ bool RefillOutlineStore ()
 			Runtime.AssertMainThread ();
 			Gdk.Threads.Enter ();
 			refreshingOutline = false;
-			if (outlineTreeStore == null || !outlineTreeView.IsRealized) {
+			if (outlineTreeStore == null || !outlineTreeView?.IsRealized == true) {
```
Therzok: Why not make it clearer rather that you're checking both null and realized, rather than having to explicitly compare with true?

PR #1814, opened 2017-01-26
Misc perf

Therzok:


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/MonoExecutionParameters.cs
```diff
@@ -262,8 +264,10 @@ public string GenerateDescription ()
 					ops.Append (", ");
 				var nameAttr = localizedDisplayNameAttributes [prop];
 				ops.Append (nameAttr.DisplayName);
-				if (!(pval is bool))
-					ops.Append (": " + GetValue (pval));
+				if (!(pval is bool)) {
+					ops.Append (": ");
+					ops.Append (GetValue (pval));
```
slluis: Being picking, `ops.Append (": ").Append (GetValue (pval))` would have been more convenient

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.PropertyGrid/EditorManager.cs
```diff
@@ -167,12 +167,9 @@ public Type GetEditorType (ITypeDescriptorContext context)
...
 		{
-			foreach (PropertyInfo member in colType.GetProperties ()) {
-				if (member.Name == "Item") {
-					if (member.PropertyType != typeof (object))
-						return member.PropertyType;
-				}
-			}
+			var member = colType.GetProperty ("Item");
+			if (member != null && member.PropertyType != typeof (object))
+				return member.PropertyType;
```
slluis: This will break if the class has more than one indexer.

Therzok: You can't have two properties with the same name though.

Therzok: Using the `Item` name check is unreliable. We should either check whether the property has `GetIndexParameters().Length > 0` or we could instead iterate `type.GetDefaultMembers ()` because indexers get placed there.

That way we iterate fewer properties (default ones and those marked with `DefaultMemberAttribute`)

main/tests/UnitTests/MonoDevelop.Components.PropertyGrid/EditorManagerTests.cs
```diff
@@ -0,0 +1,61 @@
...
+
+			public class TypeHasPrivateProperty
+			{
+				TypeHasPrivateProperty Item { get; set; }
+			}
+
+			public class TypeHasNoProperty
+			{
+				public TypeHasNoProperty Item;
+			}
```
slluis: This should be testing indexers, not properties named "Item".

Therzok:
Fixed all comments.

main/tests/UnitTests/MonoDevelop.Components.PropertyGrid/EditorManagerTests.cs
Therzok: See TypeHasIndexer

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.PropertyGrid/EditorManager.cs
```diff
@@ -167,11 +167,10 @@ public Type GetEditorType (ITypeDescriptorContext context)
 		
 		public static Type GetCollectionItemType (Type colType)
 		{
-			foreach (PropertyInfo member in colType.GetProperties ()) {
-				if (member.Name == "Item") {
-					if (member.PropertyType != typeof (object))
-						return member.PropertyType;
-				}
+			foreach (var member in colType.GetDefaultMembers ()) {
```
Therzok: This greatly reduces the number of `MemberInfo`s being returned for probing.

PR #1815, opened 2017-01-26
[DotNetCore] Fix OwnerProject null for project ref after reevaluation

mrward:
When a new project reference is added and then the project is
reevaluated the project reference is replaced with a new instance
but the OwnerProject is not set. This caused code completion to stop
working in the text editor and types were marked with red error
markers.

A new ProjectReference has a BackingEvalItem which has no properties
so on reevaluating a new ProjectReference is created since no match
is found due to the missing properties. On calling Items.SetItem
there is no items added notification since ProjectReference implements
GetHashCode and Equals and the different instances are considered to
be the same. This resulted in a null OwnerProject for the
ProjectReference and when that was written out to xml a null reference
exception was thrown. To fix this the OwnerProject is set after
reevaluation if it is null.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ItemCollection.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ItemCollection.cs
@@ -76,6 +76,7 @@ namespace MonoDevelop.Projects
 		public void SetItems (IEnumerable<T> items)
 		{
 			AssertCanWrite ();
+			items = ReuseExistingItems (items);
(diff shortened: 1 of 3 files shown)


slluis:
I think that the bug is actually in ItemCollection.SetItems. If an item being added is equal to one already present in the list, SetItems should keep the instance that's already in the list, instead of replacing it by the new instance.

mrward:
I have changed ItemCollection.SetItems to keep the original instance instead of replacing it if they are considered equal.

PR #1821, opened 2017-01-31
Make Connected Services API public

sevoku:
This change makes the following features public:
* Open the services gallery for a given project
* Open the service details view for a given service
* Query services supporting a given project
* Remove a service from a project (asking user for confirmation)

Would it make sense to open more features? Like adding a service to a project and show the confirmation dialog (allowing to add a service to all referencing projects) maybe?

cc @sgmunn

### DIFF
--- a/main/src/addins/MonoDevelop.ConnectedServices/ConnectedServices.cs
+++ b/main/src/addins/MonoDevelop.ConnectedServices/ConnectedServices.cs
@@ -88,23 +88,23 @@ namespace MonoDevelop.ConnectedServices
 		/// <summary>
 		/// Displays the service details tab for the given service
 		/// </summary>
-		public static void OpenServicesTab (this IConnectedService service)
+		public static Task OpenServicesTab (this IConnectedService service)
 		{
 			if (service == null)
 				throw new ArgumentNullException (nameof (service));
 
-			Runtime.RunInMainThread (() => OpenServicesTab (service.Project, service.Id));
+			return Runtime.RunInMainThread (() => OpenServicesTab (service.Project, service.Id));
 		}
 
 		/// <summary>
 		/// Displays the services gallery tab for the given project
 		/// </summary>
-		public static void OpenServicesTab (this DotNetProject project)
+		public static Task OpenServicesTab (this DotNetProject project)
 		{
 			if (project == null)
 				throw new ArgumentNullException (nameof (project));
 
-			Runtime.RunInMainThread (() => OpenServicesTab (project, null));
+			return Runtime.RunInMainThread (() => OpenServicesTab (project, null));
 		}
 
 		/// <summary>
--- a/main/src/addins/MonoDevelop.ConnectedServices/ConnectedServices.cs
+++ b/main/src/addins/MonoDevelop.ConnectedServices/ConnectedServices.cs
@@ -15,7 +15,7 @@ namespace MonoDevelop.ConnectedServices
 	/// <summary>
 	/// Defines a set of constants for the Connected Services addin
 	/// </summary>
-	static class ConnectedServices
+	public static class ConnectedServices
 	{
 		/// <summary>
 		/// The extension point for service providers
@@ -64,8 +64,11 @@ namespace MonoDevelop.ConnectedServices
 		/// <summary>
 		/// Displays the service details tab for the given service in the given project
 		/// </summary>
-		public static void OpenServicesTab(DotNetProject project, string serviceId = null)
+		internal static void OpenServicesTab(DotNetProject project, string serviceId)
 		{
+			if (project == null)
+				project = IdeApp.ProjectOperations.CurrentSelectedProject as DotNetProject;
+
 			ConnectedServicesViewContent servicesView = null;
 
 			foreach (var view in IdeApp.Workbench.Documents) {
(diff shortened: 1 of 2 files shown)


main/src/addins/MonoDevelop.ConnectedServices/Gui.ServicesTab/ConnectedServicesViewContent.cs
```diff
@@ -51,13 +51,16 @@ public ConnectedServicesViewContent (DotNetProject project)
...
 		{
 			var binding = ((DotNetProject)this.Project).GetConnectedServicesBinding ();
-			if (string.IsNullOrEmpty (serviceId)) {
-				var services = binding.SupportedServices;
-				this.widget.ShowGallery (services, Project);
-			} else {
+			if (!string.IsNullOrEmpty (serviceId)) {
 				var service = binding.SupportedServices.FirstOrDefault (x => x.Id == serviceId);
-				this.widget.ShowServiceDetails (service);
+				if (service != null) {
```
Therzok: We should log something on the null case, such as `service id: {0} not found`

main/src/addins/MonoDevelop.ConnectedServices/ConnectedServices.cs
```diff
@@ -83,6 +86,28 @@ public static void OpenServicesTab(DotNetProject project, string serviceId = nul
...
 
 		/// <summary>
+		/// Displays the service details tab for the given service
+		/// </summary>
+		public static void OpenServicesTab (this IConnectedService service)
+		{
+			if (service == null)
+				throw new ArgumentNullException (nameof (service));
+
+			Runtime.RunInMainThread (() => OpenServicesTab (service.Project, service.Id));
```
Therzok: What if the user wants to do something that has to wait for the UI to open?

main/src/addins/MonoDevelop.ConnectedServices/Gui.ServicesTab/ConnectedServicesViewContent.cs
sevoku: its not public and in the `null` case its supposed to show the gallery.

main/src/addins/MonoDevelop.ConnectedServices/ConnectedServices.cs
sevoku: yeah, it should be a Task

main/src/addins/MonoDevelop.ConnectedServices/Gui.ServicesTab/ConnectedServicesViewContent.cs
Therzok: Yeah, but this is not the serviceId being null case, this is the SupportedServices search not finding the requested serviceId.

Therzok: Even if it's `internal`, if someone hardcodes an ID, this will be a mess to track down.

sgmunn: yeah, logging that `service` was null would probably shorten tracking down why the gallery opened instead of a service page.

sevoku: yep, right

sevoku:
pushed requested changes, cc @Therzok @sgmunn

main/src/addins/MonoDevelop.ConnectedServices/ConnectedServices.cs
```diff
@@ -83,6 +86,28 @@ public static void OpenServicesTab(DotNetProject project, string serviceId = nul
...
 
 		/// <summary>
+		/// Displays the service details tab for the given service
+		/// </summary>
+		public static async Task OpenServicesTab (this IConnectedService service)
+		{
+			if (service == null)
+				throw new ArgumentNullException (nameof (service));
+
+			await Runtime.RunInMainThread (() => OpenServicesTab (service.Project, service.Id));
```
Therzok: Just return the Task, don't await.

Therzok: (also remove the async modifier)

main/src/addins/MonoDevelop.ConnectedServices/ConnectedServices.cs
```diff
@@ -83,6 +86,28 @@ public static void OpenServicesTab(DotNetProject project, string serviceId = nul
...
+
+		/// <summary>
+		/// Displays the services gallery tab for the given project
+		/// </summary>
+		public static async Task OpenServicesTab (this DotNetProject project)
+		{
+			if (project == null)
+				throw new ArgumentNullException (nameof (project));
+
+			await Runtime.RunInMainThread (() => OpenServicesTab (project, null));
```
Therzok: Same as above.

xamarin-release-manager:
Fixed in version 7.8.0.389 (master)

Author: Matt Ward
Commit: 8447fab59d218d1e94ed3856c418f7a0cb98121d (mono/monodevelop)

xamarin-release-manager:
Fixed in version 7.7.0.1280 (release-7.7)

Author: Matt Ward
Commit: 8447fab59d218d1e94ed3856c418f7a0cb98121d (mono/monodevelop)

PR #1823, opened 2017-01-31
Roslyn 2.0 + IVT

mhutch:
**NOTE: this requires Mono master as of 28/1 to build on Mac. It depends on MSBuild, Roslyn csc, and some recent changes to the Mono runtime.**

This works mostly, but has some regressions. Here's a list of some of the things we need to fix and some of the things we can do now we have IVT access.

- [ ] Integrate Roslyn completion providers to get full C# 7 support
- [ ] Bridge document specific options to formatting policies - see [TextPolicyDocumentOptionsProvider](https://github.com/mono/monodevelop/blob/roslyn-ivt/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TextPolicyDocumentOptionsProvider.cs) - and make sure they refresh correctly
- [ ] Bridge roslyn options to MD preferences - see [GlobalOptionPersister](https://github.com/mono/monodevelop/blob/roslyn-ivt/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/GlobalOptionPersister.cs) - and make sure they refresh correctly
- [ ] Implement IChangeSignatureOptionsService, IGenerateTypeOptionsService, IExtractInterfaceOptionsService to fix errors in code actions
- [ ] Integrate editorconfig support. Will need to figure out conflict with policies somehow.
- [ ] Integrate DiagnosticService, CodeFixMenuService and FixMultipleOccurrencesService
- [ ] Use the roslyn mechanism for loading diagnostics from MEF
- [ ] Port persistent storage service. It uses ESENT which only exists on Windows, need to port to Sqlite.
- [ ] Fix TemporaryStorageService. Had to disable it on Mac as it uses an mmap'ed file per file in the solution which completely blows past the hard file handle limit on Mac. May be a Mono bug, or perhaps we could come up with a different implementation.
- [ ] Porting CSharpEditorFormattingService / IEditorFormattingService. Looks like chunks of this could be shifted down from EditorFeatures to Features.
- [ ] Port symbol search service and set up Elfie background indexing.

Note also that we can't use Microsoft.CodeAnalysis.Workspaces.Desktop wholesale as many types fail to load due to Windows and MSBuild dependencies. We'll need to port its services, ideally to a separate "Microsoft.CodeAnalysis.Workspaces.XplatDesktop" project rather than just mashing them into MonoDevelop.Ide.


Therzok:
Assigning myself so I don't forget to look at the top list.

DavidKarlas:
notice that version-checks is pointing to roslyn-ivt branch...

PR #1831, opened 2017-02-01
Dev/dpugh/v next add platform

dpugh:


dpugh:
Adds Text.Data and Text.Logic plus exposes the platform MEF catalog.

mkrueger:
We need to discuss that in the call today.

PR #1833, opened 2017-02-02
Revert "Dev/dpugh/v next add platform"

mkrueger:
Reverts mono/monodevelop#1831


mkrueger:
Doesn't compile/some license infos are missing.

PR #1834, opened 2017-02-02
Init perf

Therzok:


main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/UpdatedNuGetPackagesProvider.cs
```diff
@@ -126,7 +129,12 @@ internal class UpdatedNuGetPackagesProvider
...
 			var package = packages
 				.Where (p => IsPackageVersionAllowed (p, packageReference))
-				.OrderByDescending (p => p.Identity.Version).FirstOrDefault ();
+				.Aggregate ((left, right) => {
+					if (left.Identity.Version >= right.Identity.Version)
+						return left;
+					return right;
+				});
+
 			if (package == null)
```
mrward: Just wondering if we could replace the above with an extension. So the code above would be something like:

`.GetHighestVersion ();`

Or if that is not possible something like:

`.Aggregate ((left, right) => HighestVersion (left, right));`

Or something else.

Therzok: I propose an IEnumerable extension in CoreExtensions, for Max/Min which take an IComparer instead of a selector.

Therzok:
@slluis @mrward fixes are up.

Therzok:
Uhhh, I need to do add MaxValueOrDefault versions. Gimme a bit more.

main/src/core/MonoDevelop.Core/CoreExtensions.cs
```diff
@@ -60,6 +60,110 @@ public static int IndexOf<T> (this IEnumerable<T> e, T item)
...
+					value = x;
+					result = item;
+					hasValue = true;
+				}
+			}
+			if (hasValue)
+				return result;
+			throw new InvalidOperationException (string.Format ("{0} contains no elements", nameof (source)));
+		}
+
```
mrward: There is very similar code in both of the MaxValue methods. Can one call the other with some default implementation of IComparer<TCompare> comparer which does the `x.CompareTo (value) > 0` comparison?

Similar comment for the MinValue methods.

Therzok: I don't think there is a way without introducing extra delegate overhead for the user. I'm not a fan of the duplicated code either, but System.Linq does the same.

PR #1838, opened 2017-02-03
Remove old CSS/HTML/JSON mappings. To be replaced by WebTooling Addin

ToddGrun:
I wanted to change my file extensions from .css2, .html2, .json2 to the real things, so I needed to update MD so that we wouldn't collide.

### DIFF
--- a/main/src/addins/AspNet/Properties/MonoDevelop.AspNet.addin.xml
+++ b/main/src/addins/AspNet/Properties/MonoDevelop.AspNet.addin.xml
@@ -164,9 +164,6 @@
 		            insertafter = "Xml"
 		           _label = "ASP.NET Files"
 		           extensions = "*.aspx,*.ashx,*.asmx,*.ascx,*.master,*.asax,*.cshtml" />
-		<FileFilter id = "Html"
-		            _label = "HTML Files"
-		            extensions = "*.htm,*.html"/>
 	</Extension>
 
 	<Extension path = "/MonoDevelop/ProjectModel/ProjectModelExtensions">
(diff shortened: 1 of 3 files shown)


ToddGrun:
@mkrueger @mhutch 

Thoughts?

mhutch:
Technically you can just insert yours before the old ones using the insertBefore attribute and they will be picked up instead when your addin is loaded.

See e.g. https://github.com/mhutch/MonoDevelop.MSBuildEditor/blob/master/Properties/Manifest.addin.xml which overrides the built in MSBuild editor.

mhutch:
That said, it only works if the thing you're inserting before has an ID...

ToddGrun:
Interesting. For the MimeType, can I specify that I want my to be insertedbefore both text/html and application/x-spark?

Also, would this work for FileFilters? I haven't tried it, but would my file filter replace theirs, or just precede it?

mkrueger:
We need to discuss it if we should remove html/json support for monodevelop. If we do it we've to remove more than just some bindings. Like the HtmlEditorExtension it would be unused in that patch.

Futuremore if we do it we should remove ASP.NET support as well.

mhutch:
@ToddGrun if you insert before html, it'll be before spark, since html is before spark.

As for ordering, it depends on how the extension point consumes the extension.

Some "list" extension points will just use all the extensions at the extension point - pretty sure file filters do this.

Other scan the extensions and find the first match - for example the DesktopService, when determining mimetype for a file, takes the first with a glob that that matches the file's extension.

Yet others will build a dictionary for perf reasons, and in this case it would effectively be the _last_ match (but there's also an `insertAfter`...).

PR #1840, opened 2017-02-06
Unit tests for Bug 44851

DavidKarlas:
This should be merged after https://github.com/mono/debugger-libs/pull/97 is merged and debugger-libs sumboulde bumped


DavidKarlas:
PR on debugger-libs + bump is done, hence this can be merged

PR #1853, opened 2017-02-09
Dev/dpugh/modify text document, Switch to dev 15 bits

dpugh:


main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.csproj
```diff
@@ -58,7 +58,7 @@
     <ProjectReference Include="..\MonoDevelop.DesignerSupport\MonoDevelop.DesignerSupport.csproj">
       <Project>{2C24D515-4A2C-445C-8419-C09231913CFA}</Project>
       <Name>MonoDevelop.DesignerSupport</Name>
-      <Private>False</Private>
+      <Private>True</Private>
```
slluis: This should not change. We don't want to local-copy other add-ins.

PR #1868, opened 2017-02-13
[Editor] Fix ImmutableText.WriteTo implementation.

Therzok:
The code would infinitely loop until it went out of bounds.

### DIFF
--- a/main/src/core/Mono.Texteditor/Mono.TextEditor.Utils/ImmutableText.cs
+++ b/main/src/core/Mono.Texteditor/Mono.TextEditor.Utils/ImmutableText.cs
@@ -293,10 +293,11 @@ namespace Mono.TextEditor.Utils
 			return new string (data);
 		}
 
-		public void WriteTo (TextWriter output, int index, int count)
+		public void WriteTo (TextWriter output, int start, int count)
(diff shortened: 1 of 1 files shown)


main/src/core/Mono.Texteditor/Mono.TextEditor.Utils/ImmutableText.cs
```diff
@@ -293,10 +293,11 @@ public string ToString (int offset, int length)
 			return new string (data);
 		}
 
-		public void WriteTo (TextWriter output, int index, int count)
+		public void WriteTo (TextWriter output, int start, int count)
 		{
-			while (index < index + count) {
-				output.Write (this [index]);
```
mkrueger: an index++ would've been enough - but y that method was never tested/used before :)

Therzok: `index < index + count` would never be false even if we did `index++`

Therzok: And we do below anyway.

PR #1869, opened 2017-02-13
[Core] Allow for msbuild builder reuse

Therzok:
For some reason, I'm still getting process disconnected exceptions, and I have no idea why.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteProjectBuilder.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteProjectBuilder.cs
@@ -50,7 +50,7 @@ namespace MonoDevelop.Projects.MSBuild
 
 		public int ReferenceCount { get; set; }
 		public DateTime ReleaseTime { get; set; }
-		Queue<TaskCompletionSource<object>> readyNotify = new Queue<TaskCompletionSource<object>> ();
+		Queue<TaskCompletionSource<bool>> readyNotify = new Queue<TaskCompletionSource<bool>> ();
 
 		public RemoteBuildEngine (RemoteProcessConnection connection)
 		{
@@ -142,6 +142,10 @@ namespace MonoDevelop.Projects.MSBuild
 			Interlocked.Decrement (ref count);
 			try {
 				alive = false;
+				lock (readyNotify) {
+					foreach (var tcs in readyNotify)
+						tcs.SetResult (false);
+				}
 				connection.Disconnect ().Ignore ();
 			} catch {
 				// Ignore
@@ -199,17 +203,17 @@ namespace MonoDevelop.Projects.MSBuild
 						return;
 
 					var tcs = readyNotify.Dequeue ();
-					tcs.SetResult (default (object));
+					tcs.SetResult (true);
 				}
 			}
 		}
 
-		public Task WaitForReady ()
+		public Task<bool> WaitForReady ()
 		{
 			if (Interlocked.CompareExchange (ref busy, busy, 0) == 0)
-				return Task.FromResult (default (object));
+				return Task.FromResult (true);
 			
-			var tcs = new TaskCompletionSource<object> ();
+			var tcs = new TaskCompletionSource<bool> ();
 			lock (readyNotify)
 				readyNotify.Enqueue (tcs);
 			return tcs.Task;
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteProjectBuilder.cs
```diff
@@ -141,6 +142,10 @@ public void Dispose ()
 			Interlocked.Decrement (ref count);
 			try {
 				alive = false;
+				lock (readyNotify) {
```
Therzok: Not sure if this is needed, the builder shouldn't shutdown until all refs are released.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -1105,21 +1105,27 @@ internal protected virtual Task<TargetEvaluationResult> OnRunTarget (ProgressMon
...
 
-					RemoteProjectBuilder builder = await GetProjectBuilder ();
+					RemoteProjectBuilder builder = await GetProjectBuilder (requestLocked).ConfigureAwait (false);
 					if (builder.IsBusy) {
-						builder.ReleaseReference ();
-						newBuilderRequested = true;
-						builder = await RequestLockedBuilder ();
+						if (requestLocked) {
+							await builder.WaitForReady ();
+							builder.Lock ();
```
alanmcgovern: we passed the `requestLocked` parameter to `GetProjectBuilder` so it should ensure that the builder is already correctly locked (if that's what was requested).

This method is also confusing because if `requestLocked` is false then we hit the else, which forces us to acquire a locked builder. Is that a mistake>

alanmcgovern: I wonder if we could design the API to be a little simpler. Could we hide `Lock`, `Unlock`, `WaitForReady` and replace it with something like:
```
public async Task<RemoteProjectBuilder> AcquireBuilder (bool locked);
public void ReleaseBuilder (RemoteProjectBuilder builder);
```
This would ensure that we either create a new dedicated builder for this command, or we re-use an existing shared builder. Once our command completes we'd release the builder, which would either let the remote process die (if it was a dedicated builder and is not needed) or it would keep the builder in the pool for re-use later if it was a shared builder.

Therzok: > This method is also confusing because if requestLocked is false then we hit the else, which forces us to acquire a locked builder. Is that a mistake>

I inverted the if/else by mistake.

Therzok: But yes, the name is confusing.

slluis:
I don't understand what this change is trying to do, and which problem is trying to solve.

Therzok:
Waiting until a builder is no longer busy so the same can be reused. I didn't understand the Lock/Unlock semantics of a builder (as it usually ended up with the builder saying that the remote process connection was disconnected) if we just ignored IsBusy and tried to lock.

slluis:
The goal of Lock/Unlock is to allow reserving a builder for a build operation, so that if another build operation needs to start we don't want it to wait until the current one has finished (we want to build in parallel). Lock/Unlock is not really necessary when running fast operations, such as when resolving assembly references. The builder will just serialize concurrent calls. So the logic could be like this:

* For fast operations:
    1. A builder is requested.
    2. If the builder is not busy, we use it to run the operation (without Lock/Unlock)
    3. If the builder is busy, we request a new builder and use it (without Lock/Unlock)
* For long operations:
    1. A builder is requested
    2. If the builder is not busy, we Lock it and run the operation.
    3. If the builder is busy, we request a new locked builder.

alanmcgovern:
What happens if a fast operation is currently in progress, then i (more or less) simultaneously queue up a fast operation, then a slow operation. The fast operation will see that the builder is not locked so it will grab that one and queue. The slow operation will see the same builder is not locked and will grab it, lock it and queue it's operation. Are we guaranteed the fast operation will execute first? Is it a FIFO queue?

If we then queued up a third fast operation it would see the builder is busy and would spin up a new one, which is as expected.

slluis:
> Are we guaranteed the fast operation will execute first? Is it a FIFO queue?

Yes, if the fast operation is invoked first, it will be executed first.

Therzok:
Superseded by https://github.com/mono/monodevelop/pull/1871

PR #1870, opened 2017-02-13
Async command update

slluis:
Adds support for async command update handlers. If an update handler returns a task, changes done in the provided CommandUpdaterInfo will be shown in the corresponding menu item.

```
[CommandUpdateHandler (ViewCommands.OpenWithList)]
public async Task OnOpenWithUpdate (CommandArrayInfo info, CancellationToken ct)
{
    info.Add (new CommandInfo ("Loading...") { Enabled = false }, null);
    var pf = (ProjectFile)CurrentNode.DataItem;
    // Some potentially slow operation
    var viewers = await DisplayBindingService.GetFileViewers (pf.FilePath, pf.Project, ct);
    info.Clear ();
    PopulateOpenWithViewers (info, viewers);
}
```

The PR also removes menu creation infrastructure from the Mac add-in, since we now have the same code in MD.Ide.


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandInfo.cs
```diff
@@ -77,47 +82,47 @@ public CommandInfo (string text, bool enabled, bool checkd)
 		
 		public string Text {
 			get { return text; }
-			set { text = value; }
+			set { text = value; NotifyChanged (); }
```
Therzok: Wouldn't we benefit more from IPropertyChanged here? We'd also have the same simplified code if we used `[CallerMemberName]` with the added benefit of being able to filter which value changed.

slluis: I don't think it is worth the additional complexity.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -486,6 +488,8 @@ internal bool ProcessKeyEvent (Gdk.EventKey ev)
 
 			for (int i = 0; i < commands.Count; i++) {
 				CommandInfo cinfo = GetCommandInfo (commands[i].Id, new CommandTargetRoute ());
+				if (cinfo.IsUpdatingAsynchronously)
+					cinfo.UpdateTask.Wait (); // Not nice, but we need a synchronous result here
```
Therzok: Maybe it's time to look into async-ifying keypresses and handling keypresses in order?

slluis:
Replaced by #1904

PR #1875, opened 2017-02-15
[AssemblyBrowser] Fix scrolling to type when doing Go To Definition

Therzok:
Before, the code was racy, as it was all attaching a ContinueWith on
LoadingTask, which caused the following semantics:
* On ViewContent loading, it would attach a task to scroll to the loaded
assembly.
* On actual Go To Definition call, it would attach a task that would
scroll to the real type we're searching for.

The order in which these two happen is inverse to the order of
attaching, we would scroll back to the first item, even though we did
all the node expansion and selection work for the real item.

The fix is to only scroll to the assembly if the current selected node
is null or equal to the root node.

### DIFF
--- a/main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserWidget.cs
+++ b/main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserWidget.cs
@@ -617,7 +617,8 @@ namespace MonoDevelop.AssemblyBrowser
 				return true;
 			return false;
 		}
-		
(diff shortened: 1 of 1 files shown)


slluis:
This won't work if an assembly has already been opened in the assembly browser (so something is already selected) and then a new assembly is opened. The new one won't be focused.

Therzok:
Nope, it does. The Load path is only called when loading a assembly widget document the first time. The path to open to a type is via Go To Definition.

Therzok:
Initially, I tried working around it by doing a CancellationTokenSource via the ViewContent, but it does not happen to run in the order I wanted.

Therzok:
Updated!

PR #1876, opened 2017-02-15
Assembly browser allocs

Therzok:


main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyLoader.cs
```diff
@@ -180,7 +180,7 @@ public string LookupAssembly (string fullAssemblyName)
 				return exe;
 
 			foreach (var asm in Runtime.SystemAssemblyService.DefaultAssemblyContext.GetAssemblies ()) {
-				if (asm.Name.ToLowerInvariant () == fullAssemblyName.ToLowerInvariant ())
+				if (string.Equals (asm.Name, fullAssemblyName, StringComparison.InvariantCultureIgnoreCase))
```
alanmcgovern: I think we want to use Ordinal here, not Invariant. Both sets of code are wrong.

Therzok: I have no qualms about doing that as a follow-up commit. Didn't want to change the behavior

alanmcgovern: 👍

PR #1880, opened 2017-02-16
[Templating] Initial support for https://github.com/dotnet/templating format and DotNetCore templates port

DavidKarlas:
This is just to get sense of things...
Changes in DotNetCore project:

 * Templates/ folder, replaced existing templates with templates from https://github.com/dotnet/templating/tree/master/template_feed
   * Only change I did is to add "category": inside Tags like this: [Link](https://github.com/mono/monodevelop/blob/new_templating/main/src/addins/MonoDevelop.DotNetCore/Templates/Projects/WebApi/CSharp/.template.config/template.json#L11)
   * This way we can add any metadata specific to MD to template, like IconId...
 * .csproj is just tracking changes in Templates/ folder
 * MonoDevelop.DotNetCore.addin.xml instead pointing at each project template, it points at folder(or subfolders) that contain .template.config/template.json file...

Changes in MonoDevelop.Ide:
 
 * .csproj and .packages have new NuGets and files, unfortunately this NuGets depend on Json.net 9.x+, we will probably have to ask templating guys to lower their requirement :(
 * Templates.addin.xml and TemplatesPathsCodon.cs add new extension point to register search paths for templates
 * MicrosoftTemplatingEngine*.cs files implement IProjectTemplatingProvider and call dotnet/templating APIs to generate projects...
 * Rest of files is just got some `Task` love

What is missing here are 2 things, setting parameters(e.g. generate .git) and better support for generating or not generating .sln file.
Atm I use XS logic to generate .sln file, which is not ideal... I will need to dig a bit into templating engine to figure out how to exlcude .sln file from template and move folders around based on that setting...


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.csproj
```diff
@@ -149,59 +163,623 @@
...
     </None>
-    <None Include="Templates\Projects\XUnitTest\XUnitTestProject.xpt.xml">
+    <None Include="Templates\Projects\WebApi\CSharp\Startup.cs">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
-    <None Include="Templates\Projects\XUnitTest\XUnitTestProject.Test.cs">
+    <None Include="Templates\Projects\WebApi\CSharp\Controllers\ValuesController.cs">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
   </ItemGroup>
```
mrward: Just wondering if there is a better way than adding every file from the template into the project. Can we submodule the dotnet/templating repository? Or maybe just have a post build task that copies files from a templates directory into the output folder?

We are also adding binary database files into the repo such as the .db. Not sure how big that file is.

main/src/addins/MonoDevelop.DotNetCore/Properties/MonoDevelop.DotNetCore.addin.xml
```diff
@@ -18,19 +18,10 @@
...
-			id="DotNetCoreEmptyWebProject"
-			file="Templates/Projects/EmptyWeb/EmptyWebProject.xpt.xml" />
-		<ProjectTemplate
-			id="DotNetCoreEmptyWebProject"
-			file="Templates/Projects/XUnitTest/XUnitTestProject.xpt.xml" />
+	<Extension path="/MonoDevelop/Ide/TemplatesPaths">
+		<TemplatesPath
+			id="DotNetCore"
+			path="Templates" />
 	</Extension>
```
mrward: Not sure about the extension name here. Can it just be "/MonoDevelop/Ide/Templates"?

Also should the extension point not indicate it is a path? So something like:

    <Template id="DotNetCore" path="Templates" />

How do templates indicate icons? I was assuming you would need to define them in the extension point.

main/src/addins/MonoDevelop.DotNetCore/Templates/Projects/StarterWeb/FSharp/Company.WebApplication1.fsproj
```diff
@@ -0,0 +1,39 @@
+<Project Sdk="FSharp.NET.Sdk;Microsoft.NET.Sdk.Web">
```
mrward: Is the F# project grouped with the C# project in the new project dialog?

OK it looks like we are using the GroupIdentity in the project templates to group together the same template for different languages. 👍

main/src/core/MonoDevelop.Ide/packages.config
```diff
@@ -1,6 +1,13 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Newtonsoft.Json" version="8.0.3" targetFramework="net45" />
+  <package id="Microsoft.TemplateEngine.Abstractions" version="1.0.0-beta1-20170208-116" targetFramework="net45" />
+  <package id="Microsoft.TemplateEngine.Core" version="1.0.0-beta1-20170208-116" targetFramework="net45" />
+  <package id="Microsoft.TemplateEngine.Core.Contracts" version="1.0.0-beta1-20170208-116" targetFramework="net45" />
+  <package id="Microsoft.TemplateEngine.Edge" version="1.0.0-beta1-20170208-116" targetFramework="net45" />
+  <package id="Microsoft.TemplateEngine.Orchestrator.RunnableProjects" version="1.0.0-beta1-20170208-116" targetFramework="net45" />
+  <package id="Microsoft.TemplateEngine.Utils" version="1.0.0-beta1-20170208-116" targetFramework="net45" />
+  <package id="Newtonsoft.Json" version="9.0.1" targetFramework="net45" />
```
mrward: Not sure we can update to Json.NET 9.0.1

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/TemplatingService.cs
```diff
@@ -137,11 +138,11 @@ internal SolutionTemplate GetTemplate (string templateId)
 			return null;
 		}
 
-		public ProcessedTemplateResult ProcessTemplate (SolutionTemplate template, NewProjectConfiguration config, SolutionFolder parentFolder)
+		public async Task<ProcessedTemplateResult> ProcessTemplate (SolutionTemplate template, NewProjectConfiguration config, SolutionFolder parentFolder)
```
mrward: 👍

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngineProjectTemplatingProvider.cs
```diff
@@ -0,0 +1,197 @@
...
+using System.IO;
+using Microsoft.TemplateEngine.Orchestrator.RunnableProjects.Config;
+using Microsoft.TemplateEngine.Orchestrator.RunnableProjects.Macros;
+using System.Threading.Tasks;
+using Mono.Addins;
+using MonoDevelop.Ide.Codons;
+
+namespace MonoDevelop.Ide.Templates
+{
+	public class MicrosoftTemplateEngineProjectTemplatingProvider : IProjectTemplatingProvider
```
mrward: Does not need to be public.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngineSolutionTemplate.cs
```diff
@@ -0,0 +1,74 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+using Microsoft.TemplateEngine.Abstractions;
+using Microsoft.TemplateEngine.Edge.Settings;
+using System.Linq;
+
+namespace MonoDevelop.Ide.Templates
+{
+	public class MicrosoftTemplateEngineSolutionTemplate : SolutionTemplate
```
mrward: Does not need to be public.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngineProjectTemplatingProvider.cs
```diff
@@ -0,0 +1,197 @@
...
+				config.GetValidProjectName (),
+				config.ProjectLocation,
+				new Dictionary<string, string> (),
+				true);
+			if (result.ResultInfo.PrimaryOutputs.Any ()) {
+				foreach (var res in result.ResultInfo.PrimaryOutputs)
+					workspaceItems.Add (await MonoDevelop.Projects.Services.ProjectService.ReadSolutionItem (new Core.ProgressMonitor (), res.Path));
+			} else {
+				foreach (var path in Directory.GetFiles (config.SolutionLocation, "*.*proj", SearchOption.AllDirectories)) {
+					if (path.EndsWith (".csproj", StringComparison.OrdinalIgnoreCase) || path.EndsWith (".fsproj", StringComparison.OrdinalIgnoreCase) || path.EndsWith (".vbproj", StringComparison.OrdinalIgnoreCase))
```
mrward: Can the file extension be defined somewhere say in the template itself? Or inferred from the language? Be nice to avoid hard coding the file extension here.

main/src/addins/MonoDevelop.DotNetCore/Templates/Projects/Console/CSharp/.template.config/template.json
```diff
@@ -0,0 +1,38 @@
...
+  "author": "Microsoft",
+  "classifications": ["Common", "Console"],
+  "name": "Console Application",
+  "groupIdentity": "Microsoft.Common.Console",
+  "identity": "Microsoft.Common.Console.CSharp",
+  "shortName": "console",
+  "tags": {
+    "language": "C#",
+    "type": "project",
+    "category": "netcore/app/general",
```
mrward: Ah OK this is where we can put extra metadata that before we had in the .xpt.xml files. So we could put things like the icon and template image in here.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.csproj
alanmcgovern: We can glob it like: `<None Include="Path\to\**\*.*" />`

slluis: I would also prefer using a reference to an external repo or directly getting the templates from NuGets.

main/src/addins/MonoDevelop.DotNetCore/Templates/Projects/Console/CSharp/.template.config/template.json
slluis: I don't like the idea of modifying the templates. There is already a "classifications" property, whose values could be mapped to existing XS categories. Icons could also be mapped to classifications.

DavidKarlas: Ok, I just thought we want engine, not also to share 100% same templates, that complicates things, since their templates don't have .sln and .gitingore... files...
I will need to talk to VS team if we want to share templates... to see how are they going to solve this problems in VS...

main/src/core/MonoDevelop.Ide/ExtensionModel/Templates.addin.xml
```diff
@@ -121,6 +126,7 @@
 
 <Extension path="/MonoDevelop/Ide/ProjectTemplatingProviders">
 	<Class id="MonoDevelop.ProjectTemplatingProvider" class="MonoDevelop.Ide.Templates.ProjectTemplatingProvider" />
+	<Class id="MonoDevelop.ProjectTemplatingProvider" class="MonoDevelop.Ide.Templates.MicrosoftTemplateEngineProjectTemplatingProvider" />
```
slluis: Set a different id.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Codons/TemplatesPathCodon.cs
```diff
@@ -0,0 +1,43 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+using Mono.Addins;
+
+namespace MonoDevelop.Ide.Codons
+{
+	[ExtensionNode (Description = "Templates scan path.")]
+	internal class TemplatesPathCodon : ExtensionNode
```
slluis: Please don't use "Codon", that's obsolete terminology. Use "ExtensionNode" instead

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngineProcessedTemplateResult.cs
```diff
@@ -0,0 +1,67 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+using System.Collections.Generic;
+using Microsoft.TemplateEngine.Edge.Template;
+using MonoDevelop.Projects;
+
+namespace MonoDevelop.Ide.Templates
+{
+	public class MicrosoftTemplateEngineProcessedTemplateResult : ProcessedTemplateResult
```
slluis: Can this be internal?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.csproj
DavidKarlas: Ok, I didn't know plan is to share/use .Net Core team templates... that changes things... since now we can't use our own templates that include .sln files or .gitignore... So I will need to talk to them on how are they planning on including all this settings in VS.net

main/src/addins/MonoDevelop.DotNetCore/Properties/MonoDevelop.DotNetCore.addin.xml
DavidKarlas: As you noticed later I was planning to use Tags inside template file, but since we plan to share templates 100% we will need to add this in .addin.xml data, which means templates from nugets won't work out of the box but will need some AddIn logic

main/src/addins/MonoDevelop.DotNetCore/Templates/Projects/StarterWeb/FSharp/Company.WebApplication1.fsproj
DavidKarlas: yes they are grouped via GroupIdentity

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngineProjectTemplatingProvider.cs
slluis: There is API in ProjectService for checking if a file is a project. You should use that.

main/src/core/MonoDevelop.Ide/packages.config
DavidKarlas: I know :( I mentioned this problem in comment

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/TemplatingService.cs
DavidKarlas: glad you like it :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngineProjectTemplatingProvider.cs
DavidKarlas: agreed, will change

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngineProjectTemplatingProvider.cs
DavidKarlas: This is obsolete code for templates from Nuget that don't have PrimaryOutput set, I will remove this code

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngineSolutionTemplate.cs
DavidKarlas: agreed, will change

main/src/core/MonoDevelop.Ide/ExtensionModel/Templates.addin.xml
DavidKarlas: agreed, will change

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Codons/TemplatesPathCodon.cs
DavidKarlas: agreed, will change

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngineProcessedTemplateResult.cs
DavidKarlas: agreed, will change

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.csproj
slluis: I'm not sure what is the problem. The .NET Core templates are project templates, not solution templates. The current approach of generating a solution if one is not provided seems correct. The generation of gitignore files is not done by templates, but by the git add-in.

DavidKarlas: What if we want solution to have more Configurations then just Debug & Release? What if we want more SolutionFolder nesting?
We can keep current way of creating .sln files, and later if problem appears rethink(shouldn't be too much problem)... We will eventually want to share everything with VS.net, once they switch to this engine we can use same way to solve this(either create .sln inside IDE or be part of template).

slluis: > What if we want solution to have more Configurations then just Debug & Release? What if we want more SolutionFolder nesting?

.NET Core projects don't use any of that, so that's not a problem.

DavidKarlas:
Pushed new version... Only thing from comments above not addressed is https://github.com/mono/monodevelop/pull/1880#discussion_r101490326 but I will address with new version of Templating engine(when PrimaryOutputs start working)...
I did some smoke tests and everything works as expected...

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.csproj
```diff
@@ -149,61 +147,40 @@
...
+        <Url>$(NuGetUrl)Microsoft.DotNet.Test.ProjectTemplates.2.0/$(TemplatesVersion)</Url>
+        <OutputPackage>$(OutputPath)Templates\Microsoft.DotNet.Test.ProjectTemplates.2.0.$(TemplatesVersion).nupkg</OutputPackage>
+      </NuGetPackage>
+      <NuGetPackage Include="Microsoft.DotNet.Web.ProjectTemplates.1.x">
+        <Url>$(NuGetUrl)Microsoft.DotNet.Web.ProjectTemplates.1.x/$(TemplatesVersion)</Url>
+        <OutputPackage>$(OutputPath)Templates\Microsoft.DotNet.Web.ProjectTemplates.1.x.$(TemplatesVersion).nupkg</OutputPackage>
+      </NuGetPackage>
+      <NuGetPackage Include="Microsoft.DotNet.Web.ProjectTemplates.2.0">
+        <Url>$(NuGetUrl)Microsoft.DotNet.Web.ProjectTemplates.2.0/$(TemplatesVersion)</Url>
+        <OutputPackage>$(OutputPath)Templates\Microsoft.DotNet.Web.ProjectTemplates.2.0.$(TemplatesVersion).nupkg</OutputPackage>
```
mrward: What happens when a new file is added to the project? Do we lose the $(TemplateVersion) properties? Do they get expanded. I think that happens with the Xamarin.Forms addin. If it does happen I am just wondering if these NuGetPackage items should be moved to a separate MSBuild file that is imported into this one.

main/src/addins/MonoDevelop.DotNetCore/Properties/MonoDevelop.DotNetCore.addin.xml
```diff
@@ -18,19 +18,43 @@
...
+			category="netcore/library/general" />
+		<Template
+			id="Microsoft.Test.xUnit.CSharp"
+			templateId="Microsoft.Test.xUnit.CSharp"
+			path="Templates/Microsoft.DotNet.Test.ProjectTemplates.1.x.1.0.0-beta1-20170216-123.nupkg"
+			icon="md-test-project"
+			category="netcore/test/general" />
+		<Template
+			id="Microsoft.Web.Empty.CSharp"
+			templateId="Microsoft.Web.Empty.CSharp"
```
mrward: By default can we not just use the id attribute for the templateId? They seem to be the same here. Would we need them to be different?

main/src/core/MonoDevelop.Ide/ExtensionModel/Templates.addin.xml
```diff
@@ -21,6 +21,11 @@
 	<ExtensionNode name="ProjectTemplate" type="MonoDevelop.Ide.Codons.ProjectTemplateCodon"/>
 </ExtensionPoint>
 
+<ExtensionPoint path = "/MonoDevelop/Ide/Templates" name = "Microsoft Templating Engine templates">
+	<Description>Microsoft templating engine project templates to be shown in the New Project dialog.</Description>
+	<ExtensionNode name="Template" type="MonoDevelop.Ide.Codons.TemplateExtensionNode"/>
+</ExtensionPoint>
```
mrward: Do we support file templates using this extension path? Previously we had two different extension points for projects and files templates. Just wondering if this handles both or how we would support file templates in the future since I think the templating engine includes some file templates.

main/src/addins/MonoDevelop.DotNetCore/Properties/MonoDevelop.DotNetCore.addin.xml
DavidKarlas: I added wanted to give ability to avoid conflicts... but now if I think about it again... `Id` scope is just for `<Template` enties, hence your point is valid, I will fix this and reuse AddIn Id

main/src/core/MonoDevelop.Ide/ExtensionModel/Templates.addin.xml
DavidKarlas: The idea is... that we will separate them by Tag in template.json file like https://github.com/dotnet/templating/blob/master/template_feed/Microsoft.DotNet.Common.ItemTemplates/content/Nuget/.template.config/template.json#L6 vs https://github.com/dotnet/templating/blob/master/template_feed/Microsoft.DotNet.Common.ProjectTemplates.1.x/content/ConsoleApplication-CSharp/.template.config/template.json#L10

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.csproj
DavidKarlas: I never seen that behaviour in Xamarin.Forms... if that is the case... I think tool doing this should be fixed...

slluis: $(TemplateVersion) should not be expanded if the project is saved. If it is, that's a bug.

mrward: I did a quick test with Xamarin Studio 6.2 and it looks like this is fine now.

DavidKarlas:
I addressed all comments and rebased on master + squashed commits to avoid having all templates files in history...

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Codons/TemplateExtensionNode.cs
```diff
@@ -0,0 +1,80 @@
...
+		string imageId;
+
+		public string ImageId {
+			get {
+				return imageId;
+			}
+		}
+
+		[NodeAttribute ("overrideName", "If template.json is outside AddIn creator control use this to change name.")]
+		string overrideName;
```
mrward: We should allow the name to be localized. So we should follow the convention that other addins used where the attribute name uses an underscore and mark the attribute as localizable. FileFilterCodon example:

    [NodeAttribute ("_label", true, "Display name of the filter.", Localizable=true)]
    string filtername = null;

main/src/core/MonoDevelop.Ide/packages.config
mrward: Mmm the templating engine that uses Json.NET v8 has not been published to myget. Looks like they merge it into a new branch called 'rel/vs2017/post-rtw' so I guess we will need to look at upgrading after this is released.

DavidKarlas:
I will wait for templating team to release new version on templating NuGet feed which will include fix for PrimaryOutputs(so I can remove hack with GetFiles(*.*proj)) and Json.Net >= 8.0.3, once that is done we can probably merge this

DavidKarlas:
Actually I changed my mind... since Json.Net change doesn't actually change anything(we will still have 8.0.3 in bin folder...) and PrimaryOutput workarounds are good enough for current templates... I'm going to merge this and change things when NuGet is available...

PR #1887, opened 2017-02-17
Mark DesignerSupport as private so it can be found when building the …

dpugh:
…MEF catalog.


Therzok:
This breaks our unique-dll-in-app-bundle check. cc @slluis

slluis:
Marius is right, add-in dlls should never local-copied by other add-ins. I replied on the xplateditor list.

PR #1889, opened 2017-02-17
Fixed 'Bug 52553 - Intellisense fails when referencing an iOS binding…

mkrueger:
… project'

The bug got introduced by a change from the F# binding which broke all other output tracked project type scenarios.


slluis:
Can you please add unit tests for this?

PR #1891, opened 2017-02-17
[vNext] Imported wildcards

mrward:
Project model changes to handle .NET Core Sdk imports.

- [x] Wildcards that are defined in an imported project now resolve files within the main root project.
- [x] Imported wildcard globs now added to root project. Prevents files being added when saving the project.
- [x] Handle .NET Core imports using forward slashes in wildcard includes and excludes:
       `<None Include="**/*" Exclude="**/*.cs" />`
- [x] Wildcard Removes now remove items added by another imported project.
- [x] Wildcard Updates now updates items added by another imported project.
- [x] Handle extra trailing slashes in item Excludes, for example:
      `BaseOutputPath=bin\`
      `<Compile Include="**/*.cs" Exclude="$(BaseOutputPath)/**" />`

**TODO**

- [x] Display these imported files in the Solution window.

The project model changes ensure the wildcard included files are available in the MSBuildProject's EvaluatedItems. However since these items have IsImported set to true they are not added to the Project.Items collection. So they do not appear in the Solution window. Ideas:

 1. Allow a ProjectExtension to change the default behaviour of Project.LoadProjectItems so it can create a ProjectItem for an imported item. ProjectExtension new method:
    `bool OnGetSupportsImportedItem (IMSBuildEvaluatedItem item)`
 2. Custom project node builder that adds the files taken from the MSBuildProject's EvaluatedItems to the Solution window. How do changes to these items get applied to the project on saving?
 3. Custom ProjectFile class that is created by the DotNetCoreProjectExtension and added to the Project.Files collection. How do changes to these items get applied to the project on saving?

If the DotNetCoreProjectExtension adds new ProjectFiles created from the MSBuildProject's EvaluatedItems to the Project.Items collection then there is currently no way to indicate that the ProjectFile is from a wildcard. This means that on saving the project the file is added as an item even though the wildcard already includes this file.


mrward:
Updated this pull request so that imported MSBuild items can optionally be loaded into the project if the ProjectExtension indicates that it supports them. This allows .NET Core projects to show wildcard imported files in the solution window.

mrward:
@slluis - This could do with a review when you have a chance. Thanks.

PR #1892, opened 2017-02-17
Optimize GettingStarted new project handling

sevoku:
This change allows GettingStarted providers to handle new project creation and removes the default action (show GS pages automatically).

GS Provides can now implement `IGettingStartedProvider.SupportedProjectCreated (Project project, ProcessedTemplateResult templateResult)` and run any desired action based on the project creation result.


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
```diff
@@ -795,26 +795,22 @@ void InstallProjectTemplatePackages (Solution sol)
...
+			// Notify supporting GettingStarted providers
+			Project firstProject = null;
+			if (templateResult.WorkspaceItems.OfType<Solution> ().Any ())
+				// this is a solution that's been instantiated, lets just look for the first project
+				firstProject = IdeApp.Workspace.GetAllProjects ().FirstOrDefault ();
+			else
+				firstProject = templateResult.WorkspaceItems.OfType<Project> ().FirstOrDefault ();
+			if (firstProject != null) {
+				var gettingStartedProvider = GettingStarted.GettingStarted.GetGettingStartedProvider (firstProject);
+				gettingStartedProvider?.SupportedProjectCreated (firstProject, templateResult);
```
mrward: Could you just pass in a flag to indicate if a solution was created here instead of passing the template result? It looks like you only need a boolean value to indicate if a solution was created. Or alternatively just pass the templateResult without the project and move all the logic above into the getting started provider - maybe a base implementation that does that logic? Or just pass the workspaceItems? It seems that this call passes the project twice to the getting started provider using two different parameters.

sevoku: I wanted it to be as flexible as possible (just in case ;)) until we have a better Template Actions API. So I'd remove the project parameter.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
```diff
@@ -795,26 +795,22 @@ void InstallProjectTemplatePackages (Solution sol)
...
-				}
+			}
+
+			// Notify supporting GettingStarted providers
+			Project firstProject = null;
+			if (templateResult.WorkspaceItems.OfType<Solution> ().Any ())
+				// this is a solution that's been instantiated, lets just look for the first project
+				firstProject = IdeApp.Workspace.GetAllProjects ().FirstOrDefault ();
+			else
+				firstProject = templateResult.WorkspaceItems.OfType<Project> ().FirstOrDefault ();
```
Therzok: Why do we grab the first project? Shouldn't we looking at something like the startup project(s) for each run config?

PR #1894, opened 2017-02-19
Bug 51960 - App freezing whilst attempting to debug on iOS Simulator

DavidKarlas:
Problem was that session.Dispose(); happened before StoppedEvent was called which meant that Breakpoints.Remove(entryBreakpoint); that Inspector AddIn did didn’t have effect because session.Dispose(); already disconnected BreakpointStore.

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
@@ -438,8 +438,6 @@ namespace MonoDevelop.Debugger
 			session.OutputWriter = null;
 			session.LogWriter = null;
 
-			sessionManager.Dispose ();
-
 			Runtime.RunInMainThread (delegate {
 				if (cleaningCurrentSession)
 					HideExceptionCaughtDialog ();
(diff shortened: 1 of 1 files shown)


jstedfast:
nice debugging!

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
```diff
@@ -455,8 +453,9 @@ static void Cleanup (SessionManager sessionManager)
 				NotifyCallStackChanged ();
 				NotifyCurrentFrameChanged ();
 				NotifyLocationChanged ();
+			}).ContinueWith ((t) => {
+				sessionManager.Dispose ();
```
alanmcgovern: `RunInMainThread` is a blocking operation, so we should just invoke `sessionManager.Dispose ()` directly instead of wrapping it in a continuation.

If we want to wrap it in a continuation we should specify the scheduler it should run so we ensure the right thing happens.

alanmcgovern: oops, it's not synchronous. Never mind that part :p

PR #1895, opened 2017-02-19
Fix unit tests broken by TextSet confusion

dpugh:


mkrueger:
Would be nice if you could use the mono formatting style - some files are already broken because of that. I'll merge that then.

PR #1896, opened 2017-02-20
C9 Bug 51960 - App freezing whilst attempting to debug on iOS Simulator

DavidKarlas:
Problem was that session.Dispose(); happened before StoppedEvent was called which meant that Breakpoints.Remove(entryBreakpoint); that Inspector AddIn did didn’t have effect because session.Dispose(); already disconnected BreakpointStore.


DavidKarlas:
Can someone approve this for C9?(QA already tested)

PR #1871, opened 2017-02-14
[Core] Allow for MSBuild remote process reuse.

Therzok:
@slluis Is this more of what you had in mind? I'm still hitting on some weird ProcessDisconnectedExceptions which I can't trace while debugging, so I guess it's a race somewhere, but I want to know this is the right approach before going further.

PR #1898, opened 2017-02-22
Reduce project builder allocation

slluis:
Fast msbuild operations don't require an exclusive lock to the builder
anymore. This should reduce builder allocation when loading a project,
since many quick targets needs to be executed after loading.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -1104,10 +1104,11 @@ namespace MonoDevelop.Projects
 				MSBuildResult result = null;
 				await Task.Run (async delegate {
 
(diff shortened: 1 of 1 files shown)


slluis:
@alanmcgovern @Therzok Alternative fix of PR #1871

Therzok:
LGTM. Tested the patch locally, at most 6 builders are spawned, but one is longer lived and takes up to 100% CPU while processing. Sounds like an improvement.

alanmcgovern:
This makes a drastic improvement to the three cases I checked:
1) Opening Main.sln
2) Opening md-addins.sln
3) Opening MTDesigner.sln

Screencast of the same: https://screencast.com/t/QT0Zo2S8Q

Once the builder starts it only takes 20-25 seconds to fully load the typesystem. That's significantly better than what it used to be and I also don't see dozens of builders starting.

PR #1899, opened 2017-02-22
Remote connection race condition

alanmcgovern:
This fixes one race condition and then rewrites some continuation based code to be async/await as it is a bit easier to reason about.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/RemoteProcessConnection.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/RemoteProcessConnection.cs
@@ -508,21 +508,30 @@ namespace MonoDevelop.Core.Execution
 					return;
 				}
 
-				lock (pendingMessageTasks) {
-					var t = Task.Run (() => {
-						msg = LoadMessageData (msg);
-						if (type == 0)
-							ProcessResponse (msg);
-						else
-							ProcessRemoteMessage (msg);
-					});
-					t.ContinueWith (ta => {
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/RemoteProcessConnection.cs
```diff
@@ -530,7 +530,8 @@ void OnConnected (IAsyncResult res)
 
 		public Task ProcessPendingMessages ()
 		{
-			return Task.WhenAll (pendingMessageTasks.ToArray ());
+			lock (pendingMessageTasks)
+				return Task.WhenAll (pendingMessageTasks.ToArray ());
```
Therzok: 👍

Therzok: Or maybe this is holding the lock a bit too much actually, until it returns a task from .WhenAll(). Lock contention should be avoided where possible, so I'd go with the more verbose way of only locking the ToArray() call.

alanmcgovern: I think it's ok to leave it as is as the extra call is super fast. If you really want it changed we can do that :)

alanmcgovern: I'll just merge it. It's much easier to read as-is and will never be a perf bottleneck :)

PR #1906, opened 2017-02-25
Add a OSX-specific and Mono FileSystemWatcher wrapper

luhenry:
This integrates the FSW from CoreFX which is OSX specific.

### DIFF
--- a/main/build/MacOSX/System_Native.cpp
+++ b/main/build/MacOSX/System_Native.cpp
@@ -1,0 +1,638 @@
+
+#include <assert.h>
+#include <cstddef>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <errno.h>
+#include <sys/syslimits.h>
+
+extern "C" int32_t SystemNative_HasOSXSupport (void);
+extern "C" int32_t SystemNative_HasOSXSupport (void)
+{
+    return 1;
+}
+
+extern "C" char* SystemNative_RealPath(const char* path);
+extern "C" char* SystemNative_RealPath(const char* path)
+{
+    assert(path != nullptr);
+    return realpath(path, nullptr);
+}
+
+/**
+ * Constants to pass into pathconf.
+ *
+ * Note - these differ per OS so these values are the PAL-specific
+ *        values; they must be converted to the correct platform
+ *        values before passing to pathconf.
+ */
+enum PathConfName : int32_t
+{
+    PAL_PC_LINK_MAX = 1,
+    PAL_PC_MAX_CANON = 2,
+    PAL_PC_MAX_INPUT = 3,
+    PAL_PC_NAME_MAX = 4,
+    PAL_PC_PATH_MAX = 5,
+    PAL_PC_PIPE_BUF = 6,
+    PAL_PC_CHOWN_RESTRICTED = 7,
+    PAL_PC_NO_TRUNC = 8,
+    PAL_PC_VDISABLE = 9,
+};
+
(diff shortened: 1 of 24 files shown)


main/build/MacOSX/System_Native.cpp
```diff
@@ -0,0 +1,636 @@
+
```
Therzok: This file is a great idea!

main/src/core/MonoDevelop.Core/MonoDevelop.FSW/FileSystemWatcher.cs
```diff
@@ -0,0 +1,561 @@
...
+
+		enum Platform
+		{
+			OSX,
+			Mono,
+		}
+
+		static FileSystemWatcher ()
+		{
+			// FIXME add detection for Mono or OSX
```
Therzok: For self reference: Maybe the whole thing can go into its separate repo, and try a `EntryPointNotFoundException` check on whether one of the System.Native pinvokes fails.

Therzok:
All in all, this looks amazing!

main/src/core/MonoDevelop.Core/MonoDevelop.FSW/OSX/FileSystemWatcher.OSX.cs
```diff
@@ -0,0 +1,504 @@
...
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
```
slluis: The above header should be removed from all files we are taking directly from .NET Core.

ludovic-henry:
@slluis I updated according to your review.

@Therzok I added the `HasOSXSupport` P/Invoke which will test if the current native executable has the `SystemNative_*` functions (https://github.com/mono/monodevelop/pull/1906/files#diff-5a37ecc510b9e231ec56752bd1e86b3dR48)

luhenry:
@Therzok anything blocking this from getting merged?

slluis:
Other than the public class issue it looks ok to me. Now we have to find a good way of using it.

luhenry:
@slluis I fixed the public class issue.

Therzok:
@slluis my plan of attack would be to split this code into its own repo, so we can submodule it where necessary and use it in the repos we need. How does it sound?

luhenry:
@Therzok @slluis  IMO it's definitely overkill to split it in its own repo, as the long term plan is to have this FSW from CoreFX directly in Mono. Having copied the source inside XS is itself a gross hack, that shouldn't be there in the first place, but which is the fastest way to get where we want to get.

My advice would simply be to copy/paste it where you need it, keep the wrapper around the OSX and Mono "backends", and once the CoreFX FSW is in Mono, simply remove the copied OSX version, and remove/inline the wrapper and call the Mono implementation directly.

PR #1909, opened 2017-02-27
[DotNetCore] Added ASP.NET Core specific RunConfiguration settings

DavidKarlas:
@slluis is this approach OK to add asp.net core specific settings for run configuration?

image

If yes... I will add all settings...

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
@@ -132,16 +132,17 @@ namespace MonoDevelop.DotNetCore
 		DotNetCoreExecutionCommand CreateDotNetCoreExecutionCommand (ConfigurationSelector configSel, DotNetProjectConfiguration configuration, ProjectRunConfiguration runConfiguration)
 		{
 			FilePath outputFileName = GetOutputFileName (configuration);
-			var assemblyRunConfiguration = runConfiguration as AssemblyRunConfiguration;
+			var dotnetCoreRunConfiguration = runConfiguration as DotNetCoreRunConfiguration;
 
 			return new DotNetCoreExecutionCommand (
-				string.IsNullOrEmpty (assemblyRunConfiguration?.StartWorkingDirectory) ? Project.BaseDirectory : assemblyRunConfiguration.StartWorkingDirectory,
+				string.IsNullOrEmpty (dotnetCoreRunConfiguration?.StartWorkingDirectory) ? Project.BaseDirectory : dotnetCoreRunConfiguration.StartWorkingDirectory,
 				outputFileName,
-				assemblyRunConfiguration?.StartArguments
+				dotnetCoreRunConfiguration?.StartArguments
 			) {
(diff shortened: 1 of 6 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -351,6 +352,10 @@ protected override void OnBeginLoad ()
 			get { return dotNetCoreMSBuildProject.HasSdk; }
 		}
 
+		public bool IsWeb {
+			get { return dotNetCoreMSBuildProject.Sdk == "Microsoft.NET.Sdk.Web"; }
```
mrward: This will not work for F# projects. The Sdk property can have multiple values. I think there is a way to get all the supported target frameworks. This is not exposed on vNext, I have a local branch with a new method in the DotNetCoreProjectExtension method you could use. Or maybe just add an IsWeb () method to the DotNetCoreProjectExtension instead which does the check but uses the list of target frameworks.

```
               internal IEnumerable<string> GetTargetFrameworks ()
               {
                       if (HasSdk)
                               return dotNetCoreMSBuildProject.TargetFrameworks;

                       return new [] { Project.TargetFramework.ToString () };
               }
```

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreRunConfiguration.cs
```diff
@@ -0,0 +1,42 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+using MonoDevelop.Core.Serialization;
+using MonoDevelop.Projects;
+
+namespace MonoDevelop.DotNetCore
+{
+	public class DotNetCoreRunConfiguration : AssemblyRunConfiguration
```
mrward: Non-public please :)

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
slluis: I wonder if there is any capability defined by asp.net projects

slluis:
The approach looks good.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
mrward: Yes there are some ProjectCapability items for projects that use the Microsoft.NET.Sdk.Web:

DotNetCoreWeb
AspNetCore
Web

However these are not available from the Project.GetProjectCapabilities method currently. They are available as items after the MSBuildProject is evaluated by the DotNetCoreProjectExtension which is done in its OnPrepareForEvaluation.

PR #1915, opened 2017-03-02
[DotNetCore] Added ASP.NET Core specific RunConfiguration settings(AppURL and LaunchBrowser)

DavidKarlas:
Pretty happy with end result:
image
Application starts on correct url and browsers launches right after server starts listening(same as VS.net does it, by trying to connect to web server address).

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreRunConfiguration.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreRunConfiguration.cs
@@ -1,0 +1,72 @@
+//
+// DotNetCoreRunConfiguration.cs
+//
+// Author:
+//       David Karlaš <david.karlas@xamarin.com>
+//
+// Copyright (c) 2017 Xamarin, Inc (http://www.xamarin.com)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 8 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreRunConfiguration.cs
```diff
@@ -0,0 +1,72 @@
...
+// THE SOFTWARE.
+using System;
+using System.Net;
+using MonoDevelop.Core.Serialization;
+using MonoDevelop.Projects;
+using System.Net.Sockets;
+
+namespace MonoDevelop.DotNetCore
+{
+	public class DotNetCoreRunConfiguration : AssemblyRunConfiguration
```
mrward: Can we make this non-public?

slluis:
What is the checkbox in Launch Url for?

DavidKarlas:
To launch or not to launch browser... I agreed it looks akward, but I wanted to keep it similar to VS:
image

slluis:
And what's the difference between Launch Url and App Url?

DavidKarlas:
App URL is URL on which server will listen.(http://localhost:8888)
Launch URL is where browser will be pointed to.(http://localhost:8888/website/i/want/to/open/without/navigating/over/20/clicks) Launch URL can also be relative, like just "index.html") which means combine App URL... which results in http://localhost:8888/index.html

DavidKarlas:
It's also possible that you are developing micro web service and you want to launch in browser totally different web address that is calling your micro web service, which makes sense to have totally different absolute(host+port) address

slluis:
Now, let's redesign that panel so that users don't have to ask the above questions. /cc @vancura @hbons

slluis:
Also, why is it necessary to specify the full App Url? shouldn't the port number be enough?

DavidKarlas:
So user can specify on which domain/ip to listen in case of remote deployment/debugging... I guess we could change logic to append http://localhost in case only port is specified...
But for now I would prefer to keep maximum compatibility with VS to not confuse users...

hbons:
I'll have a look at this.

PR #1916, opened 2017-03-02
[vNext] Support build action changing with wildcard include

mrward:
When an MSBuildItem is included with a wildcard and then its BuildAction is changed the project should add a Remove item and an Include item. The Remove item will remove it from the original
wildcard group.

Wildcard:

    <Content Include="**\*.txt" />

Changing a .txt file to be an EmbeddedResource should add the following items to the project:

    <Content Remove="TextFile.txt" />
    <EmbeddedResource Include="TextFile.txt" />

A simple fix for the above with a test has been added.

**TODO**

The following need to be tested and/or implemented.
 - [x] Add Remove and Include item when BuildAction is changed.
 - [ ] ~~MSBuild item ordering.~~ To be done in another pull request. Affects more than just Remove and Include items.
    - [ ] ~~Remove items for the same item type (e.g. None) should be grouped together.~~
    - [ ] ~~Include items should be added in their own item group after the Remove items.~~
 - [x] Changing the BuildAction back to its original value should remove the Remove item and the Include item.
    - [x] Project loaded from disk before action
    - [x] Project not reloaded before action
 - [x] Changing a property and then changing the BuildAction back to its original value should remove the Remove item and the Include item but add an Update item for the changed property.
    - [x] Project loaded from disk before action
    - [x] Project not reloaded before action
 - [x] Properties can be set on the new item (e.g. Copy to output directory).
    - [x] Project loaded from disk before action
    - [x] Project not reloaded before action
 - [x] Properties can be removed from the new item.
    - [x] Project loaded from disk before action
    - [x] Project not reloaded before action

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -3383,6 +3383,9 @@ namespace MonoDevelop.Projects
 			var existingIncludeItem = globItem.ParentProject.GetAllItems ()
 				.FirstOrDefault (i => i.Include == item.Include);
 			if (existingIncludeItem != null) {
+				foreach (var p in existingIncludeItem.Metadata.GetProperties ().ToArray ())
+					existingIncludeItem.Metadata.RemoveProperty (p.Name);
+
 				foreach (var p in item.Metadata.GetProperties ())
 					existingIncludeItem.Metadata.SetValue (p.Name, p.Value);
 			}
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -3068,8 +3068,7 @@ namespace MonoDevelop.Projects
 		{
 			None,
 			Exclude,
-			AddUpdateItem,
-			AddRemoveAndIncludeItem
+			AddUpdateItem
 		}
 
 		/// <summary>
@@ -3098,10 +3097,10 @@ namespace MonoDevelop.Projects
 				if (expandedList.Modified || loadedProjectItems.Where (i => i.WildcardItem == globItem).Count () != expandedList.Count) {
 					if (UseAdvancedGlobSupport) {
 						// Add remove items if necessary
-						foreach (var removed in loadedProjectItems.Where (i => i.WildcardItem == globItem && !expandedList.Any (newItem => newItem.ProjectItem.Include == i.Include))) {
+						foreach (var removed in loadedProjectItems.Where (i => i.WildcardItem == globItem && !expandedList.Any (newItem => newItem.ProjectItem.Include == i.Include && newItem.ProjectItem.ItemName == globItem.Name))) {
 							var file = removed as ProjectFile;
 							if (file == null || File.Exists (file.FilePath)) {
-								var removeItem = new MSBuildItem (removed.ItemName) { Remove = removed.Include };
+								var removeItem = new MSBuildItem (globItem.Name) { Remove = removed.Include };
 								msproject.AddItem (removeItem);
 							}
 							unusedItems.UnionWith (FindUpdateItemsForItem (globItem, removed.Include));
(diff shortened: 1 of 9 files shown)


mrward:
MSBuild item and item group ordering to be looked at separately in another pull request.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -3115,6 +3116,10 @@ internal virtual void SaveProjectItems (ProgressMonitor monitor, MSBuildProject
 								msproject.AddItem (it.MSBuildItem);
 							} else if (it.Action == ExpandedItemAction.AddUpdateItem) {
 								msproject.AddItem (it.MSBuildItem);
+							} else if (it.Action == ExpandedItemAction.AddRemoveAndIncludeItem) {
```
slluis: I'm wondering if this new action is necessary. The condition in line 3100 looks for items that were loaded and which are not included anymore in the project. However, it is comparing only the Include, not the item name. If the condition is changed to compare the name too, a remove item will be generated if the build action changes.
On the other hand, in SaveProjectItem, there is already a condition to check if the item name changed, and in this case it will generate a new item. Isn't this all we have to do in case of build action change?

mrward: OK eventually got this working so the existing logic is used to generate a Remove item with a small change. When looking for items to remove from the loaded items the expanded item name is compared against the glob item name. All the project items and expanded items have the new item name. Also the glob item name is now used when creating the Remove item otherwise the wrong one was added.

mrward: Updated the code to hopefully be more simple.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -3199,9 +3199,35 @@ void SaveProjectItem (ProgressMonitor monitor, MSBuildProject msproject, Project
...
+								item.BackingEvalItem = CreateFakeEvaluatedItem (msproject, it, globItem.Include, sourceItems);
+								einfo.Action = ExpandedItemAction.AddUpdateItem;
+								items.Modified = true;
+								return;
+							}
+						}
+					} else if (item.IsFromWildcardItem && item.ItemName != item.WildcardItem.Name) {
+						include = item.Include;
+						var removeItem = new MSBuildItem (item.WildcardItem.Name) { Remove = include };
+						msproject.AddItem (removeItem);
```
slluis: Why? it seems unnecessary since the Remove item will be added here: https://github.com/mono/monodevelop/blob/1de689bdf83a311a132c5412a59538e4cfe630c7/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs#L3103

mrward: Debugging this the Remove on line 3103 is not reached. [Line 3097](https://github.com/mono/monodevelop/blob/1de689bdf83a311a132c5412a59538e4cfe630c7/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs#L3097) returns false. The wildcard item has been replaced inside SaveProjectItem. The [item's BackingItem](https://github.com/mono/monodevelop/blob/1de689bdf83a311a132c5412a59538e4cfe630c7/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs#L3234) is set in that method for the new build action. So the `loadedProjectItems.Where (i => i.WildcardItem == globItem).Count () != expandedList.Count` returns true since the new build action is not considered, it no longer has a WildcardItem.

I will try something else - maybe creating the buildItem on line 3226 instead may solve this.

Another problem will be that the Remove item will be added after the Include. Without the sorting, which is in another pull request, the ordering will not match what Visual Studio 2017 does.

mrward: Mmm, creating the buildItem on line 3226 seems to result in two MSBuildItems being saved in the project file, one with a wildcard and one the correct one, but the Remove is not added.

mrward: No, do not seem to be able to get this to work creating the buildItem on line 3226. Looks like since the wildcard item was recreated it does not match the globItem so the ProjectItem for the new build action is still ignored in the check on line 3097.

slluis: Ok, let's merge as is.

PR #1921, opened 2017-03-03
[Core] Add method for getting the debug info file of a .net project

slluis:


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
```diff
@@ -714,9 +714,15 @@ void PopulateSupportFileListInternal (FileCopySet list, ConfigurationSelector co
...
-						string mdbFile = TargetRuntime.GetAssemblyDebugInfoFile (file);
-						if (File.Exists (mdbFile))
-							list.Add (mdbFile, copyIfNewer);
+						string debugFile = file + ".mdb";
+						if (File.Exists (debugFile))
+							list.Add (debugFile, copyIfNewer);
+						debugFile = Path.ChangeExtension (file, ".pdb");
+						if (File.Exists (debugFile))
+							list.Add (debugFile, copyIfNewer);
+						debugFile = Path.ChangeExtension (file, ".ppdb");
```
DavidKarlas: .ppdb is just the way we call Portable .pdb, they both have .pdb extension but different content

Therzok:
Should we also consider /debug:embedded? This means pdb info is embedded (compressed) in the assembly itself. It's an option in VS 2017.

PR #1922, opened 2017-03-03
Modify ExtensibleTextEditor.GetErrorInformationAt to return informati…

ToddGrun:
…on for *all* errors at the specified location.

https://bugzilla.xamarin.com/show_bug.cgi?id=53024

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/ExtensibleTextEditor.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/ExtensibleTextEditor.cs
@@ -26,6 +26,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 
 using Gtk;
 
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/ExtensibleTextEditor.cs
```diff
@@ -417,18 +418,28 @@ internal string GetErrorInformationAt (int offset)
 			if (line == null)
 				return null;
 
-			var error = Document.GetTextSegmentMarkersAt (offset).OfType<ErrorMarker> ().FirstOrDefault ();
-			
-			if (error != null) {
+			var errors = Document.GetTextSegmentMarkersAt(offset).OfType<ErrorMarker>();
+			StringBuilder sb = new StringBuilder();
```
slluis: To avoid unnecessary allocation of StringBuilder you could initialize sb=null and inside the loop do `if (sb == null) sb = new StringBuilder(); else sb.AppendLine()`

slluis:
Tip: if you add the bug link to the commit message, or a text like "Fixes bug #53024", the bugzilla bug will be automatically closed when the fix is merged.

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/ExtensibleTextEditor.cs
ToddGrun: Made the suggested change

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/ExtensibleTextEditor.cs
```diff
@@ -417,18 +418,27 @@ internal string GetErrorInformationAt (int offset)
...
+						GLib.Markup.EscapeText(error.Error.Message)));
+				else
+					sb.Append(GettextCatalog.GetString("<b>Error</b>: {0}",
+						GLib.Markup.EscapeText(error.Error.Message)));
 			}
-			return null;
-		}
-		
+
+			return (sb.Length > 0 ? sb.ToString() : null);
```
slluis: You need to do a null check now

ToddGrun: Good catch! That was sloppy of me, I only verified that my scenario was working.

PR #1924, opened 2017-03-06
[vNext][MSBuild] Fix global properties not available for project build

mrward:
Fixed bug #45276 - Global MSBuild properties not available when
building project programmatically
https://bugzilla.xamarin.com/show_bug.cgi?id=45276

Properties set when building a single project would not cause a
re-evaluation of the project and would not affect the build.

var context = new ProjectOperationContext ();
context.GlobalProperties.SetValue ("PackOnBuild", "true");
IdeApp.ProjectOperations.Build (project, operationContext: context);

Here PackOnBuild whilst set would not change any existing properties
that used its value when using MSBuild. This problem did not affect
xbuild.

Now the project is re-evaluated if global properties are set. These
properties are removed after the build and the project is re-evaluated
again so they do not affect the next build.

### DIFF
--- a/main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.Formats.MSBuild/ProjectBuilder.v4.0.cs
+++ b/main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.Formats.MSBuild/ProjectBuilder.v4.0.cs
@@ -63,6 +63,7 @@ namespace MonoDevelop.Projects.MSBuild
 			MSBuildResult result = null;
 			BuildEngine.RunSTA (taskId, delegate {
 				Project project = null;
+				Dictionary<string, string> originalGlobalProperties = null;
 				try {
 					project = SetupProject (configurations);
 					InitLogger (logWriter);
@@ -78,6 +79,9 @@ namespace MonoDevelop.Projects.MSBuild
 					}
 
 					if (globalProperties != null) {
+						originalGlobalProperties = new Dictionary<string, string> ();
+						foreach (var p in project.GlobalProperties)
+							originalGlobalProperties [p.Key] = p.Value;
 						foreach (var p in globalProperties)
 							project.SetGlobalProperty (p.Key, p.Value);
 						project.ReevaluateIfNecessary ();
@@ -123,7 +127,7 @@ namespace MonoDevelop.Projects.MSBuild
 					if (project != null && globalProperties != null) {
 						foreach (var p in globalProperties)
 							project.RemoveGlobalProperty (p.Key);
-						foreach (var p in engine.GlobalProperties)
+						foreach (var p in originalGlobalProperties)
 							project.SetGlobalProperty (p.Key, p.Value);
 						project.ReevaluateIfNecessary ();
 					}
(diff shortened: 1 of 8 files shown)


main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.Formats.MSBuild/ProjectBuilder.v4.0.cs
```diff
@@ -76,6 +77,12 @@ public ProjectBuilder (BuildEngine buildEngine, ProjectCollection engine, string
 						loggers = new ILogger[] { logger };
 					}
 
+					if (globalProperties != null) {
+						foreach (var p in globalProperties)
+							project.SetGlobalProperty (p.Key, p.Value);
+						project.ReevaluateIfNecessary ();
+					}
+
```
slluis: If we set the properties here, then I guess they don't need to be set again in line 82.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.Formats.MSBuild/ProjectBuilder.v4.0.cs
```diff
@@ -117,6 +124,11 @@ public ProjectBuilder (BuildEngine buildEngine, ProjectCollection engine, string
 					result = new MSBuildResult (new [] { r });
 				} finally {
 					DisposeLogger ();
+					if (project != null && globalProperties != null) {
+						foreach (var p in globalProperties)
+							project.RemoveGlobalProperty (p.Key);
```
slluis: What if the project already had a value for a property? it should be restored here.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.Formats.MSBuild/ProjectBuilder.v4.0.cs
mrward: Yeah the code setting the properties on the project instance is probably not required if they are set already on the project. I will remove those lines:

```
if (globalProperties != null)
    foreach (var p in globalProperties)
        pi.SetProperty (p.Key, p.Value);
```

mrward:
Updated.
   - No longer updating the ProjectInstance global properties.
   - Restoring the build engine's global properties for the project.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.Formats.MSBuild/ProjectBuilder.v4.0.cs
```diff
@@ -123,6 +123,8 @@ public ProjectBuilder (BuildEngine buildEngine, ProjectCollection engine, string
 					if (project != null && globalProperties != null) {
 						foreach (var p in globalProperties)
 							project.RemoveGlobalProperty (p.Key);
+						foreach (var p in engine.GlobalProperties)
+							project.SetGlobalProperty (p.Key, p.Value);
```
slluis: That will restore the value if it is defined as a global engine property, but what if the project is defining its own value?

mrward:
Updated. Restoring the project's global properties now.

PR #1926, opened 2017-03-07
[vNext] Remove Update item for wildcard after changes reverted 

mrward:
Setting CopyToOutputDirectory to PreserveNewest will generate an
Update item for a wildcard item. If CopyToOutputDirectory is then
changed back to None the Update item would not be removed from
the project file.


mrward:
This change is not working for .NET Core projects.

mrward:
OK this is working for .NET Core projects. The problem was that the project was only looking at the project containing the glob when trying to find existing Update items instead of the main project.

PR #1927, opened 2017-03-07
[Core] Memoize ConditionParser usage.

Therzok:
This reduces duplicate string parsing and condition tokenizing/ConditionExpression creation.
Reduces runtime and allocations on a simple console project from 17MB to 8.1MB and runtime from 500ms to 300ms.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
@@ -242,7 +242,7 @@ namespace MonoDevelop.Projects.MSBuild
 
 		void Evaluate (ProjectInfo project, MSBuildEvaluationContext context, MSBuildPropertyGroup group)
 		{
-			if (!string.IsNullOrEmpty (group.Condition) && !SafeParseAndEvaluate (project, context, group.Condition, true))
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
```diff
@@ -896,14 +896,21 @@ void Evaluate (ProjectInfo project, MSBuildEvaluationContext context, MSBuildTar
 				project.Targets.Add (newTarget);
 		}
 
-		static bool SafeParseAndEvaluate (ProjectInfo project, MSBuildEvaluationContext context, string condition, bool collectConditionedProperties = false)
+		Dictionary<string, ConditionExpression> conditionCache = new Dictionary<string, ConditionExpression> ();
+		static bool SafeParseAndEvaluate (ProjectInfo project, MSBuildEvaluationContext context, string condition, Dictionary<string, ConditionExpression> conditionCache = null, bool collectConditionedProperties = false)
```
slluis: If we make this method non-static we can avoid passing along conditionCache in so many places.

Therzok: Will do.

Therzok:
Updated, sorry, forgot about this.

PR #1928, opened 2017-03-07
[DotNetCore] Update .NET Core SDK download url

mrward:
Switch to main .NET Core SDK download instead of RC 4 now that
.NET Core SDK 1.0.1 has been released.


slluis:
Let's do a release from vNext

PR #1929, opened 2017-03-08
Fix null reference exception when editing .Net Core(not Web) run configuration

DavidKarlas:


slluis:
We'll release from vNext.

PR #1931, opened 2017-03-08
Disable MonoDevelop.JSon.JSonTextEditorExtension if the webtooling ad…

ToddGrun:
…din is installed, as this extension is essentially overridden by the webtooling addin's behavior.


Therzok:
Couldn't this be fixed using insertbefore/insertafter in the webtools addin manifest? No need to special case it in XS? /cc @mkrueger

ToddGrun:
The before/after mechanism doesn't help in this case, I want to completely disable this extension.

PR #1932, opened 2017-03-08
Fix for https://bugzilla.xamarin.com/show_bug.cgi?id=53165

ToddGrun:


slluis:
I'll let @mkrueger review this.

PR #1934, opened 2017-03-09
[Core] Add support for Sdk imports to MSBuildProject

slluis:
It needs to be tested. Existing hacks to support Sdks should be removed before merging this.


mrward:
Made a change to the MSBuildEvaluationContext so the MSBuildSDKsPath property is defined. Otherwise ASP.NET Core projects do not load properly and have an invalid configuration error.

PR #1952, opened 2017-03-14
Add support for SDKs fallback search paths

slluis:
Search paths for MSBuildSDKsPath are now supported, but they are handled in a special way, since MSBuild really doesn't support fallback paths for MSBuildSDKsPath. When loading a new project that has a reference to an SDK we now look for that SDK in any of the registered locations. If a location is found, the path is provided to the project builder, which sets the MSBuildSDKsPath env var before loading the project.

I have checked this with regular projects, but not with .NET Core projects. It may be necessary to remove custom SDK handling code from the .NET Core add-in.

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
@@ -29,6 +29,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using MonoDevelop.Core;
+using MonoDevelop.Projects.MSBuild;
 
 namespace MonoDevelop.DotNetCore
 {
@@ -55,9 +56,7 @@ namespace MonoDevelop.DotNetCore
 
 			MSBuildSDKsPath = Path.Combine (SdksParentDirectory, "Sdks");
 
-			// HACK: Set MSBuildSDKsPath environment variable so MSBuild will find the
-			// SDK files when building and running targets.
-			Environment.SetEnvironmentVariable ("MSBuildSDKsPath", MSBuildSDKsPath + Path.DirectorySeparatorChar);
+			MSBuildProjectService.RegisterProjectImportSearchPath ("MSBuildSDKsPath", MSBuildSDKsPath);
 		}
 
 		public void FindSdkPaths (string sdk)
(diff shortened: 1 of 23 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
```diff
@@ -55,9 +56,7 @@ public void FindMSBuildSDKsPath ()
 
 			MSBuildSDKsPath = Path.Combine (SdksParentDirectory, "Sdks");
 
-			// HACK: Set MSBuildSDKsPath environment variable so MSBuild will find the
-			// SDK files when building and running targets.
-			Environment.SetEnvironmentVariable ("MSBuildSDKsPath", MSBuildSDKsPath + Path.DirectorySeparatorChar);
+			MSBuildProjectService.RegisterProjectImportSearchPath ("MSBuildSDKsPath", MSBuildSDKsPath);
```
mrward: This will be called multiple times. Should the RegisterProjectImportSearchPath handle the same property being updated multiple times and just replace the existing value? It looks like it will add a new entry to the list.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.Formats.MSBuild/ProjectBuilder.v4.0.cs
```diff
@@ -65,6 +67,8 @@ public ProjectBuilder (BuildEngine buildEngine, ProjectCollection engine, string
 				Project project = null;
 				Dictionary<string, string> originalGlobalProperties = null;
 				try {
+					if (sdksPath != null)
+						Environment.SetEnvironmentVariable ("MSBuildSDKsPath", sdksPath);
```
mrward: Not sure exactly how project builders work. Is there a separate process for each project or are they shared for a solution?

If they are shared then I am wondering how this will work for F# projects since we are not shipping the F# SDKs with Mono. So if you have an F# .NET Core project in the solution with a C# .NET Core project they would have different sdks path, since it looks like the check for the Sdk.props file uses the MSBuild bin/Sdks directory which would work for C# projects but not for F# projects.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
slluis: I will change RegisterProjectImportSearchPath so that it doesn't add duplicates, but in any case, that method should only be called once.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.Formats.MSBuild/ProjectBuilder.v4.0.cs
slluis: Builders can be shared by multiple projects. The env var is set just before loading or running targets on a project, so it is not a problem if different projects have different SDKs paths. In fact one of the unit tests I added test this use case.

PR #1957, opened 2017-03-15
Fix build break (PackageSpecCreatorTests) and eliminate dead code.

dpugh:


main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement.Tests/MonoDevelop.PackageManagement.Tests/PackageSpecCreatorTests.cs
```diff
@@ -83,11 +83,6 @@ void AddPackageTargetFallback (string packageTargetFallback)
 			project.AddPackageTargetFallback (packageTargetFallback);
 		}
 
-		void AddPackageTargetFallback (string packageTargetFallback)
```
ToddGrun: I already checked in the build break fix

PR #1960, opened 2017-03-16
Misc opt

Therzok:


slluis:
Are those micro optimizations really worth it?

Therzok:
The `ToCharArray` came up on a linked commit in the Slack channel, easy fix, easy commit.

WRT StringBuilder optimizations, some notable optimizations are in the AstAmbience (called document path extension), SignatureMarkup (every code tooltip), AssemblyBrowser (tree construction, decompilation, xml doc loading), VCS log, editor drawing (TextEditorData) - they're worth it as they shouldn't do extra allocations causing sporadic GCs.

There's some extra code churn I made with the:
```
-			result.Append (String.Format (GettextCatalog.GetString ("<b>Declaring Type:</b>\t{0}"), type.FullName));
 +			result.Append (GettextCatalog.GetString ("<b>Declaring Type:</b>\t{0}", type.FullName));
```

But didn't it cleaned up code, so I didn't to leave them out.

Therzok:
I did a full pass so this can be easily checked later on, and first of all, for consistency.

slluis:
Just wondering if it is worth the time of you fixing all of that, and my time reviewing it.

slluis:
What I mean is that we should focus on real performance issues. I don't think IdeVersionInfo.cs is a performance bottleneck. We should not change fully functional code (even if not perfect) just for the sake of consistency.

Therzok:
I'll refrain from doing unneeded changes from now on. The idea is that those hot loops I mentioned in the previous comment showed up on profiling charts, so it ended up being worth it to do it for those. At the time, the difference between just looking at other usages of the wrong pattern (since the test scenarios I have on profiling do not cover everything) and fixing them was minor.

main/src/addins/CSharpBinding/MonoDevelop.CSharp/AstAmbience.cs
```diff
@@ -244,45 +249,50 @@ public string GetEntityMarkup (SyntaxNode e)
...
 				// sb.Append (entity.Name);
-			}*/
+			}*/
 
-			string markup = sb.ToString ();
-			if (IsObsolete (e as MemberDeclarationSyntax))
-				return "<s>" + markup + "</s>";
-			return markup;
+			if (IsObsolete (e as MemberDeclarationSyntax)) {
+				sb.Append ("</s");
```
slluis: Wrong closing tag.

main/src/addins/MonoDevelop.Gettext/MonoDevelop.Gettext/Catalog.cs
```diff
@@ -976,7 +976,8 @@ public void DeleteHeader (string key)
 				foreach (string line in value.Split (new string[] {Environment.NewLine}, StringSplitOptions.None)) {
 					if (sb.Length != 0)
 						sb.AppendLine ();
-					sb.Append ("# " + line);
+					sb.Append ("#");
```
slluis: Missed space after #

main/src/core/Mono.TextEditor.Shared/Mono.TextEditor.Utils/Diff.cs
```diff
@@ -472,7 +472,7 @@ static void WriteHunks (Queue<Hunk> qh, TextDocument baseDocument, TextDocument
 
 			int remEnd = System.Math.Min (baseDocument.LineCount, item.RemoveStart + item.Removed + item.Context);
 			for (int i = item.RemoveStart + item.Removed; i < remEnd; i++) {
-				sb.AppendLine (" " + baseDocument.GetLineText (i, false));
+				sb.AppendLine (" ").AppendLine (baseDocument.GetLineText (i, false));
```
slluis: First call should be Append, not AppendLine

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/Ambience.cs
```diff
@@ -604,7 +606,9 @@ public static string GetDocumentationMarkup (ISymbol member, string doc, Documen
 						case "seealso":
 							if (string.IsNullOrEmpty (options.HighlightParameter)) {
 								ret.Append (options.FormatHeading (GettextCatalog.GetString ("See also:")));
-								ret.Append (" " + EscapeText (xml ["cref"] + xml ["langword"]));
+								ret.Append (" ");
+								ret.Append (EscapeText (xml ["cref"]));
+								ret.Append (xml ["langword"]);
```
slluis: `xml ["langword"]` was escaped before the change

Therzok:
Feedback fixed.

PR #1962, opened 2017-03-16
6.2.0.1718 with fix win mono build

simonegli8:


simonegli8:
This fix fixes building with Mono xbuild on Windows. Building with Mono xbuild was broken since a long time on Windows (I'm talking about building EVERYTHING, not MD).
The problem was the assembly resolver was bad in the builder exe, and xbuild format v12 and v14 csproj's did not build and were not included in the windows build od MD, because they did reference the wrong assemblies.

PR #1968, opened 2017-03-16
Fix for https://bugzilla.xamarin.com/show_bug.cgi?id=53412

ToddGrun:
The issue here is that TextChangeEventArgs.GetNewOffset was using an updated position to pass to later calls. I've gone ahead and got rid of TextChange.GetNewOffset (only this caller remained), and instead of updating the position, updated a changeDelta.

The *real* fix here is to switch things over to using tracking or snapshot points, but this at least fixes the issue at hand (and potentially others), and we can debate switching over to the other data structures later.


ToddGrun:
Hold off on reviewing this for a bit, looks like I have a bug...

PR #1969, opened 2017-03-17
[Gtk] Don't show Gtk# File Templates if no Gtk# reference exists.

Therzok:
This came up on the XPlat Code Editor mailing list. Right now, this will disable gtk# file templates on a project which does not reference gtk-sharp.


Therzok:
cc @mhutch

Therzok:
This still leaves the DrawingArea tooltip available. I guess the more generic fix is providing a way to check references inside the .addin.xml. cc @mrward

PR #1970, opened 2017-03-17
[GtkCore] Disable Gtk templates if there is no Gtk# reference.

Therzok:


slluis:
Maybe we could have a more generic condition, like <HasReference assembly="gtk-sharp" />

Therzok:
Superseded by #1972

PR #1971, opened 2017-03-17
[NUnit] Concurrent-safe AbstractResultsStore

Therzok:
This is a tentative fix for a concurrency case on registering unit tests.
Based on the stacktrace in the bug report, this seems like the TestRecord
is modified concurrently, thus causing issues on getting an item that
would've been added/removed (i.e. array resize).

Bug 42513 - Test result store seems to be corrupted - internal error when running many tests


Therzok:
Best seen ignoring whitespace.

PR #1975, opened 2017-03-17
[Core] Fix caching issue when getting source files

slluis:
Fixes bug #53384

PR #1976, opened 2017-03-17
[Core] Fix caching issue when getting source files

slluis:
Fixes bug #53384


alanmcgovern:
We might need to revert this https://github.com/mono/monodevelop/pull/1975#issuecomment-288116040

PR #1977, opened 2017-03-17
Cycle9 fixed mono build on windows

simonegli8:
Building with Mono on Windows was broken.
This fix does the following things:
1. Make the projects
   MonoDevelop.Projects.Formats.MSBuild.v14.0.csproj and 
   MonoDevelop.Projects.Formats.MSBuild.v12.0.csproj
   build again on Windows, i.e. reference the Mono xbuild assemblies for them to build.
  changed Assembly resolver in Main.cs, so it first looks in the exe's path and then in the msbuild/xbuild
  bin path.
2. When running xbuild, shutdown xbuild after solution has been built, and start it when the process has 
  been killed. This is necessary, as on windows, if xbuild was not sutdown, there were open files.
3. Set the current directory to the exe path on MD start, so files pinned to the taskbar can be opened, i.e. 
  MD can be started with wrong current directory.

This pull request has been tested with 6.2.0

This pull request is a cleaned up & tested version of the closed request:
#1962

We're talking here about mono building EVERYTHING on Windows, that was broken, NOT building MD.


simonegli8:
This pull request has been tested with 6.2.0

simonegli8:
We're talking here about mono bulding EVERYTHING on Windows, that was broken, NOT building MD.

simonegli8:
This pull request is a cleaned up version of the closed request:
#1962

PR #1979, opened 2017-03-17
Fix selection to handle lead/anchor == -1

dpugh:


main/src/core/Mono.TextEditor.Platform/Platform/Text/Impl/MdView/TextSelection.cs
```diff
@@ -51,7 +51,11 @@ public VirtualSnapshotPoint ActivePoint
         {
             get
             {
-                SnapshotPoint snapshotPoint = new SnapshotPoint(_textView.TextSnapshot, _textEditor.SelectionLeadOffset);
+                int offset = _textEditor.SelectionLeadOffset;
+                if (offset == -1)
```
dpugh: An alternate fix would be to modify SelectionLeadOffset to not return -1 when selection is empty (which would mirror the behavior of SelectionRange).

PR #1981, opened 2017-03-17
Switch MD TextDocument/Undo handling to annotate the undo-initiated e…

dpugh:
…dits the same was VS does.

Also set reiterated version number on undo/redo.


ToddGrun:
LGTM

PR #1982, opened 2017-03-18
Always set workspace ActiveConfigurationId to a fully expanded one

filipnavara:
Always set workspace ActiveConfigurationId to a fully expanded one to
prevent unnecessary reloading of project files. For some projects the
SelectActiveConfiguration method first set the ActiveConfigurationId to
non-expanded version (eg. “Debug”) and later called SelectActiveRuntime
which set it again to an expanded one (eg. “Debug|Mixed Platforms”).
This resulted in double reloading of the project files.


slluis:
Thanks!

PR #1983, opened 2017-03-20
[53089][Ide] Fix IWorkbenchWindow.Closing event handling with async operations

sevoku:
If a workbench window contains an unsaved document, we ask the user to save it and run the save operation asynchronously. This is controlled by the WorkbenchWindowEventArgs.Cancel property and the result contols whether we set specific Gtk/GLib events as hanled or not (i.e. in order to allow or disallow Gtk to close a window).

The first problem here was the async EventHandler for the Closing event (https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs#L844-L868) which was not awaited and the window was closed before the Save operation had a chance to run to completion. This is solved by an EventHandler delegate returning a Task, making it possible to await the handlers subscribed to the event (see https://github.com/mono/monodevelop/pull/1983/files#diff-8da6280a955ed74b6e47759bb5b27f1fR815).

The second problem with this approach was that Gtk doesn't know anything about awaiting Tasks and if one of its event handlers leaves the synchronization context (like the asynchrounous
file saving), it will let the signal pass without waiting for its handled state to be updated. This is solved by delaying window closing/destruction (always set the initial delete/close event
handled state and close the window in a continuation task synchronized with the UI thread)

(fix bug #53089)

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/SdiWorkspaceWindow.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/SdiWorkspaceWindow.cs
@@ -38,6 +38,7 @@ using MonoDevelop.Components.Commands;
 using MonoDevelop.Ide.Extensions;
 using MonoDevelop.Ide.Gui.Content;
 using MonoDevelop.Components.DockNotebook;
+using System.Threading.Tasks;
 
 namespace MonoDevelop.Ide.Gui
 {
@@ -443,17 +444,17 @@ namespace MonoDevelop.Ide.Gui
 			}
 		}
 		
-		public bool CloseWindow (bool force)
+		public Task<bool> CloseWindow (bool force)
 		{
 			return CloseWindow (force, false);
 		}
 
-		public bool CloseWindow (bool force, bool animate)
+		public async Task<bool> CloseWindow (bool force, bool animate)
 		{
 			bool wasActive = workbench.ActiveWorkbenchWindow == this;
 			WorkbenchWindowEventArgs args = new WorkbenchWindowEventArgs (force, wasActive);
 			args.Cancel = false;
-			OnClosing (args);
+			await OnClosing (args);
 			if (args.Cancel)
 				return false;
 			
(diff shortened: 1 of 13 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/SdiWorkspaceWindow.cs
```diff
@@ -808,10 +809,10 @@ protected virtual void OnTitleChanged(EventArgs e)
 			}
 		}
 
-		protected virtual void OnClosing (WorkbenchWindowEventArgs e)
+		protected virtual async System.Threading.Tasks.Task OnClosing (WorkbenchWindowEventArgs e)
 		{
 			if (Closing != null) {
-				Closing (this, e);
+				await Closing (this, e);
```
slluis: This is not right since there can be multiple event handlers, and this invocation will return the task of the last registered handler. You should get the invocation list from the delegate and await on each handler.

sevoku: yes, definitely correct. Fixed in 9c9f2aa.

sevoku: Hhm, actually the foreach loop doesn't run the handlers at the same time. awaiting Task.WhenAll might be faster, but then all handlers would access the arguments concurrently. @slluis thoughts? would it make more sense?

slluis: I think running the handlers sequentially is ok.

CampsiteHelper:
Hi saw this defect and fix...    Not sure is the right approach in this millennium.  Why not save documents all the time?  What is the point in prompting?  Does anyone ever say no, please don't save my work?

slluis:
If the document is saved, there is no way to revert. I sometimes do temporary changes that are not intended to be saved (for example, when playing with intellisense to discover api). I guess other users do the same.

PR #1985, opened 2017-03-20
[53443][Ide] Filter recent templates by visibility in new project dialog

sevoku:
This patch adds recent template filtering by the template visibility (solution/project) to the new project dialog.

(fix bug #53443)

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
@@ -333,7 +333,10 @@ namespace MonoDevelop.Ide.Projects
 		{
 			Predicate<SolutionTemplate> templateMatch = GetTemplateFilter ();
 			templateCategories = IdeApp.Services.TemplatingService.GetProjectTemplateCategories (templateMatch).ToList ();
-			recentTemplates = IdeApp.Services.TemplatingService.RecentTemplates.GetTemplates ().ToList ();
+			if (IsNewSolution)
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
```diff
@@ -333,7 +333,10 @@ void LoadTemplateCategories ()
 		{
 			Predicate<SolutionTemplate> templateMatch = GetTemplateFilter ();
 			templateCategories = IdeApp.Services.TemplatingService.GetProjectTemplateCategories (templateMatch).ToList ();
-			recentTemplates = IdeApp.Services.TemplatingService.RecentTemplates.GetTemplates ().ToList ();
+			if (IsNewSolution)
+				recentTemplates = IdeApp.Services.TemplatingService.RecentTemplates.GetTemplates ().Where ((t) => t.IsMatch (SolutionTemplateVisibility.NewSolution)).ToList ();
```
mrward: What happens if the SolutionVisibility is All? Those project templates would be filtered out which I suspect is not what we want. I think we should be checking that the SolutionTemplateVisibility is not NewProject since that is used for project templates that are only available when adding to an existing solution.

sevoku: Isn't this exactly what `IsMatch` does: https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/SolutionTemplate.cs#L211 ?

mrward: Sorry, good point. The change is good then 👍

PR #1989, opened 2017-03-20
[DotNetCore] Hide project templates if .NET Core Sdk is not installed

mrward:
Also cache the .NET Core Sdk location and whether it is installed to prevent it from being run multiple times when checking the project templates should be hidden or not.


mrward:
Looks like this check will be need to be changed since we will only need the .NET Core runtime with a later Mono version. So the project templates should be displayed if the .NET Core runtime is installed instead of checking for the .NET Core SDK. Mono's MSBuild should handle the compiling.

mrward:
So as of today the logic should probably be something like:

 1. .NET Core SDK is installed => Show project templates.
 2. .NET Core SDK is not installed + .NET Core runtime is installed + SDKs ship with Mono => Show project templates.
 3. .NET Core SDK is not installed + .NET Core runtime is not installed + SDKs ship with Mono => Only show the .NET Standard library project template (has no need for the runtime).
 4. .NET Core SDK is not installed + SDKs not shipped with Mono => Do not show project templates.

slluis:
Yes, that logic makes sense. Maybe it would be useful to have an condition that checks if an SDK is available in any of the registered MSBuildSDKsPaths.

PR #1993, opened 2017-03-21
Finds SDK path with MSBuildProjectService.FindSdkPath() instead

dazhao-msft:


dazhao-msft:
Thanks Matt. Do you know what's going wrong with the code formatting? I compared the change in VS Code with rendering whitespace enabled, but I couldn't see any difference. Both places have the same amount of tabs.

mrward:
It is only formatting of the `MSBuildProjectService.FindSdkPath(project.TargetRuntime, project.GetReferencedSDKs());` part I was referring to which is pretty minor. The rest of the existing code has spaces before each '('. I would guess that VS Code is using a Visual Studio code style, which does not match the style used in the MonoDevelop code base. The other bits may just be different line endings, but we have mixed line endings elsewhere in the code base.

PR #1997, opened 2017-03-21
Yet another cut on adding ITextView to MD.

dpugh:
Do NOT merge this, it is only for review and commentary


main/src/core/Mono.TextEditor.Platform/PlatformCatalog.cs
```diff
@@ -89,7 +89,8 @@ private static CompositionContainer CreateContainer()
                     try
                     {
                         var assemblyFilePath = assemblyNode.Addin.GetFilePath(assemblyNode.FileName);
-                        var assembly = Assembly.LoadFrom(assemblyFilePath);
+                        var assemblyName = AssemblyName.GetAssemblyName(assemblyFilePath);
+                        var assembly = Assembly.Load(assemblyName);
```
slluis: This won't work if the add-in is located is located in its own directory. Assembly.Load won't find the assembly.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditor.cs
```diff
@@ -50,6 +50,8 @@ namespace MonoDevelop.Ide.Editor
 	public sealed class TextEditor : Control, ITextDocument, IDisposable
 	{
 		readonly ITextEditorImpl textEditorImpl;
+		public readonly Microsoft.VisualStudio.Text.Editor.ITextView TextView;
```
slluis: I would use a property here, it makes the api less rigid.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.csproj
```diff
@@ -140,6 +140,27 @@
...
+      <HintPath>..\..\..\packages\Microsoft.VisualStudio.Text.Logic.15.0.26201\lib\net45\Microsoft.VisualStudio.Text.Logic.dll</HintPath>
+      <Private>True</Private>
+      <SpecificVersion>False</SpecificVersion>
+    </Reference>
+    <Reference Include="Microsoft.VisualStudio.Text.UI">
+      <HintPath>..\..\..\packages\Microsoft.VisualStudio.Text.UI.15.0.26201\lib\net45\Microsoft.VisualStudio.Text.UI.dll</HintPath>
+      <Private>True</Private>
+      <SpecificVersion>False</SpecificVersion>
+    </Reference>
+    <Reference Include="System.ComponentModel.Composition" />
```
slluis: A guess all of those packages are open source, right?

main/src/core/Mono.TextEditor.Platform/Platform/Text/Impl/MdView/TextEditorFactoryService.cs
```diff
@@ -31,7 +31,7 @@ namespace WebToolingAddin
     /// Provides a VisualStudio Service that aids in creation of Editor Views
     /// </summary>
     [Export(typeof(ITextEditorFactoryService))]
-    internal sealed class TextEditorFactoryService : ITextEditorFactoryService
+    public sealed class TextEditorFactoryService : ITextEditorFactoryService
```
dpugh: Doesn't need to be., changing back.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.csproj
dpugh: All the MS.VisualStudio ones are or will be. I'm not sure about System.ComponentModel.Composition.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/SystemAssemblyService.cs
```diff
@@ -274,8 +274,17 @@ internal static AssemblyName GetAssemblyNameObj (string file)
 		public static string GetAssemblyName (string file)
 		{
 			return AssemblyContext.NormalizeAsmName (GetAssemblyNameObj (file).ToString ());
-		}
-		
+		}
+
+		public static Assembly AssemblyLoad (string asmPath) {
+			if (Environment.OSVersion.Platform == PlatformID.Win32NT) {
```
mhutch: It might demonstrate intent a bit better to condition this on runtime rather than OS (`Type.GetType ("Mono.Runtime") != null` is the documented way to detect Mono) but not a big deal.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.csproj
mhutch: System.ComponentModel.Composition is part of the BCL, included with Mono as well as .NET

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/SystemAssemblyService.cs
mhutch: Also, TBH `MonoDevelop.Core.Platform` might be a more appropriate place? SystemAssemblyService is more about targeting than hosting.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditor.cs
```diff
@@ -1503,22 +1509,22 @@ internal void UpdateBraceMatchingResult (BraceMatchingResult? result)
...
-            }
-            set
-            {
-                Runtime.AssertMainThread();
-                textEditorImpl.IndentationTracker = value;
-            }
-        }
-
-        public event EventHandler FocusLost { add { textEditorImpl.FocusLost += value; } remove { textEditorImpl.FocusLost -= value; } }
+		[EditorBrowsable(EditorBrowsableState.Advanced)]
```
mhutch: BTW, here's a .editorconfig with MD code style https://github.com/mono/monodevelop/blob/roslyn-ivt/.editorconfig

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/SystemAssemblyService.cs
mhutch: Also, given that it's very non-obvious what this is for to someone without context, it would be good to have some comments :)

main/src/core/Mono.TextEditor.Platform/Platform/Text/Impl/MdView/TextCaret.cs
```diff
@@ -12,7 +12,7 @@
 using Mono.TextEditor;
 using MonoDevelop.Ide.Editor;
 
-namespace WebToolingAddin
+namespace Microsoft.VisualStudio.Text.Editor.Implementation
 {
     public class TextCaret : ITextCaret
```
dpugh: No. At one point in time, MEF exported needed to be public but they got better. We've got a lot of leftover code that should be switched public -> internal but it hasn't been a priority.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditor.cs
dpugh: I'll make this change.

main/src/core/Mono.TextEditor.Platform/Platform/Text/Impl/MdView/TextCaret.cs
mkrueger: TextCaret is in Mono.TextEditor.Shared if that is a class that's going to be re used it should move to Ide.

dpugh: ITextCaret is defined in in the MS.VS.Text.UI.dll we get from nuget.

I'll change the impl to internal.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditor.cs
dpugh: The original file was mixed tabs and spaces. The changes below change the spaces section to tabs.

mhutch: Yup, I realize. But some of the style's still not quite right e.g. there should be a space before parens. It's not a big deal but I figured the editorconfig would help in future :)

mkrueger:
Let's discuss the API things by mail - doesn't really belong to that PR.

Therzok:
I think this should be re-targeted against master.

PR #1998, opened 2017-03-21
[NewProjectDialog] Disable 'Create a project directory within the solution directory' when adding projects to existing solution

VincentDondain:
Fix bug #53670: [New Project Dialog] Remove "Create a project directory within the solution directory" when adding projects to existing solution
(https://bugzilla.xamarin.com/show_bug.cgi?id=53670)

We need to prevent users from unchecking "Create a project directory within the solution directory" when adding projects to existing solution.

This is an action that can break projects if the user chooses to override files (users mistake because we ask confirmation but still there's no point in allowing that).

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
@@ -199,7 +199,8 @@ namespace MonoDevelop.Ide.Projects
 		void UpdateDefaultSettings ()
 		{
 			UpdateDefaultGitSettings ();
-			PropertyService.Set (CreateProjectSubDirectoryPropertyName, projectConfiguration.CreateProjectDirectoryInsideSolutionDirectory);
+			if (IsNewSolution)
+				PropertyService.Set (CreateProjectSubDirectoryPropertyName, projectConfiguration.CreateProjectDirectoryInsideSolutionDirectory);
 			PropertyService.Set (SelectedLanguagePropertyName, GetLanguageForTemplateProcessing ());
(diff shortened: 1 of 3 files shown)


VincentDondain:
@mrward legit comment, fixed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
```diff
@@ -194,21 +193,13 @@ void SetDefaultSettings ()
...
-			}
-			return PropertyService.Get (CreateProjectSubDirectoryInExistingSolutionPropertyName, true);
+			projectConfiguration.CreateProjectDirectoryInsideSolutionDirectory = PropertyService.Get (CreateProjectSubDirectoryPropertyName, true);
 		}
 
 		void UpdateDefaultSettings ()
 		{
 			UpdateDefaultGitSettings ();
-			UpdateDefaultCreateProjectDirectorySetting ();
+			PropertyService.Set (CreateProjectSubDirectoryPropertyName, projectConfiguration.CreateProjectDirectoryInsideSolutionDirectory);
```
mrward: Not tested this but this looks like it will change the default setting to true if it was false and you add a new project to an existing solution. Probably should not be setting this property if NewSolution is false. I think projectConfiguration.CreateProjectDirectoryInsideSolutionDirectory is set based on the check box which we will be forcing to true for a new project being added to a solution.

VincentDondain:
@mrward yes I agree with you and it was indeed always setting the parameter to true on add to existing solution.

While this behavior wasn't technically incorrect, since we force the user to enable this option we shouldn't override his settings.

PR #2000, opened 2017-03-22
[F#] Completion improvements

nosami:


Therzok:
Better commit messages help a lot when reviewing code. Took me a while to understand what the changes were supposed to do

PR #2001, opened 2017-03-22
[DotNetCore] Fix new fsharp files not added to project

mrward:
F# files were not being saved in the project file when they were added to the project. The default file globs were being found even though the condition means they should not be used. So the new F# file was not added to the project.

There is still a problem here if a new F# file is added before the FSharp.NET.Sdk NuGet package is restored for the first time. The new file would then not be added to the project file. After the restore completes the project is re-evaluated so the file globs are updated. Adding a F# file at that point works as expected. Not sure of a good way to fix this. Whilst we could prevent new files being added whilst the FSharp.NET.Sdk NuGet package is being restore it would be better IDE just did the right thing and added the F# file to the project.


nosami:
Should we bundle a version of FSharp.NET.SDK?

mrward:
The FSharp.NET.SDK is a separate NuGet package specified in the F# project. Shipping it will probably not help with the NuGet restore. NuGet will still try to go online even if things are in your local NuGet cache. This can take some time.

What would be good is some way to override the default .NET Core SDK globs just during that short period of time when a new F# project is created. Basically setting EnableDefaultCompileItems to false whilst evaluating the F# project. That would fix the new file not being saved in the .fsproj when the first NuGet package restore is being run.

PR #2003, opened 2017-03-22
[RoslynSymbolCompletionData] Handle Deprecated and Obsoleted attributes

VincentDondain:
Xamarin.iOS for instance is using those custom attributes and this seems to be the simplest change to handle them.
Also their `ContainingNamespace` isn't `System`.
This change is unlikely to clash with something else.


VincentDondain:
Unrelated error:
"MonoDevelop.Projects.Formats.MSBuild/BuildEngine.v4.0.cs(53,12): error CS1061: Type `Microsoft.Build.Evaluation.ProjectCollection` does not contain a definition for `SetGlobalProperty` and no extension method `SetGlobalProperty` of type `Microsoft.Build.Evaluation.ProjectCollection` could be found. Are you missing an assembly reference?"

PR #2010, opened 2017-03-23
[C9][PO] Remove Fuzzy annotation from all loc team translations.

Therzok:


slluis:
No more releases are planned for C9.

PR #2013, opened 2017-03-23
Add support for defining default policies

slluis:
This patch introduces a new Default policy set, which defines the default
policies when there are no solution-defined or user-defined policies.

Before this patch, default policies were obtained by just creating an intance
of the policy type. This is still the case when a policy is not set for the
Default policy set.

This new Default policy set is mostly useful for defining default policies
for specific file types (which was impossible to do).


main/src/addins/CSharpBinding/MonoDevelop.CSharp.Diagnostics/InconsistentNaming/NameConventionRule.cs
```diff
@@ -126,6 +126,21 @@ public NameConventionRule Clone ()
 			};
 		}
 
+		public bool Equals (NameConventionRule other)
+		{
+			var empty = new string [0];
```
Therzok: `Array.Empty<string> ()`?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.Policies/PolicySet.cs
```diff
@@ -78,11 +82,16 @@ public PolicySet ()
 		
 		protected override T GetDefaultPolicy<T> ()
 		{
+			// The default policy set always resturns a value for any type of policy.
+			if (IsDefaultSet)
+				return new T ();
```
Therzok: How does this compile without a `new()` constraint on `T`?

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Diagnostics/InconsistentNaming/NameConventionRule.cs
slluis: Not available in .NET 4.5.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.Policies/PolicySet.cs
slluis: The constraint is specified in the virtual method that this overrides.

Therzok: Missed the override part, 👍

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Diagnostics/InconsistentNaming/NameConventionRule.cs
Therzok: *patiently waiting for roslyn-ivt*

PR #2014, opened 2017-03-23
[F#] Use F# records for Formatter types

nosami:
So that we don't need to implement IEquatable and ICloneable.
Also updated obsolete Fantomas api calls


slluis:
Does it mean that the implementation of IEquatable is implicit?

nosami:
Yes, they are compiler generated. Looks something like this as C#

```cs
        //
        // Methods
        //
        [CompilerGenerated]
        public sealed override int CompareTo (object obj, IComparer comp);

        [CompilerGenerated]
        public sealed override int CompareTo (object obj);

        [CompilerGenerated]
        public sealed override int CompareTo (FSharpFormattingSettingsRecord obj);

        [CompilerGenerated]
        public sealed override bool Equals (object obj, IEqualityComparer comp);

        [CompilerGenerated]
        public sealed override bool Equals (FSharpFormattingSettingsRecord obj);

        [CompilerGenerated]
        public sealed override bool Equals (object obj);

        [CompilerGenerated]
        public sealed override int GetHashCode (IEqualityComparer comp);

        [CompilerGenerated]
        public sealed override int GetHashCode ();
    ```

File: main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpFormattingPolicyPanel.fs [hunk 1]
```diff
@@ -5,15 +5,16 @@ open MonoDevelop.Ide.Gui.Dialogs
...
+    let mutable policy = DefaultFSharpFormatting.policy
     let mutable panel = new FSharpFormattingPolicyPanelWidget()
     override __.CreatePanelWidget() =
         panel <- new FSharpFormattingPolicyPanelWidget()
         panel.Initialize()
         Control.op_Implicit panel
 
     override __.LoadFrom(p : FSharpFormattingPolicy) =
-        policy <- p.Clone()
+        let formats = p.Formats |> function null -> ResizeArray<FSharpFormattingSettings>() | _ -> p.Formats
```
Therzok: Does this copy the content of one list to another?

nosami: this line is `let formats = p.Formats ?? new List<FSharpFormattingSettings>()``

The line below should copy, but I'm going to check - maybe it doesn't for reference types.

nosami:
```
type record = { list:ResizeArray<int> };;
 type record =
   {list: ResizeArray<int>;}

let l = ResizeArray<int>(1);;
 val l : System.Collections.Generic.List<int>

l.Add 1;;
 val it : unit = ()

let x = { list = l };;
 val x : record = {list = <seq>;}

let y = { x with list=x.list };;
 val y : record = {list = <seq>;}

x.list.GetHashCode();;
 val it : int = 572225589

y.list.GetHashCode();;
 val it : int = 572225589
```

So, it's not copied. I'll update

PR #2017, opened 2017-03-23
[Build] Grab the build revision from the buildinfo file and use it to create a full version value including the revision.

sgmunn:


sgmunn:
I need this for telemetry to build in md-addins. This is used to be able to use it set the assembly version of the startup app.

PR #2018, opened 2017-03-23
[Core] Added `GetReferencedExtensionsFromFlavor` to `SolutionItem`.

viniciusjarina:
* This method is used by the XAML forms previewer and by the iOS Designer.
When a storyboard belong to a Shared Asset project we need to find the iOS project which references that shared asset project.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionItem.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionItem.cs
@@ -375,6 +375,24 @@ namespace MonoDevelop.Projects
 			return ItemExtension.OnGetReferencedItems (configuration);
 		}
 
+		public T [] GetReferencedExtensionsFromFlavor<T> (string projectTypeName, ConfigurationSelector configuration) where T : ProjectExtension
+		{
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionItem.cs
```diff
@@ -375,6 +375,24 @@ public IEnumerable<SolutionItem> GetReferencedItems (ConfigurationSelector confi
 			return ItemExtension.OnGetReferencedItems (configuration);
 		}
 
+		public T [] GetReferencedExtensionsFromFlavor<T> (string projectTypeName, ConfigurationSelector configuration) where T : ProjectExtension
```
Therzok: Why does this return an array?

viniciusjarina: No reason. Should I return an IEnumerable?

Therzok: Most of the surrounding API does. I'd say yes

PR #2019, opened 2017-03-24
Bug 53371 - "Debugger Operation Failed" when trying to debug an NUnittest after the test finishes

DavidKarlas:
Since 06a9d3ef1d Dispose was not blocking anymore for process to finish, causing dispose of console to happen before process terminated hence NRE exceptions.


slluis:
Can we avoid the wait? instead of a dispose we could have an async shutdown method.

DavidKarlas:
@slluis I know this is basically same thing... still `.Wait();` but thats how this method works e.g: https://github.com/mono/monodevelop/blob/master/main/src/addins/MonoDevelop.UnitTesting.NUnit/MonoDevelop.UnitTesting.NUnit/NUnitAssemblyTestSuite.cs#L425

In future whole RunUnitTest API will need to change to return Task<T>, if you want I can do that now...

PR #2024, opened 2017-03-24
When an ad-hoc document is renamed unload the Adhoc project

KirillOsenkov:
Otherwise it will try to find the new document in the old project and fail, causing analysisDocument to be null. And when analysisDocument is null, various language services are turned off.

The symptom was when calling Save As on a .css file completion stopped working. I've verified that after this call is made completion works fine after SaveAs on a .css file.


KirillOsenkov:
whitelist

Therzok:
whitelist

PR #2025, opened 2017-03-25
Fix "Move Type to file" code action

mhutch:
And a whole bunch of other buglets I found along the way.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
@@ -765,21 +765,7 @@ namespace MonoDevelop.Ide.TypeSystem
 			return document;
 		}
 
-		Dictionary<string, SourceText> changedFiles = new Dictionary<string, SourceText> ();
 		ProjectChanges projectChanges;
-
-		public override bool TryApplyChanges (Solution newSolution)
-		{
-			lock (changedFiles)
-				changedFiles.Clear ();
-			return base.TryApplyChanges (newSolution);
-		}
-
-		protected override void ApplyProjectChanges (ProjectChanges projectChanges)
-		{
-				this.projectChanges = projectChanges;
-				base.ApplyProjectChanges (projectChanges);
-		}
 
 		protected override void OnDocumentTextChanged (Document document)
 		{
@@ -843,8 +829,13 @@ namespace MonoDevelop.Ide.TypeSystem
 		{
 		}
 
+		//FIXME: this should NOT be async. our implementation is doing some very expensive things like formatting that it shouldn't need to do.
+		protected override void ApplyDocumentTextChanged (DocumentId id, SourceText text)
+		{
+			tryApplyState_documentTextChangedTasks.Add (ApplyDocumentTextChangedCore (id, text));
+		}
 
-		protected override async void ApplyDocumentTextChanged (DocumentId id, SourceText text)
+		async Task ApplyDocumentTextChangedCore (DocumentId id, SourceText text)
 		{
 			var document = GetDocument (id);
 			if (document == null)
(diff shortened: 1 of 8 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -1080,6 +1090,71 @@ static int ApplyChanges (Projection projection, ITextDocument data, List<TextCha
...
+		HashSet<MonoDevelop.Projects.Project> tryApplyState_changedProjects = new HashSet<MonoDevelop.Projects.Project> ();
+		List<Task> tryApplyState_documentTextChangedTasks = new List<Task> ();
+		Dictionary<string, SourceText> tryApplyState_documentTextChangedContents =  new Dictionary<string, SourceText> ();
+
+		internal override bool TryApplyChanges (Solution newSolution, IProgressTracker progressTracker)
+		{
+			// this is supported on the main thread only
+			// see https://github.com/dotnet/roslyn/pull/18043
+			// as a result, we can assume that the things it calls are _also_ main thread only
+			Runtime.AssertMainThread ();
```
DavidKarlas: Can we postpone this change to after 15.2 branching? As it's own PR. There is a lot of code across monodevelop and md-addins invoking this after .ConfigureAwait(false); or directly from background thread that will be part of this PR...

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -63,7 +64,6 @@ public class MonoDevelopWorkspace : Workspace
 		readonly MonoDevelop.Projects.Solution monoDevelopSolution;
 		object addLock = new object();
 		bool added;
-		bool internalChanges;
```
DavidKarlas: @garuma Any idea what was this commit https://github.com/mono/monodevelop/commit/f7432a40a2f495e09d16e514a898a9862d446fc8 supposed to fix, so I can test if this change broken it?

mhutch: No idea what it was fixing, but it _broke_ stuff, as files added/remove in Apply* weren't being tracked by our Roslyn workspace.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
mhutch: Is there? ApplyChanges has relatively few callers. I think it only really gets called from code fixes.

DavidKarlas: https://github.com/mono/monodevelop/search?utf8=✓&q=TryApplyChanges&type=
CodeDependency.cs is called from 4 places in md-addins
CodeGenerationService.cs has .ConfigureAwait(false); right above call
There are few more, I'm not saying this is not possible to fix, I'm just saying it would be nicer to postpone this to 15.3 since 15.2 is branching soon...

mhutch: Well, this fixes broken code actions, so... do we ship with those as known issues?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
garuma: When you tried to apply a modification to the Roslyn workspace it duplicated the content of that file. That change fixed it.

garuma: As the commit message said, the cause was that using the Roslyn TryApplyChanges method would trigger MonoDevelop project file events that would re-enter that code causing the duplication.

DavidKarlas: but what was usecase? I tried inserting event in iOS Designer and other stuff, but couldnt reproduce... was there specific code path?

mhutch: @garuma I think that could only happen if something were calling TryApplyChanges incorrectly though

garuma: As far as I could tell I was calling it the way it's supposed to work. What would be "calling TryApplyChanges incorrectly" ?

PR #2029, opened 2017-03-27
Fixed 'Bug 54010 - NullReference in SearchAndReplaceWidget kills XS'.

mkrueger:
Not exactly sure how that bug happened. The search entry threw an
exception before that search & replace widget lock. Remember: always
include full logs when reporting bugs.
Catched an exception in search & replace widget - that event handler
should never crash the IDE.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/SearchEntry.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/SearchEntry.cs
@@ -374,16 +374,17 @@ namespace MonoDevelop.Components
 		}
 
 		event EventHandler requestMenu;
+		object requestMenuLock = new object ();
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components/SearchEntry.cs
```diff
@@ -374,16 +374,17 @@ protected virtual void OnRequestMenu (EventArgs e)
 		}
 
 		event EventHandler requestMenu;
+		object requestMenuLock = new object ();
 		public event EventHandler RequestMenu {
 			add {
-				lock (requestMenu) {
+				lock (requestMenuLock) {
```
Therzok: Ouch, locking on multicast delegates should be a code analysis error.

PR #2031, opened 2017-03-27
[Highlighting] Implement :ascii: class

Therzok:
This adds support for :ascii: regex class, for 0-127 char values.
Fixes css and C++ syntax files which were triggering 'unknown unicode category : ^ascii'

cc @mkrueger: I'm not sure this is supposed to be just _printable_ ascii characters, which in turn would be 0x20 - 0x7E.


DavidKarlas:
I can still see this messages in log with latest master

PR #2034, opened 2017-03-27
Remove HttpRequest logic due to unintended side-effects

sgreenmsft:
Remove HttpRequest logic due to unintended side-effects, reverting back to TCP logic.  Add brief delay to allow server process to initialize.


dazhao-msft:
@sgreenmsft what is the unintended side-effects?

sgreenmsft:
For WebApi projects, it's possible for the launch URL to be invalid, and the service returns a 404.  That's treated by the HttpWebRequest as an exception, which was ignored and caused the mechanism to loop, repeatedly requesting the same invalid resource over and over again for two minutes.

PR #2037, opened 2017-03-28
Fix errors caused by UI elements access from background threads

DavidKarlas:
PathEntry constructor is not called from UI Thread so postpone contraction of UI elements until they are actually requested from UI thread

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PathBar.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PathBar.cs
@@ -73,17 +73,19 @@ namespace MonoDevelop.Components
 			set;
 		}
 
+		AtkCocoaHelper.AccessibilityElementProxy accessible;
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PathBar.cs
```diff
@@ -73,30 +73,30 @@ public class PathEntry
...
-
-		PathEntry ()
-		{
-			//Accessible = new AtkCocoaHelper.AccessibilityElementButtonProxy ();
-			Accessible = AccessibilityElementProxy.ButtonElementProxy ();
-			Accessible.Identifier = "Breadcrumb";
-			Accessible.PerformPress += OnPerformShowMenu;
+			get {
+				if (accessible == null) {
+					//Accessible = new AtkCocoaHelper.AccessibilityElementButtonProxy ();
```
iainx: We can lose this old comment

PR #2049, opened 2017-03-28
[Core] Fix build error when .NET Core project is referenced

mrward:
When a non .NET Core project references a .NET Core project the
build would fail with an error similar to:

Error MSB4019: The imported project
"~/Library/Caches/VisualStudio/7.0/MSBuild/8209_1/Sdks/Microsoft.NET.Sdk/Sdk/Sdk.props" was not found.

The MSBuildSDKsPath was not defined when building the project that
references the .NET Core project. Now a check is made of the
referenced projects and the MSBuildSDKsPath is set there is a
.NET Core project referenced.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteProjectBuilder.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteProjectBuilder.cs
@@ -72,6 +72,7 @@ namespace MonoDevelop.Projects.MSBuild
 		{
 			var builder = await LoadProject (projectFile, sdksPath).ConfigureAwait (false);
 			var pb = new RemoteProjectBuilder (projectFile, builder, this);
+			pb.SdksPath = sdksPath;
 			lock (remoteProjectBuilders) {
 				remoteProjectBuilders.Add (pb);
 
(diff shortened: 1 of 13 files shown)


slluis:
Maybe it is already the case, but the builder should be re-created if SDKs from references change.

slluis:
So maybe GetProjectBuilder could keep in a field the list of SDKs that were used to configure the builder, and check if that list changed before reusing the builder.

slluis:
Actually, it only needs to keep the MSBuildSDKsPath that was selected. If the new path differs, the builder can't be reused.

mrward:
Yeah this was not working if the references changed.

If you added a reference to a .NET Core project, from a .NET Framework project, then the solution would not compile since the sdks path was not set for the project builder. This is now working. Although I put the SdksPath on the remote builder. I guess it should be a field in the Project itself since we are doing that with other properties, such as lastBuildToolsVersion.

mrward:
Should I move the last used SdksPath to the [Project](https://github.com/mono/monodevelop/blob/ca265a7bc25079468bd2c97e48937ad4b41fe5a5/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs#L1333-L1336) class with the other fields? Or is it OK on the RemoteProjectBuilder?

slluis:
It's ok

PR #2053, opened 2017-03-29
Rename UnsubscibeAnalysisdocument -> UnsubscribeAnalysisDocument.

KirillOsenkov:
Remove dead code (parentSolution is never used).

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
@@ -170,7 +170,7 @@ namespace MonoDevelop.Ide.Gui
 				window.ViewContent.Project.Modified += HandleProjectModified;
 			window.ViewsChanged += HandleViewsChanged;
 			window.ViewContent.ContentNameChanged += delegate {
-				UnsubscibeAnalysisdocument ();
+				UnsubscribeAnalysisDocument ();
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
```diff
@@ -861,9 +861,6 @@ Task EnsureAnalysisDocumentIsOpen ()
 						var node = TypeSystemService.GetTypeSystemParserNode (Editor.MimeType, BuildAction.Compile);
 						if (Editor.MimeType == "text/x-csharp" || node?.Parser.CanGenerateAnalysisDocument (Editor.MimeType, BuildAction.Compile, new string[0]) == true) {
 							var newProject = Services.ProjectService.CreateDotNetProject ("C#");
-							var parentSolution = new Solution ();
-							parentSolution.RootFolder.Items.Add (newProject);
-							parentSolution.AddConfiguration ("Debug", true);
```
KirillOsenkov: Can you please confirm that this is indeed not needed? Or is there a reason I'm not seeing why this code is here?

Therzok: You're right. It seems 2 solutions are created for an adhoc project, one which is just not used at all, and the adhoc solution set below.

PR #2055, opened 2017-03-29
[UnitTests] Disabled some non working tests.

mkrueger:
These tests are no longer supported on that layer. TextChanging event
now works differently it's no longer synced.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.TextMate/TextMateCompletionTextEditorExtension.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.TextMate/TextMateCompletionTextEditorExtension.cs
@@ -117,6 +117,7 @@ namespace MonoDevelop.Ide.Editor.TextMate
 		{
 			if (inactive)
 				return await base.HandleCodeCompletionAsync (completionContext, triggerInfo, token);
+			Console.WriteLine ("mime type :" + Editor.MimeType);
 			if (Editor.MimeType == "text/plain")
 				return null;
 			if (!IdeApp.Preferences.EnableAutoCodeCompletion)
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.TextMate/TextMateCompletionTextEditorExtension.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.TextMate/TextMateCompletionTextEditorExtension.cs
@@ -117,7 +117,6 @@ namespace MonoDevelop.Ide.Editor.TextMate
 		{
 			if (inactive)
 				return await base.HandleCodeCompletionAsync (completionContext, triggerInfo, token);
-			Console.WriteLine ("mime type :" + Editor.MimeType);
 			if (Editor.MimeType == "text/plain")
 				return null;
 			if (!IdeApp.Preferences.EnableAutoCodeCompletion)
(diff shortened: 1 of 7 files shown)


slluis:
Can you add alternative tests for TextChanging?

mkrueger:
Added one - however it's now all based upon the MS data structures - IMO no need to test that on our level. Mid term our level will go away - we shouldn't maintain another editor API.

The other ones that fail I opened a bug 

https://bugzilla.xamarin.com/show_bug.cgi?id=54245

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.TextMate/TextMateCompletionTextEditorExtension.cs
```diff
@@ -117,6 +117,7 @@ static Task<IEnumerable<string>> GetAllWords(ITextSource source, int offset)
 		{
 			if (inactive)
 				return await base.HandleCodeCompletionAsync (completionContext, triggerInfo, token);
+			Console.WriteLine ("mime type :" + Editor.MimeType);
```
Therzok: Debug code

main/tests/UnitTests/MonoDevelop.Ide.Gui/CompletionListWindowTests.cs
```diff
@@ -466,6 +466,7 @@ public void TestAutoCompleteEmptyMatchOff ()
 		};
 
 		[Test]
+		[Ignore ("punctuation completion is now handled in backend")]
```
slluis: What does it mean? why can't this be tested?

main/tests/UnitTests/MonoDevelop.Ide.Editor/TextSourceTestBase.cs
```diff
@@ -94,6 +94,7 @@ public void TestCreateReader()
 		}
 
 		[Test]
+		[Ignore ("Broken - FIXME")]
```
slluis: Is there a bug reported for this? if not, there should be one before adding the Ignore.

main/tests/UnitTests/MonoDevelop.Ide.Editor/TextSourceTestBase.cs
```diff
@@ -103,6 +104,7 @@ public void TestCreateReaderAt()
...
```
slluis: Same as above.

main/tests/UnitTests/MonoDevelop.Ide.Editor/ReadonlyTextDocumentTestBase.cs
```diff
@@ -50,6 +50,7 @@ public void TestLineCount()
 
 
 		[Test]
+		[Ignore ("Broken - FIXME")]
```
slluis: Bug #?

main/tests/UnitTests/MonoDevelop.Ide.Gui/CompletionListWindowTests.cs
mkrueger: It's no longer handled. If a backend requires that it needs to implement it.

slluis: Should we remove the test then?

mkrueger: I did in a later commit - it's right they're useless to us.

PR #2067, opened 2017-03-29
Fix comment in winbuild.bat (# is not a valid comment in .bat files)

KirillOsenkov:
'#' is not working as a comment in windows .bat files, resulting in:
```
'#' is not recognized as an internal or external command,
operable program or batch file.
```


slluis:
# also doesn't work as a comment in markdown!

PR #2071, opened 2017-03-30
Bug 54156 - double clicking on error bubble in editor crashes

DavidKarlas:
This happens in case of ErrorPad being in AutoShow pad… Problem is that DockerFrame.UseWindowsForTopLevelFrames == false on Windows which results in ContainerWindow=null, hence exception

@iainx This was added for a11y to put title on window on Mac.. what will outcome of not setting this on Windows since it's not implemented as window...


DavidKarlas:
tnx, thought so

PR #2079, opened 2017-03-30
[Mac] Make the status bar offset bigger on ElCap/Sierra

iainx:
Fixes BXC #54275


slluis:
I'm wondering, if the problem is Sierra, why old builds of XS running in the same machine don't have this problem?

iainx:
I had changed how the progress bar frame was calculated in the earlier accessibility patch: https://github.com/iainx/monodevelop/commit/00a830eb7ee4d6ccedd1d83e5983ac3ec24bd955 (at the bottom, not sure how to link to an individual change)

iainx:
Is another relevant bug: https://bugzilla.xamarin.com/show_bug.cgi?id=47107

PR #2081, opened 2017-03-30
Pass --targetprofile:netcore compiler arg to dotnetcore apps

nosami:


nosami:
Fixes this
screen shot 2017-03-30 at 5 25 35 pm

PR #2082, opened 2017-03-30
[F#] Faster project reference parsing

nosami:
Was seeing operation cancelled messages from the compiler when
passing in the output file name along with the source files
themselves for project references.

The key just needs to be unique for each project.. and as we have the
source code, there is no need for the compiler to parse the assemblies.

This results in _much_ faster time before intellisense etc. works


nosami:
Before
screen shot 2017-03-30 at 8 25 08 pm
After
screen shot 2017-03-30 at 8 33 39 pm

PR #2083, opened 2017-03-30
[TextEditor] Use the byte offset into the UTF8 string

bratsche:
PangoLayout's IndexToPos wants the byte offset into the
string and is storing data in UTF-8. Since we're currently
passing in the character offset the red wavy underlines are
misplaced when text before the error includes any multi-byte
characters.

Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=53000

### DIFF
--- a/main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/TextSegmentMarker.cs
+++ b/main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/TextSegmentMarker.cs
@@ -24,6 +24,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 using System;
+using System.Text;
 using Cairo;
 using Mono.TextEditor.Highlighting;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/TextSegmentMarker.cs
```diff
@@ -141,12 +142,17 @@ void InternalDraw (int markerStart, int markerEnd, MonoTextEditor editor, Cairo.
 			} else {
 				int start = startOffset < markerStart ? markerStart : startOffset;
 				int end = endOffset < markerEnd ? endOffset : markerEnd;
-				int /*lineNr,*/ x_pos;
+				int /*lineNr,*/ x_pos;
+
+				var substring = layout.Text.Substring (0, end - startOffset);
```
Therzok: We can probably get a `fixed char*` from the string and do a `GetByteCount (charPtr, end - startOffset)`

This is a hot path for rendering, we shouldn't allocate strings.

mkrueger: Damn UTF8 coordinates in pango. To fix that change layout to LineMetrics metrics. That contains the layout. 
Use metrics.Layout.TranslateToUTF8Index for coordinate transformation. curIndex/byteIndex should be 0 at the start. That method is built for forward reading btw.

bratsche:
I changed this to use `TranslateToUTF8Index()` as @mkrueger suggested.

PR #2089, opened 2017-03-31
[F#] Semantic highlighting fix

nosami:
In the following code

```fsharp
type o = System.Object
```

the word System is highlighted incorrectly. Fixed to highlight as a
namespace.

### DIFF
--- a/main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpSyntaxMode.fs
+++ b/main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpSyntaxMode.fs
@@ -292,7 +292,7 @@ module Patterns =
                 | GenericParameter _ ->
                     makeSeg EditorThemeColors.UserTypesTypeParameters
                 | Namespace _ ->
-                    makeSeg "source"
+                    makeSeg "source.fs"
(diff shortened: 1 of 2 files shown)


main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpSyntaxMode.fs
```diff
@@ -292,7 +292,7 @@ module Patterns =
                 | GenericParameter _ ->
                     makeSeg EditorThemeColors.UserTypesTypeParameters
                 | Namespace _ ->
-                    makeSeg "source"
+                    makeSeg "source.fs"
```
Therzok: What is source.fs in this case? Seems like the main scope - https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/syntaxes/FSharp/F%23.sublime-syntax#L7

nosami: It's the default text colour, we're not using any special rule for namespaces

PR #2096, opened 2017-03-31
[DotNetCore] Use NuGet to restore template packages

mrward:
MSBuild now just copies the .nupkg files to the Templates directory
instead of building a console app to download the NuGet packages.


mrward:
Going to verify this works on Wrench.

PR #2099, opened 2017-04-03
Bug 43827 - Namespaces of new files aren't set identical to Visual Studio

DavidKarlas:


DavidKarlas:
Going to reopen against d15-2

PR #2100, opened 2017-04-03
Bug 43827 - Namespaces of new files aren't set identical to Visual Studio

DavidKarlas:


DavidKarlas:
was w8ing your opinion on this before merging :)

slluis:
The default value of policy types should not change. The default value can now be set using a policy resource, in the same way this is done for default formatting options: https://github.com/mono/monodevelop/blob/d15-2/main/src/core/MonoDevelop.Ide/ExtensionModel/Policies.addin.xml#L16.

DavidKarlas:
@slluis can you check if this is what you meant?

PR #2111, opened 2017-04-04
[NuGet] Fix null reference on restoring packages

mrward:
If the solution contained .NET Core and non .NET Core projects
then the restore for the .NET Core projects would fail with a
null reference exception since the projects list was not populated.

### DIFF
--- a/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/RestoreAndCheckForUpdatesAction.cs
+++ b/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/RestoreAndCheckForUpdatesAction.cs
@@ -109,24 +109,21 @@ namespace MonoDevelop.PackageManagement
 					cancellationToken);
 
 				packagesToRestore = packages.ToList ();
-				if (packagesToRestore.Any (package => package.IsMissing)) {
-					return true;
-				}
 			}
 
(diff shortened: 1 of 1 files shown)


mrward:
Mmm, looks like this fixed one null ref but added a new one :(

main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/RestoreAndCheckForUpdatesAction.cs
```diff
@@ -109,24 +109,21 @@ bool AnyNuGetAwareProjects ()
...
 			}
 
 			if (nugetAwareRestorer != null) {
 				var projects = await nugetAwareRestorer.GetProjectsRequiringRestore (nugetAwareProjects);
 				nugetAwareProjectsToBeRestored = projects.ToList ();
-				return nugetAwareProjectsToBeRestored.Any ();
 			}
 
-			return false;
+			return packagesToRestore.Any (package => package.IsMissing) ||
```
Therzok: packagesToRestore can be null?

mrward: @Therzok Yep that is the bug this change introduces and I missed - already fixed in another pull request.

PR #2116, opened 2017-04-05
Bug 54423 - Code completion automatically closes parenthesis, instead…

DavidKarlas:
… of offering completions


slluis:
There is another fix for this bug proposed in https://github.com/mono/monodevelop/pull/2115

PR #2124, opened 2017-04-06
[Refactoring] Implemented quick fix menu

mkrueger:
We had that with nrefactory and it got removed because of speed
reasons. However now roslyn & refactoring essentials is much faster
and we support async command updates.
Time to bring back that feature.


slluis:
Wrong separators:
captura de pantalla 2017-04-06 a las 12 19 49

slluis:
When there isn't any code fix available, it should show a single disabled item "No code fixes available". Also, when loading the options it should show "Loading"

mkrueger:
Added 'Loading' & 'No code fixes available'

PR #2127, opened 2017-04-06
d

mkrueger:
It was a bom remove issue


slluis:
Hmm, but the bug is that the BOM is added, not that it is removed.

mkrueger:
It's only 50% - I made a test case for the bug - it's related to the way the VS code loads the files - the encoding detection is simply wrong in the VS code. If david can't fix that I remove the code that let VS handle the loading of the files - doesn't make that much sense IMO.

PR #2128, opened 2017-04-06
MSBuild 15 hosting changes.

KirillOsenkov:
This allows MonoDevelop on Windows to build its own source code without errors.

1. In MsNetTargetRuntime, set msbuildDir to the path from VS 2017 MSBuild bin folder.

2. In MSBuildProjectService, in LoadDefaultProjectImportSearchPaths on Windows use MSBuild.exe.config instead of MSBuild.dll.config.

3. When creating a new builder directory in \Cache\MSBuild, copy the original MSBuild.exe.config from MSBuild 15 and modify it to specify explicit paths for the following properties:
  MSBuildExtensionsPath, MSBuildExtensionsPath32, MSBuildToolsPath, MSBuildToolsPath32, MSBuildSDKsPath, RoslynTargetsPath, VCTargets

4. If on Windows, copy 8 MSBuild .dll files from VS 2017 MSBuild 15.0 Bin directory to the local builder directory. This is needed in particular for building WPF applications. The builder process has AppDomain.CurrentDomain.AssemblyResolve event handler where it specifies where to load the MSBuild binaries from. However MarkupCompiler from PresentationBuildTasks.dll creates new AppDomains and inside those AppDomains it fails to load Microsoft.Build.Framework.dll. I found no way on desktop CLR to specify AssemblyResolve for other AppDomains so the simplest solution is to copy the .dll locally.

5. Add a helper in ConfigFileUtilities.cs. This is in its own type so I can write a unit-test for it. If it's in MSBuildProjectService I can't unit-test because it has a pretty invasive static constructor that doesn't run in a unit-test environment. In the future we can move other helpers from MSBuildProjectService to this file.

6. In Main.cs for the builder process, no need to set MSBUILD_EXE_PATH. We will take the path of http://source.dot.net/#Microsoft.Build/SharedUtilities/BuildEnvironmentHelper.cs,331d766061bf9097

7. Add an assembly resolver for System.Collections.Immutable which is a new .dll needed by both Roslyn and MSBuild 15.

8. When on Windows, load assemblies into the Load context and not the LoadFrom context.


main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/MsNetTargetRuntime.cs
```diff
@@ -153,13 +158,13 @@ void AddPackages (TargetFramework fx, RegistryKey fxKey)
...
-			// Probe for Dev15 location and use MSBuild from there.
+			// Probe for Dev15 location and use MSBuild from there.
 			using (RegistryKey vsReg = Registry.LocalMachine.OpenSubKey (@"SOFTWARE\Microsoft\VisualStudio\SxS\VS7", false)) {
-				if (vsReg != null) {
-					string vsPath = (string)vsReg.GetValue ("15.0");
-					string path = Path.Combine (vsPath, "MSBuild", toolsVersion, "Bin");
-					if (File.Exists (Path.Combine (path, "MSBuild.exe"))) {
-						return path;
+				if (vsReg != null) {
+					string vsPath = (string)vsReg.GetValue (MSBuildVersion);
```
mhutch: Can't we just use `toolsVersion` here?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/ConfigFileUtilities.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Linq;
+using System.Xml.Linq;
+
+namespace MonoDevelop.Projects.MSBuild
+{
+	class ConfigFileUtilities
+	{
+		internal static void SetSubelementAttribute (XElement element, string subelementName, string attributeName, string attributeValue)
+		{
+			var appContextSwitchOverrides = element.Elements ().FirstOrDefault (e => e.Name.LocalName == subelementName);
```
mhutch: What's the significance of the name `appContextSwitchOverrides`?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1,30 +1,30 @@
-﻿// MSBuildProjectService.cs
```
mhutch: This diff is mostly whitespace changes, it's difficult to see what actually changed :(

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/Main.cs
```diff
@@ -96,13 +94,26 @@ Assembly MSBuildAssemblyResolver (object sender, ResolveEventArgs args)
...
+					// to load it only from there. If that fails, then let that exception
+					// escape
+					return AssemblyLoad (fullPath);
 				} else
 					return null;
+			}
+
+			public static Assembly AssemblyLoad (string asmPath)
+			{
+				if (Environment.OSVersion.Platform == PlatformID.Win32NT) {
```
mhutch: This should ideally be conditioned on runtime, not platform.

`var isDotNet = Type.GetType("Mono.Runtime") == null`

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/ConfigFileUtilities.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Linq;
+using System.Xml.Linq;
+
+namespace MonoDevelop.Projects.MSBuild
+{
+	class ConfigFileUtilities
+	{
+		internal static void SetSubelementAttribute (XElement element, string subelementName, string attributeName, string attributeValue)
```
mhutch: IMO this name is a little misleading considering that in some cases it appends the value to the existing value.

Maybe AppendSubelementAttributeValue ?

radical:
Could you minimize the whitespace changes, ideally restricted to the lines that have non-whitespace changes? Also, it would be useful to split this into smaller commits.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/ConfigFileUtilities.cs
KirillOsenkov: SetOrAppendSubelementAttributeValue?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/ConfigFileUtilities.cs
KirillOsenkov: Oops.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
KirillOsenkov: Sorry I've unified line endings to LF accidentally.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/Main.cs
KirillOsenkov: Fixed.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/ConfigFileUtilities.cs
KirillOsenkov: Renamed to SetOrAppendSubelementAttributeValue.

KirillOsenkov:
Apologies, fixed the whitespace and responded to all suggestions. I will squash when merging to get rid of temporary commits.

main/tests/UnitTests/UnitTests.csproj
```diff
@@ -89,7 +89,6 @@
     <ProjectReference Include="..\..\src\core\MonoDevelop.Core\MonoDevelop.Core.csproj">
       <Project>{7525BB88-6142-4A26-93B9-A30C6983390A}</Project>
       <Name>MonoDevelop.Core</Name>
-      <Private>False</Private>
```
Therzok: This shouldn't be changed.

KirillOsenkov: Why? Because of a potential double-write from MonoDevelop.TextEditor.Tests? The reason I enabled copy local on this reference is because without it, it only gets copied to \build\tests by chance (when building MonoDevelop.TextEditor.Tests). As such, editing MonoDevelop.Core and then running the UnitTests will result in stale (old) MonoDevelop.Core in the \build\tests directory because the project actually responsible for deploying the .dll doesn't know that it should rebuild.

By expressing this dependency explicitly we guarantee that whenever we change Core and then re-run UnitTests, the latest Core .dll will get deployed to the \build\tests directory where the tests will pick it up. At a slight risk of a file locked (two projects building simultaneously and trying to copy the file at the same time). However from experience the risk of race should be quite low.

Therzok: 👍 then

slluis: Core MD assemblies and add-in assemblies must never be local copied. Assemblies are loaded by the add-in engine. If there are two copies of  the same assembly in different places, things will break. We execute tests using mdtool (for example https://github.com/mono/monodevelop/blob/master/main/tests/UnitTests/UnitTests.csproj#L13), which takes care of initializing the add-in engine and loading the required assemblies. If MD.Core is being local copied to the tests directory, there is some unit test project doing it wrong.

KirillOsenkov: This is now reverted back to <Private>false</Private>.

KirillOsenkov:
Fixed, thanks for all the comments.
How is this looking now?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1235,55 +1236,89 @@ static string GetLocalMSBuildExeLocation (TargetRuntime runtime)
...
 
-			foreach (var configFile in Directory.GetFiles (Path.GetDirectoryName (GetLocalMSBuildExeLocation (runtime)), "*.config.original")) {
+			return localExe;
+		}
 
+		// Creates an MSBuild config file with the search paths registered by add-ins.
+		static void UpdateMSBuildExeConfigFile (TargetRuntime runtime, string binDir, string configFile)
+		{
+			// strip the ".original" suffix
 				var localConfigFile = configFile.Substring (0, configFile.Length - 9);
```
mhutch: Indentation of this and previous line is inconsistent

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1235,55 +1236,89 @@ static string GetLocalMSBuildExeLocation (TargetRuntime runtime)
...
-					string binDir;
-					GetNewestInstalledToolsVersion (runtime, true, out binDir);
+			if (Platform.IsWindows) {
+				configFile = Path.Combine (binDir, "MSBuild.exe.config");
+			}
 
-					// This is required for MSBuild to properly load the searchPaths element (@radical knows why)
-					SetMSBuildConfigProperty (toolset, "MSBuildBinPath", binDir, false, true);
+				var doc = XDocument.Load (configFile);
+			var configuration = doc.Root;
```
mhutch: Indentation of this and previous line is inconsistent

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1235,55 +1236,89 @@ static string GetLocalMSBuildExeLocation (TargetRuntime runtime)
...
+
+			var toolset = configuration.Elements ("msbuildToolsets").FirstOrDefault ()?.Elements ("toolset")?.FirstOrDefault ();
+			if (toolset != null) {
+				// This is required for MSBuild to properly load the searchPaths element (@radical knows why)
+				SetMSBuildConfigProperty (toolset, "MSBuildBinPath", binDir, false, true);
 
 					//this must match MSBuildBinPath w/MSBuild15
 					SetMSBuildConfigProperty (toolset, "MSBuildToolsPath", binDir, false, true);
 
+				var extensionsPath = Path.GetDirectoryName (Path.GetDirectoryName (binDir));
```
mhutch: Indentation of this and previous line is inconsistent

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1235,55 +1236,89 @@ static string GetLocalMSBuildExeLocation (TargetRuntime runtime)
...
+				var sdksPath = Path.Combine (extensionsPath, "Sdks");
+				SetMSBuildConfigProperty (toolset, "MSBuildSDKsPath", sdksPath);
+
+				var roslynTargetsPath = Path.Combine (binDir, "Roslyn");
+				SetMSBuildConfigProperty (toolset, "RoslynTargetsPath", roslynTargetsPath);
+
+				var vcTargetsPath = Path.Combine (extensionsPath, "Common7", "IDE", "VC", "VCTargets");
+				SetMSBuildConfigProperty (toolset, "VCTargetsPath", vcTargetsPath);
+
 					var projectImportSearchPaths = doc.Root.Elements ("msbuildToolsets").FirstOrDefault ()?.Elements ("toolset")?.FirstOrDefault ()?.Element ("projectImportSearchPaths");
```
mhutch: Indentation of this and previous line is inconsistent

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1235,55 +1236,89 @@ static string GetLocalMSBuildExeLocation (TargetRuntime runtime)
...
 
-				var doc = XDocument.Load (configFile);
-				var toolset = doc.Root.Elements ("msbuildToolsets").FirstOrDefault ()?.Elements ("toolset")?.FirstOrDefault ();
-
-				if (toolset != null) {
-					
-					string binDir;
-					GetNewestInstalledToolsVersion (runtime, true, out binDir);
+			if (Platform.IsWindows) {
+				configFile = Path.Combine (binDir, "MSBuild.exe.config");
```
mhutch: Does this mean we can drop the one we ship now?

main/tests/UnitTests/UnitTests.csproj
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
mhutch: Whitespace change?

main/tests/UnitTests/MonoDevelop.Projects/MSBuildTests.cs
```diff
@@ -1,45 +1,46 @@
-﻿// MSBuildTests.cs
```
mhutch: Whitespace change

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/Main.cs
```diff
@@ -44,9 +44,6 @@ class MainClass
 		[STAThread]
 		public static void Main (string[] args)
 		{
-			// This is required for MSBuild to properly load the .exe.config configuration file for this executable.
-			Environment.SetEnvironmentVariable ("MSBUILD_EXE_PATH", typeof(MainClass).Assembly.Location);
```
mhutch: @radical do the other changes mean we no longer need this?

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/MsNetTargetRuntime.cs
```diff
@@ -55,8 +55,11 @@ public MsNetTargetRuntime (bool running)
 			
 			string programFilesX86 = GetProgramFilesX86 ();
 			newFxDir = programFilesX86 + "\\Reference Assemblies\\Microsoft\\Framework";
-			msbuildDir = programFilesX86 + "\\MSBuild";
-			
+
+			msbuildDir = GetMSBuildBinPath ("15.0"); // C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\bin
```
mhutch: It's a shame this 15.0 is hardcoded here, but I guess it's a consequence of `GetMSBuildExtensionsPath` not taking a `toolsVersion` parameter. I think this is fine for now but it would be nice to clean up in future.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/ConfigFileUtilities.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Linq;
+using System.Xml.Linq;
+
+namespace MonoDevelop.Projects.MSBuild
+{
+	class ConfigFileUtilities
+	{
+		internal static void SetOrAppendSubelementAttributeValue (XElement element, string subelementName, string attributeName, string attributeValue)
```
radical: nit-pick: maybe `SetOrAppendSubelement*With*AttributeValue` ?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1235,55 +1236,89 @@ static string GetLocalMSBuildExeLocation (TargetRuntime runtime)
...
+				var doc = XDocument.Load (configFile);
+			var configuration = doc.Root;
+
+			var runtimeElement = configuration.Element ("runtime");
+			ConfigFileUtilities.SetOrAppendSubelementAttributeValue (runtimeElement, "AppContextSwitchOverrides", "value", "Switch.System.IO.UseLegacyPathHandling=false");
+
+			var toolset = configuration.Elements ("msbuildToolsets").FirstOrDefault ()?.Elements ("toolset")?.FirstOrDefault ();
+			if (toolset != null) {
+				// This is required for MSBuild to properly load the searchPaths element (@radical knows why)
+				SetMSBuildConfigProperty (toolset, "MSBuildBinPath", binDir, false, true);
```
radical: Maybe used argument names for the last two - `SetMSBuildConfigProperty (toolset, "MSBuildBinPath", binDir, append: false, insertBefore: true);`.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1235,55 +1236,89 @@ static string GetLocalMSBuildExeLocation (TargetRuntime runtime)
 
 			var dirId = Process.GetCurrentProcess ().Id.ToString () + "_" + runtime.InternalId;
 			var exesDir = UserProfile.Current.CacheDir.Combine ("MSBuild").Combine (dirId);
+			var originalExe = GetMSBuildExeLocationInBundle (runtime);
```
radical: IMHO, tracking these config files is very confusing. IIUC, this is getting the *builder* exe, hence the `localConfigOriginal` is for the builder and not msbuild itself. I would suggest updating the variable and method names to make this clear.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1235,55 +1236,89 @@ static string GetLocalMSBuildExeLocation (TargetRuntime runtime)
...
-			// Creates an MSBuild config file with the search paths registered by add-ins.
+				UpdateMSBuildExeConfigFile (runtime, binDir, localConfigOriginal);
+			}
 
-			foreach (var configFile in Directory.GetFiles (Path.GetDirectoryName (GetLocalMSBuildExeLocation (runtime)), "*.config.original")) {
+			return localExe;
+		}
 
+		// Creates an MSBuild config file with the search paths registered by add-ins.
+		static void UpdateMSBuildExeConfigFile (TargetRuntime runtime, string binDir, string configFile)
```
radical: `configFile` here, IIUC, is the config file that you want to *read/get* the updates from. Renaming this would be very useful.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
radical: Maybe move this out to the caller and take the "source" and "destination/new" config file paths, explicitly as parameters for this method. And update the method name to reflect that.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
radical: This should be on the caller side too. It will be easier to see what is being updated/combined using what.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1235,55 +1236,89 @@ static string GetLocalMSBuildExeLocation (TargetRuntime runtime)
...
+				configFile = Path.Combine (binDir, "MSBuild.exe.config");
+			}
 
-					// This is required for MSBuild to properly load the searchPaths element (@radical knows why)
-					SetMSBuildConfigProperty (toolset, "MSBuildBinPath", binDir, false, true);
+				var doc = XDocument.Load (configFile);
+			var configuration = doc.Root;
+
+			var runtimeElement = configuration.Element ("runtime");
+			ConfigFileUtilities.SetOrAppendSubelementAttributeValue (runtimeElement, "AppContextSwitchOverrides", "value", "Switch.System.IO.UseLegacyPathHandling=false");
```
radical: What is this for? Could you add a comment please? Is it required for non-windows platforms too?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1235,55 +1236,89 @@ static string GetLocalMSBuildExeLocation (TargetRuntime runtime)
...
+				SetMSBuildConfigProperty (toolset, "MSBuildToolsPath32", binDir);
+
+				var sdksPath = Path.Combine (extensionsPath, "Sdks");
+				SetMSBuildConfigProperty (toolset, "MSBuildSDKsPath", sdksPath);
+
+				var roslynTargetsPath = Path.Combine (binDir, "Roslyn");
+				SetMSBuildConfigProperty (toolset, "RoslynTargetsPath", roslynTargetsPath);
+
+				var vcTargetsPath = Path.Combine (extensionsPath, "Common7", "IDE", "VC", "VCTargets");
+				SetMSBuildConfigProperty (toolset, "VCTargetsPath", vcTargetsPath);
```
radical: Is this required for non-windows platforms?

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/Main.cs
radical: This should still be required since we are using our own config file and want `msbuild` to load that. @KirillOsenkov Does it work without this? On mac and windows?

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/Main.cs
```diff
@@ -96,13 +94,26 @@ Assembly MSBuildAssemblyResolver (object sender, ResolveEventArgs args)
...
+					return AssemblyLoad (fullPath);
 				} else
 					return null;
+			}
+
+			public static Assembly AssemblyLoad (string asmPath)
+			{
+				if (Type.GetType("Mono.Runtime") == null) {
+					// we need assemblies to be loaded in the Load context
+					// the AssemblyName on Win32 contains the full path information
```
radical: maybe `s,Win32,.NET runtime,` ?

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/Main.cs
```diff
@@ -96,13 +94,26 @@ Assembly MSBuildAssemblyResolver (object sender, ResolveEventArgs args)
 				}
 
 				string fullPath = Path.Combine (msbuildBinDir, asmName.Name + ".dll");
-				if (File.Exists (fullPath)) {
-					// If the file exists under the msbuild bin dir, then we need
-					// to load it only from there. If that fails, then let that exception
-					// escape
-					return Assembly.LoadFrom (fullPath);
+				if (File.Exists (fullPath)) {
+					// If the file exists under the msbuild bin dir, then we need
```
radical: whitespace changes?

main/tests/UnitTests/MonoDevelop.Projects/MSBuildTests.cs
```diff
@@ -2795,6 +2796,17 @@ public void FrameworkAssemblyVersionNotStored ()
 
 			p.Dispose ();
 		}
+
+		[Test]
+		public void TestMutateConfigFile ()
```
radical: This isn't testing mutating a config file. It is testing `SetOrAppendSubelementAttributeValue`.

main/tests/UnitTests/MonoDevelop.Projects/MSBuildTests.cs
```diff
@@ -1,45 +1,46 @@
...
-// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-// THE SOFTWARE.
-//
-//
-
+﻿// MSBuildTests.cs
+//
```
radical: whitespace changes?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
KirillOsenkov: Yes but I wanted to make this as a separate commit later in master. I'd like to get this in to d15-2 to unblock building on Windows, and the window keeps narrowing. Feels like the follow-up changes we can do in master later. Just deleting the .config file I'd be cautious about without proper testing.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/Main.cs
KirillOsenkov: Yes we now take the path of detecting MSBuild from TryFromMSBuildAssembly: http://source.dot.net/#Microsoft.Build/SharedUtilities/BuildEnvironmentHelper.cs,92

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/MsNetTargetRuntime.cs
KirillOsenkov: Yes we can clean up in master afterwards.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
KirillOsenkov: Fixed.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
KirillOsenkov: Fixed.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
KirillOsenkov: Fixed.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
KirillOsenkov: Fixed.

main/tests/UnitTests/UnitTests.csproj
KirillOsenkov: My changes accidentally added a BOM, I removed it.

main/tests/UnitTests/MonoDevelop.Projects/MSBuildTests.cs
KirillOsenkov: Sort usings used \r\n. Converted back to \n.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
KirillOsenkov: Fixed.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
KirillOsenkov: This entry was in the original "template" config file and it enables better path normalization: https://blogs.msdn.microsoft.com/jeremykuhne/2016/06/21/more-on-new-net-path-handling/. Since I don't know the history of why that entry is there (but it's a good thing to have), I made sure to preserve that.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
KirillOsenkov: I don't think so.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/Main.cs
KirillOsenkov: Fixed.

main/tests/UnitTests/MonoDevelop.Projects/MSBuildTests.cs
KirillOsenkov: Fixed.

main/tests/UnitTests/MonoDevelop.Projects/MSBuildTests.cs
KirillOsenkov: Fixed.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/Main.cs
KirillOsenkov: Fixed.

KirillOsenkov:
Note that this PR is now abandoned in favor of: https://github.com/mono/monodevelop/pull/2206

slluis:
So can this be closed?

PR #2129, opened 2017-04-07
Fixed 'Bug 54631 - First two characters in project file removed when

mkrueger:
adding file or folder'.

The encodings handle the boms. Stripping them before just strips the
first chars of the file.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Text/TextFileUtility.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Text/TextFileUtility.cs
@@ -161,17 +161,9 @@ namespace MonoDevelop.Core.Text
 			return encoding.GetString (bytes, start, bytes.Length - start);
 		}
 
+		[Obsolete("Use encoding.GetString")]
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Core.Text/TextFileUtility.cs
```diff
@@ -161,17 +161,9 @@ public static string GetText (byte[] bytes, out Encoding encoding)
 			return encoding.GetString (bytes, start, bytes.Length - start);
 		}
 
+		[Obsolete("Use encoding.GetString")]
 		public static string GetText (byte[] bytes, Encoding encoding)
```
slluis: If it is obsolete, please remove the method.

mkrueger: I thought we're in an API stability phase ?

slluis: We'll be in API stability phase when we release this to stable.

PR #2136, opened 2017-04-07
[MSBuild] Fix item metadata evaluation issue

slluis:
Item metadata was not correctly evaluated since the item context was not
always set when evaluating an item.


slluis:
Not critical for 15.2, I'll merge to master.

Therzok:
Retargeted to master, should be free to merge.

PR #2146, opened 2017-04-11
Fixed 'Bug 54990 - Default C# Code Formatting options seem

mkrueger:
unreasonable'

### DIFF
--- a/main/src/core/MonoDevelop.Ide/ExtensionModel/Policies.addin.xml
+++ b/main/src/core/MonoDevelop.Ide/ExtensionModel/Policies.addin.xml
@@ -13,7 +13,7 @@
 	<Policies resource="MonoTextStylePolicy.xml" />
 </Extension>
 <Extension path = "/MonoDevelop/ProjectModel/PolicySets/Default" >
-	<Policies resource="MonoTextStylePolicy.xml" />
+	<Policies resource="VisualStudioTextStylePolicy.xml" />
 </Extension>
 <Extension path = "/MonoDevelop/ProjectModel/PolicySets/VisualStudio" >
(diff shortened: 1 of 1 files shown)


DavidKarlas:
I agree with this, but lets wait for Lluis to approve it

slluis:
The 8 space tab was explicitly requested by miguel.

migueldeicaza:
We have a bit of an host operating system problem here.

Unix editors, terminals and other tools default to interpreting the tab character as stopping at the next column which is a multiple of 8.

Changing the tab size to 4 is generally problematic for interop purposes.   Because while you might be able to configure your text editor, the Unix console and every other Unix tool has baked in the knowledge that tabs are 8 spaces so what this means is that code formatted with editors that change this default produce unreadable text on the Console and in other editors.

For a good discussion on the subject see this:

https://www.jwz.org/doc/tabs-vs-spaces.html

This can be summarized as "Avoid encoding the TAB character into files to avoid interop problems".   Which seems to be the most sensible thing to do.   I am not sure if this is something we want to tackle now, but it is the only reasonable fix for what looks like a complete mess when you are not using editors that have been configured to a different setting than 8 (which is what Unix defaults to).

In the short-term, what we can do is make tabs 4 spaces, for the sake of interop with Visual Studio on Windows for the Visual Studio style.  And we keep the other styles at 8.

slluis:
@migueldeicaza we define two formatting styles. The **Mono** style uses 8 space wide tabs for indenting. The **Visual Studio** style uses 4 spaces for indenting (it doesn't use tab chars). So we already have both options. The question is, which style should VS for Mac use by default?

migueldeicaza:
Let us use the VS style for VS For Mac by default.

slluis:
So be it!

PR #2158, opened 2017-04-11
[CSharpBinding] add InternalsVisibleTo for MonoDevelop.Unity

jbevain:


mhutch:
Are you going to public sign with the same public key as MonoDevelop?

KirillOsenkov:
If you need to specify IVT to a different key, here are some tips.

The logic to generate the `obj\DebugWin32\GeneratedInternalsVisibleTo.cs` files is in the `GenerateInternalsVisibleToFile` target here:
https://github.com/mono/monodevelop/blob/master/main/msbuild/GenerateInternalsVisibleTo.targets#L22-L33

I think if you specify 
```
<InternalsVisibleToAttribute Include="System.Runtime.CompilerServices.InternalsVisibleToAttribute">
  <_Parameter1>UnitTests, PublicKey=002400000c800000940000000602000000240000525341310004000001000100e1290d741888d13312c0cd1f72bb843236573c80158a286f11bb98de5ee8acc3142c9c97b472684e521ae45125d7414558f2e70ac56504f3e8fe80830da2cdb1cda8504e8d196150d05a214609234694ec0ebf4b37fc7537e09d877c3e65000f7467fa3adb6e62c82b10ada1af4a83651556c7d949959817fed97480839dd39b
</Parameter1>
</InternalsVisibleToAttribute>
```

Then it might work.

jbevain:
@mhutch, I don't think we'll use MD's key, so we need to look into that. Thanks!

jbevain:
@KirillOsenkov that looks fragile :)

The alternative would be to simply add an attribute in CSharpBinding's AssemblyInfo. @mhutch would that work for you?

mhutch:
Can we just add a supported metadata key for this?

I'd like to keep all the IVTs in one place rather than scattering then across multiple files.

jbevain:
@mhutch sure we can, I was just assuming that the less infrastructure changes in d15-2 the better. I'll look into it.

mhutch:
@jbevain try https://github.com/mhutch/monodevelop/commit/61931252693f27ed447e8e9843a3d2493ec55480

mhutch:
https://github.com/mono/monodevelop/pull/2169

jbevain:
@mhutch looking good thanks! Will your PR it against d15-2? I'll rebase my changes on top of yours then.

mhutch:
@jbevain https://github.com/mono/monodevelop/pull/2171

jbevain:
@mhutch I've updated this PR accordingly, thanks!

jbevain:
@mhutch hey, anything missing to get #2171 and this merged in d15-2?

mhutch:
Nope, you should be good!

PR #2165, opened 2017-04-12
Bug 54651 - Cannot debug .NET Core test project

DavidKarlas:


DavidKarlas:
need to change few more things to be more robust

PR #2176, opened 2017-04-13
[Ide] Properly handle mnemonics in chinese/korean languages

Therzok:
Bug 53081 - Chinese translations include the mnemonics key


Therzok:
@slluis updated, handles more cases now.

PR #2186, opened 2017-04-13
[Core] use 64bit if 32bit is not preferred/required

alanmcgovern:
This seems like a saner way of approaching the problem of
'should Xamarin Studio start 64bit or 32bit'. The PE header
has a clear set of values we can base our decision on.
* Is the assembly compiled for x86? Use 32bit
* Is the assembly compiled for x64? Use 64bit
* Is the assembly platform agnostic and prefers 32bit? Use 32bit
* Is the assembly platform agnostic and *does not* prefer 32bit? Use 64bit

This change implements the above logic, which will allow us to easily
choose the correct bitness of the process Xamarin Studio starts by
setting the correct set of flags on our binary.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/MonoTargetRuntime.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/MonoTargetRuntime.cs
@@ -351,7 +351,7 @@ namespace MonoDevelop.Core.Assemblies
 				monoPath = Path.Combine (MonoRuntimeInfo.Prefix, "bin", "mono32");
 				if (File.Exists (monoPath))
 					return monoPath;
-			} else if ((peKind & IKVM.Reflection.PortableExecutableKinds.PE32Plus) != 0) {
+			} else {
 				monoPath = Path.Combine (MonoRuntimeInfo.Prefix, "bin", "mono64");
 				if (File.Exists (monoPath))
 					return monoPath;


Therzok:
This LGTM

jbevain:
Side note: using either Cecil or IKVM.Reflection to read this flag feels overkill.

mhutch:
@jbevain there isn't really a better alternative, and we only do it when launching processes

mhutch:
Small correction - `mono` isn't multiarch, it's a symlink to `mono32` which is switching over to `mono64` soon.

Nonetheless, it seems reasonable to let Mono handle the arch-independent default. If your binary is arch-independent, why does it matter which runtime it runs on?

If you need 64-bit, you can _already_ mark your exe as 64bit, like we do for VSMac itself.

alanmcgovern:
This is the sanest way of handling apps referencing Xamarin.Mac.dll, which is platform specific. My argument is that the flags do specify a clear preference and we are not quite following the preference.

What we want is to create a test runner assembly which prefers 64bit but runs under any bitness. We've successfully done that but unfortunately XS won't DTRT. 

One workaround to both mono and XS not respecting the preference is to locally run corflags (which means shipping an equivalent to be used as part of the build, or compiling one as part of the build) so we can change the bitness of the testrunner executable to force it to 64bit only (urgh). 

Or we could create two nugets, one for 64bit and one for 32bit. Due to other bugs/limitations we can't ship a nuget with a 33bit only version and 64bit only version and still have XS be able to load/run the tests in-IDE.

mhutch:
There is no way in .NET to specify a preference for 64bit, only a requirement.

alanmcgovern:
There is a way to say you *prefer* 32bit. If you do not specify that you *prefer* 32bit then 64bit should be used. Either way I believe this is a reasonable interpetation of the 2 flags involved, which (afaik) matches how MS.NET interprets the flags:

```
Is the assembly compiled for x86? Use 32bit
Is the assembly compiled for x64? Use 64bit
Is the assembly platform agnostic and prefers 32bit? Use 32bit
Is the assembly platform agnostic and does not prefer 32bit? Use 64bit
```

/cc @garuma as he was looking into that the other day and might be able to confirm the behaviour there.

mhutch:
IIRC it was documented for .NET that, when none of these flags are set, it uses the default runtime, which happens to be 64bit on 64bit machines. For Mono on Mac the default runtime is currently 32bit, though that it changing.

Anyways, I'm not averse to this change, I'd just like the commit message to be accurate. All it's doing is overriding the system default runtime when no flags are set.

mhutch:
I think it might also be good to structure the `if` to have a explicit 64-bit and ILOnly codepaths to make the intent clear. Maybe even log a warning if the assembly requires 64bit and mono64 isn't present (we can't error out, since it possible that `mono` is 64bit and `mono64` is not present).

alanmcgovern:
upstream mono has already changed the symlink so we can ignore this. It's just been a horrible annoyance for a year. We need a 32bit symlink for 1/2 our products and a 64bit symlink for the other 1/2 because everyone likes to rely on the system default being 'correct' :(

PR #2193, opened 2017-04-14
[Ide] Set the focus to the last item in the container

Therzok:
Bug 54670 - focus improperly set to toolbar on changing documents


Therzok:
While I still don't like the fix, this works for now to fix the bug and doesn't regress editor handling. We should be able to get the widget to focus from the DockNotebookTab itself or any DockNotebookTab content should only contain 1 widget which should handle focus itself.

PR #2201, opened 2017-04-17
https://bugzilla.xamarin.com/show_bug.cgi?id=53033

ToddGrun:
The windows save-as dialog was not able to select the UTF8 codepage, even when the OpenFileDialogHandler.BuildEncodingsCombo took in the UTF8 encoding. UTF8 and UTF16 have the same WindowsCodePage (1200) which happens to match the UTF16 CodePage but not the UTF8 CodePage (65001)


Therzok:
:+1:

PR #2204, opened 2017-04-17
[Core] Remove custom framework API

mhutch:
It's not useful to be able to define custom frameworks
inside MD, only MSBuild frameworks actually work now.


main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/TargetFramework.cs
```diff
@@ -79,10 +68,6 @@ internal TargetFramework (TargetFrameworkMoniker id)
 			Assemblies = new AssemblyInfo[0];
 		}
 		
-		public bool Hidden {
```
slluis: Just for the sake of not breaking the API, can you make it just return true?

mhutch: or false :)

slluis: Yeah, false :)

mhutch:
@slluis i addressed the feedback

mhutch:
@slluis updated!

mhutch:
@slluis rebased again

PR #2206, opened 2017-04-17
Set MSBuild properties on Windows builds

KirillOsenkov:
Edit the .config file on Windows to contain correct values for MSBuild properties. These are sufficient to be able to build MonoDevelop on Windows using itself.

On Windows we take the original MSBuild.exe.config file as a starting point and modify it to set the properties we need.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
@@ -258,7 +258,8 @@ namespace MonoDevelop.Projects.MSBuild
 				string binDir;
 				GetNewestInstalledToolsVersion (runtime, true, out binDir);
 
-				var configFile = Path.Combine (binDir, "MSBuild.dll.config");
+				var configFileName = Platform.IsWindows ? "MSBuild.exe.config" : "MSBuild.dll.config";
+				var configFile = Path.Combine (binDir, configFileName);
 				if (File.Exists (configFile)) {
 					var doc = XDocument.Load (configFile);
 					var projectImportSearchPaths = doc.Root.Elements ("msbuildToolsets").FirstOrDefault ()?.Elements ("toolset")?.FirstOrDefault ()?.Element ("projectImportSearchPaths");
@@ -1243,6 +1244,12 @@ namespace MonoDevelop.Projects.MSBuild
 			string binDir;
 			GetNewestInstalledToolsVersion (runtime, true, out binDir);
 
+			if (Platform.IsWindows) {
+				// on Windows copy the official MSBuild.exe.config from the VS 2017 install
+				// and use this as the starting point
+				originalExeConfig = Path.Combine (binDir, "MSBuild.exe.config");
+			}
+
 			if (!Directory.Exists (exesDir)) {
 				// Copy the builder to the local dir, including the debug file and config file.
 				Directory.CreateDirectory (exesDir);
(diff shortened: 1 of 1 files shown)


KirillOsenkov:
Note that this PR has two commits. The first one has its own PR:
https://github.com/mono/monodevelop/pull/2205

If that one is pushed first I can re-send the second commit separately so its easier to review.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1169,11 +1170,11 @@ static string GetNewestInstalledToolsVersion (TargetRuntime runtime, bool requir
 			dictionary.Add ("SolutionPath", Path.GetFullPath (slnFile));
 			dictionary.Add ("SolutionName", Path.GetFileNameWithoutExtension (slnFile));
 			dictionary.Add ("SolutionFilename", Path.GetFileName (slnFile));
-			dictionary.Add ("SolutionDir", Path.GetDirectoryName (slnFile) + Path.DirectorySeparatorChar);
```
mhutch: Whitespace changes

KirillOsenkov: These are by design. The only CRLF lines in the entire file. Thought to sneak this in.

mhutch: Hah, ok. They're probably my fault in the first place anyway then.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1235,73 +1235,82 @@ static string GetLocalMSBuildExeLocation (TargetRuntime runtime)
...
+					var os = Platform.IsMac ? "osx" : Platform.IsWindows ? "windows" : "unix";
+					XElement searchPaths = projectImportSearchPaths.Elements ("searchPaths").FirstOrDefault (sp => sp.Attribute ("os")?.Value == os);
+					if (searchPaths == null) {
+						searchPaths = new XElement ("searchPaths");
+						searchPaths.SetAttributeValue ("os", os);
+						projectImportSearchPaths.Add (searchPaths);
 					}
-					doc.Save (localConfigFile);
+					foreach (var path in GetProjectImportSearchPaths (runtime, false))
+						SetMSBuildConfigProperty (searchPaths, path.Property, path.Path, true, false);
```
radical: `..., append:true, insertBefore:false` .

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1235,73 +1235,82 @@ static string GetLocalMSBuildExeLocation (TargetRuntime runtime)
...
+							File.Copy (dll, destination);
+					}
+				}
 
 				searchPathConfigNeedsUpdate = true;
 			}
 
 			if (searchPathConfigNeedsUpdate) {
 				// There is already a local copy of the builder, but the config file needs to be updated.
 				searchPathConfigNeedsUpdate = false;
```
radical: I wonder whether it might be better to set this to `false` after the call to `UpdateMSBuildExeConfigFile`. I don't know the context in which this gets set, so I'm not sure if it will break something. But it feels more correct :) Somebody more familiar with the code can comment better. Just a thought!

KirillOsenkov: It all depends on the threading guarantees around this code.

On one side, if there's a risk that multiple threads can enter this path, setting the flag to false before invocation makes more sense because the other thread won't enter the Update method again.

On the other hand I do agree with you that setting the flag to false afterwards has the benefit that it's only set to false if the method exits cleanly. If it throws and fails to update the flag should remain true.

Either way more thought has to go in to do it right (both threading/locking and exception handling). Since it's not related to my PR I'd avoid changing anything here for d15-2.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
KirillOsenkov: Fixed.

PR #2207, opened 2017-04-18
Fixed 'Bug 53380 - [Webtools] Editor inserts BOMs sometimes'

mkrueger:
It's a fairly complex bug to fix. The behavior of VS.NET and XS
clearly differs. VS just uses the system default locale to dertmine
encoding. That's Encoding.Default on mono that's always UTF8 with bom.
So the vs loading routines assume that ascii files are always UTF8+BOM
on mono. Unfortunately we need to be able to consume the locales and
sometimes it clashes with UTF8 - the text file utility contains logic
to prevent these clashes. CP1252 files would be converted to UTF8
otherwise - it's not just a BOM bug - it's more an encoding
detection+incompatibilty bug - mosly caused by a bad design decision
from VS.NET.

Only way to fix that is using the text file utililty logic - it has
work arounded that issue since years.

### DIFF
--- a/main/src/core/Mono.TextEditor.Platform/Platform/Text/Impl/TextModel/TextDocumentFactoryService.cs
+++ b/main/src/core/Mono.TextEditor.Platform/Platform/Text/Impl/TextModel/TextDocumentFactoryService.cs
@@ -31,9 +31,9 @@ namespace Microsoft.VisualStudio.Text.Implementation
         [Import]
         internal GuardedOperations _guardedOperations { get; set; }
 
-        #endregion
+		#endregion
 
-        internal static Encoding DefaultEncoding = Encoding.Default; // Exposed for unit tests.
+		internal static Encoding DefaultEncoding = new UTF8Encoding (encoderShouldEmitUTF8Identifier: false); // Exposed for unit tests.
 
         #region ITextDocumentFactoryService Members
 
--- a/main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Document/TextDocument.cs
+++ b/main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Document/TextDocument.cs
@@ -282,11 +282,16 @@ namespace Mono.TextEditor
 				foldedSegments.Remove (e.Node);
 		}
 
-		public TextDocument(string fileName, string mimeType)
+		public TextDocument (string fileName, string mimeType)
 		{
-			var contentType = GetContentTypeFromMimeType(mimeType);
+			var contentType = GetContentTypeFromMimeType (mimeType);
+			Encoding enc;
+			var text = TextFileUtility.GetText (fileName, out enc);
+			var buffer = PlatformCatalog.Instance.TextBufferFactoryService.CreateTextBuffer (text ?? string.Empty,
+			                                                                                 PlatformCatalog.Instance.TextBufferFactoryService.InertContentType);
 
-			this.VsTextDocument = PlatformCatalog.Instance.TextDocumentFactoryService.CreateAndLoadTextDocument(fileName, contentType ?? PlatformCatalog.Instance.ContentTypeRegistryService.UnknownContentType);
+			this.VsTextDocument = PlatformCatalog.Instance.TextDocumentFactoryService.CreateTextDocument (buffer, string.Empty);
+			this.VsTextDocument.Encoding = enc;
 
 			this.Initialize();
 		}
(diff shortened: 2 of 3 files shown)


main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Document/TextDocument.cs
```diff
@@ -282,11 +282,16 @@ void HandleFoldSegmentTreetreeNodeRemoved (object sender, RedBlackTree<FoldSegme
...
-			var contentType = GetContentTypeFromMimeType(mimeType);
-
-			this.VsTextDocument = PlatformCatalog.Instance.TextDocumentFactoryService.CreateAndLoadTextDocument(fileName, contentType ?? PlatformCatalog.Instance.ContentTypeRegistryService.UnknownContentType);
+			var contentType = GetContentTypeFromMimeType (mimeType);
+			Encoding enc;
+			var text = TextFileUtility.GetText (fileName, out enc);
+			var buffer = PlatformCatalog.Instance.TextBufferFactoryService.CreateTextBuffer (text ?? string.Empty,
+			                                                                                 PlatformCatalog.Instance.TextBufferFactoryService.InertContentType);
+
+			this.VsTextDocument = PlatformCatalog.Instance.TextDocumentFactoryService.CreateTextDocument (buffer, string.Empty);
```
Therzok: Shouldn't we be passing the fileName here? The old version did.

KirillOsenkov: Yes, the text document needs to be associated with the file.

Also why are you not using CreateAndLoadTextDocument? It has smarts about using the right storage for large files. If you just want to detect the right encoding consider using IEncodingDetector or providing your own:
https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.text.iencodingdetector.aspx

main/src/core/MonoDevelop.TextEditor.Tests/Mono.TextEditor.Tests/DocumentTests.cs
```diff
@@ -28,6 +28,8 @@
 
 using System;
 using NUnit.Framework;
+using System.IO;
+using System.Text;
```
KirillOsenkov: Consider sorting the usings here.

main/src/core/MonoDevelop.TextEditor.Tests/Mono.TextEditor.Tests/DocumentTests.cs
```diff
@@ -176,5 +178,19 @@ public void TestBufferCreationIssue()
...
+		/// </summary>
+		[Test]
+		public void TestBug53380 ()
+		{
+			var path = Path.GetTempFileName ();
+			File.WriteAllText (path, "Hello World", Encoding.ASCII);
+			var document = new TextDocument (path, "text");
+
+			Assert.AreEqual (0, document.Encoding.GetPreamble ().Length);
+			File.Delete (path);
```
KirillOsenkov: Consider deleting the file in a finally block.

main/src/core/Mono.TextEditor.Platform/Platform/Text/Impl/TextModel/TextDocumentFactoryService.cs
```diff
@@ -29,11 +29,11 @@ internal sealed partial class TextDocumentFactoryService : ITextDocumentFactoryS
...
-        internal GuardedOperations _guardedOperations { get; set; }
-
-        #endregion
-
-        internal static Encoding DefaultEncoding = Encoding.Default; // Exposed for unit tests.
+        internal GuardedOperations _guardedOperations { get; set; }
+
+		#endregion
+
+		internal static Encoding DefaultEncoding = new UTF8Encoding (encoderShouldEmitUTF8Identifier: false); // Exposed for unit tests.
```
KirillOsenkov: My understanding is that this change alone should be sufficient to fix the bug. Curious whether we really need to change how we load text documents.

mkrueger: It's enough to fix that bug but not enough to load randomly encoded documents which will not load with UTF8 encoding.

main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Document/TextDocument.cs
mkrueger: We need the text anyways - large file optimization is nonsense for us atm since the whole text is requested after load so we don't save anything here.

That's the fastest way of fixing the encoding problems. Using a new infrastructure is a mid term solution. 

I fixed the file name setting - didn't know that the vs documents controls the file name as well.

main/src/core/Mono.TextEditor.Platform/Platform/Text/Impl/TextModel/TextDocumentFactoryService.cs
mkrueger: btw. using the vs encoding detection routines is a good hint - but using the tfu is the fastest way. Large file optimization is atm nonsense since the whole text is requested after loading - so it'll make loading in fact slower.

(And our editor hasn't become faster using the vs document - in fact we already lost speed)

However if we fix that the whole text is required it'll make sense to revisit that - for now using the text file utility is the fastest way fixing the issue.

KirillOsenkov:
LGTM!

PR #2210, opened 2017-04-18
Add default fallback path for Xcode.app

Therzok:
Bug 53324 - Svn is not working on latest version of Xamarin Studio 6.2

cc @mhutch


Therzok:
A `dir` call with `enoent` check could be done if needed.

jstedfast:
Looks like a reasonable approach

PR #2212, opened 2017-04-18
[Icons][Ide] .NET Core platform icons

vancura:
New icons for the .NET Core projects.

layer-1

untitled-1


mrward:
Do we want to make any changes to the .NET Standard project icon/image? This project template will create a Sdk style project and will target .NET Standard instead of .NET Core, so presumably it should not have a .NET Core icon, but maybe it should have a multiplatform image.

newprojectnetstandardlibrary

Currently it uses the same image as the .NET Framework class library project.

vancura:
Indeed, that's a great point. I'll take a look at it.

PR #2216, opened 2017-04-19
Fixed 'Bug 55298 - Autocomplete () doesn't work'

mkrueger:
Different approach to solve the problem - now the constructor is
preselected. Unfortunately it's not possible to do the former approach
- there are some use cases for a bigger list. Analyzing these cases is
IMO not 100% possible - that patch allows fluent typing flow.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionListWindow.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionListWindow.cs
@@ -735,9 +735,13 @@ namespace MonoDevelop.Ide.CodeCompletion
 			//sort, sinking obsolete items to the bottoms
 			//the string comparison is ordinal as that makes it an order of magnitude faster, which 
 			//which makes completion triggering noticeably more responsive
-			if (!completionDataList.IsSorted)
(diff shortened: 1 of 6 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionListWindow.cs
```diff
@@ -735,9 +735,13 @@ bool FillList ()
 			//sort, sinking obsolete items to the bottoms
 			//the string comparison is ordinal as that makes it an order of magnitude faster, which 
 			//which makes completion triggering noticeably more responsive
-			if (!completionDataList.IsSorted)
-				completionDataList.Sort (ListWidget.GetComparerForCompletionList (completionDataList));
-
+			if (!completionDataList.IsSorted) {
+				try {
+					completionDataList.Sort (ListWidget.GetComparerForCompletionList (completionDataList));
```
mkrueger: Shouldn't usually happen - but I once accidently inserted identical items in the list and the UI shouldn't crash on list sorting exceptions.

slluis: Next time add this comment in the code. It will be lost if you add it to the PR.

PR #2217, opened 2017-04-19
Fixed 'Bug 55329 - [regression] IntelliSense not working with

mkrueger:
excluding conditional ItemGroup'

This patch prevents the document in beeing in state Project != null &&
AnalysisDocument == null.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
@@ -178,9 +178,8 @@ namespace MonoDevelop.Ide.Gui
 
 		void TypeSystemService_WorkspaceItemLoaded (object sender, EventArgs e)
 		{
-			if (IsAdHocProject)
-				return;
 			UnsubscribeAnalysisDocument ();
+			UnloadAdhocProject ();
 			EnsureAnalysisDocumentIsOpen ().ContinueWith (delegate {
 				if (analysisDocument != null)
 					StartReparseThread ();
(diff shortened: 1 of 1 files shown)


Therzok:
Best seen in https://github.com/mono/monodevelop/pull/2217/files?w=1

slluis:
What will happen if a file is opened while the type system is still being loaded and there is yet no type system document for the file? won't then the file be assigned an ad-hoc project? will it get the real roslyn project when type information is finally loaded for the file?

mkrueger:
Worked for me in the monodevelop solution. However it may be because the documents are created very quickly but the parsing/loading lags behind. Anyways that patch fixed it.

During reload files could be included in the project via item group condition change.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
```diff
@@ -846,54 +845,56 @@ Task EnsureAnalysisDocumentIsOpen ()
...
 				RoslynWorkspace = TypeSystemService.GetWorkspace (this.Project.ParentSolution);
+				if (RoslynWorkspace == null) // Solution not loaded yet
+					return Task.CompletedTask;
 				SubscribeRoslynWorkspace ();
 				analysisDocument = FileName != null ? TypeSystemService.GetDocumentId (this.Project, this.FileName) : null;
 				if (analysisDocument != null) {
 					TypeSystemService.InformDocumentOpen (analysisDocument, Editor);
+					return Task.CompletedTask;
 				}
-			} else {
```
Therzok: Either the removal of the else clause without adding a return Task.CompletedTask in the previous if condition or the change from `IsAdHocProject` to `UnloadAdhocProject` turned this into an infinite loop.

PR #2220, opened 2017-04-19
Fixes #55309

nosami:


slluis:
Not necessary since we are merging d15-2 to master.

PR #2221, opened 2017-04-19
[Loc] Allow native file dialogs to be localized using the locale

Therzok:
Unless we explicitly set these keys, NSOpenPanel and NSSavePanel do not get localized.

ಠ_ಠ cocoa


Therzok:
Yeah, you need to specify that the app supports these locales so it tries to get the file dialogs up.

PR #2228, opened 2017-04-20
Load perf

Therzok:
@slluis this improves project load performance by a lot. Do we want this in 15.2?

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -1409,7 +1409,7 @@ namespace MonoDevelop.Projects
 			// Check project references.
 			foreach (var projectReference in dotNetProject.References.Where (pr => pr.ReferenceType == ReferenceType.Project)) {
 				if (traversedProjects.Contains (projectReference.ProjectGuid))
-					return;
+					continue;
 
 				var p = projectReference.ResolveProject (ParentSolution);
 				if (p != null)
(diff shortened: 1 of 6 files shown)


Therzok:
Numbers incoming.

Therzok:
GetMSBuildSdkPath time from 54248ms to 1681ms on Main.sln.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -1405,10 +1407,13 @@ void GetReferencedSDKs (TargetRuntime runtime, Project project, ref HashSet<stri
 				return;
 
 			// Check project references.
-			foreach (var projectReference in dotNetProject.References.Where (pr => pr.ReferenceType == ReferenceType.Project)) {
+			foreach (var projectReference in dotNetProject.References.Where (pr => pr.ReferenceType == ReferenceType.Project)) {
+				if (traversedProjects.Contains (projectReference.ProjectGuid))
+					return;
```
slluis: This should be a `continue`.

Therzok: Yes, mishap.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectReference.cs
```diff
@@ -96,6 +96,8 @@ public class ProjectReference : ProjectItem, ICloneable
 				package = value;
 			}
 		}
+
+		internal string ProjectGuid { get => projectGuid; }
```
Therzok: Can we make this public? More consumers of `ProjectReference.ResolveProject` could use a fast-path of marking as traversed by GUID, instead of doing a linear search for the _actual_ project first.

slluis: It could be public.

slluis:
How do number change after the fix?

Therzok:
I'm gathering them now, together with another perf improvement.

Therzok:
Takes 1850ms now.

PR #2229, opened 2017-04-20
Fix builder crash

slluis:
Fix race in project builder.
Don't reuse the workDoneEvent event anymore. In some cases it may
happen that the event is signaled twice (once when the work is
finished, and maybe once if the work is aborted just after it
finished. In those cases the second time the event is signaled it
could actually signal that the next task has been completed, which is
wrong. Also improve handling of thread abort exceptions, those
exceptions should never generate an error.
Should fix bug #52645 - XS cannot recover if the external MSBuild
builder dies

Fix deadlock in RemoteProcessConnection.
When processing a message response, the response should be notified
outside the messageWaiters lock, otherwise it may deadlock if the task
that handles the response is blocked in another lock.

### DIFF
--- a/main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/BuildEngine.Shared.cs
+++ b/main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/BuildEngine.Shared.cs
@@ -1,4 +1,4 @@
-// 
+﻿// 
 // ProjectBuilder.cs
 //  
 // Author:
@@ -43,7 +43,7 @@ namespace MonoDevelop.Projects.MSBuild
 {
 	partial class BuildEngine
 	{
-		static readonly AutoResetEvent workDoneEvent = new AutoResetEvent (false);
+		static AutoResetEvent workDoneEvent;
 		static ThreadStart workDelegate;
 		static readonly object workLock = new object ();
 		static Thread workThread;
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/BuildEngine.Shared.cs
```diff
@@ -43,7 +43,7 @@ namespace MonoDevelop.Projects.MSBuild
 {
 	partial class BuildEngine
 	{
-		static readonly AutoResetEvent workDoneEvent = new AutoResetEvent (false);
+		static AutoResetEvent workDoneEvent;
```
Therzok: Let's ditch the possible race condition. Why not make `STARunner` take an object parameter and pass in the `AutoResetEvent` via `Thread.Start`?

Therzok: There is no race condition, my bad. For some reason, I saw https://github.com/mono/monodevelop/blob/b28abe3ce9bcf55a7dd6e35293cc17fd6f652cca/main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/BuildEngine.Shared.cs#L256 as an assignment.

Still, the change would make it clearer that we're not changing which reset event is used.

PR #2231, opened 2017-04-20
[Core] Smarter import evaluation

Therzok:
This commit introduces the inheritance of the file existence cache at
MSBuildEvaluationContext level.

By inheriting the parent exists evaluation cache, we end up doing less
IO on grabbing Condition='Exists(x)'.

The lock on the evaluation cache introduces an extra 300ms locking
contention, but Evaluate(ProjectInfo, MSBuildEvaluationCache,
MSBuildImport, bool) ends having a total lock contention time reduced by
3000ms. (8900ms -> 5612ms)

Time spent in the above method is reduced from 57639ms to 48601ms.


slluis:
Maybe a better solution would be to use an immutable dictionary, and only lock when modifying it.

Therzok:
ImmutableDictionary addition is a costly operation. I think this is fine for now.

PR #2232, opened 2017-04-21
Fixed 'Bug 55100 - Source Analysis C# rules makes the tooltip

mkrueger:
unusable'.


mkrueger:
The purpose is a question for the roslyn team - in case of that issue it marks every location of the variable to be inlined. Which doesn't make that much sense IMO - it only produces code issue overflow - it's better to have less hot regions for code issues.

That's why I just ignore the additional locations for now.

PR #2237, opened 2017-04-21
Fixed 'Bug 55455 - Code completion lost when moving files'

mkrueger:
Fixed some potential exceptions found during trying out various
workarounds.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
@@ -882,7 +882,7 @@ namespace MonoDevelop.Ide.TypeSystem
 			if (!isOpen) {
 				delta = ApplyChanges (projection, data, changes);
 				var formatter = CodeFormatterService.GetFormatter (data.MimeType);
-				if (formatter.SupportsPartialDocumentFormatting) {
+				if (formatter != null && formatter.SupportsPartialDocumentFormatting) {
 					var mp = GetMonoProject (CurrentSolution.GetProject (id.ProjectId));
 					string currentText = data.Text;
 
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -1366,7 +1366,9 @@ void OnFileRemoved (object sender, MonoDevelop.Projects.ProjectFileEventArgs arg
 					var id = data.GetDocumentId (fargs.ProjectFile.FilePath);
 					if (id != null) {
 						ClearDocumentData (id);
-						OnDocumentRemoved (id);
+						try {
+							OnDocumentRemoved (id);
+						} catch (Exception) {}
```
slluis: Please don't do use empty catch blocks. You should log the exception or, if the exception is expected, do a specialized catch.

mkrueger:
It's the same than active configuration changed. Reloading single projects is no longer supported.
As far as I can tell the solution should be updated with a new project & TryApplyChanges should be called. However our try apply changes is not really suited for that kind of updates yet - so I chose that approach instead.

(TryApplyChanges takes a new solution as well - so updating projects is bad in any case)

DavidKarlas:
Btw, my 2. is not addressed yet(`if (!args.Any (x => x.Hint == "TargetFramework" || x.Hint == "References"))` should probably be extended to also handle `Files` changes)

slluis:
@mkrueger what's the performance impact of this change?

mkrueger:
@DavidKarlas: it's in the project model - as you said. I don't touch that for this issue - we've had that before and it can make side effects.

@lluis: That it takes 1-2 seconds longer until the typesystem is updated after file add/move.

DavidKarlas:
I'm not saying project model should be changed.
I'm just saying that once project model will be fixed to not report "References" changed when file is moved inside project, this fix won't work anymore... Because `if (!args.Any (x => x.Hint == "TargetFramework" || x.Hint == "References"))` filter...

mkrueger:
>once project model will be fixed<

PR #2238, opened 2017-04-21
Optimize Condition parsing

Therzok:


main/src/core/MonoDevelop.Core/CoreExtensions.cs
```diff
@@ -191,23 +191,52 @@ public static Func<R> Memoize<R> (this Func<R> f)
...
 				map.Add (a, value);
 				return value;
 			};
 		}
 
-		public static Func<T1, T2, R> Memoize<T1, T2, R> (this Func<T1, T2, R> f)
+		public static Func<T1, R> MemoizeWithLock<T1, R> (this Func<T1, R> f)
+		{
+			var map = new Dictionary<T1, R> ();
+			var lockObject = new object ();
```
KirillOsenkov: Consider locking on the map itself to avoid an extra allocation. It's not visible outside anyway.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild.Conditions/Token.cs
```diff
@@ -91,6 +91,25 @@ public override string ToString ()
 
 			return String.Format ("\"{0}\" at character position {1}", tokenValue, Position);
 		}
+
+		public override int GetHashCode()
+		{
+			return tokenValue.GetHashCode () ^ tokenType.GetHashCode () ^ position.GetHashCode ();
```
KirillOsenkov: Can tokenValue ever be null?

main/src/core/MonoDevelop.Core/CoreExtensions.cs
Therzok: Good point.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild.Conditions/Token.cs
Therzok: Yeah, it can on EOF/BOF token types. I'll add a conditional.

Therzok:
Fixed both issues.

PR #2240, opened 2017-04-23
[msbuild] Fix v4/v12 assemblies use with older mono

radical:


radical:
Same as https://github.com/mono/monodevelop/pull/2239

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1275,6 +1275,38 @@ static string GetLocalMSBuildExeLocation (TargetRuntime runtime)
...
+					var assemblies = new string[] {
+								"Microsoft.Build.Tasks.v4.0.dll",
+								"Microsoft.Build.Utilities.v4.0.dll",
+								"Microsoft.Build.Tasks.v12.0.dll",
+								"Microsoft.Build.Utilities.v12.0.dll" };
+
+					foreach (var asm in assemblies) {
+						var src = Path.Combine(binDir, asm);
+						var dest = Path.Combine (exesDir, asm);
+						if (File.Exists (src))
```
KirillOsenkov: You should add && !File.Exists (dest) so that if the file already exists File.Copy won't throw.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1275,6 +1275,38 @@ static string GetLocalMSBuildExeLocation (TargetRuntime runtime)
...
+				//
+				if (Platform.IsMac || Platform.IsLinux) {
+					var assemblies = new string[] {
+								"Microsoft.Build.Tasks.v4.0.dll",
+								"Microsoft.Build.Utilities.v4.0.dll",
+								"Microsoft.Build.Tasks.v12.0.dll",
+								"Microsoft.Build.Utilities.v12.0.dll" };
+
+					foreach (var asm in assemblies) {
+						var src = Path.Combine(binDir, asm);
```
KirillOsenkov: Not sure if there's anything in our coding guidelines, but I'd avoid abbreviations and use "assembly" instead of "asm", "source" instead of "src" and "destination" instead of "dest".

KirillOsenkov: Also a space is missing after Path.Combine

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/Main.cs
```diff
@@ -76,11 +76,7 @@ Assembly MSBuildAssemblyResolver (object sender, ResolveEventArgs args)
 							"Microsoft.Build.Framework",
 							"Microsoft.Build.Tasks.Core",
 							"Microsoft.Build.Utilities.Core",
-							"System.Reflection.Metadata",
-							"Microsoft.Build.Tasks.v4.0",
-							"Microsoft.Build.Utilities.v4.0",
-							"Microsoft.Build.Tasks.v12.0",
-							"Microsoft.Build.Utilities.v12.0" };
+							"System.Reflection.Metadata"};
```
KirillOsenkov: Not related to your PR but this reminds me to ask: Should System.Collections.Immutable be in this list? Normally System.Reflection.Metadata and System.Collections.Immutable always go together ;)

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -1275,6 +1275,38 @@ static string GetLocalMSBuildExeLocation (TargetRuntime runtime)
...
+				//
+				// But for the v4.0/v12.0 assemblies, the runtime is able to resolve them from the GAC and
+				// so the event handler never gets fired.
+				//
+				// To ensure that msbuild is able to load the facade assemblies, we copy them over next to
+				// the builder. This is temporary though. It can be removed when xbuild is removed from mono,
+				// and we can put these in the proper locations in mono (GAC/facades?).
+				//
+				if (Platform.IsMac || Platform.IsLinux) {
+					var assemblies = new string[] {
```
KirillOsenkov: As this code is called pretty often, consider allocating the array once in a static field outside the method body.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
KirillOsenkov: Actually never mind, I think the entire block is only called if the builder directory doesn't exist.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
radical: Ah, thanks. Missed this switching between msbuild style and mono!

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/Main.cs
radical: `System.Reflection.Metadata` was added here to workaround a bug where the file was not present next to `MSBuild.dll`.

radical:
Updated based on the feedback.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild.Shared/Main.cs
KirillOsenkov: Yes, that's fine. I'm saying maybe we should add System.Collections.Immutable to this list as well since normally it is also required to be present next to MSBuild dlls. But if you were fixing a particular scenario and if it worked for that case, that's fine then.

slluis:
Closing since this fix is already in master.

PR #2243, opened 2017-04-24
Fixed 'Bug 55476 - TypeSystem task loop'

mkrueger:
That makes the work around with adding/removing event handlers
obsolete. That didn't really work because the event was triggered on
UI thread.


Therzok:
Seems to fix the root cause in my bug report.

PR #2244, opened 2017-04-24
[54483] Fix XwtPopup Target screen bounds calculation

sevoku:
This PR fixes Xwt based Popup positioning relative to targets with different screen coordinate systems. This includes only native Cocoa popups on top of Gtk/Cocoa targets, other constellations need to be fixed as soon as we make use of them and are marked with `TODO` comments.
(fixes bug #54483)

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Xwt/XwtPopup.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Xwt/XwtPopup.cs
@@ -70,6 +70,8 @@ namespace MonoDevelop.Components
 			get { return eventProvided; }
 		}
 
+		protected Toolkit Toolkit { get; private set; }
+
 		protected virtual void CurrentPositionChanged ()
 		{
 		}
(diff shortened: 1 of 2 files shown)


slluis:
I feel it would be easier to fix this if we had a method in Xwt.Toolkit for getting the Xwt screen coordinates for a given native widget.

sevoku:
Theoretically yes, but this would add dependencies on the other toolkits (assembly refs) or we'd need to use reflection excessively.

slluis:
I mean a method for getting the XWT screen coordinates of a widget native to the toolkit. I don't understand why you need to do cross-toolkit conversions. Can you give an example?

sevoku:
Hhm, The problem is that for the conversion we need to know the screen coords of the widget and the actual display bounds. Getting the screen coords in the coord system of the target could be moved to Xwt, but we'd still need to get its Display information, which is specific to the destination toolkit. So it would save us just 2-3 lines on the MD side. It won't be useful in any other case, cause the screen coords are only valid in the Xwt context (you can't use the Xwt bounds to align a native widget).

sevoku:
Code simplified and moved major parts to Xwt: https://github.com/mono/xwt/pull/682

PR #2252, opened 2017-04-24
Update Roslyn to latest stable 

mhutch:


mhutch:
Split the roslyn upgrade and the bugfix into #2254 and #2253

PR #2253, opened 2017-04-24
Update Roslyn to 2.1

mhutch:


slluis:
I think the Newtonsoft.Json upgrade may break stuff.

slluis:
Ignore my comment. I forgot that we are already using Newtonsoft.Json 9.0.1. I saw the change in packages.config and I thought it was an upgrade.

Therzok:
This broke RefactoringEssentials on Windows. @mkrueger do we need to bump roslyn there too?

PR #2257, opened 2017-04-25
Fixed 'Bug 55456 - Incorrect behavior and exception when trying to

mkrueger:
move a file that is modified in the editor'

The real bug was that the open file closed when moving to another
location. This patch ensures that the move file event is handled
properly.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
@@ -666,14 +666,14 @@ namespace MonoDevelop.Ide.Gui
 					var views = new ViewContent [viewContentCollection.Count];
 					viewContentCollection.CopyTo (views, 0);
 					foreach (var content in views) {
-						if (content.ContentName.StartsWith (e.SourceFile, FilePath.PathComparison)) {
+						if (e.SourceFile.IsChildPathOf (content.ContentName)) {
 							content.ContentName = e.TargetFile + content.ContentName.Substring (e.SourceFile.ToString ().Length);
 						}
 					}
 				} else {
 					foreach (var content in viewContentCollection) {
 						if (content.ContentName != null &&
-						    FilePath.PathComparer.Compare (content.ContentName, e.SourceFile) == 0) {
+						    content.ContentName == e.SourceFile) {
 							content.ContentName = e.TargetFile;
 							return;
 						}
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
```diff
@@ -658,7 +658,30 @@ public Properties GetStoredMemento (ViewContent content)
...
-		
+
+		void CheckFileMoved (object sender, FileCopyEventArgs args)
+		{
+			foreach (var e in args) {
+				if (e.IsDirectory) {
+					var views = new ViewContent [viewContentCollection.Count];
+					viewContentCollection.CopyTo (views, 0);
+					foreach (var content in views) {
+						if (content.ContentName.StartsWith (e.SourceFile, FilePath.PathComparison)) {
```
slluis: This will do the wrong thing if two directories have the same initial substring. You should use e.SourceFile.IsChildPathOf.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
```diff
@@ -658,7 +658,30 @@ public Properties GetStoredMemento (ViewContent content)
...
+							content.ContentName = e.TargetFile + content.ContentName.Substring (e.SourceFile.ToString ().Length);
+						}
+					}
+				} else {
+					foreach (var content in viewContentCollection) {
+						if (content.ContentName != null &&
+						    FilePath.PathComparer.Compare (content.ContentName, e.SourceFile) == 0) {
```
slluis: Just do `e.SourceFile == content.ContentName`. The overloaded operator for FilePath already uses PathComparer.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
```diff
@@ -658,7 +658,30 @@ public Properties GetStoredMemento (ViewContent content)
...
-		
+
+		void CheckFileMoved (object sender, FileCopyEventArgs args)
+		{
+			foreach (var e in args) {
+				if (e.IsDirectory) {
+					var views = new ViewContent [viewContentCollection.Count];
+					viewContentCollection.CopyTo (views, 0);
+					foreach (var content in views) {
+						if (e.SourceFile.IsChildPathOf (content.ContentName)) {
```
slluis: My bad, this is not right, the comparison is reversed. This should actually be `((FilePath)content.ContentName).IsChildPathOf (e.SourceFile)`

PR #2258, opened 2017-04-25
Fixed 'Bug 55308 - Save as, then open original CSS or HTML file, you

mkrueger:
get "An Autosave file has been found for this file"'


mkrueger:
AutoSave needs to be removed before the new content name is set. That's why that code got moved prior to the content name update.

PR #2261, opened 2017-04-25
[Svn] Make unix svn resilient to native crashes

Therzok:
Formerly, this code would use library paths to open libsvn and libapr accordingly.

This meant that it would use the search paths for each library that would end up loading.

In the case of Xcode not being located and svn being installed via brew, it would load mismatched dlls for the two
libraries.

dlopen-ing libsvn will ensure the right apr is loaded for the probed svn lib

Bug 47563 - Visual Studio for mac crashes on opening a solution

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl.Subversion.Unix/MonoDevelop.VersionControl.Subversion.Unix/LibSvnClient.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl.Subversion.Unix/MonoDevelop.VersionControl.Subversion.Unix/LibSvnClient.cs
@@ -34,7 +34,8 @@ namespace MonoDevelop.VersionControl.Subversion.Unix {
 			// crash. dlopen-ining libsvn_client allows us to handle the right mix-match of svn and apr, since it loads
 			// apr via the linker flags on the binary.
 			if (Core.Platform.IsMac) {
-				dlopen ("libsvn_client-1.0.dylib", 0x1);
+				if (dlopen ("libsvn_client-1.0.dylib", 0x1) == IntPtr.Zero)
+					return null;
 				return new LibSvnClient2 ();
 			}
 			try {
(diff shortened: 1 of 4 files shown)


main/src/addins/VersionControl/MonoDevelop.VersionControl.Subversion.Unix/MonoDevelop.VersionControl.Subversion.Unix/LibSvnClient.cs
```diff
@@ -17,12 +17,19 @@ protected LibSvnClient ()
...
 		}
-		
+
+		[DllImport ("libc")]
+		extern static IntPtr dlopen (string name, int mode);
+
 		public static LibSvnClient GetLib ()
 		{
+			if (Core.Platform.IsMac) {
+				dlopen ("libsvn_client-1.0.dylib", 0x1);
```
slluis: Can you please add a comment explaining why is this being done?

Therzok: Yes.

slluis: Also, will this load the corresponding libapr too?

Therzok: Yes, it loads everything accordingly, all the right deps for libsvn_client. Not sure why Mono isn't resolving libapr to the one libsvn loads via the dlloader, but ¯\_(ツ)_/¯

slluis: if this dlopen fails, `LibApr GetLib()` will return a LibAprX that doesn't match LibSvnClientX.

Therzok: > if this dlopen fails, LibApr GetLib() will return a LibAprX that doesn't match LibSvnClientX.

Fixed via https://github.com/mono/monodevelop/pull/2261/commits/8c444700b1eb57b7a8391866db4cb74804926a0f

Therzok: Returning null will mean that the apr check won't even be done.

PR #2263, opened 2017-04-25
Bug 55158 - XS hangs if MSBuild returns 2000 errors

DavidKarlas:
Majority of fix is in checking if Width actually changed, since it turns out notification is sent multiple times even when size is not changed… While looking at GTK+ code I noticed there is much better way to invalidate recalculation of width for all rows: https://github.com/GNOME/gtk/blob/gtk-2-24/gtk/gtktreeviewcolumn.c#L3749-L3765


DavidKarlas:
@slluis based on my testing this is not regression compared to stable XS(6.3). I think change is small enough to be merged into 15.2 imo.

Therzok:
@DavidKarlas as a side note, https://github.com/bratsche/gtk is what reflects our gtk source code.

alanmcgovern:
It fixed it for me!

PR #2265, opened 2017-04-25
Fixed 'Bug 55557 - Weird formatting behavior'

mkrueger:


Therzok:
We should probably add unit tests for this.

PR #2273, opened 2017-04-26
Make sure the comparer used to compare completion items is stable.

KirillOsenkov:
Should fix https://bugzilla.xamarin.com/show_bug.cgi?id=55634

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/ListWidget.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/ListWidget.cs
@@ -728,22 +728,29 @@ namespace MonoDevelop.Ide.CodeCompletion
 				}
 			}
 			filteredItems.Sort (delegate (int left, int right) {
-				int rank1, rank2;
 				var data1 = dataList [left];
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/ListWidget.cs
```diff
@@ -728,22 +728,29 @@ static CompletionListFilterResult DefaultFilterWords (ICompletionDataList dataLi
 				}
 			}
 			filteredItems.Sort (delegate (int left, int right) {
-				int rank1, rank2;
 				var data1 = dataList [left];
 				var data2 = dataList [right];
 				if (data1 == null || data2 == null)
-					return 0;
+					return left.CompareTo (right);
```
slluis: This doesn't look correct to me. If data1 != null and data2 == null, data1 should go first, no matter the original position. I think returning 0 was wrong.

KirillOsenkov: You're right. Fixed. How does it look now?

PR #2284, opened 2017-04-27
Fix new file formatting

slluis:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/SingleFileDescriptionTemplate.cs
```diff
@@ -310,8 +310,8 @@ public virtual Stream CreateFileContent (SolutionFolderItem policyParent, Projec
 			var doc = TextEditorFactory.CreateNewDocument ();
 			doc.Text = content;
 			
-			TextStylePolicy textPolicy = policyParent != null ? policyParent.Policies.Get<TextStylePolicy> ("text/plain")
-				: MonoDevelop.Projects.Policies.PolicyService.GetDefaultPolicy<TextStylePolicy> ("text/plain");
+			TextStylePolicy textPolicy = policyParent != null ? policyParent.Policies.Get<TextStylePolicy> (mime)
+				: MonoDevelop.Projects.Policies.PolicyService.GetDefaultPolicy<TextStylePolicy> (mime);
 			string eolMarker = TextStylePolicy.GetEolMarker (textPolicy.EolMarker);
```
mrward: Just wondering what happens if the mime is null here.

slluis: I will add a check.

PR #2288, opened 2017-04-27
Projection buffer support

dpugh:
Various changes to support VS/Venus-type projection in MD.

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SourceEditorView.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SourceEditorView.cs
@@ -265,13 +265,15 @@ namespace MonoDevelop.SourceEditor
 			public DocumentAndLoaded (string fileName, string mimeType) {
 				if (AutoSave.AutoSaveExists(fileName)) {
 					// Don't load the document now, let Load() handle it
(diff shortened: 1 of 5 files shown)


main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SourceEditorView.cs
```diff
@@ -265,13 +265,15 @@ private struct DocumentAndLoaded {
...
 					// Don't load the document now, let Load() handle it
-					this.Document = new TextDocument();
-					this.Document.MimeType = mimeType;
-					this.Document.FileName = fileName;
+					this.Document = new TextDocument(string.Empty, fileName, mimeType);
 
 					this.Loaded = false;
 				} else {
-					this.Document = new TextDocument(fileName, mimeType);
+					// HACK we really need to be told explicitly that the document is for an existing file (that we should load) or a new file (which will have the given name/mime type)
```
Therzok: ViewContent.IsFile is not enough?

dpugh: No ... the caller of this function knows whether the user is doing an open file vs. a new file which is really what we'd like to know here rather than having to infer it.

PR #2289, opened 2017-04-27
Fix bug 55004 - Css property difficult to see on some dark Color Themes

slluis:


slluis:
Adjusted colors for all dark themes.

PR #2292, opened 2017-04-28
[Project] Add VS-like AppliesTo extension condition

mhutch:
This condition can be used to specify that a project extension, file template, or option panel apply to projects that match a particular capability expression.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
@@ -69,6 +69,7 @@ namespace MonoDevelop.Projects
 
 				extensionContext = AddinManager.CreateExtensionContext ();
 				extensionContext.RegisterCondition ("ItemType", new ItemTypeCondition (GetType ()));
+				ExtensionContext.RegisterCondition ("AppliesTo", new AppliesToCondition (this));
 
 				OnInitialize ();
 				InitializeExtensionChain ();
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileTemplate.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileTemplate.cs
@@ -41,6 +41,7 @@ using MonoDevelop.Core;
 using MonoDevelop.Ide.Codons;
 using MonoDevelop.Ide.Gui;
 using MonoDevelop.Projects;
+using MonoDevelop.Projects.Extensions;
 
 namespace MonoDevelop.Ide.Templates
 {
@@ -59,8 +60,6 @@ namespace MonoDevelop.Ide.Templates
 		public string Description { get; private set; } = String.Empty;
 
 		public List<FileDescriptionTemplate> Files { get; private set; } = new List<FileDescriptionTemplate> ();
-
-		public static List<FileTemplate> fileTemplates = new List<FileTemplate> ();
 
 		public IconId Icon { get; private set; } = String.Empty;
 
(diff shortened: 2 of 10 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
```diff
@@ -70,6 +70,11 @@ internal void EnsureInitialized ()
 				extensionContext = AddinManager.CreateExtensionContext ();
 				extensionContext.RegisterCondition ("ItemType", new ItemTypeCondition (GetType ()));
 
+				var project = this as Project;
+				if (project != null) {
+					extensionContext.RegisterCondition ("AppliesTo", new AppliesToCondition (project);
+				}
+
 				OnInitialize ();
```
slluis: Instead of doing a `Project` type check here, it would be better to register this condition in the `OnInitialize` method of the Project class. `extensionContext` is accessible through a public property.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileTemplate.cs
```diff
@@ -191,58 +190,51 @@ internal static FileTemplate LoadFileTemplate (RuntimeAddin addin, XmlDocument x
...
+			extensionContext.RegisterCondition ("AppliesTo", new AppliesToCondition (project));
+			extensionContext.RegisterCondition ("FlavorType", new FlavorTypeCondition (project));
+			extensionContext.RegisterCondition ("ProjectTypeId", new ProjectTypeIdCondition (project));
+
 			var list = new List<FileTemplate> ();
-			foreach (var t in fileTemplates) {
-				if (t.IsValidForProject (project, projectPath))
-					list.Add (t);
+			foreach (var node in extensionContext.GetExtensionNodes<ProjectTemplateCodon> (EXTENSION_PATH)) {
+				var template = LoadTemplate (node);
```
slluis: With this change templates are going to be loaded every time GetFileTemplates and GetFileTemplateByID is invoked, which may be time and memory consuming. A solution would be to cache the loaded template instance in ProjectTemplateCodon.

mhutch: It doesn't seem to hurt. Opening the New File dialog is still ~instant. And caching the templates is memory consuming too :)

mhutch:
@slluis I addressed your feedback 🙂

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
mhutch: Ok, so it turns out that causes an error to be logged about a nonexistent condition ID when getting extensions for workspace objects that aren't projects.

Thoughts?

mhutch:
See also https://github.com/mono/mono-addins/pull/62 which fixes an issue using condition attributes

mhutch:
see also https://github.com/mono/mono-addins/pull/61

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
```diff
@@ -70,12 +70,23 @@ internal void EnsureInitialized ()
 				extensionContext = AddinManager.CreateExtensionContext ();
 				extensionContext.RegisterCondition ("ItemType", new ItemTypeCondition (GetType ()));
 
+				ExtensionContext.RegisterCondition (
+					"AppliesTo",
+					this is Project
+					? new AppliesToCondition ((Project)this)
+					: (ConditionType)new FalseCondition ());
```
slluis: I have a proposal. AppliesToCondition can take a WorkspaceObject in the constructor, then Evaluate can check if the object is a Project and check the capabilities in that case, or return false for other types of objects. In this way we don't need to check for Project here and we don't need FalseCondition.

mhutch: done 😄

slluis: 👍

PR #2301, opened 2017-04-28
Fix document save logic

slluis:
Fixed bug in the logic that saves a document when it is closed.
args.Cancel was always set to true, which was wrong since
the Close method of the document would always return false,
even when properly saving the document.

This change was done to fix an issue with asyc save
of documents, which I think was properly fixed by
https://github.com/mono/monodevelop/commit/9c9f2aa021b3f48cb865f4307ae9b59372611b25

Fixes bug #55713 - Selecting recent solution when unsaved files
open does not open solution


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
```diff
@@ -855,21 +855,21 @@ Pad WrapPad (PadCodon padContent)
 					GettextCatalog.GetString ("If you don't save, all changes will be permanently lost."),
 					AlertButton.CloseWithoutSave, AlertButton.Cancel, viewContent.IsUntitled ? AlertButton.SaveAs : AlertButton.Save);
 				if (result == AlertButton.Save) {
-					args.Cancel = true;
```
Therzok: We shouldn't trigger `OnDocumentClosing` if we canceled.

slluis: Fixed

PR #2303, opened 2017-04-28
Master fix55670

mkrueger:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/syntaxes/CSharp/C#.sublime-syntax
```diff
@@ -74,6 +74,10 @@ contexts:
   numbers:
     - match: '((\b\d+\.?\d+)|(\.\d+))([eE][+-]?\d*)?(F|f|D|d|M|m)?\b'
       scope: constant.numeric.float.source.cs
+    - match: '\d+(([eE][+-]?\d*)?(F|f|D|d|M|m)|([eE][+-]?\d*))\b'
```
KirillOsenkov: Will this match the '.f5' in `0.5f`? Because it is already matched by the previous expression (with dots)? Will the two overlap? Not sure how I'd fix this.

Another thing I notice: isn't it equivalent to: `\d+(([eE][+-]?\d*)?(F|f|D|d|M|m)?)\b`

KirillOsenkov: Maybe unify this with the previous expression using ( | ) so that they're mutually exclusive?

mkrueger: how can it if it's already matched ?

KirillOsenkov: Oh, I see. Are you saying that once the first expression matched, the second won't look in that span anymore? I don't know how this file format works.

mkrueger: ah ok didn't know that - yes that's how it works first expressions first, if matched they're matched and the matching begins after that again. So if the .5f is matched by the first expression the second won't be run.

It's a nice in theory but in practice ALL matches are matched against the line and the 1st match (lowest offset) is taken, after that all valid matches are run and then the first match is taken again.
This makes textmate grammers very slow to run compared to our older 'greedy' coloring system.

PR #2321, opened 2017-05-02
D15 2 allocs

Therzok:


main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Document/TextDocument.cs
```diff
@@ -45,7 +45,7 @@
 
 namespace Mono.TextEditor
 {
-	class TextDocument : ITextDocument
+	class TextDocument : ITextDocument, IDisposable
```
Therzok: Not sure how to detect the close, so I made this IDisposable.

PR #2323, opened 2017-05-02
Fixed 'Bug 55705 - Quick Fix doesn't do anything'

mkrueger:
Was caused by a projection parsing error race. The lock should be
enough to fix it but now we've immutable data structures which add
more thread safety.


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -616,7 +619,7 @@ IEnumerable<DocumentInfo> GenerateProjections (MonoDevelop.Projects.ProjectFile
 					false
 				);
 			}
-			projectionList.Add (entry);
+			projectionList = projectionList.Add (entry);
```
slluis: Don't we need locking here?

mkrueger: No that's in generation - after generation updates occur and there could be races.

PR #2325, opened 2017-05-03
[Ide] TextFileNavigation points remove text source version on document

mkrueger:
close.

Versions become invalid after closing the document & leak documents in
memory.


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/TextFileNavigationPoint.cs
```diff
@@ -34,27 +34,38 @@
...
 			line = location.Line;
 			column = location.Column;
-			offset = buffer.CaretOffset;
 		}
-		
+
+		protected override void OnDocumentClosing ()
+		{
+			// text source version becomes invalid on document close.
+			var editor = Document.Editor;
```
Therzok: Well, we don't get rid of the buffer version reference here.

PR #2326, opened 2017-05-03
[F#] Only show Debug F# script menu for F# files.

nosami:
Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=55873


slluis:
What are those .sln changes for? Can you make a PR for d15-2 without those?

nosami:
That .sln file is for my personal use. It doesn't affect anybody else.

I'll change the d15-2 PR though.

nosami:
Updated https://github.com/mono/monodevelop/pull/2327

PR #2328, opened 2017-05-03
[Core] Handle unknown MSBuild property being evaluated during build

mrward:
Fixed bug #55751 - [iOS] Uploading iOS Forms PCL app from Xamarin
Studio to Test Cloud fails build step with "Unknown MSBuild Failure"
https://bugzilla.xamarin.com/show_bug.cgi?id=55751

Two fixes have been made. Both on their own fix the above bug. One fix is in the BinaryMessage so it handles a null dictionary value. Another fix is in the project builders for xbuild and MSBuild so they return an empty string for unknown properties that are being evaluated instead of null, which allows the BinaryMessage to send the message back to the IDE.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/BinaryMessage.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/BinaryMessage.cs
@@ -257,7 +257,7 @@ namespace MonoDevelop.Core.Execution
 			} else if (et == typeof(string)) {
 				bw.Write ((byte)TypeCode.String);
 				foreach (var v in (string [])val)
-					bw.Write (v);
+					bw.Write (v ?? "");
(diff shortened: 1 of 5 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/BinaryMessage.cs
```diff
@@ -257,7 +257,7 @@ void WriteArray (BinaryWriter bw, object val)
 			} else if (et == typeof(string)) {
 				bw.Write ((byte)TypeCode.String);
 				foreach (var v in (string [])val)
-					bw.Write (v);
+					bw.Write (v ?? "");
```
slluis: Writing a null value in this context should be allowed, but that will require a more complex solution. Let's commit this for now.

PR #2329, opened 2017-05-03
Master fix formatting

mkrueger:


main/tests/UnitTests/MonoDevelop.CSharpBinding/CSharpTextEditorIndentationTests.cs
```diff
@@ -49,7 +50,7 @@ class CSharpTextEditorIndentationTests : TestBase
 	{
 		const string eolMarker = "\n";
 
-		internal static TestViewContent Create (string input, Ide.Editor.ITextEditorOptions options = null)
+		internal static TestViewContent Create (string input, Ide.Editor.ITextEditorOptions options = null, bool createWithProject = false)
```
slluis: What is this new parameter used for? there isn't any reference to it.

DavidKarlas: It's used on line 144 and it's set to true from unit tests

PR #2335, opened 2017-05-03
Bug 55869 - Can not inspect variables in .Net core applications

DavidKarlas:
There are legit cases of type being null, like “Static members” and “Non-public members” groups, but also when evaluating variables(hover, repl…) .NET Core debugger doesn’t set Type which we in past considered as error, now we just ignore it and set it to empty string…


main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
```diff
@@ -84,17 +84,11 @@ public object GetRawValue (ObjectPath path, EvaluationOptions options)
 
 		public ObjectValue GetValue (ObjectPath path, EvaluationOptions options)
 		{
-			string shortName = name;
-			var indexOfSpace = name.IndexOf (' ');
-			if (indexOfSpace != -1)//Remove " [TypeName]" from variable name
-				shortName = name.Remove (indexOfSpace);
```
slluis: Why is this code being removed?

DavidKarlas: In past `name` had "variableName [TypeName]" but now it is split into two variables `Name` and `Type` hence this trimming of variable name after space is not needed anymore(probably it also caused bugs in examples like "Static members".

PR #2336, opened 2017-05-03
Bug 55875 - Call Stack Disappears when a thread selection is made whi…

DavidKarlas:
…le debugging .NET Core application

Problem was that Backtrace.DebuggingSession was null, that was based because Backtrace.Attach(session); was not called, this was caused by setting backtrace directly on ThreadInfo constructor instead via ThreadInfo.Backtrace method which takes care of calling Backtrace.Attach(session);, same was needed to be done by OnGetThreads->GetThreads and OnGetProcesses->GetProcesses

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VSCodeDebuggerSession.cs
+++ b/main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VSCodeDebuggerSession.cs
@@ -87,12 +87,10 @@ namespace MonoDevelop.Debugger.VsCodeDebugProtocol
 			var threadsResponse = protocolClient.SendRequestSync (new ThreadsRequest ());
 			var threads = new ThreadInfo [threadsResponse.Threads.Count];
 			for (int i = 0; i < threads.Length; i++) {
-				var backtrace = this.GetThreadBacktrace (threadsResponse.Threads [i].Id);
 				threads [i] = new ThreadInfo (processId,
-										  threadsResponse.Threads [i].Id,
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VSCodeDebuggerSession.cs
```diff
@@ -280,8 +278,8 @@ protected void HandleEvent (object sender, EventReceivedEventArgs obj)
 					}
 					currentThreadId = body.ThreadId ?? -1;
 					//TODO: what happens if thread is not specified?
-					args.Process = OnGetProcesses () [0];
-					args.Thread = GetThread (args.Process, (long)body.ThreadId);
+					args.Process = GetProcesses () [0];
+					args.Thread = args.Process.GetThreads ().Single (t => t.Id == currentThreadId);
```
Therzok: Semantic changed. Isn't it `SingleOrDefault`?

DavidKarlas: It would fail on next line anyway, feels better if it fails fast

slluis:
Where is ThreadInfo calling Backtrace.Attach(session)?

DavidKarlas:
It's called from here: https://github.com/mono/debugger-libs/blob/master/Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs#L1026 and that code is called from https://github.com/mono/debugger-libs/blob/master/Mono.Debugging/Mono.Debugging.Client/ThreadInfo.cs#L79
Since 1st method is internal I can't call directly.

PR #2340, opened 2017-05-03
D15 2 more perf

Therzok:


slluis:
Reviewers, I need more eyes on this.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionData.cs
```diff
@@ -197,6 +197,11 @@ public virtual bool IsOverload (CompletionData other)
 			return DisplayText == other.DisplayText;
 		}
 
+		public override int GetHashCode ()
```
slluis: Why add this? there is no Equals() override.

Therzok: Custom IEqualityComparer which uses IsOverload as Equals

Therzok: I could probably get rid of this and just make the GetHashCode<T> return it.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Features/SemanticHighlighting/SemanticHighlightingVisitor.cs
```diff
@@ -537,13 +537,9 @@ public override void VisitIdentifierName(Microsoft.CodeAnalysis.CSharp.Syntax.Id
...
-					symbol = semanticModel.GetSymbolInfo (node, cancellationToken).Symbol;
-					// var sym = vds.Variables[0].Initializer != null ? vds.Variables[0].Initializer.Value as LiteralExpressionSyntax : null;
-					if (symbol == null || symbol.Name != "var") {
-						Colorize(node.Span, varKeywordTypeColor);
-						return;
-					}
+				if (vds != null && vds.Variables.Count == 1) {
+					Colorize(node.Span, varKeywordTypeColor);
+					return;
 				}
```
DavidKarlas: @Therzok can you explain a bit why we don't need semantic confirmation for `VariableDeclarationSyntax`, but we need it on line 533 for `ForEachStatementSyntax`?

Therzok: That wasn't hit a lot, I just fixed whatever showed up via profiling. I guess that isn't needed either.

Therzok: On the other hand, that does not check whether the symbol is null or whether its name is var. Probably, we don't need to check the symbol there too.

PR #2346, opened 2017-05-03
Huge leak fixes proper rebased

Therzok:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/NavigationHistoryService.cs
```diff
@@ -58,16 +59,21 @@ static NavigationHistoryService ()
...
 			IdeApp.Workspace.LastWorkspaceItemClosed += delegate {
 				history.Clear ();
-				closedHistory.Clear ();
 				OnHistoryChanged ();
+				closedHistory.Clear ();
+				OnClosedHistoryChanged ();
+			};
+
+			IdeApp.Workbench.DocumentOpened += delegate (object sender, DocumentEventArgs e) {
+				closedHistory.RemoveAll(np => (np.Item1 as DocumentNavigationPoint)?.FileName == e.Document.FileName);
```
alanmcgovern: Should this emit `OnClosedHistoryChanged` too?

alanmcgovern: Would we be better with an ObservableCollection here?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/NavigationHistoryService.cs
```diff
@@ -226,9 +236,12 @@ static void SwitchToCurrent ()
 
 		public static async void OpenLastClosedDocument () {
 			if (HasClosedDocuments) {
-				var tuple = closedHistory.Pop ();
+				int closedHistoryIndex = closedHistory.Count - 1;
+				var tuple = closedHistory[closedHistoryIndex];
+				closedHistory.RemoveAt (closedHistoryIndex);
```
alanmcgovern: Should this emit `OnClosedHistoryChanged` ?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeTemplates/EditTemplateDialog.cs
```diff
@@ -115,6 +115,12 @@ public EditTemplateDialog (CodeTemplate template, bool isNew)
 			UpdateVariables ();
 		}
 
+		protected override void OnDestroyed ()
+		{
+			textEditor.TextChanged -= DocumentTextReplaced;
```
alanmcgovern: Should we also unhook `textEditor.CaretPositionChanged` ?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildEvaluationContext.cs
```diff
@@ -47,7 +47,7 @@ class MSBuildEvaluationContext: IExpressionContext
 	{
 		Dictionary<string,string> properties = new Dictionary<string, string> (StringComparer.OrdinalIgnoreCase);
 		static Dictionary<string, string> envVars = new Dictionary<string, string> ();
```
alanmcgovern: MSBuild treats env vars as properties, so if properties are case-insensitive then it's likely that env vars need to be case insensitive too. (Someone correct me if I'm wrong about that :p )

main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Document/TextDocument.cs
```diff
@@ -220,6 +222,14 @@ protected void Initialize()
 			this.diffTracker.SetTrackDocument(this);
 		}
 
+		public void Dispose()
+		{
+			this.TextBuffer.Changed -= this.OnTextBufferChanged;
+			this.TextBuffer.ContentTypeChanged -= this.OnTextBufferContentTypeChanged;
+			this.TextBuffer.Properties.RemoveProperty(typeof(ITextDocument));
+			this.VsTextDocument.FileActionOccurred -= this.OnTextDocumentFileActionOccured;
```
alanmcgovern: Should we set `SyntaxMode = null` to ensure we disconnect from it's events too?

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Parser/CSharpParsedDocument.cs
```diff
@@ -462,37 +462,33 @@ public override void VisitBlock (Microsoft.CodeAnalysis.CSharp.Syntax.BlockSynta
...
+						errors = emptyErrors;
+					} catch (Exception e) {
+						LoggingService.LogError ("Error while getting diagnostics.", e);
+						errors = emptyErrors;
+					}
+				}
+			} finally {
+				if (locked)
+					errorLock.Release ();
			}
```
alanmcgovern: This looks like an oopsed newline :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/NavigationHistoryService.cs
Therzok: Yes.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/NavigationHistoryService.cs
Therzok: Hmmm, maybe?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeTemplates/EditTemplateDialog.cs
Therzok: We could, yes, but that didn't cause a leak.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildEvaluationContext.cs
Therzok: yes, but we're between a rock and a hard place with this. We query the env-var as is, so if we don't preserve case sensitivity here, someone might end up debugging why:
`export a="ASDF"`
and in msbuild:
`$(A) == ''ASDF"` // -> returns false

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Parser/CSharpParsedDocument.cs
Therzok: VS bit me.

main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Document/TextDocument.cs
Therzok: We can.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeTemplates/EditTemplateDialog.cs
alanmcgovern: If `TextChanged` did then i don't see how `CaretPositionChanged` wouldn't also leak. It's an event on the same class!

Therzok: I misexplained. I mean, the main issue I fixed there (doc open/close leak) didn't manifest from caret change. I just unsubscribed _that_ event from everywhere, didn't open the templates dialog to see if it leaks.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Features/Completion/CompletionResult.cs
```diff
@@ -121,11 +97,19 @@ internal void AddRange (IEnumerable<CompletionData> completionData)
...
+		class CompletionOverloadEqualityComparer : IEqualityComparer<CompletionData>
+		{
+			public bool Equals (CompletionData x, CompletionData y)
+			{
+				return x.IsOverload (y) && y.IsOverload(x);
+			}
+
+			public int GetHashCode (CompletionData obj)
+			{
+				return obj.DisplayText.GetHashCode ();
```
slluis: You are assuming here that DisplayText is used to determine if a CompletionData is an overload, but that may not be true since IsOverload is a virtual method, so it may have a different implementation (for example https://github.com/mono/monodevelop/blob/master/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/RoslynSymbolCompletionData.cs#L479).

One possible solution could be to have a corresponding GetOverloadHashCode() virtual method, but this is a breaking change since an extension that already implements IsOverload but not GetOverloadHashCode may generate wrong overload results. I also find this naming a bit confusing, it is not clear what GetOverloadHashCode would have to return.

To make the concept more clear, I would introduce the concept of "Overload Group", which is what groups entries that need to be shown together as overloads. We could then have two methods: OverloadGroupEquals(CompletionData), and GetOverloadGroupHashCode(). The implementation would be basically the same as IsOverload/GetOverloadHashCode, but I think it is more clear. Another option is to just add an OverloadGroup property and use that for comparing, but that could be less efficient since it would involve the allocation of an additional object in the case of RoslynSymbolCompletionData.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Features/SemanticHighlighting/SemanticHighlightingVisitor.cs
```diff
@@ -537,13 +537,9 @@ public override void VisitIdentifierName(Microsoft.CodeAnalysis.CSharp.Syntax.Id
 					}
 				}
 				var vds = node.Parent as VariableDeclarationSyntax;
-				if (vds != null && vds.Variables.Count == 1) {
-					symbol = semanticModel.GetSymbolInfo (node, cancellationToken).Symbol;
```
slluis: Why remove this check?

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Parser/CSharpParsedDocument.cs
```diff
@@ -462,37 +462,33 @@ public override void VisitBlock (Microsoft.CodeAnalysis.CSharp.Syntax.BlockSynta
...
 		{
 			var model = GetAst<SemanticModel> ();
 			if (model == null)
-				return Task.FromResult (emptyErrors);
+				return emptyErrors;
 			
-			if (errors == null) {
-				return Task.Run (async delegate {
-					bool locked = await errorLock.WaitAsync (Timeout.Infinite, cancellationToken).ConfigureAwait (false);
+			bool locked = await errorLock.WaitAsync (Timeout.Infinite, cancellationToken).ConfigureAwait (false);
```
slluis: We should do a preventive check before calling WaitAsync:
```csharp
if (errors != null)
    return errors;
```

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.CodeActions/CodeActionEditorExtension.cs
```diff
@@ -163,6 +162,9 @@ void HandleCaretPositionChanged (object sender, EventArgs e)
 							var provider = cfp.GetCodeFixProvider ();
 							if (!provider.FixableDiagnosticIds.Any (diagnosticIds.Contains))
 								continue;
+
+							Func<Diagnostic, bool> providerIdsContain = d => provider.FixableDiagnosticIds.Contains (d.Id);
```
slluis: Please add comments explaining why this is being factored out.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.CodeActions/CodeRefactoringService.cs
```diff
@@ -99,7 +99,11 @@ static CodeRefactoringService ()
 			var parsedDocument = doc.ParsedDocument;
 			var actions = new List<ValidCodeAction> ();
 			if (parsedDocument == null)
+				return actions;
+			var analysisDocument = doc.AnalysisDocument;
+			if (analysisDocument == null)
 				return actions;
+
 			var model = await doc.AnalysisDocument.GetSemanticModelAsync (cancellationToken);
```
slluis: For clarity, reuse analysisDocument here

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.Wrappers/SemanticHighlightingSyntaxMode.cs
```diff
@@ -162,7 +163,13 @@ public void Dispose()
 				var segments = new List<ColoredSegment> (syntaxLine.Segments);
 				int endOffset = segments [segments.Count - 1].EndOffset;
 				try {
-					var tree = lineSegments.FirstOrDefault (t => t.Item1 == line);
```
slluis: Please add a comment explaining the reason for the change.

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SourceEditorView.cs
```diff
@@ -1072,6 +1072,7 @@ public override void Dispose ()
 			currentDebugLineMarker = null;
 
 			RemoveMarkerQueue ();
+			widget.TextEditor.Document.Dispose ();
 			widget.Dispose ();
```
slluis: Why explicitly dispose the document? shouldn't `widget.Dispose ()` dispose everything inside?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockItemTitleTab.cs
```diff
@@ -178,91 +180,94 @@ public void SetLabel (Gtk.Widget page, Xwt.Drawing.Image icon, string label)
...
+			tabIcon.Image = icon;
+
+			if (!string.IsNullOrEmpty (label)) {
+				labelWidget.Parent.Show ();
+				labelWidget.Name = label;
+				btnDock.Name = string.Format ("btnDock_{0}", labelNoSpaces ?? string.Empty);
+				btnClose.Name = string.Format ("btnClose_{0}", labelNoSpaces ?? string.Empty);
+
+				string realLabel, realHelp;
+				if (string.IsNullOrEmpty (label)) {
```
slluis: This will always be false since it is inside `if (!string.IsNullOrEmpty (label))...`

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockItemTitleTab.cs
```diff
@@ -178,91 +180,94 @@ public void SetLabel (Gtk.Widget page, Xwt.Drawing.Image icon, string label)
...
+				}
+				else {
+					realLabel = GettextCatalog.GetString ("Close {0}", label);
+					realHelp = GettextCatalog.GetString ("Close the {0} pad", label);
+				}
+				btnClose.Accessible.SetLabel (realLabel);
+				btnClose.Accessible.Description = realHelp;
+			}
+			else
+				labelWidget.Parent.Hide ();
```
slluis: The UpdateVisualStyle method does `labelWidget != null` to determine if pad has a label or not. This change will break the check.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/ISyntaxHighlighting.cs
```diff
@@ -51,7 +51,7 @@ public HighlightedLine (ISegment textSegment, IReadOnlyList<ColoredSegment> segm
 	/// <summary>
 	/// The basic interface for all syntax modes
 	/// </summary>
-	public interface ISyntaxHighlighting
+	public interface ISyntaxHighlighting : IDisposable
```
slluis: This is an api breaking change. If we really need this for the optimizations, we need a PR for d15-2 ASAP.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/ScopeStack.cs
```diff
@@ -52,30 +52,34 @@ public sealed class ScopeStack : IEnumerable<string>
...
 			Count = 1;
 		}
 
-		ScopeStack (string first, ImmutableStack<string> immutableStack, int count)
+		ScopeStack (string first, ImmutableStack<string> immutableStack, int count, ScopeStack parent)
 		{
 			this.FirstElement = first;
 			this.stack = immutableStack;
 			this.Count = count;
+			this.parent = parent;
```
slluis: Please add a comment explaining why we keep 'parent' in memory.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Features/Completion/CompletionResult.cs
Therzok: Yeah, I had the IsOverload and GetIsOverloadHashcode approach initially, but I was thinking that having this hashcode as a base would be enough, since I'm not sure how an overload would have a different name. So, choosing the bucket would come from the display text (which should be the same for all overloads), right?

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Features/SemanticHighlighting/SemanticHighlightingVisitor.cs
Therzok: We already check IsVar above. This check means 'if `var` binds to a type or not`, which is always true.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/ISyntaxHighlighting.cs
Therzok: This already is in d15-2

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Parser/CSharpParsedDocument.cs
Therzok: Sure.

Therzok:
Feedback fixed.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Features/Completion/CompletionResult.cs
Therzok: But yes, I can do that. If someone matches overloads by something other than DisplayText.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/ScopeStack.cs
Therzok: Fixed above on the constructor.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.ClassOutline/CSharpOutlineTextEditorExtension.cs
```diff
@@ -247,8 +245,9 @@ static void OutlineTreeIconFunc (TreeViewColumn column, CellRenderer cell, TreeM
 			}
 		}
 
-		void OutlineTreeTextFunc (TreeViewColumn column, CellRenderer cell, TreeModel model, TreeIter iter)
+		static void OutlineTreeTextFunc (TreeViewColumn column, CellRenderer cell, TreeModel model, TreeIter iter)
 		{
+			var astAmbience = new AstAmbience (TypeSystemService.Workspace.Options);
```
mkrueger: Is it really better to recreate this object every time instead of re-using it ? How about a Lazy<AstAmbience> ?

Therzok: Adding in a lazy would still require an instance reference, which is what I'm avoiding here.

Given that AstAmbience is now a struct, this shouldn't have any impact.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Features/SemanticHighlighting/SemanticHighlightingVisitor.cs
```diff
@@ -529,21 +529,15 @@ public override void VisitIdentifierName(Microsoft.CodeAnalysis.CSharp.Syntax.Id
 			base.VisitIdentifierName(node);
 			ISymbol symbol = null;
 			if (node.IsVar) {
+				// We don't need to lookup symbol information here, as var is colored as a keyword.
```
mkrueger: Why you don't need that ?

class var {} 

....

foreach (var a in X) {}

is no keyword

Therzok: Oh, okay, I'll revert this change then

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.Wrappers/SemanticHighlightingSyntaxMode.cs
```diff
@@ -162,7 +163,15 @@ public void Dispose()
...
+					Tuple<IDocumentLine, HighlightingSegmentTree> tree = null;
+
+					// This code should not have any lambda capture linq, as it is a hot loop.
+					foreach (var segment in lineSegments) {
+						if (segment.Item1 == line) {
```
mkrueger: Won't work with the new editor infrastructure. Never compare lines with == now. Lines can change between getline calls.

Therzok: This is just unrolling the first or default. I'll fix this, but it's not caused by this code.

Therzok: So, how should I fix this?

PR #2357, opened 2017-05-04
Fixed 'Bug 55947 - VSfMac hangs when searching an API in Assembly

mkrueger:
Browser'

It was no hang it was just slow.


slluis:
Let's skip 15.2, and merge for 15.3.

PR #2359, opened 2017-05-04
[F#] Fix brace matcher when matching on parens

nosami:
Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=55916


Therzok:
Next time, please make the commit to extract the code into a different method into its own commit. HAd a hard time picking up the actual changes.

nosami:
It wasn't an extract method, as it was rewritten to make fewer LocationToOffset calls

PR #2360, opened 2017-05-04
[F#] Fix brace matcher when matching on parens

nosami:
Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=55916


slluis:
Too late for 15.2.

PR #2364, opened 2017-05-04
Check if keybinding file exists before opening it

KirillOsenkov:
This avoids a first-chance exception on startup when we try to read from a file that doesn't exist.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/KeyBindingService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/KeyBindingService.cs
@@ -168,7 +168,12 @@ namespace MonoDevelop.Components.Commands
 		public static void LoadCurrentBindings (string defaultSchemaId)
 		{
 			XmlTextReader reader = null;
-			
+
+			if (!File.Exists (ConfigFileName)) {
+				ResetCurrent (defaultSchemaId);
+				return;
+			}
+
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/KeyBindingService.cs
```diff
@@ -168,7 +168,12 @@ internal static string GetCommandKey (Command cmd)
 		public static void LoadCurrentBindings (string defaultSchemaId)
 		{
 			XmlTextReader reader = null;
-			
+
+			if (!File.Exists (ConfigFileName)) {
```
Therzok: I'd argue that IO is slower than handling an exception which does the same thing. Depends on where we want to end up.

KirillOsenkov: If you're optimizing for performance, perhaps. But if you're optimizing for debuggability of the startup path, File.Exists wins hands down :)

Also keep in mind that a single extra File.Exists at the startup path won't tip the perf scale in any way, compared to what I just saw we're doing everywhere :)

Therzok: :+1: okay, sure, I was just mentioning that the exception is handled below.

PR #2373, opened 2017-05-05
Fixed 'Bug 55987 - Completion on XML closing tags is broken'

mkrueger:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/ListWindow.cs
```diff
@@ -375,12 +375,13 @@ public KeyActions PostProcessKey (KeyDescriptor descriptor)
 					var selectedItem = list.SelectedItemIndex;
 					if (selectedItem < 0 || selectedItem >= DataProvider.ItemCount)
 						return KeyActions.CloseWindow;
-					var text = DataProvider.GetText (selectedItem);
-					if (!text.Substring (0, Math.Min (text.Length, CurrentPartialWord.Length)).EndsWith (descriptor.KeyChar.ToString (), StringComparison.Ordinal)) {
-						return KeyActions.Process | KeyActions.CloseWindow;
```
slluis: why was this added and why remove it now?

mkrueger: Has historical reasons - our completion engine was built for c# with some xml hacks.  That was a best fit for completing with punctuation. If a punctuation like '/' or so is in the display string it should not complete. 

During the webtools work some of the hack stuff was removed because they needed some more special casing. The webtools work was mostly to move responsibility of the list handling to the backends. Some of the xml special casing seems not to work anymore so I switched the xml extension to the new commit char system and that code interferes with the backend processing.

PR #2375, opened 2017-05-05
Use a ignore case comparer for propertiesWithTransforms

Therzok:


mrward:
It looks like the line endings have changed so it looks like the entire file has changed.

PR #2379, opened 2017-05-07
Mono todo optimization

Therzok:


Therzok:
Something's weird, as in `System.MonoTODOAttribute` is not found. Have to investigate why.

Therzok:
Fixed. Seems like `GetTypeByMetadataName` returns null if multiple symbols are matched.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Diagnostics/MonoTODO/MonoTODODiagnosticAnalyzer.cs
```diff
@@ -58,15 +59,34 @@ sealed class MonoTODODiagnosticAnalyzer : DiagnosticAnalyzer
...
-				(nodeContext) => {
-					Diagnostic diagnostic;
-					if (TryFindMonoTODO(nodeContext.SemanticModel, nodeContext.Node, out diagnostic, nodeContext.CancellationToken))
-						nodeContext.ReportDiagnostic (diagnostic);
-				},
-				syntaxKindsOfInterest);
+			context.RegisterCompilationStartAction (compilationContext => {
+				var compilation = compilationContext.Compilation;
+				var monoTodoAttributeExists = GetSearchAssemblies (compilation)
+				                                   .Any (assemblySymbol => assemblySymbol.GetTypeByMetadataName (MonoTODOAttributeName) != null);
```
mkrueger: Would maybe make more sense to filter for assembly name ?

Therzok: MonoTODOAttribute is shipped in every bcl assembly that uses it.

Therzok: Also, it's not really slow, we already bind all symbols because of the error diagnostics.

slluis:
@mkrueger do you approve?

PR #2391, opened 2017-05-08
Initial work to support SDK resolvers

slluis:
Changes done so far:
* MD.Core now references the MSBuild assemblies from the msbuild dir. The assemblies need to be explicitly loaded since they are not in the gac.
* I copied SdkResolution.cs and related files from MSBuild and adapted it to MD. Paths in BuildEnvironmentHelper are harcoded, that needs to be fixed.
* DefaultMSBuildEngine now gets the sdk path from SdkResolution. This is working correctly.
* I commented out the line that sets the value of MSBuildSDKsPath as an env var in ProjectBuilder.v4.0.cs. If that env var is not defined, MSBuild should be using the resolvers to find the SDKs, but this is not working, resolvers are not found.

We can't merge this because the last bit doesn't work, and builds fail. This could be because MD loads the msbuild libraries from a temp directory.


radical:
The first issue is that msbuild is looking for `SdkResolvers/` next to the other assemblies in the builder directory. Copying that over (https://gist.github.com/radical/bdc824ceafddb36c162805fdee7144cd) gets to the next issue:
```
Mono: DllImport error loading library '/Users/ankit/Library/Caches/MonoDevelop/7.0/MSBuild/3303_1/SdkResolvers/Microsoft.DotNet.MSBuildSdkResolver/libhostfxr.dylib': 'dlopen(/Users/ankit/Library/Caches/MonoDevelop/7.0/MSBuild/3303_1/SdkResolvers/Microsoft.DotNet.MSBuildSdkResolver/libhostfxr.dylib, 9): no suitable image found.  Did find:
        /Users/ankit/Library/Caches/MonoDevelop/7.0/MSBuild/3303_1/SdkResolvers/Microsoft.DotNet.MSBuildSdkResolver/libhostfxr.dylib: mach-o, but wrong architecture
```

It's a 64bit library and the builder is being run with `mono32`, which is because of https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/MonoTargetRuntime.cs#L350-L351 .
```
			if ((peKind & (IKVM.Reflection.PortableExecutableKinds.Required32Bit | IKVM.Reflection.PortableExecutableKinds.Preferred32Bit)) != 0) {
				monoPath = Path.Combine (MonoRuntimeInfo.Prefix, "bin", "mono32");
```

Switching the builder to be built for x64, fixes the above issue and I don't get any error about not being able to find a SDK. But I'm not sure whether we should switch
the builder to 64bit, at this last moment.

mhutch:
@radical is `msbuild` running 64bit?

radical:
@mhutch Yeah, good point. `msbuild` script just uses the default `mono`, which is a 64bit binary on `5.2.0`. So, we are running `msbuild` in a 64bit process, in the command line case

mrward:
The MSBuildSDKsPath is used in the [MSBuildEvaluationContext](https://github.com/mono/monodevelop/blob/3394cea4b87614eab7b10eb56b5dbecfd1def4d5/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildEvaluationContext.cs#L113) and the [MSBuildProjectService](https://github.com/mono/monodevelop/blob/a21dc0de5a6296dc4e2b0a0f9fcfd38379fb86c6/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs#L1352). 
The MSBuildSDKsPath here would point to Mono's MSBuild Sdks path instead of the path found by the resolvers. Should this be using the same path that the resolver is using?

The MSBuildSDKsPath property is used by some of the Microsoft.NET.Sdk.Web sdk files so I would guess this would result in the wrong files would be imported for .NET Core 2.0.

slluis:
My testing shows that MSBuild always assigns Mono's MSBuild Sdks path to MSBuildSDKsPath. This looks wrong.

slluis:
@radical I did the changes you proposed: SdkResolvers dir is now copied over (https://github.com/mono/monodevelop/pull/2391/commits/004cf4fc473efeff6454cd7d3928a017648bd7e8#diff-7d47d1b630cad3bebf76466c5c32107eR1316), and builders are now built as 64bit apps (https://github.com/mono/monodevelop/pull/2391/commits/004cf4fc473efeff6454cd7d3928a017648bd7e8#diff-1901772d7d7f30b6dd7e0f13923b4e45R22).

The log shows that the resolver and its native dll are correctly found and loaded. However, I still get the same build error. Could you please check this out and see if it works for you? did I miss anything?

radical:
@slluis it is working for me. I can see it using targets from dotnet like 

`Target "_ComputeLockFileCopyLocal" in file "/usr/local/share/dotnet/sdk/2.0.0-preview1-005972/Sdks/Microsoft.NET.Sdk/build/Microsoft.PackageDependencyResolution.targets":`

I tried it with:
```

<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.0</TargetFramework>
  </PropertyGroup>

</Project>
```

mrward:
The MSBuildSDKsPath property pointing to Mono's MSBuild Sdks directory might not be a problem, at least not for .NET Core 2.0. The .NET Core 2.0 preview 1 sdks do not seem to use it in any of their MSBuild targets.

The sdk-resolvers branch seems OK to me for netstandard1.0 and 2.0 and console projects that target netcoreapp1.0 and 2.0.

ASP.NET Core projects seem to trigger a stackoverflow exception if you create or open them.

mrward:
It looks like the 1.0.4 and 2.0 preview 1 sdks are using a new type of import in the Microsoft.NET.Sdk.Web/Sdk/Sdk.props which is not currently supported. Instead the Sdk.props file is imported repeatedly until the stack overflows.

Microsoft.NET.Sdk.Web/Sdk/Sdk.props: 
```
<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">

  <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.props" />
  
  <Import Sdk="Microsoft.NET.Sdk.Web.ProjectSystem" Project="Sdk.props" />
  
  <Import Sdk="Microsoft.NET.Sdk.Publish" Project="Sdk.props" />
  
</Project>
```

slluis:
Those commits should fix all pending issues. It add support for Sdk attribute in imports and it adds an internal MonoDevelop resolver for SDKs registered by addins. I'll add some unit tests.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
```diff
@@ -481,6 +481,7 @@ static void unixSetProcessName (string name)
 
 		static void LoadMSBuildLibraries ()
 		{
+			// Explicitly load the msbuild libraries since they are not installed in the GAC
 			var path = systemAssemblyService.CurrentRuntime.GetMSBuildBinPath ("15.0");
 			System.Reflection.Assembly.LoadFrom (System.IO.Path.Combine (path, "Microsoft.Build.dll"));
```
mhutch: Would it make sense for this to use Platform.AssemblyLoad?

mhutch:
Any chance I could get a public APIs for enumerating SDKs and resolving SDKs, for the MSBuild extension?

main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
slluis: Yeah

slluis:
Sure, I can add some public api for that.

mhutch:
Thanks!

Something like this would be great:

```
string GetMSBuildSdkPath (string sdkID);
IEnumerable<string> GetMSBuildSdkIDs ();
```

slluis:
@mhutch I just realized that I can't easily implement that API. The SDK resolvers don't provide information about which SDKs are available, they just return the path for a given SDK.

mhutch:
@slluis yes, I realizes it wouldn't be complete, but we have the info for a subset - the ones in MD and in the core SDK directory

slluis:
Ok, I think I fixed the unit test failures. There we two issues:
1) The custom MD resolver was not exposed as a public class in release builds.
2) The whole Sdks folder needs to be copied to the local builder directory, otherwise msbuild won't
find it and it won't use the Mono shipped targets when the .net core ones are not found.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.MSBuildResolver/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,51 @@
+﻿//
+// AssemblyInfo.cs
+//
+// Author:
+//       lluis <>
+//
+// Copyright (c) 2017 ${CopyrightHolder}
+//
```
mrward: Another missing copyright here.

slluis:
Copyright headers fixed.

PR #2392, opened 2017-05-08
[C#] Optimize parameter hinting handling.

Therzok:
This change fixes parameter hinting so that it works similar to how Roslyn handles it. This will do for now, until we switch to roslyn's ISignatureHelpProvider

Best viewed side-by-side, as nothing of the old code remains, uses the same mechanism as here: http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp.Features/SignatureHelp/InvocationExpressionSignatureHelpProvider.cs,54


mkrueger:
IMO we can use the roslyn parameter hinting - now we've internals access there is nothing that prevents it.

Why not switch to the roslyn version - if you invest work in that ?

slluis:
Let's make a lane to verify test are green.

Therzok:
Lane created.

Therzok:
I have a patch which started the work to switch to Roslyn's parameter hinting engine. https://github.com/mono/monodevelop/commits/swap-parameter-hinting

Therzok:
Tests are green.

PR #2395, opened 2017-05-09
Bug 55875 - Call Stack Disappears when a thread selection is made whi…

DavidKarlas:
Backport from d15-2 branch


slluis:
Please PR for d15-3-preview.

PR #2406, opened 2017-05-10
[F#] Adds tagged comments to the Tasks pad.

nosami:
Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=55396

### DIFF
--- a/main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpParsedDocument.fs
+++ b/main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpParsedDocument.fs
@@ -40,10 +40,9 @@ type FSharpParsedDocument(fileName, location: DocumentLocation option) =
                         tokens 
                         |> List.mapi (fun line tokenList -> line+1, tokenListToComment tokenList)
                         |> List.choose(fun (line, s) ->
-                                            let tag =
-                                                specialCommentTags 
-                                                |> Set.tryFind(fun tag -> s.StartsWith tag)
-                                            tag |> Option.bind(fun t -> Some <| Tag(t, s, DocumentRegion(line, 1, line, 1))))
+                                            specialCommentTags 
+                                            |> Set.tryFind(fun tag -> s.StartsWith tag)
+                                            |> Option.bind(fun t -> Some <| Tag(t, s, DocumentRegion(line, 1, line, 1))))
 
                     return ResizeArray(tokensByLine) :> IReadOnlyList<_>
                 | None -> return ResizeArray() :> IReadOnlyList<_>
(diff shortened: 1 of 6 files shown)


Therzok:
This PR unbumps xwt.

nosami:
Oops. Fixed

main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpParsedDocument.fs
```diff
@@ -1,25 +1,54 @@
...
     member x.ParsedLocation = location
 
     member val UnusedCodeRanges : Range.range list option = None with get, set
 
+    override x.GetTagCommentsAsync(cancellationToken) =
+        let tokenListToComment (tokenList: FSharpTokenInfo list, lineText: string) =
+            let comment = 
+                tokenList
+                |> List.filter(fun t -> t.CharClass = FSharpTokenCharKind.LineComment)
+                |> List.fold(fun acc token -> acc + lineText.[token.LeftColumn..token.RightColumn]) ""
```
Therzok: won't this create many intermediate strings? Wouldn't string.Join do a better job here?

nosami: Good catch - fixed

main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpParsedDocument.fs
```diff
@@ -1,25 +1,56 @@
...
 
     member val UnusedCodeRanges : Range.range list option = None with get, set
 
+    override x.GetTagCommentsAsync(cancellationToken) =
+        let tokenListToComment (tokenList: FSharpTokenInfo list, lineText: string) =
+            let comment = 
+                tokenList
+                |> List.filter(fun token -> token.CharClass = FSharpTokenCharKind.LineComment)
+                |> List.map(fun token -> lineText.[token.LeftColumn..token.RightColumn])
+                |> String.concat ""
```
Therzok: Which overload of string.Concat is being used? If it's [this one](https://msdn.microsoft.com/en-us/library/dd991828(v=vs.110).aspx), we're still going to allocate a lot of strings.

nosami:
```fs
        [<CompiledName("Concat")>]
        let concat sep (strings : seq<string>) =  
            String.Join(sep, strings)
```

Therzok: We need one which goes via an array variant: https://github.com/mono/mono/blob/master/mcs/class/referencesource/mscorlib/system/string.cs#L3285

Otherwise, we'd be creating strings on each iteration. Doing `ToArray()` outweighs the allocations done by strings.

Therzok: https://github.com/mono/mono/blob/master/mcs/class/referencesource/mscorlib/system/string.cs#L122

That overload seems to use a stringbuilder. So that's worse than having another array allocated before calling the array variant.

PR #2417, opened 2017-05-11
Bump required Mono version to 5.2 in monostub.m

KirillOsenkov:


decriptor:
I admit that I didn't do this when bumping to 5.0, but would it be worth adding a comment in the commit as to why a bump is required or even details/description here?

I'm assuming this is because of the MEF fix in mono?

KirillOsenkov:
Yup, a fix that unblocks VSMEF.

Would you like the comment directly in the source?

decriptor:
I don't know. I was mostly just throwing it out there for comment.

My thought was this way someone doesn't wonder why there was a bump and try to move it back for some reason.

I just happened to see your PR to bump mono in the other repo :)

PR #2430, opened 2017-05-15
[A11y] Make the property grid fully keyboard accessible

iainx:
Tab/Shift-Tab moves through the rows and <space> collapses/expands rows

Fixes BXC #53711


iainx:
This isn't complete, so I'm going to close it

PR #2431, opened 2017-05-15
D15 2 fixes

mkrueger:


mkrueger:
Bug:

https://bugzilla.xamarin.com/show_bug.cgi?id=56446

slluis:
@mkrueger would it be possible to add unit tests for this bug?

slluis:
This actually seems to fix two bugs: 56446 and 56257.

PR #2434, opened 2017-05-15
[D15.3][VCS] Don't set tabs to spaces due to default policy in the diff editor.

Therzok:
When setting the local text, the text setter would end up formatting the tabs to spaces due to how the policy was set. Disable that setting.


slluis:
Closing since I'll merge master to the preview branch.

PR #2439, opened 2017-05-16
Fixed 'Bug 56453 - Fold #regions by default '

mkrueger:
In older versions the text editor & ide fold segment data structures
were different - no need to do the conversion anymore. IsCollapsed
wasn't converted, that's why it didn't work anymore.

### DIFF
--- a/main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Document/TextDocument.cs
+++ b/main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Document/TextDocument.cs
@@ -1251,7 +1251,7 @@ namespace Mono.TextEditor
 		object foldSegmentTaskLock = new object ();
 		Task foldSegmentTask;
 
-		public void UpdateFoldSegments (List<FoldSegment> newSegments, bool startTask = false, bool useApplicationInvoke = false, CancellationToken masterToken = default(CancellationToken))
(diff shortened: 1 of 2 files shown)


main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Document/TextDocument.cs
```diff
@@ -1303,11 +1303,12 @@ void RemoveFolding (FoldSegment folding)
...
-		HashSet<FoldSegment> UpdateFoldSegmentWorker (List<FoldSegment> newSegments, out bool update, CancellationToken token = default(CancellationToken))
+		HashSet<FoldSegment> UpdateFoldSegmentWorker (IEnumerable<IFoldSegment> segments, out bool update, CancellationToken token = default(CancellationToken))
 		{
 			var oldSegments = new List<FoldSegment> (FoldSegments);
 			int oldIndex = 0;
 			bool foldedSegmentAdded = false;
+			var newSegments = segments.ToList ();
 			newSegments.Sort ();
 			var newFoldedSegments = new HashSet<FoldSegment> ();
 			foreach (FoldSegment newFoldSegment in newSegments) {
```
Therzok: type in `segments` is IFoldSegment`, yet the loop takes `FoldSegment`.

PR #2440, opened 2017-05-16
[D15.2][VCS] Don't set tabs to spaces due to default policy in the diff editor.

Therzok:
When setting the local text, the text setter would end up formatting the tabs to spaces due to how the policy was set. Disable that setting.


Therzok:
QA approved.

PR #2443, opened 2017-05-16
[Build] Introduce new `is-preview` command to configure.sh

sgmunn:


sgmunn:
https://github.com/xamarin/md-addins/pull/1906

PR #2444, opened 2017-05-16
Make CompletionData comparer stable.

KirillOsenkov:
Instead of having a DataItemComparer externally to CompletionData, have a CompletionData.Comparer singleton and use that.

Make sure all logic is concentrated in one place and does comparisons one step at a time.

The whitespace changes are the result of unifying the entire file to use LF consistently.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionData.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionData.cs
@@ -159,7 +159,13 @@ namespace MonoDevelop.Ide.CodeCompletion
 		{
 			int IComparer<CompletionData>.Compare (CompletionData a, CompletionData b)
 			{
-				return Compare (a, b);
+				if (a == b)
+					return 0;
+				if (a != null && b == null)
+					return -1;
+				if (a == null && b != null)
+					return 1;
+				return a.CompareTo (b);
 			}
 		}
 
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionData.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionData.cs
@@ -39,7 +39,7 @@ namespace MonoDevelop.Ide.CodeCompletion
 	public class CompletionData : IComparable
 	{
 		protected CompletionData () {}
-		
+
 		public virtual IconId Icon { get; set; }
 		public virtual string DisplayText { get; set; }
 		public virtual string Description { get; set; }
@@ -112,7 +112,7 @@ namespace MonoDevelop.Ide.CodeCompletion
 		public CompletionData (string text) : this (text, null, null) {}
 		public CompletionData (string text, IconId icon) : this (text, icon, null) {}
 		public CompletionData (string text, IconId icon, string description) : this (text, icon, description, text) {}
-		
+
 		public CompletionData (string displayText, IconId icon, string description, string completionText)
 		{
 			this.DisplayText = displayText;
@@ -140,7 +140,7 @@ namespace MonoDevelop.Ide.CodeCompletion
 			var currentWord = GetCurrentWord (window, descriptor);
 			window.CompletionWidget.SetCompletionText (window.CodeCompletionContext, currentWord, CompletionText);
 		}
-		
+
 		public override string ToString ()
 		{
 			return string.Format ("[CompletionData: Icon={0}, DisplayText={1}, Description={2}, CompletionText={3}, DisplayFlags={4}]", Icon, DisplayText, Description, CompletionText, DisplayFlags);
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionData.cs
```diff
@@ -150,30 +150,54 @@ public override string ToString ()
 
 		public virtual int CompareTo (object obj)
 		{
-			if (!(obj is CompletionData))
-				return 0;
```
KirillOsenkov: I think the issue was here - we were saying "I'm equal to null" when compared to a null item. Not sure if this was actually ever hit though.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionData.cs
```diff
@@ -39,7 +39,7 @@ namespace MonoDevelop.Ide.CodeCompletion
 	public class CompletionData : IComparable
 	{
 		protected CompletionData () {}
-		
+
```
KirillOsenkov: Sorry I formatted the document, I can undo whitespace changes if needed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/ListWidget.cs
```diff
@@ -797,19 +797,7 @@ static CompletionListFilterResult DefaultFilterWords (ICompletionDataList dataLi
...
 		}
 
-
-		class DataItemComparer : IComparer<CompletionData>
-		{
-			public int Compare (CompletionData a, CompletionData b)
-			{
-				if (a == b)
-					return 0;
-				if (a is IComparable && b is IComparable)
```
Therzok: This breaks the case where completion data defines it's own comparison methods.

See this for example: https://github.com/mono/monodevelop/blob/master/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/RoslynSymbolCompletionData.cs#L441-L467

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionData.cs
```diff
@@ -150,30 +150,54 @@ public override string ToString ()
...
-			if (result == 0) {
-				var aIsImport = (a.DisplayFlags & DisplayFlags.IsImportCompletion) != 0;
-				var bIsImport = (b.DisplayFlags & DisplayFlags.IsImportCompletion) != 0;
-				if (!aIsImport && bIsImport)
-					return -1;
-				if (aIsImport && !bIsImport)
-					return 1;
-				if (aIsImport && bIsImport)
-					return StringComparer.Ordinal.Compare (((CompletionData)a).Description, ((CompletionData)b).Description);
-				var ca = a as CompletionData;
```
KirillOsenkov: This was probably a remnant from older code, but now the cast is not needed since they're already CompletionData in the first place.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionData.cs
```diff
@@ -205,13 +229,13 @@ public virtual bool OverloadGroupEquals (CompletionData other)
 			return DisplayText == other.DisplayText;
 		}
 
-		public virtual int GetOverloadGroupHashCode ()
-		{
-			return DisplayText.GetHashCode ();
+		public virtual int GetOverloadGroupHashCode ()
+		{
+			return DisplayText.GetHashCode ();
```
KirillOsenkov: Unified line breaks to LF while I was here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/ListWidget.cs
```diff
@@ -830,7 +818,7 @@ internal void SelectEntry (CompletionSelectionStatus match)
 		internal static IComparer<CompletionData> GetComparerForCompletionList (ICompletionDataList dataList)
 		{
 			var concrete = dataList as CompletionDataList;
-			return concrete != null && concrete.Comparer != null ? concrete.Comparer : new DataItemComparer ();
+			return concrete != null && concrete.Comparer != null ? concrete.Comparer : CompletionData.Comparer;
```
KirillOsenkov: Avoid allocating it.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/ListWidget.cs
KirillOsenkov: Ouch. Good catch. I didn't realize CompletionData has derived classes. Fixing...

KirillOsenkov: Fixed.

KirillOsenkov:
I found another place where I think we were unstable: 
http://source.monodevelop.com/#MonoDevelop.CSharpBinding/MonoDevelop.CSharp.Completion/RoslynSymbolCompletionData.cs,454

mkrueger:
There is only 1 use case for that comparison: code template completion items - seems to work happens in every completion list. There are no other items than roslynsymbolcompletion datas beside that in the c# case.

KirillOsenkov:
Sorry Mike I'm trying to parse `"seems to work happens in every completion list."` - what do you mean? ;)

@Therzok - how's it looking now?

PR #2446, opened 2017-05-17
Fixed 'Bug 56541 - Attempting to use "_" as a tuple-discard value

mkrueger:
auto-completes instead'


slluis:
It would be nice to have a unit test

mkrueger:
It's not really value changing - I'll replace our completion engine after 15.3 with the roslyn one then we get rid of maintaining a completion engine.

PR #2449, opened 2017-05-17
[Ide] Adds feature in MonoDevelop to pin opened tabs

netonjm:
Adds feature in MonoDevelop to pin opened tabs
This feature allows to lock some of your most used windows (tabs)
to easily detect and open like VS2013 added.
- Adds Pin/Unpin tab feature clicking right button (contextual menu) in tab or Shortcut key
- Adds feature enabled based in file: ~/enable-pin-tabs


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.DockNotebook/TabStrip.cs
```diff
@@ -855,9 +886,17 @@ void DrawTab (Context ctx, DockNotebookTab tab, Gdk.Rectangle allocation, Gdk.Re
 			
 			tab.CloseButtonActiveArea = closeButtonAlloation.Inflate (2, 2);
 
+			var spinButtonAllocation = new Cairo.Rectangle (closeButtonAlloation.X - rightPadding - (tabPinnedImage.Width / 2) - PinButtonMarginRight,
```
sevoku: I think `rightPadding` shouldn't be substracted here. We want to allocate the Pin left of the close button with a distance of `PinButtonMarginRight`. `rightPadding` is the padding of the whole content inside the tab and its already included in `closeButtonAlloation.X`.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.DockNotebook/TabStrip.cs
```diff
@@ -867,11 +906,17 @@ void DrawTab (Context ctx, DockNotebookTab tab, Gdk.Rectangle allocation, Gdk.Re
...
+			if (drawPinButton)
+				ctx.DrawImage (this, (tab.IsPinned ? tabPinnedImage : tabUnPinnedImage).WithAlpha ((pinButtonHovered ? 1.0 : 0.5) * tab.Opacity), spinButtonAllocation.X, spinButtonAllocation.Y);
+
 			// Render Text
 			double tw = tabBounds.Width - (leftPadding + rightPadding);
 			if (drawCloseButton || tab.DirtyStrength > 0.5)
 				tw -= closeButtonAlloation.Width / 2;
 
+			if (drawPinButton || tab.DirtyStrength > 0.5)
+				tw -= spinButtonAllocation.Width / 2;
```
sevoku: This can be optimized, atm it limits the text width still based on the `closeButtonAlloation`:
grafik

 I think its better to use `tw -= (tabBounds.Width - spinButtonAllocation.Right) + (spinButtonAllocation.Width / 2)`, this will substract the space between the right corner of the tab and the right corner of the pin button + half of its width.

sevoku:
Also @vancura might want to review the assets:
grafik

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.DockNotebook/TabStrip.cs
sevoku: oh, its different, we always show both buttons on hover, so it should be an if/elseif, if only the close button is visible, then L915, otherwise what I wrote before.

vancura:
I'll check this out when I have time (I don't currently, I will travel to SF/Redmond in a couple of hours, so later).

But for now, have you seen this? For the record:

https://trello.com/c/1XxzqYi8/29-c8-tentative-xs-transient-editor-buffers
https://trello.com/c/iTrk8Lnv/12-feature-request-add-right-side-temporary-buffer-tab
https://trello.com/c/cpA9Ud6l/651-document-tab-pinning?menu=filter&filter=native

And we had a couple of documents back on Xamarin Google Docs I don't have anymore (obviously), but maybe we can recover them, somehow.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdePreferences.cs
```diff
@@ -86,6 +86,11 @@ internal IdePreferences ()
 		public readonly ConfigurationProperty<bool> DefaultHideMessageBubbles = ConfigurationProperty.Create ("MonoDevelop.Ide.DefaultHideMessageBubbles", false);
 		public readonly ConfigurationProperty<ShowMessageBubbles> ShowMessageBubbles = ConfigurationProperty.Create ("MonoDevelop.Ide.NewShowMessageBubbles", MonoDevelop.Ide.ShowMessageBubbles.ForErrorsAndWarnings);
 
+		public static Lazy<bool> TabFeatureEnabled = new Lazy<bool> (() => {
+			var path = System.IO.Path.Combine (Environment.GetFolderPath (Environment.SpecialFolder.Personal), "enable-pin-tabs");
```
mhutch: Please use the standard preferences mechanism, not a custom file in the home directory.

mhutch:
I think the feature is ok to include and enable by default. It's present in VS2017.

The transient tab feature is something we still need to do, but it's separate from this. 

The option to enable/disable this feature should be a normal PropertyService property, not a custom mechanism.

I think the tab color may be a little too much. In VS, pinned tabs don't change color at all.

Therzok:
> I think the tab color may be a little too much. In VS, pinned tabs don't change color at all.

Maybe we should always draw the pin icon?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.DockNotebook/DockNotebook.cs
```diff
@@ -41,6 +42,12 @@ namespace MonoDevelop.Components.DockNotebook
 	class DockNotebook : Gtk.VBox
 	{
 		List<DockNotebookTab> pages = new List<DockNotebookTab> ();
+		List<DockNotebookTab> stickedPages {
```
Therzok: Please use `UpperCamelCase` for properties

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.DockNotebook/DockNotebook.cs
```diff
@@ -328,6 +338,20 @@ void UpdateIndexes (int startIndex)
 				((DockNotebookTab)pages [n]).Index = n;
 		}
 
+		void OnTabPinned (DockNotebookTab sender, bool value)
+		{
+			if (pages.Count == 1)
+				return;
+			if (value) {
+				if (stickedPages.Count > 0)
```
Therzok: These operations are not a simple getter, they actually compute the ordering. On each of the branches, we end up computing `stickedPages` 3 times, and `normalPages` twice or one time.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.DockNotebook/DockNotebook.cs
Therzok: Actually, maybe provide this information as local variables inside `OnTabPinned`. That way, we won't have this used anywhere where it's not needed.

mhutch:
On pinned tabs? Yeah, That's what VS does.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
```diff
@@ -83,7 +85,77 @@ protected override void Run ()
...
+			var deleteCache = new System.Collections.Generic.List<Document> ();
+			var w1 = (SdiWorkspaceWindow)active.Window;
+			foreach (var item in w1.TabControl.Tabs.Where(s => !s.IsPinned)) {
+				var workspaceWindow = item.Content as SdiWorkspaceWindow; 
+				if (workspaceWindow != null && workspaceWindow.Document != null)
+					deleteCache.Add (workspaceWindow.Document);
+			}
+
+			foreach (Document doc in IdeApp.Workbench.Documents.ToArray ()) {
+				if (deleteCache.Exists(d => d == doc))
```
Therzok: `deleteCache.Contains (doc)`

netonjm:
wow! this was a very good feedback thanks to all!

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.DockNotebook/DockNotebook.cs
mhutch: also, s/sticked/pinned/

vancura:
Yes, just a pinned icon please. You'll need to icon I guess? :)

vancura:
Icons: https://github.com/xamarin/icons/commit/e5f980b218f3f86af94b23d44d73d615c792a770

PR #2453, opened 2017-05-18
Fix unit tests.

mkrueger:


main/src/core/MonoDevelop.TextEditor.Tests/Mono.TextEditor.Tests/LineSplitterTests.cs
```diff
@@ -43,7 +43,7 @@ public void TestLastLineCreation ()
 			var buffer = ImmutableText.Empty;
 			LineSplitter splitter = new Mono.TextEditor.LineSplitter ();
 			buffer = buffer.InsertText (0, "1\n2\n3\n");
-			splitter.TextReplaced (null, new TextChangeEventArgs (0, "", buffer.ToString ()));
```
Therzok: You don't need these changes if you fix the public API.

mkrueger: The API was wrong - old & new offsets need to be set for a valid text change

Therzok: Then add new constructors which set new offset to old offset.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Text/TextChangeEventArgs.cs
```diff
@@ -143,7 +143,16 @@ public class TextChangeEventArgs : EventArgs
 		/// <summary>
 		/// Creates a new TextChangeEventArgs object.
 		/// </summary>
-		public TextChangeEventArgs(int offset, int newOffset, string removedText, string insertedText)
+		[Obsolete ("Use TextChangeEventArgs (int offset, int newOffset, string removedText, string insertedText)")]
```
slluis: This is asking to use another constructor that is internal.

mkrueger: This shouldn't be public to begin with. TextChangeEventArgs are created by the editor not user code.

mkrueger: Dev philosophy changes - today I would only make very few things public. That's the only way of keeping code flexible enough :)

PR #2454, opened 2017-05-18
[Core] Made the constructors public.

mkrueger:
That's mostly for consistency.


mkrueger:
We won't do any 'fix' pass for that layer anyways. It'll just get replaced.

PR #2458, opened 2017-05-19
[Core] Add limited support for multi target framework projects

mrward:
Fixed bug #55508 - Support multiple target framework projects
https://bugzilla.xamarin.com/show_bug.cgi?id=55508

SDK style projects that target multiple frameworks can now be opened
and will show source files and dependency information. Currently
the support is limited. The project is treated as though it only has
one target framework which is the first one specified in the
TargetFrameworks property in the project file.

This is not a complete solution but it is better than the existing support. NuGet restore will only consider one target framework which is incorrect. Also it is not currently supported to install a NuGet package conditionally into the supported target frameworks if it is not compatible with all of them.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
@@ -945,6 +945,53 @@ namespace MonoDevelop.Projects.MSBuild
 				return Array.Empty<string> ();
 		}
 
+		/// <summary>
+		/// If an SDK project targets multiple target frameworks then this returns the first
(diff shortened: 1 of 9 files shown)


mrward:
Moved some logic from the project builder onto the Project and DotNetProject classes. There is still logic in the MSBuildProject and the MSBuildProjectInstance.

Looking at the MSBuildProjectInstance there seem to be cases where it is created and evaluated before the Project can set any global variables on the instance, such as MSBuildProject.Evaluate and EvaluateAsync.

slluis:
I still don't like this approach. Let me think a bit about it.

mrward:
Moved the target frameworks logic out of MSBuildProject and MSBuildProjectInstance and into the Project class. Also added a test.

PR #2461, opened 2017-05-21
Restore breakpoints for pinnedwatches

harukamm:
when loading user preferences.
Fixed bug #52527 : Hard crash while debugging simple console app
https://bugzilla.xamarin.com/show_bug.cgi?id=52527
This crash happens because it seems not to restore tracing breakpoints
even though it restores LiveUpdate property when loading user
preferences.

The same crash doesn't happen with this patch, but not useful yet due
to bug #52528


slluis:
@DavidKarlas can you review?

DavidKarlas:
Yep will review both bug fixes today

PR #2462, opened 2017-05-21
[NuGet] Support non .NET Core projects using PackageReference

mrward:
Non .NET Core Sdk style projects that use PackageReference in their
MSBuild project file are now supported. Previously the Solution
window would not show any packages for the project, would not allow
restoring, and would install new NuGet packages using a packages.config
file. Now if the project has a PackageReference the Packages folder
shows the installed packages and can be used to update or install
NuGet packages which will also create PackageReference MSBuild items.
If the project has no PackageReferences then by default a
packages.config file will be created. There is no UI option to opt-in
to using PackageReferences currently so the project file will need
to be edited to include at least one PackageReference before the
default behaviour of using a packages.config file is overridden.

### DIFF
--- a/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/PackageReferenceNuGetProject.cs
+++ b/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/PackageReferenceNuGetProject.cs
@@ -1,0 +1,304 @@
+﻿//
+// PackageReferenceNuGetProject.cs
+//
+// Author:
+//       Matt Ward <matt.ward@xamarin.com>
+//
+// Copyright (c) 2016 Xamarin Inc. (http://xamarin.com)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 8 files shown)


main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/PackageReferenceNuGetProject.cs
```diff
@@ -0,0 +1,304 @@
...
+		{
+			if (project.Items.OfType<ProjectPackageReference> ().Any ())
+				return new PackageReferenceNuGetProject (project);
+
+			return null;
+		}
+
+		public virtual Task SaveProject ()
+		{
+			return project.SaveAsync (new ProgressMonitor ());
```
Therzok: Maybe give a project monitor as a parameter, so people can hook it up

main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/PackageReferenceNuGetProject.cs
```diff
@@ -0,0 +1,304 @@
...
+				project.Items.Remove (packageReference);
+			}
+		}
+
+		public override async Task<bool> UninstallPackageAsync (
+			PackageIdentity packageIdentity,
+			INuGetProjectContext nuGetProjectContext,
+			CancellationToken token)
+		{
+			bool removed = await Runtime.RunInMainThread (() => {
```
Therzok: This pattern seems to be recurring. Maybe refactor it into a method like:
`async Task<bool> ExecuteAndSave (Func<bool> toExecute)`?

main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/PackageReferenceNuGetProject.cs
mrward: The progress is already shown in the status bar when running the restore. Also this method is for unit tests to avoid a real save.

Therzok: Ah, okay 👍

PR #2483, opened 2017-05-25
[NuGetizer3000] Visual Studio on windows compatibility

mrward:
Fixes several NuGetizer 3000 bugs including:

 1. 56701 - Windows created nuspec can not be loaded on a mac
 2. 56706 - Can not create a nuget package with Standard net libraries
 3. 52657 - Creating a new Multiplatform Library requires a
description
 4. Packages folder empty on re-opening NuGet packaging project
 5. Fix Dependencies folder being displayed if latest NuGet.Build.Packaging 0.1.276 NuGet package is installed.
 6. Create MSBuild targets during restore for NuGet packaging project if they are missing.
 7. Update to latest stable NuGet.Build.Packaging 0.1.276 in project templates.

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
@@ -74,9 +74,7 @@ namespace MonoDevelop.DotNetCore
 
 		bool IsDotNetCoreProject (DotNetProject project)
 		{
-			var properties = project.MSBuildProject.EvaluatedProperties;
-			return properties.HasProperty ("TargetFramework") ||
-				properties.HasProperty ("TargetFrameworks");
+			return project.MSBuildProject.Sdk != null;
 		}
 
 		protected override bool OnGetCanReferenceProject (DotNetProject targetProject, out string reason)
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
@@ -74,7 +74,21 @@ namespace MonoDevelop.DotNetCore
 
 		bool IsDotNetCoreProject (DotNetProject project)
 		{
-			return project.MSBuildProject.Sdk != null;
+			return project.MSBuildProject.Sdk != null && HasSupportedFramework (project);
+		}
+
+		/// <summary>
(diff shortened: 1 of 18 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -74,9 +74,7 @@ protected override bool OnGetSupportsFramework (TargetFramework framework)
 
 		bool IsDotNetCoreProject (DotNetProject project)
 		{
-			var properties = project.MSBuildProject.EvaluatedProperties;
-			return properties.HasProperty ("TargetFramework") ||
-				properties.HasProperty ("TargetFrameworks");
+			return project.MSBuildProject.Sdk != null;
```
slluis: Is this really correct? A project may use an Sdk and not be a .net core project.

mrward: The method name is probably a bit misleading. There is another check made for the target framework elsewhere.

There is also another question of whether we should split out Sdk project support from the .NET Core project extension. Currently they are mixed together and probably should be separated out.

mrward: However it does look wrong so I will change it to check the target framework is supported.

mrward:
Restricted the extension so it is only enabled for sdk projects that target .NET Core or .NET Standard.

mrward:
I think restricting the supported frameworks to netcoreapp and netstandard will break the IoT addin since it currently uses net461 as its target framework. So for now I may have to allow .NETFramework.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
mhutch: +1 on separating them out.

mhutch:
@mrward it'll break any non.NET Core project using sdks, surely? That includes Azure functions.

See also https://github.com/mono/monodevelop/commit/2b2c26b588755f1eb953555c2fc7cadb0c9b496e

mrward:
Yeah I think I am going to revert aa14e80 for now. I think we should separate out sdk project file support from the DotNetCoreProjectExtension at some point.

mrward:
Reverted the change restricting the extension.

mhutch:
Agreed, and we should probably move it down to core seeing how Sdk projects are becoming more common.

PR #2487, opened 2017-05-26
[Ide] Adds feature in MonoDevelop to pin opened tabs

netonjm:
Created a new PR because we need rebase:

Adds feature in MonoDevelop to pin opened tabs
This feature allows to lock some of your most used windows (tabs)
to easily detect and open like VS2013 added.

Adds Pin/Unpin tab feature clicking right button (contextual menu) in tab or Shortcut key
Adds enable/disable configuration options

Includes fixes from:
https://github.com/mono/monodevelop/pull/2449


Therzok:
I think we want to drop the pinned images and only use the pin icon, right @vancura ?

vancura:
We can possibly just use the pin-up icon with opacity treatment. I would like to to take a look at the icon once again, I think it could be sharper, but I can't do it here (now in SF), I need my tools.

This icon:

tab-pin-up-9 2x

vancura:
It's good enough for now. I'll take a look at it when I am back.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
```diff
@@ -83,7 +86,70 @@ protected override void Run ()
...
+					doc.Close();
+			}
+		}
+	}
+
+	class PinTabHandler : PinnedCommandHandler
+	{
+		protected override void Update (CommandInfo info)
+		{
+			info.Visible = info.Enabled = DefaultSourceEditorOptions.Instance.EnablePinTabs;
```
sevoku: The command has a key binding and its public visible (in key bindings panel and global search), hence it should be only enabled if there is a document loaded (`IdeApp.Workbench.ActiveDocument != null)`).

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
```diff
@@ -83,7 +86,70 @@ protected override void Run ()
...
+			var tabControl = activeWindow.TabControl;
+			return tabControl.Tabs.FirstOrDefault (item => (item.Content as SdiWorkspaceWindow).Equals (activeWindow));
+		}
+	}
+
+	class CloseAllExceptPinnedHandler : PinnedCommandHandler
+	{
+		protected override void Update (CommandInfo info)
+		{
+			info.Visible = info.Enabled = DefaultSourceEditorOptions.Instance.EnablePinTabs;
```
sevoku: This command should be disabled if there are no open documents (i.e. with `info.Enabled = IdeApp.Workbench.Documents.Count != 0;)`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
Therzok: You can't actually bring up the command without having a document open. This is a context menu on a docknotebook tab.

sevoku: For me its listed in the Global search results without any open documents (just run VSfM and type "pin")

Therzok: I... Forgot about that. :+1: agreed then

netonjm:
Good catch guys! Changes are done! :-)
Maybe we could squash all commits

decriptor:
I had filed a feature request for this a few years back. https://bugzilla.xamarin.com/show_bug.cgi?id=6367

netonjm:
And now comes true!

PR #2489, opened 2017-05-26
[A11y] Prevent a crash when interacting with the text editor

iainx:


main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.cs
```diff
@@ -194,7 +195,17 @@ Rectangle GetFrameForRange (AtkCocoa.Range range)
 
 			int GetLineForIndex (int index)
 			{
-				return Margin.Document.GetLineByOffset (index).LineNumber;
+				if (Margin == null || Margin.Document == null) {
+					Console.WriteLine ("No margin or document");
```
slluis: Don't use Console.WriteLine for logging. Use LoggingService. Also the condition can be simplified to Margin?.Document == null

slluis: Actually, do we need to log this at all?

iainx: Guess we don't really.
Will fix

PR #2491, opened 2017-05-28
Extension Chain optimization

Therzok:


Therzok:
I want to bring https://github.com/mono/monodevelop/pull/2491/commits/1706b02f558efacb778b63bf5f5c0b221e8d9a46 into discussion.

So, the idea of `FindNextImplementation` is to only return an item when there's at least one override/newslot definition or a serializable field/property. The question is _why_. It seems like wasted CPU for something that esentially saying: if the class is derived from the type.

Probably we should also check that the type is not the same, but it seems like pretty much pointless to do that check.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
```diff
@@ -387,15 +387,13 @@ void InitializeExtensionChain ()
...
-			foreach (var e in tempExtensions)
-				e.Dispose ();
+			extensionChain.Dispose ();
 
 			// Now create the final extension chain
 
 			extensions.Reverse ();
 			var defaultExts = CreateDefaultExtensions ().ToList ();
 			defaultExts.Reverse ();
-			extensionChain.SetDefaultInsertionPosition (defaultExts.FirstOrDefault ());
```
Therzok: this was done for no reason, maybe it was supposed to be done on the new extension chain that was created?

Therzok:
Adding DON'T MERGE YET, as I want to add unit tests for this.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
slluis: Yes, it should be called in the final extension chain. Not sure why this is not a problem right now.

Therzok: Pushed a fix.

slluis:
The reflection check is done to avoid unnecessary chained calls. For example, with the current approach, if an extension overrides a single method of DotNetProjectExtension, only methods from DotNetProjectExtension will be chained through this instance. However, all methods from base classes (ProjectExtension, SolutionItemExtension, WorkspaceObjectExtension) will not include this instance in their chain. If we only check that the extension is a subclass, then methods from the classes at the top of the hierarchy will always be chained, since almost all extensions extensions will subclass them.

Therzok:
Hmmm, I guess the best of both worlds then is to cache the methods that get reflected for each extension chain?

slluis:
Yes, we can cache the info.

Therzok:
Will revive by fixing it to maintain the old behaviour and caching the reflection checks.

PR #2492, opened 2017-05-28
[Templates] Allow for project references to be valid too.

Therzok:
Bug 56093 - [Forms] Templates don't appear if there's a project reference with Xamarin.Forms


slluis:
I'm not sure which problem is this trying to solve. What's the use case?

Therzok:
People who work on Xamarin.Forms can no longer use xamarin.forms specific templates because they reference it by project.

Therzok:
Ping.

slluis:
I don't like this solution, it changes the semantics of the condition. Let's use a different condition.

PR #2495, opened 2017-05-29
Fixed 'Bug 56890 - Rename a XAML type doesn't rename the file.'

mkrueger:
Fix is to disable that option in non source code case. Renaming in generated code is complex in general case.

### DIFF
--- a/main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.Rename/RenameItemDialog.cs
+++ b/main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.Rename/RenameItemDialog.cs
@@ -36,6 +36,7 @@ using System.Linq;
 using Microsoft.CodeAnalysis.Text;
 using Microsoft.CodeAnalysis;
 using System.Threading.Tasks;
+using RefactoringEssentials;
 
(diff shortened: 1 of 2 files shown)


DavidKarlas:
1. Why is it checking just FirstOrDefault() location and not all?
2. This is not fixing bug(it should rename source file) but its just hiding option to rename it

mkrueger:
1. That's a good question - I would need use cases for that where it fails. Usually the first is either a real c# file or a generated one. But maybe partial is a problem here - don't know if it's possible in Xaml case. But I'll add checks they don't hurt IMO.
 
2. That's fixing the bug. There are 2 reasons: the type system has no connection to the "original" file so it can't access it. And the 1:1 scenario of mapping type names to file names is just for real source like c#. And not for designer generated files. The generated files usually are not >real< and are only in memory or created during build - so renaming these would not work or they would be overwritten.
The files where they're coming from could be container files like gtk.stetic or so and renaming that to a type name doesn't make sense in the general case.
So the rename option IS just for real c# files.

mkrueger:
Seems there was another bug - the rename option was always showing - but in fact it should only show up if the file is following the naming pattern of file name == class name. Only in that case the rename operation should offer the rename file help for the user :).

PR #2498, opened 2017-05-29
Updated dotnet/templating engine, templates, added functionality and fixed 2 bugs

DavidKarlas:
To fix bug 56777 it was impossible to know launchUrl based on project(like we used to do until now)… Hence now we enable “IncludeLaunchSettings” which we later use to set initial launchUrl…
To get “IncludeLaunchSettings” templates and engine needed to be upgraded and TemplateParameterExtensionNode was added so we can set Parameters as needed by template
There was also bug with depending on PrimaryOutput from template engine(where we assumed only .csproj will be result), now we only use PrimaryOutputs which are used by PostActions, since templates now include OpenFile post actions I also implemented that
To be able to better serialise Web project run configuration we now save/load values depending if project is Web or not…
Fixed 2 bugs:
Bug 56777 - New ASP.NET Core Web API project doesn't launch browser to http://localhost:<port>/api/values
Bug 55114 - Don't serialize default values for run configurations

P.S: launchSettings.json format is not documented, hence we don’t write changes to it or use it for reading settings


DavidKarlas:
@mrward you recently worked in this area, can you review, sadly I didn't see your changes(updated engine+parameter logic), merged it...

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/AssemblyRunConfiguration.cs
```diff
@@ -50,7 +50,7 @@ internal protected override void Initialize (Project project)
 			StartAction = StartActions.Project;
 		}
 
-		[ItemProperty (DefaultValue = "")]
+		[ItemProperty (DefaultValue = StartActions.Project)]
```
slluis: Don't change this please.

DavidKarlas: That is there to fix https://bugzilla.xamarin.com/show_bug.cgi?id=55114 what is side effect isn't it default looking at https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/AssemblyRunConfiguration.cs#L50

main/src/addins/MonoDevelop.DotNetCore/Properties/MonoDevelop.DotNetCore.addin.xml
```diff
@@ -276,47 +276,55 @@
...
 			_overrideDescription="Creates a new ASP.NET Web API Core web project."
-			path="Templates/Microsoft.DotNet.Web.ProjectTemplates.1.x.1.0.0-beta2-20170505-222.nupkg"
+			path="Templates/Microsoft.DotNet.Web.ProjectTemplates.1.x.1.0.0-beta2-20170523-241.nupkg"
 			icon="md-netcore-empty-project"
 			imageId="md-netcore-empty-project"
 			wizard="MonoDevelop.DotNetCore.ProjectTemplateWizard"
 			condition="UseNetCore1x=true"
-			category="netcore/app/aspnet" />
+			category="netcore/app/aspnet">
+				<Parameter name="IncludeLaunchSettings" value="true" />
```
mrward: Parameters defined with the DefaultParameters are already passed to the templating engine if the template supports them. So do we need another way to do this with this new Parameter element here? We should instead be able to do:

    defaultParameters="IncludeLaunchSettings=true"

If there is something not working here, say IncludeLaunchSettings is not defined as a known parameter by the template, so is not being set then we could just set all the parameters defined by DefaultParameters and not check if the template defines them.

DavidKarlas:
Addressed comments

PR #2501, opened 2017-05-29
Bug 56268 - [VSfM][Build 7.1.0.582]-Unable to see the breakpoint arro…

DavidKarlas:
…w when breakoint is hit.

Basically same as https://github.com/mono/monodevelop/commit/db35c5cf1acbd602e77b432896b3a91e68f99465 but expanded also to IconMargin beside TextSegment(Red and Yellow background)

### DIFF
--- a/main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Document/TextDocument.cs
+++ b/main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Document/TextDocument.cs
@@ -1608,6 +1608,13 @@ namespace Mono.TextEditor
 			return GetTextSegmentMarkersAt (line).OfType<DocumentLineTextSegmentMarker> ().Select (m => m.Marker);
 		}
 
+		public IEnumerable<TextLineMarker> GetMarkersOrderedByInsertion (DocumentLine line)
+		{
+			if (line == null)
+				return Enumerable.Empty<TextLineMarker> ();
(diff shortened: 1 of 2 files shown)


main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Document/TextDocument.cs
```diff
@@ -1608,6 +1608,13 @@ public IEnumerable<TextLineMarker> GetMarkers (DocumentLine line)
 			return GetTextSegmentMarkersAt (line).OfType<DocumentLineTextSegmentMarker> ().Select (m => m.Marker);
 		}
 
+		public IEnumerable<TextLineMarker> GetMarkersOrderedByInsertion (DocumentLine line)
+		{
+			if (line == null)
+				return Enumerable.Empty<TextLineMarker> ();
+			return OrderTextSegmentMarkersByInsertion(GetTextSegmentMarkersAt (line)).OfType<DocumentLineTextSegmentMarker> ().Select (m => m.Marker);
```
Therzok: This is really slow - OrderBy allocates a new collection. Do we want this on every draw operation?

Therzok: Ugh, I had no idea how to fix the issue stated in the commit message, maybe we should have a way to OrderTextSegmentMarkersByInsertion faster

PR #2505, opened 2017-05-30
	[Perf] Optimize loading of SystemAssemblyService

Therzok:
This brings down RegisterSystemAssemblies from 618ms to 44ms, most of it IO bound. Total File.Exists calls down from 8919 to 402.


Therzok:
The total number of iterations in the loop is also smaller, this is a big win

PR #2508, opened 2017-05-30
Master fix56531

mkrueger:


mkrueger:
Note: The PasteFormattingRule is for fixing an issue where paste inserts unwanted new lines. Like pasting in a lambda expression. Couldn't be fixed earlier due to missing internals access :(

PR #2509, opened 2017-05-30
[Addin] changed access modifiers and added some properties   

gitexperience:
These classes and properties are being referenced by the CBinding Addin .


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/ExtensibleTreeView.cs
```diff
@@ -598,7 +598,7 @@ internal ITreeBuilder CreateBuilder (Gtk.TreeIter it)
 			}
 		}
 
-		internal ITreeBuilderContext BuilderContext {
+		public ITreeBuilderContext BuilderContext {
```
slluis: This is not intended to be public.

main/src/core/MonoDevelop.Ide/gtk-gui/MonoDevelop.Ide.Projects.IncludeNewFilesDialog.cs
```diff
@@ -2,7 +2,7 @@
 // This file has been generated by the GUI designer. Do not modify.
 namespace MonoDevelop.Ide.Projects
 {
-	internal partial class IncludeNewFilesDialog
+	public partial class IncludeNewFilesDialog
```
slluis: It should not be public.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/SystemFileNodeBuilder.cs
```diff
@@ -43,7 +43,7 @@
 
 namespace MonoDevelop.Ide.Gui.Pads.ProjectPad
 {
-	class SystemFileNodeBuilder: TypeNodeBuilder
+	public class SystemFileNodeBuilder: TypeNodeBuilder
```
slluis: Why public?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/CompletionTextEditorExtension.cs
```diff
@@ -52,7 +52,7 @@ public class CompletionTextEditorExtension : TextEditorExtension
 		bool autoHideCompletionWindow, autoHideParameterWindow;
 
 		ICompletionWidget completionWidget;
-		internal virtual ICompletionWidget CompletionWidget
+		public virtual ICompletionWidget CompletionWidget
```
slluis: Please explain why this needs to be public.

mhutch:
@slluis the CBinding implements a custom project type for CMake projects that derives directly from SolutionItem: https://github.com/mhutch/cbinding/blob/master/CBinding/CMake/CMakeTarget.cs. Because of this, the only way it's able to access existing solution pad functionality (without completely reimplementing _everything_) is to subclass and customize the node builders.

My feeling is that maybe we should factor out a base FolderNodeBuilder and FileNodeBuilder that operate on some kind of IFile and IFolder interfaces or something?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/CompletionTextEditorExtension.cs
gitexperience: Because we need this property in CBinding :- https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/Projection/ProjectedFilterCompletionTextEditorExtension.cs#L48

Therzok: protected internal would be better then.

gitexperience: Yes it is .. PR Updated 
Thanks!

slluis:
> My feeling is that maybe we should factor out a base FolderNodeBuilder and
> FileNodeBuilder that operate on some kind of IFile and IFolder interfaces or something?

Yes, that's what should be done.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/CompletionTextEditorExtension.cs
slluis: I know it is for the CBinding. I want a good reasoning for each private API that is made public.

PR #2511, opened 2017-05-30
[Mac] Use static registrar!

Therzok:


main/build/MacOSX/monostub.m
```diff
@@ -1,404 +0,0 @@
-//gcc -m32 monostub.m -o monostub -framework AppKit
```
Therzok: this is a symlink so we don't break existing code.

slluis:
What's this change for?

Therzok:
View the PR by-commit, as the github UI can't understand what went on with the symlink and rename detection fails.

Therzok:
The first commit now introduces an explanation of what this change does

Therzok:
Rebased. The previous build only had debugger unit tests failures which are unrelated. Merging.

PR #2513, opened 2017-05-30
[Core] Fix bad merge conflict resolution in 11206a0039385143de65594f9…

Therzok:
…01dafbfc2426f30

It contained a mix of 6f06eaed471f4a41e40adb24328615de8afc7c97 and the previous version.

@slluis would adding a test that tests this on a project which references System.Collections.Immutable from .NET 4.5 after its built be ok?


slluis:
I think so.

Therzok:
Added tests!

PR #2516, opened 2017-05-31
Misc optimizations on typing.

Therzok:


main/src/core/MonoDevelop.Core/MonoDevelop.Core/Properties.cs
```diff
@@ -103,7 +103,8 @@ TypeConverter GetConverter (Type type)
 		
 		public T Get<T> (string property, T defaultValue)
 		{
-			defaultValues = defaultValues.SetItem (property, defaultValue);
+			if (!defaultValues.Contains (new KeyValuePair<string, object> (property, defaultValue)))
```
Therzok: I don't know why we do this. Why do we cache the default values (and re-use them in the Get<T> doesn't request a default value?). We should be sane and return a `default(T)` if the overload without a default value is used.

Therzok: cc @mkrueger @slluis

slluis:
You should separate unrelated optimizations in different PRs. We can't validate the PR if you keep adding changes.

Therzok:
Sorry, last commit wasn't meant to be done on this PR.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/Properties.cs
mkrueger: Has mostly historical reasons - I think the cache is no longer needed and can be removed.

Therzok: Will change if I see this being time consuming in the future.

PR #2517, opened 2017-05-31
Add new IdeApp.RunWhenIdle method

slluis:
IdeApp.RunWhenIdle() runs an action when the IDE is idle, that is, when the user is not performing any action and there is no modal dialog being shown.

This method should be used, for example, to show a notification to the user. The method will ensure that the dialog is shown when the user is not interacting with the IDE.


slluis:
There is an example of the NuGet add-in using it to show the license acceptance dialog.

Therzok:
Can't we use GLib.Idle.Add?

Therzok:
Ah, I see now, there's 500/1000ms timeouts.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
```diff
@@ -431,6 +431,96 @@ public static void Run ()
...
+		/// </summary>
+		/// <param name="action">Action to execute</param>
+		/// <remarks>
+		/// This method should be used, for example, to show a notification to the user. The method will ensure
+		/// that the dialog is shown when the user is not interacting with the IDE.
+		/// </remarks>
+		public static void RunWhenIdle (Action action)
+		{
+			Runtime.AssertMainThread ();
+			idleActions.Enqueue (action);
```
Therzok: Can't we handle dispatching to the UI thread here, instead of forcing it at callsites? We can handle concurrency with a simple lock.

slluis: Yeah, I can do that.

PR #2518, opened 2017-05-31
[Ide] Optimize text drawing.

Therzok:
This prevents allocating an extra char array on drawing line parts. Use short lived pinning of strings instead of this.


mkrueger:
I really like that change - the API of GetByteCount is missing a string variant with index/length access :(

Therzok:
Can't merge due to Lluis' requested changes.

PR #2521, opened 2017-05-31
Fix 43307

nosami:
Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=43307.

The issue was that `/4.5-api/` assemblies were being used for reference.Package when an assembly references a GAC package. These don't work at runtime.


slluis:
Do we need Newtonsoft.Json.xml in the repo?

nosami:
@lluis: git error... was supposed to be Newtonsoft.Json.dll . I don't really like adding the binary to the repo, but I can't see another way to test

slluis:
Can't it be added as NuGet reference?

nosami:
Yeah... I thought about doing that... might get transient network error failures doing that though. If that's ok, then I'll go down that route instead

slluis:
It's ok, we already depend on nuget restoration for building XS.

nosami:
Ok, will change it to restore the package first

Therzok:
So, I think this should be fixed in the core of SystemAssemblyService, so each target framework uses the right directory, right now, I think it's stopping at the first one hit by the paths provided by the runtime.

PR #2522, opened 2017-05-31
[51520][Ide][Mac] Add Selector handling to default Edit commands

sevoku:
Many native Cocoa widgets can handle default
editing commands through selectors. If the active
NSView is not an instance of NSText, check whether
it responds to the matching selector (cut/copy/paste)
and enable the command.

(fix bug #51520)

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/EditCommands.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/EditCommands.cs
@@ -155,6 +155,10 @@ namespace MonoDevelop.Ide.Commands
 	
 	internal class DefaultCopyHandler: CommandHandler
 	{
+		#if MAC
+		static readonly ObjCRuntime.Selector copySelector = new ObjCRuntime.Selector ("copy:");
+		#endif
+
 		protected override void Run ()
 		{
 			#if WIN32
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/EditCommands.cs
```diff
@@ -381,17 +417,19 @@ protected override void Update (CommandInfo info)
...
+			// HACK: GdkQuartzView always responds to "selectAll:" without handling it
+			info.Enabled |= macfocus is AppKit.NSText || (macfocus.Class.Name != "GdkQuartzView" && macfocus?.RespondsToSelector (selectAllSelector) == true);
 #endif
 
 #if WIN32
 			var winfocus = System.Windows.Input.Keyboard.FocusedElement;
 			info.Enabled |= winfocus != null;
 #endif
+
+			info.Bypass = !info.Enabled;
```
Therzok: Uhhhh, is this intended?

Therzok: Can't wrap my head around the possible implications, do native file dialogs select all work?

sevoku: Yes, this was intended, see https://github.com/mono/monodevelop/commit/328ee3d349131d3de5dec5299944f130d2129a51 and bug 39596. This is required to allow Gtk widgets to handle the key presses internally.

Select all and the other commands work in native dialogs, but not from the menu (key press works, but selecting the command from the menu doesn't, and it didn't work before).

PR #2523, opened 2017-05-31
String tag model

Therzok:


Therzok:
@VincentDondain your change can be simplified now!

PR #2525, opened 2017-05-31
[Ide] Allow New Project dialog to resize to fit text

mrward:
Fixed bug #54704 - Truncation on dialog "Configure your new Single View App"
https://bugzilla.xamarin.com/show_bug.cgi?id=54704

With German selected as the IDE language the New Project dialog
would cut off some of the controls on the left hand side of the
final page of the dialog. This was because the dialog had a fixed
width.


slluis:
Some UI elements are now misaligned: https://screencast.com/t/2RtHeN0pH6dp
Blank Forms app still has issues: https://screencast.com/t/Oc2nUgqZx0
Final page of blank Forms app is too big: https://screencast.com/t/s2zqC6OjCm7
Resizing doesn't work for icon label: https://screencast.com/t/cnypag1hq
Several issues on the Mac wizard: https://screencast.com/t/Qx1dPJQbdwdN
And TvOS: https://screencast.com/t/LZrLxTGRE

sevoku:
I don't think that we can simply set `Resizable = true`, cause this would allow the user to resize the dialog which will definitely break various custom widgets (like the app name allocation on the right side).
A better approach could be to enable ellipsis or wrapping for the labels in the middle and use something like `WidthChars` instead of `WidthRequest` for the entries. Or maybe all wizard pages using `Gtk.Table` for the layout should set a `WidthRequest` for the whole content and request everything to fill the table, rather than to set a fixed width for each entry.

slluis:
Making the dialog resizable should not be a problem if the wizard pages are correctly implemented. Setting a fixed size and then ellipsize labels when there is plenty of available space in a screen is not a good solution.

sevoku:
AFAIK a Gtk window should grow if necessary, even if `Resizable == false` as long as it has a default size and no fixed width/height. I don't think that it should be resizable by the user.

slluis:
Sure, I just mean that the wizard pages should correctly adapt to window size changes, which can be caused by labels being longer than expected in some languages, by other pages being bigger, or (if we allow it) a manual resize of the window. I also don't think it is necessary to make the window resizable by the user.

mrward:
- The dialog can no longer be resized. It will still expand to fit its contents if it needs to.
 - Fixed the final page taking up too much width (The custom UI controls, such as Test Cloud, now extend the full width on the left hand side. They were not using the last column available).
 - Gone through all the wizards so that they will now resize instead of being truncated if text is longer than expected. This needed changes in md-addins. Wizards in monodevelop have also been updated to resize.

Also created a **new-project-dialog-resize-to-fit-test** build lane which includes the md-addins changes.

PR #2527, opened 2017-06-01
Fix condition for enabling NuGet build options panel

slluis:
The IsDotNetProjectOnly condition is not correct because it is using
the project selected in the solution pad to check the type, and
that may not be the project for which the options dialog is being
opened. The right solution is to use the ItemType condition to
check the type.

This commit also includes a bump of Mono.Addins that fixes a bug
in Not conditions.

Fixes bug #55605 - [VSFM Preview 9] Solution option from IDE
toolbar doesn't open when cursor is set on project in solution explorer

(NuGet panel was incorrectly enabled in this context, and it was
crashing)


slluis:
Don't merge, I want to make sure the Mono.Addins changes don't break the build.

PR #2532, opened 2017-06-01
[MicrosoftTemplateEngine] Get default parameters from ITemplateInfo

VincentDondain:
With the new template engine, parameters can have default values:

```
"symbols": {
    "MinimumOSVersion": {
      "type": "parameter",
      "description": "Overrides the Info.plist's MinimumOSVersion",
      "replaces": "$MinimumOSVersion",
      "datatype": "string",
      "defaultValue": "8.0"
    }
}
```

Therefore we need to combine those "DefaultParameters" values with the ones set by the ExtensionNode.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngineSolutionTemplate.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngineSolutionTemplate.cs
@@ -68,10 +68,12 @@ namespace MonoDevelop.Ide.Templates
 			var parameters = new List<string> ();
 			var cacheParameters = templateInfo.CacheParameters.Where (m => !string.IsNullOrEmpty (m.Value.DefaultValue));
 
+			if (!cacheParameters.Any ())
+				return defaultParameters;
+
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngineSolutionTemplate.cs
```diff
@@ -53,12 +54,32 @@ internal MicrosoftTemplateEngineSolutionTemplate (TemplateExtensionNode template
...
 
 			//HasProjects = (template.SolutionDescriptor.EntryDescriptors.Length > 0);
 		}
+
+		string MergeDefaultParameters (string defaultParameters)
+		{
+			List<TemplateParameter> priorityParameters = null;
+			var parameters = new List<string> ();
+			var cacheParameters = templateInfo.CacheParameters.Where (m => !string.IsNullOrEmpty (m.Value.DefaultValue));
+
```
mrward: Just wondering if there is no cacheParameters then we could just return the defaultParameters here and not do any more processing.

VincentDondain: Good point!

PR #2534, opened 2017-06-02
Copy MSBuild assemblies to output.

KirillOsenkov:
On Windows when monodevelop is evaluating projects, it can't find the MSBuild assemblies anywhere. Make sure the assemblies get copied to \build\bin.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
@@ -124,15 +124,15 @@
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="Microsoft.Build">
       <HintPath>$(MSBuild_OSS_BinDir)Microsoft.Build.dll</HintPath>
-      <Private>False</Private>
+      <Private>True</Private>
     </Reference>
     <Reference Include="Microsoft.Build.Framework">
       <HintPath>$(MSBuild_OSS_BinDir)Microsoft.Build.Framework.dll</HintPath>
-      <Private>False</Private>
+      <Private>True</Private>
     </Reference>
     <Reference Include="Microsoft.Build.Utilities.Core">
       <HintPath>$(MSBuild_OSS_BinDir)Microsoft.Build.Utilities.Core.dll</HintPath>
-      <Private>False</Private>
+      <Private>True</Private>
     </Reference>
     <Reference Include="ICSharpCode.SharpZipLib">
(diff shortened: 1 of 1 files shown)


mhutch:
AFAICT this is supposed to be handled by https://github.com/mono/monodevelop/blob/666c3a3914a668283f085c2d418b6a3d036eab29/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs#L482

slluis:
Yes, this should not be necessary.

KirillOsenkov:
OK I've done some research and here's what I found regarding loading assemblies from a codebase on Windows. It turns out that even though we load assemblies using:
```
    var asmName = AssemblyName.GetAssemblyName (asmPath);
    return Assembly.Load (asmName);
```
it still doesn't get promoted from the LoadFrom context to the Load context for some reason (I don't understand exactly why and what is happening).

But loading these assemblies doesn't work automatically. Another workaround I found is this:
```
		static void LoadMSBuildLibraries ()
		{
			// Explicitly load the msbuild libraries since they are not installed in the GAC
			var path = systemAssemblyService.CurrentRuntime.GetMSBuildBinPath ("15.0");
			SystemAssemblyService.LoadAssemblyFrom (System.IO.Path.Combine (path, "Microsoft.Build.dll"));
			SystemAssemblyService.LoadAssemblyFrom (System.IO.Path.Combine (path, "Microsoft.Build.Framework.dll"));
			SystemAssemblyService.LoadAssemblyFrom (System.IO.Path.Combine (path, "Microsoft.Build.Utilities.Core.dll"));

			AppDomain.CurrentDomain.AssemblyResolve += CurrentDomain_AssemblyResolve;
		}

		private static System.Reflection.Assembly CurrentDomain_AssemblyResolve (object sender, ResolveEventArgs args)
		{
			if (args.Name.StartsWith ("Microsoft.Build", StringComparison.OrdinalIgnoreCase)) {
				foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies ()) {
					if (string.Equals (assembly.FullName, args.Name, StringComparison.OrdinalIgnoreCase)) {
						return assembly;
					}
				}
			}

			return null;
		}
```
But I'm not sure what's easier, just copy the assemblies locally (but then they'll be double-loaded) or subscribe to the AssemblyResolve event (and make it conditional CLR-only).

KirillOsenkov:
I've opened https://github.com/mono/monodevelop/pull/2546 as an alternative.

Sorry that the assembly loading on Windows is so... interesting.

KirillOsenkov:
Merged the other PR that uses AssemblyResolve.

PR #2536, opened 2017-06-02
Bug 56509 - No Razor template for Views folder

DavidKarlas:


main/src/addins/AspNet/Templates/AspNetCore/MVCViewImportsPage.xft.xml
```diff
@@ -0,0 +1,24 @@
+﻿<?xml version="1.0"?>
+<Template
+	originator="Matt Ward"
```
mrward: Pretty sure I did not create this :)

main/src/addins/AspNet/Projects/AspNetCoreFileTemplateCondition.cs
```diff
@@ -29,18 +29,32 @@
...
 namespace MonoDevelop.AspNet.Projects
 {
 	class AspNetCoreFileTemplateCondition : FileTemplateCondition
-	{
+	{
+		string minimalFrameworkVersion;
 		public override void Load (XmlElement element)
-		{
+		{
+			minimalFrameworkVersion = element.GetAttribute ("minimalFrameworkVersion");
```
mrward: Should we parse the version here and store it in a Version here instead of storing just the string?

main/src/addins/AspNet/Projects/AspNetCoreFileTemplateCondition.cs
```diff
@@ -29,18 +29,32 @@
...
 
 		public override bool ShouldEnableFor (Project proj, string projectPath)
 		{
-			return proj.GetProjectCapabilities ().Contains ("AspNetCore");
+			if (proj.GetProjectCapabilities ().Contains ("AspNetCore")) {
+				if (string.IsNullOrEmpty (minimalFrameworkVersion))// If minimal version it's not set consider it's supporting all frameworks
+					return true;
+				var dpn = (DotNetProject)proj;
+				if (Version.TryParse (dpn.TargetFramework.Id.Version, out var version))
+					return version.CompareTo (Version.Parse (minimalFrameworkVersion)) != -1;//-1 means it's lower, 0 same, 1 greater
```
mrward: I would be tempted to change this to just do `return version >= minimumVersion`

main/src/addins/AspNet/Projects/AspNetCoreFileTemplateCondition.cs
mrward: Also wonder if this should be called minimumFrameworkVersion or minFrameworkVersion instead of minimalFrameworkVersion. It could even be minVersion maybe? Since then the file template would have something like: `<AspNetCore minVersion="2.0" />` which seems to read quite well.

main/src/addins/AspNet/Templates/AspNetCore/ViewPageRazor.xft.xml
```diff
@@ -0,0 +1,25 @@
+﻿<?xml version="1.0"?>
+<Template originator="Piotr Dowgiallo"
+          created="2012/06/15"
+          lastModified="2012/06/15">
+
+	<!-- Template Header -->
+	<TemplateConfiguration>
+		<_Name>ASP.NET MVC View Page (Razor)</_Name>
+		<Icon>md-html-file-icon</Icon>
+		<_Category>ASP.NET MVC</_Category>
```
mrward: Is this category correct? Just wondering since the template file is in a AspNetCore folder. If this is for a normal ASP.NET project then it probably should be in a different folder in the project.

DavidKarlas:
Addressed all comments

PR #2537, opened 2017-06-02
Master fix56626

mkrueger:


main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring/QuickFixMenuHandler.cs
```diff
@@ -46,7 +46,7 @@ class QuickFixMenuHandler : CommandHandler
 			quickFixMenu.CommandInfos.Add (new CommandInfo (GettextCatalog.GetString ("Loading..."), false, false), null);
 			info.Add (quickFixMenu);
 			try {
-				var menu = await CodeFixMenuService.CreateFixMenu (editor, ext.GetCurrentFixes (), cancelToken);
+				var menu = await CodeFixMenuService.CreateFixMenu (editor, ext.GetCurrentFixesAsync (cancelToken).Result, cancelToken);
```
slluis: You should await the GetCurrentFixesAsync() call.

mkrueger: For some reason that doesn't work - didn't find a cause the stack trace didn't say much to me.

slluis: If it doesn't work, let's fix it.

slluis:
That NotifyChanged() call should not be necessary. I'll see if I can fix it in the command system.

slluis:
I'll fix the notification issue on master.

slluis:
Can you please create a PR for d15-3?

PR #2546, opened 2017-06-02
Avoid assembly load failure on Windows.

KirillOsenkov:
For some reason even though we call Assembly.Load from the codebase, it still doesn't seem to get promoted from LoadFrom to the Load context.

A solution is to use AssemblyResolve event and just resolve to the assembly that's already loaded.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
@@ -496,9 +496,9 @@ namespace MonoDevelop.Core
 		/// This is necessary on Windows because without it, even though the assemblies are already loaded
 		/// the CLR will still throw FileNotFoundException (unable to load assembly).
 		/// </summary>
-		private static System.Reflection.Assembly CurrentDomain_AssemblyResolve (object sender, ResolveEventArgs args)
+		static System.Reflection.Assembly CurrentDomain_AssemblyResolve (object sender, ResolveEventArgs args)
 		{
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
```diff
@@ -486,6 +486,27 @@ static void LoadMSBuildLibraries ()
...
+			if (Type.GetType ("Mono.Runtime") == null) {
+				AppDomain.CurrentDomain.AssemblyResolve += CurrentDomain_AssemblyResolve;
+			}
+		}
+
+		/// <summary>
+		/// This is necessary on Windows because without it, even though the assemblies are already loaded
+		/// the CLR will still throw FileNotFoundException (unable to load assembly).
+		/// </summary>
+		private static System.Reflection.Assembly CurrentDomain_AssemblyResolve (object sender, ResolveEventArgs args)
```
Therzok: Nit: redundant accessibility modifier.

KirillOsenkov: Thanks, fixed.

main/Main.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26228.4
+VisualStudioVersion = 15.0.26403.0
```
Therzok: Woops :p

KirillOsenkov: Yeah reverted.

PR #2552, opened 2017-06-05
Bump mono to 5.4.0.0

luhenry:
cc @slluis


slluis:
This is not correct. monodevelop/master must track md-addins/master.

PR #2556, opened 2017-06-06
Make main UI keyboard accessible

iainx:
Make the main UI navigable by the tab key.
Obviously you can't tab out of the main text editor though, but it seems like that's not a thing that can be done on native macOS anyway.


Therzok:
_grabs like 5 coffees_

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PathBar.cs
```diff
@@ -789,5 +814,61 @@ protected override void OnDestroyed ()
...
+
+					if (focusedPathIndex < 0) {
+						ret = false;
+					}
+				}
+				break;
+			}
+
+			if (ret) {
+				alreadyHaveFocus = true;
```
Therzok: `alreadyHaveFocus = ret`

PR #2557, opened 2017-06-06
Xm registrar fixes

Therzok:


main/build/MacOSX/Makefile.am
```diff
@@ -10,6 +10,8 @@ MD_CONFIGURE=$(top_srcdir)/../scripts/configure.sh
 EXTERNAL=../../external
 MONOSTUB_DEFINES=-DXM_REGISTRAR
 MONOSTUB_STATIC_LINK=$(EXTERNAL)/Xamarin.Mac.registrar.full.a
+HAS_EXTERN_C_XM_REGISTRAR_DEFINE=$(shell ./check-xm-extern.sh)
```
Therzok: Self: remove HAS

main/build/MacOSX/check-xm-extern.sh
```diff
@@ -0,0 +1,2 @@
+#/usr/bin/env bash
+(nm ../../external/Xamarin.Mac.registrar.full.a | grep " _xamarin_create_classes_Xamarin_Mac" >/dev/null) && echo "-DEXTERN_C"
```
Therzok: Self: add comment explaining

slluis:
Can you explain what are the changes for?

Therzok:
@slluis external/Makefile changes are so we now copy each file when it changes, not depend on Xamarin.Mac.dll being changed.

The other change is a check for specific versions of Xamarin.Mac. Newer Xamarin.Mac versions (master and I think d15-3) export the xamarin_create_classes function as "C", aka the compiler doesn't mangle the function name. This is needed so we can link the right function.

Therzok:
Updated commits with better explanations. Merging.

PR #2560, opened 2017-06-06
[F#] Add target frameworks and don't reference mscorlib.

nosami:
Fixes unit tests. msbuild doesn't reference mscorlib by default, so the IDE
shouldn't either


nosami:
build lane https://code.internalx.com/Wrench/BuildDetails?buildId=76b0fb7aca23803b3861ddfec965bec0

slluis:
Tests are still failing on that lane.

nosami:
I just removed those tests. They are invalid. They test that mscorlib is implicitly added behind the scenes. This PR changes that behaviour.

nosami:
All tests are green now. The original failing test is now passing, and the invalid tests have been removed.

PR #2562, opened 2017-06-06
Perf ui

Therzok:
If we want this in, we need to:
~1. Merge this: https://github.com/mono/gtk-sharp/pull/200 - the reflection code would be too complicated to take into account for the Objects dictionary being in the Object class, plus, ToggleRef is not publicly accessible, therefore we need to use fully untyped reflection until we can get the array of pointers.~
~2. Bump mono to a build with this gtk# version~

Not needed to do anything special anymore.


Therzok:
https://www.screencast.com/t/7wpPlkpBN - Check the bottom right
https://www.screencast.com/t/5qW4FJDng3

main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/LeakHelpers.cs
```diff
@@ -0,0 +1,178 @@
...
+		static int remainingEqualChangedCount = 5;
+		static bool SummaryTimeoutHandler ()
+		{
+			int gobjCount = GObjectDict.Count;
+			int nsobjCount = NSObjectDict.Count;
+
+			GC.Collect ();
+			GC.Collect ();
+			GC.Collect ();
+			GC.WaitForPendingFinalizers ();
```
alanmcgovern: The finalizers for Gtk can only do their job properly if the main loop is pumping. We might have better accuracy if you do 3 GC calls and wait for finaliser calls, separated by 1 second, and then perform the analysis of live objects

main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/Options.cs
```diff
@@ -31,6 +31,10 @@ namespace PerformanceDiagnosticsAddIn
 	public static class Options
 	{
 		public static readonly ConfigurationProperty<string> OutputPath = ConfigurationProperty.Create ("PerformanceDiagnosticsAddIn.OutputPath", System.IO.Path.Combine (Environment.GetFolderPath (Environment.SpecialFolder.Personal), "Desktop"));
+		//public static readonly ConfigurationProperty<bool> ShowUICounters = ConfigurationProperty.Create ("PerformanceDiagnosticsAddIn.ShowUICounters", false);
+
+		public static readonly bool HasMemoryLeakFeature = System.IO.File.Exists (System.IO.Path.Combine (Environment.GetFolderPath (Environment.SpecialFolder.Personal), ".xs-enable-leak-check"));
+		public static readonly bool HasMemoryLeakFeaturePad = HasMemoryLeakFeature && System.IO.File.Exists (System.IO.Path.Combine (Environment.GetFolderPath (Environment.SpecialFolder.Personal), ".xs-enable-leak-check-pad"));
```
alanmcgovern: By convention we don't prefix with . as it's hard to create on windows

main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/LeakHelpers.cs
Therzok: This method is queued recursively, guaranteeing that for example, unrefs done on chained objects (Gtk.TextTagTable + Gtk.TextTag) both get finalized.

Therzok: We queue the timeout again with `return true` to run a second time.

main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/Options.cs
Therzok: Okay, sure

main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/LeakHelpers.cs
alanmcgovern: Oh i see now :P It helps when i have a large screen to read the entire diff!

main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/Options.cs
```diff
@@ -31,6 +31,10 @@ namespace PerformanceDiagnosticsAddIn
 	public static class Options
 	{
 		public static readonly ConfigurationProperty<string> OutputPath = ConfigurationProperty.Create ("PerformanceDiagnosticsAddIn.OutputPath", System.IO.Path.Combine (Environment.GetFolderPath (Environment.SpecialFolder.Personal), "Desktop"));
+		//public static readonly ConfigurationProperty<bool> ShowUICounters = ConfigurationProperty.Create ("PerformanceDiagnosticsAddIn.ShowUICounters", false);
```
alanmcgovern: Why is this new code commented out?

main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/GlobalOptionsDialog.cs
```diff
@@ -29,13 +30,24 @@ public override Control CreatePanelWidget()
...
 			vbox.ShowAll();
 			return vbox;
 		}
 
 		public override void ApplyChanges()
 		{
 			Options.OutputPath.Value = folderEntry.Path;
+
+			//if (Options.HasMemoryLeakFeature)
+				//Options.ShowUICounters.Value = uiCountersCheck.Active;
```
alanmcgovern: new code commented out?

Therzok: Was the intention of having something for users, not just us.

alanmcgovern:
👍 then!

main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/PerformanceDiagnosticsAddIn.csproj
```diff
@@ -10,7 +10,7 @@
     <OutputType>Library</OutputType>
     <RootNamespace>PerformanceDiagnosticsAddIn</RootNamespace>
     <AssemblyName>PerformanceDiagnosticsAddIn</AssemblyName>
-    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.7</TargetFrameworkVersion>
```
Therzok: Note to self: Revert this to 4.6.1 and fixup ValueTuple package.

PR #2564, opened 2017-06-06
[A11y] Set the help description on the font buttons

iainx:
Fixes BXC #53523

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Fonts/FontChooserPanelWidget.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Fonts/FontChooserPanelWidget.cs
@@ -26,6 +26,7 @@
 using System;
 using System.Collections.Generic;
 using MonoDevelop.Core;
+using MonoDevelop.Components.AtkCocoaHelper;
 using Gtk;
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Fonts/FontChooserPanelWidget.cs
```diff
@@ -80,6 +81,9 @@ public FontChooserPanelWidget ()
 				var hBox = new HBox ();
 				var setFontButton = new Button ();
 				setFontButton.Label = FontService.FilterFontName (GetFont (desc.Name));
+
+				var descStr = string.Format (GettextCatalog.GetString ("Set the font options for {0}"), GettextCatalog.GetString (desc.DisplayName));
```
slluis: This can be simplified to:

```csharp
var descStr = GettextCatalog.GetString ("Set the font options for {0}", GettextCatalog.GetString (desc.DisplayName));
```

iainx: I knew that, but for some reason forgot :)

Fixed now

PR #2569, opened 2017-06-06
[Ide] Fix null reference exception in New Project dialog

mrward:
Selecting a different category in the New Project dialog would
throw a null reference exception in the language cell renderer due
to the template not being available initially.

Also wondering if we should make LanguageCellRenderer class non-public on master and on the d15-3 branch. It is only used by the New Project dialog.


Therzok:
I'm ok with making it internal

slluis:
Yes, it should be internal.

PR #2574, opened 2017-06-07
Fixed 'Bug 16804 - External console run on Gnome 3.10 terminal

mkrueger:
problem'

Ported patch from Matthias Gliwka which fixed the gnome-terminal
problem. I added some minor fixes & verified that it works on a recent
ubuntu.


gliwka:
@mkrueger Thanks for mainlining this patch :-)

PR #2576, opened 2017-06-07
[Ide] Don't bind Page+Up Page+Down keys by default

sevoku:
This makes the Page+Up/Page+Down keys less discoverable, but its the only way to resolve the conflict between the command manager and default Gtk key handler inside the text editor without bigger command system design changes.

(fixes bug #55143)


slluis:
What do you mean by "This makes the Page+Up/Page+Down keys less discoverable"?

sevoku:
Less discoverable means that they bindings are not shown in the key bindings preferences panel. The initial idea behind adding them to the mappings was to make them visible to the user (espacially for default bindings that differ between Windows and Mac) and as a side effect it would make them work outside of the editor (for any text in the IDE, Gtk and native).

PR #2582, opened 2017-06-07
[A11y] Add a Restart panel and make it accessible

iainx:
Split the Restart panel out into a separate widget, make it accessible.

Fixes BXC #53523

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/RestartPanel.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/RestartPanel.cs
@@ -1,0 +1,66 @@
+﻿//
+// RestartPanel.cs
+//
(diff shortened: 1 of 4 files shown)


iainx:
Ideally the IDE theme panel would use the restart panel, but BOOOOO STETIC!

iainx:
This actually fixes 53447, not sure where I got 53523 from...

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/RestartPanel.cs
```diff
@@ -0,0 +1,66 @@
+﻿//
+// RestartPanel.cs
+//
+// Author:
+//       iain <>
```
Therzok: bad copyright

PR #2583, opened 2017-06-07
[A11y] Fix crash when the tabstrip is reset on debugging

iainx:
When the tabstrip was being focused it could try to access a tab with a negative index and crash


alanmcgovern:
Does this need to be backported to 15.3, or did the original change land after we branched?

iainx:
@alanmcgovern the original change landed after we branched

PR #2584, opened 2017-06-08
Don't crash the main process when a remote builder crashes

KirillOsenkov:
I'm investigating https://bugzilla.xamarin.com/show_bug.cgi?id=55660 and I'm noticing a few things that happen when the remote process crashes.

1. Calling Task.ContinueWith with a delegate that throws will crash the entire process.
2. It is happening that while we're iterating over messageWaiters in AbortPendingMessages() a new message is added (presumably reentrancy when the lock is pumping the timer adds a new one on the same thread, so the lock is ineffective). Simple fix is to capture the set of messages to drain and iterate over an immutable snapshot.
3. Several threads are calling SetException on the task source resulting in an InvalidOperationException. Calling TrySetException instead avoids that exception.

Generally, I feel like synchronization in this area needs to be thought about as we should abort more gracefully when the process terminates and avoid trying to contact it again, resulting in a cascade of secondary failures that make things even worse.

This fix is not yet ready because though the IDE doesn't crash anymore, it continues "building" indefinitely and never stops the build.

I need to find a better way to stop in case of exception to signal that the build should stop.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/RemoteProcessConnection.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/RemoteProcessConnection.cs
@@ -81,7 +81,7 @@ namespace MonoDevelop.Core.Execution
 		public bool IsReachable {
 			get {
 				var s = status;
-				return s != ConnectionStatus.ConnectionFailed && s != ConnectionStatus.Disconnected;
+				return process != null && s != ConnectionStatus.ConnectionFailed && s != ConnectionStatus.Disconnected;
 			}
 		}
 
@@ -298,6 +298,8 @@ namespace MonoDevelop.Core.Execution
 		{
 			// any exception bubbling up from here would crash the process
 			try {
+				// somehow capture that the process has exited
+				process = null;
 				if (!stopping)
 					AbortConnection (isAsync: true);
 			}
@@ -312,6 +314,11 @@ namespace MonoDevelop.Core.Execution
 
 		public Task<BinaryMessage> SendMessage (BinaryMessage message)
 		{
+			// do not attempt to send new messages if the process has already exited
+			if (process == null || process.IsCompleted) {
+				return Task.FromException<BinaryMessage> (new Exception ("Process not running"));
+			}
+
 			message.ReadCustomData ();
 			var interceptor = Interceptor;
 			if (interceptor != null && !interceptor.PreProcessMessage (message))
@@ -447,8 +454,13 @@ namespace MonoDevelop.Core.Execution
 		void StopPinger ()
 		{
 			if (pinger != null) {
-				pinger.Dispose ();
-				pinger = null;
+				// not sure but it seems that if multiple places call StopPinger we should lock here
+				lock (pingerLock) {
+					if (pinger != null) {
+						pinger.Dispose ();
+						pinger = null;
+					}
+				}
 			}
 		}
 
@@ -457,7 +469,11 @@ namespace MonoDevelop.Core.Execution
 			if (process != null)
 				stopping = true;
 
-			AbortConnection (isAsync: isAsync);
+			if (stopping) {
+				// only abort connection if the process is still there
+				AbortConnection (isAsync: isAsync);
+			}
+
 			StopPinger ();
 
 			if (listener != null) {
@@ -472,7 +488,10 @@ namespace MonoDevelop.Core.Execution
 				connection.Close ();
 				connection = null;
 			}
-			process.Cancel ();
+
+			if (stopping) {
+				process.Cancel ();
+			}
 		}
 
 		void OnConnected (IAsyncResult res)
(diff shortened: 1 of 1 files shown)


KirillOsenkov:
https://bugzilla.xamarin.com/show_bug.cgi?id=55660 has a simple repro: build the project containing this:

```

  <UsingTask TaskName="Crash" TaskFactory="CodeTaskFactory" AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.Core.dll">
    <Task>
      <Reference Include="mscorlib" />
      <Using Namespace="System" />
      <Code Type="Fragment" Language="cs">
        <![CDATA[
Environment.FailFast("Crash");
return true;
        ]]>
      </Code>
    </Task>
  </UsingTask>

  <Target Name="Crash" AfterTargets="AfterBuild">
    <Crash />
  </Target>
```

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/RemoteProcessConnection.cs
```diff
@@ -430,7 +435,9 @@ void AbortConnection (string message = null, bool isAsync = false)
 		void AbortPendingMessages ()
 		{
 			lock (messageWaiters) {
-				foreach (var m in messageWaiters.Values)
+				// capture the original values because they can change while we're enumerating
+				var originalValues = messageWaiters.Values.ToArray ();
+				foreach (var m in originalValues)
```
alanmcgovern: If the collection can change, is it possible a value will be *added* while we're enumerating or is it only possible for them to be *removed*? If they can be added then we'd silently discard a value here because we clear the list at the end without checking for newly added items.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/RemoteProcessConnection.cs
```diff
@@ -605,7 +612,9 @@ void NotifyResponse (MessageRequest req, BinaryMessage res)
 				if (res.GetArgument<bool> ("IsInternal") && !string.IsNullOrEmpty (msg)) {
 					msg = "The operation failed due to an internal error: " + msg + ".";
 				}
-				req.TaskSource.SetException (new RemoteProcessException (msg) { ExtendedDetails = res.GetArgument<string> ("Log") });
+
+				// if someone else had already set the exception, don't crash
+				req.TaskSource.TrySetException (new RemoteProcessException (msg) { ExtendedDetails = res.GetArgument<string> ("Log") });
 			} else {
 				req.TaskSource.SetResult (res);
```
alanmcgovern: Shouldn't this be a `TrySet` too ?

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/RemoteProcessConnection.cs
```diff
@@ -605,7 +612,9 @@ void NotifyResponse (MessageRequest req, BinaryMessage res)
 				if (res.GetArgument<bool> ("IsInternal") && !string.IsNullOrEmpty (msg)) {
 					msg = "The operation failed due to an internal error: " + msg + ".";
 				}
-				req.TaskSource.SetException (new RemoteProcessException (msg) { ExtendedDetails = res.GetArgument<string> ("Log") });
```
alanmcgovern: Should the `req.TaskSource.SetException (new Exception ("Connection closed"));` above this be a TrySet as well?

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/RemoteProcessConnection.cs
KirillOsenkov: What I saw is the value was in fact added. That's a good point. I'll think of better synchronization here (because we can't rely on lock here apparently).

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/RemoteProcessConnection.cs
KirillOsenkov: Yes this entire code needs to be audited for concurrency.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/RemoteProcessConnection.cs
KirillOsenkov: Yup. I suspect the code needs to be more aware of concurrency and avoid doing some of the work when the connection abruptly terminates.

KirillOsenkov:
I've added another commit which just marks places worth investigating, but the fixes aren't nearly good enough. Just something to start an investigation.

KirillOsenkov:
@slluis this may be helpful for the builder process crashing investigation...

KirillOsenkov:
This is now fixed by https://github.com/mono/monodevelop/pull/4692

PR #2586, opened 2017-06-08
[DotNetCore] Fix ASP.NET Core project build error with old SDK

mrward:
Fixed bug #57195 - Xamarin.Forms with ASP.NET Mobile Backend does not
build
https://bugzilla.xamarin.com/show_bug.cgi?id=57195

Building an ASP.NET Core project when .NET Core SDK 1.0.3 or lower
was installed, or if no .NET Core SDK was installed, would result in
the build failing with the error:

error MSB4057: The target "Build" does not exist in the project.

With .NET Core SDK 1.0.4 or later installed this build error does not
occur. The problem is that with older .NET Core SDKs the ASP.NET Core
project uses the Microsoft.Web.Sdk.Web MSBuild imports. The older
Microsoft.Web.Sdk.Web  imports use the MSBuildSdksPath property when
importing other MSBuild targets files. This is not defined and results
in nothing being imported and so the build targets are not available.
To fix this the MSBuildSDKsPath property is defined globally if an
old .NET Core sdk is installed, or if no .NET Core SDK is installed.

This will need fixing in the d15-3 branch too.

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/MSBuildSdksPathGlobalPropertyProvider.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/MSBuildSdksPathGlobalPropertyProvider.cs
@@ -1,0 +1,82 @@
+﻿//
+// MSBuildSdksPathGlobalPropertyProvider.cs
+//
+// Author:
+//       Matt Ward <matt.ward@xamarin.com>
+//
+// Copyright (c) 2017 Xamarin Inc. (http://xamarin.com)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+using System;
+using System.Collections.Generic;
+using MonoDevelop.Projects.MSBuild;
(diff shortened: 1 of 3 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/MSBuildSdksPathGlobalPropertyProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+		Dictionary<string, string> properties;
+
+		#pragma warning disable 67
+		public event EventHandler GlobalPropertiesChanged;
+		#pragma warning restore 67
+
+		public IDictionary<string, string> GetGlobalProperties ()
+		{
+			if (properties == null) {
+				properties = new Dictionary<string, string> ();
```
slluis: I think this method should be thread safe

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/MSBuildSdksPathGlobalPropertyProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+
+		string GetMSBuildSDKsPath ()
+		{
+			if (DotNetCoreSdk.IsInstalled) {
+				var latestVersion = DotNetCoreSdk.Versions.FirstOrDefault ();
+				if (latestVersion != null && latestVersion < minimumVersion) {
+					return DotNetCoreSdk.MSBuildSDKsPath;
+				}
+			} else if (MSBuildSdks.Installed) {
+				return MSBuildSdks.MSBuildSDKsPath;
```
slluis: Why this? shouldn't this be fixed in Mono?

mrward: I tried not setting this but it did not work. Without this the ASP.NET Core project fails to build with same build error `The target "Build" does not exist in the project.`.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/MSBuildSdksPathGlobalPropertyProvider.cs
mrward: I can add a lock around this code.

mrward:
Updated the property creation logic.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/MSBuildSdksPathGlobalPropertyProvider.cs
slluis: Couldn't this be due to Mono shipping outdated sdk targets?

mrward: Correct, yes. If Mono updates the .NET Core sdk they are shipping to 1.0.4 then MSBuildSdks.MSBuildSDKsPath would not be need to be returned here.

Mono 5.2.0.171 (2017-04) currently has Microsoft.NET.Sdk.Web MSBuild targets that still use `MSBuildSdksPath` when importing files.

slluis: OK, can you talk Marek to that they are updated?
We can commit this meanwhile, and remove that part of the condition when Mono is updated.

PR #2600, opened 2017-06-09
[Gettext] Do not include CDATA tags in translations

mrward:
CDATA tags that were inside translatable xml elements were included
in the extracted translation text when the files are scanned. The
CDATA tags themselves should not be displayed to the user so they are
now excluded with the regular expression. CDATA tags are used in
the project templates and were not translated since the text being
passed to the translation catalog would not include these tags and
would not find a match.


mrward:
@Therzok - I guess we will probably want to run some script to remove the existing CDATA tags from the .po files we already have.

Therzok:
We had some code which parsed PO strings containing CData and escaped it. This will help remove the hack from the code, but will create new strings to translate. I'm ok with this change, though, I'm not sure whether we actually need it.

PR #2601, opened 2017-06-09
[Ide] Fix file templates not translated

mrward:
Fixed bug #53635 - Many "New file" dialogs/strings are unlocalized.
https://bugzilla.xamarin.com/show_bug.cgi?id=53635

File templates that require a custom localizer were not translated
since Gettextcatalog.GetString was always used.

Requires [md-addins change](https://github.com/xamarin/md-addins/pull/2077) before this will translate the file templates.


Therzok:
This looks good to me. Don't see another way to do it so it's picked up automatically. Maybe we should inject the IAddinLocalizer attribute at build time into all the md-addins assemblies?

PR #2603, opened 2017-06-12
[Ide] Fix project template categories not added to translations

mrward:
The project template categories were defined in the .addin.xml file
using an attribute without an underscore so they were not added
to the translations. The project template categories now use _name
for the category name so they are added to the translations. To
maintain backwards compatibility the name attribute without the
underscore is still supported.

    <Category id="library" _name="Library">

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Codons/TemplateCategoryCodon.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Codons/TemplateCategoryCodon.cs
@@ -27,6 +27,7 @@
 using Mono.Addins;
 using MonoDevelop.Ide.Templates;
 using MonoDevelop.Core;
+using System;
 using System.Linq;
 
 namespace MonoDevelop.Ide.Codons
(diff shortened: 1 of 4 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Codons/TemplateCategoryCodon.cs
```diff
@@ -27,16 +27,21 @@
...
+using System;
 using System.Linq;
 
 namespace MonoDevelop.Ide.Codons
 {
 	[ExtensionNode (Description="A template category.")]
 	internal class TemplateCategoryCodon : ExtensionNode
 	{
+		[ObsoleteAttribute ("This is ignored when generating translations.")]
 		[NodeAttribute ("name", "Name of the category.", Localizable=true)]
```
Therzok: The fact that Localizable is true means that prefixing with underscore will mark it for translation.

Therzok: See: https://github.com/mono/monodevelop/blob/master/main/src/addins/MonoDevelop.Gettext/MonoDevelop.Gettext.addin.xml#L114-L115

Therzok: Basically, you don't need a new node attribute. You can just set `Localizable=true` and the addin manager will check for both `name` and `_name`.

You need to prefix with `_` in the .addin.xml for the gettext scanner to pick up the string.

Therzok: And I'm sooooo wrong. Sorry. AddinManager doesn't do that automatically...

PR #2605, opened 2017-06-12
Debug leak

Therzok:


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/TabStrip.cs
```diff
@@ -100,8 +100,8 @@ public void AddTab (DockItemTitleTab tab)
 
 		void HandleRemoved (object o, RemovedArgs args)
 		{
-			Gtk.Widget w = args.Widget;
-			w.ButtonPressEvent -= OnTabPress;
+			var w = (DockItemTitleTab)args.Widget;
```
Therzok: From what I've seen, the widget can't be anything other than a dockitemtitletab, or can it?

iainx: Don't think so?

PR #2607, opened 2017-06-12
Remote process crash instrumented

KirillOsenkov:
@slluis I've found that instrumenting the RemoteProcessConnection.cs like this really helps to debug and investigate what's going on.

Of course this is just for debugging, do not merge.


KirillOsenkov:
Also this one @slluis

KirillOsenkov:
This is now fixed by https://github.com/mono/monodevelop/pull/4692

PR #2609, opened 2017-06-12
Changes needed to build MD on Windows

KirillOsenkov:
This demonstrates a workaround of how to build Monodevelop on Windows.

The problem right now is we took a dependency on SdkResolver APIs that will only appear in MSBuild 15.3. Since the chances are the majority don't have the preview of VS 2017 Update 3 with these changes installed, the MD build will fail.

These three changes make it work again. Important: for this to work you need to:

 1. C:\\> `git clone https://github.com/Microsoft/MSBuild`
 2. C:\\> `cd MSBuild`
 3. C:\MSBuild> `cibuild --Scope Compile`

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
@@ -90,7 +90,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <PropertyGroup>
-    <MSBuild_OSS_BinDir Condition="'$(OS)' == 'Windows_NT'">$(MSBuildToolsPath)\</MSBuild_OSS_BinDir>
+    <MSBuild_OSS_BinDir Condition="'$(OS)' == 'Windows_NT'">C:\MSBuild\bin\Bootstrap\MSBuild\15.0\Bin\</MSBuild_OSS_BinDir>
     <!-- when building with xbuild on Unix, $(MSBuildToolsPath) is like `$prefix/lib/mono/xbuild/*/bin`
 	 when building with msbuild on Unix, $(MSBuildToolsPath) is like `$prefix/lib/mono/msbuild/*/bin`
 	 Prefer referencing msbuild 15.* assemblies over 14.1 . At runtime, we use correct one anyway
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
@@ -1,4 +1,4 @@
-//
+﻿//
 // Runtime.cs
 //
 // Author:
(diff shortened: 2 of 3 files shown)


KirillOsenkov:
CC @Therzok @bratsche @mhutch just FYI

KirillOsenkov:
And here's a little ApplyPatch script that automates applying these changes to your repo:
https://github.com/KirillOsenkov/ApplyPatch

Warning: some tweaking may be required if the paths on your machine are different.

Therzok:
@kirillosenkov or even easier with git. Make the changes, `git stash` to register them. And when you need them (i.e. after stash or after git reset --hard), just git stash apply ;)

DavidKarlas:
Maybe you just need to sprinkle some Environment Variables magic on master branch :)

EgorBo:
I managed to run monodevelop on windows with the following steps:
1) Install VS2017U3
2) Update System.ValueType to 4.4.0 in MonoDevelop.Core project
3) Unload MonoDevelop.Projects.Formats.MSBuild.v12.0 and 3) Unload MonoDevelop.Projects.Formats.MSBuild.v14.0
4) #ifdef some macOS specific code in main\src\addins\PerformanceDiagnostics\PerformanceDiagnostics\LeakHelpers.cs:L42

KirillOsenkov:
Yes, now that VS2017U3 is released this PR is no longer necessary. Just install VS 15.3. 

@EgorBo make sure you're using the DebugWin32 Configuration in the Configuration manager. The Debug configuration won't work on WIndows.

@Therzok can you check Egor's point 4 above? main\src\addins\PerformanceDiagnostics\PerformanceDiagnostics\LeakHelpers.cs:L42 If he chooses DebugWin32 the perf diagnostics shouldn't even build, right?

EgorBo:
@KirillOsenkov yeah I was using DebugWin32 but had to do "unload" anyway.
regarding 2 - if I don't do it - it compiles but fail during runtime with "Could not load file or assembly 'System.ValueTuple, Version=4.0.1.0"
so after update this:
```
<Reference Include="System.ValueTuple">
  <HintPath>..\..\..\packages\System.ValueTuple.4.3.1\lib\netstandard1.0\System.ValueTuple.dll</HintPath>
</Reference>
```
become this:
```
<Reference Include="System.ValueTuple, Version=4.0.2.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51, processorArchitecture=MSIL">
  <HintPath>..\..\..\packages\System.ValueTuple.4.4.0\lib\net461\System.ValueTuple.dll</HintPath>
</Reference>
```

KirillOsenkov:
Ah, regarding the ValueTuple you need a binding redirect in the app.config. I keep forgetting to push it to the master branch. https://github.com/mono/monodevelop/pull/2864/files

And I think VS pends a change to Main.sln which you need to undo, then it won't include those projects you had to unload. I usually first build from command line using winbuild.bat, then open in the IDE.

PR #2614, opened 2017-06-13
Fixed 'Bug 57392 - ctor code snippet is producing an nameless

mkrueger:
constructor'

We should consider replacing our code template system with the VS.NET
one or rework the system.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeTemplates/ExpansionObject.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeTemplates/ExpansionObject.cs
@@ -38,6 +38,7 @@ using System.Linq;
 using MonoDevelop.Ide.Editor;
 using MonoDevelop.Ide.Editor.Extension;
 using System.Web.SessionState;
+using System.Threading;
 
 namespace MonoDevelop.Ide.CodeTemplates
 {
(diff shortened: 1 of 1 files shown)


DavidKarlas:
Shouldn't `.WaitAndGetResult (default (CancellationToken));` just be `.Result;`?

We should consider removing(marking obsolete) `DocumentContext.ParsedDocument` if it's not reliable and move code to `await GetSemanticModelAsync`...

Afaik `GetSemanticModelAsync` never accesses UI thread and always call `.ConfigureAwait(false);`, so this `.Result;` should harmless.

If Roslyn has counterpart for this code and we really have plan to replace this with Roslyn logic, I guess `.Result;` is OK for now...

/cc: @slluis can you approve this?

mkrueger:
.Result is harmless here - right. The WaitAndGetResult () is a pattern from roslyn which I took - they're using it this way.

DavidKarlas:
Is this called from UI thread? In that case it can be multiple seconds beachball... I guess still better then current bug...

mkrueger:
multiple seconds is only possible before the project is loaded - the get semantic call is usually way faster than 'seconds'. But y better than the current bug and for template expansion there is no other way than delay atm :/

Therzok:
Yes, GetSemanticModelAsync is instant in most cases.

PR #2615, opened 2017-06-13
Fixed 'Bug 57017 - Toggling "unsafe" now requires a complete project

mkrueger:
reload to take effect'


mkrueger:
Any further problems with that one ?

PR #2632, opened 2017-06-14
[Ide] Fix invalid source being removed spew

Therzok:
The GSource handling code is now faster and these races happen easier now.


Therzok:
This is surfaced as a result of changing how GLib.Source.Remove works:
https://github.com/mono/gtk-sharp/pull/201/commits/6bb221c3c29ac9cd466d671c490d2210d99c28d2#diff-0a7f70ae80d005fc1bdb645c68bb1e81R72

Before, it would guard against double removal, now it doesn't anymore, so user errors surface.

PR #2635, opened 2017-06-14
Bug 55746 - [Wizard/Nuget] Tooltip showing xml character reference (&apos;)

DavidKarlas:


DavidKarlas:
@slluis I updated PR to not escape, I assumed `GLib.Markup.EscapeText` call was just swapped between `.Text` and `.Markup` assignment, which was not true, tnx for noticing that...
Now I looked at when `GLib.Markup.EscapeText` should be used and all assignments to `.Markup` correctly call `GLib.Markup.EscapeText` only on user input text.

PR #2649, opened 2017-06-15
[Core] Fix null ref exception on creating a new project

mrward:
Creating the configuration when creating a project would throw a null
reference exception since the CSharpCompilerParameters was trying to
notify its parent project that a property had changed but it was
not associated with a parent configuration.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetCompilerParameters.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetCompilerParameters.cs
@@ -35,7 +35,7 @@ namespace MonoDevelop.Projects
 	public abstract class DotNetCompilerParameters
 	{
 		public DotNetProject ParentProject {
-			get { return ParentConfiguration.ParentItem; }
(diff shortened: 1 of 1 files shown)


DavidKarlas:
Seems to me like exception will still happen here https://github.com/mono/monodevelop/blob/master/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CSharpCompilerParameters.cs#L117

KirillOsenkov:
Can't resist to plug http://source.monodevelop.com/#MonoDevelop.CSharpBinding/MonoDevelop.CSharp.Project/CSharpCompilerParameters.cs,112
:)

DavidKarlas:
@KirillOsenkov tbh, I 1st checked my theory at source.monodevelop.com, but since it sometimes feels maybe outdated, I went also in IDE to check for references... Later linked to github :)

KirillOsenkov:
I've only recently made it more-or-less reliably auto-update daily. Was struggling with some azure auth issues, but should be fixed now. Please let me know if you notice that it's more than a day old :) Thanks!

PR #2650, opened 2017-06-15
Add VSMEF prerequisites

KirillOsenkov:
This is the first stage of adding VSMEF. It adds NuGet packages but doesn't do anything else yet. 

This makes it easier to prototype and test things.

### DIFF
--- a/main/src/core/MonoDevelop.Startup/app.config
+++ b/main/src/core/MonoDevelop.Startup/app.config
@@ -83,6 +83,10 @@
 				<bindingRedirect oldVersion="0.0.0.0-1.0.31.0" newVersion="1.0.31.0" />
 			</dependentAssembly>
 			<dependentAssembly>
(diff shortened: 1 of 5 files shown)


main/src/core/MonoDevelop.Startup/app.config
```diff
@@ -83,6 +83,10 @@
 				<bindingRedirect oldVersion="0.0.0.0-1.0.31.0" newVersion="1.0.31.0" />
 			</dependentAssembly>
 			<dependentAssembly>
+				<assemblyIdentity name="System.Threading.Tasks.Dataflow" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
+				<bindingRedirect oldVersion="0.0.0.0-4.6.1.0" newVersion="4.6.1.0" />
```
Therzok: 😎

PR #2651, opened 2017-06-15
Add a MEF CompositionManager.

KirillOsenkov:
Adds an (unused for now) CompositionManager.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
@@ -1,0 +1,163 @@
+﻿// CompositionManager.cs
+//
+// Author:
+//   Kirill Osenkov <https://github.com/KirillOsenkov>
+//
+// Copyright (c) 2017 Microsoft
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
```diff
@@ -0,0 +1,163 @@
...
+			ReadAssembliesFromAddins (assemblies, "/MonoDevelop/Ide/TypeService/MefHostServices");
+
+			// spawn discovery tasks in parallel for each assembly
+			var tasks = new List<Task<DiscoveredParts>> (assemblies.Count);
+			foreach (var assembly in assemblies) {
+				var task = Task.Run (() => Discovery.CreatePartsAsync (assembly));
+				tasks.Add (task);
+			}
+
+			foreach (var task in tasks) {
```
Therzok: Maybe it's better to use Task.WhenAny here to save some time spent

Therzok: I.e. 

```
while (list.Count > 0) {
    var task = await Task.WhenAny (list);
    catalog = catalog.AddParts (task.Result);
    list.Remove (task);
}

KirillOsenkov: I was thinking about that. If you notice in my original PR I await Task.WhenAll and then iterate.

I've thought to save some time (potentially) by not waiting on all of them to complete, and start processing the first one while the others are still running.

Yours is even better in terms of perf, but what I'm worried about is it adds non-determinism to the composition. So the composition errors we'd get (and it's inevitable, people will run into composition errors all the time as part of regular development) would be non-deterministic depending on which parts were resolved in what order.

We should measure my approach and your approach and see if the optimization you propose is worth it. The overall idea is to hide the composition in a background thread while we're busy showing the start page and doing other stuff, so not sure we even need to optimize this.

KirillOsenkov: So I've tried the original approach, my current approach and your approach, and I see no difference whatsoever on a warm filesystem.
https://github.com/KirillOsenkov/MefHost/commit/d9f3de0dc4540a9f190138f62e26e5cbe91708c2

So probably either way is OK. I'd still be curious to see if you maybe see any difference on your hardware.

Therzok: Non-determinism makes sense here, sure, not sure if the perf gain is worth it, just wondering why we did this sequentially.

There will be cases where assemblies will be inserted in deterministic (i.e. `insertbefore` in extension point), so maybe this method of handling it is better

PR #2658, opened 2017-06-19
Enable VSMEF parallel to the old composition

KirillOsenkov:
A set of fixes that enables the new VSMEF composition live side-by-side with the old compositions for now. This will allow the WebTools addin to target the new composition while everything else is unchanged.


Therzok:
Looks good to me!

PR #2660, opened 2017-06-20
[Ide] Allow shell PathBar and Toolbar to coexist

Therzok:
By making a document implement IPathedDocumentEx.CanCoexist, you can attach a stock pathbar and a toolbar.

Bug 40652 - PathBar is broken for the forms previewer

cc @garuma @alanmcgovern @chkn 

This is currently untested, but this would be the only option to fix this without breaking API.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Content/IPathedDocument.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Content/IPathedDocument.cs
@@ -39,6 +39,11 @@ namespace MonoDevelop.Ide.Gui.Content
 		
 		event EventHandler<DocumentPathChangedEventArgs> PathChanged;
 	}
+
+	public interface IPathedDocumentEx
+	{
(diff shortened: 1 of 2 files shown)


Therzok:
The reason this is a new interface and doesn't extend IPathedDocument is that I want to add this property into IPathedDocument as soon as we can break API.

slluis:
I don't like this approach. Why do we need a CanCoexist property? why not just show the toolbar if something is added to it?

Therzok:
https://github.com/mono/monodevelop/commit/1926b377370542fc4375f888e1b503b6b987b741

It would esentially be reverting this commit which explains why it's good to have a mode. Maybe the Android designer needa to disable it and it should be enabled by default.

There needs to be a marker somehow, because the Android designers wants the default pathbar gone.

chkn:
@Therzok @garuma Any word on this? If I can't use the built-in path bar then I'll have to resort to an xwt one.

Therzok:
This has been fixed in the new document model.

PR #2670, opened 2017-06-21
[d15-3][Core] Remove custom framework definitions

mrward:
We get all target framework definitions via MSBuild now.

Fixes old MD framework names overriding the system ones.

Main reason for this change is to allow .NET 3.5 to be available in the project options for Unity since Unity is based on .NET 3.5. This change also shows the .NET 3.5/3.0 assemblies in the references dialog if the project targets .NET 3.5.

Only small problem that I saw when testing this change is that the few .NET 3.5 specific assemblies (e.g.Microsoft.Build.Utilities.v3.5.dll) are not resolved by MSBuild. MSBuild seems to not look in the correct 3.5-api folder. The other GAC assemblies for .NET 3.0, which falls back to the .NET 2.0 folder, seem OK. However I suspect Unity will not be using the v3.5 MSBuild dlls.


slluis:
Let's wait for a green build before merging.

PR #2675, opened 2017-06-23
Fixed 'Bug 57594 - Numbers syntax highlighted in namespace names,

mkrueger:
method names and property names'


slluis:
Shouldn't this be fixed upstream?

mkrueger:
Which upstream ? We >are< upstream :)

slluis:
I thought we get this from somewhere else, I guess I got confused.

mkrueger:
Ah we're not using the 'official' textmate grammar. It's too slow for us atm - it's replicating the c# grammar to get syntax info out of the text mate engine. We're already have roslyn for that.

PR #2676, opened 2017-06-23
Pool strings

Therzok:


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildEngineManager.cs
```diff
@@ -34,6 +34,8 @@ class MSBuildEngineManager: IDisposable
 		object localLock = new object ();
 		bool disposed;
 
+		internal Core.StringInternPool Pool { get; } = new Core.StringInternPool ();
```
Therzok: This is the only shared state I can put it on, the Engine itself is null for imported projects.

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SourceEditorView.cs
```diff
@@ -3194,7 +3194,7 @@ void ClearBracketMarkers ()
 				TextEditor.Options.ZoomChanged += value;
 			}
 			remove {
-				TextEditor.Options.ZoomChanged += value;
+				TextEditor.Options.ZoomChanged -= value;
```
KirillOsenkov: WOW. Great catch!

slluis:
What's the impact of this change?
Also please create a build lane to verify nothing breaks before merging.

Therzok:
https://github.com/mono/monodevelop/commit/a0b973750f1abfa03cdad0525ca8e4723907c33e

We save around 50MB of managed memory with Main.sln open.

Therzok:
Test failures unrelated.

PR #2679, opened 2017-06-23
Fix quick fix menu load issue

slluis:
Changed the way the quick fix menu is shown. It used to be a command
that generated a submenu with the fixes. Now the quick fix menu is
defined as an item set that contains the quick fix command array.

This is a workaround to a Cocoa issue. When a submenu is made visible,
there seems to be no way to hide it, even when the item that defines
the submenu is removed. By changing the way the quick fix menu is
generated, that menu doesn't need to be closed and recreated, it is
just updated.

Fixes bug #57704


slluis:
@manish do you want to verify before merging?

PR #2681, opened 2017-06-23
Offload work to Roslyn

Therzok:
This is a WIP to cleanup storage service implementations. Getting IPersistentStorageService2 working would mean we can cache typesystem and syntax info


Therzok:
I'll split the persistent service into another PR. That will imply doing a lot more hoops to get the thing merged.

Therzok:
This won't work properly without some things specified in https://github.com/mono/monodevelop/pull/2684

PR #2683, opened 2017-06-23
Move the WPF UI stub types to a stub Microsoft.VisualStudio.Text.UI.W…

KirillOsenkov:
…pf.dll.

That assembly is compiled from:
https://github.com/KirillOsenkov/TextUIWpfStub

It has the same signature as the official one but only has the 5 types we need.

This is necessary to be able to include Microsoft.CodeAnalysis.EditorFeatures.dll into our MEF composition.


KirillOsenkov:
This is not ready for review yet, just for @ToddGrun and myself for now.

ToddGrun:
@dpugh -- let's discuss on Monday

KirillOsenkov:
This is no longer needed.

PR #2684, opened 2017-06-24
Roslyn persistent storage services

Therzok:
Follow-up to https://github.com/mono/monodevelop/pull/2681


Therzok:
There are a few pending things to do before merging:
1. Figure out why there is nothing written to the index. I tried hacking the solution size threshold to be 1byte, still nothing written, even though I do SyntaxTreeIndex.PrecalculateAsync.
2. Go through OSSTool to approve ManagedEsent and SQLitePCL.raw

DavidKarlas:
I'm a bit confused about where is Sqlite used(I guess some Roslyn implementation of Storage is using it), but what does MonoDevelopPersistentStorageServiceFactory do?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopPersistentStorageLocationService.cs
```diff
@@ -0,0 +1,56 @@
+﻿//
+// MonoDevelopPersistentStorageServiceFactory.cs
```
DavidKarlas: Sync this

Therzok: :+1:

DavidKarlas:
I'm wondering since all Roslyn NuGets are in MonoDevelop.Core... maybe this Sqlite nugets should be there, other then that 👍

Therzok:
@DavidKarlas Will do, this is WIP still, trying to figure out how to use it.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
```diff
@@ -99,6 +99,21 @@
     <MSBuild_OSS_BinDir Condition="'$(OS)' == 'Unix' and '$(MSBuild_OSS_BinDir)' == '' and Exists('$(MSBuildToolsPath)\..\..\..\msbuild\15.0\bin\MSBuild.dll')">$(MSBuildToolsPath)\..\..\..\msbuild\15.0\bin\</MSBuild_OSS_BinDir>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Esent.Interop, Version=1.9.4.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
```
Therzok: ManagedEsent was added manually to MD.Core in https://github.com/mono/monodevelop/commit/81eb2bbdc2a80836cec2fcadccce6e016ccfc26c, so it never got restored.

Therzok:
Code-wise, it's good now. We got stuff running in parallel atm (i.e. roslyn diagnostic analyzers and our own custom code).

Should we remove them in a later PR or do it here? Before merging, we still need to go through OSSTool to include sqlite and managedesent (even though only used for composition, not called at all) as a dep.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopPersistentStorageLocationService.cs
```diff
@@ -0,0 +1,56 @@
...
+namespace MonoDevelop.Ide.TypeSystem
+{
+	[ExportWorkspaceService (typeof (IPersistentStorageLocationService), ServiceLayer.Host), Shared]
+	class MonoDevelopPersistentStorageLocationService : IPersistentStorageLocationService
+	{
+		public bool IsSupported (Workspace workspace) => workspace is MonoDevelopWorkspace;
+
+		public string GetStorageLocation (Solution solution)
+		{
+			return solution.FilePath != null ? Path.GetDirectoryName (solution.FilePath) : null;
```
Therzok: Maybe we should mirror the directory VS uses? Or something similar.

Therzok:
Submitted OSSTool for approval. Hunting down LoaderException issues with Microsoft.CodeAnalysis.Workspaces.Desktop on Mac.

Therzok:
Current issues:
1. MMap storage doesn't work on Mac. Seems like it's still running out of file handles.
2. We're in an impossible-to-solve position right now with having to use `Host` workspace kind to enable incremental analyzers and having to use `MonoDevelop` workspace kind to use our own temporary storage.

Therzok:
I will bump roslyn to 2.3.0-beta2 as part of this PR. They added an improvement for MemoryMappedFile handling.

We also now depend on https://github.com/mono/mono/pull/5116

Therzok:
Ignore https://github.com/mono/monodevelop/pull/2684/commits/e368a9bdd64d7afce0d47ba525611653d62f248a please

Therzok:
Even with https://github.com/mono/mono/pull/5116 and https://github.com/dotnet/roslyn/pull/19624 we run out of handles on OS X with Main.sln. Will do a few more tests tomorrow to see if I'm using the wrong version, but this is probably not going to work.

We might have to ask roslyn to change the TemporaryStorageService layer from Host to Default so we can plug our own.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopPersistentStorageLocationService.cs
Therzok: `${SolutionDir}/.vs/${SolutionName}/v15` before removing DON'T MERGE label.

Therzok:
Rebased on master.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopPersistentStorageLocationService.cs
slluis: We should store the data out of the solution dir if possible.

Therzok: UserConfig.Cache?

Therzok: i.e. `~/Library/Caches/VisualStudio/7.0/Roslyn/${SolutionName}/`?

slluis: Maybe just this: https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService.cs#L361

Therzok: The advantage of having the cache inside the solution directory means that every solution with projects that have the same name won't invalidate the cache of another project with the same name.

That TypeSystemService method won't work, as the cache is per solution, not per project.

slluis: I just pointed to an example of how it can be done. The same approach can be used for solution data.

Therzok: Makes sense! :+1:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -130,20 +133,22 @@ internal void OpenSolutionInfo (SolutionInfo sInfo)
...
 
-		internal MonoDevelopWorkspace (MonoDevelop.Projects.Solution solution) : base (services, "MonoDevelop")
+		internal MonoDevelopWorkspace (MonoDevelop.Projects.Solution solution) : base (services, "Host")
 		{
 			this.monoDevelopSolution = solution;
 			this.Id = WorkspaceId.Next ();
 			if (IdeApp.Workspace != null && solution != null) {
 				IdeApp.Workspace.ActiveConfigurationChanged += HandleActiveConfigurationChanged;
 			}
+			DiagnosticProvider.Enable (this, DiagnosticProvider.Options.Syntax);
```
Therzok: Note: don't do this when source analysis is disabled!

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.CodeActions/CodeFixMenuService.cs
```diff
@@ -217,10 +217,12 @@ static void ConfirmUsage (string id)
 
 		static async Task<ImmutableArray<Diagnostic>> GetDiagnosticsForDocument (ImmutableArray<DiagnosticAnalyzer> analyzers, Microsoft.CodeAnalysis.Document doc, ImmutableHashSet<string> diagnostics, CancellationToken token)
 		{
+			var sln = doc.Project.Solution;
```
KirillOsenkov: I thk it's btr to avd abbrvs (use "solution")

Therzok: k

NuGet.config
```diff
@@ -8,5 +8,6 @@
 		<add key="VSTest" value="https://dotnet.myget.org/F/vstest/" />
 		<add key="Templating" value="https://dotnet.myget.org/F/templating/" />
 		<add key="Azure AppService" value="https://www.myget.org/F/azure-appservice/api/v3/index.json" />
+		<add key="Roslyn MyGet" value="https://dotnet.myget.org/F/roslyn/api/v3/index.json" />
```
KirillOsenkov: One thing to be aware of is that this feed is transient and packages may disappear. It is always better to have them push the packages to nuget.org such that we can be sure they don't disappear and break our builds.

Therzok: We can update to official beta3 as it has the mmap fixes, but I'm using this for testing for now. Will revert to official one when I switch back to this branch.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopPersistentStorageLocationService.cs
mhutch: @slluis I think we should store it under the `.vs` folder adjacent to the solution. This has a few advantages:

* consistent with VS
* goes away when the solution is deleted
* easy for users to delete if there are issues

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopTemporaryStorageServiceFactory.cs
```diff
@@ -19,169 +19,11 @@ namespace MonoDevelop.Ide.TypeSystem
 	[ExportWorkspaceServiceFactory (typeof (ITemporaryStorageService), MonoDevelopWorkspace.ServiceLayer), Shared]
 	sealed class MonoDevelopTemporaryStorageServiceFactory : IWorkspaceServiceFactory
```
mkrueger: Can't you just remove that service factory  the Microsoft.CodeAnalysis.Host.TemporaryStorageServiceFactory should work then ? 

That one was just to replace the roslyn one with a custom implementation.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopPersistentStorageLocationService.cs
mkrueger: Use

public static string GetCacheDirectory (string fileName, bool forceCreation = false)

That works and same name is no problem because the full name is the key.

I think storing data in the solution directory is a bad approach. The project could be on a network drive or on a read only file system etc.

The cache goes away as well if it's not touched for a longer time.

Therzok:
Open VSTS issues are marked as not critical, this can go in.

Therzok:
This is good to go.

Therzok:
QA approved

PR #2688, opened 2017-06-26
[d15-3][DotNetCore] Downgrade .NET Core 1.x project templates

mrward:
Fixed bug #57760 - F# .NET Core projects do not compile
https://bugzilla.xamarin.com/show_bug.cgi?id=57760

Use older .NET Core 1.x project templates so that the F# projects
compile without any errors when only the .NET Core 1.0.4 SDK is
installed. Later versions of the F# project templates fail to build
due to an error about duplicate Compile items.

Since the .NET Core 1.x project templates for C# and F# are included
in the same NuGet package they have both been downgraded to the same
version.


slluis:
What's the effect in the c# templates? what changes?

mrward:
C# .NET Core project templates compared with stable VSMac 7.0:

 1. New project templates have a launchSettings.json file - used by VSMac to determine launch url.
 2. New web api project has a DotNetCliToolReference:
```
<DotNetCliToolReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Tools" Version="1.0.1" />
```
 3. New web app project has a PackageTargetFallback property, a new ValidationScriptsPartial.cshtml file, and a DotNetCliToolReference
```
    <DotNetCliToolReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Tools" Version="1.0.1" />
```
 4. New xunit project has a IsPackable false property and uses a preview VSTest version.
```
<PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0-preview-20170427-09" />
```
VS 2017 on Windows (15.3.0 preview 2.1) is using preview versions of the Microsoft.NET.Test.Sdk NuGet package for their test projects. VSMac 7.0 currently uses the 15.0.0 stable version of Microsoft.NET.Test.Sdk. Presumably at some point 15.3.0 will be stable.

Another possibility here would be to downgrade the 1.x project templates to match the version currently shipping with VSMac 7.0. I did not think you could mix and match beta1 and beta2 templates with a beta2 templating engine but that seems to work. However doing that we lose the fix where launching the web api project would use a launch url that showed the web api results since the old project templates do not have a launchSettings.json file.

manish:
This has been tested and no issues were found.

@Aguilex can we merge this?

Aguilex:
@mrward what about the missing Linux x64 tests?

mrward:
@Aguilex Looking at all the other pull requests they seem to have the same problem. It does not seem specific to this pull request. The wrench builds are the important ones.

manish:
As per @slluis we don't need to worry about Linux checks

@Aguilex can you merge this PR please?

PR #2693, opened 2017-06-27
[Ide] Use Markup for Key bindings prefs warnings

sevoku:
(fixes bug #57547)


slluis:
Why use the Markup property? there is no markup in the text.

sevoku:
This is a hack for a `CoreText` issue. Cocoa does not handle wrapping and sizing in some corener cases right. If a text has multiple lines without markup, it calculates the size correctly, but fails to render all lines. And if `AttributedString` is set (aka Markup) Cocoa fails to calculate the size correctly if wrapping is disabled. For the Markup sizing issue I've added a workaround in https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/TooltipPopoverWindow.cs#L177-L181.
Normal Text works correctly in 99%, but in this special case (maybe because of the special characters?) it fails, but can be fixed by using `AttributedString`/`Markup` without actual string attributes.

PR #2697, opened 2017-06-27
Moved ASP.NET Core templates and logic to separated AddIn

DavidKarlas:


main/src/addins/MonoDevelop.AspNetCore/Properties/AddinInfo.cs
```diff
@@ -0,0 +1,16 @@
...
+using System;
+using Mono.Addins;
+using Mono.Addins.Description;
+
+[assembly:Addin ("AspNetCore", 
+                 Namespace = "MonoDevelop",
+                 Version = MonoDevelop.BuildInfo.Version,
+                 Category = "Web Development")]
+
+[assembly:AddinName (".NET Core Support")]
```
mrward: ASP.NET Core Support?

mrward: Should we also move the ASP.NET Core specific code from the [DotNetCoreRunConfiguration class](https://github.com/mono/monodevelop/blob/master/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreRunConfiguration.cs) and move it into the AspNetCore addin? We could create an AspNetCoreRunConfiguration class that maybe derives from DotNetCoreRunConfiguration.

Could we have an AspNetCoreProjectExtension that is available for project's that have the AspNetCore project capability?

mrward: Also maybe some parts of the DotNetCoreExecutionHandler should be moved across into an AspNetCoreExecutionHandler? There is code in there for launching the browser.

mrward: Also maybe the ASP.NET Core specific parts of the DotNetCoreRunConfigurationEditorWidget should be split out and moved into the AspNetCore addin? Although maybe the whole thing should go across since it looks like if the project is not a web project there is no extra UI added.

DavidKarlas: Last time I tried splitting DotNetCoreRunConfiguration into 2 was problem with breaking API :( I will take another look

DavidKarlas:
@mrward addressed comments and converted more code to AspNetCore, unfortunately for backward compatibility reasons, I didn't delete code in DotNetCore addin, this can be done when we bump to VSfM 8.0 version

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -0,0 +1,83 @@
...
+using System;
+using MonoDevelop.Core;
+using MonoDevelop.Core.Execution;
+using MonoDevelop.DotNetCore;
+using MonoDevelop.Projects;
+
+namespace MonoDevelop.AspNetCore
+{
+	[ExportProjectModelExtension]
+	public class AspNetCoreProjectExtension : DotNetCoreProjectExtension
```
mrward: This probably should not be public.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreRunConfigurationEditor.cs
```diff
@@ -51,116 +51,26 @@ public override Control CreateControl ()
...
 		}
 
 		public override bool Validate ()
 		{
-			return widget.ValidateCore ();
+			return widget.Validate ();
 		}
 	}
 
 	class DotNetCoreRunConfigurationEditorWidget : DotNetRunConfigurationEditorWidget
```
mrward: Can this class be deleted now? Does not seem to be used.

DavidKarlas: No because it still has function, that is call `DotNetRunConfigurationEditorWidget` with `false` in constructor which results in Mono runtime selection pad removed

DavidKarlas:
Removed public from project extension class

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/DotNetCoreExecutionHandler.cs
```diff
@@ -0,0 +1,107 @@
...
+using MonoDevelop.Ide;
+using System;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Net.Sockets;
+
+namespace MonoDevelop.AspNetCore
+{
+	class AspNetCoreExecutionHandler : IExecutionHandler
```
mrward: Filename is DotNetCoreExecutionHandler.cs but the class name is AspNetCoreExecutionHandler.

DavidKarlas: Fixed

PR #2698, opened 2017-06-27
Roslyn search category

Therzok:
Currently targeting `roslyn-persistent-storage-services`. Will switch to master when the former is merged.

https://www.screencast.com/t/t1wVpFEwr

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp/ProjectSearchCategory.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp/ProjectSearchCategory.cs
@@ -152,18 +152,35 @@ namespace MonoDevelop.CSharp
 							if (!MatchesTag (searchPattern.Tag, result))
 								continue;
 
-							int rank = 0;
-							if (result.MatchKind == NavigateToMatchKind.Exact)
-								rank = result.Name.Length * 11;
-							else if (result.MatchKind == NavigateToMatchKind.Prefix)
-								rank = result.Name.Length * 5;
-							else if (result.MatchKind == NavigateToMatchKind.Substring)
-								rank = result.Name.Length * 2;
-							else
-								rank = result.Name.Length;
+							int laneLength = result.NameMatchSpans.Length;
+							int index = laneLength > 0 ? result.NameMatchSpans [0].Start : -1;
 
-							if (result.IsCaseSensitive)
-								rank *= 2;
+							int rank = 0;
+							if (result.MatchKind == NavigateToMatchKind.Exact) {
+								rank = int.MaxValue;
+							} else if (result.MatchKind == NavigateToMatchKind.Prefix) {
+								rank = int.MaxValue - (result.Name.Length - 1) * 10 - index;
+							} else {
+								int patternLength = searchPattern.Pattern.Length;
+								rank = searchPattern.Pattern.Length - result.Name.Length;
+								rank -= index;
+
+								rank += laneLength * 100;
+
+								// Favor matches with less splits. That is, 'abc def' is better than 'ab c def'.
+								int baseRank = (patternLength - laneLength - 1) * 5000;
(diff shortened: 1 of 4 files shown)


main/src/addins/CSharpBinding/MonoDevelop.CSharp/ProjectSearchCategory.cs
```diff
@@ -369,161 +128,56 @@ static SymbolCache GetSymbolInfos (CancellationToken token)
...
+					).ConfigureAwait (false);
+
+					foreach (var results in aggregatedResults) {
+						foreach (var result in results) {
+							if (!MatchesTag (searchPattern.Tag, result))
+								continue;
+
+							int rank = 0;
+							if (result.MatchKind == NavigateToMatchKind.Exact)
+								rank = result.Name.Length * 11;
```
Therzok: @mkrueger I need your brain on this.

Calculating the match rank here, what would be best? We know what spans are matched (start, end, length), the name of the symbol that gets matched and the match kind.

Therzok:
Remaining TODO:
* [ ] Fix performance, it takes a bit to give initial results and subsequent results
* [ ] Fix match ranking

main/src/addins/CSharpBinding/MonoDevelop.CSharp/ProjectSearchCategory.cs
```diff
@@ -369,161 +128,73 @@ static SymbolCache GetSymbolInfos (CancellationToken token)
...
+												searchPattern.Pattern,
+												token))
+					).ConfigureAwait (false);
+
+					foreach (var results in aggregatedResults) {
+						foreach (var result in results) {
+							if (!MatchesTag (searchPattern.Tag, result))
+								continue;
+
+							int laneLength = result.NameMatchSpans.Length;
```
Therzok: Tagging @slluis here, I took bits of the match ranking from lane matcher (to avoid allocations and double processing of strings - matching is already done by roslyn, we just need to calculate a rank). Hopefully this is correct enough to give good results.

slluis: This needs to be verified with unit tests.

slluis: Also, which matching algorithm does Roslyn use? can't we keep the matching logic we have now?

Therzok: The results we get by passing our filter is the same matching algorithm VSWindows has. We can re-calculate match ranking for items coming in from roslyn as data, currently I wanted to relay everything there.

It is similar to the lane matcher, as it handles string skipping, but doesn't offer a rank for a given item.

Therzok: Based on the filter string, they have 3 matchers:
https://github.com/dotnet/roslyn/blob/master/src/Workspaces/Core/Portable/PatternMatching/PatternMatcher.cs#L74-L101

Their matcher also supports patterns such as the one requested in this bug: https://bugzilla.xamarin.com/show_bug.cgi?id=56636

slluis: I'd like to know how the VS matcher compares to ours. I don't want to lose functionality if possible.

Therzok: Would unit tests comparing results from static data suffice or would you rather have integration tests for this?

slluis: Static data is ok

PR #2702, opened 2017-06-28
[Core] Optimize MSBuild evaluation whitespace allocations.

Therzok:
This reduces the number of duplicate strings from 40MB to 30MB. We now
avoid substringing and creating a lot of string junk and reusing simple
whitespace strings.

Also cache the first part of the string, as most of them are just
newlines

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core/StringTable.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core/StringTable.cs
@@ -462,7 +462,7 @@ namespace MonoDevelop.Core
         private string AddItem(string chars, int start, int len, int hashCode)
         {
             var text = chars.Substring(start, len);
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Core/StringTable.cs
```diff
@@ -462,7 +462,7 @@ private string AddItem(char[] chars, int start, int len, int hashCode)
         private string AddItem(string chars, int start, int len, int hashCode)
         {
             var text = chars.Substring(start, len);
-            AddCore(chars, hashCode);
+            AddCore(text, hashCode);
```
Therzok: Bug fix for this and unit test submitted upstream.

Therzok:
Wrench build failed on restoring md-addins nuget packages. Ran the test suite locally, green for me.

PR #2704, opened 2017-06-28
[Core] Drop another 10MB of duplicate strings

Therzok:
Intermediate parsing strings can be cached in the pool. Total string allocations coming in from the Project Model dropped a lot.
Previously, MSBuildEvaluationContext.EvaluateReference would create 250MB of strings, now it only allocates 3.5MB total

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildEvaluationContext.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildEvaluationContext.cs
@@ -500,12 +500,15 @@ namespace MonoDevelop.Projects.MSBuild
 			i += 2;
 			int j = FindClosingChar (str, i, ')');
 			if (j == -1) {
-				val = str.Substring (start);
+				lock (project.Pool)
+					val = project.Pool.Add (str, start, str.Length - start);
 				i = str.Length;
 				return false;
 			}
 
-			string prop = str.Substring (i, j - i).Trim ();
+			string prop;
+			lock (project.Pool)
+				prop = project.Pool.Add (str, i, j - i).Trim ();
 			i = j + 1;
 
 			bool res = false;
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildEvaluationContext.cs
```diff
@@ -500,12 +500,15 @@ bool EvaluateReference (string str, List<MSBuildItemEvaluated> evaluatedItemsCol
...
+				lock (project.Pool)
+					val = project.Pool.Add (str, start, str.Length - start);
 				i = str.Length;
 				return false;
 			}
 
-			string prop = str.Substring (i, j - i).Trim ();
+			string prop;
+			lock (project.Pool)
+				prop = project.Pool.Add (str, i, j - i).Trim ();
```
alanmcgovern: Why are you trimming the pool output? Surely we should intern the trimmed value?

Therzok: That means modifying the intern pool code so it supports a trim mode, since we don't do a substring _unless_ it's a cache miss.

The total amount of strings allocated by trim is 20kb across a few hundred/thousand objects, would not say it's a bottleneck. Can do though, but not sure we should block on the lower 1%

alanmcgovern: My concern was that we'd cache an untrimmed string and then we'd repeatedly trim it after the fact and then *not* pool that trimmed version. It may not be a problem in practice.

basically i was thinking of:
```
prop = project.Pool.Add (str, i, j - i)
prop = project.Pool.Add (prop.Trim ());
```

Therzok: Yeah, it's not a problem in practice. Let's keep the code simple. With that code, you'd allocate the trimmed string anyway, so there is no real benefit to save these 20kbs.

Having support for trimming in StringInternPool means writing yet-another-start-end-whitespace checker to avoid doing the trim itself on checking.

PR #2706, opened 2017-06-29
[NuGet] Allow NuGet action to be cancelled on closing solution

mrward:
Fixed bug #53795 - Failure to install blocks the IDE
https://bugzilla.xamarin.com/show_bug.cgi?id=53795

Previously when closing the solution or IDE when a NuGet package
action is still running a dialog would be displayed saying that the
solution cannot be closed until the NuGet operation was completed.
This was true with NuGet v2 but with NuGet v3 and above the NuGet
operations can be cancelled. So now the dialog allows the current
operation to be cancelled when the Yes button is clicked. If the
operation is taking a while a busy spinner image will be displayed in
the dialog until the dialog is closed or the operation is cancelled
and then the solution will be closed.

installingnugetpackagesonclosesolution

The above dialog is not shown if a only a NuGet restore is being run. In this case a request will be made to cancel the restore and the solution will be closed.

### DIFF
--- a/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement.Commands/PackageManagementStartupHandler.cs
+++ b/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement.Commands/PackageManagementStartupHandler.cs
@@ -25,14 +25,15 @@
 // THE SOFTWARE.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using MonoDevelop.Components.Commands;
 using MonoDevelop.Core;
 using MonoDevelop.Ide;
+using MonoDevelop.PackageManagement.Gui;
 using MonoDevelop.Projects;
 using NuGet.Common;
-using System.Collections.Generic;
 
 namespace MonoDevelop.PackageManagement.Commands
 {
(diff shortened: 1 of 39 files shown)


slluis:
I don't think it is necessary to show that dialog at all. If a NuGet operation is in progress and the user closes the solution, NuGet should just be cancelled. Maybe we could show a dialog if the NuGet operation is not stopping for some reason.

mrward:
Maybe we can do something similar to how the debugger works with the debugger is busy dialog. So we would show a dialog if NuGet is taking longer than expected to stop, with an option to just continue with closing the solution.

One problem here is that if say an NuGet package install is taking a while, then closing the solution would cancel the installation. Or maybe that is not a problem.

slluis:
It may be a problem if canceling the installation can break the solution.

mrward:
I was more thinking of a user wondering why a NuGet package was not installed in the project the next time they opened it.

When you cancel an install then NuGet will rollback the changes.

slluis:
Ok, I was thinking more about restoring packages.

mrward:
Maybe we should split the actions into two separate groups. Those that can be silently cancelled, such as restore, and the rest (install, update, remove) which we should show a dialog before cancelling.

For the restore case if that is taking too long to cancel maybe we can show dialog indicating that NuGet is busy and offer to close the solution anyway.

slluis:
Makes sense.

mrward:
Updated this. The dialog is shown if an install or uninstall is currently running when the solution is being closed. If a just restore is being run then no dialog will be displayed and instead the restore will be cancelled and the solution will be closed.

PR #2718, opened 2017-07-05
Bug 57840 - Adding .NET Standard project to existing solution/directo…

DavidKarlas:
…ry structure rewrites/reformats *all* source code on disk

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngineProjectTemplatingProvider.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngineProjectTemplatingProvider.cs
@@ -218,7 +218,7 @@ namespace MonoDevelop.Ide.Templates
 			// Format all source files generated during the project creation
 			foreach (var p in workspaceItems.OfType<Project> ()) {
 				foreach (var file in p.Files)
-					if (!filesBeforeCreation.Contains ((string)file.FilePath)) //Format only newly created files
+					if (!filesBeforeCreation.Contains ((string)file.FilePath, FilePath.PathComparer)) //Format only newly created files
 						await FormatFile (p, file.FilePath);
 			}
 			processResult.SetFilesToOpen (filesToOpen);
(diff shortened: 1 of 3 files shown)


DavidKarlas:
Since all file paths are obtained from file system just simple string compare feels sufficient using `.Contains(string)` or should I change to use FilePath comparer?

DavidKarlas:
Nvm, I just added PathComparer, better safe then sorry

slluis:
Can this be unit tested?

DavidKarlas:
I will take a look if I can unit test this

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngineProjectTemplatingProvider.cs
```diff
@@ -147,6 +147,9 @@ public IEnumerable<SolutionTemplate> GetTemplates ()
 			var parameters = GetParameters (solutionTemplate, config);
 			var templateInfo = solutionTemplate.templateInfo;
 			var workspaceItems = new List<IWorkspaceFileObject> ();
+
+			var filesBeforeCreation = Directory.GetFiles (config.ProjectLocation, "*", SearchOption.AllDirectories);
```
Therzok: I wonder if it's better to use a HashSet, rather than an Array here. The number of files can be big, this leads to O(n^2) iterations in the checks below

DavidKarlas:
@slluis added unit tests, lets wait for Wrench to process

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/ProjectTemplatingProvider.cs
```diff
@@ -82,7 +82,7 @@ ProjectCreateInformation CreateProjectCreateInformation (NewProjectConfiguration
 			cinfo.ProjectName = config.ProjectName;
 			cinfo.SolutionName = config.SolutionName;
 			cinfo.ParentFolder = parentFolder;
-			cinfo.ActiveConfiguration = IdeApp.Workspace.ActiveConfiguration;
+			cinfo.ActiveConfiguration = IdeApp.Workspace?.ActiveConfiguration ?? new SolutionConfigurationSelector (string.Empty);
```
Therzok: Wouldn't defaulting to `ConfigurationSelector.Default` be better here?

slluis:
There is one last issue that should be fixed for bug 57840. If the project is being added to an existing solution, we should use the solution policy to format the files instead of the project policy.

DavidKarlas:
@slluis check last commit which uses solution formatting rules... problem is that at time we are doing formatting project is not added to solution yet, that is done later at https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs#L641
So if `parentFolder` has value, it means we are adding project to existing solution + new project doesn't have it's own specific formatting rules yet anyway, hence just use `parentFolder` policies...

slluis:
Can you also make a PR for master?

PR #2720, opened 2017-07-05
Fix bug #42063 - Evaluated MSBuild properties don't respect global pr…

slluis:
…operty providers


main/tests/UnitTests/MonoDevelop.Projects/MSBuildProjectTests.cs
```diff
@@ -1460,6 +1461,35 @@ public void AddIncludeItemBeforeUpdateItemOfSameKind ()
 			Assert.AreEqual (expectedXml, xml);
 			p.Dispose ();
 		}
+
+		[Test]
+		public void GlobalPropertyProvider ()
```
Therzok: Maybe it's worth adding a positive test to check whether the property is overridden if there's two custom global property providers which set the same key?

slluis: Can do

PR #2721, opened 2017-07-05
Fix bug #40942 - XS is leaving empty properties in the project file

slluis:
A new flag is generated while evaluating properties: DefinedMultipleTimes.
This flag is used to determine if a property is defined only once in the
project file, or if it is defined more than once (or in other imported
project files). In the first case, property definitions that set the default
value can be safely removed, since we know that the property definition is not
overriding another definition.


slluis:
Don't merge yet, I want to add unit tests.

PR #2722, opened 2017-07-05
[NuGet] Add support for AssetTargetFallback

mrward:
Fixed bug #57093 - Add support for AssetTargetFallback
https://bugzilla.xamarin.com/show_bug.cgi?id=57093

.NET Core 2.0 preview 2 uses an AssetTargetFallback MSBuild property
defined in one of the imported SDK MSBuild files. This is used instead
of the PackageTargetFallback property when determining if a NuGet
package is compatible. Currently the AssetTargetFallback property
is set to net461 to allow .NET Core projects to reference NuGet
packages that include assemblies that target the full .NET Framework.
The supported fallback frameworks are now added to the generated
project.assets.json file when a .NET Core 2.0 project is restored.
This then allows .NET Core 2.0 projects to use .NET Framework
NuGet packages.

In order to fix this bug NuGet was upgraded from 4.3.0.2418 to 4.3.0.4199. There were several changes to the NuGet API in this upgrade, including the removal of nine assemblies. So I have added a card to the 15.4 Trello board for this branch to be tested/approved by QA to make sure NuGetizer 3000, NuGet and .NET Core have not regressed since this is a larger change than I was expecting to have to make to fix the original bug.


mrward:
This has now been tested/verified by QA.

PR #2733, opened 2017-07-07
Replace SharpZipLib package with JetBrains.SharpZipLib.Stripped

sevoku:
The bump is required to support ZIP files with
versions not supported by the old SharpZipLib 0.86


mkrueger:
Works on linux.

Therzok:
Works on Windows!

PR #2734, opened 2017-07-07
Fixed 'Bug 57866 - [VSFeedback Ticket] #443161 - quick actions new

mkrueger:
type tooltip preview obscures the context menu'

Was a complex fix because os x doesn't let me get the submenu position
- so the solution was a work around: position the tooltip around the
context menu but not at the right side.


slluis:
What are all those changes in ContextMenuExtensionsMac.cs required for?

mkrueger:
Basically a code cleanup. I hoped to get the submenu location somehow and removed my old work around - but it seems not to be a way to get it :/.

slluis:
@mkrueger @Therzok we were having a discussion here. What doesn't work about getting the location of a menu?

Therzok:
Oh, I missed the comments.

PR #2739, opened 2017-07-10
Fixed 'Bug 58023 - Semantic colorization disappears on method

mkrueger:
parameters'

The bug was in that cache layer - but caching on this level isn't
required at all it needs to be done in the semantic highlighting
itself. Which is the case for our c# semantic highlighting as well for
the roslyn one so I decided to remove that.


slluis:
@mkrueger will you make a PR for 15.4?

PR #2745, opened 2017-07-10
[Core] Expose shared cache methods to StringInternPool.

Therzok:
The normal instance methods are really good only for serial code. If parallel code uses it, it's better to use the thread safe global cache rather than the local cache.

Fixes some abnormal behaviour in unlucky scenario where lock contention in bg
can reach 20s for evaluating Main.sln

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core/StringInternPool.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core/StringInternPool.cs
@@ -56,5 +56,20 @@ namespace MonoDevelop.Core
 		{
 			return table.Add (chars);
 		}
+
+		public static string AddShared (StringBuilder chars)
+		{
+			return StringTable.AddShared (chars);
+		}
+
+		public static string AddShared (string chars)
+		{
+			return StringTable.AddShared (chars);
+		}
+
+		public static string AddShared (string chars, int start, int len)
+		{
+			return StringTable.AddShared (chars, start, len);
+		}
 	}
 }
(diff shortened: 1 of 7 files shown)


slluis:
Won't strings will stay in memory forever now?

Therzok:
They're evicted as soon as a string has the same hashcode.

Even before, the cache would write-back to the shared string pool, now it's just not writing to a local pool anymore, as that one is not threadsafe.

Therzok:
I'll do some analysis to see how the cache is laid out on different project loads (without manually evicting strings)

slluis:
Sure, but the string in the pool will stay there forever. If you open and close a solution, all strings used by the solution won't be freed.

Therzok:
Simple project with 4 solutions and nothing really special:

String Count: 1889 - Strings Lengths: 271599

After main.sln is opened:
String Count: 27596 - Strings Lengths: 6179707

After main.sln is closed and the initial project is opened again:
String Count: 27596 - Strings Lengths: 6179707

Opening md-addins.sln afterwards:
String Count: 45567 - Strings Lengths: 9150683

So, that means we have 6mb of strings that will live when smaller projects are opened. But switching to larger solutions will thrash the cache eventually, as the string count max is 65k.

String Lengths is basically `entries.Sum (str => str?.Length ?? 0)`.

There are wins and losses to be had here:
* Reopening Main.sln will lead to allocation-less evaluation for example.
* There's not much thrashing going on until you reach the 65k limit.
* It's ~20mb of managed strings being kept in memory, which will eventually get evicted. On the other hand, before caching them like this, we had 50mb of string duplicates just when having Main.sln open. So, we're trading here.
* The main culprit of fixing the issues by using this technique was duplicate strings. That meant we could do allocation-less string interning.

This code does not affect the static shared cache, the initial intern pool PRs did this anyway, now I'm just eliminating lock contention.

I'm fine with a technically 20MB of managed strings being kept in memory for the sake of faster project model evaluation and fewer GCs.

PR #2747, opened 2017-07-10
[A11y] Fix the keyboard focus on the editor

iainx:
Now that the whole of the shell UI can be focused, lots of controls like to
steal the keyboard focus meaning that when we expect to be able to type in the
editor, we often can't. To fix this, when opening a new document, we have the
editor grab focus.


slluis:
Can you please update this PR like https://github.com/mono/monodevelop/pull/2746?

iainx:
@slluis sorry, I missed your request for this, and forgot about it.

PR #2445, opened 2017-05-16
Unify project & assembly ref resolution

mhutch:
For early feedback.

* We now call the ResolveReference target instead of ResolveAssemblyReferences. This means that the resultant set of ReferencePath items is more complete.

* Project references are now resolved from ReferencePath items.

* Resolved AssemblyReference items have full metadata. This means, for example, that we can identify which references are implicit, which come from the framework, which come from specific PackageReferences, etc. This should allow the Dependencies node to be more accurate.

* Since xbuild doesn't have this metadata, I also removed support for xbuild.

PR #2753, opened 2017-07-11
[DotNetCore] Support transitive project references

mrward:
Fixed bug #57990 - Transitively referenced SDK projects do not show up
in IntelliSense
https://bugzilla.xamarin.com/show_bug.cgi?id=57990

With three .NET Core projects: LibC references LibB references LibA
The types defined by LibA were not available to LibC since it was not
directly referencing LibA. .NET Core projects support transitive
project references so the types defined in LibA should be available to
LibC.

 - [x] Provide DotNetProject API to return referenced projects with alias information
 - [x] Type system now using new DotNetProject API when getting project references
 - [x] .NET Core addin returns transitive project references
 - [x] DotNetCoreProject API needs a review - Is there a better API or way to do this?
 - [ ] ~~Remove code duplication in DotNetProject. New method has code similar to the existing GetReferencedAssemblyProject.~~
 - [x] Tests to be added for new DotNetProject API
 - [ ] ~~Tests to be added for type system changes when finding project references~~ Cannot see a simple way to test this.


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -608,5 +608,39 @@ void UpdateHiddenFiles (IEnumerable<ProjectFile> files)
...
+		}
+
+		/// <summary>
+		/// Recursively gets all project aliases for .NET Core projects. Calling
+		/// base.OnGetReferencedAssemblyProjectAliases returns the directly referenced project aliases.
+		/// </summary>
+		void GetProjectAliases (HashSet<string> traversedProjects, List<DotNetProjectAliases> projectAliases, ConfigurationSelector configuration)
+		{
+			foreach (var projectAlias in base.OnGetReferencedAssemblyProjectAliases (configuration)) {
+				if (traversedProjects.Contains (projectAlias.Project.ItemId))
```
Therzok: `if (!traversedProjects.Add())`?

slluis:
In PR https://github.com/mono/monodevelop/pull/2445 @mhutch is introducing new API for getting all information about references. AssemblyReference will have an Aliases property. In fact, the list of references is obtained from an msbuild target, so maybe for .net core projects it will take into account transitive references, which would make this PR unnecessary. It should be verified.

mhutch:
FWIW, I did some more testing on my PR and found a blocking bug in MSBuild: https://github.com/dotnet/sdk/issues/1393

mrward:
OK I have updated the code to use the better API from the unify project/assembly references pull request PR #2445. The changes included from the other PR were the new DotNetProject.GetReferences method and the extra properties on the AssemblyReference.

Also added some tests.

mhutch:
Some of that code looks suspiciously similar to mine :)

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
```diff
@@ -873,13 +873,25 @@ public Task<IEnumerable<AssemblyReference>> GetReferencedAssemblies (Configurati
...
+		public Task<List<AssemblyReference>> GetReferences (ConfigurationSelector configuration)
+		{
+			return BindTask (async ct => {
+				return await ProjectExtension.OnGetReferences (configuration, ct);
+			});
+		}
+
+		public Task<List<AssemblyReference>> GetReferences (ConfigurationSelector configuration, CancellationToken token)
+		{
+			return ProjectExtension.OnGetReferences (configuration, token);
```
slluis: For correctness you should use BindTask here too. The token provided to OnGetReferences would be a combination of the `token` parameter and the token provided by BindTask, created with `CancellationTokenSource.CreateLinkedTokenSource ()`

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -620,5 +620,52 @@ void UpdateHiddenFiles (IEnumerable<ProjectFile> files)
...
+		protected override async Task<List<AssemblyReference>> OnGetReferences (
+			ConfigurationSelector configuration,
+			System.Threading.CancellationToken token)
+		{
+			var references = new List<AssemblyReference> ();
+
+			var traversedProjects = new HashSet<string> ();
+			traversedProjects.Add (Project.ItemId);
+
+			await GetProjectAssemblyReferences (traversedProjects, references, configuration, token);
```
slluis: Isn't OnGetReferences() supposed to return all references, not only project references?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/AssemblyReference.cs
```diff
@@ -32,14 +32,58 @@ namespace MonoDevelop.Projects
 {
 	public sealed class AssemblyReference
 	{
+		Dictionary<string, string> metadata;
+
 		public AssemblyReference (FilePath path, string aliases = null)
 		{
 			FilePath = path;
-			Aliases = aliases ?? "";
+			metadata = new Dictionary<string, string> { { "Aliases", aliases } };
```
slluis: For consistency with other API, I'd like metadata to be made available through a Metadata property of type `IReadOnlyPropertySet`, like here https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/TargetEvaluationResult.cs#L58.

This property set can be created like this: https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs#L1287.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/AssemblyReference.cs
```diff
@@ -26,20 +26,71 @@
...
+		}
+
+		string GetMetadata (string name)
+		{
+			return metadata.GetValue (name);
+		}
+
+		bool MetadataIsTrue (string name)
+		{
+			return string.Equals (GetMetadata (name), "true", StringComparison.OrdinalIgnoreCase);
```
slluis: MSBuildPropertyGroupEvaluated already implements boolean conversion logic, so this would be better: `return metadata.GetValue(name, false);`

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/AssemblyReference.cs
```diff
@@ -26,20 +26,71 @@
...
+
+		public AssemblyReference (FilePath path, IReadOnlyPropertySet metadata)
+		{
+			FilePath = path;
+			this.metadata = metadata;
 		}
 
 		public FilePath FilePath { get; private set; }
-		public string Aliases { get; private set; }
+		public string Aliases => GetMetadata ("Aliases") ?? "";
```
slluis: better use of api: `metadata.GetValue ("Aliases","")`

PR #2755, opened 2017-07-11
[Mac] Make GetApplicationUrl tests resilient to any kind of system.

Therzok:


Therzok:
Possibly should add some comments:
CF calls do not work if the extension is not `plist`. So using the `Info.plist.in` does not work.

main/tests/MacPlatform.Tests/CoreFoundationTests.cs
```diff
@@ -56,19 +57,52 @@ public static void TestURL ()
 		[Test]
 		public static void TestApplicationUrls ()
 		{
-			string test = "/Applications/Xamarin Studio.app/Contents/Info.plist";
-			string[] results = MonoDevelop.MacInterop.CoreFoundation.GetApplicationUrls (test, MonoDevelop.MacInterop.CoreFoundation.LSRolesMask.All);
+			string dir = nameof (TestApplicationUrl);
```
slluis: Shouldn't this be a temp dir?

Therzok: That's an even better idea!

Therzok:
PR updated

PR #2761, opened 2017-07-13
Fixed 'Bug 58060 - [VSFeedback Ticket] #454760 - "Find References"

mkrueger:
fails leaving an error in the log'

The bug was in the find all references command. Not in the find
references command.


DavidKarlas:
Can you reproduce the bug? I think code change itself is good and makes a lot of sense, but I don't see how it fixes bug.

mkrueger:
lookup.Symbol.ContainingType == null in the search type case. Never used that command on a type - so it crashes in that case. I'll look into writing a test for it. 

Basically it uses the VS.NET find references  - when we get the EditorFeatures we could use their commands directly…

PR #2764, opened 2017-07-13
Add VS editor references to MonoDevelop.Core.

KirillOsenkov:
MonoDevelop.Core needs explicit references to VS Editor .dlls because they are being added implicitly by Microsoft.CodeAnalysis.EditorFeatures.dll but as indirect references they are being resolved from the VSSDK folder. By referencing these explicitly we make sure they're uniformly resolved from the single location as MonoDevelop.Ide.

This avoids double-writes during the build for these assemblies.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="4.0">
   <Import Project="..\..\..\MonoDevelop.props" />
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


slluis:
Shouldn't all those VS editor assemblies be referenced in MD.Ide?

KirillOsenkov:
Yes, they are referenced by MD.Ide. However MD.Core references Microsoft.CodeAnalysis.EditorFeatures.dll, which in turn has these dependencies. So when RAR (ResolveAssemblyReference) runs for MD.Core, it visits the reference closure and finds the editor binaries, but doesn't know where to resolve them from, and instead of taking them from the NuGet package (like MD.Ide) it resolves them from VS SDK (and it's never what we want, because we don't want our build to depend on VS SDK installed).

By having these references explicitly (I just copied them from MD.Ide, just made sure they're not copy local to avoid a double-write) RAR uses the same set of assemblies for both MD.Core and MD.Ide.

KirillOsenkov:
Oh, and the reason I'm making this change is to prevent unnecessary overbuilding. Earlier, MD.Core would see the editor binaries in output from the NuGet package, and overwrite them with the editor binaries from VS SDK, thus causing a rebuild of everything. Then, MD.Ide would see the "wrong" editor binaries in the output, and overwrite them again with the versions from the NuGet package. They would keep flip-flopping like this, causing unnecessary file copies and invalidating parts of our build. 
With my change the incremental build should be slightly faster because the editor binaries don't have to be copied to the output.

slluis:
My question was more about why MD.Core is referencing Microsoft.CodeAnalysis.EditorFeatures.dll, since it is not used there.

KirillOsenkov:
If https://github.com/mono/monodevelop/pull/2780 is committed I will close this one.

PR #2765, opened 2017-07-13
Adds a Diagnostics menu item to put a mark in the log.

iainx:
Adding a mark in the log can be helpful to find a point in a busy log. Mark the
log before doing an action, and then it can be easier to work out when you
started the action


Therzok:
Should this only be enabled in debug mode?

PR #2768, opened 2017-07-13
[policies] Fix NewLinesForBracesInTypes for Mono C# policy

VincentDondain:
According to the mono coding guidelines document: http://www.mono-project.com/community/contributing/coding-guidelines/#where-to-put-braces
(where to put braces in particular), "Classes and namespaces go like if statements, differently than methods".

good:

namespace N {
    class X {
        ...
    }
}

bad:

namespace N
{
    class X
    {
        ...
    }
}

So, to summarize:

Statement                      Brace position
Namespace                      same line
Type                           same line
Method (including ctor)        new line
Properties                     same line
Control blocks (if, for…)      same line
Anonymous types and methods    same line


slluis:
The formatting of the Mono policy was set many years ago, and I don't see the need to change it now.

PR #2775, opened 2017-07-14
Fixed 'Bug 58139 - Paste corrupts inserted text'

mkrueger:
It's just a guess what could go wrong. The new text is now formatted
before it's inserted in the text - not afterwards. I assume that there
may be a race when formatting it with the on the fly formatter
afterwards.
I couldn't reproduce the reported bug.


mkrueger:
Don't merge - need the confirmation that the bug is fixed. I assume it's a race somewhere - couldn't repro the bug.

slluis:
If we apply this solution, doesn't it mean that it won't be possible to undo the formatting action?

mkrueger:
Undo is handled in the buffer and at the end it's a buffer change.

PR #2778, opened 2017-07-16
Update to StrongNamer 0.0.6 that supports better build incrementality.

KirillOsenkov:
StrongNamer 0.0.6 has a fix that avoids rewriting a signed binary if it already exists and the Mvid matches.


KirillOsenkov:
Here's the strong namer fix: https://github.com/dsplaisted/strongnamer/pull/15

nosami:
Thanks!

PR #2780, opened 2017-07-17
Cleanup MD.Core and MD.Ide references.

KirillOsenkov:
Move most Roslyn references to MD.Ide.
Remove some unneeded references from MD.Core.
Sort references in MD.Core and MD.Ide.
Remove ProductVersion and SchemaVersion which are never useful.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
@@ -4,8 +4,6 @@
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProductVersion>8.0.30703</ProductVersion>
-    <SchemaVersion>2.0</SchemaVersion>
     <ProjectGuid>{7525BB88-6142-4A26-93B9-A30C6983390A}</ProjectGuid>
     <OutputType>Library</OutputType>
     <AssemblyName>MonoDevelop.Core</AssemblyName>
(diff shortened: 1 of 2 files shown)


KirillOsenkov:
I've verified that the actual references that end up baked into MD.Core.dll and MD.Ide.dll are the same as before. If it builds and the CI build passes it should give us enough confidence that I didn't mess anything up. I tried to be careful when transplanting references and sorting.

mkrueger:
It's ok to leave some references in core

main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
```diff
@@ -134,24 +112,41 @@
...
+    </Reference>
+    <Reference Include="Mono.Cecil.Rocks">
+      <HintPath>..\..\..\packages\Mono.Cecil.0.10.0-beta5\lib\net40\Mono.Cecil.Rocks.dll</HintPath>
+    </Reference>
+    <Reference Include="monodoc, Version=1.0.0.0, Culture=neutral, PublicKeyToken=0738eb9f132ed756" />
+    <Reference Include="Mono.Posix" />
+    <Reference Include="mscorlib" />
+    <Reference Include="Newtonsoft.Json">
+      <HintPath>..\..\..\packages\Newtonsoft.Json.9.0.1\lib\net45\Newtonsoft.Json.dll</HintPath>
+    </Reference>
```
mrward: We seem to have two different Newtonsoft.Json versions being referenced. Is this going to cause a problem? MonoDevelop.Core references 9.0.1 whilst MonoDevelop.Ide references 8.0.3. I think we should be using 8.0.3 everywhere.

KirillOsenkov: This is a good point. I'll send a separate PR to unify all Newtonsoft.Json references to 8.0.3.

PR #2781, opened 2017-07-17
Fixed 'Bug 58139 - Paste corrupts inserted text'

mkrueger:
It's just a guess what could go wrong. The new text is now formatted
before it's inserted in the text - not afterwards. I assume that there
may be a race when formatting it with the on the fly formatter
afterwards.
I couldn't reproduce the reported bug.


slluis:
@mkrueger can you please fix the conflicts?

PR #2782, opened 2017-07-17
[57111][Ide] Fix KeyBindingSet compare with parent Pt. 2

sevoku:
* fix parent selection for compare
* skip unchanged and empty bindings in schemes

(fixes bug #57111)


slluis:
Would it be possible to have unit tests for this?

sevoku:
Yeah, I'll write some

sevoku:
@slluis I've added unit tests for loading, saving and comparing `KeyBindingSet`, since its all related to the same issue. However its not a real compare (in terms of IComparable) impl, it supports only sets with the same parent or the parent itself, but it doesn't work in the other direction (`x.Equals(y) == y.Equals(x)` is not guaranteed to be `true`). So its just enough for the selection of the right sceme in the options panel, but not to compare any set with any other one. I've added some comments as a reminder.

sevoku:
@slluis what about #2783? do we need the tests also in 15.4?

PR #2783, opened 2017-07-17
[57111][Ide] Fix KeyBindingSet compare with parent Pt. 2

sevoku:
#2782 for `d15-4`

* fix parent selection for compare
* skip unchanged and empty bindings in schemes

(fixes bug #57111)


slluis:
Let's add the tests too.

PR #2784, opened 2017-07-17
Fixes BXC #53727

Jammer77:
Xamarin-Studio]Accessibility:MAS06:OSX:keyboard:Button got Expanding when tab through that.


iainx:
The bug report is that it shouldn't expand (it's confusingly written), but in my opinion, it's a bad bug report and I disagree with it being a bug.

PR #2786, opened 2017-07-17
[NuGet] Transitive assembly references not available until solution reloaded

mrward:
Fixed bug #58009 - Transitive assembly references not available until
project reloaded
https://bugzilla.xamarin.com/show_bug.cgi?id=58009

With a solution containing three .NET Standard projects: LibC
references LibB which references LibA. If the Newtonsoft.Json NuGet
package was installed into LibA the types from this NuGet packages
are not available in LibB or LibC until the solution is closed and
re-opened again. Closing and re-opening the solution refreshes the
reference information for the type system. To fix this when a NuGet
package is installed into a .NET Core project the projects that
reference this project have their reference information refreshed
by calling NotifyModified ("References") which causes the type system
to get the reference information again from MSBuild. Types from
the installed NuGet packages are then available in projects that
reference this updated project either directly or indirectly.

### DIFF
--- a/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/DotNetProjectExtensions.cs
+++ b/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/DotNetProjectExtensions.cs
@@ -254,7 +254,7 @@ namespace MonoDevelop.PackageManagement
 		public static IEnumerable<DotNetProject> GetReferencingProjects (this DotNetProject project)
 		{
 			var projects = new List<DotNetProject> ();
-			var traversedProjects = new Dictionary<string, bool> ();
+			var traversedProjects = new Dictionary<string, bool> (StringComparer.OrdinalIgnoreCase);
 			traversedProjects.Add (project.ItemId, true);
 
 			foreach (var currentProject in project.ParentSolution.GetAllDotNetProjects ()) {
--- a/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/DotNetProjectExtensions.cs
+++ b/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/DotNetProjectExtensions.cs
@@ -216,5 +216,83 @@ namespace MonoDevelop.PackageManagement
 			return File.Exists (propsFileName) &&
 				File.Exists (targetsFileName);
 		}
+
+		/// <summary>
+		/// If a NuGet package is installed into a .NET Core project then all .NET Core projects that
+		/// reference this project need to have their reference information updated. This allows the
+		/// assemblies from the NuGet package to be made available to the other projects since .NET
+		/// Core projects support transitive references. This method calls NotifyModified for each
+		/// project that references it, as well as for the project itself, passing the hint 'References'
+		/// which will cause the type system to refresh its reference information, which will be taken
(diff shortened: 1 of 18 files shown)


main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/RestoreNuGetPackagesInNuGetIntegratedProject.cs
```diff
@@ -75,12 +98,44 @@ public bool HasPackageScriptsToRun ()
...
+			var projects = project.DotNetProject.GetReferencingProjects ();
+			if (!projects.Any ())
+				return;
+
+			referencingProjects = new List<BuildIntegratedNuGetProject> ();
+			foreach (var referencingProject in projects) {
+				var projectProxy = new DotNetProjectProxy (referencingProject);
+				var currentNuGetProject = solutionManager.GetNuGetProject (projectProxy) as BuildIntegratedNuGetProject;
+				if (currentNuGetProject != null) {
+					referencingProjects.Add (currentNuGetProject);
```
Therzok: Will this happen to restore a project multiple times?

Therzok: Nevermind, duplicate check is done in GetReferencingProjects.

main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/DotNetProjectExtensions.cs
```diff
@@ -216,5 +216,97 @@ public static bool DotNetCoreNuGetMSBuildFilesExist (this DotNetProject project)
...
+			}
+		}
+
+		/// <summary>
+		/// Returns all projects that directly or indirectly referencing the specified project.
+		/// </summary>
+		public static IEnumerable<DotNetProject> GetReferencingProjects (this DotNetProject project)
+		{
+			var projects = new List<DotNetProject> ();
+			var traversedProjects = new Dictionary<string, bool> ();
```
Therzok: Don't we want a case insensitive guid check here?

main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/RestoreNuGetPackagesInNuGetIntegratedProject.cs
mrward: Only if there is a bug in the code somewhere :)

main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/DotNetProjectExtensions.cs
mrward: Yeah I can add that. .NET Core projects do not have a guid in their project file so it may not be a problem. The guid taken from the solution so it should match. However it makes sense to indicate it explicitly that case does not matter.

mrward: The dictionary is now using a case insensitive string comparer.

PR #2789, opened 2017-07-18
Fixed 'Bug 58102 - FocusCurrentDocument command doesn't work if you

mkrueger:
are currently on a pad that has auto-hide'

This command used to be just to pulse the caret posititon without
focussing. However it's now called focus document - so the purpose
changed.


slluis:
Actually, Document.Select() should already be grabbing the focus after merging https://github.com/mono/monodevelop/pull/2746. Can you please verify?

nosami:
The bug with auto-hide pads is not fixed after merging #2746 (although that does fix focus from docked pads)

This fix works for me with auto-hide pads.

slluis:
@mkrueger ping

mkrueger:
@pong - The fix works for me. I'm not sure how 'select' is defined - selection doesn't 100% mean grab focus -  it could be just a bring to front. It's the question how to define Document.Select() behavior.

PR #2790, opened 2017-07-18
[F#] Add library/exe switcher in project options. Fixes #28783

nosami:


main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpProjectCompilerOptions.fs
```diff
@@ -0,0 +1,54 @@
...
+    let boxChild (control:obj) =
+        let child = (downcast control : Box.BoxChild) 
+        child.Expand <- false
+        child.Fill <- false
+
+    let entries = 
+        [| "Executable"
+           "Library"
+           "Executable with GUI"
+           "Module" |] |> Array.map GettextCatalog.GetString
```
mrward: @Therzok - Is the above syntax supported when finding the strings to translate? I guess these are defined elsewhere in the C# addin so maybe it does not matter.

Therzok:
As in, I'd take the C# code, and make it a `DotNetCompileTargetSelector` class in `MonoDevelop.Components.Ide` and re-use that in both C# and F#.

main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpProjectCompilerOptions.fs
```diff
@@ -0,0 +1,54 @@
...
+
+    let combo = new ComboBox(entries)
+
+    do
+        Stetic.BinContainer.Attach (this) |> ignore
+        let label = new Label(GettextCatalog.GetString "Compile target:")
```
mrward: I guess we also need the accessibility adding for the new UI.

main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpProjectCompilerOptions.fs
Therzok: No, it is not. It would be pretty complicated to do with a regex scanner

nosami:
Pretty sure that not all of the options map over

Therzok:
CompileTarget seems to, and that would be the target of the common code.

nosami:
Ready for feedback

PR #2794, opened 2017-07-18

jbevain:
This is a common scenario for Unity developers, as all attributes defined in Unity's API are not suffixed with `Attribute`.


Therzok:
https://github.com/mono/monodevelop/blob/master/main/tests/MonoDevelop.CSharpBinding.Tests/Features/CodeCompletion/PortedRoslynTests/SymbolCompletionHandlerTests.cs

Attribute completion testing is done here.

jbevain:
Test added and passes locally.

PR #2795, opened 2017-07-18

jbevain:
This is a common scenario for Unity developers, as all attributes defined in Unity's API are not suffixed with `Attribute`.


slluis:
Too late for 15.3, sorry.

PR #2796, opened 2017-07-18

jbevain:


jbevain:
Waiting on #2794.

slluis:
Can you please add the unit tests here too?

jbevain:
Sure thing, here it is.

PR #2797, opened 2017-07-18
Unify all references to NewtonSoft.Json.dll to 10.0.3.

KirillOsenkov:


KirillOsenkov:
@alanmcgovern @kzu folks, were there any requirements for any MonoDevelop assemblies to be referencing a specific version of Newtonsoft.Json? I'm thinking of uniformly referencing Newtonsoft.Json 10.0.3 (10.0.0.0) and using binding redirects in MonoDevelop.exe.config to redirect all previous versions.

Would this be a problem for any MonoDevelop binary consumer?

mrward:
There are other components from md-addins that use Json.NET. Presumably we would want to update those too? Assuming they do not break VS.

alanmcgovern:
github.com/xamarin/md-addins/commit/beb5336ad875e3b8549efdcfbfac911443016378

KirillOsenkov:
As far as I can tell, it's fine to leave md-addins pointing to 8.0.3. When those components run inside MonoDevelop the binding redirect in MonoDevelop.exe.config will redirect from 8.0.0.0 to 10.0.0.0.

alanmcgovern:
Yeah - basically i'm not sure which components are being loaded inside a VS2015 process, but those components need to compile against the API surface provided by the older newtonsoft.json. As long as that's the case then this looks fine to me!

alanmcgovern:
@sgmunn Can you remember why we had to downgrade before? Was it that sometihng actually *broke* because of the newer version of the dependency, or was it just a missing binding redirect to ensure assemblies referencing the older version remapped to the newer version? I vaguely recall the new version actually broke something account/authentication related.

decriptor:
I thought it was VS 2015 related?

sgmunn:
not specifically - it was XVS related, but i couldn't tell you what version(s) of VS was involved.

PR #2801, opened 2017-07-19
Fixed 'Bug 58219 - Keyword(Access) not working in custom text color

mkrueger:
theme'

Didn't think there were themes that used the old keyword colors.
They're a direct inheritance of SharpDevelop :). However there isn't
any reason we don't support that in the textmate theme - we still
control that.


mkrueger:
Not a big bug, but a regression.

PR #2804, opened 2017-07-19
Stop the error pad resizing column widths constantly

iainx:
The error pad is constantly resizing the description column, which causes it to recalculate all the column widths,
update all the texts and strings.

This is playing havoc with the accessibility, even with only two errors shown

Instead of resizing the description column as needed, just set it to expand and give it the full width


iainx:
https://www.screencast.com/t/0aiINflBLx0 shows this happening, watch the scroll bar at the right. This is with accessibility turned off as well

iainx:
Closed because it breaks multiline

PR #2807, opened 2017-07-20
Fixed 'Bug 58255 - [VSFeedback Ticket] #438405 - Navigating XAML files

mkrueger:
with option-arrow isn't Mac-like'

It's fixed by changing the defaults.


mkrueger:
The question is - do we really need an option to select the option-left/right behavior ? 
I think we're the only one allowing that :)

mhutch:
I think we should just have a single word breaking behavior, but I'm not sure exactly what it should be. On Mac, for example, this behavior is inconsistent between Xcode, VSCode and Cocoa textareas.

Out of the behaviors we have already, the "Unix" one is most similar to Cocoa, and the "Windows" one is most similar to Xcode and VSCode. For that reason, I agree that the "Windows" one is best in the short term.

However, it's interesting to note that Roslyn customizes the word breaking behavior specifically for C#. I think this makes sense. Ideally longer-term we would do some research and tuning to get behaviors that feels natural and useful for each language.

cc @jmatthiesen for awareness

PR #2808, opened 2017-07-20
Fixed 'Bug 58218 - [C# Code Formatting Policy] Make the Mono policy

mkrueger:
follow the Mono coding guidelines and create a new separate
MonoDevelop policy'

### DIFF
--- a/main/src/addins/CSharpBinding/CSharpBinding.addin.xml
+++ b/main/src/addins/CSharpBinding/CSharpBinding.addin.xml
@@ -209,17 +209,17 @@
 		<Generator class="MonoDevelop.CSharp.Refactoring.CSharpCodeGenerator" mimeType = "text/x-csharp" />
 	</Extension>
 
+
+	<Extension path = "/MonoDevelop/ProjectModel/PolicySets/Mono" >
+		<Policies resource="MonoDevelopCSharpPolicy.xml" />
+	</Extension>
+
+	<Extension path = "/MonoDevelop/ProjectModel/PolicySets/MonoDefault" >
+		<Policies resource="MonoCSharpPolicy.xml" />
+		<Policies resource="MonoTextStylePolicy.xml" />
+	</Extension>
+
 	<Extension path = "/MonoDevelop/ProjectModel/PolicySets" >
-		<PolicySet id="Mono" _name="MonoDevelop style" allowDiffSerialize="false">
-			<Policies resource="InvariantTextStylePolicy.xml" />
-			<Policies resource="MonoDevelopCSharpPolicy.xml" />
-		</PolicySet>
-		
-		<PolicySet id="MonoDefault" _name="Mono" allowDiffSerialize="false">
-			<Policies resource="MonoCSharpPolicy.xml" />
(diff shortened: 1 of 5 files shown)


Therzok:
We need to migrate affected solutions to use MD style.

main/src/addins/CSharpBinding/CSharpBinding.addin.xml
```diff
@@ -208,19 +208,23 @@
 	<Extension path = "/MonoDevelop/TypeSystem/CodeGenerators">
 		<Generator class="MonoDevelop.CSharp.Refactoring.CSharpCodeGenerator" mimeType = "text/x-csharp" />
 	</Extension>
-	
-	<Extension path = "/MonoDevelop/ProjectModel/PolicySets/Mono" >
-		<Policies resource="MonoCSharpPolicy.xml" />
-	</Extension>
 
 	<Extension path = "/MonoDevelop/ProjectModel/PolicySets" >
+		<PolicySet id="Mono" _name="MonoDevelop style" allowDiffSerialize="false">
```
slluis: This is not correct since it will try to add a new Mono policy set. The above declaration that uses the "/MonoDevelop/ProjectModel/PolicySets/Mono" extension path was correct.

main/src/addins/CSharpBinding/CSharpBinding.addin.xml
```diff
@@ -208,19 +208,23 @@
...
+			<Policies resource="InvariantTextStylePolicy.xml" />
+			<Policies resource="MonoDevelopCSharpPolicy.xml" />
+		</PolicySet>
+		
+		<PolicySet id="MonoDefault" _name="Mono" allowDiffSerialize="false">
```
slluis: Let's declare this policy set in MD.Core and extend it here.

main/src/addins/CSharpBinding/CSharpBinding.addin.xml
mkrueger: But why did it work - I didn't change that.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.addin.xml
```diff
@@ -227,6 +227,9 @@
 		<PolicySet id="Mono" _name="Mono" allowDiffSerialize="true">
```
slluis: Let's hide this policy (`visible="false"`)

main/src/core/MonoDevelop.Core/MonoDevelop.Core.addin.xml
```diff
@@ -227,6 +227,9 @@
...
 		<!--	<Policies resource="" /> -->
 		</PolicySet>
+		<PolicySet id="MonoDefault" _name="MonoDefault" allowDiffSerialize="true">
```
slluis: Please change name to "Mono"

slluis:
So the plan is to hide the old Mono policy and rename the new one to just Mono. Projects using the old policy will keep the old settings, but in the policy selector it will show "Custom" instead of "Mono".

mkrueger:
ping

PR #2809, opened 2017-07-20
[Packaging] Make msbuild property provider thread safe

mrward:
Fixed bug #58244 - Error loading project - NuGetAuthoringPath item
already added
https://bugzilla.xamarin.com/show_bug.cgi?id=58244

Prevent the MSBuild property NuGetAuthoringPath being added twice
to the same dictionary by making the property provider thread safe.

### DIFF
--- a/main/src/addins/MonoDevelop.Packaging/MonoDevelop.Packaging/MSBuildGlobalPropertyProvider.cs
+++ b/main/src/addins/MonoDevelop.Packaging/MonoDevelop.Packaging/MSBuildGlobalPropertyProvider.cs
@@ -33,7 +33,7 @@ namespace MonoDevelop.Packaging
 {
 	class MSBuildGlobalPropertyProvider : IMSBuildGlobalPropertyProvider
 	{
-		Dictionary<string, string> properties;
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.Packaging/MonoDevelop.Packaging/MSBuildGlobalPropertyProvider.cs
```diff
@@ -34,6 +34,7 @@ namespace MonoDevelop.Packaging
 	class MSBuildGlobalPropertyProvider : IMSBuildGlobalPropertyProvider
 	{
 		Dictionary<string, string> properties;
+		readonly object propertiesLock = new object ();
```
Therzok: I think this is one of the cases where `Lazy<T>` simplifies code, as it already handles synchronization.

Therzok: I think this is one of the cases where `Lazy<T>` simplifies code, as it already handles synchronization.

mrward:
Switched this to use Lazy<T>

PR #2810, opened 2017-07-20
Fixes BXC #53775

Jammer77:
Implemented focus for EventBoxTooltip

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/EventBoxTooltip.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/EventBoxTooltip.cs
@@ -37,8 +37,11 @@ namespace MonoDevelop.Components
 	public class EventBoxTooltip : IDisposable
 	{
 		EventBox eventBox;
-		string tip;
 		TooltipPopoverWindow tooltipWindow;
+		ImageView image;
+		Pixbuf normalPixbuf;
+		Pixbuf activePixbuf;
+		string tip;
 		bool mouseOver;
 
 		public Atk.Object Accessible {
@@ -51,11 +54,6 @@ namespace MonoDevelop.Components
 		/// The EventBox should have Visible set to false otherwise the tooltip pop window
 		/// will have the wrong location.
 		/// </summary>
-
-		ImageView image;
-		Pixbuf normalPixbuf;
-		Pixbuf activePixbuf;
-
 		public EventBoxTooltip (EventBox eventBox)
 		{
 			this.eventBox = eventBox;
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components/EventBoxTooltip.cs
```diff
@@ -43,24 +45,47 @@ public class EventBoxTooltip : IDisposable
...
+
 		/// <summary>
 		/// The EventBox should have Visible set to false otherwise the tooltip pop window
 		/// will have the wrong location.
 		/// </summary>
+
+		ImageView image;
+		Pixbuf normalPixbuf;
+		Pixbuf activePixbuf;
+
```
mrward: This has moved the comment away from the constructor. I would have put these new fields with the eventBox field at the top of the class.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/EventBoxTooltip.cs
```diff
@@ -146,6 +171,39 @@ public void Dispose ()
 
 		public TaskSeverity? Severity { get; set; }
 		public PopupPosition Position { get; set; }
-	}
+	}
+#region Extension method for Pixbuf
```
mrward: Not sure we need a region here.

mrward: If this is a useful class should we extract it into a PixbufExtensions class?

iainx:
Looks good to me once Matt's comments have been fixed

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/EventBoxTooltip.cs
Therzok: I think this should be better suited for ImageService, as doing this means we recreate pixbufs on every switch. ImageService has caching for pixbufs in different variants.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/EventBoxTooltip.cs
```diff
@@ -146,6 +171,39 @@ public void Dispose ()
...
+	internal static class PixbufExtension
+	{
+		public static unsafe Pixbuf ColorShiftPixbuf (this Pixbuf src, byte shift = 120)
```
Therzok: The whole method doesn't need to be unsafe, you can use an unsafe block for the math code.

slluis:
@iainx @Jammer77 This fix may work, but it is not entirely correct. Conversion from Xwt.Image to Pixbuf should be avoided unless there is really no other option. The problem is that a Pixbuf is a bitmap, while an Xwt.Image may not be a bitmap, so when doing the conversion, some information may be lost. Most of our icons have low and high resolution variants. ImageView handles variant selection automatically, so that when the image is rendered in a retina display, the high res version is used. If the image is converted to a pixbuf and then converted back to an Xwt.Image, this capability is lost.

A solution would be to draw the focus in the image using Xwt.ImageBuilder. It would not be possible to do color shifting, but we can do something else to show the focus.

slluis:
In any case, ColorShiftPixbuf should not be public.

PR #2812, opened 2017-07-21
Fixed 'Bug 44384 - Rename refactoring on constructor should rename

mkrueger:
class'


mkrueger:
Used the roslyn GetRenamableSymbolAsync method to determine the renameable symbol.

alanmcgovern:
This broke the build so i reverted it here: https://github.com/mono/monodevelop/commit/82ebce67e220e1e96df2c6ecbb70bf0cc7b7c2a1

PR #2814, opened 2017-07-21
Information Adjustment pt_BR.po

ralmsdeveloper:
msgstr "Bancos de dados SQLite" to msgstr "Bancos de dados SQL Server" Line 109


Therzok:
Hi. Unfortunately, we cannot take this pull request as the translations for pt_BR are done by our internal localization team (and the changes are lost on import). Will open a ticket to signal this string in case it is not fixed.

PR #2829, opened 2017-07-26
D15 4 Create an mpack for Azure Functions

sgmunn:


sgmunn:
related: https://github.com/xamarin/md-addins/pull/2274

PR #2834, opened 2017-07-27
[F#] Add test for Xamarin.Forms template using a shared project

mrward:
The portable class library project is being tested already. The shared
project needs other parameters configured so it is created instead of
the portable class library project.

### DIFF
--- a/main/external/fsharpbinding/MonoDevelop.FSharp.Tests/TemplateTests.fs
+++ b/main/external/fsharpbinding/MonoDevelop.FSharp.Tests/TemplateTests.fs
@@ -54,7 +54,7 @@ type ``Template tests``() =
 
     let templatesDir = FilePath(".").FullPath.ToString() / "buildtemplates"
 
-    let test (tt:string) =
+    let testWithParameters (tt:string, parameters:string) =
         if not MonoDevelop.Core.Platform.IsMac then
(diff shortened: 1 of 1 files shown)


main/external/fsharpbinding/MonoDevelop.FSharp.Tests/TemplateTests.fs
```diff
@@ -54,7 +54,7 @@ type ``Template tests``() =
 
     let templatesDir = FilePath(".").FullPath.ToString() / "buildtemplates"
 
-    let test (tt:string) =
+    let testWithParameters (tt:string, parameters:string) =
```
nosami: You probably don't need to declare that parameters is of type `string` as it can be inferred from it's usage.

Also, you should probably use F# param syntax as opposed to C# tuple form.

so `let testWithParameters (tt:string) parameters =` and remove the parens and comma from the invocation.

Minor quibbles though - really happy to have the new test, thanks!

mrward: My F# needs some work :)

nosami: It doesn't need to be fixed to merge :)

PR #2836, opened 2017-07-28
Access C#/F# (platform selection) dropdown list box using keyboard BXC #53729

Jammer77:


mrward:
What happens if a template is selected with multiple languages and one of the categories, say General is double clicked? Just wondering since there was a bug here that was fixed in https://github.com/mono/monodevelop/commit/5f23a5444831bd0c2907c96501885963fc8036a8

It looks to me as though double clicking a category would show the pop-up which I am not sure is the expected behaviour. The IsSolutionTemplateOnActivatedRow method was added to check if the row being activated had a solution template. Just wondering if the code should be changed to be something like:

			var template = GetSolutionTemplateOnActivatedRow ((Gtk.TreeView)o, args);
                        if (template == null)
                           return;

			if (template.AvailableLanguages.Count <= 1) {
 				if (CanMoveToNextPage) {
 					MoveToNextPage ();
 				}
 			} else {
			    HandlePopup (template, 0);
 			}

iainx:
Some points

* If you double click the row, the menu pops up and then disappears quickly, this shouldn't happen.
* You need to provide more details about what you've done in your commit message. For this commit something like 

> Show language selector menu when activating the template row
>
> When activating the template row, show the language selector menu if there is more than 1 language option, otherwise activating moves to the next page
> Fixes BXC #53729

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -583,9 +583,15 @@ void FocusWidget (Widget widget)
 
 		void TreeViewRowActivated (object o, RowActivatedArgs args)
 		{
-			if (CanMoveToNextPage && IsSolutionTemplateOnActivatedRow ((Gtk.TreeView)o, args)) {
-				MoveToNextPage ();
-			}
+			var template = GetSelectedTemplate ();
```
iainx: The popup menu appears and then disappears when you double click on a row

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -145,6 +148,10 @@ void HandlePopup (SolutionTemplate template, uint eventTime)
 		[GLib.ConnectBefore]
 		void TemplatesTreeViewButtonPressed (object o, ButtonPressEventArgs args)
 		{
+			var currentClickTime = DateTime.Now;
+			doubleClicked = doubleClickPeriod > (currentClickTime - lastClickTime).TotalMilliseconds;
```
Therzok: I think it would be simpler to detect double clicking via:
`args.Event.Type == Gdk.EventType.TwoButtonPress`

Jammer77: It does'n work.

Jammer77: args.Event.Type always Gdk.EventType.ButtonPress

mrward: I guess the problem here is that double clicking the row, does not mean you are double clicking the language button. That would fire the row activated event which with the new code changes shows the language selector.

On Windows and the Mac I believe the allowed duration between clicks being registered as a double click is configurable. So I do not think we should be hard coding 500ms here.

Is there another way we can detect this through UI events or mouse events?

Jammer77: I'v tried to use SystemInformation.DoubleClickTime, but in Mono it's not implemented.

Jammer77: Therzok  show correct way, but it does not work with  TreeView

Therzok: Can you try subscribing to the treeview's parent widget? That _should_ do the trick

Jammer77: Ok. Sure. I'll try.

iainx:
Ok, this seems to work good for me now...

@slluis Are we allowed to squash these commits before committing them because this is a branch on the Mono repo?

iainx:
One thing that might be a nice addition, but maybe I'm being silly, is to call NSBeep () when pressing space doesn't do anything because there's no menu to display?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -583,9 +603,9 @@ void FocusWidget (Widget widget)
 
 		void TreeViewRowActivated (object o, RowActivatedArgs args)
 		{
-			if (CanMoveToNextPage && IsSolutionTemplateOnActivatedRow ((Gtk.TreeView)o, args)) {
+			if (CanMoveToNextPage && !isLastPressedKeySpace)
```
mrward: The original code used to check if the row double clicked had a template. I think you can double click the template category and move to the next page if the IsSolutionTemplateOnActivateRow method is not called.

https://github.com/mono/monodevelop/commit/5f23a5444831bd0c2907c96501885963fc8036a8

slluis:
Yes, it is ok to squash the commits.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -173,7 +173,7 @@ private void TemplatesTreeViewKeyPressed (object o, KeyPressEventArgs args)
 				if (template.AvailableLanguages.Count > 1)
 					HandlePopup (template, 0);
 				else
-					AppKit.AppKitFramework.NSBeep ();
+					System.Media.SystemSounds.Beep.Play ();
```
Therzok: 👍

Jammer77:
I think bug fixed!

PR #2839, opened 2017-07-31
Fixed 'Bug 58473 - [VSFeedback Ticket] #460850 - Complete Word hotkey

mkrueger:
doesn't work if "Show completion list after a character is typed" is
unchecked'

Command broke during API changes for the webtools team. I added a test
for that option.


Therzok:
Best seen in ?w=1

PR #2840, opened 2017-07-31
[d15-4][DotNetCore] Fix build error for SDK projects targeting .NET Framework

mrward:
Fixed bug #58439 - SDK Style projects targeting .NET Core 4.6.1 fails
to build even when .NET Core 2.0 preview2 is installed
https://bugzilla.xamarin.com/show_bug.cgi?id=58439

Building a SDK style project that targeted the .NET Framework would
fail to build returning an error about the .NET Core SDK not being
installed even though it was installed. The project would also have
an error icon in the Solution window with an error message about
the .NET Core SDK not being installed.

The problem was that the check added to test for the .NET Core 2.0
SDK being installed for .NET Core 2.0 projects was indicating that
the .NET Framework was not supported. This resulted in the error
about the .NET Core SDK not being installed being reported. This
check has been changed so if the project's target framework is not
.NET Core or .NET Standard then it is assumed it is supported.


manish:
Builds https://code.internalx.com/Wrench/BuildDetails?buildId=796f3eb5b36ee363f7bf4833f8c3f82

PR #2843, opened 2017-07-31
BXC #53783

Jammer77:
If the solution is correct, then I implement it on all such backends.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs
@@ -523,6 +523,19 @@ namespace MonoDevelop.Components.AtkCocoaHelper
 
 			nsa.AccessibilityLinkedUIElements = newLinkedElements;
 		}
+
+		public static void MakeAccessibilityAnnouncement (this Atk.Object o,  string message)
+		{
+			if (o == null)
+				return;
+			var nsObject = AtkCocoaMacExtensions.GetNSAccessibilityElement (o) as NSObject;
+			if (nsObject == null)
+				return;
+			var dictionary =
+				new NSDictionary (NSAccessibilityNotificationUserInfoKeys.AnnouncementKey, new NSString (message),
+								  NSAccessibilityNotificationUserInfoKeys.PriorityKey, NSAccessibilityPriorityLevel.High);
+			NSAccessibility.PostNotification (nsObject, NSAccessibilityNotifications.AnnouncementRequestedNotification, dictionary);
+		}
 	}
 
 	public class AccessibilityElementProxy : IAccessibilityElementProxy
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs
@@ -528,7 +528,7 @@ namespace MonoDevelop.Components.AtkCocoaHelper
 		{
 			if (o == null)
 				return;
-			var nsObject = AtkCocoaMacExtensions.GetNSAccessibilityElement (o) as NSObject;
+			var nsObject = GetNSAccessibilityElement (o) as NSObject;
 			if (nsObject == null)
 				return;
 			var dictionary =
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
@@ -34,8 +34,6 @@ using MonoDevelop.Components.AutoTest;
 using MonoDevelop.Components.Commands;
 using MonoDevelop.Ide.Templates;
 using MonoDevelop.Ide.Gui;
-using AppKit;
-using Foundation;
 using MonoDevelop.Core;
 using System.Threading.Tasks;
 using MonoDevelop.Ide.Projects;
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
@@ -34,6 +34,9 @@ using MonoDevelop.Components.AutoTest;
 using MonoDevelop.Components.Commands;
 using MonoDevelop.Ide.Templates;
 using MonoDevelop.Ide.Gui;
+using AppKit;
+using Foundation;
+using MonoDevelop.Core;
 
 namespace MonoDevelop.Ide.Projects
 {
@@ -59,7 +62,7 @@ namespace MonoDevelop.Ide.Projects
 			templatesTreeView.Selection.SelectFunction = TemplatesTreeViewSelection;
 			templatesTreeView.RowActivated += TreeViewRowActivated;
 			cancelButton.Clicked += CancelButtonClicked;
-			nextButton.Clicked += (sender, e) => MoveToNextPage ();
+			nextButton.Clicked += NextButtonClicked;
 			previousButton.Clicked += (sender, e) => MoveToPreviousPage ();
 
 			nextButton.CanDefault = true;
@@ -68,6 +71,37 @@ namespace MonoDevelop.Ide.Projects
 			// Setup the treeview to be able to have a context menu
 			var actionHandler = new ActionDelegate (templatesTreeView);
 			actionHandler.PerformShowMenu += PerformShowMenu;
+		}
+
+		void NextButtonClicked (object sender, EventArgs e)
+		{
+			bool isLastPage = controller.IsLastPage;
+			MoveToNextPage ();
+			if (isLastPage) {
+				var message = GenerateProjectCreatedMessage ();
+				ShowAccessibityNotification (message);
+			}
+		}
+
+		void ShowAccessibityNotification (string message)
+		{
+			var accessibleObject = this.Accessible;
+			if (accessibleObject == null)
+				return;
+			var nsObject = AtkCocoaMacExtensions.GetNSAccessibilityElement (accessibleObject) as NSObject;
+			if (nsObject == null)
+				return;
+			var dictionary =
+				new NSDictionary (NSAccessibilityNotificationUserInfoKeys.AnnouncementKey, new NSString (message),
+								  NSAccessibilityNotificationUserInfoKeys.PriorityKey, NSAccessibilityPriorityLevel.High);
+			NSAccessibility.PostNotification (nsObject, NSAccessibilityNotifications.AnnouncementRequestedNotification, dictionary);
+		}
+
+		string GenerateProjectCreatedMessage ()
+		{
+			var message = GettextCatalog.GetString ("{0} successfully created");
+			var template = controller.SelectedTemplate;
+			return String.Format (message, template.Name);
 		}
 
 		public void ShowDialog ()
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
@@ -74,7 +74,7 @@ namespace MonoDevelop.Ide.Projects
 		}
 
 		void ProjectCreationFailed (object obj, EventArgs args) => ShowProjectCreationAccessibityNotification (true);
-		void ProjectCreationSucceed (object obj, EventArgs args) => ShowProjectCreationAccessibityNotification (false);
+		void ProjectCreated(object obj, EventArgs args) => ShowProjectCreationAccessibityNotification (false);
 		async void NextButtonClicked (object sender, EventArgs e) => await MoveToNextPage ();
 
 		void ShowProjectCreationAccessibityNotification (bool hasError)
@@ -104,7 +104,7 @@ namespace MonoDevelop.Ide.Projects
 		{
 			this.controller = controller;
 			controller.ProjectCreationFailed += ProjectCreationFailed;
-			controller.ProjectCreationSucceed += ProjectCreationSucceed;
+			controller.ProjectCreated += ProjectCreated;
 			languageCellRenderer.SelectedLanguage = controller.SelectedLanguage;
 			topBannerLabel.Text = controller.BannerText;
 			LoadTemplates ();
@@ -263,7 +263,7 @@ namespace MonoDevelop.Ide.Projects
 				projectConfigurationWidget.Destroy ();
 
 			controller.ProjectCreationFailed -= ProjectCreationFailed;
-			controller.ProjectCreationSucceed -= ProjectCreationSucceed;
+			controller.ProjectCreated -= ProjectCreated;
 
 			base.Destroy ();
 		}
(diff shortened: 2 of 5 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -68,6 +71,37 @@ public GtkNewProjectDialogBackend ()
...
+				return;
+			var dictionary =
+				new NSDictionary (NSAccessibilityNotificationUserInfoKeys.AnnouncementKey, new NSString (message),
+								  NSAccessibilityNotificationUserInfoKeys.PriorityKey, NSAccessibilityPriorityLevel.High);
+			NSAccessibility.PostNotification (nsObject, NSAccessibilityNotifications.AnnouncementRequestedNotification, dictionary);
+		}
+
+		string GenerateProjectCreatedMessage ()
+		{
+			var message = GettextCatalog.GetString ("{0} successfully created");
```
mrward: How should we handle if the project creation fails? There is no real way to know if calling Create on the project controller succeeded or not. Also MoveToNextPage is an async method which is not awaited before adding this accessibility text to the button.

Therzok: Doesn't need to be a method, because this whole thing can be reduced to a simple:
`GettextCatalog.GetString ("{0} successfully created", template.Name);`

mrward: For example, if you put a space in the project name and click the Create button then the project creation would fail with an error message, leaving the dialog still open, but here we would have accessibility text that said the project template was created successfully.

Also when the create finishes successfully the dialog is closed. Not sure how this will work with voice over.

mrward: Looking at the code it looks like if after calling the controller's Create method if the dialog is closed by a call to its CloseDialog method then that means the project was created. Otherwise it failed and an error message should have been shown, unless it crashed completely.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -34,7 +34,10 @@
 using MonoDevelop.Components.Commands;
 using MonoDevelop.Ide.Templates;
 using MonoDevelop.Ide.Gui;
-
+using AppKit;
```
iainx: This file can't depend on AppKit, because it needs to build on both Linux and Windows as well. You'll need to create the ShowAccessibilityNotification method as an extension method on Atk.Object in AtkCocoaMacExtensions that does the work, and an empty method of the same name in AtkCocoaNoopExtensions

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -68,6 +71,37 @@ public GtkNewProjectDialogBackend ()
...
+		{
+			bool isLastPage = controller.IsLastPage;
+			MoveToNextPage ();
+			if (isLastPage) {
+				var message = GenerateProjectCreatedMessage ();
+				ShowAccessibityNotification (message);
+			}
+		}
+
+		void ShowAccessibityNotification (string message)
```
iainx: Accessibility is spelled incorrectly here.

This method should be in AtkCocoaMacExtensions with as
`public static void ShowAccessibilityNotification (this Atk.Object o, string message)`

and the same method signature in AtkCocoaNoopExtensions

iainx: The code will then be called as `this.Accessible.ShowAccessibilityNotification (message);`

iainx: Would the method be better named as `MakeAccessibilityAnnouncement` because we're not actually showing a notification?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs
```diff
@@ -523,6 +523,19 @@ public static void AddLinkedUIElement (this Atk.Object o, params Atk.Object [] l
 
 			nsa.AccessibilityLinkedUIElements = newLinkedElements;
 		}
+
+		public static void MakeAccessibilityAnnouncement (this Atk.Object o,  string message)
+		{
+			if (o == null)
+				return;
+			var nsObject = AtkCocoaMacExtensions.GetNSAccessibilityElement (o) as NSObject;
```
iainx: Shouldn't need the namespace here

iainx:
You've not removed the `using AppKit;` and `using Foundation;` lines from GtkNewProjectDialogBackend.cs

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -73,16 +75,18 @@ public GtkNewProjectDialogBackend ()
...
+		void ShowProjectCreationAccessibityNotification (bool hasError)
 		{
-			bool isLastPage = controller.IsLastPage;
-			MoveToNextPage ();
-			if (isLastPage) {
-				var message = GenerateProjectCreatedMessage ();
-				ShowAccessibityNotification (message);
-			}
+			var projectTemplate = controller.SelectedTemplate;
+			var messageText = GettextCatalog.GetString (hasError ? "{0} creation fail" : "{0} successfully created",
```
iainx: You can't use expressions inside `GettextCatalog.GetString`

It needs to be written the long way unfortunately, because the parser that scans the files looking for translatable strings can't handle expressions.

`var messageText = hasError ? GettextCatalog.GetString ("{0} creation fail") : GettextCatalog.GetString ("{0} successfully created");

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -122,16 +119,17 @@ public void CloseDialog ()
 		public void RegisterController (INewProjectDialogController controller)
 		{
 			this.controller = controller;
+			controller.ProjectCreationFailed += ProjectCreationFailed;
+			controller.ProjectCreationSucceed += ProjectCreationSucceed;
```
iainx: These should be removed when the dialog is closed. They might not cause a ref cycle at the moment, but better safe than sorry

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
```diff
@@ -697,6 +703,8 @@ public void MoveToPreviousPage ()
 
 			IsNewItemCreated = true;
 			UpdateDefaultSettings ();
+			var result = ProjectCreationSucceed?.BeginInvoke(this, new EventArgs (), new AsyncCallback ((_) => {} ),null);
+			result.AsyncWaitHandle.WaitOne ();
```
mrward: Out of interest - why do we need to do this? Does it not work if the ProjectCreationSucceed event is fired directly?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -88,16 +103,17 @@ public void CloseDialog ()
...
```
mrward: To be consistent the second event I think should be called ProjectCreationSucceeded. Or maybe it could just be ProjectCreated.

Naming is difficult 😄

I guess it would be nice if the Create method returned a bool to indicate success or failure however that would break the API so we cannot do that for version 7.*. Probably have to look at this later on.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -68,8 +71,20 @@ public GtkNewProjectDialogBackend ()
...
+		}
+
+		void ProjectCreationFailed (object obj, EventArgs args) => ShowProjectCreationAccessibityNotification (true);
+		void ProjectCreationSucceed (object obj, EventArgs args) => ShowProjectCreationAccessibityNotification (false);
+		async void NextButtonClicked (object sender, EventArgs e) => await MoveToNextPage ();
 
+		void ShowProjectCreationAccessibityNotification (bool hasError)
+		{
+			var projectTemplate = controller.SelectedTemplate;
+			var textTemplate= hasError ? "{0} creation fail" : "{0} successfully created";
```
Therzok: this won't get translated. Change this to:
```
string messageText;
if (hasError)
    messageText = GettextCatalog.GetString ("{0} failed to create");
else
    messageText = GettextCatalog.GetString ("{0} successfully created");
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
Jammer77: Without it dialog destroyed before VO show message.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
Therzok: `CreateFailed` and `Created` work for me.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
Therzok: Solved.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
```diff
@@ -610,8 +614,10 @@ public void MoveToPreviousPage ()
 			if (wizardProvider.HasWizard)
 				wizardProvider.BeforeProjectIsCreated ();
 
-			if (!await CreateProject ())
+			if (!await CreateProject ()) {
+				ProjectCreationFailed?.Invoke (this, new EventArgs ());
```
Therzok: `new EventArgs ()` -> `EventArgs.Empty`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
mrward: Yeah we could use CreateFailed and Created. Since you may be creating a solution + projects. So I guess that would cover everything.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
```diff
@@ -697,6 +703,8 @@ public void MoveToPreviousPage ()
 
 			IsNewItemCreated = true;
 			UpdateDefaultSettings ();
+			var result = ProjectCreationSucceed?.BeginInvoke(this, new EventArgs (), new AsyncCallback ((_) => {} ),null);
```
Therzok: `new EventArgs ()` -> `EventArgs.Empty`. The event invocation is synchronous. Why do we need the async invocation here if we wait for it right below?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
Therzok: There's also a few gotchas here:
1) We're using UI methods from an async bg call. That means we're prone to gtk errors
2) There's no EndInvoke. If any handler throws, we'll leak a threadpool thread.
3) It's sync-waited. I see no reason for this.

If what we want is to run on the UI thread, you can do this:
```
var tcs = new TaskCompletionSource<bool> ();
Gtk.Application.Invoke ((sender, args) => {
    ProjectCreationSucceed?.Invoke (this, EventArgs.Empty);
    tcs.SetResult (true);
}
await tcs.Task;
```

Jammer77: Thanks for help.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
Jammer77: Done

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
Jammer77: done

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
Jammer77: Done

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -68,8 +71,20 @@ public GtkNewProjectDialogBackend ()
...
+		}
+
+		void ProjectCreationFailed (object obj, EventArgs args) => ShowProjectCreationAccessibityNotification (true);
+		void ProjectCreated(object obj, EventArgs args) => ShowProjectCreationAccessibityNotification (false);
+		async void NextButtonClicked (object sender, EventArgs e) => await MoveToNextPage ();
 
+		void ShowProjectCreationAccessibityNotification (bool hasError)
+		{
+			var projectTemplate = controller.SelectedTemplate;
+			var textTemplate= hasError ? "creation fail" : "successfully created";
```
Therzok: Change to:
```
string messageText;
if (hasError)
    messageText = GettextCatalog.GetString ("{0} failed to create", projectTemplate.Name);
else
    messageText = GettextCatalog.GetString ("{0} successfully created", projectTemplate.Name);
```

Jammer77: Done

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
Jammer77: I did "Created"

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
Jammer77: Done!

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
Jammer77: Done!

Jammer77:
I did everything?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
```diff
@@ -697,6 +703,14 @@ public void MoveToPreviousPage ()
 
 			IsNewItemCreated = true;
 			UpdateDefaultSettings ();
+
+			var tcs = new TaskCompletionSource<bool> ();
+			Gtk.Application.Invoke ((sender, args) => {
```
slluis: You could use Runtime.RunInMainThread(), which already returns a Task.

PR #2846, opened 2017-08-01
[F#] Fixes #57061

nosami:
Remove workaround that was put in place for a previous version of the
compiler. This allows project references to work. Fixes #57061


nosami:
addprojectreference

PR #2848, opened 2017-08-01
[Mac] Make GetApplicationUrl tests resilient to any kind of system.

Therzok:


Therzok:
Has no impact on the release, just stabilizes the tests so they don't fail on bots which don't have XS.app installed.

PR #2851, opened 2017-08-01
Update metadata for an upcoming update

jmatthiesen:
Updating per a discussion with @mhutch . Thanks!

### DIFF
--- a/main/src/addins/MonoDevelop.AzureFunctions/Properties/AddinInfo.cs
+++ b/main/src/addins/MonoDevelop.AzureFunctions/Properties/AddinInfo.cs
@@ -7,8 +7,8 @@ using Mono.Addins.Description;
 	Version = MonoDevelop.BuildInfo.Version,
 	Category = "Ide")]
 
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.AzureFunctions/Properties/AddinInfo.cs
```diff
@@ -7,8 +7,8 @@
 	Version = MonoDevelop.BuildInfo.Version,
 	Category = "Ide")]
 
-[assembly: AddinName ("Azure Functions")]
-[assembly: AddinDescription ("Azure Functions")]
++[assembly: AddinName ("Azure Functions development (Preview)")]
```
Therzok: This doesn't compile. Remove the leading `+` symbols.

Therzok: This doesn't compile. Remove the leading + symbols.

PR #2854, opened 2017-08-01
[A11y] Add option to disable accessibility [d15-4]

iainx:
Can be set through the UI or through defaults
defaults write MonoDevelop com.monodevelop.AccessibilityEnabled true


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.OptionPanels/D152AccessibilityPanel.cs
```diff
@@ -0,0 +1,135 @@
...
+using MonoDevelop.Ide.Gui.Dialogs;
+using MonoDevelop.Core;
+
+using Gtk;
+
+using Foundation;
+
+namespace MonoDevelop.Ide.Gui.OptionPanels
+{
+	public class D152AccessibilityPanel : OptionsPanel
```
mrward: Does this need to be public? Not that I am expecting anyone to take a dependency on it :)

PR #2860, opened 2017-08-03
Fixed 'Bug 56288 - Spaces converted to tabs on paste with C# source

mkrueger:
code option to "Use default settings from 'Text file'"'


mkrueger:
Getting the mime type inheritance chain like:

				var types = MonoDevelop.Ide.DesktopService.GetMimeTypeInheritanceChain (MonoDevelop.CSharp.Formatting.CSharpFormatter.MimeType);
				var textPolicy = indent.DocumentContext.Project.Policies.Get<Ide.Gui.Content.TextStylePolicy> (types);

Would've fixed that as well. But just taking the text editor options as base here is faster and bugs will be more discoverable.

PR #2862, opened 2017-08-03
Fixed 'Bug 58570 - cut & paste always inserts tabs in xml'.

mkrueger:
There is no real need for using policies here - the editor options are
already set according to the policies.


main/src/addins/Xml/Editor/BaseXmlEditorExtension.cs
```diff
@@ -550,8 +550,8 @@ protected string GetLineIndent (int line)
 			//because that means there are closing tags on the line, and they de-indent the line they're on
 			var endElementDepth = GetElementIndentDepth (Tracker.Engine.Nodes);
 			var elementDepth = Math.Min (endElementDepth, startElementDepth);
-
-			//FIXME: use policies
+			if (Editor.Options.TabsToSpaces)
+				return new string (' ', elementDepth * Editor.Options.TabSize);
```
KirillOsenkov: I wonder if these strings need caching. Doesn't look like this codepath is hot, so maybe not.

PR #2864, opened 2017-08-03
Add binding redirect for System.ValueTuple as AzureFunctions is refer…

ToddGrun:
…encing a different version than others.


mhutch:
No objections to merging this as it seems like a useful change in general.

That said, Functions only references ValueTuple for its commandline func.exe tool, so I'm concerned how/why that copy is getting loaded into the MD process. Do you have repro instructions?

KirillOsenkov:
I think Todd might be bringing his own copy with WebTooling, so Azure Functions has nothing to do with it (except for being the same version as the one getting loaded). In any case, I've checked that it's safe to unify on 4.0.1.1.

KirillOsenkov:
Turns out it was trying to load 4.0.1.0 because Roslyn references it. Since we deploy 4.0.1.1 we need the redirect.

PR #2868, opened 2017-08-04
Fixed 'Bug 53583 - Using IExtendingLineMarker causes redraw issues'

mkrueger:
This was caused by some old code trying to manage the addition/removal
of line markers. Was a left over of times where line markers weren't
stored in segment trees. I removed the old handling code it was mostly
unused anyways.


nosami:
Looking forward to trying this out!

nosami:
signature-help Works great, thanks!

PR #2883, opened 2017-08-08
Add Less, Scss and TypeScript item templates

mlorbetske:
Not sure if this is fully correct yet or targeting the right branch

cc @mhutch


vancura:
Ok, ignore my comments above, I am waiting for an approval on icons and then I'll push them here.

PR #2888, opened 2017-08-09
Fixed 'Bug 58077 - [VSFeedback Ticket] #461622 - Unable to save some

mkrueger:
files after leaving VSfM open for extended period; affected file
content is corrupt'

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SourceEditorView.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SourceEditorView.cs
@@ -777,7 +777,7 @@ namespace MonoDevelop.SourceEditor
 					}
 				}
 				try {
-					this.Document.VsTextDocument.Save();
+					MonoDevelop.Core.Text.TextFileUtility.WriteText (fileName, Document);
 				} catch (InvalidEncodingException) {
 					var result = MessageService.AskQuestion (GettextCatalog.GetString ("Can't save file with current codepage."), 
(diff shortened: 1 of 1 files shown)


slluis:
Were you able to reproduce bug #58077? Do you know what's the root of the problem?

mkrueger:
https://github.com/mono/monodevelop/blob/master/main/src/core/Mono.TextEditor.Platform/Platform/Text/Impl/TextModel/TextDocument.cs#L736

That's the only location in the code where that could happen - it's inside the texd model save logic. However it shouldn't be used at all because we've a unix save logic that uses atomar file moves for saving.

By bypassing that I also bypass the defects - they silently ignore errors and may create random files like in the bug description. No I haven't reproduced the bug - silently ignoring exceptions is bad in any case. If they fear duplicate files they should check with file exists if it's really a dup - if not something else is the problem.

KirillOsenkov:
Adding @dpugh - David, the VS editor has non-trivial document save logic (create a temp file, rename, etc). Could it be that there's a bug in our copy of the Text Document Save logic? From the bug description https://bugzilla.xamarin.com/show_bug.cgi?id=58077 it is worrying. 

@mkrueger - your change will likely fix the problem, but I'm concerned that by not calling the VS text document save a lot of logic won't run that we may need. Perhaps the dirty state management? VS will perpetually think the document is dirty.

mkrueger:
we've a non trivial save logic as well - but we're using unix sysctl commands for that so using the vs editor one shouldn't be done in any case on unix/os x.

mkrueger:
The root cause may not even be in the vs saving - could be somewhere else.

PR #2889, opened 2017-08-09
Fixed 'Bug 58077 - [VSFeedback Ticket] #461622 - Unable to save some

mkrueger:
files after leaving VSfM open for extended period; affected file
content is corrupt'


slluis:
The fix needs to be verified.

slluis:
Already merged in master, won't merge it in 15.4.

PR #2891, opened 2017-08-10
Fixed 'Bug 58488 - Breakpoint highlighting not visible'

mkrueger:
+ Turned on highlighting
+ The current line marker is now displayed in the empty space left of
the line numbers and not on top of the icon marker.


slluis:
The invocation marker should be shown in that area too:
captura de pantalla 2017-08-16 a les 18 56 56

mkrueger:
This makes it in fact easier to implement :).

DavidKarlas:
image
Looks like this is not handling files with 100+ lines nicely, this looks like it's overlapping.

vancura:
Please ping me or add @hbons and/or me as reviewers next time you create a design-related PR. We need to know about changes like this one. Thanks!

PR #2892, opened 2017-08-10
LESS, SCSS, TS, JS icons

vancura:
This brings LESS, SCSS, TS, JS icons.

There's a problem when clicking http://vncr.in/lxDj, it seems I broke something but I have no idea where it could be. @Therzok also took a look and is helpless. Log: https://gist.github.com/vancura/79cda003c600143a0ac2ef697549908f


vancura:
Depends on https://bugzilla.xamarin.com/show_bug.cgi?id=58697.

vancura:
https://bugzilla.xamarin.com/show_bug.cgi?id=58697 is now fixed in master, I think it's safe to merge this now. @mhutch? What do you think?

PR #2894, opened 2017-08-11
[F#] Fix a race condition when merging ItemGroups

nosami:
Sometimes when a file is added to the project, it doesn't yet exist
in the MSBuildItems. Fix by checking for Project Files that don't yet
exist in MSBuildItems. Fixes #57689

### DIFF
--- a/main/external/fsharpbinding/MonoDevelop.FSharp.Tests/ProjectTests.fs
+++ b/main/external/fsharpbinding/MonoDevelop.FSharp.Tests/ProjectTests.fs
@@ -105,7 +105,7 @@ type ProjectTests() =
                                         File.Create(path).Dispose()
                                         project.AddFile(path, buildAction) |> ignore)
 
-                do! project.SaveAsync(monitor) |> Async.AwaitTask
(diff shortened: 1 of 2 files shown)


mrward:
The fix looks OK to me. I am just wondering if we know where the race problem exists. The code that was changed is called after the project calls base.OnWriteProject so it seems odd to me that the file is not added to the MSBuildProject at this point.

    override x.OnWriteProject(monitor, msproject) =
        base.OnWriteProject(monitor, msproject)
        fixProjectFormatForVisualStudio msproject

I have seen something similar before with .NET Core projects sometimes where when a file was deleted from an F# project it would be removed from the Files collection and the project saved before the file was deleted from disk, which would cause the project to have a RemoveItem added when it was not needed. So I guess this may be similar - file added to the Files collection, project is saved, before the file exists on disk. Then the original F# code before this pull request was applied was checking for the file to exist, which it did not, so it removed it from the MSBuildProject.

mrward:
QA has also [verified this has been fixed](https://bugzilla.xamarin.com/show_bug.cgi?id=57689#c9).

PR #2895, opened 2017-08-14
Bug 58164 - Device/Simulator list is disabled upon creation of new iOS project

DavidKarlas:
Moved selection of target device from SelectActiveConfiguration to after target devices are actually set


DavidKarlas:
Opened Trello board for QA: https://trello.com/c/zTC7RWwQ/61-vsfm-bug-58164-device-simulator-list-is-disabled-upon-creation-of-new-ios-project

PR #2903, opened 2017-08-14
[NUnit] Recognize unit-test projects when using PackageReference NuGet references instead of packages.config

garuma:
This needed to use some (currently) internal types of MonoDevelop.PackageManagement for which a temporary InternalsVisibleTo was added.

### DIFF
--- a/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/ProjectPackageReference.cs
+++ b/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/ProjectPackageReference.cs
@@ -75,11 +75,31 @@ namespace MonoDevelop.PackageManagement
 
 		NuGetFramework GetFramework ()
 		{
+			if (Project == null)
+				return NuGetFramework.AnyFramework;
 			string framework = Project.GetDotNetCoreTargetFrameworks ().FirstOrDefault ();
 			if (framework != null)
 				return NuGetFramework.Parse (framework);
 
 			return NuGetFramework.UnsupportedFramework;
+		}
+
+		public bool IsAtLeastVersion (Version version)
+		{
+			var packageReference = CreatePackageReference ();
+			var requestedVersion = new NuGetVersion (version);
+			var comparer = VersionComparer.VersionRelease;
+			if (packageReference.HasAllowedVersions) {
+				var versionRange = packageReference.AllowedVersions;
+				if (versionRange.HasLowerBound) {
+					var result = comparer.Compare (versionRange.MinVersion, requestedVersion);
+					return versionRange.IsMinInclusive ? result <= 0 : result < 0;
+				}
+			} else if (packageReference.PackageIdentity.HasVersion) {
+				var packageVersion = packageReference.PackageIdentity.Version;
+				return comparer.Compare (requestedVersion, packageVersion) <= 0;
+			}
+			return false;
 		}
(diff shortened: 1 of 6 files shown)


main/src/addins/MonoDevelop.PackageManagement/AssemblyInfo.cs
```diff
@@ -7,3 +7,8 @@
 [assembly: AssemblyDescription("NuGet addin for MonoDevelop")]
 [assembly: AssemblyVersion ("1.0")]
 [assembly: AssemblyCopyright ("MIT X11")]
+
+/* Temporary workaround to call some internal API from the NUnit addin.
+ * Should be removed once this API has become stable and is exposed publicly
+ */
+[assembly: InternalsVisibleTo (@"MonoDevelop.UnitTesting.NUnit, PublicKey=002400000c800000940000000602000000240000525341310004000001000100e1290d741888d13312c0cd1f72bb843236573c80158a286f11bb98de5ee8acc3142c9c97b472684e521ae45125d7414558f2e70ac56504f3e8fe80830da2cdb1cda8504e8d196150d05a214609234694ec0ebf4b37fc7537e09d877c3e65000f7467fa3adb6e62c82b10ada1af4a83651556c7d949959817fed97480839dd39b")]
```
mrward: There are other InternalsVisibleTo defined as elements in the .csproj file itself. However I guess it is OK here.

main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/ProjectPackageReference.cs
```diff
@@ -75,13 +75,33 @@ VersionRange GetVersion ()
...
 		}
 
+		public bool IsAtLeastVersion (Version version)
+		{
+			var packageReference = CreatePackageReference ();
+			var requestedVersion = new NuGetVersion (version);
+			var comparer = VersionComparer.VersionRelease;
+			if (packageReference.HasAllowedVersions) {
+				var versionRange = packageReference.AllowedVersions;
+				if (versionRange.HasLowerBound) {
```
mrward: Do we want to support upper bounds only? `(, 2.0.0]`. Although most of the PackageReference projects I have seen just use `1.0.0-beta-*`.

main/src/addins/MonoDevelop.PackageManagement/AssemblyInfo.cs
garuma: I hadn't seen that my bad, I will move it there.

main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/ProjectPackageReference.cs
garuma: For the purpose of that method (checking a minimum) I felt that by definition having a upper bound only means it will always be false (assuming you aren't checking for the very first version of the package). I can add though.

I hadn't seen the second syntax type, is that also handled through `AllowedVersions` or is there something else to check?

mrward: The second type using * should work with the code already written.

Yeah, good point about the upper bound since you cannot tell what version is actually being used. It could be any version up to that. So I guess the code is OK.

PR #2910, opened 2017-08-16
Fix code completion in .xaml when code completion is triggered in mid…

DavidKarlas:
…dle of element name like `<Entr`

### DIFF
--- a/main/src/addins/Xml/Editor/BaseXmlEditorExtension.cs
+++ b/main/src/addins/Xml/Editor/BaseXmlEditorExtension.cs
@@ -370,7 +370,7 @@ namespace MonoDevelop.Xml.Editor
 			}
 			
 			//attribute name completion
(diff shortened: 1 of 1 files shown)


mrward:
Can we create unit tests for this fix?

slluis:
Yes please

DavidKarlas:
There is unit test for this at https://github.com/xamarin/md-addins/blob/master/Xamarin.Forms.Addin.Tests/XamlCodeCompletion.cs#L344 but it's not enabled on wrench atm, I plan to this tomorrow...

decriptor:
Since this affects all xml files should there be a more generic test?

DavidKarlas:
@decriptor Unfortunately there is no unit tests in Xml addin for code completion since it would take a bit longer to set that up and since this case is already unit tested in Xamarin.Forms.Addin unit tests I opened bug so this is addressed in future https://bugzilla.xamarin.com/show_bug.cgi?id=58968

PR #2911, opened 2017-08-16
Fix out of date type system projections(.xaml) for files in SharedProject

DavidKarlas:
Problem was that .xaml file projections were updated only for currently selected project(iOS or Android), hence if user .cs file was using different project file was not seeing changes in .xaml(new or renamed fields), this was pretty big problem because designer removes ability to change selected project for user...

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
@@ -1010,13 +1010,12 @@ namespace MonoDevelop.Ide.Gui
 			string mimeType = editor.MimeType;
 			CancelOldParsing ();
 			var token = parseTokenSource.Token;
-			var projectsContainingFile = (adhocProject ?? Project)?.ParentSolution?.GetProjectsContainingFile (currentParseFile);
+			var currentProject = adhocProject ?? Project;
+			var projectsContainingFile = currentProject?.ParentSolution?.GetProjectsContainingFile (currentParseFile);
 			if (projectsContainingFile == null || !projectsContainingFile.Any ())
-				projectsContainingFile = new Project [] { adhocProject ?? Project };
+				projectsContainingFile = new Project [] { currentProject };
 			ThreadPool.QueueUserWorkItem (delegate {
 				foreach (var project in projectsContainingFile) {
-					if (project is SharedAssetsProject)
-						continue;
 					var projectFile = project?.GetProjectFile (currentParseFile);
 					TypeSystemService.AddSkippedFile (currentParseFile);
 					var options = new ParseOptions {
@@ -1033,6 +1032,8 @@ namespace MonoDevelop.Ide.Gui
 						TypeSystemService.ParseProjection (options, mimeType, token).ContinueWith (task => {
 							if (token.IsCancellationRequested)
 								return;
+							if (currentProject != project)
+								return;
 							Application.Invoke ((o, args) => {
 								// this may be called after the document has closed, in that case the OnDocumentParsed event shouldn't be invoked.
 								var taskResult = task.Result;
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
```diff
@@ -1010,50 +1010,55 @@ string GetCurrentParseFileName ()
...
-				if (project != null && TypeSystemService.CanParseProjections (project, mimeType, currentParseFile)) {
-					TypeSystemService.ParseProjection (options, mimeType, token).ContinueWith (task => {
-						if (token.IsCancellationRequested)
-							return;
-						Application.Invoke ((o, args) => {
-							// this may be called after the document has closed, in that case the OnDocumentParsed event shouldn't be invoked.
-							var taskResult = task.Result;
-							if (isClosed || taskResult == null || token.IsCancellationRequested)
+				foreach (var project in projectsContainingFile) {
+					if (project is SharedAssetsProject)
```
slluis: Why skip shared projects?

DavidKarlas: My reasoning was that before my change `Project` was never set to SharedAssetsProject(because it PathBar made sure either iOS or Droid projects are selected(or any other kind))...
Also not sure how much sense it makes to update parsedDocument for sharedDocument since it doesn't have references and everything else needed...

slluis: If we need to hardcode exceptions on specific project types, we are doing something wrong. Also, I'm wondering if this change will play well with the code in PathedDocumentTextEditorExtension, since that class seems to also update the workspace.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
```diff
@@ -1010,50 +1010,56 @@ string GetCurrentParseFileName ()
...
-						});
-					}, TaskContinuationOptions.OnlyOnRanToCompletion);
-				} else {
-					TypeSystemService.ParseFile (options, mimeType, token).ContinueWith (task => {
-						if (token.IsCancellationRequested)
-							return;
-						Application.Invoke ((o, args) => {
-							// this may be called after the document has closed, in that case the OnDocumentParsed event shouldn't be invoked.
-							if (isClosed || task.Result == null || token.IsCancellationRequested)
+							if (currentProject != project)
```
slluis: `project` is the `foreach` loop variable, and won't be captured in this lambda.

slluis:
@mkrueger can you please review?

slluis:
Would it be possible to write unit tests for this?

DavidKarlas:
I can probably write unit test, I will try

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
mkrueger: Shared projects are an exception in the model they're not represented in the roslyn type system so these need to be skipped somehow.

mkrueger:
Hard to tell which side effects can occur - however basically it's correct but I'm not 100% sure if it's needed. But had a side effect on rename - I don't except any issues with that.

I just have one concern: shared projects in general. We've basically a bunch of hacks to get that in - and in roslyn it's not better. There are bugs caused by the roslyn model. In general we need a good way to handle shared files in a sane way. ATM hacks are scattered accross differnt layers & addins. 
I think we should take the VS.NET/Roslyn approach when they 'fix' it properly - in the mean time we'll hack our way around and hope the sky isn't falling down.

/ approved

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
slluis: If they are not represented in the type system, then that's the check we have to do, not explicitly check for SharedProject.

PR #2914, opened 2017-08-17
[Core] Support parsing MSBuild Exists condition with unquoted property

mrward:
Fixed bug #58644 - Condition parsing issue with function calls
https://bugzilla.xamarin.com/show_bug.cgi?id=58644

.NET Core 2.0 SDK uses conditions that pass properties to the
Exists function without single quotes. These are now supported:

    <PropertyGroup Condition="Exists($(FileName))">

Previously the condition would only be evaluated correctly if it had
single quotes around the property passed to the Exists function:

    <PropertyGroup Condition="Exists('$(FileName)')">


slluis:
Can you please create a PR for d15-4?

PR #2920, opened 2017-08-21
Fix 53648

Jammer77:


main/src/addins/MacPlatform/MainToolbar/ButtonBar.cs
```diff
@@ -198,6 +211,35 @@ void UpdateButton (IButtonBarButton button, int idx)
...
+				hasFocus = value;
+				RebuildSegments ();
+			}
+		}
+		uint focusedSegment = 0; 
+		public bool IncreaseFocusIndex()
+		{
+			bool result = true;
+			focusedSegment++;
+			if (this.buttons.Count () <= focusedSegment+1) {
```
Therzok: Use `buttons.Count` instead of `buttons.Count()`.

Jammer77: I'v done it.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/ViewCommandHandlers.cs
```diff
@@ -36,6 +36,8 @@
 using MonoDevelop.Components.Commands;
 using MonoDevelop.Ide.Commands;
 using MonoDevelop.Ide.Editor;
+using MonoDevelop.Components.AtkCocoaHelper;
+using System.Linq;
```
Therzok: Why is Linq needed?

Jammer77: Solved!

PR #2930, opened 2017-08-23
Fixed 'Bug 58955 - Refactoring essentials seems to die and block rosl…

mkrueger:
…yn from working'


slluis:
What is the actual fix?

Therzok:
AST visitors have a `throw NotSupportedException()` on `DefaultVisit`, that means that every time the language evolves (i.e. local functions), it'll end up throwing on parsing the AST.

The fix is this: https://github.com/icsharpcode/RefactoringEssentials/commit/96cf74664fd5b8b305faad1664c16050afaaca0d#diff-15d0a1ed0b4c777b7bed33a097d0758eR774

mkrueger:
y exactly. However there were lots of code commented out but not in the tests - I needed to fix that otherwise the project didn't build.

PR #2931, opened 2017-08-23
[Ide] Fix Diagnostics menu not being localized.

Therzok:


Therzok:
The removed strings were commented out in RefactoringEssentials.

PR #2932, opened 2017-08-23
Bump MD version

slluis:
This bump is required if we are going to do a SR.


Aguilex:
@slluis what does it fix?

slluis:
It doesn't fix any bug, but the version bump is required in case we have to do a service release from this branch.

PR #2934, opened 2017-08-23
[SourceEditor] Fixed debug marker icon spacing.

mkrueger:


DavidKarlas:
Maybe we can add this width based on if debugging is in progress or something like that? Some screenshots would be welcome to see how much extra space is this...

mkrueger:
No - that would produce text jumping when hitting the run button and another jump when the process is terminated. With lines below 1000 LOC it's barely visible. Above 1000 LOC it's visible - 16px more.

vancura:
Please ping me or add @hbons and/or me as reviewers next time you create a design-related PR. We need to know about changes like this one. Thanks!

PR #2937, opened 2017-08-23
[Ide] Close file stream in RecentFileStorage constructor

mrward:
The file used for storage is opened with an exclusive lock in the
RecentFileStorage's constructor and the stream associated with the
file is not closed. This can prevent the SaveRecentFiles method
from being able to acquire the file lock and write to the file.

This seemed to cause the RecentTemplates_TwoTemplatesInGroupDifferentLanguage_TreatedAsDifferentRecentTemplate test on Wrench to fail sometimes.


Therzok:
Nice catch!

PR #2939, opened 2017-08-24
Bug 55129 - Closing quote missing in XAML when autocompleting

DavidKarlas:


mrward:
Changing the default to true now means this comment is out of date:

https://github.com/mono/monodevelop/blob/2cbd63a06178bfd39b8cf30a41ee226e93406089/main/src/addins/Xml/Editor/XmlEditorOptions.cs

PR #2941, opened 2017-08-24
[NuGet] Support imported package references

mrward:
Fixed bug #58842 - Support imported PackageReferences
https://bugzilla.xamarin.com/show_bug.cgi?id=58842

If a project has no PackageReference items but imports a file that
does have PackageReference items the packages were not restored
on opening its solution. This was because only the PackageReference
items were checked inside the project itself to determine if the
project used any packages. Now the evaluated items is also checked
so any imported PackageReference items are detected and a restore
will then be run on opening the solution.

This only affected non .NET Core projects. Also the restore itself
already handled imported PackageReference items when generating the
project.assets.json file.

Imported PackageReference items are not currently displayed in the
Packages folder.


mrward:
This has been verified/approved by QA.

PR #2942, opened 2017-08-24
[Ide] Support installing mpack directly from a file open operation

garuma:
This for instance allows an mpack to be installed simply by drag&dropping it on the dock icon on Mac.

Depends on https://github.com/mono/mono-addins/pull/84


garuma:
Updated commit to include required mono-addins bump now that the PR there is merged.

Mergeable whenever appropriate!

PR #2943, opened 2017-08-24
Delete a workaround that is no longer necessary.

KirillOsenkov:
MSBuild 15.3 is out and the bug has been fixed.


slluis:
Is this fix already in Mono?

KirillOsenkov:
The bug was in one of the preview releases of MSBuild 15.3. Building MonoDevelop failed if you used that version of MSBuild on Windows. They've fixed the bug quickly, and my original workaround was purely preventative. I think we got lucky and the bug never even made it to Mono in the first place.

PR #2947, opened 2017-08-25
Work around an unhandled exception in ServiceHub

bratsche:
I couldn't find a place to simply catch this exception anywhere that worked, so here's another attempt to work around this issue. I've also reported it upstream so hopefully it'll be fixed there.

This is to solve the log error:

```
ERROR [2017-08-25 09:35:06Z]: An unhandled exception has occured. Terminating Visual Studio? False
System.Net.Sockets.SocketException (0x80004005): Success
  at Microsoft.ServiceHub.Utility.SocketClient+<ConnectAsync>d__0.MoveNext () [0x00231] in <77315aa4fd5d49b8add7ad003b8a13a2>:0
```

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
@@ -546,7 +546,7 @@ namespace MonoDevelop.Ide
 		{
 			// This is a workaround for a specific issue in ServiceHub where some 'success' event triggers an exception
 			// We should remove this once it's fixed upstream.
-			if (ex?.InnerException is System.Net.Sockets.SocketException && ex.InnerException.Message == "Success")
+			if (ex?.InnerException is System.Net.Sockets.SocketException sockEx && sockEx.Message == "Success")
 				return;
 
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
```diff
@@ -544,6 +544,11 @@ static void SetupExceptionManager ()
 		
 		static void HandleException (Exception ex, bool willShutdown)
 		{
+			// This is a workaround for a specific issue in ServiceHub where some 'success' event triggers an exception
+			// We should remove this once it's fixed upstream.
+			if (ex?.InnerException is System.Net.Sockets.SocketException && ex.InnerException.Message == "Success")
```
Therzok: Would look better with c#7's:
`ex?.InnerException is System.Net.Sockets.SocketException sockEx && sockEx.Message == "Success"`

bratsche: Cool thanks. I've updated it to use that syntax now.

bratsche:
@mrward Ah, sorry.. I didn't see that PR. So I guess it's already fixed there.

PR #2951, opened 2017-08-25
Align version with 15.3.3 SR version

sgmunn:


sgmunn:
I was too slow - https://github.com/mono/monodevelop/pull/2950

PR #2954, opened 2017-08-29
[Gettext] Add option to output sorted translation files.

Therzok:
This will not sort the messages.po. But that shouldn't be an issue, as we don't import that, it's only modified when generating it via this tool.


slluis:
What's this for?

Therzok:
So when importing localization from loc team, we can keep the line diff to a minimum

Therzok:
https://github.com/mono/monodevelop/pull/2953/files has huge diff noise just because of the diff algorithm considering lines similar and not completely removed.

PR #2955, opened 2017-08-30
[Core] Re-evaluate AppliesTo condition on capability change

mrward:
If the project extension uses an AppliesTo attribute which is
initially false and then a new project capability is added and the
project is re-evaluated then the AppliesTo attribute is now
re-evaluated. This will then cause the project extension to be added
to the extension chain. If the project extension was attached to the
project and the project capability is removed then the project
extension will now be removed when the project is re-evaluated.

Also added a test for this in MonoDevelop.Core.Tests but it relies on the AzureFunctions project
extension which I do not like. Ideally this would be independent of that addin being available. Maybe instead a test addin could be created or registered somehow that defines its own project extension that uses the AppliesTo attribute.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
@@ -50,6 +50,7 @@ namespace MonoDevelop.Projects
 		bool isShared;
 		object localLock = new object ();
 		ExtensionContext extensionContext;
+		AppliesToCondition appliesToCondition;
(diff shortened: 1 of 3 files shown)


slluis:
> The best solution is to re-create the exception context...

I mean the extension context, of course.

mrward:
Changed the code to re-create the extension context. Also had to switch to using the node id when checking for loadedNodes since the new context will create new SourceExtensionNodes so the original check was failing.

Creating a new node and passing it to WorkspaceObject.RegisterCustomExtension results it in having an empty string for the node id. Not sure if this is a problem since it looks like this is only used for testing. The test that uses the CustomCapabilityNode passes since its id is an empty string and that will match one of loadedNodes.

PR #2958, opened 2017-08-30
a11y: Increase smallest font size

vancura:
Pango’s `<small>` font size is too small, resulting in `9pt` font. We can’t globally override this font size, so I replaced the tag with `<span font=“11”>`.

Requires:
- https://github.com/mono/mono-tools/pull/49
- https://github.com/mono/mono-addins/pull/86

slack-imgs com


vancura:
Ok, this seems to be bigger than I originally thought.

vancura:
Closing the issue for now. We'll fix font size problems case by case.

PR #2964, opened 2017-08-30
[NuGet] Support RestoreProjectStyle MSBuild property

mrward:
Fixed bug #58985 - Support RestoreProjectStyle MSBuild property
https://bugzilla.xamarin.com/show_bug.cgi?id=58985

If a project sets the RestoreProjectStyle property to be
PackageReference then it should be restored as though it has
PackageReferences even if it does not have any.

    <RestoreProjectStyle>PackageReference</RestoreProjectStyle>

This fixes runtime problems when a project references a .NET Standard
project that has PackageReferences since without this assemblies
from the NuGet packages are not copied to the output directory and
would have to be added to the main project.

### DIFF
--- a/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/DotNetProjectExtensions.cs
+++ b/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/DotNetProjectExtensions.cs
@@ -195,6 +195,12 @@ namespace MonoDevelop.PackageManagement
 				.Any (projectItem => StringComparer.OrdinalIgnoreCase.Equals (projectItem.Include, packageId));
 		}
 
+		public static bool HasPackageReferenceRestoreProjectStyle (this DotNetProject project)
+		{
+			string restoreStyle = project.ProjectProperties.GetValue ("RestoreProjectStyle");
+			return StringComparer.OrdinalIgnoreCase.Equals (restoreStyle, "PackageReference");
+		}
+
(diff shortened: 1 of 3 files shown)


main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/DotNetProjectExtensions.cs
```diff
@@ -195,6 +195,12 @@ public static bool HasPackageReference (this DotNetProject project, string packa
 				.Any (projectItem => StringComparer.OrdinalIgnoreCase.Equals (projectItem.Include, packageId));
 		}
 
+		public static bool HasPackageReferenceRestoreProjectStyle (this DotNetProject project)
+		{
+			string restoreStyle = project.MSBuildProject.EvaluatedProperties.GetValue ("RestoreProjectStyle");
```
slluis: You should use `project.ProjectProperties.GetValue ("RestoreProjectStyle")`

mrward: I was going to use that originally but I changed my mind and used evaluated properties instead. Would that support a property defined in an imported file? Or maybe we do not need to support that.

slluis: It would, ProjectProperties are evaluated properties.

mrward:
Changed the code to use ProjectProperties.

PR #2972, opened 2017-08-31
[59123] Don't autohide pad while its dock item is hovered

sevoku:
PR for `master`: #2971


slluis:
@Aguilex @AdrianMurphy @jmatthiesen if you are OK with this, somebody please merge

PR #2978, opened 2017-09-01
Fix 53648  

Jammer77:
VO notification for "Select All" and "Remove/Add breakpoint"

### DIFF
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -280,6 +280,9 @@ namespace MonoDevelop.MacIntegration
 			return true;
 		}
 
+
+		//BUG 53648 - that code reason of "divider" message in Voice Over
+		/*
 		static void OnCommandActivating (object sender, CommandActivationEventArgs args)
 		{
 			if (args.Source != CommandSource.Keybinding)
@@ -293,13 +296,15 @@ namespace MonoDevelop.MacIntegration
 				}
 			}
 		}
+		*/
+		
 
 		void InitApp (CommandManager commandManager)
 		{
 			if (initedApp)
 				return;
 
-			commandManager.CommandActivating += OnCommandActivating;
+			//commandManager.CommandActivating += OnCommandActivating; //that code reason of "divider" message in Voice Over
 
 			//mac-ify these command names
 			commandManager.GetCommand (EditCommands.MonodevelopPreferences).Text = GettextCatalog.GetString ("Preferences...");
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -281,8 +281,6 @@ namespace MonoDevelop.MacIntegration
 		}
 
 
-		//BUG 53648 - that code reason of "divider" message in Voice Over
-		/*
 		static void OnCommandActivating (object sender, CommandActivationEventArgs args)
 		{
 			if (args.Source != CommandSource.Keybinding)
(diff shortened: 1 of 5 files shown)


main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -280,6 +280,9 @@ public override bool SetGlobalMenu (CommandManager commandManager, string comman
 			return true;
 		}
 
+
```
iainx: This method is needed to flash the menu, so while removing it might stop the Divider message, it also removes necessary functionality.

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -293,13 +296,15 @@ static void OnCommandActivating (object sender, CommandActivationEventArgs args)
...
+		*/
+		
 
 		void InitApp (CommandManager commandManager)
 		{
 			if (initedApp)
 				return;
 
-			commandManager.CommandActivating += OnCommandActivating;
+			//commandManager.CommandActivating += OnCommandActivating; //that code reason of "divider" message in Voice Over
```
iainx: This code is needed as well.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
```diff
@@ -225,6 +225,7 @@ public DefaultWorkbench()
 			DockNotebook.NotebookChanged += NotebookPagesChanged;
 
 			Accessible.SetIsMainWindow (true);
+			Accessible.SetRole (AtkCocoa.Roles.AXButton);
```
iainx: Why are you making the root window have a button role?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
```diff
@@ -1135,6 +1136,12 @@ protected override bool OnFocusInEvent (Gdk.EventFocus evnt)
 			return base.OnFocusInEvent (evnt);
 		}
 
+		protected override void OnSetFocus (Widget focus)
```
iainx: What is this method for, it doesn't seem related to the breakpoint code

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
Jammer77: Because i have no choice:

public enum Roles
        {
            AXButton,
            AXCell,
            AXColumn,
            AXGroup,
            AXGrowArea,
            AXImage,
            AXLink,
            AXList,
            AXMenuButton,
            AXPopUpButton,
            AXRadioButton,
            AXRow,
            AXRuler,
            AXSplitGroup,
            AXSplitter,
            AXStaticText,
            AXTabGroup,
            AXTextArea
        };

iainx: You can add the missing role to the Roles enum, but why does it need a custom role set? GtkWindows should already have the AXWindow role set

Jammer77: Ok.

iainx:
I made some tweaks to the FlashMenu method so that it reads the command name so we don't need to go through and add the code to every command.

PR #2980, opened 2017-09-01
Bug 58764 - XAML Editor jumps to a new line when trying to insert property quote marks

DavidKarlas:
Problem was that when typing “” first time and deleting with backspace IDE still thought it’s in “SkipCharSession” which is keeping track of when to skip second “ if user types after writing string… Fix for this issue is inspired by commit fa40fd0ed which fixed it for C# brackets… I also added logic needed for unit tests + removed unused field from C# session handler.


decriptor:
I wonder if this'll fix our bug too: https://bugzilla.xamarin.com/show_bug.cgi?id=46531

DavidKarlas:
@decriptor yes, this is same issue, closed it as duplicate

PR #2981, opened 2017-09-01
[A11y] Reset the user defaults before restarting the app

iainx:
NSUserDefaults seems to not want to synchronise when restarting the app, so reset it which seems to make it work.

Fixes BXC #59138


Therzok:
Debug.WriteLine is NOP in release builds, lgtm!

iainx:
Ok, think this fixes it.

iainx:
QA have okayed this.

slluis:
Ready to merge.

PR #2983, opened 2017-09-04
[A11y] Make VO read the correct menu item description

iainx:
When pressing a shortcut key VoiceOver kept saying "Divider". This change makes it read the command name instead
Based on work by Dmytro Ovcharov

Fixes BXC #53648


iainx:
This is a tweaked version of https://github.com/mono/monodevelop/pull/2978#pullrequestreview-60082567

PR #2984, opened 2017-09-04
Add UseStandardResourceNames msbuild setting...

nosami:
... to generate resource identifiers the same way that C# does.
This allows having the same resource name in different folders.
Follow on commit to https://github.com/Microsoft/visualfsharp/pull/3352#issuecomment-316743690

### DIFF
--- a/main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpProject.fs
+++ b/main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpProject.fs
@@ -167,7 +167,7 @@ type FSharpProject() as self =
         let globalGroup = msproject.GetGlobalPropertyGroup()
         // Generate F# resource names the same way that C# does
         // See https://github.com/Microsoft/visualfsharp/pull/3352
-        globalGroup.SetValue ("UseStandardResourceNames", "true", "false", true)
+        globalGroup.SetValue ("UseStandardResourceNames", x.UseStandardResourceNames, "false", true)
 
         maybe {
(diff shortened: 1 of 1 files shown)


main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpProject.fs
```diff
@@ -161,9 +164,13 @@ type FSharpProject() as self =
     override x.OnWriteProject(monitor, msproject) =
         base.OnWriteProject(monitor, msproject)
         fixProjectFormatForVisualStudio msproject
-        //Fix pcl netcore and TargetFSharpCoreVersion
         let globalGroup = msproject.GetGlobalPropertyGroup()
+        // Generate F# resource names the same way that C# does
+        // See https://github.com/Microsoft/visualfsharp/pull/3352
+        globalGroup.SetValue ("UseStandardResourceNames", "true", "false", true)
```
mrward: Should this not use the UseStandardResourceNames property value? Otherwise I do not think you can set UseStandardResourceNames to false in the .fsproj and override this behaviour. It seems to set it back to true when the file is changed. Although not sure if you want to override this - unless you have some code that relies on this behaviour.

nosami: Yes, you're right - it should use the property

nosami: Good catch, thanks :)

PR #2990, opened 2017-09-05
[Roslyn] Bump to 2.3.2

Therzok:


main/src/core/MonoDevelop.Startup/app.config
```diff
@@ -108,43 +108,43 @@
 			</dependentAssembly>
 			<dependentAssembly>
 				<assemblyIdentity name="Microsoft.CodeAnalysis" publicKeyToken="31bf3856ad364e35" culture="neutral" />
-				<bindingRedirect oldVersion="0.0.0.0-2.3.0.0" newVersion="2.3.0.0" />
+				<bindingRedirect oldVersion="0.0.0.0-2.3.2.0" newVersion="2.3.2.0" />
```
KirillOsenkov: This is wrong because even though the package version is 2.3.2 the .dll version is still 2.3.0.0!

main/src/core/MonoDevelop.Startup/app.config
```diff
@@ -108,43 +108,43 @@
...
 			</dependentAssembly>
 			<dependentAssembly>
 				<assemblyIdentity name="Microsoft.CodeAnalysis.CSharp" publicKeyToken="31bf3856ad364e35" culture="neutral" />
-				<bindingRedirect oldVersion="0.0.0.0-2.3.0.0" newVersion="2.3.0.0" />
+				<bindingRedirect oldVersion="0.0.0.0-2.3.2.0" newVersion="2.3.2.0" />
```
KirillOsenkov: Same here. I'll fix.

KirillOsenkov: Lol `<!-- DO NOT TRUST CORECLR PACKAGE VERSIONS. LOOK AT THE ASSEMBLY VERSION OF THE ASSEMBLY, NOT THE PACKAGE VERSION -->` same should be valid for Roslyn and other packages :)

PR #2993, opened 2017-09-06
IconMargin accessibility implemented.

Jammer77:
Now user can see all IconMargin accessibility children, and navigate them. 
And user can select any markers by VO.
Also i did reset accessibility children method in AtkCocoaMacHelper and multiple adding Accessibility proxies .


Jammer77:
MarkerAccessible class needs Widget for Accessible.GtkParent.

AccessibilityElementProxy class instance does not work well without it.

PR #2994, opened 2017-09-06
[A11y] Only emit the container::add/remove signal on macOS

iainx:
On macOS we need to emit the add/remove signals for the accessibility system to work correctly. Normally doing it triggers a critical warning in Gtk, but the Xamarin macOS Gtk has code to work around it. Gtk on Linux/Windows do not so the code spams Criticals to the log.

Fix this by only emiting on macOS

Fixes BXC #57083


iainx:
@slluis ARGH.

iainx:
@slluis Done

PR #3015, opened 2017-09-11
[d15-4][Ide] Fix being unable to open Find in Files dialog

mrward:
Fixed bug #59373 - Unable to open Find in Files dialog if Current
Document was selected
https://bugzilla.xamarin.com/show_bug.cgi?id=59373

Accessibility was being configured on the searchentryFileMask when
it was not created. Depending on the search option selected in the
'Look in' drop down list the searchentryFileMask is not always
created.

Not sure if this meets the requirements for inclusion in 15.4. It affects searching in the current document or selection, but you could use the inline search instead here. However if one of those options were selected then the dialog cannot be re-opened until all documents are closed or the solution is closed.

It is a regression from VS Mac 7.1


mrward:
This has now been verified and approved by QA.

slluis:
@Aguilex @jmatthiesen @AdrianMurphy approved by QA. Can you please merge?

PR #3021, opened 2017-09-11
[58836][Ide] Fix breadcrumb dropdown sizing

sevoku:
Layout must be configured during initialization for initial window sizing.

(fixes bug #58836)


mkrueger:
Doesn't work for me for:

public class Foo1234567890123456789012345678901234567890123456
{ }

sevoku:
@mkrueger how exactly? Can you make a screenshot?

This is how it should look like:
grafik

grafik
grafik

So no ellipsis/wrap if `width<400`, ellipsis if `text-length>400`.

mkrueger:
Don't have another foo in the list - that's the only one.

sevoku:
Ok, it seems to be good on Linux too,
before:
grafik
with this patch:
grafik
But still not perfect:
grafik

It's related to the missing HiDPI support on Linux, making the list exceed the 450px limit, but still better than before.

PR #3028, opened 2017-09-12
Add support for batching builds

slluis:
This commit introduces some changes in the msbuild support classes to allow
building several projects in a single MSBuild session. By doing this, build
times on big solution are significantly reduced since MSBuild can resuse
cached project information. Here is a summary of changes:

All remote builder handles has been removed from MSBuildProjectService and
moved to a new RemoteBuildEngineManager class. This new class takes care
of starting and disposing builders, and it also manages build sessions.

SolutionExtension has new virtual OnBeginBuildOperation and
OnEndBuildOperation methods that are called when a build session starts
and ends. Build sessions are implicit for now. For example, when
a project with dependencies is built (which implies building several
projects) OnBeginBuildOperation is called before the first project is built,
and OnEndBuildOperation after the last project is built.

MSBuildSolutionExtension is a new solution extension that is in charge
of starting and finishing build sessions. Build sessions are specific
to MSBuild, so this is handled in an MSBuild specific extension.

RemoteProjectBuilder has been simplified. ResolveAssemblyReferences()
and ResolvePackageDependencies() have been moved to the DotNetProject
class, and are now implemented using the standard RunTarget method.
Also the class is now used through a IRemoteProjectBuilder interface,
which is disposable, so clients can just request a builder and
dispose it when done.

A new ClearCachedData() has been added to SolutionItem, and it can
be called at any time to dispose cached data related to the project.
DotNetProject now caches assembly and package references, so it
overrides OnClearCachedData() to clear them when there are changes
in the project.

Builder handling inside the Project class has been simplified.
The project doesn't hold a reference to a project builder anymore,
it gets the builder when it needs it to run a target, and releases
it when done. There is also an explicit distinction between
long operations (such as Build), and short operations (such as
resolving references). This is used by the builder manager to
optimize the use of remote builders.

Build log handling had to change a bit since when using build sessions
MSBuild only allows specifying loggers at session level, not at
project level, while MD handles log at project level. The solution
is to have a single build session logger in the builder, and
redirect the log to a specific project logger every time a new
project build starts.

Still pending:

 * [x] Proper handling of parallel build support
 * [x] Make sure that old builders (xbuild) still work


slluis:
This is now ready for review.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteBuildEngineManager.cs
```diff
@@ -0,0 +1,675 @@
...
+			}
+
+			// One builder per solution
+			string builderKey = runtime.Id + " # " + solutionFile + " # " + group + " # " + requiresMicrosoftBuild;
+
+			RemoteBuildEngine builder = null;
+
+			// Find builders which are not being shut down
+
+			var candiateBuilders = builders.GetBuilders (builderKey).Where (b => !b.IsShuttingDown && (!b.IsBusy || allowBusy));
```
Therzok: typo `candidate`

PR #3030, opened 2017-09-13
[CSharpBinding] Fixed unit tests.

mkrueger:
The VS.NET formatting way is less intrusive than we were so some tests
needed to be removed.


mkrueger:
Mostly a fix for a bugfix that caused unit test failures. That resulted that on the fly formatting was broken. The new code should only run on statement formatting not general formatting.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Formatting/OnTheFlyFormatter.cs
```diff
@@ -57,12 +57,47 @@ public static void Format (TextEditor editor, DocumentContext context, int start
 			Format (policyParent, mimeTypeChain, editor, context, startOffset, endOffset, exact, optionSet: optionSet);
 		}
 
-		public static void FormatStatmentAt (TextEditor editor, DocumentContext context, MonoDevelop.Ide.Editor.DocumentLocation location, OptionSet optionSet = null)
+		public async static void FormatStatmentAt (TextEditor editor, DocumentContext context, MonoDevelop.Ide.Editor.DocumentLocation location, OptionSet optionSet = null)
```
DavidKarlas: This async void is harmless imo, problem is if user continues typing...
and `await analysisDocument.GetSyntaxTreeAsync()` returns after user typed some new text... What will formatting do to new text user just entered? Probably mess will appear...
I think this can be solved with CancelationToken which cancels when user types new text...

PR #3034, opened 2017-09-14
Fix bug 58445 - Closing a solution with unsaved changes and selecting "Save" before closing, solution doesn't close

slluis:
This was fixed in d15-2, but for some unknown reason the fix didn't make it to master. It is a regression.


slluis:
@Aguilex @jmatthiesen @AdrianMurphy QA approved, can you please merge?

PR #3035, opened 2017-09-14
Block selection improved

KeptPanther:
Block selection now works as in Visual Studio for Windows
Virtual spaces are highlighted when using a block selection
Paste from clipboard now works correctly when using a block selection
Fixed bug with jumping caret when moving selected text


mkrueger:
ty I fixed some issues - it's an improvement :-)

new pr is on :  https://github.com/mono/monodevelop/pull/3113

mkrueger:
Ok reopening - the changes in 3113 weren't enough - there are some issues with tab indenting it doesn't look right. If I've some time I'll fix it but for now I don't have it so I reopen that one. 

The fixes from #3113 should be ok. The PR failed 1 unit test. However if possible add some block selection tests. I'm not sure if the commented out code inside the dnd code is required in some use cases.  But I haven't looked at/tested that code part too deeply.

PR #3041, opened 2017-09-14
Bug 51136 - Breakpoint Action has no effect

DavidKarlas:


slluis:
We should have unit tests for this.

DavidKarlas:
https://github.com/mono/debugger-libs/pull/154
Unit tests added inside debugger-libs repository

P.S: Unfortunately unit tests are not ran against .NET Core yet

PR #3046, opened 2017-09-14
[MacInterop] Throw AppleScriptException instead of Exception

jstedfast:
When we fail to load the OSA component, throw an AppleScriptException
instead of a general Exception to make it easier to catch.

Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=57438


jstedfast:
The MakeSimulatorGrabFocus() method already catches AppleScriptException (and TimeoutException) and just ignores it, but it was not catching Exception, so this corner-case exception was making it thru and breaking things.

I think having this code throw AppleScriptException probably makes more sense than catching Exception, so this is my proposed fix.

PR #3047, opened 2017-09-14
Fixes performance issue running msbuild targets

slluis:
The problem was that when a project is modified and then refreshed in
the remote builder, the builder would unload not only the modified
project, but all loaded projects. This issue was introduced in this
commit: mono/monodevelop@ebf6c0a, when trying to fix another bug
(23435).

This fix was slowing down project loading because during source analysis,
several projects were marked for refresh, which caused all loaded projects
to be unloaded, so projects had to be loaded again and again. It was
much worse when installing .NET Core 2.0 because for some reason,
.NET Standard projects are much slower to evaluate when it is installed,
(but that's a different issue), and Main.sln has one of such projects:
RefactoringEssentials.

The bug that the original commit was fixing was that target files imported
by an msbuild project are not reloaded if they change in disk. This patch
fixes this specific reloading problem, and then reverts the previous fix.

Fixes bug #58586 - Evaluation is slow with .netcore 2.x installed


slluis:
Created a build lane.

PR #3052, opened 2017-09-15
D15 4 dotnetcore fsharp duplicate files build error

nosami:
Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=59491


nosami:
Build lane - https://code.internalx.com/Wrench/BuildDetails?buildId=f07687698df95b1c9f2aa27cc3b539

mrward:
Do we need a test for the imported files? The test written just seems to check the slashes are handled.

nosami:
> Do we need a test for the imported files? The test written just seems to check the slashes are handled.

Maybe. What does it mean for a file to be imported? The file needs to exist and has already been loaded into a project?

mrward:
Probably just needs a file, such as a JavaScript file, in a wwwroot directory or subdirectory and a netcoreapp1.1 project. The JavaScript file would not be explicitly defined in the .fsproj but would be imported by the .NET Core SDKs that ships with Mono. Then the fsproj would need to be loaded using the ProjectService.

Would need to see if that works without running an msbuild restore to pull down the FSharp NuGet packages. Originally I was going to just create a test for this in the DotNetCore addin. Not sure if you want to add one to the F# addin.

nosami:
Ah, I understand now. It doesn't sound F# specific, so I don't think there is any need to test this within the F# test suite.

mrward:
I guess so - although the bug was in the F# addin :)

Originally I was going to write a test in the .NET Core addin and fix the import problem in the F# addin mainly because it would be quicker for me than adding a test in the F# addin tests.

mrward:
The error is not specific to .NET Core projects. If you have a project that imports another msbuild file that defines some files then that should create imported files. The [test added for the IsImported property](https://github.com/mono/monodevelop/pull/3052/commits/de299d347a13ee25bcc2414cc211619e24f31e82) uses an import that defines files with a wildcard - athough it should work with the wildcard replaced with some files in an ItemGroup.

main/external/fsharpbinding/MonoDevelop.FSharp.Tests/TemplateTests.fs
```diff
@@ -161,6 +164,28 @@ type ``Template tests``() =
         let proj = sln.Items.[0] :?> FSharpProject
         proj.IsPortableLibrary |> should equal true
 
+    [<Test;AsyncStateMachine(typeof<Task>)>]
+    member x.``Can build dotnet core MVC web app``()=
```
mrward: 👍 I guess we should change to use a .NET Core 1.1 project so the test works on Wrench without needing .NET Core 2.0 SDK installed. .NET Core 1.1 should work with just Mono installed since it ships with the .NET Core v1 sdk.

nosami: Yeah...ok... although I should probably ask #releng for an .NET Core 2.0 SDK installation. This test works fine on my machine.

mrward: Yeah it makes sense to have .NET Core 2.0 SDK being installed on Wrench. It would make it possible to do other .NET Core 2.0 specific tests in the future. 👍

nosami:
Hmm... 1.1 app fails with the following

```
   /Users/builder/Library/Caches/VisualStudio/7.0/MSBuild/27723_1/Sdks/FSharp.NET.Sdk/Sdk/Sdk.targets(21,5): warning : The 'CreateManifestResourceNames' target should be overriden in 'FSharp.NET.Sdk' package. Maybe you need to add 'FSharp.NET.Sdk' package to fsproj or run restore
   /Users/builder/Library/Caches/VisualStudio/7.0/MSBuild/27723_1/Sdks/FSharp.NET.Sdk/Sdk/Sdk.targets(27,5): warning : The 'CoreCompile' target should be overriden in 'FSharp.NET.Sdk' package. Maybe you need to add 'FSharp.NET.Sdk' package to fsproj or run restore
   /Library/Frameworks/Mono.framework/Versions/5.4.0/lib/mono/msbuild/15.0/bin/Microsoft.Common.CurrentVersion.targets(4064,5): error MSB3030: Could not copy the file "obj/Debug/netcoreapp1.1/aspnetcoremvc11.dll" because it was not found.
```

It works for me locally. I think I should just ignore this test and reinstate with a 2.0 test once releng have 2.0 installed on CI.

nosami:
This fix was verified [here](https://bugzilla.xamarin.com/show_bug.cgi?id=59491#c8)

slluis:
@Aguilex @jmatthiesen can you merge? it fixes a regression

PR #3054, opened 2017-09-18
[mac] Bump the minimum nursery size in mono

alanmcgovern:
Let's try with 8MB and see if it improves things.


alanmcgovern:
I've been running this for about a week now and haven't noticed any reduction in performance. Hopefully there is a benefit for the heavy-allocation case :)

PR #3056, opened 2017-09-18
Bug 51324 - Deadlock initializing the Threads window after starting a debug session

DavidKarlas:
What this commit does... It takes fetching for session, processes and threads needed by Thread Pad off Main thread to thread pool...

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ThreadsPad.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ThreadsPad.cs
@@ -40,6 +40,9 @@ using MonoDevelop.Ide;
 using MonoDevelop.Components.AutoTest;
 using System.ComponentModel;
 using System.Linq;
+using System.Threading;
+using System.Collections.Generic;
+using System.Threading.Tasks;
 
 namespace MonoDevelop.Debugger
 {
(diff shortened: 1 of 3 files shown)


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ThreadsPad.cs
```diff
@@ -203,26 +235,26 @@ void Update ()
 			store.Clear ();
 
 			try {
-				if (DebuggingService.GetSessions ().SelectMany (s => s.GetProcesses ()).Count () > 1) {
-					foreach (var session in DebuggingService.GetSessions ()) {
-						foreach (var process in session.GetProcesses ()) {
+				if (sessions.SelectMany (s => s.processes).Count () > 1) {
```
Therzok: This will iterate the whole processes list. Maybe do a simpler check, like Skip().Any()?

DavidKarlas: Readability is imo currently better, it says when there is more then 1 process... Skip+Any, yes it says same thing just not as clear...
Performance... This method is caller rarely so imo not a big issue...

Therzok: I'd go with a manual loop for readability

PR #3057, opened 2017-09-18
Master fix59360

mkrueger:
Unfortunately screenshots don't seem to contain the mouse cursor - it's basically the inverted os x ibeam without shadow. That's what sublime text/vs code seem to be using.

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.cs
@@ -490,8 +490,8 @@ namespace Mono.TextEditor
 		{
 			xtermCursorInverted = xtermCursor;
 			if (Platform.IsMac) {
-				var img = xtermCursor.Image.ToXwtImage();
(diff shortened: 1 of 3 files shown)


main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.cs
```diff
@@ -594,6 +633,8 @@ protected internal override void OptionsChanged ()
 
 			DisposeLayoutDict ();
 			caretX = caretY = -LineHeight;
+			Console.WriteLine("set cursor");
```
Therzok: Debug code

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/MacCursorImage.cs
```diff
@@ -0,0 +1,54 @@
...
+// THE SOFTWARE.
+using System;
+using System.IO;
+using AppKit;
+using CoreImage;
+using MonoDevelop.Core;
+
+namespace MonoDevelop.Ide.Editor
+{
+	class MacCursorImage
```
Therzok: Should be in #if MAC

PR #3058, opened 2017-09-18
Include all available MSBuild item metadata when returning evaluated items

garuma:
The previous code would only include custom metadata from the item and none of the default provided ones (such as FullPath).

Using MetadataNames and GetMetadataValue instead ensures every bit of metadata that an item normally exposes is returned to the caller.


slluis:
We need unit tests for this.

garuma:
Good point, added.

It doesn't seem like the old <4.0 API supports the equivalent thing easily so I only test in case msbuild is activated.

PR #3059, opened 2017-09-19
Xaml 15.5

DavidKarlas:
@mhutch how do you feel about changes in BaseXmlEditorExtension.cs?
@decriptor can you test this with Android extension? Mostly around : and . behaviour.


main/src/addins/Xml/Editor/BaseXmlEditorExtension.cs
```diff
@@ -418,7 +418,8 @@ public override Task<ICompletionDataList> HandleCodeCompletionAsync (CodeComplet
 			}
 
 			//element completion
-			if (currentChar == '<' && tracker.Engine.CurrentState is XmlRootState ||
+			if ((currentChar == '<' && tracker.Engine.CurrentState is XmlRootState) ||
+				((currentChar == '.' || currentChar == ':') && tracker.Engine.CurrentState is XmlNameState) ||
```
mhutch: Surely the `.` based triggering should be in the XAML subclass? "Properties" aren't a thing in XML.

mhutch: And if the trigger is a colon, then the completion should handle that explicitly, IMO.

DavidKarlas: Before I go changing this... Can we define when should code completion appear, in .xml in VS 2017 it appears on anything?
https://bugzilla.xamarin.com/show_bug.cgi?id=59202#c1

decriptor:
I *think* `.` might trip us up a bit, but I'm not sure it is really used outside of a hardcoded `dimen` in which case we don't care about the autocompletion window. I'd go ahead with this and I can add it to the list of things I handle -> https://github.com/xamarin/designer/blob/master/Xamarin.Designer.Android/Xamarin.AndroidDesigner.MonoDevelop/AndroidXmlEditorExtension.cs#L169

PR #3061, opened 2017-09-19
Fixed 'Bug 59260 - VS4M and MonoDevelop not adapting to applied fonts'

mkrueger:
The line gap is not supported by pango therefore I use the os x
routines to determine the line height. However that only works on os x
- on linux the fluent fonts don't have that gap. But there is no
single gtk editor (and even sublime text) which supports that setting
there.
Underline position metrics are now taken into account for underlines
as well.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/OSXEditor.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/OSXEditor.cs
@@ -32,14 +32,12 @@ using MonoDevelop.Core;
 
 namespace MonoDevelop.Ide.Editor
 {
-	class MacCursorImage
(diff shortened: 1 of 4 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/OSXEditor.cs
```diff
@@ -32,14 +32,12 @@
...
-	class MacCursorImage
+	class OSXEditor
 	{
 		static Xwt.Drawing.Image image;
 
-		public static Xwt.Drawing.Image Image
-		{
-			get
-			{
+		public static Xwt.Drawing.Image IBeamCursorImage {
```
Therzok: There's a usage of this which didn't get renamed?

PR #3064, opened 2017-09-20
[TextEditor] Fix zooming.

mkrueger:


Therzok:
Should be backported to 15.5

mkrueger:
Why ?

PR #3074, opened 2017-09-21
Fixed 'Bug 59627 - C#6 Interpolated String Insertion is Weird'

mkrueger:


mkrueger:
That's another feature that originates in editorfeatures. I re synced the source and now the bug is fixed.

PR #3080, opened 2017-09-24
Fix bxc #58815 build breakage

radical:
NuGet.Build.Packaging was bumped[1] to fix the build in mono-addins.
So, we pick up that change from mono-addins. And update
NuGet.Build.Packaging to v0.2.0 to be in sync.

--
1. https://github.com/mono/mono-addins/pull/88


luhenry:
@radical we can just merge master into mono-2017-08, no need to cherry-pick this single commit. I will do the bumping tomorrow.

radical:
@luhenry sure. https://github.com/mono/monodevelop/pull/3081 is the PR for master.

PR #3081, opened 2017-09-24
Bump NuGet.Build.Packaging and mono-addins.

radical:
NuGet.Build.Packaging was bumped[1] to fix the build in mono-addins.
So, we pick up that change from mono-addins. And update
NuGet.Build.Packaging to v0.2.0 to be in sync.

bxc: #58815

--
1. https://github.com/mono/mono-addins/pull/88


radical:
Should I do this for any branch other than master and 2017-08?

PR #3084, opened 2017-09-25
Fsharp status bar tooltips

nosami:
Useful if you don't use the mouse.

statusbartooltips

### DIFF
--- a/main/external/fsharpbinding/MonoDevelop.FSharp.Gui/FSharpSettingsWidget.cs
+++ b/main/external/fsharpbinding/MonoDevelop.FSharp.Gui/FSharpSettingsWidget.cs
@@ -8,13 +8,14 @@ namespace MonoDevelop.FSharp.Gui
 		{
 			this.Build ();
 		}
-		public Gtk.CheckButton AdvanceLine { get { return advanceToNextLineCheckbox; } }
-		public Gtk.CheckButton CheckCompilerUseDefault { get { return checkCompilerUseDefault; } }
+        public Gtk.CheckButton AdvanceLine => advanceToNextLineCheckbox;
+        public Gtk.CheckButton CheckCompilerUseDefault => checkCompilerUseDefault;
 
-		public Gtk.Button ButtonCompilerBrowse { get { return buttonCompilerBrowse; } }
-		public Gtk.Entry EntryCompilerPath { get { return entryCompilerPath; } }
-		public Gtk.CheckButton CheckHighlightMutables { get { return checkHighlightMutables; } }
-		public Gtk.CheckButton CheckTypeSignatures { get { return checkTypeSignatures; } }
+        public Gtk.Button ButtonCompilerBrowse => buttonCompilerBrowse;
+        public Gtk.Entry EntryCompilerPath => entryCompilerPath;
+        public Gtk.CheckButton CheckHighlightMutables => checkHighlightMutables;
+        public Gtk.CheckButton CheckTypeSignatures => checkTypeSignatures;
+        public Gtk.CheckButton CheckStatusBarTooltips => checkStatusBarTooltips;
 
-	}
+    }
 }
--- a/main/external/fsharpbinding/MonoDevelop.FSharp.Gui/gtk-gui/MonoDevelop.FSharp.Gui.FSharpSettingsWidget.cs
+++ b/main/external/fsharpbinding/MonoDevelop.FSharp.Gui/gtk-gui/MonoDevelop.FSharp.Gui.FSharpSettingsWidget.cs
@@ -14,8 +14,6 @@ namespace MonoDevelop.FSharp.Gui
 
 		private global::Gtk.HSeparator hseparator4;
 
-		private global::Gtk.HSeparator hseparator3;
-
 		private global::Gtk.HBox hbox1;
 
 		private global::Gtk.Label label2;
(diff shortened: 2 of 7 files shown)


iainx:
This might be better in the lower left of the bottom bar?

Or might be worth revisiting after #3100 gets merged

iainx:
But also, the new UI doesn't have accessibility hints.

nosami:
> This might be better in the lower left of the bottom bar?

Yeah, I think so too. No idea how to do that though :)

nosami:
footerstatusbar

Modified to use the footer status bar area.

decriptor:
That looks really cool, but would it be better if it was a bar in the bottom of the text editor (if possible). I'm thinking about the eye travel distance. It isn't horrible on a horizontal monitor, but it could be a fair bit on a vertical monitor?

nosami:
> but would it be better if it was a bar in the bottom of the text editor (if possible)

That space at the bottom is currently doing nothing right now, seems a shame to waste it.

decriptor:
true, maybe cycle cool random F# tips ;)

slluis:
There was a conscious decision of removing the status bar when XS was redesigned a few years ago. I think it would make more sense to show the status in a bar in the text editor, but I'd like to have the opinion from @vancura @hbons @migueldeicaza.

vancura:
Window status area is indeedly not the right place to show this information. We need something better. Let me think.

vancura:
If I understand this well, the information currently shown in the toolbar status area is just a quick help, right? If that's right, I'd rather show it on a keypress in a bubble on top of the TextEditor. Something that JetBrains use to inspect declaration of the method.

nosami:
@vancura - the idea is that it isn't distracting and is automatic. Every other editor/IDE has this. It's also currently an option that is disabled by default - you have to enable it.

vancura:
This is what Rider has, is this the same thing? Invoked with ⌘P.

screenflow

vancura:
By the way, I also think it's way too detached if it's in toolbar or footer. It should close to where you're looking at.

nosami:
>By the way, I also think it's way too detached if it's in toolbar or footer. It should close to where you're looking at. 
I'd like to have the same feature but in a bar beneath the editor if it's not possible to use the footer area. I've been using this feature for years in other editors and I find working in VS for Mac jarring for this reason. I don't want to use the mouse to see the tooltip.

>This is what Rider has, is this the same thing? Invoked with ⌘P.

No, not quite the same thing, that's parameter help.

Certainly possible to display a tooltip with a keybind, but I wouldn't want it to be the only way to get this information. 
Here's the same thing in `VS code`

vscode

vancura:
I got it. Well, the problem is we don't have any permanent panel we could display it on, like VSCode's footer. Toolbar status area is a wrong place to use for such a feature, but I don't think we should re-introduce footer just because of this.

@maryanne-exe @hbons Thoughts?

nosami:
I just found a screenshot from VS (Windows) showing almost the same thing

vswindows

I don't mind _where_ the information is displayed so long as it can be displayed _somewhere_ - I was happy with it in the top area :)

vancura:
Sorry to rain on your parade. We just need to come up with some better place.

maryanne-exe:
Agree that status bar isn't the right place for these hints, let's leave it for IDE related updates.
Cons of displaying in the line (either top or bottom):
- The line will constantly flash in the user's peripheral view: can be annoying;
- Distance eyes need to travel is significant as was mentioned here;

I think it's reasonable to have a hotkey to display the hint.
Pros:
- We can use the same tooltip that popups on mouse hover: consistent with the current solution and less of implementation work;
- The user may decide to enable or disable the hints.
Cons:
- Discoverability of a new hotkey. We can add a hotkey suggestion in a tooltip that shown on the mouse event (e.g. at the bottom of a tooltip) so the user may learn in the UI. Also, use the same shortcut as other IDEs use (e.g. Rider).

vancura:
@nosami: Do you think a keyboard shortcut to invoke the feature would be too much?

nosami:
@vancura - It's not difficult at all. 

I feel like it that would be only be half as useful though. I would still want a way to always see the current type information without having to press some shortcut key constantly while evaluating some unfamiliar code.

The popup tooltip also obscures the view of the code which makes things a little awkward when trying to determine how a block of code works.

F#'s strong type inference means that type information isn't visible in the actual source code, and this type of tooltip becomes far more valuable than it would be for a C# developer.

vancura:
It would be really a nice feature if we find a good spot to draw it on. Let's summarize the requirements:

- it needs to be fast
- no keyboard shortcut to show it
- it needs to draw near the caret

Am I right?

vancura:
If that's what we need, let's move this to xamarin/design where we discuss major design features and issues (sorry public people here). I am sorry that we need to postpone your branch, but let's first make sure this looks and behaves really nice, it could be a killer feature, and not only for F#.

nosami:
@vancura - yes, ideally. I'm not opposed to having a shortcut provided that there is an option to have it always visible as well.

Not too sure about it being near the caret though as that moves constantly... and anywhere near the caret would obscure other code. I'd prefer somewhere static, either in the IDE frame, or per editor frame.

vancura:
Please see the xamarin/design issue above. Also, I am marking this PR as not-to-be-merged yet.

migueldeicaza:
This should be added as well to our online help pad that we have discussed.

migueldeicaza:
The shortcut is undiscoverable, so I rather not rely on that.    For F# we need this kind of help.

PR #3085, opened 2017-09-25
Fixed 'Bug 59704 - Crash when invoking intellisense from xaml file in

mkrueger:
Forms project'

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
@@ -542,8 +542,10 @@ namespace MonoDevelop.Ide.TypeSystem
 
 			public void Dispose ()
 			{
-				foreach (var p in Projections)
-					p.Dispose ();
+				Runtime.RunInMainThread (delegate {
+					foreach (var p in Projections)
+						p.Dispose ();
+				});
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -541,9 +541,11 @@ internal class ProjectionEntry : IDisposable
...
 
 			public void Dispose ()
-			{
-				foreach (var p in Projections)
-					p.Dispose ();
+			{
+				Runtime.RunInMainThread (delegate {
+					foreach (var p in Projections)
+						p.Dispose ();
+				});
```
mrward: Does this other Dispose of the projections need to be run with the UI thread? https://github.com/mono/monodevelop/blob/68d0a9de3ab40b9fa27680ba2ed6c15b43edb9c5/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs#L233

mrward: There is another projection Dispose that looks like it does not run on the UI thread:

https://github.com/mono/monodevelop/blob/7b23234c1e8f83495a040954be8be3ce7bc0d279/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs#L500

DavidKarlas:
I'm already proposing this change in Xaml15.5 pull request, this fix has 1 side effect which is addressed here: https://github.com/mono/monodevelop/pull/3059/files#diff-8e0b88a8fac20866f890fa1208f120d1R655
So imo, either we need to accept and merge Xaml15.5 or this PR needs `if (CurrentCompletionContext != null) {` check in CompletionTextEditorExtension.cs

mkrueger:
y the change from david fixes that

PR #3087, opened 2017-09-25
Fix Roslyn versions.

KirillOsenkov:
Even though the NuGet package version is 2.3.2, the assembly version inside is still 2.3.0.0.


slluis:
We need a PR for 15.5 too, right?

Therzok:
Affects Windows only or mono with `assembly-loader=strict` which we don't use

PR #3098, opened 2017-09-26
Remove Preview moniker from builds

sgmunn:


sgmunn:
I'm sure there was a PR for this but I think perhaps it was merged too soon and reverted. Anyway, here it is again to be merged just prior to release to stable

PR #3100, opened 2017-09-27
[StatusService] Separate the status from the statusbar

iainx:
Create a StatusService which keeps track of multiple status messages and how they change. The Statusbar simply becomes
one of many possible listeners to this StatusService.


iainx:
Obsoletes PR #1172

PR #3105, opened 2017-09-28
Fixed 'Bug 51491 - No way to disable cursor blink'

mkrueger:


Therzok:
Does this hook into the native Mac one? Doesn't seem like Gtk picks up `CursorBlink` global option from Mac.

mkrueger:
I don't know - the bug is for linux. This is a question for cody. If OS X has such an option gtk should pick it up.

PR #3107, opened 2017-09-28
Fixed 'Bug 59249 - Wrong inspection behavior'

mkrueger:


Therzok:
Can we add some tests for this? Both positive and negative?

mkrueger:
the refactoring test suite is no longer nunit for some reason - I no longer maintain it so I can't add tests. Furthermore we've problems loading the refactoring essentials project something with a wrong framework. So I don't have a chance adding unit tests on monodevelop.

PR #3109, opened 2017-09-28
Work around a Gtk bug on Windows where mouse coordinates are not repo…

KirillOsenkov:
…rted correctly.

Bug https://bugzilla.xamarin.com/show_bug.cgi?id=57086 outlines the symptoms.


KirillOsenkov:
It would be nice to get this in as other people who are using windows are starting to run into the same issue. I've verified that this fixes it on Windows and shouldn't affect other platforms.

PR #3113, opened 2017-09-29
Kept panther blockselection

mkrueger:


mkrueger:
Ok seems that there are 1-2 little issues with it need to fix that first.

nosami:
Looking forward to this. I use block selection often from the Vim addin.

mkrueger:
We"ve unit tests for block selection - that's purely a visual thing.

mkrueger:
I requested some changes - it looks better when just using spaces but tab indentation completely screws it. 

I don't think it can be implemented w/o changing drawing code.

PR #3118, opened 2017-09-29
Implemented 'Bug 39274 - [FEATURE REQUEST] Procedure line separator

mkrueger:
(as per VS)'

The line separator service is a roslyn feature. When EditorFeatures is
available we can directly plug it in.


Therzok:
Do we have a trello card?

mkrueger:
https://trello.com/c/JcT9AjWo/2923-vsmac-procedure-line-separators

it's in vsts as well

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Features/LineSeparators/CSharpLineSeparatorService.cs
```diff
@@ -0,0 +1,323 @@
...
+using Microsoft.CodeAnalysis.Host.Mef;
+using Microsoft.CodeAnalysis.Shared.Extensions;
+using Microsoft.CodeAnalysis.Text;
+using MonoDevelop.Ide.Editor.Extension;
+using Roslyn.Utilities;
+
+namespace Microsoft.CodeAnalysis.Editor.CSharp.LineSeparator
+{
+	[ExportLanguageService (typeof (ILineSeparatorService), LanguageNames.CSharp), Shared]
+	internal class CSharpLineSeparatorService : ILineSeparatorService
```
Therzok: This will have to go when we merge in roslyn 2.6.0 bump

mkrueger: I knew where I got the line separator service from  - we can get rid of some things when we get editorfeatures.

mhutch:
@vancura could you take a look at the visual design for this?

hbons:
Can you post a screenshot of what this does? Can't make it out from the checkbox description. :)

vancura:
Here's the screenshot, I've just compiled the branch:

screen shot 2017-10-31 at 16 26 04

vancura:
@mkrueger: Can we have a different source code line height around the line? So the source has some gap around it?

vancura:
Minimal source is pretty ugly:

screen shot 2017-10-31 at 16 39 02

vancura:
@mkrueger: What does exactly this line separate? I thought it's between methods, but even classes are not separated. Do you have a list?

screen shot 2017-10-31 at 16 40 14

vancura:
With that gap around the line I mean something like this:

tmp

Can we have it?

mkrueger:
no gap here - we should keep the vs.net look&feel. It's a bad feature if you ask me - gaps between lines don't make it better. We should never turn it on as default - it's just for the 2% of vs.net users.

mkrueger:
Work item: https://devdiv.visualstudio.com/DevDiv/_workitems?id=509961&_a=edit

mkrueger:
@vancura : no I don't have a list. Nor did I implement what is separated. It's from the vs team and I didn't find any list. And no I can't change the list however we can always ask for upstream changes.

vancura:
Got it. I'll design something – I mean a style of line, if possible, and definitely a color – as soon as possible (I have other higher priority issues to work on right now).

What can we do about that line? Can we use a dashed line? Can we set gaps between dots or dashes?

Also, worth noting, would it be possible to have a setting in Settings (/c @hbons) to set the TextEditor line-height? Our competition usually has it, and it makes sense for typographic control freaks like me. Thoughts?

vancura:
Styling task: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/530251

PR #3119, opened 2017-09-29
Debugger source lookups

DavidKarlas:
Card:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/500134

This pull request adds support also for Portable .pdb(beside .mdb).
Adds UI familiar to VS2017 developer when source is missing instead going to Diasassembly view directly:
VSfM:
image
VS2017:
image

It also adds options to manage "Debug Sources Files" into Solution options
image

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/NoSourceView.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/NoSourceView.cs
@@ -59,13 +59,13 @@ namespace MonoDevelop.Debugger
 				headerLabel.Markup = GettextCatalog.GetString ("{0} file not found", $"<b>{fileName}</b>");
 				box.PackStart (headerLabel);
 				var actionsBox = new HBox ();
-				var buttonBrowseAndFind = new Button (GettextCatalog.GetString ($"Browse and find {fileName}"));
+				var buttonBrowseAndFind = new Button (GettextCatalog.GetString ("Browse and find {0}", fileName));
 				buttonBrowseAndFind.Clicked += OpenFindSourceFileDialog;
 				actionsBox.PackStart (buttonBrowseAndFind);
 				box.PackStart (actionsBox);
 				if (IdeApp.ProjectOperations.CurrentSelectedSolution != null) {
 					var manageLookupsLabel = new Label ();
-					manageLookupsLabel.Markup = GettextCatalog.GetString ("Manage the locations used to find source files in the {0}", $"<a href=\"clicked\">{GettextCatalog.GetString ("Solution Options")}</a>");
+					manageLookupsLabel.Markup = GettextCatalog.GetString ("Manage the locations used to find source files in the {0}", "<a href=\"clicked\">" + GettextCatalog.GetString ("Solution Options") + "</a>");
 					manageLookupsLabel.LinkClicked += (sender, e) => {
 						if (IdeApp.ProjectOperations.CurrentSelectedSolution == null)
 							return;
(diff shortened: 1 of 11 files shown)


DavidKarlas:
@vancura Can you review 1st image?
3rd image is same component as in Preferences for something else:
image

mhutch:
Looking good!

This list of search locations should probably be fullscreen. I'd also suggest adding some text like this to the first screen:

You can manage the locations used to find source files in the Solution Options.

DavidKarlas:
Fixed to use full options window and added text explaining lookup locations can be managed in solution options.

image

image

vancura:
Please simplify, nearly all of this text is useless:

screen shot 2017-10-02 at 17 33 40

Let me know when you fix this so I can re-review. Thanks!

DavidKarlas:
@vancura 
image

vancura:
Perfect! Can you please show me how it will look when there are multiple buttons side by side?

DavidKarlas:
Currently there are no other actions, not sure in which case there are in VS, hence I made dummy buttons
image

vancura:
Amazing! Thanks!

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/NoSourceView.cs
```diff
@@ -55,21 +55,25 @@ Widget CreateContent (bool disassemblyNotSupported)
...
-
-				var browseAndFindLabel = new Label ();
-				browseAndFindLabel.Markup = $"    - <a href=\"clicked\">{GettextCatalog.GetString ("Browse and find {0}...", fileName)}</a>";
-				browseAndFindLabel.LinkClicked += OpenFindSourceFileDialog;
-				headerLabel.Font = label.Font.WithScaledSize (2);
-				optionsLabel.Font = label.Font.WithScaledSize (1.5);
-				box.PackStart (browseAndFindLabel);
-				box.PackStart (new Label (GettextCatalog.GetString ("You can manage the locations used to find source files in the Solution Options.")));
+				var actionsBox = new HBox ();
+				var buttonBrowseAndFind = new Button (GettextCatalog.GetString ($"Browse and find {fileName}"));
```
Therzok: This won't get picked up by the regex parser.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/NoSourceView.cs
```diff
@@ -80,7 +84,7 @@ Widget CreateContent (bool disassemblyNotSupported)
 			}
 			if (!disassemblyNotSupported) {
 				var labelDisassembly = new Label ();
-				labelDisassembly.Markup = GettextCatalog.GetString ("You can {0} in the Disassembly window.", $"<a href=\"clicked\">{GettextCatalog.GetString ("view disassembly")}</a>");
+				labelDisassembly.Markup = GettextCatalog.GetString ("View disassembly in the {0}", $"<a href=\"clicked\">{GettextCatalog.GetString ("Disassembly Tab")}</a>");
```
Therzok: Not sure if the 2nd string will be picked up by the regex parser. Better split this into a non-interpolated string, same with manageLookupsLabel above.

decriptor:
`Browse and find` feels redundant to me. I'm not sure if just `Find File.cs` or `Locate File.cs` is better?
What about instead of `Manage the locations...`, `Manage source file locations in Solution Options` (not sure if that changes the meaning)?

Instead of `View disassembly in the Disassembly Tab` maybe just `View assembly` with that text being the link. Would it be inferred that it'll be disassembled?

DavidKarlas:
I prefer to stay closer to VS2017 text...

slluis:
@DavidKarlas please fix conflicts

PR #3123, opened 2017-09-29
Remove some translations that are triggering a crash

bratsche:
It seems to be related to using markup. Remove these for now.

https://bugzilla.xamarin.com/show_bug.cgi?id=58759


bratsche:
QA verified now

PR #3126, opened 2017-10-02
Remove some translations that are triggering a crash

bratsche:
It seems to be related to using markup. Remove these for now.

https://bugzilla.xamarin.com/show_bug.cgi?id=58759

This is https://github.com/mono/monodevelop/pull/3123 but cherry-picked for d15-5


slluis:
Can you please create a card for this in https://trello.com/b/hiTkeyIB/platform-155-release ?

slluis:
QA accepted.

bratsche:
@slluis There was already this card, but we thought it was Japanese/Chinese specific at the time https://trello.com/c/cqKjObjc/36-xs-japanese-chinese-crasher

Should there be another card or should it be included in the existing card?

bratsche:
@slluis Actually I see that QA has already been updating that card with this branch and build. So I think we can just use the existing card.

PR #3132, opened 2017-10-03
Support Xamarin.Forms 2.4

mrward:
Various fixes for [bug 59286](https://bugzilla.xamarin.com/show_bug.cgi?id=59286).

The changes made to Project.cs are enough to get Xamarin.Forms 2.4 mostly working in a .NET Standard project. However the changes may not be enough to handle other MSBuild update items added by other NuGet packages. Also I am wondering if there is a better way to handle the code that runs on the UI thread to update the existing ProjectFile's fields - maybe somehow have this run when the file is added to the project before the saving is run.

Testing this branch the following problems have been identified and are not yet fixed. Renaming seems to be problematic independent of using Xamarin.Forms 2.4.

 - Add ContentPage with xaml then change build action to None for .xaml file.
 - Add ContentPage with xaml and then remove .xaml but do not delete .xaml file
 - Add ContentPage with xaml and then remove .xaml.cs file but do not delete .xaml.cs file
 - Add ContentPage with xaml. Then both .xaml and .xaml.cs file removed from the project but not deleted. Then edit .csproj and remove associated MSBuild items
 - Add ContentPage with xaml then rename .xaml file
 - Add ContentPage with xaml then rename .xaml.cs file

More detailed test steps:

https://gist.github.com/mrward/0c6f15ac2d781d1c154b76766e87c505

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
@@ -226,10 +226,6 @@ namespace MonoDevelop.Projects.MSBuild
 					list.Insert (index++, new MSBuildImport { Sdk = sdkPath, Project = "Sdk.props" });
 					list.Add (new MSBuildImport { Sdk = sdkPath, Project = "Sdk.targets" });
 				}
-				var nugetPropsPath = $"$(BaseIntermediateOutputPath)\\{pi.Project.FileName.FileName}.nuget.g.props";
-				var nugetTargetsPath = $"$(BaseIntermediateOutputPath)\\{pi.Project.FileName.FileName}.nuget.g.targets";
-				list.Insert (index, new MSBuildImport { Project = nugetPropsPath, Condition = $"Exists('{nugetPropsPath}')" });
-				list.Add (new MSBuildImport { Project = nugetTargetsPath, Condition = $"Exists('{nugetTargetsPath}')" });
 				objects = list;
 			}
 
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
@@ -417,9 +417,20 @@ namespace MonoDevelop.Projects.MSBuild
 
 		void UpdateProperties (ProjectInfo project, MSBuildEvaluationContext context, MSBuildItem item, MSBuildItemEvaluated evaluatedItem)
 		{
+			var rootProject = project.GetRootMSBuildProject ();
+
 			foreach (var p in item.Metadata.GetProperties ()) {
 				if (string.IsNullOrEmpty (p.Condition) || SafeParseAndEvaluate (project, context, p.Condition, true)) {
-					var evaluatedProp = new MSBuildPropertyEvaluated (project.Project, p.Name, p.Value, context.EvaluateString (p.Value)) { Condition = p.Condition };
+					string evaluatedValue = context.EvaluateString (p.Value);
+					string unevaluatedValue = p.Value;
+
+					if (rootProject != item.ParentProject) {
+						// Use the same evaluated value as the property value so expanded metadata properties from a wildcard
+						// item are not saved in the project file.
+						unevaluatedValue = evaluatedValue;
+					}
+
+					var evaluatedProp = new MSBuildPropertyEvaluated (project.Project, p.Name, unevaluatedValue, evaluatedValue) { Condition = p.Condition };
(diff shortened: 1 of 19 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
```diff
@@ -362,31 +366,52 @@ void Evaluate (ProjectInfo project, MSBuildEvaluationContext context, MSBuildIte
...
+				}
 			} else
-				UpdateEvaluatedItemInAllProjects (project, item, update, trueCond, it);
+				UpdateEvaluatedItemInAllProjects (project, null, item, update, trueCond, it);
+		}
+
+		void AddUpdateToGlobInclude (ProjectInfo project, MSBuildItem item, string update)
+		{
+			do {
+				foreach (var globInclude in project.GlobIncludes.Where (g => g.Item.Name == item.Name)) {
```
Therzok: Please do not use Linq here. This lambda will end up being created in a loop (and there's an outer loop too that calls `AddupdateToGlobInclude`). `item.Name` ends up captured.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
```diff
@@ -362,31 +366,52 @@ void Evaluate (ProjectInfo project, MSBuildEvaluationContext context, MSBuildIte
...
 		{
 			do {
-				UpdateEvaluatedItem (project, item, include, trueCond, it);
+				UpdateEvaluatedItem (project, context, item, include, trueCond, it);
 				project = project.Parent;
 			} while (project != null);
 		}
 
-		static void UpdateEvaluatedItem (ProjectInfo project, MSBuildItem item, string include, bool trueCond, MSBuildItemEvaluated it)
+		void UpdateEvaluatedItem (ProjectInfo project, MSBuildEvaluationContext context, MSBuildItem item, string include, bool trueCond, MSBuildItemEvaluated it)
```
Therzok: Why was this made non-static?

mrward: Because one of the methods being called needs to use an instance method. The instance method was SafeParseAndEvaluate which I did not want to change to allow it be static.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
mrward: What should I change the code to?

Therzok: Unroll the link condition into the foreach and `continue` on `g.Item.Name != item.Name`

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
Therzok: 👍 okay, it's fine then

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
```diff
@@ -362,31 +366,52 @@ void Evaluate (ProjectInfo project, MSBuildEvaluationContext context, MSBuildIte
...
+				UpdateEvaluatedItemInAllProjects (project, null, item, update, trueCond, it);
+		}
+
+		void AddUpdateToGlobInclude (ProjectInfo project, MSBuildItem item, string update)
+		{
+			do {
+				foreach (var globInclude in project.GlobIncludes.Where (g => g.Item.Name == item.Name)) {
+					if (globInclude.Updates == null)
+						globInclude.Updates = new List<GlobInfo> ();
+					var regex = new Regex (ExcludeToRegex (update));
```
Therzok: `update` is invariant here. We could create fewer regex if we take the regex as a parameter in `AddUpdateToGlobInclude` and create it outside the foreach in `UpdateItem`

mrward:
Also the SaveNetStandardProjectWithXamarinFormsVersion24PackageReference test seems to fail on Wrench. Seems to work if .NET Core 2.0 sdk is installed locally. Without .NET Core 2.0 sdk installed locally then the test fails.

mrward:
Also seems to work if just the .NET Core 1.0 sdk is installed which seems to suggest there is a problem with the .NET Core sdk that ships with Mono 5.4.0.201 (2017-06/71277e78f6e).

mrward:
The Core tests are now passing on Wrench. The problem was that the generated nuget.g.props and nuget.g.targets files were being evaluated twice. We do not need to add these imports ourselves before evaluating the project since Mono's Microsoft.Common.props and Microsoft.Common.targets now import them.

PR #3138, opened 2017-10-04
Switch some projects to use PackageReference instead of packages.config.

KirillOsenkov:
To make it easier to update VS Editor and Roslyn references in the future I'm switching to use <PackageReference> in the projects that use Roslyn and/or Editor.

One thing to know about PackageReference is that it works in non-SDK-style projects. A project can opt-in to supporting this either by having at least one PackageReference or setting <RestoreProjectStyle>PackageReference</RestoreProjectStyle>.

If ProjectB references ProjectA and both have opted in to use PackageReference, all PackageReferences from ProjectA will be available in ProjectB transitively.

In some projects you'll notice that I'm just adding a System.ValueTuple. This opts in and inherits the PackageReferences from MD.Core.

To make it easier to update versions I'm extracting the Editor and Roslyn versions into an MSBuild property and using that. This way we can bump the version by changing a single line.


mrward:
Wrench build is failing. Assemblies are being copied to various addin directories causing the duplicate file check to fail.

KirillOsenkov:
Yup, looking.

KirillOsenkov:
I'm afraid this approach isn't going to work. We don't have enough control over `<PackageReference>` items to prohibit copying all those to output everywhere except the original reference.

I've filed https://github.com/NuGet/Home/issues/5986 to track this, but it looks like we can't use PackageReference at this moment because of the duplicate files.

KirillOsenkov:
Actually I think I found a way to salvage it.

Let's try `<CopyNuGetImplementations>false</CopyNuGetImplementations>` to turn off copying dependencies from NuGet packages that come transitively from referenced projects.

KirillOsenkov:
Rebased on top of master to resolve conflicts, create a new wrench lane to test.

KirillOsenkov:
I'm abandoning this approach because of the bug above and I've already found a different workaround.

alanmcgovern:
Let me take a quick look...

alanmcgovern:
@KirillOsenkov We should swap over to the 'standard' mechanism to avoid local copying :) It now builds for me though!

Therzok:
Can this be closed?

PR #3142, opened 2017-10-04
[Debugger] Handle special message "Next statement can't be in another…

DavidKarlas:
… function." and make it translatable

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebugCommands.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebugCommands.cs
@@ -673,7 +673,15 @@ namespace MonoDevelop.Debugger
 				DebuggingService.SetNextStatement (doc.FileName, doc.Editor.CaretLine, doc.Editor.CaretColumn);
 			} catch (Exception e) {
 				if (e is NotSupportedException || e.InnerException is NotSupportedException) {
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebugCommands.cs
```diff
@@ -673,7 +673,15 @@ protected override void Run ()
 				DebuggingService.SetNextStatement (doc.FileName, doc.Editor.CaretLine, doc.Editor.CaretColumn);
 			} catch (Exception e) {
 				if (e is NotSupportedException || e.InnerException is NotSupportedException) {
-					MessageService.ShowError ("Unable to set the next statement to this location.");
+					string message;
+					if (e is NotSupportedException)
+						message = e.Message;
+					else
+						message = e.InnerException.Message;
+					if (message == "Unable to set the next statement. The next statement cannot be set to another function.")
```
mrward: Will this exception text always be in English?

DavidKarlas: yes, because Debugger-libs don't handle localisation https://github.com/mono/debugger-libs/pull/159

mrward: Just wondering 👍

PR #3143, opened 2017-10-04
[AutoTest] Add Support for NSTableView

manish:


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest.Results/NSObjectResult.cs
```diff
@@ -181,16 +189,45 @@ public override List<AppResult> NextSiblings ()
...
 		{
 			return GetProperties (ResultObject);
 		}
 
 		public override bool Select ()
 		{
+			if (ResultObject is NSTableView) {
+				var control = (NSTableView)ResultObject;
+				control.SelectRow (index, true);
+				control.PerformClick (0, index);
```
Therzok: `return true` here?

manish: Oh crap. Nice catch

PR #3146, opened 2017-10-05
[Ide][59844] Fix unicode markup parsing

sevoku:
Pango parses a markup without respecting multibyte
characters (byte based), hence the attribute indexes
need to be converted, in order to match the encoded
string representation.

(fixes bug #59844)

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PangoUtil.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PangoUtil.cs
@@ -257,4 +257,52 @@ namespace MonoDevelop.Components
 			});
 		}
 	}
+
+	internal class TextIndexer
+	{
+		int [] indexToByteIndex;
+		int [] byteIndexToIndex;
+
+		public TextIndexer (string text)
+		{
+			SetupTables (text);
+		}
+
+		public int IndexToByteIndex (int i)
+		{
+			if (i >= indexToByteIndex.Length)
+				// if the index exceeds the byte index range, return the last byte index + 1
+				// telling pango to span the attribute to the end of the string
+				// this happens if the string contains multibyte characters
+				return indexToByteIndex [i - 1] + 1;
+			return indexToByteIndex [i];
+		}
+
+		public int ByteIndexToIndex (int i)
+		{
+			return byteIndexToIndex [i];
+		}
+
+		public void SetupTables (string text)
+		{
+			if (text == null) {
+				this.indexToByteIndex = new int [0];
+				this.byteIndexToIndex = new int [0];
+				return;
+			}
+
+			var arr = text.ToCharArray ();
+			int byteIndex = 0;
+			int [] indexToByteIndex = new int [arr.Length];
+			var byteIndexToIndex = new System.Collections.Generic.List<int> ();
+			for (int i = 0; i < arr.Length; i++) {
+				indexToByteIndex [i] = byteIndex;
+				byteIndex += System.Text.Encoding.UTF8.GetByteCount (arr, i, 1);
+				while (byteIndexToIndex.Count < byteIndex)
+					byteIndexToIndex.Add (i);
+			}
+			this.indexToByteIndex = indexToByteIndex;
+			this.byteIndexToIndex = byteIndexToIndex.ToArray ();
+		}
+	}
 }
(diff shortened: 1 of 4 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PangoUtil.cs
```diff
@@ -257,4 +257,52 @@ void Destroy ()
...
+
+		public int ByteIndexToIndex (int i)
+		{
+			return byteIndexToIndex [i];
+		}
+
+		public void SetupTables (string text)
+		{
+			if (text == null) {
+				this.indexToByteIndex = new int [0];
```
Therzok: `Array.Empty<int> ()`

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PangoUtil.cs
```diff
@@ -257,4 +257,52 @@ void Destroy ()
...
+				this.byteIndexToIndex = new int [0];
+				return;
+			}
+
+			var arr = text.ToCharArray ();
```
Therzok: We don't need a char array here. `text.Length` will give us `arr.Length`. and [`GetByteCount(char *, int)`](https://msdn.microsoft.com/en-us/library/c7f50zt7(v=vs.110).aspx) will replicate the behaviour for `GetByteCount(char[], int, int)` without allocations

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PangoUtil.cs
```diff
@@ -257,4 +257,52 @@ void Destroy ()
...
+			int byteIndex = 0;
+			int [] indexToByteIndex = new int [arr.Length];
+			var byteIndexToIndex = new System.Collections.Generic.List<int> ();
```
Therzok: pre-allocate with `arr.Length`

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PangoUtil.cs
```diff
@@ -257,4 +257,52 @@ void Destroy ()
...
+			for (int i = 0; i < arr.Length; i++) {
+				indexToByteIndex [i] = byteIndex;
+				byteIndex += System.Text.Encoding.UTF8.GetByteCount (arr, i, 1);
+				while (byteIndexToIndex.Count < byteIndex)
+					byteIndexToIndex.Add (i);
+			}
+			this.indexToByteIndex = indexToByteIndex;
+			this.byteIndexToIndex = byteIndexToIndex.ToArray ();
```
Therzok: Let's keep it a list, and not an array?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PangoUtil.cs
Therzok: as in, the could would look like:
```
			int byteIndex = 0;
			int [] indexToByteIndex = new int [text.Length];
			var byteIndexToIndex = new System.Collections.Generic.List<int> ();
			fixed (char *p = text) {
					for (int i = 0; i < text.Length; i++) {
						indexToByteIndex [i] = byteIndex;
						byteIndex += System.Text.Encoding.UTF8.GetByteCount (p + i, 1);
						while (byteIndexToIndex.Count < byteIndex)
							byteIndexToIndex.Add (i);
					}
			}
```

Therzok:
Optimizations above should be applied to the original Xwt code too.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PangoUtil.cs
```diff
@@ -257,4 +257,56 @@ void Destroy ()
...
+		static System.Collections.Generic.List<int> emptyList = new System.Collections.Generic.List<int> ();
+		int [] indexToByteIndex;
+		System.Collections.Generic.List<int> byteIndexToIndex;
+
+		public TextIndexer (string text)
+		{
+			SetupTables (text);
+		}
+
+		public int IndexToByteIndex (int i)
```
Therzok: Maybe we should get rid of `IndexToByteIndex` and replace it with a GetCharCount.

PR #3147, opened 2017-10-05
[Ide][59844] Fix unicode markup parsing

sevoku:
(#3146 for `d15-5`)

Pango parses a markup without respecting multibyte
characters (byte based), hence the attribute indexes
need to be converted, in order to match the encoded
string representation.

(fixes bug #59844)

(cherry picked from commit bd8e2072c9a299451b55af315f8bf033a080cfd3)

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PangoUtil.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PangoUtil.cs
@@ -257,4 +257,52 @@ namespace MonoDevelop.Components
 			});
 		}
 	}
+
+	internal class TextIndexer
+	{
+		int [] indexToByteIndex;
+		int [] byteIndexToIndex;
+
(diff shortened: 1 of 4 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PangoUtil.cs
```diff
@@ -257,4 +257,52 @@ void Destroy ()
...
+
+		public int ByteIndexToIndex (int i)
+		{
+			return byteIndexToIndex [i];
+		}
+
+		public void SetupTables (string text)
+		{
+			if (text == null) {
+				this.indexToByteIndex = new int [0];
```
Therzok: `Array.Empty<int> ()`

Therzok: Moving reviews to master version

PR #3158, opened 2017-10-06
Fix warnings

nosami:


main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpPathExtension.fs
```diff
@@ -222,9 +222,9 @@ type FSharpPathExtension() as x =
       
             if ownerProjects.Count > 1 then
                 let p = x.DocumentContext.Project
-                newPath.Add (PathEntry(icon = ImageService.GetIcon(p.StockIcon.Name, Gtk.IconSize.Menu),
-                                      markup = GLib.Markup.EscapeText (p.Name),
-                                      Tag=p))
+                newPath.Add (new PathEntry(icon = ImageService.GetIcon(p.StockIcon.Name, Gtk.IconSize.Menu),
```
Therzok: PathEntry is now IDisposable btw. That helps clear some leaks

nosami: That's actually the reason for the warning that I was seeing. Need to add `new` on IDisposables. Am I supposed to dispose them?

Therzok: Oh, I missed the ones in CSharp implementation of PathedDocumentExtension. As long as PathEntry.Accessible is not queried, you should be safe with not disposing it.

PR #3167, opened 2017-10-09
[Xml] Always show code completion when typing(not just Cmd+Space or start)

DavidKarlas:
To match VS2017 behaviour

This is basically same as https://github.com/mono/monodevelop/commit/fb7a266e308c670011611e12916e1d0d7219b63e but to also include attribute value(didn't realise at time that VS2017 also does this for attribute values)


slluis:
Maybe we should honor the "Show completion list after a character is typed" option?

DavidKarlas:
This PR is replaced by https://github.com/mono/monodevelop/pull/3169 which adds setting

DavidKarlas:
Actually we will go without setting and keep Xml code completion always agressive

PR #3169, opened 2017-10-09
[Xml] Added "Automatically show code completion" to Behavior->Xml settings

DavidKarlas:
And added support for it to all 3 Xml cases(Element, Attribute and AttributeValue)

### DIFF
--- a/main/src/addins/Xml/Editor/XmlEditorOptionsPanelWidget.cs
+++ b/main/src/addins/Xml/Editor/XmlEditorOptionsPanelWidget.cs
@@ -34,7 +34,7 @@ namespace MonoDevelop.Xml.Editor
 {
 	class XmlEditorOptionsPanelWidget : VBox
 	{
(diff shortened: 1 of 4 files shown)


slluis:
There is already an option for this, “Show completion list after a character is typed” in the Intellisense panel. No need to add a new one.

main/src/addins/Xml/Editor/XmlEditorOptionsPanelWidget.cs
```diff
@@ -48,27 +48,36 @@ public XmlEditorOptionsPanelWidget ()
 			                                                          GettextCatalog.GetString ("Check to enable automatic punctuation insertion"));
 			showSchemaAnnotationCheck = new CheckButton (GettextCatalog.GetString ("Show schema annotation"));
```
Therzok: This widget is not parented, so it will be leaked. Can we add a `OnDestroyed` that forces this check to be destroyed?

DavidKarlas:
Going back to https://github.com/mono/monodevelop/pull/3167

PR #3174, opened 2017-10-10
Fixed 'Bug 60073 - Error squiggles do not appear for standalone files'

mkrueger:
The error handling was moved to the parsed document logic - this never
belonged to the core. However a small API change was needed but the
old API is intact. However the old API didn't break but the obsolete
method should be removed in future versions.
The caching in c# parsed document isn't needed roslyn caches already.


slluis:
Maybe it would be better to provide an `IsAdhocProject` property in ParseOption when calling `TypeSystemParser.Parse()`. The c# parser can then store that flag in `CSharpParsedDocument` and that class can check that flag in `GetErrorAsync()` or any other method that may need a different behavior.

PR #3180, opened 2017-10-11
Fix build on Linux

phoyd:
Project did not build on Linux because of wrong casing for Directory.Build.props. (see https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build)


mrward:
@phoyd - Thanks 👍

PR #3183, opened 2017-10-11
Fix for missing imports in generated vbnet projects.

phoyd:
Fix for missing import in generated vbnet projects. (see https://bugzilla.xamarin.com/show_bug.cgi?id=55445)


slluis:
Thanks!

PR #3185, opened 2017-10-12
Fixed 'Bug 60055 - UI hang on opening TypeScript file'

mkrueger:


mkrueger:
The fix is correct in the way it fixes the bug in the highlighting engine that can cause an infinite loop. 

I'll look at the typescript highlighing I'm not sure if that's the problem - it seems that my regex translation has an issue. But even if it is fixed the underlying problem is that the highlighting engine can infinite loop - that is what this fix is preventing.

slluis:
We need a more complete fix, this is not enough. Maybe the IDE doesn't hang, but now there is going to have delay every time that code is hit (Neeraj reported this in the bug report). So we need to: 1) Disable the whole grammar when the problem is detected, at least for the current file. 2) Inform about the problem in the log.

PR #3195, opened 2017-10-13
Bugfix14525

Jammer77:
UnitTest pad fixes.
Correct statuses for groups implemented, and refreshing groups and root element


main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
```diff
@@ -60,7 +62,71 @@ protected UnitTestGroup (string name, WorkspaceObject ownerSolutionItem): base (
...
+		}
+
+		public override TestStatus Status {
+			get {
+				return base.Status;
+			}
+			set {
+				UpdateStatusFromChildren ();
+				base.Status = value;
+			}
```
slluis: This is not correct. UpdateStatusFromChildren should be called from UnitTest.Status on its parent.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTest.cs
```diff
@@ -45,7 +45,7 @@ public abstract class UnitTest: IDisposable
 		string name;
 		IResultsStore resultsStore;
 		protected UnitTestResult lastResult;
-		UnitTestGroup parent;
```
Therzok: I wouldn't change this to `UnitTest`. If the parent can only be an `UnitTestGroup`, keep this and `SetParent` as UnitTestGroup. Since it derives from `UnitTest`, `public Parent` can be left unchanged.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
```diff
@@ -60,7 +62,53 @@ protected UnitTestGroup (string name, WorkspaceObject ownerSolutionItem): base (
...
+			};
+
+			return result;
+		}
+
+		internal void UpdateStatusFromChildren ()
+		{
+			var calculatedResult = GetLastResultDynamicaly ();
+			var storedResult = base.GetLastResult ();
+			if(calculatedResult != storedResult){
```
slluis: You need to use .Equals() to compare here.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestService.cs
```diff
@@ -105,7 +105,11 @@ static void OnExtensionChange (object s, ExtensionNodeEventArgs args)
 		public static AsyncOperation RunTest (UnitTest test, MonoDevelop.Projects.ExecutionContext context)
 		{
 			var result = RunTest (test, context, IdeApp.Preferences.BuildBeforeRunningTests);
-			result.Task.ContinueWith (t => OnTestSessionCompleted (), TaskScheduler.FromCurrentSynchronizationContext ());
+			result.Task.ContinueWith (t => {
+											 RefreshResult (test?.Parent);
```
slluis: Why call RefreshResult? it should not be necessary.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestService.cs
```diff
@@ -185,7 +185,14 @@ internal static Task RunTest (UnitTest test, MonoDevelop.Projects.ExecutionConte
 		public static AsyncOperation RunTests (IEnumerable<UnitTest> tests, MonoDevelop.Projects.ExecutionContext context)
 		{
 			var result = RunTests (tests, context, IdeApp.Preferences.BuildBeforeRunningTests);
-			result.Task.ContinueWith (t => OnTestSessionCompleted (), TaskScheduler.FromCurrentSynchronizationContext ());
+			result.Task.ContinueWith (t => { 
+											  tests.Select (tst => tst?.Parent)
+				                                   .Where (pr => pr !=null)
+				                                   .ToList ()
+				                                   .ForEach (RefreshResult);
```
slluis: This should not be necessary.

main/src/addins/MonoDevelop.UnitTesting/Gui/TestPad.cs
```diff
@@ -1,4 +1,4 @@
-﻿//
```
slluis: This file only has whitespace changes. Please revert all changes in this file.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
```diff
@@ -60,7 +62,53 @@ protected UnitTestGroup (string name, WorkspaceObject ownerSolutionItem): base (
...
+			var skipped = results.Sum (t => t.Skipped);
+			var resultStatus = ResultStatus.Inconclusive;
+			var uniqe = results.Distinct ();
+			var count = uniqe.Count ();
+			historicResult = Tests.Any (t => t.IsHistoricResult); ;
+
+			if (count == 1)
+				resultStatus = uniqe.First ().Status;
+			else if(uniqe.Any(u => u.IsFailure))  
+				resultStatus = ResultStatus.Inconclusive;
```
slluis: This is not right. This method should just calculate the last result from the children tests, not change any status. If status needs to change, it should be done elsewhere.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTest.cs
```diff
@@ -411,7 +412,7 @@ protected virtual void OnTestChanged ()
 			});
 		}
 		
-		protected virtual void OnTestStatusChanged ()
+		public virtual void OnTestStatusChanged ()
```
slluis: This should not be public.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTest.cs
```diff
@@ -37,20 +37,20 @@
...
 		IResultsStore resultsStore;
-		UnitTestResult lastResult;
+		protected UnitTestResult lastResult;
 		UnitTest parent;
 		TestStatus status;
 		WorkspaceObject ownerSolutionItem;
 		SolutionItem ownerSolutionEntityItem;
 		UnitTestResultsStore results;
-		bool historicResult;
+		protected bool historicResult;
```
slluis: I don't what lastResult and historicResult to be part of the public api. Rather than make those properties protected, it would be better to have an internal method for updating lastResult and other fields.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
```diff
@@ -60,7 +62,53 @@ protected UnitTestGroup (string name, WorkspaceObject ownerSolutionItem): base (
...
-		
+
+		public override void ResetLastResult ()
+		{
+			foreach (var t in new List<UnitTest> (Tests))
+				t.ResetLastResult();
+			base.ResetLastResult ();
+		}
+
+		public UnitTestResult GetLastResultDynamicaly ()
```
slluis: This should not be public.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
Jammer77: This method doesn't change status it just calculate status from children for returned UnitTestResult object instance.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTest.cs
Jammer77: I added protected setter to existed public property  IsHistoricResult. Is it right?

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
slluis: You're right, by bad.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTest.cs
```diff
@@ -141,14 +141,15 @@ public UnitTestResult GetLastResult ()
...
-		public void ResetLastResult ()
+		public virtual void ResetLastResult ()
 		{
 			historicResult = true;
 			OnTestStatusChanged ();
 		}
 
 		public bool IsHistoricResult {
 			get { return historicResult; }
+			protected set { historicResult = value; }
```
slluis: Please make this internal. A protected property is still visible to external subclasses.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTest.cs
```diff
@@ -37,22 +37,22 @@
...
+
 namespace MonoDevelop.UnitTesting
-{
+{
 	public abstract class UnitTest: IDisposable
 	{
 		string name;
 		IResultsStore resultsStore;
-		UnitTestResult lastResult;
+		protected UnitTestResult lastResult;
```
slluis: Please add an `internal` setter method for changing lastResult.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
```diff
@@ -60,7 +62,54 @@ protected UnitTestGroup (string name, WorkspaceObject ownerSolutionItem): base (
 				return false;
 			}
 		}
-		
+
+		public override void ResetLastResult ()
+		{
+			foreach (var t in new List<UnitTest> (Tests))
```
Therzok: Why do we construct a new list here? Is it modified while enumerated?

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
```diff
@@ -60,7 +62,54 @@ protected UnitTestGroup (string name, WorkspaceObject ownerSolutionItem): base (
...
+				t.ResetLastResult();
+			base.ResetLastResult ();
+		}
+
+		static UnitTestResult GetLastResultDynamicaly (IEnumerable<UnitTest> tests)
+		{
+			var results = tests.Select (tst => tst.GetLastResult ()).ToList ();
+			var passed = results.Sum (t => t.Passed);
```
Therzok: This method is really inefficient. This iterates a list of test results (which can possibly thousands) multiple times. Why not manually iterate once and keep track of all the values?

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestResult.cs
```diff
@@ -224,6 +224,35 @@ public void Add (UnitTestResult res)
...
+			};
+			return unknowObject.GetHashCode ();
+		}
+
+		public override bool Equals (object obj)
+		{
+			var unitTestResult =  obj as UnitTestResult;
+			if (unitTestResult == null)
+				return false;
+			return GetHashCode () == unitTestResult.GetHashCode ();
```
slluis: This is not right. If two objects have the same hash that doesn't mean they are equal.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestGroup.cs
```diff
@@ -60,7 +62,67 @@ protected UnitTestGroup (string name, WorkspaceObject ownerSolutionItem): base (
...
+				skipped += res.Skipped;
+
+				if(res.Status != lastStatus)
+					uniqeCount++;
+				
+				lastStatus = res.Status;
+			}
+
+			if (uniqeCount == 1)
+				resultStatus = tests.FirstOrDefault ().GetLastResult ().Status;
```
slluis: You can just do `resultStatus = lastStatus;`

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestResult.cs
```diff
@@ -224,6 +224,54 @@ public void Add (UnitTestResult res)
...
+		{
+			return  firstResult.Status == secondResult.Status &&
+					firstResult.IsFailure == secondResult.IsFailure &&
+					firstResult.IsSuccess == secondResult.IsSuccess &&
+					firstResult.IsInconclusive == secondResult.IsInconclusive &&
+					firstResult.IsNotRun == secondResult.IsNotRun &&
+					firstResult.Passed == secondResult.Passed &&
+					firstResult.Errors == secondResult.Errors &&
+					firstResult.Failures == secondResult.Failures &&
+					firstResult.ErrorsAndFailures == secondResult.ErrorsAndFailures &&
```
slluis: You are comparing `Errors` and `Failures`, so there is no point in comparing `ErrorsAndFailures`, which is implemented as `Errors + Failures`. Same for most properties you are comparing, they are redundant.

PR #3199, opened 2017-10-13
Revert Roslyn to 2.3.0-beta1 to make it compatible with the VS Editor…

KirillOsenkov:
… again.


KirillOsenkov:
I've ran the Metadata compat tool before & after reverting Roslyn and I've verified that the binary incompatibility is gone:

image

Never mind the ones that came back, they've always been there :)

KirillOsenkov:
This should fix https://bugzilla.xamarin.com/show_bug.cgi?id=59910

KirillOsenkov:
Wrench is green.

alanmcgovern:
Oh, i thought this was a PR against master, but it's against d15-5. Why have we not reverted master too?

KirillOsenkov:
It's not necessary to reverse master because I'm preparing a better fix - upgrading the Editor "up" to match Roslyn, as opposed to downgrading Roslyn to match Editor.

Here's the PR: https://github.com/mono/monodevelop/pull/3202

Todd and I will need to test it tomorrow though before it can go in.

PR #3200, opened 2017-10-13
Add Microsoft.VisualStudio.Threading.

KirillOsenkov:
This is in preparation for updating the editor.


KirillOsenkov:
This is intentional. We don't want to duplicate the entries in packages.config because we rely on the fact that MonoDevelop.Core will restore them first. We don't need the actual references in Core yet, so adding them to Ide for now.

Longer term I'll move the VSMEF composition from Ide down to Core, in which case it will all make sense again.

PR #3205, opened 2017-10-16
[Core] Add support for multi-dimensional array writing.

Therzok:
The problem with the old code is that it assumed that any Array was uni-dimensional. Therefore, it would end up with an invalid cast exception.
When writing any array into the binary message, flatten the array. Thus, when reading the array back, we read it as uni-dimensional. It should not impact anything.

There is a fast-path for rank1 arrays, as the foreach is rewritten to handle generalized iteration.

Bug 59805 - Visual Studio for Mac potential issue on unit testing support


slluis:
I don't understand this fix. If we serialize a multidimensional array as a 1 dimension array, won't this still crash when deserializing? the receiver will get a 1 dimension array when it expects something different.

Therzok:
array.Length returns 4 for a `int[2,2]`. Deserializing will work properly.

Therzok:
I can fix deserialization for multi-dimensional arrays.

Therzok:
Ready for re-review.

PR #3207, opened 2017-10-17
Coverity fixes

Therzok:


Therzok:
Up for reviews now.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockItem.cs
```diff
@@ -215,20 +215,20 @@ public DockItemToolbar GetToolbar (DockPositionType position)
...
-				if (toolbarTop == null)
-					toolbarTop = new DockItemToolbar (this, DockPositionType.Top);
+					if (toolbarTop == null)
+						toolbarTop = new DockItemToolbar (this, DockPositionType.Top);
 					return toolbarTop;
 				case DockPositionType.Bottom:
 					if (toolbarBottom == null)
-					                     toolbarBottom = new DockItemToolbar (this, DockPositionType.Bottom);
+	                    toolbarBottom = new DockItemToolbar (this, DockPositionType.Bottom);
 					return toolbarBottom;
```
jstedfast: argh, what is going on with this indenting? how is this line indented less than the case label?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockItem.cs
```diff
@@ -215,20 +215,20 @@ public DockItemToolbar GetToolbar (DockPositionType position)
...
 				case DockPositionType.Left:
 					if (toolbarLeft == null)
-					                     toolbarLeft = new DockItemToolbar (this, DockPositionType.Left);
+	                    toolbarLeft = new DockItemToolbar (this, DockPositionType.Left);
 					return toolbarLeft;
```
jstedfast: same with this line

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockItem.cs
Therzok: ¯\_(ツ)_/¯

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockItem.cs
iainx: gah... spaces vs tabs somehow... it looks fine on my editor

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/CairoExtensions.cs
```diff
@@ -538,7 +538,14 @@ public static ImageSurface LoadImage (Assembly assembly, string resource)
...
 			System.IO.File.WriteAllBytes (tmp, buffer);
 			var img = new ImageSurface (tmp);
-			System.IO.File.Delete (tmp);
+			try {
+				System.IO.File.Delete (tmp);
+			} catch (Exception e) {
+				LoggingService.LogWarning ($"Unable to delete {tmp} due to exception {e}");
+
+				// Only want to dispose when the Delete failed
+				img.Dispose ();
```
Therzok: You want a `throw` here

Therzok: You want a throw here after the Dispose, otherwise you'll return a non-null disposed image.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/NavigationHistoryService.cs
```diff
@@ -360,21 +360,25 @@ static void ResetCountChanges (object sender, TextFileEventArgs args)
...
-			bool closedHistoryChanged = false;
-			foreach (ProjectFileRenamedEventInfo args in e) {
-				foreach (NavigationHistoryItem point in history) {
+			foreach (NavigationHistoryItem point in history) {
+				foreach (ProjectFileRenamedEventInfo args in e) {
 					var dp = point.NavigationPoint as DocumentNavigationPoint;
-					historyChanged &= (dp != null && dp.HandleRenameEvent (args.OldName, args.NewName));
-					closedHistoryChanged &= (dp != null && dp.HandleRenameEvent (args.OldName, args.NewName));
-					if (historyChanged && closedHistoryChanged)
+					if (dp?.HandleRenameEvent (args.OldName, args.NewName) == true) {
```
slluis: This will cut the loop short. There can be more than one history item that needs to be renamed.

Therzok: HandleRenameEvent just returns true/false, it doesn't do any work

slluis: It indeed does work: https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/DocumentNavigationPoint.cs#L138

Therzok: Ooooooh. I missed that.

Therzok:
Fixed both comments.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/NavigationHistoryService.cs
```diff
@@ -360,19 +358,25 @@ static void ResetCountChanges (object sender, TextFileEventArgs args)
...
+			bool historyChanged = false, closedHistoryChanged = false;
+
+			foreach (NavigationHistoryItem point in history) {
+				foreach (ProjectFileRenamedEventInfo args in e) {
 					var dp = point.NavigationPoint as DocumentNavigationPoint;
-					historyChanged &= (dp != null && dp.HandleRenameEvent (args.OldName, args.NewName));
-					closedHistoryChanged &= (dp != null && dp.HandleRenameEvent (args.OldName, args.NewName));
-					if (historyChanged && closedHistoryChanged)
-						break;
+					historyChanged |= (dp?.HandleRenameEvent (args.OldName, args.NewName)).GetValueOrDefault ();
```
slluis: The | operator won't execute HandleRenameEvent once historyChanged becomes true.

Therzok: Does it short-circuit?

Therzok: It won't short-circuit. The handler can have side-effects, thus the runtime cannot optimize it out.

Therzok:
```
class Something
		{
			public int X = 0;
			public bool HandleRename ()
			{
				X++;
				return true;
			}
		}
		{
			var s = new Something();

			bool changed = false;
			for (int i = 0; i < 10; ++i)
				changed |= s.HandleRename();

			Console.WriteLine(s.X);
		}
```

This prints 10.

PR #3210, opened 2017-10-17
[CSharpBinding] Use roslyn classifiers for highlighting.

mkrueger:


Therzok:
578 calls to TextViewMargin.GetLayout - 14324ms before
650 calls to TextViewMargin.GetLayout - 880ms now

:+1: from me

mkrueger:
Fixed a little issue but seems now to work for me - but we need to test it a few days.

PR #3211, opened 2017-10-17
[Ide] Made code completion work fast again.

Therzok:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/CompletionTextEditorExtension.cs
```diff
@@ -190,7 +190,7 @@ public override bool KeyPress (KeyDescriptor descriptor)
 				var caretOffset = Editor.CaretOffset;
 				var token = completionTokenSrc.Token;
 				try {
-					var task = HandleCodeCompletionAsync (CurrentCompletionContext, new CompletionTriggerInfo (CompletionTriggerReason.CharTyped, descriptor.KeyChar), token);
+					var task = Task.Run (() => HandleCodeCompletionAsync (CurrentCompletionContext, new CompletionTriggerInfo (CompletionTriggerReason.CharTyped, descriptor.KeyChar), token));
 					if (task != null) {
```
jstedfast: Don't we have to wait on the HandleCodeCompletionAsync() method in the Task.Run()?

Therzok: A Task returning method is the same as a normal method. It will run synchronously on the calling thread.

The Task.Run will end when the inner code finishes. This moved the code off the UI thread itself. Adding `async() => await HandleCodeCompletionAsync ().ConfigureAwait (false)` is just useless overhead

slluis: This change is ignoring the task returned by HandleCodeCompletionAsync. The task returned by Task.Run and assigned to `task` will complete when the call to HandleCodeCompletionAsync returns, but it should complete when the task returned by HandleCodeCompletionAsync completes. To avoid this problem you should call Unwrap() on the result of Task.Run().

Therzok: The overload used is:
https://msdn.microsoft.com/en-us/library/hh160382(v=vs.110).aspx

This is not a `Task<Task>`, thus not unwrappable.

slluis: Fair enough

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/CompletionTextEditorExtension.cs
```diff
@@ -190,7 +190,7 @@ public override bool KeyPress (KeyDescriptor descriptor)
 				var caretOffset = Editor.CaretOffset;
 				var token = completionTokenSrc.Token;
 				try {
-					var task = HandleCodeCompletionAsync (CurrentCompletionContext, new CompletionTriggerInfo (CompletionTriggerReason.CharTyped, descriptor.KeyChar), token);
+					var task = Task.Run (() => HandleCodeCompletionAsync (CurrentCompletionContext, new CompletionTriggerInfo (CompletionTriggerReason.CharTyped, descriptor.KeyChar), token));
```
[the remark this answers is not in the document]
KirillOsenkov: @Therzok this is all going away soon

PR #3215, opened 2017-10-18
Master csharpclassifiers opt

Therzok:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/StackMatchExpression.cs
```diff
@@ -34,7 +34,7 @@ namespace MonoDevelop.Ide.Editor.Highlighting
 {
 	public abstract class StackMatchExpression
 	{
-		public abstract Tuple<bool, ScopeStack> MatchesStack (ScopeStack scopeStack, ref string matchExpr);
+		public abstract (bool, ScopeStack) MatchesStack (ScopeStack scopeStack, ref string matchExpr);
```
Therzok: Not sure if the API break here is worth it. I'd rather push this in.

Therzok: I can provide another overload for this if needed.

mkrueger: Doesn't matter much - we use tuples in several places ... maybe we should refactor all of them. IMO our code should be refactored more often and make use of new .NET features.

PR #3219, opened 2017-10-18
[Core] Ditch immutable collections where not needed.

Therzok:
Or use a builder to append blocks.


Therzok:
Rebased on master so VSTS works.

Therzok:
Rebased again on top of master with the updater fix.

PR #3228, opened 2017-10-19
Update Roslyn to 2.6.0

KirillOsenkov:


main/src/addins/CSharpBinding/MonoDevelop.CSharp/ProjectSearchCategory.cs
```diff
@@ -240,7 +241,8 @@ public IEnumerable<DeclaredSymbolInfoWrapper> GetAllTypes(string tag, Cancellati
 					if (kind == SyntaxKind.ConstructorDeclaration ||
 						kind == SyntaxKind.IndexerDeclaration)
 						continue;
-					if (syntaxFactsService.TryGetDeclaredSymbolInfo (current, out DeclaredSymbolInfo info)) {
+					var stringTable = Roslyn.Utilities.StringTable.GetInstance (); // TODO: fix this
```
Therzok: This should go into `SearchAsync` right before `project.Documents` iteration.

Therzok: And we could also actually cache it there.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/EditorWaitIndicator.cs
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		public IWaitContext StartWait (string title, string message, bool allowCancel)
+		{
+			return new WaitContext (title, message, allowCancel);
+		}
+
+		private sealed class WaitContext : IWaitContext
+		{
+			public WaitContext (string title, string message, bool allowCancel)
```
Therzok: This should be implemented via the status bar API.

KirillOsenkov:
Build is green. Let's get this in as this will unblock a lot of work by various teams. @Therzok feel free to log bugs for things that I should clean up/change, but I'd like to do the non-essential cleanup in subsequent PRs, as both WebTools and TypeScript teams are waiting for this.

Therzok:
Let me check if the persistent storage has been fixed in 2.6. If it hasn't, upstream needs to fix the sqlite db not being created and we need new binaries.

KirillOsenkov:
Would you be OK with merging this now and then updating the binaries again when needed? At least three people are blocked waiting on this work. I'd be happy to update binaries again when we have the storage fix.

Therzok:
Yes, definitely, go ahead and merge if it works.

PR #3233, opened 2017-10-20
[Refactoring] If we build debug builds, enable RoslynClrHeapAnalyzer

Therzok:
This will provide us with code issues for writing code. Also, serves
as an infrastructure on how to add more analyzers that are specifically
for MonoDevelop developers usage


Therzok:
Sample of what it does. It's useful for allocations in code.

screen shot 2017-10-20 at 11 19 04

PR #3245, opened 2017-10-24
[52989][Editor] Fix search widget shortcuts

sevoku:
This fixes default shortcuts inside the search widget, which overlap with default editor key bindings. Might fix similar issues with other IDE parts, which lead to unexpected text changes in the editor.


jmatthiesen:
I think you may be OK here, but please validate that Visual Studio (Windows) key bindings include Shift+Enter for Find Previous as well as Shift + F3 for Find Previous.

mrward:
VS 2017 on Windows does use shift+enter and shift+F3 for find previous in the text editor's inline search.

jmatthiesen:
I see where my response was confusing - I meant to say that we should confirm our Visual Studio (Windows) bindings also mirror that behavior from VS2017 on Windows.

PR #3246, opened 2017-10-24
[511658][Ide] Don't remove missing favorite projects

sevoku:
Don't clean up favorite recent projects automatically
based on file existance. A project might be located
on an external drive and temporary unavailable.
Remove a project from favorites only if it is unstarred
from the welcome page, or actively deleted.

(fixes bug #511658)

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentOpen.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentOpen.cs
@@ -127,9 +127,10 @@ namespace MonoDevelop.Ide.Desktop
 			}
 		}
 		
-		protected override void OnNotifyFileRemoved (string fileName)
+		public override void NotifyFileRemoved (string fileName)
 		{
 			try {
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentOpen.cs
```diff
@@ -200,7 +189,7 @@ public IList<RecentFile> GetFiles ()
 		public abstract void ClearFiles ();
 		public abstract void AddFile (string fileName, string displayName);
 		public abstract void AddProject (string fileName, string displayName);
-		public abstract void NotifyFileRemoved (string filename);
+		protected abstract void OnNotifyFileRemoved (string filename);
```
mrward: Isn't this an API break? We have gone from `public abstract void NotifyFileRemoved` to `public void NotifyFileRemoved`.

sevoku: yep, sorry, I missed that it's public

PR #3248, opened 2017-10-24
[UnitTesting] Fix xUnit test messages not displayed  …

mrward:
xUnit tests support an ITestOutputHelper interface which can be
passed to the constructor of the test. This has a WriteLine method
which can be used to output messages whilst running the test. When
this was used the text displayed in the Test Results window would
show:

    Microsoft.VisualStudio.TestPlatform.ObjectModel.TestResultMessage

Instead of the actual text. Includes some tests for the test result generation.

Removed the NuGet addin project reference which is not needed.

Disabled some tests that only work if .NET Core is installed so the tests will run on Wrench without failing.


mrward:
The UnitTesting addin tests are not currently being run on Wrench. Added a build step for the vstest-test-output-not-displayed build lane which would need to be included on the master lane.

PR #3249, opened 2017-10-25
Fixed bug where .NET 4.6 project using SDK .csproj was debugged as .NET Core

DavidKarlas:
This is fixed by making check if project is .NET Core or not more strict

Unfortunately adding unit test for this would not be simple :(

### DIFF
--- a/main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/VsTestProjectTestSuite.cs
+++ b/main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/VsTestProjectTestSuite.cs
@@ -78,7 +78,7 @@ namespace MonoDevelop.UnitTesting.VsTest
 			if (assemblyFileName.IsNull)
 				return false;
 			if (Project is DotNetProject dnp) {
-				if (Project.HasFlavor<DotNetCoreProjectExtension> ())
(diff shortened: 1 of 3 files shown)


main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/VsTestProjectTestSuite.cs
```diff
@@ -78,7 +78,7 @@ protected override bool OnCanRun (IExecutionHandler executionContext)
 			if (assemblyFileName.IsNull)
 				return false;
 			if (Project is DotNetProject dnp) {
-				if (Project.HasFlavor<DotNetCoreProjectExtension> ())
+				if (Project.HasFlavor<DotNetCoreProjectExtension> () && dnp.TargetFramework.Id.Identifier.IndexOf ("netcoreapp", StringComparison.OrdinalIgnoreCase) != -1)
```
mrward: I would be tempted to use the existing methods in the .NET Core addin to do this check. You could add an InternalsVisibleTo in the .NET Core project so you can use the framework extension methods in the .NET Core addin maybe?

Then you could write `dnp.TargetFramework.IsNetCoreApp ()`.

DavidKarlas:
@mrward tnx for suggestion

PR #3250, opened 2017-10-25
[Ide] Added Or and ProjectCapability file template conditions

DavidKarlas:
I will also open PR for Xamarin.Forms addin which will make use of this new condition

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/OrFileTemplateCondition.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/OrFileTemplateCondition.cs
@@ -1,0 +1,61 @@
+﻿//
+// OrFileTemplateCondition.cs
+//
+// Author:
+//       David Karlaš <david.karlas@microsoft.com>
+//
+// Copyright (c) 2017 Microsoft Corporation
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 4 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/OrFileTemplateCondition.cs
```diff
@@ -0,0 +1,61 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+using System.Xml;
+using System.Linq;
+using MonoDevelop.Projects;
+
+namespace MonoDevelop.Ide.Templates
+{
+	public class OrFileTemplateCondition : FileTemplateCondition
```
mrward: Does this need to be public?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/ProjectCapabilityFileTemplateCondition.cs
```diff
@@ -0,0 +1,47 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+using System.Xml;
+using MonoDevelop.Projects;
+
+namespace MonoDevelop.Ide.Templates
+{
+	public class ProjectCapabilityFileTemplateCondition : FileTemplateCondition
```
mrward: Does this need to be public?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/ProjectCapabilityFileTemplateCondition.cs
```diff
@@ -0,0 +1,47 @@
...
+	{
+		string capabilityExpression;
+		public override void Load (XmlElement element)
+		{
+			capabilityExpression = element.GetAttribute ("CapabilityExpression");
```
mrward: Not sure I like this attribute name. You end up using the following in the file template:

    `<ProjectCapability CapabilityExpression="XamarinForms" />`

The attribute name suggests you can write an expression? Could we just use:

`<ProjectCapability>XamarinForms</ProjectCapability>`

Or

`<ProjectCapability name="XamarinForms" />`

DavidKarlas: This was done intentionally to point that out, that user can write expression https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs#L803-L812

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/ProjectCapabilityFileTemplateCondition.cs
```diff
@@ -0,0 +1,63 @@
...
+			if (!expressionHasValue && !nameHasValue)
+				throw new InvalidOperationException ("Name or Expression must be set.");
+			if (nameHasValue && expressionHasValue)
+				throw new InvalidOperationException ("Set Name or Expression, but you can not set both at same time.");
+			useName = nameHasValue;
+		}
+
+		public override bool ShouldEnableFor (Project proj, string projectPath)
+		{
+			if (!(proj is DotNetProject dnp))
```
slluis: Capabilities are handled by the Project class, it doesn't need to be a DotNetProject.

PR #3256, opened 2017-10-26
Fixed 'Bug 60152 - Extension methods have wrong icon in CodeCompletion'

mkrueger:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/RoslynCompletionData.cs
```diff
@@ -90,10 +90,15 @@ public abstract class RoslynCompletionData : CompletionData
 				}
 				var modifier = GetItemModifier ();
 				var type = GetItemType ();
-				return "md-" + modifier + type;
+				var hash = modifier.GetHashCode() ^ type.GetHashCode();
+				if (!IconIdCache.ContainsKey(hash))
+					IconIdCache [hash] = "md-" + modifier + type;
```
slluis: This is not right, there is no guarantee `hash` is unique for a modifier and type.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/RoslynCompletionData.cs
```diff
@@ -90,10 +90,15 @@ public abstract class RoslynCompletionData : CompletionData
 				}
 				var modifier = GetItemModifier ();
 				var type = GetItemType ();
-				return "md-" + modifier + type;
+				var hash = modifier.GetHashCode() ^ type.GetHashCode();
+				if (!IconIdCache.ContainsKey(hash))
```
Therzok: Are we sure all the possible values do not collide? Better have a unit test asserting this.

slluis: Is this optimization really worth it?

Therzok: We create this string for each icon in the completion widget, so if you have a 1 million RoslynCompletionData elements, you end up creating 1 million IconIds.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/RoslynCompletionData.cs
mkrueger: It is - there are just a few strings I checked that the combos are unique.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/RoslynCompletionData.cs
mkrueger: In general we should never do string allocations like that if they can be cached.

mkrueger: @Therzok : Y I checked that there is no collision. Hashes can collide but we're speaking of 64 string combinations here and with such a low number a hash collision is very, very, very, very (insert some verys here) unlikely.

I added a unit test to ensure that…

PR #3261, opened 2017-10-26
Removed redundant nesting in unit test tree.

Jammer77:


main/src/addins/MonoDevelop.UnitTesting/Gui/TestNodeBuilder.cs
```diff
@@ -94,15 +102,49 @@ public override void BuildNode (ITreeBuilder treeBuilder, object dataObject, Nod
...
+				var unitTestGroup = test.Tests.First () as UnitTestGroup;
+				BuildChildNodes (unitTestGroup, builder);
+				return;
+			}
+			builder.AddChildren (test.Tests);
+		}
+
+		static bool ContainsSingleUnitTestGroup(UnitTestGroup test) => 
+						test.Tests.Count () == 1 &&
+		       			test.Tests.OfType<UnitTestGroup> ().Count () == 1;
```
slluis: This is very inefficient. Never use the Count() extension method in a collection that has a Count property.

main/src/addins/MonoDevelop.UnitTesting/Gui/TestNodeBuilder.cs
```diff
@@ -94,15 +102,49 @@ public override void BuildNode (ITreeBuilder treeBuilder, object dataObject, Nod
...
+			if (ContainsSingleUnitTestGroup (test)) {
+				BuildChildNodes (test, builder);
+				return;
+			}
 			builder.AddChildren (test.Tests);
 		}
 
+		void BuildChildNodes (UnitTestGroup test, ITreeBuilder builder)
+		{
+			if (test == null || test.Tests == null)
```
slluis: This check is unnecessary, it will never be true.

main/src/addins/MonoDevelop.UnitTesting/Gui/TestNodeBuilder.cs
```diff
@@ -94,15 +102,49 @@ public override void BuildNode (ITreeBuilder treeBuilder, object dataObject, Nod
...
+				return;
+
+			bool isSolution = test is SolutionFolderTestGroup;
+			if (!isSolution && ContainsSingleUnitTestGroup(test)) {
+				var unitTestGroup = test.Tests.First () as UnitTestGroup;
```
slluis: Use the collection indexer instead of `First()`, that is `test.Tests[0]`

main/src/addins/MonoDevelop.UnitTesting/Gui/TestNodeBuilder.cs
```diff
@@ -69,8 +70,15 @@ public override void BuildNode (ITreeBuilder treeBuilder, object dataObject, Nod
...
 
+			var singleTestSuffix = String.Empty;
+			if (test is UnitTestGroup unitTestGroup) 
+				singleTestSuffix =  GetSuxffix (unitTestGroup);
+
 			var title = RemoveGenericArgument (test.Title);
-			title = test.Title;
+			title = String.IsNullOrEmpty (singleTestSuffix) ?
+						  test.Title :
+						  $"{test.Title}{singleTestSuffix}";
```
slluis: There is no gain in doing the condition and using interpolated strings here, this can just be `test.Title + singleTestSuffix`

main/src/addins/MonoDevelop.UnitTesting/Gui/TestNodeBuilder.cs
```diff
@@ -94,15 +102,49 @@ public override void BuildNode (ITreeBuilder treeBuilder, object dataObject, Nod
...
 		}
 
+		static string GetSuxffix (UnitTestGroup unitTestGroup)
+		{
+			var result = String.Empty;
+			if (!(unitTestGroup is SolutionFolderTestGroup))
+				if (ContainsSingleUnitTestGroup (unitTestGroup)) {
+					    var testCollection = unitTestGroup.Tests;
+						var singleChildTestGroup = testCollection.FirstOrDefault () as UnitTestGroup;
+						result = $".{singleChildTestGroup.Title}{GetSuxffix (singleChildTestGroup)}";
```
Therzok: This is inefficient. This translates to recursive multiple string.Format calls being done. This can be easily unrolled into a `foreach/for` iteration with a stringbuilder.

main/src/addins/MonoDevelop.UnitTesting/Gui/TestNodeBuilder.cs
```diff
@@ -129,21 +130,18 @@ public override void BuildChildNodes (ITreeBuilder builder, object dataObject)
...
 				return;
 			}
 			builder.AddChildren (test.Tests);
 		}
 
 		static bool ContainsSingleUnitTestGroup(UnitTestGroup test) => 
-						test.Tests.Count () == 1 &&
-		       			test.Tests.OfType<UnitTestGroup> ().Count () == 1;
+						test.Tests.Count  == 1 &&
+		       			test.Tests.OfType<UnitTestGroup> ().Count() == 1;
```
Therzok: This will still do an O(n) operation. Maybe check if `SingleOrDefault() != null`

main/src/addins/MonoDevelop.UnitTesting/Gui/TestNodeBuilder.cs
```diff
@@ -94,15 +100,49 @@ public override void BuildNode (ITreeBuilder treeBuilder, object dataObject, Nod
...
+				var unitTestGroup = test.Tests[0] as UnitTestGroup;
+				BuildChildNodes (unitTestGroup, builder);
+				return;
+			}
 			builder.AddChildren (test.Tests);
 		}
 
+		static bool ContainsSingleUnitTestGroup(UnitTestGroup test) => 
+						test.Tests.Count  == 1 &&
+		       			test.Tests.OfType<UnitTestGroup> ().Count() == 1;
```
Therzok: Check if `test.Tests.OfType<UnitTestGroup> ().SingleOrDefault () != null`

main/src/addins/MonoDevelop.UnitTesting/Gui/TestNodeBuilder.cs
```diff
@@ -94,15 +100,49 @@ public override void BuildNode (ITreeBuilder treeBuilder, object dataObject, Nod
 			}
 		}
 
+		static string GetSuxffix (UnitTestGroup unitTestGroup)
```
mrward: Typo in the method name here. Should get `GetSuffix`.

main/src/addins/MonoDevelop.UnitTesting/Gui/TestNodeBuilder.cs
```diff
@@ -94,15 +100,49 @@ public override void BuildNode (ITreeBuilder treeBuilder, object dataObject, Nod
...
+			if (!isSolution && ContainsSingleUnitTestGroup(test)) {
+				var unitTestGroup = test.Tests[0] as UnitTestGroup;
+				BuildChildNodes (unitTestGroup, builder);
+				return;
+			}
 			builder.AddChildren (test.Tests);
 		}
 
+		static bool ContainsSingleUnitTestGroup(UnitTestGroup test) => 
+				test.Tests.OfType<UnitTestGroup> ().SingleOrDefault () != null;
```
slluis: This is not correct, it is not checking that there is a single child that is a group. It must check that the group is the only child. Please don't use linq for this, it is not necessary.

Therzok: I misread this as operating on two different collections initially. It's still good to do the `test.Tests.Count == 1`.

Sorry for the last time, this should actually be: `test.Tests.Count == 1 && test.Tests[0] is UnitTestGroup`

main/src/addins/MonoDevelop.UnitTesting/Gui/TestNodeBuilder.cs
```diff
@@ -94,15 +100,49 @@ public override void BuildNode (ITreeBuilder treeBuilder, object dataObject, Nod
 			}
 		}
 
+		static string GetSuxffix (UnitTestGroup unitTestGroup)
+		{
+			var stringBuilder = new StringBuilder();
+			if (!(unitTestGroup is SolutionFolderTestGroup))
+				while (unitTestGroup != null)
```
slluis: The check for `SolutionFolderTestGroup` should also be done for child items.

PR #3266, opened 2017-10-26
Make SignatureMarkupCreator public.

KirillOsenkov:
We need to reuse this for Quick Info in Razor.

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp/SignatureMarkupCreator.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp/SignatureMarkupCreator.cs
@@ -51,7 +51,7 @@ using System.Threading;
 
 namespace MonoDevelop.CSharp
 {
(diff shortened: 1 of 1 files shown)


mkrueger:
C# binding shouldn't have public API - why do you need that ? Can you use ivt ?

In general the signure markup generator shouldn't be used anymore now roslyn should be used to generate tooltips.

KirillOsenkov:
I understand that what I'm proposing is dirty, but it's just a quick temporary workaround to get us unblocked for now. With Razor, there's a lot of moving pieces and I'm trying to get obstacles out of the way.

As you're saying, the entire SignatureMarkupCreator will need to be revisited in the future anyway. This fix is the simplest thing that solves our problem for now. If you want, I can even file a bug to make it internal again in the future?

Please? Pretty please?

Therzok:
This means that we'll expose this as API for a release if we forget about this. We can't remove the API afterwards.

PR #3267, opened 2017-10-26
Make sure the temporary storage service factory is at the Host layer.

KirillOsenkov:
Otherwise it doesn't get loaded for some workspaces.


Therzok:
Do we need this in now? Persistent storage is in QA, and it should be merged, then we'll use roslyn's temporary storage service factory.

KirillOsenkov:
Yeah not really. I can wait.

PR #3274, opened 2017-10-27
Xml indent and formatting fixes

DavidKarlas:
When reviewing this I suggest looking at changes per commit since commits explain change and it's smaller chunk to digest


decriptor:
Other than an extra newline being added when hitting enter between two elements this seems fine.
ie, `<Element1><SubElement2>`

DavidKarlas:
@decriptor addressed issue with different elements being next to each other

PR #3277, opened 2017-10-30
[d15-5][NuGet] Fix transitive dependencies for project.json

mrward:
Fixed bug #60433 - Transitive dependencies not available when using
project.json
https://bugzilla.xamarin.com/show_bug.cgi?id=60433

With two projects A and B that both use project.json files, project B
referencing project A, the NuGet package dependencies used by project
A were not available to project B transitively. This was because the
project reference information was not being added to the package spec
when a restore occurred so the assemblies from the transitively
referenced NuGet packages were not being added to the
project.lock.json file for Project B.


mrward:
This has now been verified/accepted by QA.

PR #3278, opened 2017-10-30
[CSharp] Performance optimisation, code completion list appearence blocked UI for up to 1 second

DavidKarlas:
When typing something like `backtrace.GetFrame(i);` after I typed `i` typing would get stuck for almost 1 second, so I profiled and noticed that problem is in list sorting and calls to ToDisplayString which put a lot of pressure on GC... Later I checked how many times it's called... turns out... out of 6,6k completion items `DisplayText` was called 181,7k...
Solution is cache all DisplayText up front...

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/ImportSymbolCompletionData.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/ImportSymbolCompletionData.cs
@@ -37,6 +37,7 @@ namespace MonoDevelop.CSharp.Completion
 	{
 		CSharpCompletionTextEditorExtension completionExt;
 		ISymbol type;
+		string displayText;//This is just for caching, because Sorting completion list can call DisplayText many times
 		bool useFullName;
 
 		public ISymbol Symbol { get { return type; } }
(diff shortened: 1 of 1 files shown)


main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/ImportSymbolCompletionData.cs
```diff
@@ -48,7 +49,7 @@ class ImportSymbolCompletionData : CompletionData
 		}
 		static CompletionItemRules rules = CompletionItemRules.Create (matchPriority: -10000);
         public override CompletionItemRules Rules => rules;
-		public override string DisplayText { get => type.ToDisplayString (); }
+		public override string DisplayText { get => displayText; }
```
Therzok: I'd still use a `Lazy<T>` here. The current code forces the display text calculation on start. Maybe the completion item will not be displayed, say you have `20000` items which contain `i` and `40000` that don't contain `i`, this will not optimize for the best case.

DavidKarlas: I think we filter by DisplayText, hence it will be called at least 1x

DavidKarlas: Actually filter is called after Sort... but even if that is fixed... We will always have to call DisplayText at least once even if filtering is fixed to be done before Sort...

PR #3281, opened 2017-10-30
Fix case sensitivity error importing Microsoft.CSharp.targets

directhex:


directhex:
Not guilty, RE VSTS build failure.

PR #3299, opened 2017-11-01
[Core] Save user preferences in subdirectory

mrward:
The solution's user preferences file (SolutionName.userprefs) is now
saved in a subdirectory. For MonoDevelop this is:

  .md/SolutionName/v7/UserPrefs.xml

For Visual Studio:

  .vs/SolutionName/mac-v7/UserPrefs.xml

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceItem.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceItem.cs
@@ -423,29 +423,22 @@ namespace MonoDevelop.Projects
 
 		internal string GetPreferencesFileName ()
 		{
-			return GetPreferencesFileName (BrandingService.ApplicationName, BuildInfo.CompatVersion);
+			return GetPreferencesFileName (BuildInfo.CompatVersion);
 		}
 
-		internal string GetPreferencesFileName (string appName, string appVersion)
+		internal string GetPreferencesFileName (string appVersion)
 		{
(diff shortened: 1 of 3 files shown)


slluis:
I'm not sure we need .md for MonoDevelop. We could just use .vs for MD and VSMac.

mrward:
The implementation is based on that suggested by @mhutch in the VSTS work item.

If we do not do this what should happen on Linux or Windows for the directory part that includes the version number? Presumably we do not want mac-v7 on these operating systems or do we? I am assuming `mac` was added to prevent it clashing with VS on Windows.

slluis:
We can use the `xs` prefix instead of `mac`.

mrward:
So same location for MonoDevelop and VS Mac: `.vs/SolutionName/xs-v7/UserPrefs.xml`

mrward:
Updated the code to use the new location.

slluis:
I don't think the version number in the `xs-v7` should be tied to the VSMac version. I would just use `xs`. If we ever have to break the format, we can add the version number.

mrward:
Removed the version from the directory name now.

mrward:
QA have now tested/verified this pull request.

mhutch:
@mrward @Therzok we should probably have an API for this directory so the persistent storage can get it, right? :)

Therzok:
Currently has its own https://github.com/mono/monodevelop/commit/2d36e086f4f79e1c7dd81f5b4ab7b04439ef00c0#diff-bc6326fe96c7b2ac6c3563e5b2fce924R51

PR #3300, opened 2017-11-02
[Ide][Mac] Enable Alert sheeting on Mac

sevoku:
This patch enables alert dialog sheeting on Mac:

grafik

### DIFF
--- a/main/src/addins/MacPlatform/Dialogs/MacAlertDialogHandler.cs
+++ b/main/src/addins/MacPlatform/Dialogs/MacAlertDialogHandler.cs
@@ -156,8 +156,8 @@ namespace MonoDevelop.MacIntegration
 				}
 				
 				if (!data.Message.CancellationToken.IsCancellationRequested) {
-
-					var result = (int)alert.RunModal () - (long)(int)NSAlertButtonReturn.First;
+					NSWindow parent = data.TransientFor ?? IdeApp.Workbench.RootWindow;
+					var result = (int)(parent == null ? alert.RunModal () : alert.RunSheetModal (parent)) - (long)(int)NSAlertButtonReturn.First;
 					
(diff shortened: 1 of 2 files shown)


VincentDondain:
Fixed by adding:

```
if (alert.Window.IsSheet && alert.Window.SheetParent != null)
    alert.Window.SheetParent.EndSheet (alert.Window);
```

right before `NSApplication.SharedApplication.AbortModal ();` in `MacAlertDialogHandler.cs`.

main/src/addins/MacPlatform/Dialogs/MacAlertDialogHandler.cs
```diff
@@ -149,14 +149,19 @@ public bool Run (AlertDialogData data)
...
+									NSApplication.SharedApplication.AbortModal ();
 							}
 						});
 					});
 				}
 				
 				if (!data.Message.CancellationToken.IsCancellationRequested) {
-					NSWindow parent = data.TransientFor ?? IdeApp.Workbench.RootWindow;
+					NSWindow parent = null;
+					if (IdeTheme.UserInterfaceTheme != Theme.Dark || MacSystemInformation.OsVersion < MacSystemInformation.HighSierra) // sheeting is broken on High Sierra with dark NSAppearance
```
Therzok: Did you mean `&&` instead of `||` here?

Therzok: No, ignore my comment.

sevoku: nono, it's true if macOS version is lower than `HighSierra` (-> always sheet), or the theme is not `Dark` (sheet only with light theme on High Sierra).

PR #3305, opened 2017-11-03
Fix bug from commit https://github.com/mono/monodevelop/commit/6f8fa8…

Gankov:
https://github.com/mono/monodevelop/commit/6f8fa89dfb4fa4487df4127e25a2ecadee2d57af


Therzok:
Nice catch!

PR #3307, opened 2017-11-03
Fixed 'Bug 60082 - "Go to Declaration" doesn't work'

mkrueger:
IdString generation was very basic so it failed in almost any advanced case. Unfortunately cecil doesn't have an id string generator implemented so a custom one was needed.

### DIFF
--- a/main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/UnresolvedIdStringProvider.cs
+++ b/main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/UnresolvedIdStringProvider.cs
@@ -1,0 +1,198 @@
+﻿//
+// UnresolvedIdStringProvider.cs
+//
+// Author:
+//       Mike Krüger <mikkrg@microsoft.com>
+//
+// Copyright (c) 2017 Microsoft Corporation
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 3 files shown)


main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/UnresolvedIdStringProvider.cs
```diff
@@ -0,0 +1,198 @@
...
+					for (int i = start; i < parameters.Count; i++){
+						if (i > start)
+							sb.Append (',');
+						AppendTypeName (sb, parameters[i].Type, false, GetTypeParameterCount(entity));
+					}
+					sb.Append (')');
+				}
+			}
+			if (member.SymbolKind == SymbolKind.Operator && (member.Name == "op_Implicit" || member.Name == "op_Explicit")) {
+				sb.Append ('~');
```
Therzok: Shouldn't we actually be appending `implicit` or `explicit` here?

mkrueger: why ? member name is already appended at line 67.
According to c# language spec. 4.3.1 for implicit & explict operators:

Conversion operators have a trailing “~” followed by the return type.

namespace Acme {
 	class Widget: IProcess 	{ 		
public static explicit operator int(Widget x) {...} 		
public static implicit operator long(Widget x) {...} 	
}
 }

"M:Acme.Widget.op_Explicit(Acme.Widget)~System.Int32" "M:Acme.Widget.op_Implicit(Acme.Widget)~System.Int64"

At that point the string is at:
"M:Acme.Widget.op_Explicit(Acme.Widget)'

this code only appends the ~ suffix with return type.

Y I looked at the spec implementing that - id strings are very complex and I wanted a working version. Even if that code will be thrown away in the future. We should replace the decompiler backend.  I need to ask the ILSpy guys what's up with their rewrite…

mkrueger: ... but even if we replace it we probably need our own id string mapping :/. So having one working implementation isn't the worst thing that can happen.

Therzok: :+1: sounds good then

PR #3308, opened 2017-11-03
Add type check status bar icon. Fixes #57979

nosami:


nosami:
> otherwise we'll end up with an icon for each languange service

I thought this too... but does it matter?

nosami:
I'll change it anyway.

slluis:
It does matter, we don't want to duplicate logic.

nosami:
OK. Just using the methods that were already available. Not sure why I didn't do that to begin with :)

slluis:
The existing ShowStatusIcon/HideStatusIcon methods can't be reused as is, since they handle some workspace loading events. My proposal was to add new static Show/HideTypeInformationGatheringIcon() methods in TypeSystemService, and have then have MonoDevelopWorkspace and the F# language service use those.

nosami:
Ok, will update

nosami:
Done

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -135,29 +132,16 @@ static void OnLoadingFinished (EventArgs e)
...
-				workspacesLoading--;
-				if (workspacesLoading == 0 && statusIcon != null) {
-					statusIcon.Dispose ();
-					statusIcon = null;
-					OnLoadingFinished (EventArgs.Empty);
-					WorkspaceLoaded?.Invoke (this, EventArgs.Empty);
-				}
-			});
+			TypeSystemService.HideTypeInformationGatheringIcon ();
+			OnLoadingFinished (EventArgs.Empty);
```
Therzok: We now no longer trigger this event on the UI thread.

nosami:
@Therzok Good catch, fixed

PR #3311, opened 2017-11-03
[d15-5] Remove items being added to the project for non .NET Core projects

mrward:
- Fixes remove item added when UseAdvancedGlobSupport is false 

 Removing a file but not deleting it would add an MSBuild remove item
for non .NET Core projects that were not using the advanced file glob
support.

 - Fixes remove item added when removing file link

Deleting a file link in a .NET Core project would add a remove item
for the file instead of just removing the link from the project.


mrward:
This has been tested/accepted by QA.

PR #3318, opened 2017-11-06
Remove translations

bratsche:
This was already approved for 15.5 but never committed to master yet.

I can't remember if we intentionally didn't commit these to master because we want to try to update Pango first or if it was just an oversight. So I'm making this PR so we don't lose track of this.


bratsche:
Marius has a proper fix for this issue, so closing.

PR #3322, opened 2017-11-07
Removed .Wait(100) in GetCachedChunks which was causing small UI hangs while typing

DavidKarlas:
I also added extra `if (token.IsCancellationRequested) return;` so we don't schedule stuff on UI thread for no reason...
I did some testing and everything looks good `if(task.Complete==false)`(rerendering happens fine) I checked also code...

Typing feels more fluid...

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.cs
@@ -1330,12 +1330,13 @@ namespace Mono.TextEditor
 			}
 			var token = cacheSrc.Token;
 			var task = doc.SyntaxMode.GetHighlightedLineAsync (line, token);
-			task.Wait (100);
 			if (task.IsCompleted) {
 				cachedLines [lineNumber] = task.Result;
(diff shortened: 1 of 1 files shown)


alanmcgovern:
Couldn't we just make this a standard `await` ?

DavidKarlas:
I’m afraid not, because this is happening inside expose event, pretty sure postponing rendering could have unintended consequences

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.cs
```diff
@@ -1335,15 +1336,9 @@ public ChunkDescriptor (TextDocument doc, DocumentLine line, int offset, int len
...
-				Runtime.RunInMainThread (delegate {
-					if (token.IsCancellationRequested)
-						return;
-					cachedLines [lineNumber] = t.Result;
-					Document.CommitLineUpdate (line);
-				});
-			});
+				cachedLines [lineNumber] = t.Result;
+				Document.CommitLineUpdate (line);
+			}, token, TaskContinuationOptions.OnlyOnRanToCompletion, Runtime.MainTaskScheduler);
```
alanmcgovern: delightful!

PR #3324, opened 2017-11-08
Bug 60552 - Debugger "Mouse Overs" do not support copy paste

DavidKarlas:


DavidKarlas:
Actually this is fixing just 1 part of bug, will reopen when 2nd part is fixed as well...

DavidKarlas:
Fixed whole bug, hence reopened

PR #3330, opened 2017-11-10
Fixed 'Bug 60485 - RE RedundantBaseQualifierAnalyzer throws System.Ty…

mkrueger:
…peLoadException'


mkrueger:
RedundantBaseQualifierAnalyzer works - the call that caused that may be because of a roslyn version mismatch. however there were some issues with RE and 2.6.0 which got fixed.

PR #3332, opened 2017-11-11
Update VS Editor implementation to 15.0.4-pre.

KirillOsenkov:


KirillOsenkov:
This will fix the string resource issues we're seeing.

PR #3335, opened 2017-11-13
Fix #504170 No syntax highlighting for bindings

DavidKarlas:
I took xml.json copied it to xaml.json and added support for markupExtensions highlighting


mrward:
Do the VSCode syntaxes support inheriting/re-using extension syntaxes? So instead of copying the .xml one you could just re-use it and extend it with the specific .xaml support. It looks like this is what the XSL syntax is doing somehow with `"include": "text.xml"`

DavidKarlas:
I don't think it's possible to override behavior of "doubleQouteString" from another file which is needed to add markupExtension pattern matching.

mkrueger:
I assume it works for you so I approve it but as said it clashes with the xml one.

Maybe it's worth to upstream the changes to https://github.com/Microsoft/vscode

PR #3336, opened 2017-11-13
Separate .binlog processing from UI

rodrmoya:


rodrmoya:
This has now the most basic view we can have with folding:
screen shot 2017-11-14 at 10 15 09

Doesn't look perfect, but I think it's enough to start building on top of it

PR #3338, opened 2017-11-13
Change ITagBasedSyntaxHighlightingFactory to accept an ITextView

KirillOsenkov:
We need ITextView so we can save MD ITextDocument to pass changed lines to HighlightingStateChanged event.

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/VSEditor/ITagBasedSyntaxHighlightingFactory.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/VSEditor/ITagBasedSyntaxHighlightingFactory.cs
@@ -7,6 +7,7 @@ namespace Microsoft.VisualStudio.Platform
     using System;
     using System.Collections.Generic;
     using Microsoft.VisualStudio.Text;
+    using Microsoft.VisualStudio.Text.Editor;
     using MonoDevelop.Ide.Editor.Highlighting;
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


main/src/addins/MonoDevelop.SourceEditor2/VSEditor/ITagBasedSyntaxHighlightingFactory.cs
```diff
@@ -18,6 +19,6 @@ namespace Microsoft.VisualStudio.Platform
     /// </remarks>
     public interface ITagBasedSyntaxHighlightingFactory
     {
-        ISyntaxHighlighting CreateSyntaxHighlighting (ITextBuffer textBuffer);
+        ISyntaxHighlighting CreateSyntaxHighlighting (ITextView textView);
     }
```
mrward: This is a public interface. Does this break the public API?

KirillOsenkov: This is only public so that WebToolingAddin can call it. No one else calls it.

In fact, this will fix this problem we're currently seeing from BinaryCompatChecker:

```
In assembly 'WebToolingAddin, Version=1.0.6500.21499, Culture=neutral, PublicKeyToken=null': Unable to resolve member reference 'MonoDevelop.Ide.Editor.Highlighting.ISyntaxHighlighting Microsoft.VisualStudio.Platform.ITagBasedSyntaxHighlightingFactory::CreateSyntaxHighlighting(Microsoft.VisualStudio.Text.Editor.ITextView)' from assembly 'MonoDevelop.SourceEditor, Version=2.6.0.0, Culture=neutral, PublicKeyToken=3ead7498f347467b'
```

PR #3341, opened 2017-11-14
Fixed 'Bug 60622 - Can't get syntax highlighting/colorization for

mkrueger:
Xamarin.iOS bindings'.


mkrueger:
I couldn't reproduce the bug but from what it looks it was caused by a reinitialization of the text editor chain so that the RoslynClassificationHighlighting wasn't updated correctly after the reinitialization. Setting back to the old default highlighting in Dispose should fix any reinitialization issues.

The update highlighting sets back to default too in case the file got lost from the roslyn type system - so it defaults back to the syntactic highlighting.

PR #3346, opened 2017-11-14
[F#] Fixes #524437. Restore F# packages sooner

nosami:
So that StrongNamer is restored before the F# addin is built.

### DIFF
--- a/main/Makefile.am
+++ b/main/Makefile.am
@@ -29,6 +29,8 @@ NUGET_RESTORE = mono external/nuget-binary/nuget.exe restore -DisableParallelPro
 restore-packages:
 	@$(NUGET_RESTORE)
 	msbuild /t:Restore /p:RestoreDisableParallel=true external/RefactoringEssentials/RefactoringEssentials/RefactoringEssentials.csproj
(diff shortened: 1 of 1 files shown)


nosami:
re: https://github.com/xamarin/md-addins/pull/2672

main/Makefile.am
```diff
@@ -29,6 +29,8 @@ NUGET_RESTORE = mono external/nuget-binary/nuget.exe restore -DisableParallelPro
 restore-packages:
 	@$(NUGET_RESTORE)
 	msbuild /t:Restore /p:RestoreDisableParallel=true external/RefactoringEssentials/RefactoringEssentials/RefactoringEssentials.csproj
+	mono external/fsharpbinding/.paket/paket.bootstrapper.exe
+	pushd . && cd external/fsharpbinding && mono .paket/paket.exe restore && popd
```
KirillOsenkov: technically you could just do
`pushd external/fsharpbinding && mono .paket/paket.exe restore && popd`, right?
just checking to validate my noob knowledge of the shell :)

nosami: Hmm.. yes. I always use `pushd .` out of habit :)

PR #3349, opened 2017-11-15
Add Microsoft.CodeAnalysis.EditorFeatures.dll to the MEF composition.

KirillOsenkov:


KirillOsenkov:
It's still broken, composition crashes, I'm investigating.

KirillOsenkov:
It works on Windows, but on a Mac Mono is much more strict about missing assemblies, and so the lack of a few assemblies is a deal breaker.

We're missing:
Microsoft.VisualStudio.RemoteControl.dll
Microsoft.VisualStudio.CodingConventions.dll
Microsoft.VisualStudio.ImageCatalog
Microsoft.VisualStudio.Imaging

KirillOsenkov:
FYI @jasonmalinowski

KirillOsenkov:
Blockers listed in this Roslyn bug:
https://github.com/dotnet/roslyn/issues/22959

KirillOsenkov:
FYI @minestarks

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/InlineRenameService.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Editor;
+using Microsoft.CodeAnalysis.Text;
+
+namespace MonoDevelop.Ide.Composition
+{
+	[Export(typeof(IInlineRenameService))]
+	internal class InlineRenameService : IInlineRenameService
```
jasonmalinowski: This one should have been fixed with the newer packages. Are you still on the older ones?

Therzok:
Cc @slluis @mkrueger With this, we can replace our refactorings/code fixes with the Roslyn CodeFixService/CodeRefactoringService implementations.

slluis:
Don't we have to implement IInlineRenameService and ISignatureHelpBroker first?

KirillOsenkov:
@slluis no, these are very short-term hacks and will be deleted. These types will be moved out of EditorFeatures.dll to their EditorFeatures.Wpf.dll so we won't need to provide these anymore. The stubs are fine for now just to make MEF composition happy. As long as we don't use Roslyn's inline rename and Signature Help, these will never get called.

KirillOsenkov:
This project is now tracking the remaining work on EditorFeatures refactoring on the Roslyn side:
https://github.com/dotnet/roslyn/projects/32

KirillOsenkov:
@slluis @manish it would be nice for this to land in master to unblock a lot of work. What extra validation do we want here? 

/cc @DavidKarlas @Therzok

PR #3350, opened 2017-11-15
Fixed 'Bug 60178 - Completion window is being displayed on a different

mkrueger:
monitor'


mkrueger:
Didn't look at the problem - I just copied the code we're using everywhere else - that works.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/ParameterInformationWindow.cs
```diff
@@ -109,7 +109,7 @@ void Content_BoundsChanged (object sender, EventArgs e)
 			var ctx = Widget.CurrentCodeCompletionContext;
 			var lineHeight = (int)Ext.Editor.LineHeight;
 			var cmg = ParameterInformationWindowManager.CurrentMethodGroup;
-			var geometry = Visible ? Screen.VisibleBounds : Xwt.MessageDialog.RootWindow.Screen.VisibleBounds;
```
sevoku: Wouldn't this just fix it:
```var geometry = Visible ? Screen.VisibleBounds : Xwt.MessageDialog.RootWindow.Screen.VisibleBounds;```

mkrueger:
I use xwt now.

I thought DesktopService is the right way to do - we should really resolve that.

https://devdiv.visualstudio.com/DevDiv/_workitems?id=527373&_a=edit

PR #3352, opened 2017-11-15
Implemented ProgressMonitor.ReportObject()

slluis:
This method adds support for using arbitrary objects for reporting status.


main/src/core/MonoDevelop.Core/MonoDevelop.Core.ProgressMonitoring/AggregatedProgressMonitor.cs
```diff
@@ -47,6 +47,7 @@ public enum MonitorAction
 		Tasks = 0x20,
 		Cancel = 0x40,
 		FollowerCancel = 0x80,	// when the follower is cancelled, the whole aggregated monitor is cancelled.
+		ReportObject = 0x10,
```
mrward: Seems to be using the same enum value as for Dispose. Should this be 0x100 ?

slluis: Ouch

PR #3354, opened 2017-11-15
[Core] Fix failing progress monitor test

mrward:
The MonitorAction.All enum value did not include the ReportObject value.


slluis:
Thanks!

PR #3355, opened 2017-11-15
Fix 520293

nosami:
Fix unused opens analyzer bugs by switching over to new compiler API.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/520293


KirillOsenkov:
Please update the app.config file for MonoDevelop.Startup :)

PR #3357, opened 2017-11-15
Fix issues in app.config files.

KirillOsenkov:
BinaryCompatChecker found a few issues :)

I've verified that our FSharp.Core.dll is 4.4.1.0, FSharp.Compiler.Service.dll is 16.0.2.0 and the VS Editor publicKeyTokens were wrong.


nosami:
I bumped FSharp.Compiler.Service.dll to 17.0.1 [here](https://github.com/mono/monodevelop/pull/3355) :)

PR #3359, opened 2017-11-16
Fixed '51873-Implement abstract class adds code without any spaces'

mkrueger:
Seems that this is a roslyn bug - the GetChangedSolutionAsync call is
missing all post processing for the code generation. Fortunately there
is an internal method forcing the post procession to be done.

### DIFF
--- a/main/src/addins/MonoDevelop.Refactoring/MonoDevelop.CodeActions/CodeFixMenuService.cs
+++ b/main/src/addins/MonoDevelop.Refactoring/MonoDevelop.CodeActions/CodeFixMenuService.cs
@@ -407,7 +407,7 @@ namespace MonoDevelop.CodeActions
 				var oldSolution = documentContext.AnalysisDocument.Project.Solution;
 				var updatedSolution = oldSolution;
 				using (var undo = editor.OpenUndoGroup ()) {
-					updatedSolution = await act.GetChangedSolutionAsync (new RoslynProgressTracker (), token);
+					updatedSolution = await act.GetChangedSolutionInternalAsync(true, token);
 					documentContext.RoslynWorkspace.TryApplyChanges (updatedSolution, new RoslynProgressTracker ());
 				}
 				await TryStartRenameSession (documentContext.RoslynWorkspace, oldSolution, updatedSolution, token);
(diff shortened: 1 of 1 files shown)


mrward:
Is there a test we can add for this to catch if Roslyn breaks us again?

mkrueger:
That's a good question.
I suppose there will be changes there when the diagnostic service is used - marius is working on that so it doesn't make sense atm to add tests for that layer then. After that we should 100% add some tests to ensure that the integration doesn't break.

Therzok:
@mkrueger Code fixes are not in yet. We'll switch to CodeFixMenuService when EditorFeatures is added into the composition.

KirillOsenkov:
Did we file a Roslyn issue?

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.CodeActions/CodeFixMenuService.cs
```diff
@@ -407,7 +407,7 @@ public ContextActionRunner (TextEditor editor, CodeAction act)
 				var oldSolution = documentContext.AnalysisDocument.Project.Solution;
 				var updatedSolution = oldSolution;
 				using (var undo = editor.OpenUndoGroup ()) {
-					updatedSolution = await act.GetChangedSolutionAsync (new RoslynProgressTracker (), token);
+					updatedSolution = await act.GetChangedSolutionInternalAsync(true, token);
```
KirillOsenkov: 1. Could you please use named arguments so the reader easily knows that `true` is for?

2. Please add a comment describing the situation and why we're calling this method, ideally link to the Roslyn issue and leave a TODO to fix this.

decriptor:
Probably the same system, but I've noticed the same issue with extract method as well

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.CodeActions/CodeFixMenuService.cs
mkrueger: I regard named arguments as redundant code if not needed - IMO bad behavior. If it's not easy to know what the true is for it's a failure in the API IMO - which is the case here.

mkrueger:
Y all code actions suffer from that.

CyrusNajmabadi:
> action.GetChangedSolutionAsync is missing post processing in the generated code

That's by design.  If you need the post processing, you should call: GetOperationsAsync

PR #3361, opened 2017-11-16
Fix stetic warnings

iainx:
Fixes Stetic's code generation when you delete the gtk-gui/*.cs files
Generates the class with the correct public/internal visibility as specified in the gui.stetic file
Ignores warning 436 when building Stetic files because Stetic helper classes are defined multiple times


mrward:
Some of the classes have changed from private to public for some reason. The CSharpFormattingPolicyPanelWidget is one example.

Minor: The generated code seems to be using a VS style instead of the MonoDevelop style.

iainx:
They've changed from internal to public, that's because the gui.stetic files doesn't have the correct settings for them. I fixed the ones that broke the build, I'll go through and fix the rest up

As for the formatting, it's using whatever the system default policy is

mrward:
I guess that means Stetic does not use the formatting policy defined by the solution.

iainx:
https://github.com/mono/monodevelop/blob/078ca687c7ce42ed86ce47bd742f2dc61060816a/main/src/addins/MonoDevelop.GtkCore/MonoDevelop.GtkCore.GuiBuilder/GuiBuilderService.cs#L580

Is the code that formats the file, I can look into that again and see if it's using the solution one

main/src/addins/MonoDevelop.GtkCore/libsteticui/CodeGeneratorPartialClass.cs
```diff
@@ -11,15 +11,15 @@ namespace Stetic
 {
 	internal static class CodeGeneratorPartialClass
 	{
-		public static void GenerateProjectGuiCode (SteticCompilationUnit globalUnit, CodeNamespace globalNs, CodeTypeDeclaration globalType, GenerationOptions options, List<SteticCompilationUnit> units, ProjectBackend[] projects, ArrayList warnings)
+		internal static void GenerateProjectGuiCode (SteticCompilationUnit globalUnit, CodeNamespace globalNs, CodeTypeDeclaration globalType, GenerationOptions options, List<SteticCompilationUnit> units, ProjectBackend[] projects, ArrayList warnings)
```
Therzok: Why internal? It's an internal class.

mrward:
Yeah that looks like the formatting code. It is using the PolicyService.InvariantPolicies - which might be on purpose. Seems like using the project's policy being used would format it correctly if it was used.

https://github.com/mono/monodevelop/blob/078ca687c7ce42ed86ce47bd742f2dc61060816a/main/src/addins/MonoDevelop.GtkCore/MonoDevelop.GtkCore.GuiBuilder/GuiBuilderService.cs#L581

Therzok:
@iainx the generated code is still not right. Looking at it with ?w=1, we have:

Removed functionality - 
 https://github.com/mono/monodevelop/pull/3361/commits/6b71ca4d0aebe115461034a7a02d2a0d578c23b5?w=1#diff-aa9f3502558e1239494ba664201ff5f3L115
Bad escape https://github.com/mono/monodevelop/pull/3361/commits/6b71ca4d0aebe115461034a7a02d2a0d578c23b5?w=1#diff-aa9f3502558e1239494ba664201ff5f3R280
Removed widgets - https://github.com/mono/monodevelop/pull/3361/commits/6b71ca4d0aebe115461034a7a02d2a0d578c23b5?w=1#diff-c99c3f6242b31016eb0c330b71d6af7eL178
https://github.com/mono/monodevelop/pull/3361/commits/6b71ca4d0aebe115461034a7a02d2a0d578c23b5?w=1#diff-c99c3f6242b31016eb0c330b71d6af7eL204
https://github.com/mono/monodevelop/pull/3361/commits/6b71ca4d0aebe115461034a7a02d2a0d578c23b5?w=1#diff-3d4b5391ee85d15ebea448bba07067fbL159
https://github.com/mono/monodevelop/pull/3361/commits/6b71ca4d0aebe115461034a7a02d2a0d578c23b5?w=1#diff-5f8466b5a24d8943438379b40fe93ac2L243
https://github.com/mono/monodevelop/pull/3361/commits/6b71ca4d0aebe115461034a7a02d2a0d578c23b5?w=1#diff-6d9e8dbd5e784ce53e5b439b600f8d51L132

etc

PR #3364, opened 2017-11-16
Add dynamic channel support

drasticactions:
This PR allows us to setup dynamic channel support, by removing hardcoded references to "Stable," "Beta," and "Alpha".


main/src/core/MonoDevelop.Core/MonoDevelop.Core.Setup/UpdateLevel.cs
```diff
@@ -24,13 +24,121 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+using System;
+
 namespace MonoDevelop.Core.Setup
 {
-	public enum UpdateLevel
+	public class UpdateLevel
```
slluis: Please don't change public API. I propose the following solution:

1) Create a new UpdateChannel class with this implementation.
2) In UpdateService add a new UpdateChannel property, which by default takes the value of the current UpdateLevel property.
3) In UpdateService change the UpdateLevel property return a conversion of UpdateChannel to UpdateLevel. For non-default levels it could just return Alpha.
4) No change will be required in all the classes currently using UpdateLevel in this repository.
5) Change code in md-addins to use UpdateChannel instead of UpdateLevel.

drasticactions: 💯 on board. I would have done this sooner but you asked for the PR now 😉.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Updater/UpdateService.cs
```diff
@@ -85,6 +85,17 @@ static void ScheduleUpdateRun ()
 			}
 		}
 
+		public static UpdateChannel UpdateChannel {
+			get {
+				// Returned Saved Update Level
+				var prop = PropertyService.Get ("MonoDevelop.Ide.AddinUpdater.UpdateChannel", UpdateChannel.Stable.Id);
+				return new UpdateChannel (prop, prop, "", 0);
```
KirillOsenkov: add named arguments so we can see what these values mean?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Updater/UpdateService.cs
```diff
@@ -85,6 +85,17 @@ static void ScheduleUpdateRun ()
 			}
 		}
 
+		public static UpdateChannel UpdateChannel {
+			get {
+				// Returned Saved Update Level
+				var prop = PropertyService.Get ("MonoDevelop.Ide.AddinUpdater.UpdateChannel", UpdateChannel.Stable.Id);
```
slluis: If the `MonoDevelop.Ide.AddinUpdater.UpdateChannel` property is not set, it should return the channel corresponding to the currently active UpdateLevel.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Updater/UpdateService.cs
```diff
@@ -85,6 +85,17 @@ static void ScheduleUpdateRun ()
 			}
```
slluis: UpdateLevel should now be implemented by invoking UpdateChannel.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Updater/UpdateService.cs
drasticactions: I've updated it in 633a9b44f9df5498d71a777a5109a6402633778d to show that it's the "UpdateChannelId" that's being used to create the "New" `UpdateChannel`.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Updater/UpdateService.cs
drasticactions: Updated in 633a9b44f9df5498d71a777a5109a6402633778d to use the active or default `UpdateLevel` if `UpdateChannel` is not set.

KirillOsenkov:
Could you rebase to resolve merge conflicts?

drasticactions:
@KirillOsenkov I fixed the conflicts, it should be correct now.

KirillOsenkov:
Hmm, now md-addins fails to build:
```
Xamarin.Ide.Updater/UpdateService.cs(101,5): error CS0200: Property or indexer 'UpdateService.UpdateLevel' cannot be assigned to -- it is read only [/Users/vsts/agent/2.126.0/work/1/s/md-addins/Xamarin.Ide/Xamarin.Ide/Xamarin.Ide.csproj]
2018-01-12T19:20:15.3254740Z 
2018-01-12T19:20:15.3268350Z     226 Warning(s)
2018-01-12T19:20:15.3282070Z     1 Error(s)
```

See the VSTS log for details. You can also download the .binlog here:
https://devdiv.visualstudio.com/_apis/resources/Containers/1653619?itemPath=drop%2Fmdaddins.binlog

drasticactions:
I think the issue is that it's pointing at the wrong `md-addins`. Now it's pointing to master, which doesn't have the changes I made there, hence it failing.

@slluis I think `version-checks` needs to be changed so it points to the right `md-addins`, right?

KirillOsenkov:
Yeah, probably when resolving the conflict you took the md-addins commit that is later, but doesn't have your change. Maybe rebase your change to md-addins so that it's latest again and update version-checks to point to that?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.OptionPanels/AddInsOptionsPanel.cs
```diff
@@ -98,17 +98,6 @@ public void Store ()
...
-			if (checkUnstable.Active) {
-				if (radioBeta.Active)
-					UpdateService.UpdateLevel = UpdateLevel.Beta;
-				else if (radioAlpha.Active)
-					UpdateService.UpdateLevel = UpdateLevel.Alpha;
-				else if (radioTest.Active)
-					UpdateService.UpdateLevel = UpdateLevel.Test;
-			} else
-				UpdateService.UpdateLevel = UpdateLevel.Stable;
-
```
slluis: Why remove this code?

drasticactions: I removed the setter from `UpdateLevel`, making this code break. I could change it over to use `UpdateChannel`, but as far as I can tell, the radio buttons the addins page refers to don't exist on that page anymore, so those values are never set. Whatever the `UpdateLevel` (or now, `UpdateChannel`) the user has set in the Updater dialog transfers over to addins dialog. 

If that's not the case, I'm more than happy to add that code back and have it point to `UpdateChannel` instead. I just didn't see it being used.

PR #3369, opened 2017-11-16
Implement passing of binlog files to the document viewer

rodrmoya:


main/src/core/MonoDevelop.Core/MonoDevelop.Core.ProgressMonitoring/ProgressEvent.cs
```diff
@@ -0,0 +1,43 @@
...
+
+using System;
+
+namespace MonoDevelop.Core.ProgressMonitoring
+{
+	public class ProgressEvent
+	{
+	}
+
+	public class ProjectStartedProgressEvent : ProgressEvent
```
slluis: This should be in MonoDevelop.Projects namespace

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -0,0 +1,115 @@
...
+		{
+			BuildOutput = output;
+		}
+
+		protected override void OnObjectReported (object statusObject)
+		{
+			switch (statusObject) {
+			case ProjectStartedProgressEvent pspe:
+				if (File.Exists (pspe.LogFile)) {
+					BuildOutput.Load (pspe.LogFile);
```
slluis: You can't load the log file in the ProjectStarted event, since there will be no log yet. You'll have to do it when the project build has finished.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.ProgressMonitoring/ProgressEvent.cs
rodrmoya: Changed

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
rodrmoya: Yes, the name might be misleading, see the last commits, the LogFile is just added to a list, and only processed when asked to. But yes, I'll change it to add it only when the ProjectFinished event is passed around

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.Formats.MSBuild/BuildEngine.cs
```diff
@@ -92,7 +92,7 @@ internal string GetUnsavedProjectContent (string file)
 			}
 		}
 
-		void BeginBuildOperation (IEngineLogWriter logWriter, MSBuildVerbosity verbosity, ProjectConfigurationInfo [] configurations)
+		void BeginBuildOperation (IEngineLogWriter logWriter, string binglogFilePath, MSBuildVerbosity verbosity, ProjectConfigurationInfo [] configurations)
```
Therzok: `binLogFilePath` is typo'd

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteBuildEngine.cs
```diff
@@ -216,16 +216,20 @@ internal void Shutdown ()
 		/// <summary>
 		/// Indicates that a build session is starting
 		/// </summary>
-		public async Task BeginBuildOperation (TextWriter logWriter, MSBuildLogger logger, MSBuildVerbosity verbosity, ProjectConfigurationInfo[] configurations)
+		public async Task BeginBuildOperation (ProgressMonitor monitor, MSBuildLogger logger, MSBuildVerbosity verbosity, ProjectConfigurationInfo[] configurations)
 		{
-			buildSessionLoggerId = RegisterLogger (logWriter, logger);
+			buildSessionLoggerId = RegisterLogger (monitor.Log, logger);
 			try {
+				var binlogPath = Path.ChangeExtension (Path.GetTempFileName (), "binlog");
```
slluis: GetTempFileName creates a temp file which is never deleted. Also the binlog file doesn't seem to be deleted.

rodrmoya: right, the problem is when do we remove it? that's why I used GetTempFileName, which uses the user's $TMP folder, which gets cleaned on reboots, at least on Mac and Linux. For Windows it might be a problem indeed, AFAIK the $TMP dirs are not cleaned up there

rodrmoya: Ah yes, we can remove it when disposing BuildOutput I guess?

slluis: Yes

rodrmoya: Ok, added code to do this, only for temporary .binlog files

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputViewContent.cs
```diff
@@ -53,10 +59,27 @@ public BuildOutputViewContent (FilePath filename)
...
-				return filename.FileName;
+				return filename.FileName ?? GettextCatalog.GetString ("Build Output");
 			}
 		}
+
+		public override void Dispose ()
+		{
+			control.Dispose ();
+		}
 	}
```
rodrmoya: this is missing a call to `base.Dispose` right @jtorres ?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/ErrorListPad.cs
```diff
@@ -1031,28 +1019,26 @@ static int CategoryIterSort (TreeModel model, TreeIter a, TreeIter z)
 
 		internal void FocusOutputView ()
 		{
-			logBtn.Active = true;
-			HandleLogBtnToggled (this, EventArgs.Empty);
+			HandleLogBtnClicked (this, EventArgs.Empty);
 		}
```
rodrmoya: since there's no `outputView` anymore, what does this do @jtorres ?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/MSBuildOutputProcessor.cs
```diff
@@ -0,0 +1,122 @@
...
+				return;
+			}
+
+			Clear ();
+			base.Process ();
+
+			try {
+				binlogReader.Replay (FileName);
+			} catch (Exception) {
+			}
```
slluis: You should at least log the exception here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputProcessor.cs
```diff
@@ -191,25 +144,45 @@ private void ProcessNode (TextEditor editor, BuildOutputNode node, int tabPositi
...
+		}
+
+		public void Dispose ()
+		{
+			Dispose (true); 
+		}
+
+		void Dispose (bool disposing)
+		{
+			if (disposing) {
```
Therzok: If you don't do anything on the finalizer, please don't create one. A finalizer forces the object to be promoted from gen0.

rodrmoya: hmm, what do you mean? it is doing something, removing the files

rodrmoya: Also, the GC.SuppressFinalize call in the destructor is supposed to avoid that, right?

mrward: I think GC.SuppressFinalize should be called in the Dispose method so the finalizer is not run. Currently it is being called in the destructor.

Therzok: This removes the files only when you `Dispose (true)`, which is not done in the finalizer. Yes, SuppressFinalize should be done in Dispose. Please remove the finalizer alltogether.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/ErrorListPad.cs
jtorres: @rodrmoya it's referenced [here](https://github.com/mono/monodevelop/blob/e3037364ddeadfc1216d447da70588df004a8c22/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Tasks/TaskListEntry.cs#L265)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputViewContent.cs
jtorres: @rodrmoya hmmm what you mean?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/ErrorListPad.cs
rodrmoya: right, so seems it's not needed, as we don't have an `outputView` anymore

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputViewContent.cs
rodrmoya: yes, I think so, @mrward @Therzok ?

Therzok: Yes.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteBuildEngineManager.cs
```diff
@@ -357,7 +357,9 @@ internal static object StartBuildSession (TextWriter tw, MSBuildLogger logger, M
 				foreach (var b in builders.GetAllBuilders ())
 					if (b.BuildSessionId == session) {
 						b.BuildSessionId = null;
-						await b.EndBuildOperation ();
+
+						var sessionInfo = session as SessionInfo;
```
Therzok: Hard cast if we don't null check.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -29,33 +29,65 @@
...
+		}
+
+		void SetupBuildOutput (BuildOutput output)
 		{
-			this.filename = filename;
+			BuildOutput = output;
+			ProcessLogs (false);
+
+			BuildOutput.OutputChanged += (sender, e) => {
+				if (Runtime.IsMainThread) {
```
Therzok: Why does processing of all the projects have to be done on the UI thread?

Therzok: At least some of the work done in ProcessLogs is non-UI dependent.

Therzok: Make sure to switch other callers of this method to delegate to background thread (i.e. from button clicked) when processing

PR #3370, opened 2017-11-16
Switch the VS Editor implementation to Release

KirillOsenkov:
This gets rid of an assert that degrades dogfooding experience.


KirillOsenkov:
For now I've published a Release build of Microsoft.VisualStudio.Text.Implementation.dll so that we stop seeing that assert during Undo.

I will later investigate and fix the root cause, but for now this should unblock the dogfooding experience. Sorry for the pain.

PR #3373, opened 2017-11-16
[build] fix issues bundling live player

sgmunn:
Really bump the size of the .dmg when creating it. There does not appear to be a suitable way to calculate this accurately and playing trial and error trying to find a value that works is painful and likely to break again. This is a hammer approach, but since all the extra space in the .dmg is compressed away to nothing it doesn't really matter that much.


main/build/MacOSX/make-dmg-bundle.sh
```diff
@@ -34,7 +34,7 @@ rm -f "$DMG_FILE.master"
  	
 # Compute an approximated image size in MB, and bloat by 1MB
```
alanmcgovern: maybe we should put the explanation in this comment :)

PR #3375, opened 2017-11-17
Rollback PR #3211 temporarily.

KirillOsenkov:
It breaks WebTooling completion for several reasons:

1. Sometimes HandleCodeCompletionAsync used to return null, which was semantically significant for WebTooling. Now it never returns null, but always returns a task.

2. WebTooling is not yet prepared to be able to run on a background thread. We need to do work to make our code properly thread-safe.

3. The timings have changed, so certain logic in WebTooling async completion is now wrong (we check if a completion is already up and don't return anything, and that check returns different results based on whether we're still on the UI thread or on a background thread now).

We will be revisiting the entire Completion story pretty soon with Roslyn EditorFeatures so I'm sure we'll arrive at a good solution that works for everybody. For now we want to keep the current WebTooling working in master, because it would take us a lot of work to adapt for this change.

So while we agree that this is a good change in theory (frees up the UI thread) it is currently breaking so we'd like to pursue a tactical fix short term.


KirillOsenkov:
This rolls back #3211

PR #3376, opened 2017-11-17
Fixed 'Bug 60724 - No way of changing behavior of Automatically insert

mkrueger:
parentheses after completion in the IDE'

This behavior should be disabled completely because it's gone in the
next iteration of the completion engine.


mkrueger:
Build lane:
https://code.internalx.com/Wrench/BuildDetails?buildId=e643cd3b23777eea37bfb47a188c3f5b

slluis:
Mike, we are not doing more 15.4 releases.

PR #3380, opened 2017-11-17
Fixed '526670-Parameter tooltip sometimes doesn't show information'

mkrueger:


mkrueger:
Worked most of the time because usually parameter length finds the right overload but in case of overloads with the same parameter list length the old code took the first method it could find.

PR #3385, opened 2017-11-20
Implemented referenced assemblies loading error in the unit test tree.

Jammer77:


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectReference.cs
```diff
@@ -498,6 +498,8 @@ public static ProjectReference RenameReference (ProjectReference pref, string ne
 
 							if (!ownerProject.CanReferenceProject (p, out reason))
 								return reason;
+						} else {
+							return GettextCatalog.GetString ("Assembly not found");
```
mrward: Not sure this message is correct. The code is looking for a project not an assembly here. So shouldn't the message be Project not found?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectReference.cs
```diff
@@ -628,7 +630,7 @@ void AddRequiredPackages (List<string> result, SystemPackage fromPackage)
 		
 		void UpdatePackageReference ()
 		{
-			if (referenceType == ReferenceType.Package && ownerProject != null) {
+			if ((referenceType == ReferenceType.Package || referenceType == ReferenceType.Project) && ownerProject != null) {
```
mrward: Not sure that if the reference is a project that the code inside the if should be called. The code is using the AssemblyContext to find a reference which I do not believe will not work for a project reference. So I do not think this is correct. If the notfound flag needs to be set for a project reference I think it makes more sense to check the project exists with different code and not use the code in this if statement.

main/src/addins/MonoDevelop.UnitTesting.NUnit/MonoDevelop.UnitTesting.NUnit/NUnitProjectTestSuite.cs
```diff
@@ -68,6 +69,16 @@ public NUnitProjectTestSuite (DotNetProject project, NUnitVersion version): base
 			this.project = project;
 			project.NameChanged += OnProjectRenamed;
 			IdeApp.ProjectOperations.EndBuild += OnProjectBuilt;
+			CheckReferencesErrors ();
+		}
+
+		void CheckReferencesErrors()
+		{
+			var errorReferenceCollection = project.References.Where (reference => !string.IsNullOrEmpty (reference.ValidationErrorMessage))
+			                                      			 .ToList ();
```
mrward: I would use the IsValid property on the ProjectReference instead of checking if the ValidationMessage is null or empty. The IsValid property does this check.

main/src/addins/MonoDevelop.UnitTesting/Services/UnitTest.cs
```diff
@@ -55,6 +55,8 @@ public abstract class UnitTest: IDisposable
 
 		public virtual bool CanMergeWithParent => false;
 
+		public string ReferenceErrorMessage { get; protected set; }
```
slluis: Errors might not be related to references. Let's rename to just "ErrorMessage".

main/src/addins/MonoDevelop.UnitTesting.NUnit/NUnit3Runner/NUnitTestRunner.cs
```diff
@@ -136,11 +137,28 @@ public NunitTestInfo GetTestInfo (string path, string[] supportAssemblies)
...
 				return null;
 		}
+
+		bool CheckXmlForError(XmlElement root, out string result)
+		{
+			var elements = root.GetElementsByTagName ("properties");
+			var skipReasonString = string.Empty;
+			foreach (XmlElement element in elements) 
+				//Reason why i parse InnerXml - GetAttribute returns empty strings
+				if (element.InnerXml.Contains ("name=\"_SKIPREASON\""))
```
slluis: `<properties>` contains a list of `<property>` elements. You have to get the attribute from the property elements. If GetAttribute() is returning an empty string you are doing something wrong.

main/src/addins/MonoDevelop.UnitTesting.NUnit/MonoDevelop.UnitTesting.NUnit/NUnitProjectTestSuite.cs
```diff
@@ -68,6 +69,16 @@ public NUnitProjectTestSuite (DotNetProject project, NUnitVersion version): base
 			this.project = project;
 			project.NameChanged += OnProjectRenamed;
 			IdeApp.ProjectOperations.EndBuild += OnProjectBuilt;
+			CheckReferencesErrors ();
+		}
+
+		void CheckReferencesErrors()
```
slluis: Let's revert this. The other check should be enough.

PR #3392, opened 2017-11-21
Fixed '527398 - "Move type to ..." doesn't move class to a new file'

mkrueger:


mkrueger:
According to the roslyn team GetOperationsAsync is the thing we need to call here.

decriptor:
Thanks! I've really missed this! :) 👍

PR #3397, opened 2017-11-23
Preserve the app bundle directory instead of moving it into .dmg

KirillOsenkov:
We need to do more checks on the bundle directory after the .dmg is created (such as compatibility-check).


KirillOsenkov:
Any idea why the verify-codesign fails for this change in Wrench?

KirillOsenkov:
`Visual Studio.app\Contents\MacOS\MonoDoc.app\Contents\MonoBundle\libintl.dylib` used to be just the text `libintl.8.dylib`, now it's an entire binary file. Not sure if it's my fault? Or is it related?

alanmcgovern:
We need to preserve symlinks when copying. I’ll check it later!

alanmcgovern:
LGTM! merging this like a boss as CI is happy!

KirillOsenkov:
👍

PR #3398, opened 2017-11-23
Master fixeditor

mkrueger:


mkrueger:
Should improve things. But I need feedback from Mikayla before merging.

main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Document/DiffTracker.cs
```diff
@@ -91,6 +91,9 @@ void TrackDocument_TextChanged (object sender, MonoDevelop.Core.Text.TextChangeE
 				var startLine = trackDocument.GetLineByOffset (change.NewOffset);
 				var endLine = trackDocument.GetLineByOffset (change.NewOffset + change.InsertionLength);
 				var lineNumber = startLine.LineNumber;
+				var oldState = lineStates [lineNumber];
```
Therzok: Nice

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
```diff
@@ -3050,7 +3048,8 @@ void OnDocumentStateChanged (object s, TextChangeEventArgs args)
 				var change = args.TextChanges[i];
 				var start = editor.Document.OffsetToLineNumber (change.NewOffset);
 				var end = editor.Document.OffsetToLineNumber (change.NewOffset + change.InsertionLength);
-				editor.Document.CommitMultipleLineUpdate (start, end);
+				if (start != end)
```
Therzok: Shouldn't we still be comitting a line update?

Therzok: And can we batch line updates without comitting them unti lthe end?

mkrueger: I already reomved it - it's not needed in general because in line updates are already handled. That's just for programatical updates that don't call update explicitly.

However it doesn't hurt in the current line case - because updates are batched :-).

mkrueger:
Improves things a lot - we should merge it.

PR #3401, opened 2017-11-24
Make sure updater is disabled when running tests

slluis:


Therzok:
What I meant is, `make test` will disable the updater for the user. Or `mdtool run-md-tests` will.

slluis:
The setting is set but not saved, so it should be ok. In any case I'll try to find a better solution, just in case there is a test that ends saving preferences.

slluis:
Pushed a better fix.

PR #3406, opened 2017-11-27
[DotNetCore] File watcher

mrward:


main/external/fsharpbinding/MonoDevelop.FSharp.Tests/ProjectTests.fs
```diff
@@ -112,6 +112,7 @@ type ProjectTests() =
                 project.FileName <- FilePath(projectPath)
                 project.AddFile(path / "Controllers\\HomeController.fs", "Compile") |> ignore
                 do! project.SaveAsync(monitor)
+                project.Dispose()
```
nosami: better to `use! project.SaveAsync(monitor)` instead

nosami: Actually, this isn't right as `project.SaveAsync` returns void.

Better to change this line -
 https://github.com/mono/monodevelop/blob/860d7d1e23693e7da2f00ee7b5cc74802d1b0082/main/external/fsharpbinding/MonoDevelop.FSharp.Tests/ProjectTests.fs#L111

to
```fs
use project = solutionItem :?> DotNetProject
```
to avoid the manual `.Dispose()`. It works like `using` in C#, but doesn't need an explicit block.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectNodeBuilder.cs
```diff
@@ -220,6 +222,30 @@ void OnSystemFileDeleted (object sender, FileEventArgs args)
 			}
 		}
 
+		void OnSystemFileCreated (object sender, FileEventArgs args)
+		{
+			if (!args.Any (f => f.IsDirectory))
+				return;
+
+			// When a folder is created, we need to refresh the parent if the folder was created externally.
```
slluis: Why is this necessary? When a file is added to a project, the folder that contains it will also be added.

mrward: If you move a folder b, which has files, from say `projectfolder/a/b` to `projectfolder/b` then it would not be displayed in the solution window.

The problem was that the ProjectNodeBuilder
was using the FileService.FileRemoved to refresh the Solution pad.
Since the new directory had no associated ProjectFolder no tree
node could be found to refresh the UI. Now the FileService.FileCreated
event is used and the directory's parent is found and refreshed.

Maybe there is a better way to do this.

mrward:
This has been approved by QA.

PR #3407, opened 2017-11-27
[529486][Ide] Improve dialog auto-parenting logic

sevoku:
Don't use "special" top level windows (popups, pads, etc.) for message dialog auto-parenting.
First of all this is important for sheeting alerts on macOS, which should never be parented to pads or other "special" windows. But also in general we shouldn't use pads, or special windows (popups, menus, etc.) for dialog auto-parenting, not only on Mac.

(fixes bug #529486)


mhutch:
IIRC the original intent was that, if a modal dialog pops up another modal dialog, the second dialog should be parented to the first so that it displays on top of it.

PR #3408, opened 2017-11-27
Master optimizequicktasks

mkrueger:


main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.QuickTasks/QuickTaskOverviewMode.IdleUpdater.cs
```diff
@@ -0,0 +1,213 @@
...
+				state.ColorCache [IndicatorDrawingState.UnderlineErrorColor] = SyntaxHighlightingService.GetColor (mode.TextEditor.EditorTheme, EditorThemeColors.UnderlineError);
+				state.ColorCache [IndicatorDrawingState.UnderlineWarningColor] = SyntaxHighlightingService.GetColor (mode.TextEditor.EditorTheme, EditorThemeColors.UnderlineWarning);
+				state.ColorCache [IndicatorDrawingState.UnderlineSuggestionColor] = SyntaxHighlightingService.GetColor (mode.TextEditor.EditorTheme, EditorThemeColors.UnderlineSuggestion);
+				state.ColorCache [IndicatorDrawingState.BackgroundColor] = SyntaxHighlightingService.GetColor (mode.TextEditor.EditorTheme, EditorThemeColors.Background);
+
+				ResetEnumerators ();
+
+				cr = new Cairo.Context (surface.Surface);
+				cr.Scale (displayScale, displayScale);
+				GLib.Idle.Add (RunHandler);
```
Therzok: Can we check once more that the token is not cancelled before queuing up the idle?

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.QuickTasks/QuickTaskOverviewMode.IndicatorDrawingCache.cs
```diff
@@ -0,0 +1,148 @@
...
+				case DiagnosticSeverity.Hidden:
+					return BackgroundColor;
+				default:
+					throw new ArgumentOutOfRangeException ();
+				}
+			}
+
+			public bool Equals (IndicatorDrawingState other)
+			{
+				if (other == null)
```
Therzok: This seems like a really slow method to call in the Idle handler. If there are a lot of usages/tasks, we'll end up iterating the whole thing.

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.QuickTasks/QuickTaskOverviewMode.IdleUpdater.cs
mkrueger: It's running on the UI thread so it's never canceled there we're not in a multi thread situation here.

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.QuickTasks/QuickTaskOverviewMode.IndicatorDrawingCache.cs
mkrueger: Even if it's 100000 it's nothing :)

mkrueger: var s1 = new IndicatorDrawingState ();
				var s2 = new IndicatorDrawingState ();

				for (int i = 0; i < 1000000; i++) {
					s1.UsageRectangles.Add ((0, 0));
					s1.TaskRectangles.Add ((0, 0));
					s2.UsageRectangles.Add ((0, 0));
					s2.TaskRectangles.Add ((0, 0));
				}

				Stopwatch sw = new Stopwatch ();
				sw.Start ();
				Console.WriteLine (s1.Equals (s2));
				sw.Stop ();
				Console.WriteLine (sw.ElapsedMilliseconds);

Did that - Took 20ms on my machine and that's for 2 million comparisons. Where atm we can expect ~10k in the worst case. So that is >no< problem. I exepect if computers have screens that have million pixel lines they're much faster :). 

So not an issue here since it's an O(1) method on a specific machine.

KirillOsenkov:
@mkrueger this PR adds System.ValueTuple.dll 4.4.0 (assembly version 4.0.2.0) and it gets copied to `build\AddIns\DisplayBindings\SourceEditor\System.ValueTuple.dll`. The rest of MonoDevelop uses 4.3.1 (assembly version 4.0.1.1). The binding redirects in app.config redirect everything to 4.0.1.1.

Do you need that System.ValueTuple.dll in the SourceEditor folder? Is this intentional? If not, I'd set the `<Private>False</Private>` metadata on the reference.

Also I'm OK with using 4.4.0, but then we need to upgrade the rest of MonoDevelop to use 4.4.0 instead and not forget the app.config file.

KirillOsenkov:
I'll take care of this in a separate PR.

PR #3409, opened 2017-11-27
[Core] Fix assembly check for frameworks without assemblies

sevoku:
As of cf0a3720c80b6320d7dea9163ffc11096a01020b frameworks may contain
0 assemblies. This patch fixes the broken assembly check if the framework has also
no manifest.


sevoku:
I don't know why and how that happened, but while playing around with configurations the .NET 4 Client profile had this issue for me. It might be something wrong with my installation, I'll investigate this later.

PR #3411, opened 2017-11-27
[Ide] Fix leaks in the options panel.

Therzok:


Therzok:
@mrward `Destroy()` causes `OnDestroyed` to be called, yes, but the reverse is not true.

Overriding `Destroy()` means your code is only called when the object is destroyed from managed via `Destroy`. `OnDestroyed` guarantees that the code executes even when the destroy comes from native.

Therzok:
Failures unrelated.

PR #3413, opened 2017-11-27
[C#] Fix performance issue with files opened for a long time

Therzok:
The logic about this change is not keeping around all previous versions
of a CSharpParsedDocument. What happened here was:
* ParseOptions contained a ref to the old ParsedDocument
* ParsedDocument contained a reference to the current ParseOptions
* On the next query, both of these instances are kept alive in the new
ParsedDocument.

This caused a huge chain of live document+options+others to be alive
after many edits.

### DIFF
--- a/main/tests/MonoDevelop.CSharpBinding.Tests/MonoDevelop.CSharpBinding.Parser/CSharpParsedDocumentTests.cs
+++ b/main/tests/MonoDevelop.CSharpBinding.Tests/MonoDevelop.CSharpBinding.Parser/CSharpParsedDocumentTests.cs
@@ -1,0 +1,89 @@
+﻿//
+// CSharpParsedDocumentTests.cs
+//
+// Author:
+//       Marius Ungureanu <maungu@microsoft.com>
+//
+// Copyright (c) 2017 
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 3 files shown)


slluis:
Any idea for unit testing this?

Therzok:
I wrote a test like this:
```
[TestFixture]
	public class CSharpParsedDocumentTests
	{
		[Test]
		public void DoesNotLeakPreviousDocument ()
		{
			var mem1 = GC.GetTotalMemory (true);

			// Force only the last doc to be alive at this point.
			var doc = GetParsedDocumentStress ();

			var mem2 = GC.GetTotalMemory (true);

			var delta = mem2 - mem1;
			Assert.That (delta, Is.AtMost (24));
		}

		CSharpParsedDocument GetParsedDocumentStress ()
		{
			CSharpParsedDocument old = null, doc = null;

			for (int i = 0; i < 100; ++i) {
				old = doc;

				var options = new ParseOptions {
					OldParsedDocument = old,
				};

				doc = new CSharpParsedDocument (options, "mock");
			}
			return doc;
		}
	}
```

But the delta is -334600. That's not reliable

Therzok:
Added a unit test. Without this fix, the finalized count is 0. With this patch, it's 99.

main/tests/MonoDevelop.CSharpBinding.Tests/MonoDevelop.CSharpBinding.Parser/CSharpParsedDocumentTests.cs
```diff
@@ -0,0 +1,91 @@
...
+
+		[Test]
+		public void DoesNotLeakPreviousDocument ()
+		{
+			const int DocumentCount = 100;
+
+			// Force only the last doc to be alive at this point.
+			var doc = GetParsedDocumentStress (DocumentCount);
+
+			GC.Collect ();
```
Therzok: Not sure how many GCs to trigger here. I said it would be safe to go with 4.

Therzok:
Test failed on VSTS with 98 items finalized, not 99. This is weird.

Therzok:
re-review please, force pushed test to look for at least 98 items finalized, as the mono GC can be lazy.

PR #3423, opened 2017-11-28
[StressTest] Add stress test program

mrward:
Opens the IDE, opens a solution, opens some files, edits one file,
builds and then debugs the solution.

Currently IDE memory usage is output to the console after each iteration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ obj/
 /main/packages
 /main/.nuget/
 pull-package.exe
+/main/tests/StressTest/TestProject/bin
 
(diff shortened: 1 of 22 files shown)


DavidKarlas:
I tested this and it works as expected, I really like it's in MonoDevelop repo, test is slow that is because default wait time on UserInterfaceTests.Ide.WaitForIdeIdle (); is 10seconds...
I would also like some solution reopening and some other stuff, but I guess that can be different test...
This does exactly what we wanted, hence I think we can Merge and later improve...

PR #3430, opened 2017-11-29
D15 5 leak fixes

Therzok:


main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.cs
```diff
@@ -151,6 +151,23 @@ public TextViewMarginAccessibilityProxy ()
...
+				Accessible.InsertionPointLineNumber = null;
+				Accessible.NumberOfCharacters = null;
+				Accessible.FrameForRange = null;
+				Accessible.LineForIndex = null;
+				Accessible.RangeForLine = null;
+				Accessible.StringForRange = null;
+				Accessible.RangeForIndex = null;
+				Accessible.StyleRangeForIndex = null;
+				Accessible.RangeForPosition = null;
+				Accessible.GetVisibleCharacterRange = null;
```
mkrueger: Wouldn't that all be better done in an Accessible.Dispose () ? 
That looks like code that's not belonging here.

Therzok: Could be done, yes. @iainx what do you think?

iainx: Yes, I was thinking that

Therzok: we can leave as-is for d15-5 fix proper in master

AdrianMurphy:
After talking with Greg, he noted that Lluis said this should go into 15.5.1. Not merging this for RTW, but leaving it open.

PR #3431, opened 2017-11-29
Fixed '530737 - Information tooltips are not showing method docs, but

mkrueger:
object docs'


mkrueger:
seems properties, fields & events were missing in the  monodoc provider code so it always gave the description of the containing type.

Issue #3437, opened 2017-12-01
Tooltips are unreadable with Monokai editor theme


mhutch:
Editor tooltips are unreadable when using the Monokai dark editor theme with the light UI theme

<img width="1005" alt="00000138" src="https://user-images.githubusercontent.com/183285/33496862-d1912d70-d699-11e7-8f65-e69c6972abc2.png">

> VS bug #568030

slluis:
We used to have a logic to automatically adapt colors to backgrounds. I wonder what happened to that.

sevoku:
Proposed fix: https://github.com/mono/monodevelop/pull/3888

xamarin-release-manager:
Fixed in version 7.5.0.437 (master)

Author: Vsevolod Kukol
Commit: d50b371e84a45e131376846f6c76744fdfa68e72 (mono/monodevelop)

slluis:
@sevoku can you prepare a PR for 15.6?

xamarin-release-manager:
Fixed in version 7.4.1.25 (d15-6)

Pull Request #4098 merged by: Lluis Sanchez
Author: mono
Commit: f1f64ce35353f6a4a6d85f6406bd6b620b3c02a1 (mono/monodevelop)

xamarin-release-manager:
[VS sync] The field 'Milestone' contains the value '7.4' that is not in the list of supported values

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #3888 merged by: Lluis Sanchez
Author: mono
Commit: 4f95fb9676851ad665d9fe558d23cddfb992b3ec (xamarin/monodevelop)

PR #2870, opened 2017-08-05
[Addin] (Test PR) To allow CBinding to use "Project" property

gitexperience:
CMakeProject being derived from SolutionItem is unable to access the "Project" property. The PR is intended to test the changes needed to make the "Project" property available to the CBinding's CMakeProject.

PR #3440, opened 2017-12-01
Issue #3437

gitexperience:
Since text color and tooltips color are same (i.e., monokai-white), small change in that will work for both. @mhutch, can we use any better option ?


mhutch:
I think the more general fix in this case is maybe for the tooltips to use the background color of the theme?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/themes/MonokaiStyle.json
```diff
@@ -5,7 +5,7 @@
 	"originator": "Xamarin Inc. (http://xamarin.com)",
 
 	"palette": [
-		{ "name": "monokai-white", "value": "#F8F8F2" },
+		{ "name": "monokai-white", "value": "#c9c99c" },
```
mkrueger: Why changing that ? Did the colors in monokai change at some point ?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditorViewContent.cs
```diff
@@ -154,10 +155,9 @@ void UpdateStyleParent (MonoDevelop.Projects.Project styleParent, string mimeTyp
 
 			var mimeTypes = DesktopService.GetMimeTypeInheritanceChain (mimeType);
 
-			if (styleParent != null)
-				policyContainer = styleParent.Policies;
-			else
-				policyContainer = MonoDevelop.Projects.Policies.PolicyService.DefaultPolicies;
+				policyContainer = ((styleParent as IPolicyProvider)
+									?.Policies)
+									?? MonoDevelop.Projects.Policies.PolicyService.DefaultPolicies;
```
mkrueger: Does this improve readability ? just curious about opinions.

gitexperience: @mkrueger  this was to avoid NRE if the styleParent does not implement IPolicyProvider.
These changes are from this [PR!](https://github.com/mono/monodevelop/pull/2870) . By mistake, i made these changes in my master branch and so these changes are carry forward here, waiting to get it merged.

Therzok: Then keep the if check and do
```
if (styleParent is IPolicyProvider provider)
    policyContainer = provider.Policies;
else
    policyContainer = ...
```

PR #3443, opened 2017-12-03
Added `--profiler:` option to StressTest so it can automatically  take…

DavidKarlas:
... heapshots and make more detailed reports

Primary idea here, is not so much `printreport=` option but `type=all`, which will allow us to create "reproducable" and good heapshots for later use in UI profiler.
Code outside `--profiler:` was modified as minimal as possible(parsing arguments and passing options around).
Mono.Profiling.Log was copied as source from runtime primary to make few optimisations and avoid problems with GAC being out of sync.

### DIFF
--- a/main/tests/StressTest/MonoDevelop.StressTest/StressTestApp.cs
+++ b/main/tests/StressTest/MonoDevelop.StressTest/StressTestApp.cs
@@ -37,11 +37,12 @@ namespace MonoDevelop.StressTest
 	{
 		List<string> FoldersToClean = new List<string> ();
 		ITestScenario scenario;
+		StressTestOptions.ProfilerOptions ProfilerOptions;
 
 		public StressTestApp (StressTestOptions options)
 		{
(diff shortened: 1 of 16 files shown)


main/tests/StressTest/MonoDevelop.StressTest/StressTestApp.cs
```diff
@@ -141,6 +155,11 @@ void OnCleanUp ()
 
 		void ReportMemoryUsage (int iteration)
 		{
+			System.Threading.Thread.Sleep (5000);//Make sure IDE stops doing what it was doing
```
mrward: Can we use the WaitForIdeIdle check here instead of a Thread.Sleep?

DavidKarlas: WaitForIdeIdle feels so weird, as far as I understand it's checking for status bar and has 10sec delay, I will change to WaitForIdeIdle.

PR #3455, opened 2017-12-05
Fix leaks2

Therzok:
Fixed leaks surrounding editing files with projections and ~a leak of a TextBufferUndoManager for each document that was opened.~ (filed [upstream](https://github.com/Microsoft/vs-editor-api/issues/8))

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
@@ -171,9 +171,11 @@ namespace MonoDevelop.Ide.TypeSystem
 			return Task.Run (async delegate {
 				var projects = new ConcurrentBag<ProjectInfo> ();
 				var mdProjects = solution.GetAllProjects ();
-				foreach (var p in projectionList)
-					p.Dispose ();
-				projectionList = projectionList.Clear ();
+				lock (projectionListUpdateLock) {
+					foreach (var p in projectionList)
+						p.Dispose ();
+					projectionList = projectionList.Clear ();
+				}
 				projectIdMap.Clear ();
 				projectIdToMdProjectMap = projectIdToMdProjectMap.Clear ();
 				projectDataMap.Clear ();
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -171,9 +171,11 @@ Task<SolutionInfo> CreateSolutionInfo (MonoDevelop.Projects.Solution solution, C
 			return Task.Run (async delegate {
 				var projects = new ConcurrentBag<ProjectInfo> ();
 				var mdProjects = solution.GetAllProjects ();
-				foreach (var p in projectionList)
-					p.Dispose ();
-				projectionList = projectionList.Clear ();
+				lock (projectionListUpdateLock) {
+					foreach (var p in projectionList)
+						p.Dispose ();
```
DavidKarlas: Can you make copy of list and dispose that list outside `lock`, just to be safe...

Therzok: Done.

Therzok:
Seems like the TextBufferUndoManager leak is in upstream.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -171,9 +171,14 @@ Task<SolutionInfo> CreateSolutionInfo (MonoDevelop.Projects.Solution solution, C
 			return Task.Run (async delegate {
 				var projects = new ConcurrentBag<ProjectInfo> ();
 				var mdProjects = solution.GetAllProjects ();
+				ImmutableList<ProjectionEntry> toDispose;
+				lock (projectionListUpdateLock) {
+					toDispose = projectionList;
```
KirillOsenkov: Do you ever do anything with toDispose?

Therzok: Forgot to rename! good catch

PR #3456, opened 2017-12-05
[Structured build output] Save build output doc

jtorres:
Enables Save / SaveAs, execute such commands, and initialise ErrorPad accordingly. 
Updates .binlog when new build is launched.


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputViewContent.cs
```diff
@@ -33,55 +35,86 @@ namespace MonoDevelop.Ide.BuildOutputView
...
+			Widget.Dispose ();
 			base.Dispose ();
 		}
+
+		internal async Task UpdateContent ()
+		{
+			if (ContentName != Widget.BuildOutput.FilePath) {
+				await Save (new FileSaveInformation (ContentName, null));
+			}
+		}
```
rodrmoya: Actual save support for the build output is implemented here -> https://github.com/mono/monodevelop/pull/3442 so I think this should be implemented on top of that?

PR #3486, opened 2017-12-06
[528908][F#] Fix compiler configuration panel alignments

sevoku:
(fixes bug #528908)

@nosami we've discussed that we want to add a documentation file path entry, but it's not as simple as expected. We can't use the normal Gtk file entry, cause it won't open a native file picker. For that purpose we have `MonoDevelop.Components.FileEntry` ([link](https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/FileEntry.cs)) but we can't use it without (a) referencing `MonoDevelop.Ide` and (b) using stetic (external widget import is broken). So I'd say, let's postpone that for the full preferences redesign, when we'll rewrite all the preference/option panels.

grafik


sevoku:
Yeah, it's using the assembly basename by default.

PR #3487, opened 2017-12-07
[Ide] Make document switcher shortcuts configurable

bratsche:
These shortcuts were generated dynamically, and so they were not
configurable for users who want to rebind Cmd-1 .. Cmd-9 to other
things.

Fixes VSTS #527532


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/WindowCommands.cs
```diff
@@ -92,41 +92,113 @@ protected override void Run ()
...
+	internal class OpenDocumentHandlerBase : CommandHandler
 	{
-		protected override void Update (CommandArrayInfo info)
+		int index;
+
+		// 1-based index
+		protected OpenDocumentHandlerBase (int index)
 		{
-			int i = 0;
-			foreach (Document document in IdeApp.Workbench.Documents) {
```
Therzok: This changes the MainMenu behaviour. We no longer display this at all.

screen shot 2017-12-07 at 08 06 37

We also need a handler that displays documents after the 10th one, which would mirror behaviour of `OpenDocumentListHandler` for `i > 10` and manually add the first 10 handlers [here](https://github.com/mono/monodevelop/blob/d291f726db411d0ed37120541b67a49a42e4d235/main/src/core/MonoDevelop.Ide/ExtensionModel/MainMenu.addin.xml#L266) too.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/WindowCommands.cs
```diff
@@ -96,26 +96,33 @@ internal class OpenDocumentListHandler : CommandHandler
...
 				commandInfo.Description = GettextCatalog.GetString ("Activate document '{0}'", commandInfo.Text);
 				if (document.Window.ShowNotification) {
 					commandInfo.UseMarkup = true;
 					commandInfo.Text = "<span foreground=" + '"' + "blue" + '"' + ">" + commandInfo.Text + "</span>";
 				}
 
 				//Add AccelKey
-				if (i < 10) {
+				if (IdeApp.Workbench.Documents.Count + i < 10) {
+					System.Console.WriteLine ("Setting accel key...");
```
Therzok: Debug code.

PR #3490, opened 2017-12-11
Queue alert sheets

olegoid:
In some cases, a window should show multiple alerts simultaneously.

At the moment if a window needs to show two alerts at the same time one of them will be displayed in a proper manner (sliding from the top and attached to the title bar). The second alert, however, will be just "floating" in the air.

Example:

screen shot 2017-12-11 at 2 21 35 pm

Ideally, alerts should be queued and shown one by one. Like this:

alerts

### DIFF
--- a/main/src/addins/MacPlatform/Dialogs/MacAlertDialogHandler.cs
+++ b/main/src/addins/MacPlatform/Dialogs/MacAlertDialogHandler.cs
@@ -193,13 +193,15 @@ namespace MonoDevelop.MacIntegration
 
 				if (optionButtons != null) {
 					foreach (var button in optionButtons) {
-						var option = data.Options [(int)button.Tag];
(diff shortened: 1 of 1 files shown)


sgmunn:
looks good to me, @sevoku there are other "changes" only because the code that got wrapped in OnAlertClosed was also indented so looks like more changes than just the change from RunSheetModal > BeginSheet.

olegoid:
Wait, I found a problem with PR changes.

olegoid:
The issue I mentioned in https://github.com/mono/monodevelop/pull/3490#issuecomment-350868413 should be fixed now.

olegoid:
Thanks @sevoku.

@sgmunn do we need anything else to merge this PR?

sgmunn:
I think this is good enough

PR #3504, opened 2017-12-14
Fixes Microsoft.Build reference paths to use MSBuildToolsPath property

netonjm:
MSBuildToolsPath has the correct directory for MSBuild folder in windows and mac (probably in linux too), then there is no need to add additional changes to current path:

The conditional in MSBuild_OSS_BinDir was failing and giving a empty value, without this property correctly setted, <HintPath>$(MSBuild_OSS_BinDir)XXXXX.dll</HintPath> has a wrong path and 
it didn't fail to compile becuase is taking the `Microsoft.Build.dll` from GAC folder which is different version (older) than the MSBuildToolsPath and it didnt' have some types.

`/Library/Frameworks/Mono.framework//Versions/5.8.0/lib/mono/gac/Microsoft.Build/14.0.0.0__b03f5f7f11d50a3a/Microsoft.Build.dll `

Fixed this in MonoDevelop.Core.csproj and MonoDevelop.Ide.csproj

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
@@ -93,27 +93,22 @@
     <NoWarn>1591;1573</NoWarn>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
-  <PropertyGroup>
-    <MSBuild_OSS_BinDir Condition="'$(OS)' == 'Windows_NT'">$(MSBuildToolsPath)\</MSBuild_OSS_BinDir>
-    <!-- when building with xbuild on Unix, $(MSBuildToolsPath) is like `$prefix/lib/mono/xbuild/*/bin`
+  <ItemGroup>
+  <!-- when building with xbuild on Unix, $(MSBuildToolsPath) is like `$prefix/lib/mono/xbuild/*/bin`
 	 when building with msbuild on Unix, $(MSBuildToolsPath) is like `$prefix/lib/mono/msbuild/*/bin`
 	 Prefer referencing msbuild 15.* assemblies over 14.1 . At runtime, we use correct one anyway
 	 -->
-    <MSBuild_OSS_BinDir Condition="'$(OS)' == 'Unix' and '$(MSBuild_OSS_BinDir)' == '' and Exists('$(MSBuildToolsPath)\..\..\..\msbuild\15.0\bin\MSBuild.exe')">$(MSBuildToolsPath)\..\..\..\msbuild\15.0\bin\</MSBuild_OSS_BinDir>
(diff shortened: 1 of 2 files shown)


mrward:
Using a wrench build of structured-build-output currently throws an error on loading a .cs file:

https://gist.github.com/mrward/049d490b509e5cf7c9d7e864b047958d

@netonjm was seeing this with a local build of this branch.

mrward:
Builds from wrench for the master branch seem to be OK. I have been using 7.4.0.805 (master) from wrench this morning without any problems.

rodrmoya:
well, on master the exception is not thrown, so I guess not

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.csproj
```diff
@@ -211,28 +211,18 @@
...
-      <Private>False</Private> 
-    </Reference> 
-    <Reference Include="Microsoft.Build.Framework"> 
-      <HintPath>$(MSBuild_OSS_BinDir)Microsoft.Build.Framework.dll</HintPath> 
-      <Private>False</Private> 
-    </Reference> 
-    <Reference Include="Microsoft.Build.Utilities.Core"> 
-      <HintPath>$(MSBuild_OSS_BinDir)Microsoft.Build.Utilities.Core.dll</HintPath> 
-      <Private>False</Private> 
+    <!-- when building with xbuild on Unix, $(MSBuildToolsPath) is like `$prefix/lib/mono/xbuild/*/bin`
```
radical: The reason for using `MSBuild_OSS_BinDir` is explained by this comment. This was useful when we were transitioning over from xbuild to msbuild and wanted to still be able to build with xbuild.
Accordingly, this comment can be dropped or just reduced to 

```<!-- when building with msbuild on Unix, $(MSBuildToolsPath) is like `$prefix/lib/mono/msbuild/*/bin` -->```

File: main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj [hunk 2]
```diff
@@ -93,27 +93,22 @@
     <NoWarn>1591;1573</NoWarn>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
-  <PropertyGroup>
-    <MSBuild_OSS_BinDir Condition="'$(OS)' == 'Windows_NT'">$(MSBuildToolsPath)\</MSBuild_OSS_BinDir>
-    <!-- when building with xbuild on Unix, $(MSBuildToolsPath) is like `$prefix/lib/mono/xbuild/*/bin`
+  <ItemGroup>
+  <!-- when building with xbuild on Unix, $(MSBuildToolsPath) is like `$prefix/lib/mono/xbuild/*/bin`
```
radical: Same comment update here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.csproj
```diff
@@ -211,28 +211,18 @@
...
+    when building with msbuild on Unix, $(MSBuildToolsPath) is like `$prefix/lib/mono/msbuild/*/bin`
+    Prefer referencing msbuild 15.* assemblies over 14.1 . At runtime, we use correct one anyway
+    -->
+    <Reference Include="Microsoft.Build">
+      <HintPath>$(MSBuildToolsPath)\Microsoft.Build.dll</HintPath>
```
Therzok: We're losing local copying here.

PR #3507, opened 2017-12-15
Initialize TimeCounter.TimerTraceList without any condition

manish:
Bug 540261: "A fatal error has occurred" when opening New Project Dialog
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/540261


manish:
Ran using PR build and all the tests are now passing.
https://devdiv.visualstudio.com/DevDiv/_release?definitionId=512&_a=release-contribution-tab-ms.vss-test-web.test-result-in-release-management&releaseId=65599

Surprisingly, this happens only on Sierra and not on HighSierra. This tin fix has fixed the issue, but I am not capable to assess the wider impact.

slluis:
Alternative fix: https://github.com/mono/monodevelop/pull/3510

PR #3512, opened 2017-12-15
Structured build output - Preview tabs

netonjm:
Includes a feature in TabStrip to draw new type of tabs (Preview Tabs).

We don't know what kind of tab it is since the content is attached to the document.
(for testing purposes I enabled allways this in every .binlog document, we need to change this)

I created a subset of TabContainers, a new entity to manage all data related to view.

This containers are setup on every call to OnSizeAllocated method (on window resize).

TabWidth (and TargetWidth) are the current size for all tabs, which modifies depending the size of the window (and containers).  

In draw Method are commented some functions to debug purposes to ensure the draw of containers are calculated correctly.

```
//Draw container background
//ctx.SetSourceColor (new Cairo.Color (0, 0.2, 0.3));
//ctx.Rectangle (contentStartX, Allocation.Y, contentEndX - contentStartX , Allocation.Height);
//ctx.Fill ();

//ctx.SetSourceColor (new Cairo.Color (0.7, 0.6, 0.6));
//ctx.Rectangle (previewTabContainer.ContentStartX, Allocation.Y, previewTabContainer.ContentEndX - previewTabContainer.ContentStartX, Allocation.Height);
//ctx.Fill ();
```

In Draw, we have some CalculateTabs functions to get some handlers of drawing tab methods. For  closing tabs, Normal tabs and Preview tabs. 

We walk over the tab array in each container starting from the Container.ContentStartX drawing the tab with the current TabWidth size.

We return an array of EventHandlers because we want operate (reorder) the array before invoke them (event handlers).

At the moment, I'm having an issue removing tabs from the current array. Probably could be related to my cache collections in TabContainer.Tabs. Maybe could be a good thing change this to use the collection in DockNotebookTab.cs to better consistance with the data.

DO NOT MERGE IT's not finished.


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.DockNotebook/DockNotebook.cs
```diff
@@ -131,6 +135,14 @@ public DockNotebook ()
 
 		Cursor fleurCursor = new Cursor (CursorType.Fleur);
 
+		internal bool ContainsTab (DockNotebookTab tab)
+		{
+			if (AllTabs.Any (t => tab == t)) {
+				return true;
+			}
+			return false;
+		}
```
rodrmoya: this can be simplified to `return AllTabs.Any (t => tab == t);`

rodrmoya:
As we just talked, the other issue I see here is that it always opens .binlog files, even static ones open via File->Open, in preview tabs. In that case it should use a normal tab.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.DockNotebook/DockNotebook.cs
Therzok: Depends on where `ContainsTab` is called. LINQ version allocates a capture per call.

rodrmoya: right, but that `.Any` is being called always, so instead of

`if (...Any()) return true;
return false`

it can be simplified to just return `.Any`, or am I missing something (could be :D)?

Therzok: Ohhhh, I read this as a loop at first. Yeah, simplifying is good

rodrmoya:
This seems to work great for me now, doing some more testing though

netonjm:
Ready to review guys!

PR #3514, opened 2017-12-18
Implemented  ellipsize of a long text in the unit tests context menu

Jammer77:


main/src/addins/MonoDevelop.UnitTesting/Services/AbstractUnitTestEditorExtension.cs
```diff
@@ -253,6 +253,15 @@ public override void PopupContextMenu (UnitTestLocation unitTest, int x, int y)
 							menuItem.Clicked += new TestRunner (unitTest.UnitTestIdentifier + id, project, true).Select;
 							submenu.Add (menuItem);
 
+							const int maxLabelLength = 80;
+							if (label.Length > maxLabelLength) {
+								const string gap = "...";
```
sgmunn: I'm not completely sure if this is the right approach, but I do know that the ellipsis should be a single character

"\u2026" and not 3 separate "." chars.

main/src/addins/MonoDevelop.UnitTesting/Services/AbstractUnitTestEditorExtension.cs
```diff
@@ -253,6 +253,15 @@ public override void PopupContextMenu (UnitTestLocation unitTest, int x, int y)
...
+								int remainsLength = (maxLabelLength - gap.Length) / 2;
+								string start = label.Substring (0, remainsLength);
+								string end = label.Substring (label.Length - remainsLength, remainsLength);
+								label = $"{start}{gap}{end}";
```
iainx: Need to strip any whitespace off start and end so that there isn't a space beside the …

main/src/addins/MonoDevelop.UnitTesting/Services/AbstractUnitTestEditorExtension.cs
iainx: (you can generate … on the Mac with alt+; )

PR #3515, opened 2017-12-19
[Structured build output] Save build output

jtorres:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputViewContent.cs
```diff
@@ -33,56 +35,86 @@ namespace MonoDevelop.Ide.BuildOutputView
...
-				return control;
+		public override Task Save (FileSaveInformation fileSaveInformation)
+		{
+			var result = false;
+
+			if (File.Exists (fileSaveInformation.FileName))
+				File.Delete (fileSaveInformation.FileName); //TODO: backup before removing?
+
+			File.Copy (Widget.BuildOutput.FilePath, fileSaveInformation.FileName);
+
```
rodrmoya: This won't work for the aggregated log case, which contains several files, unless it's a very simple solution with just one project, so if I've understood correctly the code, when that happens, you would just be saving the 1st .binlog, and ignore the rest?

so, when saving, you need to call the BuildOutput's Save() method, which does what it needs to concatenate all the .binlog files into one.

PR #3516, opened 2017-12-19
Implemented copy to to clipboard  from  ThreadPad

Jammer77:


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ThreadsPad.cs
```diff
@@ -146,25 +148,54 @@ void ShowPopup (Gdk.EventButton evt)
...
-				session.Stop ();
-			};
-			context_menu.Items.Add (pauseExecution);
+			var copyExecution = new ContextMenuItem (GettextCatalog.GetString ("_Copy"));
+			copyExecution.Sensitive = true;
+			copyExecution.Clicked += CopyExecution_Clicked;;
+			context_menu.Items.Add (copyExecution);
+
+			var process = store.GetValue (selected, (int)Columns.Object) as ProcessInfo;
+			if (process != null) { //User right-clicked on thread and not process
```
mrward: The comment now looks wrong here compared to the original code. Before the process was null when this comment applied. So maybe the comment should be removed?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ThreadsPad.cs
```diff
@@ -146,25 +148,54 @@ void ShowPopup (Gdk.EventButton evt)
...
-			context_menu.Items.Add (continueExecution);
-			var pauseExecution = new ContextMenuItem (GettextCatalog.GetString ("Pause"));
-			pauseExecution.Sensitive = session.IsRunning;
-			pauseExecution.Clicked += delegate {
-				session.Stop ();
-			};
-			context_menu.Items.Add (pauseExecution);
+			var copyExecution = new ContextMenuItem (GettextCatalog.GetString ("_Copy"));
+			copyExecution.Sensitive = true;
+			copyExecution.Clicked += CopyExecution_Clicked;;
```
mrward: Duplicate semi-colon at end of line.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ThreadsPad.cs
```diff
@@ -146,25 +148,54 @@ void ShowPopup (Gdk.EventButton evt)
...
-			};
-			context_menu.Items.Add (continueExecution);
-			var pauseExecution = new ContextMenuItem (GettextCatalog.GetString ("Pause"));
-			pauseExecution.Sensitive = session.IsRunning;
-			pauseExecution.Clicked += delegate {
-				session.Stop ();
-			};
-			context_menu.Items.Add (pauseExecution);
+			var copyExecution = new ContextMenuItem (GettextCatalog.GetString ("_Copy"));
+			copyExecution.Sensitive = true;
```
mrward: Not sure I like the name 'copyExecution'. The others make sense 'pauseExecution' and 'resumeExecution'. I think this shold just be copy or copyMenu.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ThreadsPad.cs
```diff
@@ -146,25 +148,54 @@ void ShowPopup (Gdk.EventButton evt)
...
+				};
+				context_menu.Items.Add (pauseExecution);
+			}
 			context_menu.Show (this, evt);
 		}
+
+		void CopyExecution_Clicked (object sender, ContextMenuItemClickedEventArgs e)
+		{
+			TreeIter selected;
+			if (!tree.Selection.GetSelected (out selected))
```
mrward: Do we need to do this check here? This check is made before the menu is opened in the ShowPopup menu.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ThreadsPad.cs
```diff
@@ -146,25 +148,54 @@ void ShowPopup (Gdk.EventButton evt)
...
+				return;
+
+			var infoObject = tree.Model.GetValue (selected, (int)Columns.Object);
+
+			string bufferText = string.Empty;
+			if(infoObject is ThreadInfo threadInfo)
```
mrward: Missing space between 'if' and first brace.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ThreadsPad.cs
```diff
@@ -146,25 +148,54 @@ void ShowPopup (Gdk.EventButton evt)
...
+				bufferText = $"{threadInfo.Id} {threadInfo.Name} {threadInfo.Location}";
+
+			if (infoObject is ProcessInfo processInfo) 
+				bufferText = $"{processInfo.Id} {processInfo.Name} {processInfo.Description}";
+
```
mrward: Can the infoObject be anything other than a ProcessInfo or ThreadInfo? Just wondering since we would end up copying an empty string. Can we copy something useful here or should we disable/hide the Copy menu instead?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ThreadsPad.cs
Jammer77: It's a not my comment and it still honest.  I'm not a big fan of comments at all.

mrward: The comment seems wrong to me. Originally the code was:

```			var process = store.GetValue (selected, (int)Columns.Object) as ProcessInfo;
 			if (process == null)
 				return;//User right-clicked on thread and not process
```

In the above code the comment seems OK and makes sense to me.

Now the code is:

```
			var process = store.GetValue (selected, (int)Columns.Object) as ProcessInfo;
			if (process != null) { //User right-clicked on thread and not process
```

So the comment looks wrong or the if statement is incorrect. If process is not null then the user right clicked on a process.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ThreadsPad.cs
Jammer77: GetValue returns object. Potentially it can be anything.  Now it's only ProcessInfo or ThreadInfo. I prefer to call overridden ToString(). But I do everything according general code style.

PR #3518, opened 2017-12-20
Fixed  ignoring  "build before running" setting when running unit test

Jammer77:
I know, but other option about a building tested project, before testing.  Bug report was about building tests project.

https://devdiv.visualstudio.com/DevDiv/_workitems?id=a2108d31-086c-4fb0-afda-097e4cc46df4&_a=query&fullScreen=true


DavidKarlas:
Btw, there is also Unit tests specific setting for this, maybe that should be used.

PR #3530, opened 2017-12-22
[Roslyn] Bump to VSWin 15.6 equivalent in vs-for-mac packages

Therzok:


KirillOsenkov:
Can you share more details about this? What is this version you're bumping to? The reason I'm asking is that Roslyn recently merged the vs-for-mac branch back into master, so now regular Roslyn master builds should have everything we need.

To be clear: I'd be happy to have a newer Roslyn in master, since this would help with adding EditorFeatures to VSMEF.

Therzok:
@KirillOsenkov any version that's newer than our vs-for-mac packages suffices for me. There's perf improvements that we need in.

slluis:
Do we need this for 15.6?

KirillOsenkov:
Oh, if this goes in, then we can add editorFeatures to the VSMEF composition?

Therzok:
@KirillOsenkov yes, then we can bump editorfeatures

@slluis yes, we should have it for 15.6 to match VS 15.6. But I think I got the wrong version, we should be targeting upstream 2.6.x, not 2.7.x

KirillOsenkov:
Let's take `2.7.0-beta3-62509-03` - this is the official version that is being inserted in VS for Windows for escrow. It's not the final build, but it's stabler and better than the random build that we took.

Also I think it's best to get this version into both d15-6 and master. This is so that we can dogfood while stabilizing and if we find an issue we need to let the Roslyn team know anyway because it'll likely be critical for them as well. Also if they discover an issue, it'll be easier for us to take their patch since we'd be on the same baseline.

Therzok:
Updated.

PR #3545, opened 2018-01-02
Fixed '530391 - Placing caret in global:: qualifier highlights whole …

mkrueger:
…document'

This fix switches over to the roslyn highlighting engine.


mkrueger:
Keywords are no longer highlighted but I suppose that is the right engine in roslyn to use ? 

Keyword highlighting is not really important - more important is the read/write coloring - is the highighting engine still a work in progress in roslyn ? 

btw. I think it's usable that way. And a good first step to get more roslyn provided features just working.

Therzok:
Yes, it seems like the roslyn engine only highlights symbols.

KirillOsenkov:
Roslyn also highlights many things, such as certain keywords:
http://source.roslyn.io/#Microsoft.CodeAnalysis.EditorFeatures/Extensibility/Highlighting/IHighlighter.cs,dd45304f10610aad,references

mkrueger:
This is the old basic highlighting engine - there is a new one that is more sophisticated. 

IDocumentHighlighter - this moved out of editior features and is more complex but it's missing keyword highlighting so far.

PR #3555, opened 2018-01-03
[A11y] Disable all the accessibility proxies if AtkCocoa is not loaded

iainx:


Therzok:
@iainx can you re-create the PR or push a dummy commit?

PR #3557, opened 2018-01-03
[d15-6][Core] Fix project reference to shared project considered invalid

mrward:
A project reference to a shared project that exists was shown as
invalid in the Solution window. It would have an error icon and have
the tooltip 'Project not found.' even though the shared project did
exist.

The problem was that the code was changed in commit d40ba5fec794edbe14ccf58c1e177944cd2a6fe9
so that if a project was not a DotNetProject then it was marked as
invalid even if it existed. The SharedAssetsProject class does
not derive from DotNetProject so the project was treated as though
it could not be found.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectReference.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectReference.cs
@@ -492,13 +492,14 @@ namespace MonoDevelop.Projects
 					}
 				} else if (ReferenceType == ReferenceType.Project) {
 					if (ownerProject != null && ownerProject.ParentSolution != null && ReferenceOutputAssembly) {
-						DotNetProject p = ResolveProject (ownerProject.ParentSolution) as DotNetProject;
-						if (p != null) {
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectReference.cs
```diff
@@ -492,13 +492,14 @@ public static ProjectReference RenameReference (ProjectReference pref, string ne
 					}
 				} else if (ReferenceType == ReferenceType.Project) {
 					if (ownerProject != null && ownerProject.ParentSolution != null && ReferenceOutputAssembly) {
-						DotNetProject p = ResolveProject (ownerProject.ParentSolution) as DotNetProject;
-						if (p != null) {
+						var p = ResolveProject (ownerProject.ParentSolution);
+						var dotNetProject = p as DotNetProject;
```
KirillOsenkov: >dotNetProject [](start = 10, length = 13)

if (p is DotNetProject dotNetProejct) ...

PR #3558, opened 2018-01-03
Avoid a first-chance exception after showing Windows File Open dialog.

KirillOsenkov:


main/src/addins/WindowsPlatform/WindowsPlatform/Dialogs/OpenFileDialogHandler.cs
```diff
@@ -159,18 +159,23 @@ static List<string> GetSelectedItems (CommonFileDialog dialog)
 			IShellItemArray resultsArray;
 			uint count;
 
-			try {
-				nativeDialog.GetSelectedItems (out resultsArray);
-			} catch (COMException ex) {
+			var hr = nativeDialog.GetSelectedItems(out resultsArray);
+			if (hr != 0x0000) {
```
ToddGrun: Probably just change this to a simple 0 to match what's below

PR #3560, opened 2018-01-03
Don't deploy Microsoft.Build.Tasks.Core.dll.

KirillOsenkov:
We don't deploy the 3 other MSBuild dlls, so let's not deploy this one.
MSBuild is taken from the toolset that's installed separately from MonoDevelop. Otherwise we might run into situations where our own copy is loaded instead of the official one and cause a mismatch.


Therzok:
Does the IDE still work without this?

Therzok:
Yep, this breaks the IDE

KirillOsenkov:
Pushed a fix. cc @rodrmoya

Therzok:
Let's see if VSTS is happy

KirillOsenkov:
VSTS is happy! :)

PR #3561, opened 2018-01-04
Fixed '527496 - Error "while update show code surroundings window"'

mkrueger:
TextEditorResolverProvider needs to be removed - resolving doesn't
work this way. Fortunately it's only used in some very old scenarios.


sgmunn:
@mkrueger can you also create a PR for 15-6 of this too, the bug milestone is 15-6 so presumably we will need it there as well

PR #3562, opened 2018-01-04
Fixed '530439 - Blank space in C# tooltips without summary'

mkrueger:


main/src/addins/CSharpBinding/MonoDevelop.CSharp.Tooltips/QuickInfoProvider.cs
```diff
@@ -86,7 +87,7 @@ static class QuickInfoProvider
 			var documentation = symbol.GetDocumentationParts (model, editor.CaretOffset, formatter, cancellationToken);
 			sb.Append ("<span font='" + FontService.SansFontName + "' size='small'>");
 
-			if (documentation != null) {
+			if (documentation != null && documentation.Any ()) {
```
Therzok: This creates the documentation parts eagerly twice. Consider creating an immutablearray out of it.

mkrueger: There is nothing eager in here - the current implementation returns list objects. If I manually make an immutablearray it would create more memory pressure. If I make a list out of it I would do the same if the implementation changes to an immutable list.

Therzok: Ah, yes, you're right!

mkrueger: I don't think that it'll ever be a basic enumerable -doesn't make much sense implementation wise :)

PR #3568, opened 2018-01-04
[NUnit] Properly enumerate project assembly reference dependencies. 

garuma:
This PR changes the way dependencies of a test project are computed by relying on the MSBuild machinery instead of being limited to only the project model.

This fixes cases where assembly references are computed on the fly e.g. when custom `ReferencePath` are specified in the project so that `<Reference />` with no hint path are still resolved (something the old system would mistakenly think of as GAC assemblies).

The commits changes the API for gathering dependencies to be asynchronous and applies the same sort of filtering as before to only load dependencies that wouldn't be otherwise loaded.

An exception here is that facades are pulled in by the underlying call to `GetReferences` without a way to differentiate them from other references, that causes assembly load errors down the line which this commit works around that by wrapping the call in `InitSupportAssemblies` in a try/catch.

NB: The scenario that called for this fix is a MonoDevelop addin testsuite which has very specific considerations when it comes to assembly loading. Since that test-suite already uses a custom console runner (via vstool\mdtool) to execute unit-tests maybe it would be a good idea in general to also allow delegation of `GetTestInfo` to a custom runner as well.

### DIFF
--- a/main/src/addins/MonoDevelop.UnitTesting.NUnit/MonoDevelop.UnitTesting.NUnit/NUnitProjectTestSuite.cs
+++ b/main/src/addins/MonoDevelop.UnitTesting.NUnit/MonoDevelop.UnitTesting.NUnit/NUnitProjectTestSuite.cs
@@ -205,21 +205,26 @@ namespace MonoDevelop.UnitTesting.NUnit
 		protected override string TestInfoCachePath {
 			get { return Path.Combine (resultsPath, storeId + ".test-cache"); }
 		}
-		
-		protected override IEnumerable<string> SupportAssemblies {
-			get {
+
+		protected override async Task<IEnumerable<string>> GetSupportAssembliesAsync ()
+		{
+			DotNetProject project = base.OwnerSolutionItem as DotNetProject;
+
+			if (project != null) {
+				var references = await project.GetReferences (IdeApp.Workspace.ActiveConfiguration).ConfigureAwait (false);
 				// Referenced assemblies which are not in the gac and which are not localy copied have to be preloaded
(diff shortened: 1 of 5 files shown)


main/src/addins/MonoDevelop.UnitTesting.NUnit/MonoDevelop.UnitTesting.NUnit/NUnitProjectTestSuite.cs
```diff
@@ -205,21 +205,26 @@ public override void GetCustomConsoleRunner (out string command, out string args
...
-				DotNetProject project = base.OwnerSolutionItem as DotNetProject;
-				if (project != null) {
-					foreach (var pr in project.References) {
-						if (pr.ReferenceType != ReferenceType.Package && !pr.LocalCopy && pr.ReferenceOutputAssembly) {
-							foreach (string file in pr.GetReferencedFileNames (IdeApp.Workspace.ActiveConfiguration))
-								yield return file;
-						}
-					}
-				}
+				var supportAssemblies = references.Where (r => !r.IsCopyLocal && (!r.IsProjectReference || r.ReferenceOutputAssembly) && !r.IsFrameworkFile && !r.IsImplicit && !IsGacReference (r))
```
Therzok: Maybe do not eagerly `ToList` here, as you're already creating a new list from this result in `RunAsyncLoadTest`?

garuma: You right, it's probably fine there.

PR #3570, opened 2018-01-05
Fix a first-chance exception in ConditionRelationalExpression

KirillOsenkov:


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild.Conditions/ConditionRelationalExpression.cs
```diff
@@ -144,6 +144,11 @@ public override bool CanEvaluateToString (IExpressionContext context)
 				return string.Equals (l, r, StringComparison.OrdinalIgnoreCase);
 			case RelationOperator.NotEqual:
 				return !string.Equals (l, r, StringComparison.OrdinalIgnoreCase);
+			case RelationOperator.GreaterOrEqual:
+			case RelationOperator.LessOrEqual:
+				if (string.Equals (l, r, StringComparison.OrdinalIgnoreCase))
```
Therzok: This changes evaluation behaviour:
https://github.com/Microsoft/msbuild/blob/master/src/Build/Evaluation/Conditionals/GreaterThanOrEqualExpressionNode.cs#L16

See how `EqualExpressionNode` has an override for string:
https://github.com/Microsoft/msbuild/blob/master/src/Build/Evaluation/Conditionals/EqualExpressionNode.cs#L37

I think the actual fix is detecting whether a fix is a `Version` then doing comparison there. (I suppose that's where it's failing to run, and string comparison won't do version comparison correctly, as `10 < 2`)

Therzok:
To expand, what I think is happening here is that `System.Version` is being used in msbuild conditions.

I suppose the targets that are failing are checking exact equality between versions.

But, given this expression

```
var v1 = Version.Parse("2.0.0.0");
var v2 = Version.Parse("2.00.0.0");
 
Console.WriteLine(v1 <= v2); // yields true
```

I suppose we'll end up not evaluating properly. The same issue exists for the `Equals` case. So, I guess we should be expanding the fix to handle `Version` in MSBuild conditions and act accordingly. Node comparison only works for versions, not generic strings, so that should cover all the cases and be compatible with MSBuild evaluation

KirillOsenkov:
I see what you mean now. Turns out MSBuild has specific support for numeric and Version literals:
http://source.dot.net/#Microsoft.Build/Evaluation/Conditionals/NumericComparisonExpressionNode.cs,42

Makes sense now. I will investigate. I didn't know they have actual support for Versions.

KirillOsenkov:
And this is the file that we're evaluating:
https://github.com/dotnet/sdk/blame/master/src/Tasks/Microsoft.NET.Build.Extensions.Tasks/msbuildExtensions/Microsoft/Microsoft.NET.Build.Extensions/Microsoft.NET.Build.Extensions.NETFramework.targets

KirillOsenkov:
This has been superseded by https://github.com/mono/monodevelop/pull/3592/

PR #3572, opened 2018-01-05
Avoid multiple first-chance exceptions when registering packages

KirillOsenkov:
Introduce SystemAssemblyService.IsManagedAssembly() (same as in Mono.Addins). Avoid throwing a first-chance exception when trying to read native .dlls.


KirillOsenkov:
Mono Addins already has an IsManagedAssembly() helper, but it's internal. I wasn't sure if it's best to reuse that one, so I just copied. We do need one in MonoDevelop and SystemAssemblyService felt like a good place for it. I plan to use it in a few more places later.

PR #3574, opened 2018-01-05
Make HelpService lazy initialized.

KirillOsenkov:
I think it's better to remove it from the startup path entirely because in 99% of cases we pay for it even though we won't need it. An on-demand slow down is probably better. This also removes another first-chance exception from the startup path.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/HelpService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/HelpService.cs
@@ -43,21 +43,6 @@ namespace MonoDevelop.Projects
 		static object helpTreeLock = new object ();
 		static HashSet<string> sources = new HashSet<string> ();
 		
-		/// <summary>
-		/// Starts loading the MonoDoc tree in the background.
-		/// </summary>
-		public static void AsyncInitialize ()
-		{
-			lock (helpTreeLock) {
-				if (helpTreeInitialized)
-					return;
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/HelpService.cs
```diff
@@ -43,21 +43,6 @@ public static class HelpService
 		static object helpTreeLock = new object ();
 		static HashSet<string> sources = new HashSet<string> ();
 		
-		/// <summary>
-		/// Starts loading the MonoDoc tree in the background.
-		/// </summary>
-		public static void AsyncInitialize ()
```
Therzok: This is public API. We can't remove it.

mhutch:
Do we really not use it? I though it was used in most cases.

KirillOsenkov:
I see, makes sense. I didn't know how much it's used, was just guessing. Good point about the API. I think I'll abandon this PR and just fix the underlying file not found in a less invasive way.

Therzok:
Delay-loading monodoc is a good fix though, so let's have an issue tracking it if we don't fix it now.

KirillOsenkov:
Abandoning this in favor of https://github.com/mono/monodevelop/pull/3580

KirillOsenkov:
Filed https://github.com/mono/monodevelop/issues/3581 to track.

PR #3579, opened 2018-01-05
Avoid a first-chance exception in WindowsPlatform ImageHelper.

KirillOsenkov:
This should be a good safe tactical fix (WindowsPlatform only), no Xwt changes needed.

### DIFF
--- a/main/src/addins/WindowsPlatform/WindowsPlatform/ImageHelper.cs
+++ b/main/src/addins/WindowsPlatform/WindowsPlatform/ImageHelper.cs
@@ -46,8 +46,10 @@ namespace WindowsPlatform
 				var resourceName = stockId + ".png";
 
 				// first check if such a resource exists to avoid a first-chance exception
-				if (typeof (ImageHelper).Assembly.GetManifestResourceStream (resourceName) != null) {
-					image = Image.FromResource (typeof (ImageHelper), resourceName);
+				using (var stream = typeof (ImageHelper).Assembly.GetManifestResourceStream (resourceName)) {
+					if (stream != null) {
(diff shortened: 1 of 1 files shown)


KirillOsenkov:
The exception happens 75 times during startup.

main/src/addins/WindowsPlatform/WindowsPlatform/ImageHelper.cs
```diff
@@ -43,9 +43,14 @@ public static Image GetStockIcon (this IconId stockId)
 				return null;
 			Image image;
 			if (!cachedIcons.TryGetValue (stockId, out image)) {
-				try {
-					image = Image.FromResource (typeof (ImageHelper), stockId + ".png");
-				} catch (InvalidOperationException) {
+				var resourceName = stockId + ".png";
+
+				// first check if such a resource exists to avoid a first-chance exception
+				if (typeof (ImageHelper).Assembly.GetManifestResourceStream (resourceName) != null) {
```
Therzok: Do we not need to dispose the stream?

KirillOsenkov: I guess so... it wouldn't hurt. Just added a commit that disposes the stream.

KirillOsenkov: Dispose will run this line:
http://referencesource.microsoft.com/#mscorlib/system/io/unmanagedmemorystream.cs,277

KirillOsenkov:
How about this PR? Can this go in?

PR #3580, opened 2018-01-06
Avoid loading Monodoc on Windows

KirillOsenkov:
Since we know it's going to fail anyway, do not attempt to load Monodoc. This prevents a first-chance exception in the startup path.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/HelpService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/HelpService.cs
@@ -65,7 +65,15 @@ namespace MonoDevelop.Projects
 			lock (helpTreeLock) {
 				if (helpTreeInitialized)
 					return;
-				
+
+				// Only attempt on Windows if we can find monodoc.xml (currently not the case).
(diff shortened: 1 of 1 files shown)


mhutch:
This doesn't avoid loading it AFAICT, it just defers the error until something accesses the HelpService.

KirillOsenkov:
Agreed, but I think it's an acceptable fix for our goal. I just want to get the exception off the startup path on Windows so it doesn't impact our productivity.

If there are strong objections, I can investigate a better fix of course, but honestly this is not the area where I'd like to be spending my time.

KirillOsenkov:
How do we feel about this? Can I get this in?

mhutch:
Look ok to me. Though the exception will just be delayed until you hit a tooltip. I'd suggest moving the check inside HelpService.InitializeHelpTree

KirillOsenkov:
You're right, this is a better and cleaner fix. 

I think I've only resisted it because it was Friday and I was tired and didn't want to do any more work ;)

KirillOsenkov:
Rebased for a cleaner history.

PR #3583, opened 2018-01-06
[Perf] Optimize startup to not hit XmlSerializer path via LazyDeseria…

Therzok:
…lizer

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Properties.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Properties.cs
@@ -206,14 +206,16 @@ namespace MonoDevelop.Core
 				writer.WriteStartElement (PropertyNode);
 				writer.WriteAttributeString (KeyAttribute, property.Key);
 				
-				if (property.Value is LazyXmlDeserializer) {
-					writer.WriteRaw (((LazyXmlDeserializer)property.Value).Xml);
-				} else if (property.Value is ICustomXmlSerializer) {
-					((ICustomXmlSerializer)property.Value).WriteTo (writer);
+				if (property.Value is LazyXmlDeserializer deserializer) {
+					writer.WriteRaw (deserializer.Xml);
+				} else if (property.Value is ICustomXmlSerializer customXmlSerializer) {
+					customXmlSerializer.WriteTo (writer);
 				} else {
 					if (!(property.Value is string) && property.Value.GetType ().IsClass) {
-						XmlSerializer serializer = new XmlSerializer (property.Value.GetType ());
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/SystemAssemblyService.cs
```diff
@@ -421,7 +421,7 @@ public TargetFrameworkMoniker GetTargetFrameworkForAssembly (TargetRuntime tr, s
 		void SaveUserAssemblyContext ()
 		{
 			List<string> list = new List<string> (userAssemblyContext.Directories);
-			PropertyService.Set ("MonoDevelop.Core.Assemblies.UserAssemblyContext", list);
```
alanmcgovern: Would it make sense to build this behaviour into the property service?

Therzok: So, I guess the check would be:
```
if (value is ICollection collection && collection.Count == 0)
    value = null;
```
?

Therzok: But yes, good idea.

alanmcgovern: yeah, something like that. If we do make this change we should also consider whether or not the deserializer should return an empty collection or a null value when the key is unset.

KirillOsenkov:
I've been also thinking about optimizing the Loading side with this (in Properties.cs):

```
			public T Deserialize<T> ()
			{
				try {
					// optimize for a common case to avoid trying to load mscorlib.XmlSerializers.xml
					// for an empty List<string>
					if (typeof(T) == typeof(List<string>) &&
						xml != null &&
						xml.Contains(@"<ArrayOfString xmlns:xsi=""http://www.w3.org/2001/XMLSchema-instance"" xmlns:xsd=""http://www.w3.org/2001/XMLSchema"" />")) {
						return (T)(object)new List<string> ();
					}
```

It takes care of both cases where we try to read List<string>.

alanmcgovern:
I don’t think we need to hardcode stuff like that into the loading side if we just fix the saving side. When those files are written to disk they just won’t have an empty array anymore, so the loader logic would become dead/unused code nearly instantly.

Therzok:
@alanmcgovern @KirillOsenkov @mrward PTAL

KirillOsenkov:
Thanks!

PR #3588, opened 2018-01-08
[linux] Add bigger icons for Linux

directhex:
Not every window manager can use the SVG directly for all cases


directhex:
This is in an effort to fix 
screenshot from 2018-01-09 10-09-10 looking like ass

PR #3592, opened 2018-01-09
Add support for comparing Versions in ConditionExpression

KirillOsenkov:
MSBuild has intrinsic support for comparing double and Version values. See http://source.dot.net/#Microsoft.Build/Evaluation/Conditionals/NumericComparisonExpressionNode.cs

Example file:
https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Extensions.Tasks/msbuildExtensions/Microsoft/Microsoft.NET.Build.Extensions/Microsoft.NET.Build.Extensions.NETFramework.targets#L18

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild.Conditions/ConditionExpression.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild.Conditions/ConditionExpression.cs
@@ -30,14 +30,16 @@ using System.Xml;
 
 namespace MonoDevelop.Projects.MSBuild.Conditions {
 	internal abstract class ConditionExpression {
-	
+
 		public abstract bool BoolEvaluate (IExpressionContext context);
 		public abstract float NumberEvaluate (IExpressionContext context);
 		public abstract string StringEvaluate (IExpressionContext context);
+		public virtual Version VersionEvaluate (IExpressionContext context) => throw new NotSupportedException();
 		
 		public abstract bool CanEvaluateToBool (IExpressionContext context);
 		public abstract bool CanEvaluateToNumber (IExpressionContext context);
 		public abstract bool CanEvaluateToString (IExpressionContext context);
+		public virtual bool CanEvaluateToVersion (IExpressionContext context) => false;
 
 		public virtual void CollectConditionProperties (ConditionedPropertyCollection properties)
 		{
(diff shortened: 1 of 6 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild.Conditions/ConditionExpression.cs
```diff
@@ -30,14 +30,18 @@
...
 namespace MonoDevelop.Projects.MSBuild.Conditions {
 	internal abstract class ConditionExpression {
-	
+
+		static readonly Version EmptyVersion = new Version ();
+
 		public abstract bool BoolEvaluate (IExpressionContext context);
 		public abstract float NumberEvaluate (IExpressionContext context);
 		public abstract string StringEvaluate (IExpressionContext context);
+		public virtual Version VersionEvaluate (IExpressionContext context) => EmptyVersion;
```
Therzok: It's internal, so we can make it abstract.

Therzok:
Can you rebase on top of latest master? I made some changes in the code in Relational Expression.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild.Conditions/ConditionExpression.cs
KirillOsenkov: Making it abstract would require adding 5 or 6 throwing implementations, and I didn't want to pollute all derived classes.

Therzok: Then this default should throw :)

KirillOsenkov:
Rebased and added a throw in the base method.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild.Conditions/ConditionFactorExpresion.cs
```diff
@@ -132,10 +123,19 @@ public override bool CanEvaluateToString (IExpressionContext context)
 			return true;
 		}
 
-		internal Conditions.Token Token {
-			get {
-				return this.token;
-			}
+		public override bool CanEvaluateToVersion (IExpressionContext context)
+		{
+			var text = StringEvaluate (context);
```
Therzok: We need to verify the token type is string.

Therzok:
LGTM.

On other note, the API around ConditionExpression should probably be changed a bit so we don't evaluate a token multiple times. We seem to do that a lot because of CanEvaluate -> Evaluate

slluis:
LGTM, but it needs unit tests.

KirillOsenkov:
Added tests and they pass. How's it looking now?

KirillOsenkov:
@Therzok valid comment about evaluating multiple times, I've noticed it too. 

Filed https://github.com/mono/monodevelop/issues/3617 to investigate.

PR #3593, opened 2018-01-09
[build] Update the .gitignore again

alanmcgovern:


alanmcgovern:
Can we just merge these when approved instead of requiring the extra round trip?

PR #3595, opened 2018-01-09
Fixed possible null reference in unit tests.

mkrueger:


mkrueger:
The QA team reported that exception - I think the issue is in one of the unit tests but checking for null here doesn't hurt.

PR #3611, opened 2018-01-10
[Tools] Improve toggle profiling tool to also output inverted call tree

Therzok:


Therzok:
Example output normal sample:
https://gist.github.com/Therzok/23b92a89d973153c5252bb96697c33bb

Side by side inverted sample:
https://gist.github.com/Therzok/317159b6eb6435a3b7fa2056018dc0b3

Therzok:
Need to fix something, it OOM'd on 500mb sample file.

Therzok:
The other alternative is to fix the JIT address to method name reporting so `filtercalltree` still works. If that's desired, we can drop this PR for that.

Therzok:
Closing, as we now have AOT.

PR #3612, opened 2018-01-10
[Ide] Moved FileRegistry into the core ide.

mkrueger:
It's a proposed fix for 524635 [Feedback] XAML files are not refreshed
after switching git branch.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DocumentRegistry.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DocumentRegistry.cs
@@ -1,0 +1,244 @@
+﻿//
+// DocumentRegistry.cs
+//
+// Author:
+//       Mike Krüger <mikkrg@microsoft.com>
+//
+// Copyright (c) 2018 Microsoft
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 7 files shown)


mkrueger:
A proposed fix for 
https://devdiv.visualstudio.com/DevDiv/Xamarin%20VS%20for%20Mac/_queries?id=524635&_a=edit&triage=true

However without the presenter implementation it would just blindly reload the .xaml files.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DocumentRegistry.cs
```diff
@@ -0,0 +1,244 @@
...
+			public DocumentInfo (Document doc)
+			{
+				this.Document = doc;
+				LastSaveTimeUtc = DateTime.UtcNow;
+				doc.Saved += Doc_Saved;
+				doc.Reloaded += Doc_Saved;
+			}
+
+
+			void Doc_Saved (object sender, EventArgs e)
```
Therzok: Can be static, `Document` is passed in as `sender`. Then, DocumentInfo does not need to be `IDisposable`

Therzok: And `DocumentInfo` does not appear to be disposed anywhere for that matter.

PR #3616, opened 2018-01-10
Editorconfig

mkrueger:
See https://devdiv.visualstudio.com/DevDiv/_workitems?id=515047&_a=edit


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileDescriptionTemplate.cs
```diff
@@ -73,7 +74,7 @@ static void OnExtensionChanged (object s, ExtensionNodeEventArgs args)
 		public abstract string Name { get; }
 		
 		public abstract void Load (XmlElement filenode, FilePath baseDirectory);
-		public abstract bool AddToProject (SolutionFolderItem policyParent, Project project, string language, string directory, string name);
+		public abstract Task<bool> AddToProject (SolutionFolderItem policyParent, Project project, string language, string directory, string name);
```
slluis: Public api break, not good.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/SingleFileDescriptionTemplate.cs
```diff
@@ -281,7 +282,7 @@ protected virtual string ProcessContent (string content, IStringTagModel tags)
 
 		// Returns a stream with the content of the file.
 		// project and language parameters are optional
-		public virtual Stream CreateFileContent (SolutionFolderItem policyParent, Project project, string language, string fileName, string identifier)
+		public virtual async Task<Stream> CreateFileContent (SolutionFolderItem policyParent, Project project, string language, string fileName, string identifier)
```
slluis: More public api breakage

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileDescriptionTemplate.cs
mkrueger: There is no way around it in my opinion.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/SingleFileDescriptionTemplate.cs
mkrueger: But needed :/

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileDescriptionTemplate.cs
slluis: Yes, there is a way around. Instead of modifying the AddToProject method, obsolete it and add a new AddToProjectAsync method that calls the old AddToProject.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/SingleFileDescriptionTemplate.cs
slluis: Use same workaround as above.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileDescriptionTemplate.cs
mkrueger: that doesn't work because I need the async for the implementation of the old AddToProject.

It's not an API change because I want to change a API it's an API change because I need the async call for not blocking the UI.

I've AsyncImpl. And NonAsyncCall

So I can do:

bool AddToProject 
Task<bool> AddToProjectAsync   -> Calling AddToProject

But where should I put the Async call in ? It can't be in bool AddToProject because I need an async call.

So all I can do is:

bool AddToProject -> empty by default
Task<bool> AddToProjectAsync   -> Async call

And call both methods in the consumer - but that's something I don't really like for obvious reasons. The question is : is there anything pluggin into that which is not our code ?

slluis: The consumer (the template engine) will only call `AddToProjectAsync()`. The default implementation of `AddToProjectAsync` calls `AddToProject`, which can be converted to virtual and and throw a NIE.

Existing template subclasses will keep working without any change because the template engine will invoke `AddToProjectAsync()`, and that will invoke `AddToProject`, so it will just work like now.

Template subclasses that what to take advantage of async can just override `AddToProjectAsync()` and ignore `AddToProject`.

main/src/addins/MonoDevelop.Packaging/MonoDevelop.Packaging/AddPlatformImplementationViewModel.cs
```diff
@@ -203,6 +203,7 @@ void AddAssemblyInfoFile (Project project)
 
 			fileTemplate.SetProjectTagModel(parameters);
 			fileTemplate.AddToProject (project.ParentFolder, project, "C#", project.BaseDirectory, "");
+			fileTemplate.AddToProjectAsync (project.ParentFolder, project, "C#", project.BaseDirectory, "");
```
slluis: AddToProjectAsync should be calling AddToProject

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileDescriptionTemplate.cs
```diff
@@ -73,7 +74,14 @@ static void OnExtensionChanged (object s, ExtensionNodeEventArgs args)
...
 		public abstract void Load (XmlElement filenode, FilePath baseDirectory);
-		public abstract bool AddToProject (SolutionFolderItem policyParent, Project project, string language, string directory, string name);
+
+		[Obsolete("Use public abstract Task<bool> AddToProjectAsync (SolutionFolderItem policyParent, Project project, string language, string directory, string name)")]
+		public virtual bool AddToProject (SolutionFolderItem policyParent, Project project, string language, string directory, string name)
+		{
+			return false;
+		}
+
+		public abstract Task<bool> AddToProjectAsync (SolutionFolderItem policyParent, Project project, string language, string directory, string name);
```
slluis: Introducing an abstract member is an API break. Please do what I described. AddToProjectAsync should be a virtual method that invokes AddToProject by default. Make AddToProject virtual and throw a NIE.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileTemplate.cs
```diff
@@ -303,7 +304,7 @@ public virtual bool CanCreateUnsavedFiles (FileDescriptionTemplate newfile, Solu
 			}
 		}
 
-		protected virtual bool CreateFile (FileDescriptionTemplate newfile, SolutionFolderItem policyParent, Project project, string directory, string language, string name)
+		protected virtual async Task<bool> CreateFile (FileDescriptionTemplate newfile, SolutionFolderItem policyParent, Project project, string directory, string language, string name)
```
slluis: API break.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileTemplate.cs
```diff
@@ -244,13 +245,13 @@ internal static FileTemplate GetFileTemplateByID (string templateID)
 			return node == null ? null : LoadTemplate (node);
 		}
 
-		public virtual bool Create (SolutionFolderItem policyParent, Project project, string directory, string language, string name)
+		public virtual async Task<bool> Create (SolutionFolderItem policyParent, Project project, string directory, string language, string name)
```
slluis: API break.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/ProjectDescriptor.cs
```diff
@@ -193,6 +193,7 @@ public void InitializeItem (SolutionFolderItem policyParent, ProjectCreateInform
 						continue;
 					fileTemplate.SetProjectTagModel (projectCreateInformation.Parameters);
 					fileTemplate.AddToProject (policyParent, project, defaultLanguage, project.BaseDirectory, null);
+					fileTemplate.AddToProjectAsync (policyParent, project, defaultLanguage, project.BaseDirectory, null);
```
slluis: No need to call both AddToProject and AddToProjectAsync

slluis: Also, the async call should be awaited.

main/src/addins/MonoDevelop.Packaging/MonoDevelop.Packaging/AddPlatformImplementationViewModel.cs
mkrueger: Why ? That would break things.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileTemplate.cs
mkrueger: It's not in a public class

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileTemplate.cs
mkrueger: It's not in a public class

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/EditorConfigService.cs
```diff
@@ -0,0 +1,121 @@
...
+		public async static Task<ICodingConventionContext> GetEditorConfigContext (string fileName, CancellationToken token = default (CancellationToken))
+		{
+			if (contextCache.TryGetValue (fileName, out ICodingConventionContext result))
+				return result;
+			result = await codingConventionsManager.GetConventionContextAsync (fileName, token);
+			if (result == null)
+				return null;
+			lock (contextCacheLock) {
+				if (contextCache.ContainsKey (fileName))
+					return contextCache [fileName];
```
slluis: Use `TryGetValue` to avoid double access.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/EditorConfigService.cs
```diff
@@ -0,0 +1,121 @@
...
+				watcher.EnableRaisingEvents = true;
+				lock (watchers) {
+					watchers.Add (directoryPath + Path.DirectorySeparatorChar.ToString () + fileName, watcher);
+				}
+			}
+
+			public void StopWatching (string fileName, string directoryPath)
+			{
+				lock (watchers) {
+					watchers.Remove (directoryPath + Path.DirectorySeparatorChar.ToString () + fileName);
```
slluis: The removed watcher should be disposed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.EditorConfig/EditorConfigDocumentOptionsProvider.cs
```diff
@@ -0,0 +1,251 @@
...
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Diagnostics.Analyzers.NamingStyles;
+using Microsoft.CodeAnalysis.ErrorLogger;
+using Microsoft.CodeAnalysis.Options;
+using Microsoft.CodeAnalysis.Shared.Utilities;
+// using Microsoft.VisualStudio.CodingConventions;
+
+namespace Microsoft.CodeAnalysis.Editor.Options
+{
+	// NOTE: this type depends on Dev15 assemblies, which is why the type is in EditorFeatures.Next. But, that library
```
slluis: Can this file be excluded? it is all commented out.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileTemplate.cs
slluis: Fair enough!

slluis:
The build is failing.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.OptionProvider/CSharpDocumentOptionsProvider.cs
```diff
@@ -0,0 +1,127 @@
...
+			}
+			var path = GetPath (document);
+			ICodingConventionContext conventions = null;
+			try {
+				if (path != null)
+					conventions = await EditorConfigService.GetEditorConfigContext (path, cancellationToken);
+			} catch (Exception e) {
+				LoggingService.LogError("Error while loading coding conventions.", e);
+			}
+			return new DocumentOptions(policy.CreateOptions (textpolicy), conventions.CurrentConventions);
```
slluis: If conventions can't be loaded it will be null and this will NRE.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/EditorConfigService.cs
mkrueger: y you're right but it's unlikely that tihs will ever be hit - but I change it

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/DefaultSourceEditorOptions.cs
```diff
@@ -474,6 +484,8 @@ public ITextEditorOptions WithTextStyle (MonoDevelop.Ide.Gui.Content.TextStylePo
 		bool tabsToSpaces = true;
 		public bool TabsToSpaces {
 			get {
+				if (context != null && context.CurrentConventions.UniversalConventions.TryGetIndentStyle (out Microsoft.VisualStudio.CodingConventions.IndentStyle result))
+					return result == Microsoft.VisualStudio.CodingConventions.IndentStyle.Spaces;
```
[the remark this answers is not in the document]
mkrueger: It's basically just a dictionary lookup what they do there - but y these can be cached now the context changed event is working - it didn't use to work in former versions :(

slluis:
We need unit tests for this.

mohakbarokar:
This PR is tested on Build https://code.internalx.com/Wrench/BuildDetails?buildId=40201e1c7c6578ea32d91eb83a5978d 

Build Config: https://gist.github.com/mohakbarokar/94108e2d85fc337996a3a18d791cdac9

Observation: "dotnet_style_predefined_type_for_locals_parameters_members = true:error" rule is not working in this build for editorconfig

We can merge this PR as it is verified.

slluis:
Still no luck with the build.

KirillOsenkov:
Sorry that I haven't been following this closely, but maybe we could file a Roslyn bug to move http://source.roslyn.io/#Microsoft.CodeAnalysis.EditorFeatures.Wpf/Options/EditorConfigDocumentOptionsProviderFactory.cs down to EditorFeatures, and then just use that?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.csproj
```diff
@@ -104,15 +104,19 @@
   <ItemGroup>
     <Reference Include="atk-sharp, Version=2.12.0.0, Culture=neutral, PublicKeyToken=35e10195dab3c99f">
       <SpecificVersion>False</SpecificVersion>
+      <Package>gtk-sharp-2.0</Package>
```
KirillOsenkov: What is the Package metadata for?

mkrueger: monodevelop insists to inject that.

mkrueger:
Just because it has the same name it doesn't mean we can use it - we need to bridge our policy system there.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.csproj
Therzok: @KirillOsenkov it's pkg-config integration, it's injected so things work when doing tarballs for linux.

slluis:
QA approved, but let's try to reduce the package dependencies before merging.

main/src/addins/CSharpBinding/CSharpBinding.csproj
```diff
@@ -181,6 +181,24 @@
       <HintPath>..\..\..\build\bin\Microsoft.VisualStudio.CodingConventions.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <Reference Include="Microsoft.VisualStudio.CodingConventions">
+      <HintPath>..\..\..\build\bin\Microsoft.VisualStudio.CodingConventions.dll</HintPath>
+    </Reference>
+    <Reference Include="Microsoft.VisualStudio.CoreUtility">
```
KirillOsenkov: @mkrueger The 5 editor references are not necessary since importing ReferencesVSEditor at the top of this file takes care of that.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="4.0">
+  <Import Project="..\..\..\packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.props" Condition="Exists('..\..\..\packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.props')" />
```
KirillOsenkov: What is this necessary for?

main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
```diff
@@ -241,6 +242,50 @@
...
+    <Reference Include="Microsoft.Win32.Primitives">
+      <HintPath>..\..\..\packages\Microsoft.Win32.Primitives.4.0.1\lib\net46\Microsoft.Win32.Primitives.dll</HintPath>
+    </Reference>
+    <Reference Include="Microsoft.VisualStudio.Text.Logic">
+      <HintPath>..\..\..\packages\Microsoft.VisualStudio.Text.Logic.15.6.241-preview\lib\net46\Microsoft.VisualStudio.Text.Logic.dll</HintPath>
+    </Reference>
+    <Reference Include="Microsoft.VisualStudio.Language.StandardClassification">
+      <HintPath>..\..\..\packages\Microsoft.VisualStudio.Language.StandardClassification.15.6.241-preview\lib\net46\Microsoft.VisualStudio.Language.StandardClassification.dll</HintPath>
+    </Reference>
+    <Reference Include="Microsoft.VisualStudio.Text.UI">
```
KirillOsenkov: Don't add these.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.csproj
```diff
@@ -201,62 +216,24 @@
...
     </Reference>
-    <Reference Include="Minimatch">
-      <HintPath>..\..\..\packages\Minimatch.2.0.0\lib\netstandard1.0\Minimatch.dll</HintPath>
+    <Reference Include="Microsoft.VisualStudio.Text.Logic">
+      <HintPath>..\..\..\build\bin\Microsoft.VisualStudio.Text.Logic.dll</HintPath>
     </Reference>
-    <Reference Include="Microsoft.VisualStudio.CodingConventions">
-      <HintPath>..\..\..\packages\Microsoft.VisualStudio.CodingConventions.1.1.20171211.2\lib\netstandard1.3\Microsoft.VisualStudio.CodingConventions.dll</HintPath>
-      <Private>true</Private>
+    <Reference Include="Microsoft.VisualStudio.Text.UI">
```
KirillOsenkov: Don't add this.

KirillOsenkov:
Please go through and make sure you're not manually adding references that are already mentioned in https://github.com/mono/monodevelop/blob/master/main/msbuild/ReferencesVSEditor.props. The reason that file exists is that we can often update VS editor dependencies in a single place.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
mkrueger: NuGet did that

main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
mkrueger: NuGet did that as well…

mkrueger:
I didn't add these - as said the nuget package did that - CodingConventions nuget is very bad regarding to references  - the problem with that is that the CodingConventions dll doesn't use many  - at least none that would require such huge list.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
Therzok: What Kirill meant is that we already import them:
https://github.com/mono/monodevelop/blob/master/main/msbuild/ReferencesVSEditor.props#L20

main/src/addins/CSharpBinding/CSharpBinding.csproj
```diff
@@ -100,7 +100,8 @@
       <Private>False</Private>
     </ProjectReference>
     <ProjectReference Include="..\MonoDevelop.UnitTesting\MonoDevelop.UnitTesting.csproj">
-      <Project>{A7A4246D-CEC4-42DF-A3C1-C31B9F51C4EC}</Project>
+      <Project>{A7A4246D-CEC4-42DF-A3C1-C31B9F51C4EC}</Project>./main/src/addins/CSharpBinding/CSharpBinding.csproj
```
Therzok: Uhm?

PR #3618, opened 2018-01-11
Update Roslyn dependencies.

KirillOsenkov:
Update VS Editor to 15.6.161-preview since that's the version referenced by Roslyn.
Remove a reference to Text.Internal.
Add a reference to Humanizer.Core 2.2.0. It is now a prerequisite.


slluis:
I guess we need this in the d15-6 branch, right?

KirillOsenkov:
Right, I will port this shortly.

PR #3621, opened 2018-01-11
[Refactoring] Avoid double hashset lookup

Therzok:


KirillOsenkov:
You mentioned the same code is in Roslyn? Shall we submit a PR to fix it in Roslyn as well?

Therzok:
No, this code does not exist in roslyn.

PR #3626, opened 2018-01-11
[Toolbar][Mac] Fix various configuration selector issues

sevoku:
This PR fixes a couple of issues with the configuration selector (Toolbar), icluding weird behaviour on HighSierra.

* [53733](https://bugzilla.xamarin.com/show_bug.cgi?id=53733): broken Tab-Focus
* [60722](https://bugzilla.xamarin.com/show_bug.cgi?id=60722): wrong hit area calculation
* [59751](https://bugzilla.xamarin.com/show_bug.cgi?id=59751): Layout/Rendering issues on HighSierra

~~TODO: the SelectorView is still not really accessible, it is possible to navigate between the cells with the Tab key,  but VoiceOver sees only one big area without any possible action.~~ done in 
8827e7a

### DIFF
--- a/main/src/addins/MacPlatform/MainToolbar/SelectorView.cs
+++ b/main/src/addins/MacPlatform/MainToolbar/SelectorView.cs
@@ -128,6 +128,18 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


iainx:
Looks good to me, and the accessibility works ok

PR #3627, opened 2018-01-11
[Core] Optimize SynchronizationContext users

Therzok:
Most notably, optimize the ProgressMonitoring stack. The old code would create lambdas with captures for every progress monitor update. This was optimized to use a cached static delegate and no capture lambdas with a passed state parameter.


Therzok:
ping

Therzok:
@BrzVlad there's a GC failure in the VS for Mac VSTS build failure

Therzok:
For reference: failed build, before rerunning it https://devdiv.visualstudio.com/DevDiv/_build/index?buildId=1281109

Therzok:
Fixed the test failure, handed off the GC crasher to @BrzVlad. It seems to have happened when allocating memory in a tight loop (such as, caused by the old code).

Therzok:
Re-review please.

PR #3643, opened 2018-01-15
Static registrar support

Therzok:


Therzok:
These changes should not impact anything. Can be approved and merged at any time.

PR #3644, opened 2018-01-15
Fixed 'Bug 61137 (`) - Cannot add Reference ( Microsoft.CodeAnalysis

mkrueger:
error)'

Deactivated until UI is implemented.


Therzok:
We need to track this somewhere, that these UIs need to be implemented before the next release. cc @slluis

mkrueger:
I make a card for each of the missing UIs.

PR #3646, opened 2018-01-15
[61102][Mac] Fix open project dialog on High Sierra

sevoku:
Changing the layout of the NSSavePanel AccessoryView does not work on High Sierra, which seems to always enforce initial constraints resulting in a wrong layout:

grafik

With this patch the accessory view layout remains always the same and it just enables/disables the "Close current workspace" checkbox:

grafik
grafik

(fixes bug #61102)


hbons:
- [ ] "Show files" -> "Show Files"

vancura:
Also, way too many spaces between _Encoding:_ and _Auto Detected_. Or is that column left-aligned? It should be right-aligned.

By any chance, have you checked what it does with more verbose languages like German? I am afraid what could happen if these labels are multiline.

sevoku:
ok, I've fixed the alignments:

grafik

and it works fine with German:

grafik

@vancura the labels are always single-line.

slluis:
Can this be merged?

PR #3648, opened 2018-01-15
[DotNetCore] Fix tests timing out on build server

mrward:
Ensure projects are disposed after running tests to prevent the file
watcher not being disposed and stopping the unit tests from finishing
on the build server.


mrward:
VSTS build does not run the .NET Core tests so that passing not verify the fix. Wrench build is green so going to merge this.

PR #3653, opened 2018-01-15
[Core] Remove Mono.TextEditor.Platform

Therzok:
This is now shipped via nuget as vs-editor-api


Therzok:
This is dead code that's not used anywhere. Should be ok to merge.

KirillOsenkov:
:+1: this was the right time to finally delete this. We've now tested vs-editor-api NuGet enough to know that we're not missing anything from the old checked-in source.

PR #3657, opened 2018-01-15
[A11y] Detect whether VoiceOver is being used or not

iainx:


main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -83,6 +83,56 @@ public class MacPlatformService : PlatformService
...
+
+		[DllImport("/usr/lib/libobjc.dylib")]
+		private static extern IntPtr imp_implementationWithBlock(ref BlockLiteral block);
+
+		[DllImport("/usr/lib/libobjc.dylib")]
+		private static extern void method_setImplementation(IntPtr method, IntPtr imp);
+
+		[MonoNativeFunctionWrapper]
+		delegate void AccessibilitySetValueForAttributeDelegate (IntPtr self, IntPtr selector, IntPtr valueHandle, IntPtr attributeHandle);
+		delegate void SwizzleDelegate (IntPtr block, IntPtr self, IntPtr valueHandle, IntPtr attributeHandle);
```
Therzok: Maybe a more specific name for this delegate? `SwizzleDelegate` makes it sound like it works for every method.

iainx: Good point.

PR #3658, opened 2018-01-16
[540610][Ide] Add Find Command command

### LINKED ISSUE #3479: No command to global search commands
Although the global search includes commands, there is no filter for commands, nor is there a command to open the search with this filter (like "Go to type" and "Go to file" for types and files).

It would also be good to bind this to *cmd-shift-p* for compat with VS Code.

> VS bug #540610

sevoku:
This PR adds a "Run Command" (VS Code binding: Ctrl+Shift+P, Mac: Cmd+Shift+P) command to search commands and shows the appropriate category in the global search menu.

grafik

(fixes bug #540610)
(closes #3479)

### DIFF
--- a/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
+++ b/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
@@ -895,6 +895,13 @@
 			macShortcut = "Alt|Meta|Shift|F"
 			_description = "Search for and replace text in all files of a directory" />
 
+	<Command id = "MonoDevelop.Ide.Commands.SearchCommands.RunCommand"
+			defaultHandler = "MonoDevelop.Components.MainToolbar.RunCommandHandler"
+			_label = "_Command..."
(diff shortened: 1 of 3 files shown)


sevoku:
@hbons it might make sense to add this to the main menu, but I'm not sure where. Maybe `Search` → `Go To` → `Command…`?

Therzok:
@sevoku I think we should maybe do a smarter implementation here.

We currently have SearchCategory as a base class, which allows you to define tags. Maybe, we should gather the tags from all the providers and filter them out somehow? Or extend the base API so each search category can extend the tags and generate commands automatically?

https://github.com/mono/monodevelop/blob/6c6deb0170e64673b8a2267c3d51ec075230088f/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/SearchCategory.cs

sevoku:
@Therzok definitely YES, but we need to spec it out more precisely (see #551666, maybe for 15.7). This PR just makes commands more discoverible and adds the VS Code binding.

PR #3663, opened 2018-01-16
Optimize project save

slluis:
Two optimizations: first, the default run configuration required for comparing
is generated only once and then cached. Generating the configuration is slow
because it needs a project evaluation.

Second, invalidation of the assembly and package references caches is now
instant. Instead of having to acquire a lock to clear the cache, there is now
a flag that can be set to signal that the cache is invalid and needs refresh.

Also added some logging.

This should fix VSTS issue #530382.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
@@ -986,16 +986,21 @@ namespace MonoDevelop.Projects
 		}
 
 		AsyncCriticalSection referenceCacheLock = new AsyncCriticalSection ();
-		Dictionary<string, List<AssemblyReference>> referenceCache = new Dictionary<string, List<AssemblyReference>> ();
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
```diff
@@ -986,16 +986,21 @@ public IEnumerable<DotNetProject> GetReferencedAssemblyProjects (ConfigurationSe
...
+		ImmutableDictionary<string, List<AssemblyReference>> referenceCache = ImmutableDictionary<string, List<AssemblyReference>>.Empty;
+		bool referenceCacheNeedsRefresh;
 
 		async Task<List<AssemblyReference>> RunResolveAssemblyReferencesTarget (ConfigurationSelector configuration)
 		{
 			List<AssemblyReference> refs = null;
 			var confId = (GetConfiguration (configuration) ?? DefaultConfiguration)?.Id ?? "";
 
+			// Check the cache before waiting for the lock, which may be very slow
+			if (!referenceCacheNeedsRefresh && referenceCache.TryGetValue (confId, out refs))
```
Therzok: Won't this race?

PR #3676, opened 2018-01-17
Fix bug #40942 - XS is leaving empty properties in the project file (15-6)

slluis:
A new flag is generated while evaluating properties: DefinedMultipleTimes.
This flag is used to determine if a property is defined only once in the
project file, or if it is defined more than once (or in other imported
project files). In the first case, property definitions that set the default
value can be safely removed, since we know that the property definition is not
overriding another definition.


parasAmbhore:
I can confirm the bug is not reproducible when using below builds.

Build Info:

Version 7.4 Preview (7.4 build 902)

Xamarin.Mac
Xamarin.iOS
Xamarin.Android

Gist Link: https://gist.github.com/GLjackyvaswani/b78a1873f508730f3d3fd2d6b49f576b

Screen-Cast Link: https://www.screencast.com/t/sDXKKHqP

PR #3678, opened 2018-01-17
[Ide] Use System.IO.Compression from NuGet, like .Core

rodrmoya:


rodrmoya:
This keeps failing in wrench, so out of ideas

rodrmoya:
yes, I think yours is safer, so I'd vote for merging yours and closing this PR

PR #3687, opened 2018-01-18
[SourceEditor] Fix file reload after save bug.

mkrueger:


alanmcgovern:
What bug was this fixing? Can we include the URL to the original bug in the commit message in future?  If i was a betting man i would say this caused a regression https://github.com/mono/monodevelop/issues/3687 , but i don't know what the original problem was :(

alanmcgovern:
Also it seems very related to our other ask - https://github.com/mono/monodevelop/issues/3929

mrward:
Not sure there was a bug report - I think it was just a case of dogfooding master and every time you saved the document it would scrollback to the top since it was being re-loading.

alanmcgovern:
Yeah, the bug i'm seeing is that my files are being reloaded after save :p

PR #3698, opened 2018-01-19
[Debugger] Can't edit function breakpoint

iainx:
Enable editing of breakpoint locations, either line, function or exception

Fixes VSTS #528267

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/BreakpointPropertiesDialog.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/BreakpointPropertiesDialog.cs
@@ -364,6 +364,12 @@ namespace MonoDevelop.Debugger
 			fb.ParamTypes = parsedParamTypes;
 		}
 
+		void SaveCatchpoint (Catchpoint cp)
+		{
+			cp.ExceptionName = entryExceptionType.Text;
+			cp.IncludeSubclasses = checkIncludeSubclass.Active;
+		}
+
 		class ParsedLocation
 		{
 			int line;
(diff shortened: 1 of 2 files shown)


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/BreakpointPropertiesDialog.cs
```diff
@@ -357,6 +357,12 @@ void SaveFunctionBreakpoint (FunctionBreakpoint fb)
 			fb.ParamTypes = parsedParamTypes;
 		}
 
+		void SaveCatchpoint (Catchpoint cp)
+		{
+			cp.ExceptionName = entryExceptionType.Text;
+			cp.IncludeSubclasses = checkIncludeSubclass.Active;
```
slluis: Will this work in the middle of a debugging session? will the breakpoing be re-bound?

iainx: Yes, it appears to in my tests.

DavidKarlas:
@iainx You probably did changes also in debugger-libs repo that you need to push/open PR for.

iainx:
@DavidKarlas yes, you're right, I didn't realise I was editing in there as well

PR filed: https://github.com/mono/debugger-libs/pull/175

Therzok:
I think this needs to be rebased on master, otherwise VSTs will just keep building this commit.

DavidKarlas:
@iainx I tried if re-bounding works, doesn't seem to work in my case:
1) Add function breakpoint to `MainClass.test1()`
2) Start debugging
3) `test1` method is hit
4) Edit function breakpoint and change it to `MainClass.test2()`
5) Continue execution
6) It should break when enters `test2` method, but it doesn't it continues execution and finished process
```
class MainClass
{
    {
        test1();
    }

    private static void test1()
    {
        test2();
    }

    private static void test2()
    {
        Console.WriteLine("test2");
    }
}
```

iainx:
@DavidKarlas Yes, you're right, I only tested moving the breakpoints when the code wasn't running, and looking at the soft debugger, the code to move breakpoints when the debugger is running hasn't been implemented, and it beyond my expertise at the moment (is it even supported in Mono?)

So, I've made it so that the breakpoints are only editable when the debugger session isn't running, which improves the situation slightly at least

ketttY:
@monojenkins rebase

ketttY:
@monojenkins rebase

ketttY:
@monojenkins rebase

ketttY:
@monojenkins rebase

sgmunn:
@monojenkins rebase

ketttY:
Current version:
EditFunctionBreakpoint1080

ketttY:
Test Case: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1052373

ketttY:
@monojenkins rebase

mrward:
Compat check error with Xamarin.Mac is known and unrelated to this change.

PR #3636, opened 2018-01-12
[Refactoring] Switch to Roslyn CodeFixService

Therzok:

PR #3699, opened 2018-01-19
Implemented '546162 - Generate overrides code fix ui'

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems?id=546162&_a=edit

Test instructions : 

* Place caret inside a new class:

class Test { 
  |
}

* Right click -> Quck Fix -> select  Generate overrides...
* Select overrides & hit ok -> the overriden members should be created
* Select cancel -> nothing should happen


Therzok:
https://github.com/mono/monodevelop/pull/3636/commits/f449a46075145d9833cfa3aad4e764795f37cab5

I implemented it here too.

Therzok:
Vaclav had some issues on the design: http://vncr.in/ovIf

Therzok:
Your version is definitely better, as it handles the pickmembers options. I'll remove the commit from my PR.

mkrueger:
btw. why did you implment it - that's what we've task assignments for.

Therzok:
@mkrueger was easy copy-paste, wanted to have it functional before handing off codefixes service to QA.

mkrueger:
https://files.slack.com/files-pri/T03C6CW9H-F8VNN57QQ/pasted_image_at_2018_01_22_10_43_am.png

mkrueger:
pasted image at 2018_01_22 10_43 am

mkrueger:
Not sure what's with the dialog buttons - the margins should be corrected by xwt automatically IMO.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.PickMembersService/PickMembersService.cs
```diff
@@ -0,0 +1,57 @@
...
+using MonoDevelop.Core;
+
+namespace MonoDevelop.Refactoring.PickMembersService
+{
+	[ExportWorkspaceService(typeof(IPickMembersService)), Shared]
+	class PickMembersService : IPickMembersService
+	{
+		PickMembersResult IPickMembersService.PickMembers(string title, ImmutableArray<ISymbol> members, ImmutableArray<PickMembersOption> options)
+		{
+			var dialog = new PickMembersDialog ();
```
sevoku: Wrap this whole thing in an `IdeApp.RunWhenIdle()` and a `Xwt.Toolkit.NativeEngine.Invoke` to make the dialog native and to make sure that you don't lock the UI.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.PickMembersService/PickMembersDialog.cs
```diff
@@ -0,0 +1,176 @@
...
+			Content = box;
+			Buttons.Add (new DialogButton (Command.Cancel));
+			Buttons.Add (okButton = new DialogButton (Command.Ok));
+			this.DefaultCommand = okButton.Command;
+
+			this.Width = 400;
+			this.Height = 321;
+			this.Resizable = false;
+
+			Show ();
```
sevoku: Don't call `Show()` in the constructor.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.PickMembersService/PickMembersService.cs
```diff
@@ -0,0 +1,57 @@
...
+			try {
+				dialog.Init (title, members, options);
+				bool performChange = dialog.Run () == Xwt.Command.Ok;
```
sevoku: shouldn't the dialog be transient? Using `dialog.Run (Xwt.MessageDialog.RootWindow)` will make it modal for the main IDE window.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.PickMembersService/PickMembersDialog.cs
```diff
@@ -0,0 +1,176 @@
...
+				}
+				UpdateOkButton ();
+			};
+
+			listViewPublicMembers.HeadersVisible = false;
+			listViewPublicMembers.DataSource = treeStore;
+			var checkBoxCellView = new CheckBoxCellView (symbolIncludedField);
+			checkBoxCellView.Editable = true;
+			checkBoxCellView.Toggled += delegate { UpdateOkButton (); };
+			listViewPublicMembers.Columns.Add ("", checkBoxCellView);
```
sevoku: I won't use an additional column for this, just pack everything in one column, or even better use a `ListBox`.

sevoku:
Also see https://github.com/mono/monodevelop/blob/master/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/LicenseAcceptanceService.cs#L48-L56 for a good native dialog example.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.PickMembersService/PickMembersService.cs
sevoku: See [LicenseAcceptanceDialog](https://github.com/mono/monodevelop/blob/master/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/LicenseAcceptanceService.cs#L48-L56) for an example

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.PickMembersService/PickMembersService.cs
mkrueger: No should be modal.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.PickMembersService/PickMembersService.cs
mkrueger: I don't think that works here because I would need to return 

res.Task.Result. And if it locks it would lock at that. However I can implement it like that - but no idea if it really would help in that case.

mkrueger: Locks the UI in that case - it's running on the UI thread and the task gets added to the idle threat and res.Task.Result is locking there.

It's IMO wrong here to use that in this case because the actions run on the UI thread.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.PickMembersService/PickMembersDialog.cs
mkrueger: /fixed

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.PickMembersService/PickMembersService.cs
sevoku: Oh ok, I didn't look into the actual context, so if it doesn't come from some background task, then `RunWhenIdle()` doesn't make sense. But can we make it native using `Xwt.Toolkit.NativeEngine.Invoke`?

slluis:
@mkrueger This has been approved. Please fix the conflicts and merge. Thanks!

PR #3700, opened 2018-01-19
[Tools] Terminal cannot be used with Tools - Open in Terminal window

iainx:
If the command is null, we don't need to run an extra bash process

Fixes VSTS #554029


slluis:
Please create a PR for 15.6

iainx:
Done https://github.com/mono/monodevelop/pull/3704

PR #3718, opened 2018-01-23
Only call GetErrorsAsync once instead of multiple times

### LINKED ISSUE #3717: Reduce the number of times that GetErrorsAsync is called
Current `GetErrorsAsync` gets called 3 times every time the document is parsed.
1. Folding (to determine if there is an error and if it should enable folding)
2. Underlines
3. Quicktasks

For 2 and 3 they happen in pretty much the same place:
https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/ErrorHandlerTextEditorExtension.cs#L84

Instead of calling `UpdateErrorUndelines` (which is spelled incorrectly) and `UpdateQuickTasks` where each one calls `GetErrorsAsync` independently it should probably get the errors and then pass them into those methods.

decriptor:
Fixes #3717


mkrueger:
At best we don't use ParsedDocument anymore for C#. Marius was working on getting rid of that.

But patch looks good so far :)

decriptor:
This patch was mostly targeting xml files, so It is probably still needed long term for us at least?

PR #3719, opened 2018-01-23
Editor features- IEditorOperations

mkrueger:
Need that for the automatic bracket API


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.csproj
```diff
@@ -117,6 +117,10 @@
       <Private>False</Private>
     </Reference>
     <Reference Include="mscorlib" />
+    <Reference Include="Microsoft.VisualStudio.Text.UI">
```
KirillOsenkov: Instead of referencing from output directory, just use this:
https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj#L9

PR #3723, opened 2018-01-24
[d15-6][Ide] Remove NuGet imports when migrating to project.json

mrward:
When a PCL project was migrated to project.json the imports for
MSBuild .targets were not removed. These imports are added by NuGet
into the generated ProjectName.nuget.props and
ProjectName.nuget.targets files on restoring when a project.json file
is used. Having the import in the project file results in the import
being used twice. When a PCL project that uses Xamarin.Forms was
migrated to project.json the build would fail with an error:

Error XF001: Xamarin.Forms targets have been imported multiple times.
Please check your project file and remove the duplicate import(s).

Now on migrating the imports from the packages that were used by
the project are removed from the project file.

VSTS 552280


neerajdeshmukh:
Verified on Version 7.4 Preview (7.4 build 929) and it is working fine.
Worked on provided steps from bug.
Build Info:
https://gist.github.com/mohakbarokar/25930ae9f39291204aebc20f5a0f2e99

Approved for Merging.

PR #3724, opened 2018-01-24
Static registrar support

Therzok:


Therzok:
Needs cleanup in version-checks before merging.

Therzok:
Seems like there are some test failures, will look.

Therzok:
Started a new build, md-addins tests were not using the stub launcher.

Therzok:
Failure unrelated. Can we merge these both in tandem?

slluis:
Rebuilding

Therzok:
This breaks non-stub launcher runs (i.e. debugging). Need to fix it somehow so libxammac.dylib is used for dev builds, not embedded libxammac.

Therzok:
Fixed debugging.

Therzok:
@slluis I updated the VSTS timeout to 20 mins, it seems to still be taking 12mins.

PR #3730, opened 2018-01-24
[Ide] Display binlog in a TreeView instead of TextEditor

rodrmoya:
The performance seems to be better, at least that's what I'm seeing,
and also, I think a TreeView allows us to customise the look of the rows
in a much easier way than with the TextEditor.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputProcessor.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputProcessor.cs
@@ -31,6 +31,7 @@ using System.Text;
 using System.Threading.Tasks;
 using MonoDevelop.Ide.Editor;
 using Gtk;
+using MonoDevelop.Core;
 
 namespace MonoDevelop.Ide.BuildOutputView
 {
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputProcessor.cs
```diff
@@ -178,6 +180,42 @@ public Task<(string, IList<IFoldSegment>)> ToTextEditor (TextEditor editor, bool
...
+		private async Task ProcessNode (TreeStore store, TreeIter parentIter, BuildOutputNode node, bool includeDiagnostics)
+		{
+			// For non-diagnostics mode, only return nodes with data
+			if (!includeDiagnostics && (node.NodeType == BuildOutputNodeType.Diagnostics ||
+										(!node.HasData && !node.HasErrors && !node.HasWarnings))) {
+				return;
+			}
+
+			TreeIter it = TreeIter.Zero;
+			await Runtime.RunInMainThread (() => {
```
slluis: This is going to cause a gui thread transition for every node added to the tree, and that is very slow. It would be better to add the nodes all at once in the gui thread.

PR #3731, opened 2018-01-25
[Core] Fix TypeInitializationException using file watcher on Windows

mrward:
A TypeInitializationException is thrown when a .NET Core project
is used on Windows. On Windows when a DllImport cannot be found a
DllNotFoundException is thrown not an EntryPointNotFoundException. To
avoid this the native Mac support check is now only run when the
platform is Mac.


mrward:
Not tested on Windows yet. Will see if I can built the code on Windows and quickly verify the watcher works there.

mrward:
Tested on Windows and the TypeInitializationException does not happen on this branch but does with master.

PR #3733, opened 2018-01-25
[Ide] Use void instead of bool for pinvoke return types

bratsche:
This matches these two functions to the C return types

See https://github.com/bratsche/gtk/blob/393f44073cf81c17d8c84138d3fa0549b6f07cf4/gdk/quartz/gdkkeys-quartz.c#L778

https://github.com/bratsche/gtk/blob/393f44073cf81c17d8c84138d3fa0549b6f07cf4/gdk/quartz/gdkglobals-quartz.c#L53


DavidKarlas:
There is same class also in Xwt

bratsche:
I'll do xwt also

PR #3735, opened 2018-01-25
Fix --verify-all MonoArg attribute

dlech:
The option is `--verify-all`, not `--verifyAll`.

This was discovered while trying to debug something that needed the `--verify-all` option set. MonoDevelop just hung when pressing F5 to start the program. After enabling the show terminal option, I discovered that it said "unknown option --verifyAll".


dlech:
Unfortuneately, I haven't been able to test this because I haveven't been able to figure out how to build monodevelop. It always fails with error about fsharp or if it does actually build, then it crashes when I run it.

PR #3737, opened 2018-01-26
Get rid of some exceptions that occur during project open

ToddGrun:
More annoying exceptions

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteBuildEngineManager.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteBuildEngineManager.cs
@@ -1,4 +1,4 @@
-﻿//
+//
 // RemoteBuildEngineManager.cs
 //
 // Author:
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteBuildEngineManager.cs
```diff
@@ -609,9 +609,10 @@ static void CleanCachedMSBuildExes ()
 				spid = spid.Substring (0, i);
 				int pid;
 				if (int.TryParse (Path.GetFileName (spid), out pid)) {
-					try {
-						// If there is a process running with this id it means the builder is still being used
-						if (Process.GetProcessById (pid) != null)
+					try {
+						// If there is a process running with this id it means the builder is still being used
+						var runningProcesses = Process.GetProcesses();
```
Therzok: What is the performance impact on startup for this?

Therzok: The reasoning: You're looking up the whole process table on every directory found. It should be cached above, I don't think this will race in a bad way.

ToddGrun: I went ahead and reverted this portion of the change. I wasn't seeing it hit during startup, but rather on project load. I measured the perf of this change and it was about 4 ms slower on my machine than the original code, probably enough to warrant me having to suffer through the exception. Kinda wish there was a TryGetProcessById.

PR #3738, opened 2018-01-26
[Padtree] Disable the default treeview search feature

iainx:
GtkTreeView has a feature that has an interactive search when you press
cmd/ctrl+f which we don't want to use. There is a bug in Gtk that reenables
the search when the model is set, so we need to hack around that

This fixes it for all uses of PadTreeView, including Search Results pad, Errors
pad, Breakpoints pad and many others.

Fixes VSTS #551327

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/PadTreeView.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/PadTreeView.cs
@@ -36,6 +36,18 @@ namespace MonoDevelop.Ide.Gui.Components
 		CellRendererText textRenderer = new CellRendererText ();
 		readonly List<CellRendererText> renderers = new List<CellRendererText> ();
 
+		static void ResetSearchColumn (object sender, EventArgs args)
+		{
+			// The Gtk documentation is incorrect for SearchColumn.
+			// SearchColumn does not get reset to -1 when the model changes,
+			// it gets reset to the first string column which then reenables the ctrl+f search feature
+			//
+			// This is probably not the behaviour expected so we work around it here.
+
+			var tree = sender as TreeView;
+			tree.SearchColumn = -1;
+		}
+
 		public PadTreeView ()
 		{
 			Init ();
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/PadTreeView.cs
```diff
@@ -36,6 +36,23 @@ public class PadTreeView : MonoDevelop.Components.ContextMenuTreeView
 		CellRendererText textRenderer = new CellRendererText ();
 		readonly List<CellRendererText> renderers = new List<CellRendererText> ();
 
+		public new TreeModel Model {
```
Therzok: If the PadTreeView is passed as `ContextMenuTreeView`, it won't call this. We probably need to figure out a better way here.

sevoku: MAybe we could use `OnPropertyNotifyEvent` instead?

iainx: I don't think there is a better solution, this one works for the majority of cases, we can document other cases.

We can "fix" the bug in our Gtk, but then it's still broken on Linux

Therzok: That doesn't work that way. It's only reported for a few objects.

I think we could add a Notification for `model` and do things there.

Therzok: via `this.AddNotification ("model", handler)`

sevoku: It won't hurt do move it to ContextMenuTreeView

Therzok: You get the same problem there, if someone stores the `ContextMenuTreeView` as `Gtk.TreeView`. Add a notification on the tree to check for model, then reset the search column. that would be easiest.

PR #3739, opened 2018-01-26
Fixes "VSTS 524616 - can't refresh the website successfully after del…

mkrueger:
…ete the content in '.cshtml' file"

https://devdiv.visualstudio.com/DevDiv/_workitems?id=524616
Was caused by an incompatibility of the vs text buffer with our text buffer due to the changed event behavior.

### DIFF
--- a/main/src/core/MonoDevelop.TextEditor.Tests/Mono.TextEditor.Tests/DocumentTests.cs
+++ b/main/src/core/MonoDevelop.TextEditor.Tests/Mono.TextEditor.Tests/DocumentTests.cs
@@ -40,7 +40,7 @@ namespace Mono.TextEditor.Tests
 		public void TestDocumentCreation ()
 		{
 			var document = new Mono.TextEditor.TextDocument ();
-			
+
 			string text = 
 			"1234567890\n" +
 			"12345678\n" +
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.TextEditor.Tests/Mono.TextEditor.Tests/DocumentTests.cs
```diff
@@ -195,5 +195,21 @@ public void TestBug53380 ()
...
+		/// VSTS 524616 can't refresh the website successfully after delete the content in '.cshtml' file
+		/// </summary>
+		[Test]
+		public void TestVSTS524616 ()
+		{
+			var document = new Mono.TextEditor.TextDocument ();
+			document.Text = "test";
+			string txt;
+			document.TextChanging += delegate {
+				txt = document.Text;
```
Therzok: txt is not used.

KirillOsenkov: In fact the entire delegate is not used.

KirillOsenkov:
Also the failure only used to happen in old Razor documents as far as I can tell - the test doesn't seem to use that?

main/src/core/MonoDevelop.TextEditor.Tests/Mono.TextEditor.Tests/DocumentTests.cs
mkrueger: not needed to be used it should only be called.

mkrueger:
y the test fails I write tests before the fix ... ty for asking

mkrueger:
The failure is not really connected to razor but the old razor code does what the unit test does. I don't plan to use the razor code in any test if it's not needed. The old razor code should go away IMO.

We're using the VS.NET webtools AFAIK - sooner or later we'll get a better razor implementation.

PR #3740, opened 2018-01-26
Fixes "VSTS 524616 - can't refresh the website successfully after del…

mkrueger:
…ete the content in '.cshtml' file"

https://devdiv.visualstudio.com/DevDiv/_workitems?id=524616
Was caused by an incompatibility of the vs text buffer with our text buffer due to the changed event behavior.

### DIFF
--- a/main/src/core/MonoDevelop.TextEditor.Tests/Mono.TextEditor.Tests/DocumentTests.cs
+++ b/main/src/core/MonoDevelop.TextEditor.Tests/Mono.TextEditor.Tests/DocumentTests.cs
@@ -40,7 +40,7 @@ namespace Mono.TextEditor.Tests
 		public void TestDocumentCreation ()
 		{
 			var document = new Mono.TextEditor.TextDocument ();
-			
+
 			string text = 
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.TextEditor.Tests/Mono.TextEditor.Tests/DocumentTests.cs
```diff
@@ -195,5 +195,21 @@ public void TestBug53380 ()
...
+
+		/// <summary>
+		/// VSTS 524616 can't refresh the website successfully after delete the content in '.cshtml' file
+		/// </summary>
+		[Test]
+		public void TestVSTS524616 ()
+		{
+			var document = new Mono.TextEditor.TextDocument ();
+			document.Text = "test";
+			string txt;
```
Therzok: txt is not used.

GLjackyvaswani:
Content is getting refreshed when delete or add any content under cshtml.
Issue BUG 524616 is working fine now.

We can resolved this PR.

PR #3742, opened 2018-01-26
Fixes #557431 - Fix Solution -> Workspace creation synchronously

DavidKarlas:


Therzok:
What I think lluis meant was:
https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/RootWorkspace.cs#L937

TypesystemService.Load call here should return a workspace.
Thus TypesystemService is 2 part:
synchronous create workspace and return it
Task.Run() load it.

Therzok:
RootWorkspace does not use the workspace list being returned, so that is the actual problem.

Therzok:
I think the unit test passes because we're lucky.

We don't wait for it to be loaded here:
http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Ide/RootWorkspace.cs,937

DavidKarlas:
I think you are confusing Workspace being created and Workspace being loaded/populated with things... Purpose of this fix is not that we get fully populated Workspace back but we get Workspace object that will be populated later in background.

slluis:
David is right, this is the fix I was proposing.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService_WorkspaceHandling.cs
```diff
@@ -118,40 +118,40 @@ public static void NotifyFileChange (string fileName, string text)
...
+				await Task.WhenAll (tasks).ConfigureAwait (false);
+			} else if (item is MonoDevelop.Projects.Solution solution) {
+				var workspace = new MonoDevelopWorkspace (solution);
+				lock (workspaceLock)
+					workspaces = workspaces.Add (workspace);
+				list.Add (workspace);
+				if (showStatusIcon)
+					workspace.ShowStatusIcon ();
+
+				await workspace.TryLoadSolution (cancellationToken).ConfigureAwait (false);
```
slluis: Maybe we need to do run this in a Task.Run() to ensure that the solution is loaded in a background thread.

DavidKarlas: TryLoadSolution is calling CreateSolutionInfo which 1st statement is Task.Run so we are OK.

DavidKarlas:
@slluis what we might want to consider is... splitting this logic in 2 methods:
Synchronous `CreateWorkspace` and Asynchronous `LoadWorkspace` to remove current implicit synchronous behavior which someone in future might remove accidentally(in theory test should catch that but still)

slluis:
Makes sense.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService_WorkspaceHandling.cs
```diff
@@ -41,7 +41,7 @@ namespace MonoDevelop.Ide.TypeSystem
 {
 	public static partial class TypeSystemService
 	{
-		static readonly MonoDevelopWorkspace emptyWorkspace;
+		static internal readonly MonoDevelopWorkspace emptyWorkspace;
```
KirillOsenkov: the order should be `internal static`

main/tests/Ide.Tests/MonoDevelop.Ide/TypeSystemServiceTests.cs
```diff
@@ -91,5 +91,23 @@ public void TestOuptutTracking_LanguageName ()
...
 			sol.Dispose();
 		}
+
+		[Test]
+		public async Task TestWorkspaceImmediatelyAvailable ()
+		{
+			//Initialize IdeApp so IdeApp.Workspace is not null
+			if (!IdeApp.IsInitialized)
+				IdeApp.Initialize (new ProgressMonitor ());
+			string solFile = Util.GetSampleProject ("console-project", "ConsoleProject.sln");
```
KirillOsenkov: >solFile [](start = 10, length = 7)

It's nice to avoid abbreviations :) solutionFile is better in my opinion :)

PR #3745, opened 2018-01-26
Consume new IFilePathRegistryService from vs platform to allow the fi…

ToddGrun:
…le path to determine the content type / mimetype

Remove the LinkTypes code that was specific to webtooling as the code now falls back to using the corresponding methods from IContentTypeRegistryService2

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/PlatformService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/PlatformService.cs
@@ -304,7 +304,12 @@ namespace MonoDevelop.Ide.Desktop
 			IContentType contentType = filePathRegistryService.GetContentTypeForPath (fileName);
 			if (contentType != PlatformCatalog.Instance.ContentTypeRegistryService.UnknownContentType) {
 				string mimeType = PlatformCatalog.Instance.MimeToContentTypeRegistryService.GetMimeType(contentType);
-				return FindMimeType (mimeType);
+				if (mimeType != null) {
+					MimeTypeNode mt = FindMimeType(mimeType);
+					if (mt != null) {
+						return mt;
+					}
+				}
 			}
 
 			foreach (MimeTypeNode mt in mimeTypeNodes) {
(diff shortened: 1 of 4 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/PlatformService.cs
```diff
@@ -296,6 +299,14 @@ static PlatformService ()
 
 		MimeTypeNode FindMimeTypeForFile (string fileName)
 		{
+			IFilePathRegistryService filePathRegistryService = CompositionManager.GetExportedValue<IFilePathRegistryService> ();
+
+			IContentType contentType = filePathRegistryService.GetContentTypeForPath (fileName);
+			if (contentType != PlatformCatalog.Instance.ContentTypeRegistryService.UnknownContentType) {
+				string mimeType = PlatformCatalog.Instance.MimeToContentTypeRegistryService.GetMimeType(contentType);
+				return FindMimeType (mimeType);
```
DavidKarlas: Is it possible that `FindMimeType` returns null but `foreach` below could find something?

Therzok: Yes, when for example you write a binary file to `asdf.cs`

Therzok: Or, for example, F# files have 3 kinds of patterns: ".fs", ".fsx", ".fsharp", all with the `mimeType` `text/x-fsharp`. There could be another `.fsi` extern with `text/x-fsharp` as the mimeType. If it's not registered in the IDE as handled by F#, it'll return null.

ToddGrun: OK, let me add a check to harden this up then.

ToddGrun: OK, I've added another commit that checks this. I could never end up getting back a null mimeType because there is an existing codepath that creates a content type when it's lookup fails.

PR #3755, opened 2018-01-30
[Editor] Reset IM context when filtering keypresses

bratsche:
This fixes issues in languages like Japanese and Chinese where
the text isn't visible until it gets committed from the preedit
buffer.

Fixes VSTS #535828


bratsche:
@slluis Thanks, fixed.

PR #3756, opened 2018-01-30
Display .binlog in a tree view

rodrmoya:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
```diff
@@ -0,0 +1,96 @@
...
+		Message,
+		Diagnostics
+	}
+
+	class BuildOutputNode : TreePosition
+	{
+		public virtual BuildOutputNodeType NodeType { get; set; }
+		public virtual string Message { get; set; }
+		public BuildOutputNode Parent { get; set; }
+		public virtual IList<BuildOutputNode> Children { get; } = new List<BuildOutputNode> ();
```
slluis: This is still creating a list by default, it shouldn't

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -238,4 +249,130 @@ protected override void OnWriteLog (string message)
...
+			this.includeDiagnostics = includeDiagnostics;
+			if (includeDiagnostics) {
+				rootNodes = buildOutput.RootNodes.ToList ();
+			} else {
+				// If not including diagnostics, we need to filter the nodes,
+				// but instead of doing so now for all, we do it on the fly,
+				// as nodes are requested
+				rootNodes = new List<BuildOutputNode> ();
+				foreach (var root in buildOutput.RootNodes) {
+					rootNodes.Add (new FilteredBuildOutputNode (root, null, includeDiagnostics));
```
slluis: Shouldn't top level nodes also be filtered?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
rodrmoya: Yes, right, was going to do it on a separate PR, but I can add it here

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
rodrmoya: The nodes to not include when not displaying diagnostics are just messages or empty Task/Target's, so root nodes (Build nodes) are not affected.

Issue #3640, opened 2018-01-12
Toggle Suggestion Mode no longer has keybinding or menu entry


mhutch:
The 'Toggle Suggestion Mode' command is missing from the Edit Menu and no longer has a keybinding. it's now only possible to find it in the keybindings panel. A number of folks have complained that VS for Mac doesn't have this command.

This is an important feature - it's even called out in the [VS introductory docs](https://docs.microsoft.com/en-us/visualstudio/ide/using-intellisense#list-members).

We should add it back to the menu and restore the *Ctrl + Alt + Space* keyboard shortcut.

> VS bug #552034

sevoku:
Yeah, it was a design descision to remove it from the main menu, and the keyboard shortcut has been removed much earlier with 3aea18098ae30eec5ea456e688f063366d34482d back in 2013 (@mkrueger do you remember the reason?). We can revert both changes easily. cc @hbons

GLjackyvaswani:
I can confirm that the above issue is not reproduiceble in below builds.

Build Info:

Visual Studio Enterprise 2017 for Mac (Preview)
Version 7.4 Preview (7.4 build 945)

Xamarin.Mac
Version: 4.2.0.8 (Visual Studio Enterprise)
Xamarin.iOS
Version: 11.8.0.8 (Visual Studio Enterprise)
Xamarin.Android
Version: 8.2.0.6 (Visual Studio Enterprise)

Gist Link: https://gist.github.com/parasAmbhore/eef24234c94652607602e7c443063f28
Screen Cast: https://www.screencast.com/t/T84eOyNAG

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #3757 merged by: Vsevolod Kukol
Author: mono
Commit: 4daa08a2f8305b11dc944b512658aea0a59b5a34 (xamarin/monodevelop)

PR #3757, opened 2018-01-31
[Ide] Add Toggle Completion Mode menu and keys

sevoku:
This PR brings back the "Edit → Toggle Completion Mode" menu entry and adds it's default _Ctrl + Alt + Space_ binding to the VS binding schemes.

grafik

EDIT: We can't add _Ctrl + Alt + Space_ as a default MonoDevelop binding, because it would overlap with _Import Symbol_.

Fixes gh #3640

### DIFF
--- a/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
+++ b/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
@@ -970,7 +970,8 @@
 			shortcut = "Control|space" />
 	<Command id = "MonoDevelop.Ide.Commands.TextEditorCommands.ToggleCompletionSuggestionMode"
 			defaultHandler = "MonoDevelop.Ide.Commands.ToggleCompletionSuggestionModeHandler"
-			_label = "Toggle Completion Suggestion Mode" />
+			_label = "Toggle Completion Mode"
+			_displayName = "Toggle Completion Suggestion Mode" />
 	<Command id = "MonoDevelop.Ide.Commands.TextEditorCommands.ShowCodeTemplateWindow"
 			_description="Inserts a snippet"
 			_label = "_Snippet..."
--- a/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
+++ b/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
@@ -970,8 +970,8 @@
 			shortcut = "Control|space" />
 	<Command id = "MonoDevelop.Ide.Commands.TextEditorCommands.ToggleCompletionSuggestionMode"
 			defaultHandler = "MonoDevelop.Ide.Commands.ToggleCompletionSuggestionModeHandler"
-			_label = "Toggle Completion Mode"
-			_displayName = "Toggle Completion Suggestion Mode" />
+			_label = "Switch Completion/Suggestion Mode"
+			_displayName = "Switch Completion/Suggestion Mode" />
 	<Command id = "MonoDevelop.Ide.Commands.TextEditorCommands.ShowCodeTemplateWindow"
 			_description="Inserts a snippet"
 			_label = "_Snippet..."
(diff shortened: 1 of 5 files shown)


mrward:
The keyboard shortcut seems to already be used by Import Symbol.

https://github.com/mono/monodevelop/blob/c5b9b039424e960912e1ac2d872ea81a6b1738d0/main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.addin.xml#L71

sevoku:
Yeah, that's the reason why I add it to the VS schemes only: 8ea5518244cac19bc80a0d58b64631f86cc9b64b
We can't add it to the default bindings because of the overlapping with the `Import Symbol` command.

mhutch:
I wonder whether there's a good way we could display the state.

e.g. maybe

```
Completion Mode > |    Complete Symbols |
                  | ✔️ Suggest Symbols  |
```

The downside is that we wouldn't display the keybinding... but we don't have one by default anyways.

Alternately we could maybe have a single command with a full name *Switch Completion/Suggestion Mode*, which would display *Switch to Completion Mode* or *Switch to Suggestion Mode* depending on the state.

Thoughts?

mhutch:
cc @maryanne-exe @hbons

maryanne-exe:
The idea of replacing “Switch to Completion Mode”/"Switch to Suggestion Mode" sounds good: action is clear and we can have a shortcut. The only problem that we can't see currently selected mode but that's easy to discover.

Another idea we can have Completion Mode by default and have “Use Suggestion mode” in the menu with a checkmark: the user can check or uncheck.

mhutch:
Makes sense! I was also thinking about usability via the global search and discoverability via search in the keybinding options panel.

maryanne-exe:
A couple more thoughts:

1. Completion Mode >  |      Complete Symbols     Key1  |
                                       | ✔️ Suggest Symbols       Key2  |

Key1 and Key2 are slightly different. We have a similar solution in View > Editor Columns. Cons: one more key to remember, Pros: discoverable via searches and has keybindings.

2. Have a toggle like I mentioned in a comment above but call it “Suggestive Completion Mode”, the user checks/unchecks. Pros: discoverable via search and has a key.

mhutch:
Not a fan of having multiple keybindings to remember for this toggle operation tbh

sevoku:
@maryanne-exe @mhutch I also like the idea of “Switch to Completion Mode”/"Switch to Suggestion Mode" or having a checkbox for "Use Suggestion mode”. A submenu is just too much for that, and it's problematic with keybindings cause it'd be a different command (array) type (we could have still an additional toggle binding, but it won't be visible in the menu).

@slluis I didn't add the tag cause all 3 commits are involved. In such cases I always put that tag in the merge commit. Can't we do it this way anymore?

mhutch:
I like the "Switch to..." as it makes the command action very clear/explicit

slluis:
@sevoku I want the fix message in a commit message because then I can generate a commit history report with a list of all fixes. Bug auto-closing is a feature that detects bug fix messages, but it is not the only one. Including the fix message in the merge commit is ok-ish, but it has some problems: it is easy to forget, you may not be the one merging, and if you cherry-pick the fix to another branch you have to remember again to set the fix message. If the message is in a commit, it all just works.

slluis:
Also, when there is more than one commit, adding the message to the last one is enough.

GLjackyvaswani:
I can confirm that the above issue is not reproduiceble in below builds.

Build Info:

Version 7.4 Preview (7.4 build 945)

Xamarin.Mac
Xamarin.iOS
Xamarin.Android

Gist Link: https://gist.github.com/parasAmbhore/eef24234c94652607602e7c443063f28

Screen Cast: https://www.screencast.com/t/T84eOyNAG

sevoku:
Ok I've update the command labels:

grafik
grafik
grafik
grafik

@slluis the last comming message includes the fixes tag now

mhutch:
Looks great, thanks!

GLjackyvaswani:
I can confirm that the above issue is not reproducible in below builds.

Build Info:

Version 7.4 Preview (7.4 build 946)

Xamarin.Mac
Xamarin.iOS
Xamarin.Android

Gist Link: https://gist.github.com/parasAmbhore/5d69ae017de4e7eb8c4569dcca3a16c7

Screen Cast: https://www.screencast.com/t/u5WSXvQG

PR #3759, opened 2018-01-31
[IDE] Use ContextMenu on the Errors and Tasks pad

iainx:
Use the cross platform ContextMenu on for the Error and Tasks pads' popup menu.
Means that
 - An ugly Gtk menu isn't used on Cocoa
 - Fixes refcycles and leaks

Create a ColumnSelectorMenu for automatically generating context menus for
hiding or showing treeview columns

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/PadTreeView.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/PadTreeView.cs
@@ -26,7 +26,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.Text;
 using Gtk;
+using MonoDevelop.Components;
+using MonoDevelop.Core;
 
 namespace MonoDevelop.Ide.Gui.Components
 {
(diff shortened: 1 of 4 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/PadTreeView.cs
```diff
@@ -120,5 +123,77 @@ internal void RegisterRenderForFontChanges (CellRendererText renderer)
...
+		}
+
+		string id;
+
+		public ColumnSelectorMenu (TreeView treeview, string restoreID, params string[] columnNames)
+		{
+			id = restoreID;
+
+			int i = 0;
+			foreach (var column in treeview.Columns) {
```
Therzok: Would be nicer here.

```
var columns = treeview.Columns;
for (int i ... ) {

}
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/PadTreeView.cs
```diff
@@ -120,5 +123,77 @@ internal void RegisterRenderForFontChanges (CellRendererText renderer)
...
+			foreach (var c in Items) {
+				var item = c as CheckBoxContextMenuItem;
+				if (item == null) {
+					continue;
+				}
+				builder.AppendFormat ("{0};", item.Checked);
+			}
+			builder.Remove (builder.Length - 1, 1);
+
+			Console.WriteLine ($" Writing {builder} to {id}");
```
Therzok: Debug code.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/PadTreeView.cs
```diff
@@ -120,5 +123,77 @@ internal void RegisterRenderForFontChanges (CellRendererText renderer)
...
+	{
+		class ColumnSelectorMenuItem : CheckBoxContextMenuItem
+		{
+			TreeViewColumn column;
+
+			public ColumnSelectorMenuItem (TreeViewColumn column, string label) : base (label)
+			{
+				this.column = column;
+
+				Checked = column.Visible;
```
Therzok: This is one case I'm not sure whether it'll leak.

We have:
TVC -> this via event handler
and
this -> TVC via field

I think we need a `WeakReference<TreeViewColumn>` here.

Therzok: An easy way to test this, without spinning up a profiler, is to make a finalizer and see if it's hit.

iainx: The menu is displayed in response to an event on the TreeView, so I'm not sure what the event is creating a ref between the TVC and this?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/PadTreeView.cs
iainx: Really? That's just personal preference surely?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/PadTreeView.cs
```diff
@@ -120,5 +123,77 @@ internal void RegisterRenderForFontChanges (CellRendererText renderer)
...
+		public ColumnSelectorMenu (TreeView treeview, string restoreID, params string[] columnNames)
+		{
+			id = restoreID;
+
+			int i = 0;
+			foreach (var column in treeview.Columns) {
+				var name = i < columnNames.Length ? columnNames[i] : null;
+
+				var item = new ColumnSelectorMenuItem (column, string.IsNullOrEmpty(name) ? column.Title : name);
+				item.Checked = treeview.Columns[i].Visible;
```
iainx: Line isn't needed...

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/PadTreeView.cs
Therzok: Well, ref via field and ref via the `OnItemClicked` instance method. A finalizer check would be enough to verify this, but I'm erring on the side of this is leaking.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/PadTreeView.cs
Therzok: Well, you're still using an indexer here, it just looks cleaner to be to see that `columnNames[i]` matches `treeview.Columns[i]`.

Therzok:
Pushed leak fix for ContextMenu in this case. The instance method subscription will no longer cause a leak, now that the `ContextMenuItem` is wrapped into a WeakReference, rather than strong reference via EventHandler.Target

PR #3771, opened 2018-02-05
[Ide] Switch BuildOutputWidget to use XWT

rodrmoya:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -232,4 +248,105 @@ protected override void OnWriteLog (string message)
...
+
+		public object GetValue (TreePosition pos, int column)
+		{
+			var node = pos as BuildOutputNode;
+			if (node != null) {
+				switch (column) {
+				case 0: // Image
+					switch (node.NodeType) {
+					case BuildOutputNodeType.Build:
+						return ImageService.GetIcon (Ide.Gui.Stock.StatusBuild, Gtk.IconSize.Menu);
```
slluis: Load the image in the constructor and store it in a field. It will be faster than getting it from the ImageService every time.

PR #3772, opened 2018-02-05
[Ide] Add start/end times to BuildOutputNode

rodrmoya:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -217,13 +217,13 @@ protected override void OnWriteLogObject (object logObject)
 					BuildOutput.Load (pspe.LogFile, true); 
 				} else {
 					currentCustomProject = new BuildOutputProcessor (pspe.LogFile, false);
-					currentCustomProject.AddNode (BuildOutputNodeType.Project, "Custom project", true);
+					currentCustomProject.AddNode (BuildOutputNodeType.Project, "Custom project", true, DateTime.Now);
```
slluis: Maybe the start/end times should be provided by the ProjectStartedProgressEvent / ProjectFinishedProgressEvent objects, it would be more accurate.

rodrmoya:
DON'T MERGE - this will conflict with #3756 , and it will be easier to redo this PR than the other one

rodrmoya:
Ready for review now

PR #3774, opened 2018-02-05
[IDE] Tool output pad not being reused - appears multiple times

iainx:
Rework the LogView so that it can handle multiple progress monitors feeding into
it and then CustomToolService just needs to use one Tool Output pad.

Fixes VSTS #527927


slluis:
How does it ensure that output from two tasks running in parallel don't overlap?

PR #3777, opened 2018-02-06
[Core] Add a fallback when a given culture does not exist in .NET

Therzok:
OS X seems to have some cultures that do not exist in .NET. Fallback to default when they're not found.

Fixes VSTS #532970


Therzok:
Failure unrelated. cc @slluis

PR #3778, opened 2018-02-06
[d15-6][Core] Add a fallback when a given culture does not exist in .NET

Therzok:
OS X seems to have some cultures that do not exist in .NET. Fallback to default when they're not found.

Fixes VSTS #532970


Therzok:
To reproduce this crash, you need to have the IDE locale set to default and OS locale set to Furlan (Friulian)

tanmay-bhattacharjee:
I can confirm that issue #532970 is not reproducible when tested with the following build:
https://devdiv.visualstudio.com/_apis/resources/Containers/1726026?itemPath=drop%2FVisualStudioForMac-Preview-7.4.0.988.dmg

To test this issue, IDE locale was set to default and OS locale set to Furlan (Friulian) and issue was not reproducible which was reproducible when tested with an old build.

This PR can be merged.
Thanks!

PR #3779, opened 2018-02-06
Fixed 'VSTS Bug 561844: [Feedback] Code snippets don't highlight

mkrueger:
defined input variables upon creation'

https://devdiv.visualstudio.com/DevDiv/_workitems?id=561844

I obsoleted the formatting infrastructure - it should be taken out and
replaced with the roslyn infrastructure.

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Formatting/OnTheFlyFormatter.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Formatting/OnTheFlyFormatter.cs
@@ -36,7 +36,6 @@ using MonoDevelop.Ide;
 using MonoDevelop.Ide.Editor;
 using MonoDevelop.Ide.Gui.Content;
 using MonoDevelop.Projects.Policies;
-using Roslyn.Utilities;
 
 namespace MonoDevelop.CSharp.Formatting
 {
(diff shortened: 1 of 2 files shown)


main/src/addins/CSharpBinding/MonoDevelop.CSharp.Formatting/OnTheFlyFormatter.cs
```diff
@@ -76,7 +75,7 @@ public static void FormatStatmentAt (TextEditor editor, DocumentContext context,
 				return;
 			using (var undo = editor.OpenUndoGroup (/*OperationType.Format*/)) {
 				try {
-					var syntaxTree = await analysisDocument.GetSyntaxTreeAsync ();
+					var syntaxTree = analysisDocument.GetSyntaxTreeAsync ().WaitAndGetResult (default (CancellationToken));
```
Therzok: Can we make this an async task and propagate that? Blocking wait here could probably run roslyn on the UI thread.

mkrueger: Not in that case needs to run on UI thread here the operation requires to alter the text - and editor.ApplyTextChanges needs to run on UI thread. The API is the problem here - only way would've been to make a new API that gives back the formatting changes instead but basically that old API and their usages would break anyways.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeFormatting/CodeFormatter.cs
```diff
@@ -53,12 +53,13 @@ internal CodeFormatter (string mimeType, AbstractCodeFormatter formatter)
 			this.formatter = formatter;
 		}
 
-		[Obsolete("Use Format (PolicyContainer policyParent, ITextSource input, ISegment segment = null) instead. This function is going to be removed.")]
+		[Obsolete ("Use roslyn instead")]
```
DavidKarlas: This code is not C# specific, stating `[Obsolete ("Use roslyn instead")]` seems wrong...

mkrueger: No not really - make a custom solution - at best a roslyn plug in :) - for your language instead. The formatting capabilities should move to the backends.

Roslyn has a generic IFormattingService infrastructure that can be used.

slluis:
So is this PR ok? @Therzok @DavidKarlas

Therzok:
LGTM!

PR #3783, opened 2018-02-06
[IDE] Errors pad autohides after a few moments

iainx:
closes the pad.

Output is hidden.

Fixes VSTS #561864


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/ErrorListPad.cs
```diff
@@ -292,7 +292,7 @@ void CreateControl ()
 				AddTask (t);
 			}
 
-			control.FocusChain = new Gtk.Widget [] { outputView };
+			control.FocusChain = new Gtk.Widget [] { outputView, sw };
```
sevoku: I was sooo blind!

PR #3785, opened 2018-02-06
[15.6][IDE] Errors pad autohides after a few moments

sevoku:
(#3783 for `d15-6`)

closes the pad.

Output is hidden.

Fixes VSTS #561864

(cherry picked from commit 1d6a8648da9661f5d16934a24a2eb1faa5323b83)


GLjackyvaswani:
I can confirm that the above issue is not reproducible in below builds.
Hence marking as QA - Approved can merged.

Build Info:

Version 7.4 Preview (7.4 build 988)

Xamarin.Mac
Xamarin.iOS
Xamarin.Android

Gist Link: https://gist.github.com/parasAmbhore/a137a8b002477d843ee97748f61a1880

PR #3786, opened 2018-02-06
[15.5][IDE] Errors pad autohides after a few moments

sevoku:
(#3783 for d15-5)

closes the pad.

Output is hidden.

Fixes VSTS #561864

(cherry picked from commit 1d6a8648da9661f5d16934a24a2eb1faa5323b83)


mohakbarokar:
Bug #561864 is tested on Build https://devdiv.visualstudio.com/DevDiv/_build/index?buildId=1356376&tab=artifacts&_a=summary

Build Gist: https://gist.github.com/mohakbarokar/55d0015ee1b6dcd16d5abbf94f458419

We can merge the PR as the Bug is not reproducible

PR #3793, opened 2018-02-07
[Git] Fix progress reporting on clone

Therzok:
This now reports that a loose object can not be found instead of crashing and shows checkout progress in the pad

Fixes VSTS #549721


Therzok:
cc @slluis. Cloning still fails, but at least it reports a backend problem now. Libgit2 seems to not handle some submodule setup done by xamarin-android.

PR #3819, opened 2018-02-09
Buildoutput breadcrumb

jtorres:
Implements Navigation bar (breadcrumb)

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
@@ -37,15 +37,19 @@ using Microsoft.Build.Framework;
 using Gtk;
 using Xwt;
 using System.Linq;
+using MonoDevelop.Ide.Gui.Content;
+using MonoDevelop.Components;
+using Xwt.Drawing;
 
 namespace MonoDevelop.Ide.BuildOutputView
 {
-	class BuildOutput : IDisposable
+	class BuildOutput : IDisposable, IPathedDocument
 	{
 		BuildOutputProgressMonitor progressMonitor;
 		readonly List<BuildOutputProcessor> projects = new List<BuildOutputProcessor> ();
 
 		public event EventHandler OutputChanged;
+		public event EventHandler<DocumentPathChangedEventArgs> PathChanged;
 
 		public ProgressMonitor GetProgressMonitor ()
 		{
@@ -166,16 +170,20 @@ namespace MonoDevelop.Ide.BuildOutputView
 			OutputChanged?.Invoke (this, EventArgs.Empty);
 		}
 
+		BuildOutputDataSource currentDataSource;
 		public BuildOutputDataSource ToTreeDataSource (bool includeDiagnostics)
 		{
 			foreach (var p in projects) {
 				p.Process ();
 			}
 
-			return new BuildOutputDataSource (this, includeDiagnostics);
+			currentDataSource = new BuildOutputDataSource (this, includeDiagnostics);
+			return currentDataSource;
 		}
 
 		bool disposed = false;
+
+		public PathEntry [] CurrentPath { get; set; }
 
 		~BuildOutput ()
 		{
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -193,12 +201,71 @@ void Dispose (bool disposing)
...
+			public void ActivateItem (int n) => buildOutput.SiblingSelected?.Invoke (this, list[n]);
+
+			public Xwt.Drawing.Image GetIcon (int n) => dataSource.GetValue (list [n], 0) as Xwt.Drawing.Image;
+
+			public string GetMarkup (int n) => list [n].Message;
+
+			public object GetTag (int n) => list [n];
+
+			public void Reset() => list = new BuildOutputNode [0];
 		}
```
rodrmoya: I would move all this code to BuildOutputWidget, which is where the UI happens, so that we can keep `BuildOutput` UI agnostic (apart from helper methods like the `ToTreeDataSource`, which even could go to an extension method.

Thus, apart from removing all UI code from this class, if all this code is in `BuildOutputWidget`, there is no need to keep track of the `currentDataSource`, as `BuildOutputWidget` already has in (in the tree view's `DataSource`).

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -147,14 +220,15 @@ static void ExpandChildrenWithErrors (TreeView tree, BuildOutputDataSource dataS
 			}
 		}
 
+		BuildOutputDataSource dataSource;
 		void ProcessLogs (bool showDiagnostics)
```
rodrmoya: There's no need really to keep a field to store the data source, it's already in the tree view, and can just be accessed with `var dataSource = treeView.DataSource as BuildOutputDataSource`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
rodrmoya: On `BuildOutputWidget` you just need to listen to the treeView's `SelectionChanged`, and that would give you a `BuildOutputNode`, which you can use to access all the needed data, that is, siblings (via its `Parent`'s `Children`, icon, message, etc), as well as to do operations on the tree, as `BuildOutputNode` implements `TreePosition`, which is the type you need to operate on the tree (for selecting, expanding, etc)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -37,6 +37,9 @@
 using Gtk;
 using Xwt;
 using System.Linq;
+using MonoDevelop.Ide.Gui.Content;
+using MonoDevelop.Components;
+using Xwt.Drawing;
```
rodrmoya: the changes in this whole file seem to just be leftovers from the previous change, as these `using` are not needed (right?) and there's no point in doing this -> https://github.com/mono/monodevelop/pull/3819/files#diff-563f5ecf5ef6e0a860fd9d77cb55312fR178

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -133,8 +156,60 @@ void Initialize ()
...
+		{
+			await Runtime.RunInMainThread (() => {
+				treeView.SelectRow (node);
+			});
+		}
+
+		void TreeView_SelectionChanged (object sender, EventArgs e)
+		{
+			var selectedNode = (sender as Xwt.TreeView).SelectedRow as BuildOutputNode;
+			if (selectedNode == null)
```
rodrmoya: here you can just use `treeView`, like a few lines below

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -169,5 +244,60 @@ void ProcessLogs (bool showDiagnostics)
...
+					throw new ArgumentNullException ("widget");
+				this.widget = widget;
+				Reset ();
+
+				var parent = DataSource.GetParent (node) as BuildOutputNode;
+				var rootsCount = DataSource.GetChildrenCount (parent);
+				list = new BuildOutputNode [rootsCount];
+				for (int i = 0; i < rootsCount; i++) {
+					list [i] = DataSource.GetChild (parent, i) as BuildOutputNode;
+				}
```
rodrmoya: hmm, why do you need this `BuildOutputNode` array, can't you just use `parent.Children`?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -172,7 +172,8 @@ public BuildOutputDataSource ToTreeDataSource (bool includeDiagnostics)
 				p.Process ();
 			}
 
-			return new BuildOutputDataSource (this, includeDiagnostics);
+			var currentDataSource = new BuildOutputDataSource (this, includeDiagnostics);
+			return currentDataSource;
 		}
```
rodrmoya: this temp var is unneeded, just `return new BuildOutputDataSource ...` should be enough

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -105,7 +124,11 @@ void Initialize ()
 				}
 			};
 
-			var toolbar = new DocumentToolbar ();
+			toolbar = new DocumentToolbar ();
+
+			toolbar.AddSpace ();
+			box = new Gtk.VBox ();
```
rodrmoya: this .AddSpace call was added to make the buttons show on the right side of the toolbar, but with the addition of the PathBar, which does exactly that (occupy all available space), the AddSpace is not needed

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -36,20 +36,26 @@
...
 namespace MonoDevelop.Ide.BuildOutputView
 {
-	class BuildOutputWidget : VBox
+	class BuildOutputWidget : VBox, IPathedDocument
 	{
 		TreeView treeView;
 		ScrollView scrolledWindow;
 		CheckBox showDiagnosticsButton;
 		Button saveButton;
+		Gtk.VBox box;
```
netonjm: we need use Xwt here

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -105,7 +124,10 @@ void Initialize ()
 				}
 			};
 
-			var toolbar = new DocumentToolbar ();
+			toolbar = new DocumentToolbar ();
+
+			box = new Gtk.VBox ();
```
netonjm: we need use Xwt here

PR #3820, opened 2018-02-09
Build output search

rodrmoya:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -368,5 +370,85 @@ public void SetValue (TreePosition pos, int column, object value)
...
+
+		#region Search functionality
+
+		List<BuildOutputNode> currentSearchMatches;
+		string currentSearchPattern;
+		int currentMatchIndex = -1;
+
+		static void SearchInNodeAndChildren (BuildOutputNode node, List<BuildOutputNode> matches, string pattern)
+		{
+			if (node.Message?.ToLower ().Contains (pattern.ToLower ()) ?? false) {
```
netonjm: if we pass the pattern parameter in ToLower from the caller method we can save the process in each iteration

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -368,5 +370,85 @@ public void SetValue (TreePosition pos, int column, object value)
...
+				foreach (var child in node.Children) {
+					SearchInNodeAndChildren (child, matches, pattern);
+				}
+			}
+		}
+
+		public BuildOutputNode FirstMatch (string pattern)
+		{
+			// Initialize search data
+			if (currentSearchMatches == null) {
```
netonjm: why not initializate like a instantiate readonly instanciated List to avoid unnecessary null checks?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
Therzok: Better to do a case insensitive comparison for performance.

netonjm: We checked to do this in the past in the Search combobox, and the performance was worst.

netonjm: I am not sure if could be related to a bug in mono implementation ¯\_(ツ)_/¯

PR #3822, opened 2018-02-09
Refactoring in BuildOutputDataSource agnostic from filter

netonjm:
Also this change allows get items or iterate in collection from BuildOutput


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -166,13 +167,30 @@ internal void RaiseOutputChanged ()
 			OutputChanged?.Invoke (this, EventArgs.Empty);
 		}
 
+
+		public List<BuildOutputNode> GetTreeRootNodes (bool includeDiagnostics) 
+		{
+			if (includeDiagnostics) {
```
rodrmoya: as talked on Slack, why not just add this `includeDiagnostics` to the already existing `GetRootNodes` method?

PR #3824, opened 2018-02-09
Linux tarball build fixes

directhex:
This ensures `make dist` produces a tarball which works

* Ensure no files needed for build are missed
* Don't bother trying to make packager-friendly tarballs with no binaries - that ship sailed years ago
* Don't call into `scripts/configure.exe` for tarball builds, only `main/` is exported in the tarball


directhex:
Why is VSTS still in pending state? This has been waiting for checks for 5 days

mrward:
@directhex I don't think VSTS supports building from remote repositories. The branch needs to be on the monodevelop repo.

directhex:
@mrward that's... not a way to solicit contributions from anyone who doesn't already have commit rights.

Issue #3806, opened 2018-02-07
Insert of two ' characters when pressing ' is incorrect for F# code


dsyme:
By default the editor inserts two ``'`` marks when you press ``'`` once.

This is wrong for F# code because ``'T`` is a type variable.

We should just disable this behaviour for this key for F# code

> VS bug #564268

mhutch:
See also https://github.com/mono/monodevelop/issues/3475

xamarin-release-manager:
> [sync] [VS-11] Comment by **xamarinc**



Fixed in version 8.6.0.802 (master)



Pull Request #3828 merged by: Mike Kr??ger  
 Author: mono  
 Commit: 253eca238de88767d32b75e81db22d19ae546fdc (xamarin/monodevelop)

PR #3828, opened 2018-02-12
Fixed 'VSTS Bug 564268: Insert of two ' characters when pressing ' is…

mkrueger:
… incorrect for F# code'

Special cased f# files in the bracket handler.

---------------

In theory the f# binding would provide an AutoBracketHandler however that infrastructure is going to be replaced by the VS.NET one - so atm the best solution is just to hack out the f# case in the default handler until the infrastructure is replaced.

Any thoughts ?


nosami:
fixes https://github.com/mono/monodevelop/issues/3806

PR #3829, opened 2018-02-12
[Ide] Fixes a crash when activating no existent PathBar type / element

jtorres:
Issue: While having a "_pathed_" doc opened, when we (i.e.) do a `Select All` and `Cut`, (the doc's length becomes zero -empty-), the PathBar entries are not updated therefore when activating some element in the path bar's DropDownList it crashes.
Fix: Updates PathBar when PathedDoc root's length is equal or less than caret offset instead of just return;

Crash repro: https://www.dropbox.com/s/lwfrc21bm9zz2jv/vsformac%20bug.mov?dl=0


Therzok:
This could be unit testable by creating an UnitTestBase which uses a texteditor with a pathed document extension attached, and verifies that PathChanged is called when the text is set to empty.

PR #3831, opened 2018-02-12
[AOT] Improve targets for AOT

Therzok:
This fixes exclusion path necessity and AOT's everything


Therzok:
Will investigate why assembly-loader=strict is failing, it's possibly a bad resolution of System.ValueTuple.

PR #3835, opened 2018-02-12
Structured build output Goto Log Build Output

netonjm:
This feature allows go to selected error/warning/message in Log Build Output line from the ErrorListPad.

To jump into the error we need all the data loaded. I changed our current code to allow do this.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
@@ -189,34 +189,33 @@ namespace MonoDevelop.Ide.BuildOutputView
 			RaiseOutputChanged ();
 		}
 
-		public IEnumerable<BuildOutputNode> GetRootNodes ()
-		{
-			foreach (var proj in projects) {
-				foreach (var node in proj.RootNodes) {
-					yield return node;
-				}
-			}
-		}
-
 		internal void RaiseOutputChanged ()
 		{
 			OutputChanged?.Invoke (this, EventArgs.Empty);
 		}
 
-
-		public List<BuildOutputNode> GetTreeRootNodes (bool includeDiagnostics)
+		public List<BuildOutputNode> GetRootNodes (bool includeDiagnostics)
 		{
 			if (includeDiagnostics) {
-				return GetRootNodes ().ToList ();
+				return GetProjectRootNodes ().ToList ();
 			} else {
 				// If not including diagnostics, we need to filter the nodes,
 				// but instead of doing so now for all, we do it on the fly,
 				// as nodes are requested
 				var nodes = new List<BuildOutputNode> ();
-				foreach (var root in GetRootNodes ()) {
+				foreach (var root in GetProjectRootNodes ()) {
 					nodes.Add (new FilteredBuildOutputNode (root, null, includeDiagnostics));
 				}
 				return nodes;
+			}
+		}
+
+		IEnumerable<BuildOutputNode> GetProjectRootNodes ()
+		{
+			foreach (var proj in projects) {
+				foreach (var node in proj.RootNodes) {
+					yield return node;
+				}
 			}
 		}
 
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
@@ -417,6 +417,16 @@ namespace MonoDevelop.Ide.BuildOutputView
 		}
 
 		#endregion
+	}
+
+	class BuildOutputDataSearch
+	{
+		readonly List<BuildOutputNode> rootNodes;
+
+		public BuildOutputDataSearch (List<BuildOutputNode> rootNodes)
+		{
+			this.rootNodes = rootNodes;
+		}
 
 		#region Search functionality
 
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
@@ -40,58 +40,6 @@ using System.Linq;
 
 namespace MonoDevelop.Ide.BuildOutputView
 {
-	static class BuildOutputNodeExtensions
-	{
-		public static BuildOutputNode SearchFirstNode (this IEnumerable<BuildOutputNode> sender, BuildOutputNodeType type, string search)
-		{
-			BuildOutputNode tmp;
-			foreach (var item in sender) {
-				tmp = item.SearchFirstNode (type, search);
-				if (tmp != null) {
-					return tmp;
-				}
-			}
-			return null;
-		}
-
-		public static void Search (this BuildOutputNode node, List<BuildOutputNode> matches, string pattern)
-		{
-			if ((node.Message?.IndexOf (pattern, StringComparison.OrdinalIgnoreCase) ?? -1) >= 0) {
-				matches.Add (node);
-			}
-
-			if (node.HasChildren) {
-				foreach (var child in node.Children) {
-					Search (child, matches, pattern);
-				}
-			}
-		}
-
-		public static BuildOutputNode SearchFirstNode (this BuildOutputNode buildOutputNode, BuildOutputNodeType type, string search)
-		{
-			if (type == buildOutputNode.NodeType) {
-				if (search == buildOutputNode.Message) {
-					return buildOutputNode;
-				} else {
-					//We don't want deep recursive into children, change to next item
-					return null;
-				}
-			}
-
-			//iterating into children
-			if (buildOutputNode.Children != null) {
-				BuildOutputNode tmp;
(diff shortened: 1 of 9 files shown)


netonjm:
Do not merge is still in progress

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -40,11 +40,80 @@
...
 namespace MonoDevelop.Ide.BuildOutputView
 {
+	public enum SearchNodesComparation
+	{
+		Match, Contains, StartsWith
+	}
+
+	static class BuildOutputNodeExtensions
+	{
+		public static IEnumerable<BuildOutputNode> SearchNodes (this IEnumerable<BuildOutputNode> sender, BuildOutputNodeType type, string search, SearchNodesComparation searchNodesComparation = SearchNodesComparation.Match)
```
Therzok: Please make the inputs and outputs not be an interface, and don't use yield return. Reason why:

This main method will create an Iterator allocation, but the problem is that the inner loop `item.SearchNode` also creates an Iterator. This makes performance horrible and GC prone.

We can get away with an iterator allocation at caller level, but iterator allocations in the hot loop is not good. We allocate iterators at both nesting levels (even inside SearchNodesBase), so that's a lot of allocations in this code.

Given that the only callers of this API only search for the first item, why not do that?

Therzok: I.e. return a `BuildOutputNode` to match the `FirstOrDefault` behaviour in the callers, and return early, rather than lazily enumerating everything.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -166,13 +203,28 @@ internal void RaiseOutputChanged ()
 			OutputChanged?.Invoke (this, EventArgs.Empty);
 		}
 
-		public BuildOutputDataSource ToTreeDataSource (bool includeDiagnostics)
+
+		public List<BuildOutputNode> GetTreeRootNodes (bool includeDiagnostics)
+		{
+			if (includeDiagnostics) {
```
rodrmoya: I think this method could be merged with the already existing `GetRootNodes`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/ErrorListPad.cs
```diff
@@ -415,6 +420,32 @@ void ShowPopup (Gdk.EventButton evt)
...
+			TreeIter iter, sortedIter;
+			if (view.Model.GetIter (out sortedIter, rows [0])) {
+				iter = filter.ConvertIterToChildIter (sort.ConvertIterToChildIter (sortedIter));
+
+				store.SetValue (iter, DataColumns.Read, true);
+
+				TaskListEntry task = store.GetValue (iter, DataColumns.Task) as TaskListEntry;
+				if (task != null) {
+					await OpenBuildOutputViewDocument ().ConfigureAwait (false);
+					if (task.Severity == TaskSeverity.Error) {
```
rodrmoya: I don't think this `.ConfigureAwait` is correct, as passing `false` to it would run the continuation on another thread, and you probably want it to run on the UI thread, which is where this is called from, and thus avoid context switches when calling Xwt.InvokeAsync in `ExpandNode` above?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/ErrorListPad.cs
```diff
@@ -940,12 +949,19 @@ internal void FocusOutputView ()
...
+
+		async Task OpenBuildOutputViewDocument () 
 		{
 			if (buildOutputViewContent == null) {
 				buildOutputViewContent = new BuildOutputViewContent (buildOutput);
 				buildOutputDoc = IdeApp.Workbench.OpenDocument (buildOutputViewContent, true);
 				buildOutputDoc.Closed += BuildOutputDocClosed;
+
+				await buildOutputViewContent.ProcessLogs (false);
 			} else if (buildOutputDoc != null) {
```
rodrmoya: I think this call to `ProcessLogs` shouldn't happen here, it is an internal thing of the `BuildOutputViewContent`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -40,11 +40,50 @@
 
 namespace MonoDevelop.Ide.BuildOutputView
 {
+	static class BuildOutputNodeExtensions
+	{
+		public static BuildOutputNode SearchFirstNode (this IEnumerable<BuildOutputNode> sender, BuildOutputNodeType type, string search)
```
rodrmoya: Since the data to be searched is the `BuildOutputDataSource` itself and we already have a search API there, I would really consolidate the "search API" into the data source itself. Maybe, instead of keeping the state of the search in the data source, as we do now, there could be just a `FirstMatch` method (with overloads for whatever stuff we want to search on), that just returns a, let's say, `BuildOutputDataSourceSearchResults` (or a shorter name 😄 ) which itself has all the state needed by https://github.com/mono/monodevelop/pull/3847 and the `NextMatch`, `PreviousMatch` API on it.

That is, as mentioned on https://github.com/mono/monodevelop/pull/3847, we should really have a single API/code path for anything we need, not N different ones depending on the case.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/ErrorListPad.cs
netonjm: Right!

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -353,32 +352,41 @@ static void ExpandChildrenWithErrors (TreeView tree, BuildOutputDataSource dataS
...
-						ExpandChildrenWithErrors (treeView, buildOutputDataSource, root);
-					}
-				});
+			processingCompletion = new TaskCompletionSource<object> ();
+
+			Task.Run (async () => {
+				try {
+					await Runtime.RunInMainThread (() => {
+
+						BuildOutput.ProcessProjects ();
```
slluis: This still runs on the UI thread, shouldn't it be done in a background thread?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputViewContent.cs
```diff
@@ -94,6 +96,21 @@ public override void Dispose ()
...
+		}
+
+		internal async Task GoToWarning (string description, string project)
+		{
+			await control.GoToWarning (description, project);
+		}
+
+		internal async Task GoToMessage (string description, string project)
+		{
+			await control.GoToMessage (description, project);
```
slluis: No need to do async/await here, just return what control.GoToMessage() returns.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/ErrorListPad.cs
```diff
@@ -981,4 +995,38 @@ public override void GetSize (Widget widget, ref Gdk.Rectangle cell_area, out in
 			}
 		}
 	}
+
+	public static class TaskListEntryExtensions
```
slluis: Please make this internal.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/ErrorListPad.cs
```diff
@@ -996,7 +996,7 @@ public override void GetSize (Widget widget, ref Gdk.Rectangle cell_area, out in
 		}
 	}
 
-	public static class TaskListEntryExtensions
+	internal static class TaskListEntryExtensions
```
Therzok: internal is redundant.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
```diff
@@ -158,6 +158,55 @@ public FilteredBuildOutputNode (BuildOutputNode master, FilteredBuildOutputNode
 
 	static class BuildOutputNodeExtensions
 	{
+		public static BuildOutputNode SearchFirstNode (this IEnumerable<BuildOutputNode> sender, BuildOutputNodeType type, string search)
```
slluis: I don't see the point of defining those methods as extension methods when you can add them directly to BuildOutputNode.

netonjm: that's a good point

PR #3838, opened 2018-02-12
Fix #560350: Issue with adding a Style to Tags in cshtml views

DavidKarlas:
Problem was that default code completion was triggering code completion on `=` and `"` which user typed but default implementation doesn't know that Xml extension already inserts this values(="") in case of `AutoInsertFragments=true`, hence we have to also specify `KeyActions.Ignore` so editor doesn't insert additional `=` and `"`. This is same behaviour as VS has.


DavidKarlas:
Btw, there is no unit test because there is no way AFAIK to simulate keyboard input on editor.

neerajdeshmukh:
Verified on Visual Studio Enterprise 2017 for Mac (Preview) Version 7.4 Preview (7.4 build 1007).
It is working fine. 

Complete build info:
https://gist.github.com/ganorkargaurav/7e8b187cee1ca7e4abf406efcca829b8

Approved for merge.

PR #3839, opened 2018-02-12
Allow IsValidInContext to return true in projection scenarios.

ToddGrun:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/VsCompletionTextEditorExtension.cs
```diff
@@ -76,13 +77,56 @@ private bool AnyCompletionItemSources (IContentType contentType)
...
+
+			return isValidInContext;
+		}
+
+		private IEnumerable<SnapshotPoint> GetSnapshotPointsAtCaret()
+		{
+			// TODO: Cache the caret and content type information?
+			SnapshotPoint caretPosition = view.Caret.Position.BufferPosition;
+
+			List<SnapshotPoint> snapshotPoints = new List<SnapshotPoint> ();
```
KirillOsenkov: I'd extract GetSnapshotPointsAtPosition() so it's useful not just for the caret, but for any SnapshotPoint

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/VsCompletionTextEditorExtension.cs
```diff
@@ -76,13 +77,56 @@ private bool AnyCompletionItemSources (IContentType contentType)
...
+			snapshotPoints.Add (caretPosition);
+
+			for (int curSnapshotPointIndex = 0; curSnapshotPointIndex < snapshotPoints.Count; curSnapshotPointIndex++) {
+				SnapshotPoint curSnapshotPoint = snapshotPoints[curSnapshotPointIndex];
```
KirillOsenkov: I'd spell out "current", not "cur" :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/VsCompletionTextEditorExtension.cs
```diff
@@ -76,13 +77,56 @@ private bool AnyCompletionItemSources (IContentType contentType)
...
+					snapshotPoints.AddRange (curProjectionSnapshot.MapToSourceSnapshots (curSnapshotPoint));
+				}
+			}
+
+			return snapshotPoints;
+		}
+
+		private IEnumerable<IContentType> GetContentTypesAtCaret()
+		{
+			IEnumerable<SnapshotPoint> snapshotPoints = GetSnapshotPointsAtCaret ();
```
KirillOsenkov: Not sure if we should worry about allocations here. Maybe special-case the for loop above to just collect the content types into a HashSet, since we don't really need the actual points.

KirillOsenkov: But on the other hand these methods are very useful in general.

PR #3841, opened 2018-02-13
[Git,VCS] Fix some task balancing in clone and report message properly

Therzok:


main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlTask.cs
```diff
@@ -56,7 +56,7 @@ protected virtual ProgressMonitor CreateProgressMonitor ()
 				LoggingService.LogError ("Version Control command failed: ", e);
 			} catch (Exception e) {
 				string msg = GettextCatalog.GetString ("Version control operation failed: ");
-				tracker.ReportError (msg, e);
+				tracker.ReportError (msg + e.Message, e);
```
slluis: We should not show the message of unknown exceptions. We are already logging it.

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
```diff
@@ -171,7 +171,7 @@ static bool OnTransferProgress (TransferProgress tp, ProgressMonitor monitor, re
 		{
 			if (progress == 0 && tp.ReceivedObjects == 0) {
 				progress = 1;
-				monitor.BeginTask (GettextCatalog.GetString ("Receiving and indexing objects"), 2 * tp.TotalObjects);
+				monitor.Log.WriteLine (GettextCatalog.GetString ("Receiving and indexing objects"), 2 * tp.TotalObjects);
```
slluis: It's a pity progress is not being properly reported anymore. What was the problem?

Therzok: Libgit2 is not giving balanced events for Transfer/Checkout progress. Even if I make captures to deal with that, we still don't have it split nicely in two.

It used to work until we decided to recursively clone submodules by default, and that's when it breaks.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlTask.cs
Therzok: Then the status bar shows just `Version control operation failed: `. LibGit2Sharp messages are pretty descriptive, so hovering over the status bar to get the full error seems good to me.

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
slluis: ok

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlTask.cs
slluis: That is, catch the exception in the git add-in, and throw it as VersionControlException, then capture VersionControlException here and report the message.

Therzok: VersionControlException wrapping is already done. Added commit which addresses this.

PR #3847, opened 2018-02-13
Structured build output search reb

jtorres:


rodrmoya:
Works great, except for 1 thing, which is that the "x of y" label in the search entry doesn't get updated.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -229,6 +240,79 @@ void UpdatePathBarEntries (PathEntry[] entries)
...
+			var dataSource = treeView.DataSource as BuildOutputDataSource;
+			if (dataSource == null)
+				return;
+
+			Find (dataSource.FirstMatch (searchEntry.Entry.Text));
+		}
+
+		void FindNext (object sender, EventArgs args)
+		{
+			var dataSource = treeView.DataSource as BuildOutputDataSource;
```
rodrmoya: The cause of that different behaviour depending on whether the user uses the buttons in the search entry or CMD-G/SHIFT-CMD-G is that, indeed, there are 2 code paths here. See the already existing `Find/FindNext/FindPrevious` methods, so both cases should really use the same code path. I guess these 2 different implementations of `Find/FindNext/FindPrevious` should be consolidated into a single one, so that both ways to iterate through matches behave the same

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -132,17 +137,20 @@ void Initialize ()
...
 
-					IsSearchInProgress = firstMatch != null;
-				}
-			};
+			searchEntry.Entry.Changed += FindFirst;
+			searchEntry.Entry.KeyReleaseEvent += SearchEntryKeyReleased;
+
+			buttonSearchBackward = new Button ();
+			buttonSearchForward = new Button ();
+			buttonSearchBackward.Clicked += (sender, args) => FindPrevious ();
```
Therzok: ref cycle here.

Therzok: Please use instance handlers and unsubscribe them.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -229,6 +247,79 @@ void UpdatePathBarEntries (PathEntry[] entries)
...
+			var dataSource = treeView.DataSource as BuildOutputDataSource;
+			if (dataSource == null)
+				return;
+			
+			IsSearchInProgress = node != null;
+
+			if (node != null) {
+				MoveToMatch (node);
+
+				resultInformLabel.Text = String.Format (GettextCatalog.GetString ("{0} of {1}"), dataSource.CurrentAbsoluteMatchIndex, dataSource.MatchesCount);
```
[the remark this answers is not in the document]
jtorres: Ok, I agree. This is, however, also present at [0]; shall I get rid of it as well? 

[0] https://github.com/mono/monodevelop/blob/2f3ec4bf6a988b1e7e75034a2a176e6705fcaa2f/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SearchAndReplaceWidget.cs#L838

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -187,6 +188,12 @@ void Initialize ()
 			PackStart (scrolledWindow, expand: true, fill: true);
 		}
 
+		void SearchEntryKeyReleased (object o, Gtk.KeyReleaseEventArgs args)
+		{
+			if (args.Event.Key == Gdk.Key.Return)
+				FindNext ();
```
slluis: I feel you should use the SearchEntry.Activated event for this, not key release.

jtorres: @slluis yep,, that's it.. thanks .. 👍

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -132,17 +137,20 @@ void Initialize ()
...
 
-			searchEntry.Entry.Changed += (sender, e) => {
-				var dataSource = treeView.DataSource as BuildOutputDataSource;
-				if (dataSource != null) {
-					var firstMatch = dataSource.FirstMatch (searchEntry.Entry.Text);
-					if (firstMatch != null) {
-						MoveToMatch (firstMatch);
-					}
+			resultInformLabel = new Label ();
+			searchEntry.AddLabelWidget ((Gtk.Label) resultInformLabel.ToGtkWidget());
```
netonjm: why not Xwt?

PR #3848, opened 2018-02-13
[NUnit] Fix being unable to use NUnit in a .NET Core test project

mrward:
The problem was that the NUnit test provider was finding a
PackageReference for NUnit, trying to load the tests for the .NET Core
project and failing. To fix this the NUnit test provider now ignores
.NET Core projects so the VS Test provider has a chance to load the
tests.

Fixes VSTS #565933


slluis:
With this fix, doesn't it mean that the vsts provider will always be used for NUnit projects? I'm not against it, but we need to make sure it works fine. An alternative solution could be to skip .net core projects in the NUnit add-in.

mrward:
The VSTS provider will be used if the project has a NuGet package that contains a test adapter. If the project just references NUnit then the original NUnit test provider is still used.

mrward:
I did originally think about just skipping .NET Core projects as a way to fix this. I can do that instead if that is a better fix.

slluis:
I think it is a better fix, since the NUnit add-in doesn't really support .NET Core projects.

mrward:
OK I will change it 👍

mrward:
Changed the fix to check for .NET Core.

mrward:
QA approved.

PR #3850, opened 2018-02-13
Azure functions bump master

jstedfast:
* Updates Azure templates to 2.0
* Added dialog for creating new Azure Functions from a template
* Moves Azure Function support to md-addins to allow easier access to proprietary bits


KirillOsenkov:
I think if you merge the latest master into this PR's branch a lot of build errors you're seeing should go away. I've assigned this bug to you: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/566834 because I think this is due to your PR targeting an older master.

mrward:
The failing tests need looking into.

 - AppliesTo_ActivateDeactivateCapability
   - I think this test needs changing so it can work without having to use code from md-addins. It was using the Azure Functions project extension, which was originally in the monodevelop repo, to run a test for the AppliesTo attribute not working after a project was re-evaluated.

 - DotNetCoreProjectTemplate_ExcludeFile
    - These seem to be failing - not sure why the excluded files are being formatted.

mrward:
I will take a look at those failing tests since I wrote them in the first place 😄

jstedfast:
Cool, thanks @mrward

mrward:
Mmm, the AppliesTo test seems more difficult to fix. Using `WorkspaceObject.RegisterCustomExtension` to register the project extension does not fix the test since the extension bypasses the conditions since they are only applied if the project extension is taken from the addin itself. This is probably why I used the AzureFunctions project extension in the test.

One workaround would be to add an internal project extension in MonoDevelop.Core which uses an AppliesTo attribute which is just used for the tests. I do not really like that though.

mrward:
Another idea would be to just move the test to md-addins where there would be an extension that uses the AppliesTo attribute. The Azure Functions addin would use it and there's a feature branch with an extension for Xamarin.Forms that uses the attribute. Ideally though since the code is in MonoDevelop.Core there would be a test in MonoDevelop.Core.Tests.

sgmunn:
@mrward you could create an addin monodevelop that is not bundled, but only used by the tests - which is what you suggested

> Maybe there's a way to create a test addin that is only used by the tests.

We would have to exclude that addin from packaging (or output outside the build folder). This would keep the test next to the code that it's testing, which seems better than being a different repo altogether.

mrward:
Had a go at using a test addin on [another branch](https://github.com/mono/monodevelop/tree/azure-functions-bump-master-fix-capability-test). Building this on Wrench to check.

mrward:
https://github.com/mono/monodevelop/pull/3866

Test is passing on Wrench and VSTS with the above change. The wrench dmg does not contain the test addin. So it seems to work at least.

sgmunn:
@slluis if you can review and approve, i would like to have this in the RTW build of VSM. This will allow us to update the azure functions addin on the gallery once we sort out the issues we have.

slluis:
Please add a description to the PR that explains what this PR is about, what is being changed and why.

PR #3851, opened 2018-02-13
Improve timing display in build output

rodrmoya:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -320,6 +320,20 @@ public TreePosition GetParent (TreePosition pos)
 			return node?.Parent;
 		}
 
+		static string GetDurationAsString (BuildOutputNode node)
```
netonjm: this could be added like an extension method

PR #3852, opened 2018-02-13
Expose an API to register Solution Crawler.

KirillOsenkov:
Since this is internal in Roslyn, we need to expose this to WebTooling and other addins that do not have IVT from Roslyn.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
@@ -1434,10 +1434,6 @@ namespace MonoDevelop.Ide.TypeSystem
 		{
 			var solutionCrawlerRegistrationService = workspace.Services.GetService<ISolutionCrawlerRegistrationService> ();
 			solutionCrawlerRegistrationService.Register (workspace);
-
-			var backgroundCompiler = new BackgroundCompiler (workspace);
-			var backgroundParser = new BackgroundParser (workspace);
-			backgroundParser.Start ();
 		}
 
 		internal static void UnregisterSolutionCrawler (Workspace workspace)
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
@@ -1430,7 +1430,7 @@ namespace MonoDevelop.Ide.TypeSystem
 			return false;
 		}
 
-		static void RegisterSolutionCrawler (Workspace workspace)
+		internal static void RegisterSolutionCrawler (Workspace workspace)
 		{
 			var solutionCrawlerRegistrationService = workspace.Services.GetService<ISolutionCrawlerRegistrationService> ();
 			solutionCrawlerRegistrationService.Register (workspace);
@@ -1440,7 +1440,7 @@ namespace MonoDevelop.Ide.TypeSystem
 			backgroundParser.Start ();
 		}
 
-		static void UnregisterSolutionCrawler (Workspace workspace)
+		internal static void UnregisterSolutionCrawler (Workspace workspace)
 		{
 			var solutionCrawlerRegistrationService = workspace.Services.GetService<ISolutionCrawlerRegistrationService> ();
 			solutionCrawlerRegistrationService.Unregister (workspace);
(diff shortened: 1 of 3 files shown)


KirillOsenkov:
@minestarks

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -1430,6 +1430,21 @@ internal bool TryGetOriginalFileFromProjection (string fileName, int offset, out
 			return false;
 		}
 
+		static void RegisterSolutionCrawler (Workspace workspace)
```
minestarks: Should these be `public`?

KirillOsenkov: I think we have IVT to WebToolingAddin, so things should work. I wouldn't want to introduce a public API since this is hopefully not too longterm.

minestarks: OK but aren't these `private` by default and therefore still uncallable from WebToolingAddin?

KirillOsenkov: ouch!! sorry about that. Good catch. I’ll fix.

KirillOsenkov: Fixed.

slluis:
Can't those methods just be public in VSMac?

KirillOsenkov:
Sure. It's just I don't think this is the shape I would like them to be frozen in. For example, we want control over whether to start BackgroundParser and BackgroundCompiler. Also dispose BackgroundCompiler if started. I didn't want to paint ourselves in the corner :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -1430,6 +1430,21 @@ internal bool TryGetOriginalFileFromProjection (string fileName, int offset, out
 			return false;
 		}
 
+		internal static void RegisterSolutionCrawler (Workspace workspace)
+		{
+			var solutionCrawlerRegistrationService = workspace.Services.GetService<ISolutionCrawlerRegistrationService> ();
+			solutionCrawlerRegistrationService.Register (workspace);
+
+			var backgroundCompiler = new BackgroundCompiler (workspace);
```
Therzok: These will indefinitely leak without disposal. We need to store them on the workspace. Make the methods instance methods and add 2 members.

Are BackgroundCompiler/Parser really needed? SolutionCrawler is already being activated via constructor.

KirillOsenkov: The parser doesn't need disposal. The compiler does, and it's a good point. I'm not sure if the background compiler is needed for diagnostics or not. 

This is not for MonoDevelopWorkspace, this is for a custom workspace in TypeScript (and potentially WebTools).

KirillOsenkov:
@minestarks could you try this without BackgroundParser and BackgroundCompiler? I think Solution Crawler alone should be sufficient for your purposes.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
Therzok: Then register it here.

https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs#L110

Therzok:
@KirillOsenkov We already register the solution crawler is source analysis is enabled:
https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs#L110

Therzok:
I don't see why this is needed.

KirillOsenkov:
As I said above, this is not for MonoDevelop workspace :)

minestarks:
@KirillOsenkov confirming we don't need background compiler / parser

Therzok:
> As I said above, this is not for MonoDevelop workspace :)

Then why put the code in MonoDevelopWorkspace? The code which has the workspace can grab the ISolutionCrawlerRegistrationService and register it.

I wouldn't add this code in MD itself.

minestarks:
@Therzok this came up because `ISolutionCrawlerRegistrationService` is internal and Roslyn doesn't have IVT to our addin.

Therzok:
> this came up because ISolutionCrawlerRegistrationService is internal and Roslyn doesn't have IVT to our addin.

Okay, cool.

@KirillOsenkov can we have these 2 as extension methods somewhere in a `WorkspaceExtensions.cs` file? Keep them internal for now, and let's :shipit: after

KirillOsenkov:
How about this?

PR #3854, opened 2018-02-14
[564512] Fix editable environmental variables

sevoku:
Fixes VSTS #564512


sevoku:
This is already included, closing.

PR #3855, opened 2018-02-14
[MSBuild] Parallelize actual MSBuild builder

Therzok:
Removed a few forced re-evaluation calls. CreateProjectInstance already
forces a re-evaluation if necessary.

The idea behind this is to make use of internal functionality of MSBuild
BuildManager to get builds running.

The only thing we cannot touch in parallel is the actual Project in MSBuild.
But, creating after ProjectInstance we can safely execute a submission
asynchronously, as all the details inside the ProjectInstance are immutable.

So, on the STAThread, we need to setup the project and create the project instance
(which will re-evaluate if necessary).

The actual target run can be done async which we do via ExecuteAsync. Then,
the caller will block on the MRE we set via submission.

To actually be executed in parallel, we need to increase the MaxNodeCount.
That is currently commented out, as MSBuild itself pinvokes CreateProcess
which is a Windows-only method.


Therzok:
Currently breaks cancellation, will need to take a look.

Therzok:
Will revive once https://github.com/Microsoft/msbuild/issues/3397 is resolved

PR #3861, opened 2018-02-15
[Structured build output] Few fixes:

jtorres:
- Implements File -> Save command handler
- Disables / Enabled Save button depending on IsDirty
- Removes SelectRow() and use MoveToMach() instead
- Enable / Disable search buttons depending on result matches
- When not matches, now displays "0 of 0" instead of "Not found"
- Implements missing PathChanged event
- Fixes: Save button should not show up when loading a log from a file (it is already saved)
- Adds a placeholder "Search Build Output";
- When pattern is null or empty, then FirstMatch returns null (This fixes -> Exclude output node itself in the breadcrumbs because they are too long. If the output element is selected, show just a parent in the breadcrumbs.)
- Removes nodes with no children in the pathBar  (Fixes: Exclude output node itself in the breadcrumbs because they are too long. If the output element is selected, show just a parent in the breadcrumbs.)
- Disables auto horizontal scrolling on the treeview

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
@@ -231,8 +231,9 @@ namespace MonoDevelop.Ide.BuildOutputView
 			if (newIndex >= CurrentPath.Length)
 				return;
 
-			if (CurrentPath [newIndex].Tag != null) {
-				MoveToMatch (CurrentPath [newIndex].Tag as BuildOutputNode);
+			var node = CurrentPath [newIndex].Tag as BuildOutputNode;
+			if (node != null && node.HasChildren) {
+				MoveToMatch (node);
 			}
 		}
 
@@ -244,7 +245,8 @@ namespace MonoDevelop.Ide.BuildOutputView
 
 			var stack = new Stack<BuildOutputNode> ();
 
-			stack.Push (selectedNode);
+			if (selectedNode.HasChildren)
+				stack.Push (selectedNode);	
 			var parent = selectedNode.Parent;
 
 			while (parent != null) {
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -188,23 +188,52 @@ void Initialize ()
...
+		}
+
+		FilePath filePathLocation;
+		async Task SaveAs ()
+		{
+			var dlg = new Gui.Dialogs.OpenFileDialog (GettextCatalog.GetString ("Save as..."), MonoDevelop.Components.FileChooserAction.Save) {
+				TransientFor = IdeApp.Workbench.RootWindow,
+				InitialFileName = ViewContentName
+			};
+			dlg.AddFilter (null, "*.binlog");
```
netonjm: reuse binlog constant

netonjm:
LGTM

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -188,23 +192,50 @@ void Initialize ()
...
 
+		async void SaveButtonClickedAsync (object sender, EventArgs e) => await Save ();
+
+		public async Task Save ()
+		{
+			if (filePathLocation == FilePath.Empty) {
+				await SaveAs ();
+			} else {
+				await Save (filePathLocation);
+			}
```
rodrmoya: A .binlog file is not editable, so saving to the already existing file makes no sense. This should always `SaveAs` I think

jtorres: hmm then why do we want to make File -> Save functional?

rodrmoya: for the Build Output view for the current build, which is not a "file", but the output of whatever builders VSfM starts for building the current solution. That is, for the non-static .binlog file case, like the Save button in BuildOutputWidget

PR #3862, opened 2018-02-15
[Refactoring] Add extension point for analyzers to be whitelisted at …

Therzok:
…runtime.

Fixes VSTS #567098 Add an extension point to provide custom diagnostic analyzer assemblies to Workspace

### DIFF
--- a/main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/MonoDevelopWorkspaceDiagnosticAnalyzerProviderService.cs
+++ b/main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/MonoDevelopWorkspaceDiagnosticAnalyzerProviderService.cs
@@ -30,7 +30,9 @@ using System.ComponentModel.Composition;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
+using Mono.Addins;
 using MonoDevelop.Core;
+using MonoDevelop.Core.AddIns;
 
 namespace MonoDevelop.AnalysisCore
 {
@@ -43,9 +45,17 @@ namespace MonoDevelop.AnalysisCore
 		static readonly AnalyzerAssemblyLoader analyzerAssemblyLoader = new AnalyzerAssemblyLoader ();
 		readonly ImmutableArray<HostDiagnosticAnalyzerPackage> hostDiagnosticAnalyzerInfo;
 
+		const string extensionPath = "/MonoDevelop/Refactoring/AnalyzerAssemblies";
+		string [] RuntimeEnabledAssemblies;
 		public MonoDevelopWorkspaceDiagnosticAnalyzerProviderService ()
 		{
+			LoadAnalyzerAssemblies ();
 			hostDiagnosticAnalyzerInfo = CreateHostDiagnosticAnalyzerPackages ();
+		}
+
+		void LoadAnalyzerAssemblies()
+		{
+			RuntimeEnabledAssemblies = AddinManager.GetExtensionNodes<AssemblyExtensionNode> (extensionPath).Select (b => b.FileName).ToArray ();
 		}
 
 		public IAnalyzerAssemblyLoader GetAnalyzerAssemblyLoader ()
@@ -58,7 +68,7 @@ namespace MonoDevelop.AnalysisCore
 			return hostDiagnosticAnalyzerInfo;
 		}
 
-		static ImmutableArray<HostDiagnosticAnalyzerPackage> CreateHostDiagnosticAnalyzerPackages ()
+		ImmutableArray<HostDiagnosticAnalyzerPackage> CreateHostDiagnosticAnalyzerPackages ()
 		{
 			var builder = ImmutableArray.CreateBuilder<HostDiagnosticAnalyzerPackage> ();
 			var assemblies = ImmutableArray.CreateBuilder<string> ();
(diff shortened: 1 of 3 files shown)


Therzok:
Code is currently duplicated, duplication is removed in codefix. I'll rebase after this is merged and send over to QA.

Therzok:
Btw, the code written to support runtime enable/disable of extensions is not doing anything, as the Host Analyzer packages is only called once.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/MonoDevelopWorkspaceDiagnosticAnalyzerProviderService.cs
```diff
@@ -43,11 +45,19 @@ partial class MonoDevelopWorkspaceDiagnosticAnalyzerProviderService : IWorkspace
...
+		string [] RuntimeEnabledAssemblies;
 		public MonoDevelopWorkspaceDiagnosticAnalyzerProviderService ()
 		{
+			LoadAnalyzerAssemblies ();
 			hostDiagnosticAnalyzerInfo = CreateHostDiagnosticAnalyzerPackages ();
 		}
 
+		void LoadAnalyzerAssemblies()
+		{
+			RuntimeEnabledAssemblies = AddinManager.GetExtensionNodes<AssemblyExtensionNode> (extensionPath).Select (b => b.FileName).ToArray ();
```
slluis: You have take into account that extensions can change. Please use AddinManager.ExtensionChanged or AddinManager.AddExtensionNodeHandler to keep the list up to date.

Therzok: I commented above. Roslyn does not reload the host analyzer packages. So it is a NOP, even if we handle it.

Therzok: I used the have code for that, but it's misleading if written. I'll add a comment explaining that roslyn does not refresh that list.

slluis: So if I install an add-in that makes use of this feature, it won't work until the IDE is restarted. Isn't there anything we can do about it? that experience is pretty bad.

KirillOsenkov: I guess this extension point is targeted first and foremost at our own external-addins (webtools, typescript). So there's not a problem since they're bundled. I'm not even sure we should open it up for everyone else at this point, but I'm not sure if there's a way to let ourselves in, but disallow others.

Very impressed that add-ins even allow live updating, without restarting. MEF technically supported recomposition, but it was so full of problems that this capability was never used.

I'm wondering if it would ease things a bit if GetExtensionNodes returned an ObservableCollection that would change and raise events when add-ins are updated.

Therzok: @slluis yep, the service which queries this extension is not a workspace service, but rather a global service:

http://source.roslyn.io/#Microsoft.CodeAnalysis.Features/Diagnostics/DiagnosticAnalyzerService.cs,31

So unless roslyn makes it per workspace, I don't see how to actually make it happen.

Therzok: Most of the things inside roslyn are designed around this being one-time init, so I don't see how we could introduce that.

Not even sure how VSWin handles analyzers from nuget packages in this case.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.addin.xml
```diff
@@ -150,4 +150,18 @@
 		<Assembly file="RefactoringEssentials.dll"/>
 	</Extension>
 
+	<ExtensionPoint path="/MonoDevelop/Refactoring/AnalyzerAssemblies">
+		<ExtensionNode name = "Assembly" type = "MonoDevelop.Core.AddIns.AssemblyExtensionNode" />
```
KirillOsenkov: Not sure you should surround the `=` in the attributes with spaces, it's not done anywhere else. XML looks weird :)

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.addin.xml
```diff
@@ -150,4 +150,18 @@
...
+	</ExtensionPoint>
+
+	<Extension path="/MonoDevelop/Refactoring/AnalyzerAssemblies">
+		<Assembly file="RefactoringEssentials" />
+		<Assembly file="Refactoring Essentials" />
```
KirillOsenkov: Why is there a space here? Space in assembly name?

Therzok: I just copy pasted what was in the previous code.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/MonoDevelopWorkspaceDiagnosticAnalyzerProviderService.cs
slluis: If MEF can't recompose, that's going to be a problem if we start using it as a supported extension mechanism in VSMac.

PR #3867, opened 2018-02-15
[IDE] Fix null reference when code uses Log.WriteLine

iainx:
The TextWriter was using the version of WriteConsoleLogText that takes no monitor. Change it to use the method that takes a monitor, but improve handling for
null monitors.

Fixes VSTS #567563

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs
@@ -553,10 +553,10 @@ namespace MonoDevelop.Ide.Gui.Components
 
 		protected void UnsafeBeginTask (LogViewProgressMonitor monitor, string name, int totalWork)
 		{
-			var indent = monitor.Indent;
-			if (!string.IsNullOrEmpty (name)) {
-				var tag = indent.Indent ();
-				buffer.TagTable.Add (tag);
+			if (!string.IsNullOrEmpty (name) && monitor != null) {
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs
```diff
@@ -567,14 +567,17 @@ protected void UnsafeBeginTask (LogViewProgressMonitor monitor, string name, int
 				UnsafeAddText (null, Environment.NewLine, null);
 			}
 
-			monitor.Marker = buffer.CreateMark (name, buffer.EndIter, false);
+			var marker = buffer.CreateMark (name, buffer.EndIter, false);
+			if (monitor != null) {
```
mrward: Do we need to add a similar null check in UnsafeEndTask?

https://github.com/mono/monodevelop/blob/5f04fee84c43f847b394261fbcdd4dcdfb012c87/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs#L606

mrward: Maybe we don't. If monitor is null then we push a null so it should not go into that if statement.

iainx: Doh, yes

iainx: Done

PR #3868, opened 2018-02-15
Implement "Copy Element Output" popup menu item

rodrmoya:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -172,6 +172,7 @@ void Initialize ()
 			treeView.Accessible.Identifier = "BuildOutputWidget.TreeView";
 			treeView.Accessible.Description = GettextCatalog.GetString ("Structured build output");
 			treeView.SelectionChanged += TreeView_SelectionChanged;
+			treeView.ButtonPressed += TreeView_ButtonPressed;
```
Therzok: This event isn't unsubscribed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -235,6 +236,27 @@ void TreeView_SelectionChanged (object sender, EventArgs e)
...
 
+		void CopyElementMenu_Clicked (object sender, EventArgs e)
+		{
+			var clickedRow = treeView.SelectedRow as BuildOutputNode;
+			if (clickedRow != null) {
+				Clipboard.SetText (clickedRow.ToString (true));
+			}
+		}
+
+		void TreeView_ButtonPressed (object sender, ButtonEventArgs e)
```
Therzok: this can be a static handler. The treeview is the sender here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
```diff
@@ -169,5 +170,26 @@ public static string GetDurationAsString (this BuildOutputNode node)
 
 			return String.Empty;
 		}
+
+		static void ToString (this BuildOutputNode node, bool includeChildren, StringBuilder result, int tabPosition)
+		{
+			result.AppendFormat ("{0}{1}{2}", new String ('\t', tabPosition), node.FullMessage, Environment.NewLine);
```
Therzok: instead of creating a new string on every node, making this O(n) allocations, make it O(logn) by calculating the indent on the parent node.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
rodrmoya: it can't really, because it calls the CopyElementMenu_Clicked, which can't be static because it references the tree view.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -235,6 +236,27 @@ void TreeView_SelectionChanged (object sender, EventArgs e)
...
+		{
+			if (e.Button == PointerButton.Right) {
+				var menu = new Menu ();
+
+				var copyElementMenu = new MenuItem (GettextCatalog.GetString ("Copy Element Output"));
```
mhutch: It's not clear to me what "element" means here

rodrmoya: This comes from the spec, but yes, maybe it's not too clear? what else should we use?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
Therzok: Well, you could set the TreeView as the `ContextMenuItem.Context` and use that in the CopyElementMenu_Clicked handler. But this is not blocking, only do it if this leaks.

rodrmoya: With the move to use ContextMenu, which has a nice overload with a completion handler, the event handler for the menu item is unsubscribed, so I think it should be ok, no leaks

PR #3871, opened 2018-02-16
Fix mutable completion list filtering

mhutch:
The controller's ResetState method reset the StartOffset
to zero. Since this method is called when a mutable list
updates, the CompletionString used for filtering mutable
lists was the entire document up to the completion point,
which of course was so long and complex it matched nothing.

Also remove a redundant ResetSizes call that caused
unnecessary nontrivial work when updating mutable lists.


mhutch:
Requesting backport to 15.6, else PackageReference inteliisense will be broken.

PR #3872, opened 2018-02-16
[d15-6Fix 567563 d156

iainx:


iainx:
Missed the ] key, got the return key instead and created this monstrosity instead

PR #3874, opened 2018-02-16
Adds feature to jump to Error/warning from the BuildOutput treeview +…

netonjm:
Adds feature to jump to Error/warning from the BuildOutput treeview + test

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputViewContent.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputViewContent.cs
@@ -92,6 +92,7 @@ namespace MonoDevelop.Ide.BuildOutputView
 
 		public override void Dispose ()
 		{
+			control.FileSaved -= FileNameChanged;
 			control.Dispose ();
 			base.Dispose ();
 		}
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
@@ -44,18 +44,6 @@ using MonoDevelop.Components.Extensions;
 
 namespace MonoDevelop.Ide.BuildOutputView
 {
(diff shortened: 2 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -191,6 +205,25 @@ void Initialize ()
 			scrolledWindow = new ScrollView { BorderVisible = false };
 			scrolledWindow.Content = treeView;
 
+			treeView.ButtonPressed += delegate (object sender, ButtonEventArgs e) {
```
jtorres: could you use a handler and unsubscribe it?, this class implements IDisposable

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputViewContent.cs
```diff
@@ -46,13 +47,15 @@ public BuildOutputViewContent (FilePath filename)
 			this.ContentName = filename;
 			control = new BuildOutputWidget (filename);
 			control.FileSaved += FileNameChanged;
+			control.TaskSelected += (s, e) => TaskSelected?.Invoke (s, e);
```
Therzok: Unsubcribe events please.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
```diff
@@ -111,6 +111,27 @@ public void AddParameter (string message, string fullMessage)
 			});
 		}
 
+		public BuildOutputNode InverseSearchFirstNode (BuildOutputNodeType type, string search = null)
+		{
+			if (type == NodeType) {
+				if (search == Message || search == null) {
+					return this;
+				} else {
```
rodrmoya: if search is null or empty, it should return no items, right?

netonjm: if search is null, it means we are only searching the first type coincidence

PR #3875, opened 2018-02-16
UI fixes rebased

jtorres:
- Implements File -> Save command handler
- Disables / Enabled Save button depending on IsDirty
- Removes SelectRow() and use MoveToMach() instead
- Enable / Disable search buttons depending on result matches
- When not matches, now displays "0 of 0" instead of "Not found"
- Implements missing PathChanged event
- Fixes: Save button should not show up when loading a log from a file (it is already saved)
- Adds a placeholder "Search Build Output";
- When pattern is null or empty, then FirstMatch returns null (This fixes -> Exclude output node itself in the breadcrumbs because they are too long. If the output element is selected, show just a parent in the breadcrumbs.)
- Removes nodes with no children in the pathBar  (Fixes: Exclude output node itself in the breadcrumbs because they are too long. If the output element is selected, show just a parent in the breadcrumbs.)
- Disables auto horizontal scrolling on the treeview

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
@@ -1,4 +1,4 @@
-﻿//
+//
 // BuildOutput.cs
 //
 // Author:
@@ -423,9 +423,10 @@ namespace MonoDevelop.Ide.BuildOutputView
 				root.Search (currentSearchMatches, currentSearchPattern);
 			}
 
-			if (currentSearchMatches.Count > 0) {
-				currentMatchIndex = 0;
-				return currentSearchMatches [0];
+				if (currentSearchMatches.Count > 0) {
+					currentMatchIndex = 0;
+					return currentSearchMatches [0];
+				}
 			}
 
 			return null;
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -418,14 +418,16 @@ public BuildOutputNode FirstMatch (string pattern)
...
 			currentMatchIndex = -1;
 
-			// Perform search
-			foreach (var root in rootNodes) {
-				root.Search (currentSearchMatches, currentSearchPattern);
-			}
+			if (!string.IsNullOrEmpty (pattern)) {
+				// Perform search
+				foreach (var root in rootNodes) {
+					root.Search (currentSearchMatches, currentSearchPattern);
```
Therzok: Since this wants the first match, why are we searching in the whole tree?

jtorres: @netonjm ^^^

rodrmoya: This should probably use the .SearchFirstMatch method instead, if it just wants the 1st match

PR #3879, opened 2018-02-16
Make dist2

Therzok:
Mirror of #3824 so VSTS builds


directhex:
Can we get this into the branch, to remove the need to wait on the 2017-04 branch (and subsequent release)?

slluis:
Into which branch?

directhex:
Apparently not needed for how the VS release process works. Sorry for noise.

PR #3880, opened 2018-02-16
Simplify keeping track of the current search

rodrmoya:
Instead of creating a `BuildOutputDataSearch` instance for the whole lifetime of the `BuildOutputWidget`, just create it on demand, when a real search is started. And also, use this to keep track of whether we are in the middle of a search or not, instead of having to update a property.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
@@ -289,14 +289,14 @@ namespace MonoDevelop.Ide.BuildOutputView
 			PathChanged?.Invoke (this, new DocumentPathChangedEventArgs (CurrentPath));
 		}
 
-		void FindFirst (object sender, EventArgs args)
+		async void FindFirst (object sender, EventArgs args)
 		{
 			var dataSource = treeView.DataSource as BuildOutputDataSource;
 			if (dataSource == null)
 				return;
 
 			currentSearch = new BuildOutputDataSearch (dataSource.RootNodes);
-			Find (currentSearch.FirstMatch (searchEntry.Entry.Text));
+			Find (await currentSearch.FirstMatch (searchEntry.Entry.Text));
 		}
 
 		public void FindNext (object sender, EventArgs args)
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
@@ -415,7 +415,7 @@ namespace MonoDevelop.Ide.BuildOutputView
 			}, cts.Token);
 		}
 
-		public bool IsSearchInProgress => currentSearch != null;
+		public bool IsSearchInProgress => currentSearch != null && currentSearch.MatchesCount > 0;
 
 		public void FocusOnSearchEntry ()
 		{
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -389,11 +392,9 @@ void ProcessLogs (bool showDiagnostics)
 					BuildOutput.ProcessProjects ();
 
 					await Runtime.RunInMainThread (() => {
-					
-						treeBuildOutputNodes = BuildOutput.GetRootNodes (showDiagnostics);
-						search = new BuildOutputDataSearch (treeBuildOutputNodes);
+						currentSearch = null;
```
jtorres: if we " reset" the `currentSearch` here, we also have to reset UI if there is a `currentSearch` active. Otherwise, it will brake the UI consistency.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -414,7 +415,7 @@ void ProcessLogs (bool showDiagnostics)
 			}, cts.Token);
 		}
 
-		public bool IsSearchInProgress { get; private set; } = false;
+		public bool IsSearchInProgress => currentSearch != null;
```
jtorres: The goal of `IsSearchInProgress` is to enable /disable the command handlers (Serarch Previous / Next) and this (according to specs) should be disabled if there are no matches, regardless the is a current search or not.

rodrmoya: That’s what it still does

jtorres: it does not: 
screenshot 2018-02-19 11 16 42

netonjm: It works but the comparation is odd, currentSearch it must exists allways, there is no need to re-allocate it, only a reset must be enough and a flag true/false it's ok to me.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -213,8 +212,13 @@ internal Task GoToMessage (string description, string project)
...
 		async Task ExpandNode (string project, BuildOutputNodeType nodeType, string message) 
 		{
+			var dataSource = treeView.DataSource as BuildOutputDataSource;
+			if (dataSource == null) {
+				return;
+			}
+
 			await processingCompletion.Task;
-			var projectNode = treeBuildOutputNodes.SearchFirstNode (BuildOutputNodeType.Project, project);
+			var projectNode = dataSource.RootNodes.SearchFirstNode (BuildOutputNodeType.Project, project);
```
netonjm: The purpose of treeBuildOutputNodes is avoid expose the BuildOutputNodes from DataSource which is a class to internally get desired data from a collection to the Tree.

rodrmoya: `treeBuildOutputNodes` = `dataSource.RootNodes` always, so why do we need a private field to store it, while we already have it?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
rodrmoya: right, fixed now, it wasn't updating the search when switching from normal to diagnostics view

PR #3882, opened 2018-02-17
[FSW] Fix FSW on Windows by also catching DllNotFoundException

Therzok:
__Internal is a Mono-ism, it won't work on .NET.

Credits to @AngelCry for finding it.


Therzok:
@mrward yes, you're right. I missed the IsMac check, was going by what was being discussed in gitter.

AngelCry:
where that check are done? i got this exception on Win7 x64, so i guess some check are failed

mrward:
The problem with the file watcher on Windows was fixed in this commit 1005d51e1a35e8a551548899db717696cb20e07d

PR #3883, opened 2018-02-17
[Build Output] Cleanup some well known messages and add tooltips to TreeView

rodrmoya:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/MSBuildOutputProcessor.cs
```diff
@@ -142,24 +143,36 @@ private void BinLog_TaskFinished (object sender, TaskFinishedEventArgs e)
...
 		{
 			if (String.IsNullOrEmpty (e.Message)) {
 				return;
 			}
 
 			switch (e.Message[0]) {
+			case 'S':
+				if (e.Message.StartsWith (SkippingTargetMessagePrefix, StringComparison.Ordinal)) {
+					// "Skipping target ..." messages
+					var parts = e.Message.Split ('"');
```
Therzok: Don't allocate here. Make a scanner, you can also do substring equality without allocating a substring.

Therzok: Same in below cases.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/MSBuildOutputProcessor.cs
```diff
@@ -142,24 +143,36 @@ private void BinLog_TaskFinished (object sender, TaskFinishedEventArgs e)
 
 			EndCurrentNode (stringPool.Add (e.Message), e.Timestamp);
```
Therzok: Several lines above, still adding the same string to the stringpool twice.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/MSBuildOutputProcessor.cs
rodrmoya: "Make a scanner", you mean use .IndexOf + .Compare, which don't do allocations?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/MSBuildOutputProcessor.cs
rodrmoya: yes, this is being fixed in #3880 so not changing it here

Issue #3896, opened 2018-02-19
[ F# ] Remove/review use of locking and Async.RunSynchronously


dsyme:
The implementation of the F# binding has uses of Async.RunSynchronously that look suspicious 

For example this code https://github.com/mono/monodevelop/blob/d4627becd4fef461a2e667ddbac41280d0d80209/main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpTextEditorCompletion.fs#L442 which contains both a very suspicious use of a lock and a nested call to RunSynchronously.

I believe this should be something like this:
```
                let! typedParseResults =
                    asyncMaybe {
                        let! document = documentContext.TryGetFSharpParsedDocument() |> async.Return
                        let! location = document.ParsedLocation |> async.Return
                        let trimmedLine = lineToCaret.TrimEnd()
                        let reparse = trimmedLine.EndsWith("->") || trimmedLine.Contains(").") || trimmedLine.EndsWith("].")
                        if location.Line = context.line && not reparse then
                            LoggingService.logDebug "Completion: got parse results from cache"
                            return! document.TryGetAst() |> async.Return
                        else
                            LoggingService.logDebug "Completion: syncing parse results"
                            // force sync
                            let projectFile = documentContext.Project |> function null -> document.FileName| proj -> proj.FileName.ToString()
                            let! ast = languageService.ParseAndCheckFileInProject(projectFile, document.FileName, 0, editor.Text, true) |> Async.map Some
                            document.Ast <- ast
                            return ast
                    }
```

> VS bug #569293

dsyme:
@nosami Do you happen to know if MD is "good" at sending cancellation requests for activity (e.g. for computing completions) when the results will no longer be needed (e.g. the completions have been dismissed)?  I assume so.  Visual Studio/Roslyn is good at sending these and it's critical to prevent ever-growing request queues.

If so the above feels right. It has the advantage that if the completion request gets cancelled then the ParseAndCheckFileInProject request will also be cancelled.

Therzok:
@dsyme completion requests are cancelled on every keypress. Doesn't seem to happen when manually triggering completion, which is a bug.

Therzok:
Fixed that: https://github.com/mono/monodevelop/pull/3906

nosami:
Fixed by https://github.com/mono/monodevelop/pull/3897

PR #3897, opened 2018-02-19
[F#] Some cleanup and notes

dsyme:
I'm reviewing various parts of the F# binding implementation, here is some cleanup and notes/comments

I need to double-check it before it's fully ready for review


dsyme:
@nosami This also fixes https://github.com/mono/monodevelop/issues/3896

nosami:
Looks good!

PR #3902, opened 2018-02-20
[Structured build output] Solution root

jtorres:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/MSBuildOutputProcessor.cs
```diff
@@ -127,6 +129,29 @@ void BinlogReader_MessageRaised (object sender, BuildMessageEventArgs e)
...
+				}
+
+				if (!string.IsNullOrEmpty (solFileName) && !string.IsNullOrEmpty (config) && !string.IsNullOrEmpty (platform))
+					break;
+			}
+
+			if (CurrentNode.NodeType == BuildOutputNodeType.Build)
+				this.CurrentNode.Message = solFileName;
+			
 			AddNode (BuildOutputNodeType.Project,
```
rodrmoya: there can be a lot of projects in a big solution build, most of which are not children of the root BuildStarted, so I think you could place the `if (CurrentNode.NodeType == BuildOutputNodeType.Build)` at the beginning and move the `foreach` to be inside that if, so that we only process the properties if needed.

netonjm:
Revised and works good!

PR #3909, opened 2018-02-20
[MacPlatform] Window merely minimises when the Close button is pressed

iainx:
A handler in MacPlatform was returning true, which blocks later event handlers
from running. Return false from it instead


slluis:
See https://github.com/mono/monodevelop/pull/3800, looks like a dup.

PR #3910, opened 2018-02-20
[Ide] Groups nodes by solution

jtorres:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -156,18 +156,21 @@ internal void RaiseOutputChanged ()
...
 		{
-			if (includeDiagnostics) {
-				return GetProjectRootNodes ().ToList ();
-			} else {
-				// If not including diagnostics, we need to filter the nodes,
-				// but instead of doing so now for all, we do it on the fly,
-				// as nodes are requested
-				var nodes = new List<BuildOutputNode> ();
-				foreach (var root in GetProjectRootNodes ()) {
+			var comparer = new BuildOutPutNodeComparer ();
```
netonjm: If we use this comparer in each GetRootNodes call, maybe we can instantiate this in the class instead reallocate it

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -156,18 +156,21 @@ internal void RaiseOutputChanged ()
...
+			var nodes = new List<BuildOutputNode> ();
+
+			foreach (var root in GetProjectRootNodes ()) {
+				//let's check if root exists according to solution level comparer
+				if (nodes.Contains (root, comparer)) {
+					var solutionNode = nodes.Find (node => comparer.Equals (node, root));
```
netonjm: why not get the nodes directly and if count > 0 then .. instead do 2 searches

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -156,18 +156,21 @@ internal void RaiseOutputChanged ()
...
+					//now we add its children to existent one since are in the same solution
+					foreach (var child in root.Children)
+						solutionNode.AddChild (child);
+				} else {
 					nodes.Add (new FilteredBuildOutputNode (root, includeDiagnostics));
```
rodrmoya: this breaks normal mode, since you don't seem to be taking into account the `includeDiagnostics` flag when adding the children. The way the `FilteredBuildOutputNode` is implemented is by doing the filtering on the fly, as children are requested, and it needs a `masterNode`, so maybe this could be implemented by adding a `AggregatedBuildOutputNode` which contains a list of all root nodes that are for the same solution, and then just pass that to `FilteredOutputNode` constructor, and do the real aggregation (Start/EndTime and anything else) in the `AggregatedBuildOutputNode`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
rodrmoya: Also, calling `.Contains` and then `Find` is a waste of time, just call `Find` and check for null on its return.

netonjm: I have fear to a Contains call with small strings :-(

rodrmoya:
Seems to work better now, but look what happens on a builder whose only child is a .sln:

screen shot 2018-02-21 at 13 54 13

I guess we'd need to special case this here -> https://github.com/mono/monodevelop/blob/4095dea7671b592f79cafa48ffb233230b26ca35/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/MSBuildOutputProcessor.cs#L153 so that we don't set the Build's node's `Message` to an empty string

rodrmoya:
Also, I'm still not sure adding the children to the `FilteredBuildOutputNode` is going to do what we want in all cases, see the implementation of `Children` in there: https://github.com/mono/monodevelop/blob/4095dea7671b592f79cafa48ffb233230b26ca35/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs#L150

That is, when you create a `FilteredBuildOutputNode` you pass a masterNode, which is what it uses to filter the children and return other properties. But when aggregating more than one `BuildOutputNode` that won't work, as we need to make:

* StartTime => MinOfStartTimeForAllTheAggregatedNodes;
* EndTime => MaxOfStartTimeForAllTheAggregatedNodes;
* HasErrors => AggregatedNodes?.Any (x => x.HasError); (and ditto for HasData and HasWarnings)

So, again, I think we need an intermediate class (`AggregatedBuildOutputNode` for instance) that contains a list of all the aggregated nodes and does what I mentioned below for the properties, and once all are aggregated, pass that to a `FilteredBuildOutputNode`. Or well, something like that, but just adding the children to a `FilteredBuildOutputNode` won't work. In fact, see what happens here:

screen shot 2018-02-21 at 14 07 40

The times are incorrect, as it is just using the timing for the first node that was added to the `FilteredOutputNode`. This same file, on the same branch, looks like this:

screen shot 2018-02-21 at 14 09 53

PR #3911, opened 2018-02-20
[C#] Add language version 7.2 to supported languages UI

Therzok:
Also add an unit test that will fail every time roslyn adds a new language version.

The output of the test is:
1) AssertWeHaveAllLanguagesInUI (MonoDevelop.CSharp.Project.CSharpLanguageVersionHelperTests.AssertWeHaveAllLanguagesInUI)
     Roslyn added new C# language versions: 'CSharp7_2', add to CSharpLanguageVersionHelper.GetKnownLanguageVersions
  Expected: 0
  But was:  1

This will ensure that we never ship without adding these values to the UI.

Fixes VSTS #565692 Development with C# 7.2


Therzok:
Done.

PR #3920, opened 2018-02-21
[Svn] Null check update monitor before trying to write to its log

Therzok:
The code captures , thus it'll use the field on the client for the monitor. Rather than capturing a local, null check the update monitor.
Fixes VSTS #560980 [Feedback] VS for Mac crashes almost instantly when removing a png file from images directory in a shared project under version control (Subversion support extension enabled)


slluis:
Is this a regression? should it be merged in 15.6?

PR #3931, opened 2018-02-22
Fix master build

mrward:
Updated reference to md-addins so it is tracking the correct branch.


mrward:
Not needed - already fixed.

PR #3940, opened 2018-02-22
[Core] Implemented string builder pooling.

mkrueger:
It's just an idea to reduce some memory traffic & object count.
We use StringBuilders all over the place so it may make sense pooling them.

The basic idea is to provide an easy way to get them and to have an implicit dispose on "ToString()" - that avoids the need of manually calling the object pool to "free" the objects.

The string builder usages I do and that I reviewed always work like that:

creation

Add stuff

tostring

<- no more usages ->

So that pattern makes sense IMO - thoughts ?


Therzok:
@mkrueger Can't we just surface this from: http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces/Formatting/StringBuilderPool.cs,8 or directly from the shared pool?

http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces/Utilities/ObjectPools/SharedPools.cs,3a2d3a9983e8d869

mkrueger:
hm, didn't know that roslyn has it - however it's internal API and not all our DLLs have ivt with roslyn assemblies.

Furthermore it uses a an allocate/free model which I want to avoid if possible.

Therzok:
@mkrueger we could just expose a MD.Core layer service like:
```
class SharedPool<T>
{
    <mirror roslyn shared pool API>
}
```

And that way we can reuse their infrastructure without duplicating code (similar to how you made CacheStringBuilder surface StringBuilder API).

I know it's not exactly the same.

Allocate/Free is Rent/Return API, just like yours, but using an Array rather than a Queue.

Therzok:
The benefit of this is that it also has capacity setting. We should consider both.

mkrueger:
Looking at the implementation - their StringBuilderPool ReturnAndFree would be an option. Looking more deeply their ReturnAndFree method seems to be broken because it does not what it should do :).

PR #3944, opened 2018-02-22
[Build Output] Add expand/collapse items to the context menu

rodrmoya:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -319,12 +319,29 @@ void TreeView_ButtonPressed (object sender, ButtonEventArgs e)
...
+				copyElementMenu.Clicked += (s, args) => ClipboardCopy (selectedNode);
 				menu.Items.Add (copyElementMenu);
 
+				var expandElementMenu = new ContextMenuItem (GettextCatalog.GetString ("Expand Element"));
+				expandElementMenu.Clicked += (s, args) => treeView.ExpandRow (selectedNode, false);
+				menu.Items.Add (expandElementMenu);
+
+				var collapseAllMenu = new ContextMenuItem (GettextCatalog.GetString ("Collapse All"));
+				collapseAllMenu.Clicked += (s, args) => {
+					var dataSource = treeView.DataSource as BuildOutputDataSource;
```
Therzok: Can the data source be anything else? If not, make it a direct cast.

rodrmoya: It shouldn't in this code path, as this is only called when we open the menu for a real row from the data source.

PR #3948, opened 2018-02-23
Adds feature to show errors and warnings when node is collapsed

netonjm:
Adds feature to show errors and warnings when node is collapsed
Fixes #570136
Fixes #570560

image

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
@@ -229,12 +229,12 @@ namespace MonoDevelop.Ide.BuildOutputView
 		}
 
 		public override int WarningCount {
-			get => nodes.FirstOrDefault ()?.WarningCount ?? 0;
+			get => nodes.Sum (x => x.WarningCount);
 			set => throw new NotImplementedException ();
 		}
 
 		public override int ErrorCount {
-			get => nodes.FirstOrDefault ()?.ErrorCount ?? 0;
+			get => nodes.Sum (x => x.ErrorCount);
 			set => throw new NotImplementedException ();
 		}
 	}
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
@@ -50,7 +50,6 @@ namespace MonoDevelop.Ide.BuildOutputView
 
 	class BuildOutputNode : TreePosition
 	{
-
 		const string ParametersNodeName = "Parameters";
 
 		public virtual BuildOutputNodeType NodeType { get; set; }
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
```diff
@@ -228,6 +232,21 @@ public void AddNode (BuildOutputNode node)
 			get => nodes.Any (x => x.HasData);
 			set => throw new NotImplementedException ();
 		}
+
+		public override int Warnings {
+			get => nodes.First ()?.Warnings ?? 0;
```
Therzok: What you want is `FirstOrDefault` here.

Therzok: Same below.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
```diff
@@ -70,6 +70,10 @@ class BuildOutputNode : TreePosition
 		public virtual string Project { get; set; }
 		public virtual int LineNumber { get; set; }
 
+		public virtual bool ShowsInformation { get; set; }
+		public virtual int Errors { get; set; }
```
rodrmoya: I'm not sure I get what this `ShowsInformation` is used for, is it to check if this node should display the errors/warnings counts when collapsed? If so, this doesn't belong to the `BuildOutputNode`, i.e. the data model, but to the cell renderer, the view.

rodrmoya:
We only display the errors/warnings indicator if there are errors/warnings, so looks a bit weird when nodes have only errors or warnings, but not both:

screen shot 2018-02-23 at 15 21 57

Should we always display both even when there are 0 errors or 0 warnings?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
netonjm: ahh yes :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
rodrmoya: Also, I think these should be named ErrorsCount and WarningsCount?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
```diff
@@ -225,6 +227,16 @@ public void AddNode (BuildOutputNode node)
...
+		public override int WarningCount {
+			get => nodes.FirstOrDefault ()?.WarningCount ?? 0;
+			set => throw new NotImplementedException ();
+		}
+
+		public override int ErrorCount {
+			get => nodes.FirstOrDefault ()?.ErrorCount ?? 0;
+			set => throw new NotImplementedException ();
+		}
 	}
```
rodrmoya: this doesn't affect this branch, as the icons are only shown for children, but the implementation of these 2 is wrong, it should be the sum of the WarningCount/ErrorCount of all the nodes, not just of the 1st node

PR #3950, opened 2018-02-23
Native Structured Build Output – Fix native context menu

sevoku:
This includes some ContextMenu API changes

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/ContextMenu.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/ContextMenu.cs
@@ -89,16 +89,15 @@ namespace MonoDevelop.Components
 			ContextMenuExtensionsGtk.ShowContextMenu (parent, x, y, this, closeHandler, selectFirstItem);
 		}
 
-		public void Show (Xwt.Widget parent, int x, int y, Action closeHandler)
+		public void Show (Xwt.Widget parent, int x, int y, Action closeHandler, bool selectFirstItem = false)
 		{
-			var gtkWidget = parent.Surface.NativeWidget as Gtk.Widget;
-			if (gtkWidget != null) {
-				Show (gtkWidget, x, y, closeHandler);
+			if (parent.Surface.NativeWidget is Gtk.Widget widget) {
+				Show (widget, x, y, closeHandler, selectFirstItem);
 				return;
 			}
 			#if MAC
-			if (Platform.IsMac && parent.Surface.NativeWidget is AppKit.NSView) {
-				ContextMenuExtensionsMac.ShowContextMenu ((AppKit.NSView)parent.Surface.NativeWidget, x, y, this, closeHandler, false);
+			if (parent.Surface.NativeWidget is AppKit.NSView view) {
+				ContextMenuExtensionsMac.ShowContextMenu ((AppKit.NSView)view, x, y, this, closeHandler, selectFirstItem);
 				return;
 			}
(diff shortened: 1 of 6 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components/ContextMenu.cs
```diff
@@ -60,12 +60,12 @@ public void CollapseSeparators ()
 				Items.RemoveAt (Items.Count - 1);
 		}
 
-		public void Show (Gtk.Widget parent, Gdk.EventButton evt)
+		public void Show (Control parent, Gdk.EventButton evt)
```
Therzok: This is public API change.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Control.cs
```diff
@@ -91,7 +91,7 @@ public T GetNativeWidget<T> ()
 			if (nativeWidget is T)
 				return (T)nativeWidget;
 			else
-				throw new NotSupportedException ();
+				return (T)ConvertToType (typeof (T), nativeWidget);
```
Therzok: Every time this is called, it'll create a new embed.

sevoku: Yes, that was bad. Reverted and fixed with 9201fdf.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/ContextMenu.cs
sevoku: Yes it is, we need a better solution if we don't allow API changes for the next release.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/ContextMenu.cs
```diff
@@ -89,12 +89,32 @@ public void Show (Gtk.Widget parent, int x, int y, Action closeHandler, bool sel
 			ContextMenuExtensionsGtk.ShowContextMenu (parent, x, y, this, closeHandler, selectFirstItem);
 		}
 
+		public void Show (Xwt.Widget parent, int x, int y, Action closeHandler)
+		{
+			var gtkWidget = parent.Surface.NativeWidget as Gtk.Widget;
```
Therzok: I'd prefer `is` pattern matching like:
```
if (parent.Surface.NativeWidget is Gtk.Widget widget) {
    Show (gtkWidget, x, y, closeHandler);
    return;
}
#if MAC
if (parent.Surface.NativeWidget is AppKit.NSView view) {
    Show (view, x, y, closeHandler);
    return;
}
#endif
throw new NotSupportedException ();
```

Or using a switch with case patterns.

sevoku: right, thanks!

sevoku:
Actually this still breaks the API. It's not possible to pass a Control because of the ambiguous call, so it also breaks the unit tests: https://github.com/mono/monodevelop/blob/structured-build-output-native-xwt-context-menu-fix/main/src/addins/MonoDevelop.UnitTesting/Services/AbstractUnitTestEditorExtension.cs#L274

sevoku:
API break is fixed now: https://github.com/mono/monodevelop/pull/3950/commits/9e8620c3587af9ffc2d1d6b2b5ead37fec2868b6

PR #3956, opened 2018-02-26
Icon refactoring

vancura:
*DON'T MERGE YET*

We have a huge chaos in file and folder icons. I am trying to clean it up here.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/ExtensionModel/StockIcons.addin.xml
+++ b/main/src/core/MonoDevelop.Ide/ExtensionModel/StockIcons.addin.xml
@@ -140,7 +140,6 @@
 	<StockIcon stockid="md-file-unit-test" resource="file-unit-test-32.png" size="Dnd" />
 	<StockIcon stockid="md-find-next" resource="find-next-16.png" size="Menu" />
 	<StockIcon stockid="md-find-prev" resource="find-prev-16.png" size="Menu" />
-	<StockIcon stockid="md-folder-assets" resource="folder-assets-16.png" size="Menu" />
 	<StockIcon stockid="md-reference-assembly" resource="reference-assembly-16.png" size="Menu" />
 	<StockIcon stockid="md-reference-folder" resource="reference-assembly-16.png" size="Menu" />
 	<StockIcon stockid="md-fs-field" resource="element-fs-field-16.png" size="Menu" />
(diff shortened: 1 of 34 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/StockIcons.cs
```diff
@@ -42,9 +41,6 @@ public class Stock
 		public static readonly IconId CloseAllDocuments = "md-close-all-documents";
 		public static readonly IconId CloseCombine = "md-close-combine-icon";
 		public static readonly IconId CloseIcon = Gtk.Stock.Close;
-		public static readonly IconId ClosedFolder = "md-closed-folder";
-		public static readonly IconId ClosedReferenceFolder = "md-closed-reference-folder";
-		public static readonly IconId ClosedResourceFolder = "md-closed-resource-folder";
```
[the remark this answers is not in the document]
vancura: OK.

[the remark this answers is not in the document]
vancura: @mrward See the last commit.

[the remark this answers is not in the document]
mrward: 👍

vancura:
Could you please review this PR? I'd like to know if this is fine before I do the same thing in `md-addins`. Thanks!

main/src/core/MonoDevelop.Ide/ExtensionModel/StockIcons.addin.xml
```diff
@@ -296,6 +240,60 @@
...
+	<StockIcon stockid="md-reference-folder" resource="folder-component-16.png" size="Menu" />
+	<StockIcon stockid="md-resource-folder" resource="folder-resource-16.png" size="Menu" />
+	<StockIcon stockid="md-solution-folder" resource="folder-solution-16.png" size="Menu" />
+	<StockIcon stockid="md-component-folder" resource="folder-component-16.png" size="Menu" />
+	<StockIcon stockid="md-assets-folder" resource="folder-assets-16.png" size="Menu" />
+	<StockIcon stockid="md-code-folder" resource="folder-code-16.png" size="Menu" />
+	<StockIcon stockid="md-services-folder" resource="folder-services-16.png" size="Menu" />
+
+	<!-- File icons -->
+	<StockIcon stockid="md-generic-file" resource="file-generic-16.png" size="Menu" />
```
sevoku: why the pattern (`md-file-*` -> `md-*-file`) change? Imo we should use the same pattern like we have for other categories (`md-prefs-*`, `md-file-*` etc.). Also the Stock IDs belong to the public API too, kind of. AddIns can reference and use them directly (instead of `Gui.Stock`). I remember that we changed some id's before and it was really hard for AddIn devs to track that, since we don't expose ALL icons throught `Gui.Stock`. @slluis?

vancura: Right, I can use `md-folder-*` and `md-file-*`. The reason why I did it was that there's a huge chaos in IDs. Take a look: sometimes they're reversed, sometimes theres an `-icon` suffix. A mess.

vancura: I might be wrong, but I really think we should clean up these IDs if we want to have more extensions in the marketplace. There should be a clear pattern, or it will look like they're glued together from different sources (which they in fact are). We might break some APIs today, but if we design a pattern now, it will pay off later in the longer run.

sevoku: I'm not against cleaning them up, I'm just for a unified pattern and for not breaking AddIns that use the IDs directly. And we should definitely remove the `-icon` suffix everywhere. If we don't care about the IDs to be changed without an API bump, then it's ok, otherwise we should sync it with the next API bump.

vancura: If @slluis agrees to rename icon IDs, I'd make sure we have same pattern for all names. Plus I'd document it so it's super-clear.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/StockIcons.cs
```diff
@@ -169,5 +154,35 @@ public class Stock
...
+		public static readonly IconId ReferenceFolder = "md-reference-folder";
+		public static readonly IconId ResourceFolder = "md-resource-folder";
+		public static readonly IconId SolutionFolder = "md-solution-folder";
+		public static readonly IconId ComponentFolder = "md-component-folder"; // Unused in monodevelop
+		public static readonly IconId AssetsFolder = "md-assets-folder"; // Unused in monodevelop
+		public static readonly IconId CodeFolder = "md-code-folder";
+		public static readonly IconId ServicesFolder = "md-services-folder";
+
+		// Deprecated folder icons
+		public static readonly IconId ClosedFolder = "md-folder";
```
Therzok: Nit: To make this easier in the future, in case we do other renames, I'd say the pattern should be:
```
public static readonly IconId Folder = "md-folder";
public static readonly IconId ReferenceFolder = "md-reference-folder";

...

public static readonly IconId ClosedFolder = Folder;
public static readonly IconId OpenFolder = Folder;
public static readonly IconId ClosedReferenceFolder = ReferenceFolder;
```

vancura: Yes. Sure, thanks!

PR #3976, opened 2018-02-26
[TextEditor] Always fetch the preedit string

bratsche:
When the window first focuses imContextNeedsReset is always set to true,
so here we set preeditString to null and it won't be displayed correctly
for Japanese and Chinese users. There may be other situations where it
also is set to true and the preedit string will disappear in the middle
of typing.

Fixes VSTS #573925

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
@@ -482,10 +482,7 @@ namespace Mono.TextEditor
 
 		void PreeditStringChanged (object sender, EventArgs e)
 		{
-			if (imContextNeedsReset)
-				preeditString = null;
-			else
-				imContext.GetPreeditString (out preeditString, out preeditAttrs, out preeditCursorCharIndex);
+			imContext.GetPreeditString (out preeditString, out preeditAttrs, out preeditCursorCharIndex);
 			if (!string.IsNullOrEmpty (preeditString)) {
 				if (preeditOffset < 0) {
 					preeditOffset = Caret.Offset;


bratsche:
@sevoku @manish I haven't gotten the QualityAssurance test running here locally, but in Slack Sevo said he thinks this is safe to commit so I went ahead and assigned reviewer and stuff.

sevoku:
@bratsche could you add a ref to the other commit 3cfb1b3d13ceb04017a6ecb19a03667e7cc5508a to the commit message pls? Something like:
> Reverts commit 3cfb1b3d13ceb04017a6ecb19a03667e7cc5508a

@manish could QA run the tests with and without the workaround and file bugs for both cases?

bratsche:
@sevoku Thanks, I pushed an update to the commit

alanmcgovern:
Wouldn't this just re-introduce the bug the original patch was trying to fix?

mkrueger:
That's the original bug that the "if" fixed : 

https://bugzilla.xamarin.com/show_bug.cgi?id=59642

Please ensure that this doesn't happen in that patch

sevoku:
@alanmcgovern @mkrueger we have discussed this on Slack with @manish already. The bug was reported by QA for automated tests only. We never had customer reports about the same. This happens only when the automated test uses our integrated automation framework (https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest.Results/GtkWidgetResult.cs#L274-L413), but not with an external native tool `clickclick` (which they use as a workaround for that issue). Persumably the problem is the way how we inject Gdk events and how it plays together with the IM on mac.

neerajdeshmukh:
Verified on Visual Studio Enterprise 2017 for Mac Version 7.4 (build 1035).
We are able to see Keyword after giving input from keypad. As scope of this ticket to track visibility of character only, Marking as QA-Approved.
Screencast: https://www.screencast.com/t/XHgBZrLxIdz
Build Gist: https://gist.github.com/neerajdeshmukh/2c3ebe5ec29945b13fc0f44a699f9b8d

After switching from any other window, character is not visible in my observation as well in this build.
But as per comment from Ashley, this is being tracked under https://devdiv.visualstudio.com/DevDiv/_workitems/edit/569768 and https://devdiv.visualstudio.com/DevDiv/_workitems/edit/573925.

Hence marking as QA-Approved as scope of the ticket is limited to visibility of character after providing input only as perf Cody comment on bug.

Approved for Merge.

Issue #3929, opened 2018-02-22
Make 'the file has changed outside the IDE' be globally re-usable


alanmcgovern:
https://github.com/mono/monodevelop/blob/34b73f265af47bc7ed1c314c45d6dbdc33745018/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SourceEditorWidget.cs#L705-L715

How much work would it take to make this globally reusable so we can have the IDE automatically display this for things like the Android Designer, iOS Designer, Forms Previewer? I don't think we should be inventing our own mechanism to handle external file changes :)

/cc @paymicro

> VS bug #571163

Therzok:
@alanmcgovern I see a few options here:
* A ViewContent derived abstract implementation in MonoDevelop which has built-in support for the the MessageBar that users can derive from
* A InfoBar derived implementation that handles this specific abstraction and can be embedded at the top of your Box. (you would still need to handle the presentation)

alanmcgovern:
My main goal is for us to *not* invent our own UI/UX for this, or to try to reimplement the whole experience in our own codebase.

What i'm hoping is (more or less) for us to be able to hook into an event and then call a method to show the IDE specific UI, which we'll then act on when the user clicks a button.

As such your first option is best, but the second option is equally OK. If we have a unified/shareable widget *and* we have a notification from the IDE, i'm 100% ok with handling the presentation of the actual 'your files have changed' widget.

mkrueger:
My2cents : provide a viewcontent and that's it :)

alanmcgovern:
We have a ViewContent already, but having to choose which ViewContent we subclass in order to activate different IDE features seems like it's not going to be scaleable.

Why can't this be enabled for any ViewContent?

mkrueger:
@alanmcgovern I do not really understand what you mean  - which ViewContent should you subclass ?

alanmcgovern:
Was this actually fixed or was the issue accidentally closed?

mkrueger:
PR bit rotted - I closed it. Was never taken. But now with all the infrastructure changes it needs to be redone if we still want that.

PR #3979, opened 2018-02-27
Fixes issue #3929 Make 'the file has changed outside the IDE' be

mkrueger:
globally re-usable https://github.com/mono/monodevelop/issues/3929

View contents need to inherit from InfoBarViewContent instead to make
use of that. The AutoSaveWarning doesn't show up automatically but the
file changed warning should work out of the box.


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/InfoBarViewContent.cs
```diff
@@ -0,0 +1,224 @@
...
+		public abstract Gtk.Widget ContentControl { get; }
+
+		void EnsureVBoxIsCreated ()
+		{
+			if (vbox != null)
+				return;
+			vbox = new VBox ();
+			vbox.SetSizeRequest (32, 32);
+			vbox.Accessible.SetShouldIgnore (true);
+			Console.WriteLine ("control:" + ContentControl);
```
Therzok: Debug code.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/InfoBarViewContent.cs
```diff
@@ -0,0 +1,224 @@
...
+		const uint CHILD_PADDING = 0;
+
+		VBox vbox;
+		InfoBar infoBar;
+
+		public bool IsInfoBarVisible {
+			get { return infoBar != null; }
+		}
+
+		public sealed override Control Control { get { EnsureVBoxIsCreated (); return vbox; } }
```
Therzok: I'd rewrite this as:
```
if (vbox == null)
    CreateVBox();
return vbox;
```

mkrueger: if null check is done in EnsureVBoxIsCreated - and no Control isn't called frequently so that a method call does something :)

nosami:
>  or we make the info bar reusable for showing custom messages

I'd like this for F#

mkrueger:
Generalizing the info bar made things much easier :)

slluis:
The InfoBar class should not be used directly by the public api, since it is a GTK widget, and we may want to switch to native at some point. I'd like to have an InfoBar class that wraps the widget, very much like what we with DocumentToolbar.

Also, instead of ShowInfoBar/RemoveInfoBar methods, I'd like to have an InfoBar property in ViewContent, with Show/Hide methods.

mkrueger:
I don't think it's good to expose the infobar as a property - it gives too much responsibility to that and I think it's good to follow the demeter law here.

The show/hide methods work only on the view content and not on the info bar - another reason not doing it.

Maybe I just make that internal - ATM we don't have another implementation for that.

mkrueger:
Show/Hide on info bar is the wrong place - it's an operation on the view content and not on the control - and it's not needed IMO.

It's visibile when the property is set and removed when it's == null.

slluis:
I don't want InfoBar creation to be delegated to ViewContent subclasses. I want to model it as a service that ViewContent can provide. So if you what to show a message in an info bar, it should be as easy as this:

``` csharp
public override Task Load (FileOpenInformation fileOpenInformation)
{
    // ... do stuff
    if (someWarning)
        InfoBar.ShowMessage ("Something went wrong");
    // ...
}

public override Task Save (FileSaveInformation fileSaveInformation)
{
    // do stuff, then hide message if visible
   InfoBar.Hide ();
}
```

That is: no need to explicitly create the info bar, ViewContent does it on demand. The actual widget is only created when ShowMessage is invoked (the InfoBar property returns a wrapper), and destroyed when calling Hide. The InfoBar property would be read-only.

mkrueger:
The info bar content is more complex it adds buttons, combo boxes and so on it's not about messages.

slluis:
I posted just an example of api use. We would obviously add support for more complex content, in a similar way we already do it for DocumentToolbar. MesageService is for showing global notifications. InfoBar is for showing a message embedded in a WebView, it is a very different feature.

mkrueger:
Took some time today to work on the API.

paymicro:
If the `InformationArea` is shown then all hiding elements is shown too.
Screencast: http://recordit.co/pome7jWe8g

mkrueger:
Should be fixed

mkrueger:
So any feedback for that ?

paymicro:
LGTM
works as it should.

alanmcgovern:
Is this ready to merge? Is there something we're blocked on?

slluis:
It's blocked on me, I need to review it again.

slluis:
Not as is. I'd like some changes. Please leave the PR open, I'll post my comments.

mkrueger:
closing - that's outdated.

PR #3980, opened 2018-02-27
Master issue3978

mkrueger:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/LineSeparatorTextEditorExtension.cs
```diff
@@ -92,16 +93,15 @@ void OptionsChanged (object sender, EventArgs e)
...
-				var line = Editor.GetLineByOffset (s.Start);
-				var marker = Editor.TextMarkerFactory.CreateLineSeparatorMarker (Editor);
-				Editor.AddMarker (line, marker);
-				newMarkers.Add (marker);
-			}
-
-			RemoveMarkers ();
-			markers = newMarkers;
+			await Runtime.RunInMainThread (delegate {
+				RemoveMarkers ();
```
slluis: Unrelated fix, it should be in another PR. But in any case, I don't think it is correct. The DocumentContext_DocumentParsed event should be fired in the UI thread.

mkrueger: Doesn't matter where it's fired - the await above can run on another thread.

slluis: > Doesn't matter where it's fired - the await above can run on another thread causing this code to run in background.

No, await will capture the current context and will switch back to it after the async call. So if the even is raised in the UI thread, execution after await will continue in the UI thread, no matter what the async call does.

mkrueger: Then I don't have a good explaination for that stack trace - I assumed that the switch back doesn't work for some reason.

mkrueger:
Found & fixed the cause - however the original approach fixed the bug as well - but now the event should always run on the UI thread.

PR #3983, opened 2018-02-27
Fixes VSTS 569813 'Find in Files' searches within hidden MacOS .DS_St…

mkrueger:
…ore files.

https://devdiv.visualstudio.com/DevDiv/_workitems?id=569813

The binary & hidden files flag was not used in the UI in any way so it's better to remove it. For VS4Mac it doesn't make sense to include non source code files so they're not useful at all.
The default is not to search in binary or hidden files - that fixes the .DS_Store issue as well.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.cs
@@ -39,11 +39,6 @@ namespace MonoDevelop.Ide.FindInFiles
 {
 	public abstract class Scope
 	{
-		public bool IncludeBinaryFiles {
-			get;
-			set;
-		}
-
 		public virtual PathMode PathMode {
 			get {
 				var workspace = IdeApp.Workspace;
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.cs
```diff
@@ -39,11 +39,6 @@ namespace MonoDevelop.Ide.FindInFiles
 {
 	public abstract class Scope
 	{
-		public bool IncludeBinaryFiles {
-			get;
-			set;
-		}
-
 		public virtual PathMode PathMode {
```
mrward: Public API break here. Mark it as obsolete instead?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.cs
```diff
@@ -280,10 +275,6 @@ public class DirectoryScope : Scope
 			get { return PathMode.Absolute; }
 		}
 
-		public bool IncludeHiddenFiles {
-			get;
-			set;
-		}
```
mrward: Another API break here. Mark it as obsolete instead?

PR #3984, opened 2018-02-27
Structured build output summary node

rodrmoya:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
```diff
@@ -93,6 +94,8 @@ public Xwt.Drawing.Image GetImage ()
 			switch (NodeType) {
 			case BuildOutputNodeType.Build:
 				return Resources.BuildIcon;
+			case BuildOutputNodeType.BuildSummary:
+				return HasErrors ? Resources.ErrorIcon : Resources.TaskIcon;
```
netonjm: If we are going to reuse GetImage for other things, this condition must be moved to the cell renderer, because not there is no way to get the real TaskIcon if HasErrors = true

rodrmoya: You are calling .GetImage from the cell renderer, right?

netonjm: yep!

rodrmoya: Then yes, no need to do this on the cell renderer, the `BuildOutputNode` knows what icons it needs to return for each `NodeType`, right?

PR #3985, opened 2018-02-27
Editor features

DavidKarlas:


slluis:
Yay!

dsyme:
It looks like [this change](https://github.com/mono/monodevelop/commit/6372601cf428eb01905f59b4c2f99aa70c3ba1d5#diff-0524bda4b72240c35996595524f79113R260) which is part of this PR means that an exception is now being raised when Roslyn Workspaces project information is not available, e.g. for F# projects, see https://github.com/mono/monodevelop/issues/4062.

The Android designer depends on null being raised for this condition, and is prepared to handle it.

Is it ok if we remove the throw of the exception and return null in this method as before?  The corresponding `GetCodeAnalysisProject` just before `GetCodeAnalysisProjectAsync` does return null in this case.

slluis:
@mkrueger see Don's comment.

mkrueger:
Y it is ok - I don't mind in that case - null is something that can be expected there.

PR #3986, opened 2018-02-27
[DotNetCore] Support .NET Core SDK 2.1.300 project templates

mrward:
- If .NET Core SDK 2.1.300 preview is installed then project templates for that version of the SDK will be available in the New Project dialog.
- The project templates installed with the .NET Core SDK 2.1 preview are used instead of having them with the DotNetCore and AspNetCore addins.
- Included one change from the aspnetcore-https-support branch around handling the application url otherwise it is not possible to run the ASP.NET Core projects.

VSTS #571396


slluis:
This has been accepted. Let's fix the conflicts, merge, and backport to 15.7.

mrward:
@monojenkins backport d15-7

PR #3990, opened 2018-02-28
[Ide] Ensured that the document parsed event is called on UI thread.

mkrueger:
If ReparseDocument was called from a background thread the parsed event was called on the same thread. This should be fixed.

Fixes VSTS 572820


GLjackyvaswani:
Have tested the bug on to the build provided, issue is seems to be fixed.

Tested on : Version 7.4 (build 1037)

Build taken from : https://devdiv.visualstudio.com/_apis/resources/Containers/1819563?itemPath=drop%2FVisualStudioForMac-7.4.0.1037.dmg

Screencast : https://www.screencast.com/t/LrH9IOywnn

PR #3995, opened 2018-02-28
Add ImageService.TryGetImageId and put imageid's on two StockIcons th…

ToddGrun:
…at webtooling uses in completion

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ImageService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ImageService.cs
@@ -192,6 +192,21 @@ namespace MonoDevelop.Ide
 			AddIcon (iconId, icon);
 		}
 
+		public static bool TryGetImageId(string stockId, out ImageId imageId)
+		{
+			foreach(var pair in imageIdToStockId)
+			{
+				if (pair.Value == stockId)
+				{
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ImageService.cs
```diff
@@ -192,6 +192,21 @@ public static void AddImage (ImageId imageId, Xwt.Drawing.Image icon)
...
+			foreach(var pair in imageIdToStockId)
+			{
+				if (pair.Value == stockId)
+				{
+					imageId = pair.Key;
+					return true;
+				}
+			}
+
+			imageId = default(ImageId);
```
decriptor: As a heads up, I believe that `ImageId` is changing to `AccessibleImageId`

ToddGrun: Yep, would like to see that changed.

slluis:
+1 on the reverse lookup dictionary

Therzok:
The reason I had asked for a reverse lookup is that the current number of image ids is way lower than the current number of stock ids, so it's wasteful to look through stock icons without image ids.

Therzok:
👍

PR #4003, opened 2018-03-01
[StaticRegistrar] Allow libvsmregistrar.dylib to not exist

Therzok:
The reason is that the designers install a custom addin version
into the root of the IDE, caused a mismatch between the dll and
the static registrar generated for it

### DIFF
--- a/main/build/MacOSX/monostub.mm
+++ b/main/build/MacOSX/monostub.mm
@@ -293,16 +293,6 @@ run_md_bundle_if_needed(NSString *appDir, int argc, char **argv)
     }
 }
 
-static void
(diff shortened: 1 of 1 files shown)


alanmcgovern:
I'd like to take this as an immediate solution to let us run our integration tests again :)

garuma:
With the addition of the environment variable we can now easily disable the static registrar thus unblocking us while waiting for improvement in the static-registrar itself to support our case better.

Working designer build: https://devdiv.visualstudio.com/DevDiv/Xamarin/_build/index?buildId=1431145&_a=summary

PR #3923, opened 2018-02-21
Update to a newer SQLitePCLRaw

directhex:
This version includes ARM builds of libe_sqlite3.so

PR #4004, opened 2018-03-01
Update to a newer SQLitePCLRaw

Therzok:
This version includes ARM builds of libe_sqlite3.so

Replaces https://github.com/mono/monodevelop/pull/3923


Therzok:
As a note, this version doesn't need binding redirects for roslyn since the assembly version is always 1.0.0.0: https://github.com/ericsink/SQLitePCL.raw/blob/d86a5b5311ff4cdbba54cda49c23a776bd888663/src/cs/AssemblyInfo.cs

Future versions will.

PR #4005, opened 2018-03-01
Changes the order of the context menu and some labels

netonjm:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads/ErrorListPad.cs
```diff
@@ -378,19 +378,15 @@ void ShowPopup (Gdk.EventButton evt)
...
-			var copy = new ContextMenuItem (GettextCatalog.GetString ("_Copy"));
-			copy.Clicked += OnTaskCopied;
-			menu.Add (copy);
-
 			var goBuild = new ContextMenuItem (GettextCatalog.GetString ("Go to Log"));
 			goBuild.Clicked += async (s, e) => await OnGoToLog (s, e);
 			menu.Add (goBuild);
 
-			var jump = new ContextMenuItem (GettextCatalog.GetString ("_Go to Task"));
+			var jump = new ContextMenuItem (GettextCatalog.GetString ("Go to Task"));
```
Therzok: Removing the underscores removes the mnemonic for non-macOS platforms. Please consider reverting this change.

netonjm: because now we have 3 Go items, make sense change to other character the underscore?

netonjm:
done!

PR #4014, opened 2018-03-02
Fix mutable completion list filtering

mhutch:
Fixes mutable completion lists again. Without this, PackageReference IntelliSense in the MSBuild extension will regress and be completely broken.


slluis:
Can you explain the use case? I'd like to add a unit test, to avoid breaking again.

mhutch:
@slluis see the commit message -

"The controller's ResetState method reset the StartOffset to zero. Since this method is called when a mutable list updates, the CompletionString used for filtering mutable lists was the entire document up to the completion point, which of course was so long and complex it matched nothing."

The test cases did not catch this as they have no document to update the CompletionString from.

mhutch:
You could easily test this with a document "a", trigger completion at the end of the document with a "b" character, then mutate the list to clear it and add a "b" item. Expect result is for the list to show "b", actual result is nothing because the filter string will be "ab" which will not batch the item "b".

Issue #4007, opened 2018-03-01
UI accessed from non-UI thread


mrward:
With Visual Studio 2017 for Mac (Preview) 7.5 Preview (7.5 build 654) (master build from Wrench), saw this entry in the IDE log:

```
AppKit.AppKitThreadAccessException: AppKit Consistency error: you are calling a method that can only be invoked from the UI thread.
  at AppKit.NSApplication.EnsureUIThread () [0x00020] in /Library/Frameworks/Xamarin.Mac.framework/Versions/4.5.0.18/src/Xamarin.Mac/AppKit/NSApplication.cs:107 
  at AppKit.NSAccessibilityElement.set_AccessibilityRole (System.String value) [0x00001] in /Library/Frameworks/Xamarin.Mac.framework/Versions/4.5.0.18/src/Xamarin.Mac/AppKit/NSAccessibilityElement.g.cs:3558 
  at MonoDevelop.Components.AtkCocoaHelper.RealAccessibilityElementProxy.SetRole (System.String role, System.String description) [0x00000] in /Users/builder/data/lanes/432/e3d445bc/source/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs:1133 
  at MonoDevelop.Components.AtkCocoaHelper.RealAccessibilityElementProxy.SetRole (MonoDevelop.Components.AtkCocoaHelper.AtkCocoa+Roles role, System.String description) [0x00000] in /Users/builder/data/lanes/432/e3d445bc/source/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs:1141 
  at MonoDevelop.Components.AtkCocoaHelper.AccessibilityElementProxy.SetRole (MonoDevelop.Components.AtkCocoaHelper.AtkCocoa+Roles role, System.String description) [0x00000] in /Users/builder/data/lanes/432/e3d445bc/source/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs:685 
  at MonoDevelop.Components.PathEntry.get_Accessible () [0x0001a] in /Users/builder/data/lanes/432/e3d445bc/source/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PathBar.cs:81 
[...]

PR #4017, opened 2018-03-02
Fixes issue #4007 UI accessed from non-UI thread

mkrueger:
https://github.com/mono/monodevelop/issues/4007


DavidKarlas:
Imo, it would be nicer fix to remove `.ConfigureAwait(false);` from line 747

slluis:
ConfigureAwait can be removed, but that won't fix the bug, because the await is done from a background thread, so it will continue on a background thread.

DavidKarlas:
I did check both calls to Update, from what I understand both come from UI thread, hence my comment

slluis:
We need a PR for d15.7 too.

DavidKarlas:
Whops, I just saw this is wrapped in Task.Run, nvm my comment

PR #4031, opened 2018-03-03
Fix japanse translation

asciian:
I changed the message that we get it when we install extensions.

This is a fix of translation:
- **インストール** means **install**
- **アンインストール** means **uninstall**

Thank you.


Therzok:
Hey, thanks for the PR! Unfortunately, we cannot merge this as the localization files for Japanese are handled by the localization team and any manual changes are discarded on re-import.

But I have opened a ticket upstream so they can take a look and fix this and it'll be brought in on the next import.

PR #4033, opened 2018-03-05
Fixed VSTS 569956: Moving lines changes whitespace

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/569956

The original bug can't be reproduced however the editorconfig settings were not taken in that use case so this patch fixes this. But it is unrelated to the original bug report.


main/src/addins/CSharpBinding/MonoDevelop.CSharp.Formatting/CSharpFormatter.cs
```diff
@@ -56,16 +56,29 @@ class CSharpFormatter : AbstractCodeFormatter
...
 			if (lineSegment == null)
 				return;
 
 			try {
-				var policy = policyParent.Get<CSharpFormattingPolicy> (MimeType);
-				var textpolicy = policyParent.Get<TextStylePolicy> (MimeType);
-				var tracker = new CSharpIndentEngine (policy.CreateOptions (textpolicy));
+				Microsoft.CodeAnalysis.Options.OptionSet options = null;
+
+				foreach (var doc in IdeApp.Workbench.Documents) {
```
Therzok: Is there no simpler way to get the AnalysisDocument from the editor?

mkrueger: Good question - maybe with GetContent - haven't tried.

Therzok: Was just wondering if we can do a direct lookup rather than a search. We should keep UI thread work at minimum

mkrueger: No - even with get content not really a chance. The whole MonoDevelop.Ide.Gui.Document system is problematic - would've made more sense to join that with the ViewContent - by design ViewContent had the responsibilities of Document but MonoDevelop changed that.

There was a discussion some years back about that - so atm there is no connection to the Document that way back.

PR #4035, opened 2018-03-05
Fixed VSTS 530391 - Placing caret in global:: qualifier highlights wh…

mkrueger:
…ole document

This fix switches over to the roslyn highlighting engine.


Therzok:
Only DocumentContext can be null in that scenario. The workspace is always set and the services are always created.

mkrueger:
No can't be null :)

PR #4046, opened 2018-03-05
Fixes VSTS 572337: Saving changes causes the ViewContent to be reloaded

mkrueger:
https://devdiv.visualstudio.com/DevDiv/Xamarin%20VS%20for%20Mac/_workitems/edit/572337

Moving the file registry to central point requires to set the skip change to that point as well otherwise all other view content implementations are broken.


alanmcgovern:
Just pinging about this one again - we do need it in 15.7 to avoid a regression there.

mkrueger:
Should've been merged.

alanmcgovern:
👍  Yeah, lluis pointed it out to me the other day separately to this. Thanks!

PR #4053, opened 2018-03-06
[61249][Mac] Trigger global search when initiated from the editor

sevoku:
Fixes bxc #61249

### DIFF
--- a/main/src/addins/MacPlatform/MainToolbar/MainToolbar.cs
+++ b/main/src/addins/MacPlatform/MainToolbar/MainToolbar.cs
@@ -325,6 +325,9 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 			}
 			set {
 				searchEntry.StringValue = value;
+				// BXC #61249: setting StringValue programmatically won't raise the Changed event,
+				// but we want to start a new search when the controller updates the search text
+				SearchEntryChanged?.Invoke (searchEntry, EventArgs.Empty);
 			}
 		}
 					


main/src/addins/MacPlatform/MainToolbar/MainToolbar.cs
```diff
@@ -325,6 +325,9 @@ public void RebuildToolbar (IEnumerable<ButtonBarGroup> groups)
 			}
 			set {
 				searchEntry.StringValue = value;
+				// BXC #61249: setting StringValue programmatically won't raise the Changed event,
+				// but we want to start a new search when the controller updates the search text
+				SearchEntryChanged?.Invoke (searchEntry, EventArgs.Empty);
```
Therzok: What about the code above, with the SearchCategory? Does the go to file command work?

Therzok: The reason I'm asking is because if it isn't enough to trigger:
https://github.com/mono/monodevelop/pull/4053/files#diff-04ae4a0efff46021123c79451e4e7913R73

We should also fix it there.

sevoku: Hhm, do you mean got to file with the name selected in the editor? We don't have that, or did I miss it? Actually it'd be a good feature...

Therzok: No, I mean these 3 handlers which autofill the search category.

https://github.com/mono/monodevelop/blob/6bb421103a69c8b28fd723322cead5f2eb1469c6/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/MainToolbarCommandHandlers.cs#L39

These suffer from the same problem, as they only set the search category.

Therzok: See also https://github.com/mono/monodevelop/pull/4053/files#diff-04ae4a0efff46021123c79451e4e7913R313

sevoku: So, I've double checked it now, the other commands are not affected, because they don't involve the editor selection. There we just set the category and set the caret to the end of the search box so the user can start typing (which will trigger the search then), but we don't want to trigger the search directly, because there is nothing to search for.
However this would be a great feature, if we'd have the same behaviour for those commands. i.e if the user has selected something in the editor and fires the "Got To File", we'd fill "file:<selected text>" in the searchbox and start a search for the selection.

PR #4055, opened 2018-03-06
[d15-7][Forms] Further improve Xamarin.Forms support in .NET Standard projects

mrward:
Various .NET Standard - Xamarin.Forms problems addressed:

Add ContentPage with xaml then change build action to None for .xaml file.
Add ContentPage with xaml. Then both .xaml and .xaml.cs file removed from the project but not deleted. Then edit .csproj and remove associated MSBuild items
Add ContentPage with xaml then rename .xaml file
Add ContentPage with xaml then rename .xaml.cs file
Moving/copying xaml files from existing PCL/shared projects to .NET Standard project

VSTS 505912


mrward:
Mmm, looks like the Xamarin.Forms addin tests are not shutting own properly. The tests are all run, results are output, but the test run does not seem to completely end. Seems to be happening VSTS and on Wrench too.

mrward:
Unable to reproduce any hang locally using `mdtool.exe run-md-tests Xamarin.Forms.Addin.Tests.dll`.

mrward:
Looks like latest master also has the same problem.

mrward:
OK the Xamarin.Forms tests are OK on Wrench now.

PR #4068, opened 2018-03-06
Fix potential hang in Ide tests

slluis:
Make sure all solutions and projects created during test execution are properly disposed.


slluis:
It didn't fix the problem :/

slluis:
It's working now.

PR #4070, opened 2018-03-07
Fixed 'Bug 576819: Extract Interface... OK button is disabled when

mkrueger:
dialog is initially shown (works in VS2017)'

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/576819

Found & fixed an issue with the cell view update as well.

### DIFF
--- a/main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.ExtractInterface/ExtractInterfaceDialog.cs
+++ b/main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.ExtractInterface/ExtractInterfaceDialog.cs
@@ -108,7 +108,12 @@ namespace MonoDevelop.Refactoring.ExtractInterface
 			listViewPublicMembers.DataSource = treeStore;
 			var checkBoxCellView = new CheckBoxCellView (symbolIncludedField);
 			checkBoxCellView.Editable = true;
-			checkBoxCellView.Toggled += delegate { UpdateOkButton ();};
+			checkBoxCellView.Toggled += delegate { 
+				GLib.Timeout.Add (20, delegate {
+					UpdateOkButton (); 
+					return false;
+				});
+			};
 			listViewPublicMembers.Columns.Add ("", checkBoxCellView);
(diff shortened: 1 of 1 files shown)


mkrueger:
See : 

https://github.com/mono/xwt/issues/795

Therzok:
cc @sevoku

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.ExtractInterface/ExtractInterfaceDialog.cs
```diff
@@ -108,7 +108,12 @@ public ExtractInterfaceDialog ()
 			listViewPublicMembers.DataSource = treeStore;
 			var checkBoxCellView = new CheckBoxCellView (symbolIncludedField);
 			checkBoxCellView.Editable = true;
-			checkBoxCellView.Toggled += delegate { UpdateOkButton ();};
+			checkBoxCellView.Toggled += delegate { 
+				GLib.Timeout.Add (20, delegate {
```
Therzok: Can we track this hack in a github issue so we don't forget?

mkrueger: It's here : https://github.com/mono/xwt/issues/795

Therzok: Created a issue tracking this on our side https://github.com/mono/monodevelop/issues/4075.

Easier than an issue in xwt with cross references, so we don't lose track of it.

mkrueger: Not a big thing to track if you ask me - we've some time outs for similiar behaviors. It's not that this code won't run in ~10 years :) - it's even likely that it will because no one cares about working code and refactoring :/

netonjm: ✋

PR #4078, opened 2018-03-07
[Mac] Check for mono 5.10.124, or 5.8.0.130 to signal a mono which ca…

Therzok:
…n't handle csc compiled xammac


Therzok:
Rebased again to get the hang fix in.

PR #4099, opened 2018-03-08
Update to versions of Editor and Roslyn shipped with VS 15.6.

KirillOsenkov:


KirillOsenkov:
I've done some spelunking and found the versions of Editor and Roslyn that shipped in VS 15.6 RTM.

**Editor:**
15.6.319-preview+g5bc180380a.RR
SHA: 5bc180380aa63ff90b2e3aeb6a9acfeac2bd16c5
NuGet:
https://vside.myget.org/feed/vssdk/package/nuget/Microsoft.VisualStudio.Text.Logic/15.6.27415

**Roslyn:**
2.7.0-beta3-62620-07
SHA: e873e69306527a0424e97af2985370556f474019
NuGet: https://dotnet.myget.org/feed/roslyn/package/nuget/Microsoft.CodeAnalysis.CSharp/2.7.0-beta3-62620-07

KirillOsenkov:
Turns out that in that particular NuGet package of the editor, the HintPaths are `lib\net45` instead of `lib\net46`. I've let the Editor team know about this, but for now, fixing build.

Also filed this issue:
https://github.com/mono/monodevelop/issues/4102

Because of this bug it was building fine on Windows as, though the HintPaths were wrong, it was resolving the editor binaries from the VS SDK folder:

image

slluis:
Do we have any pressing need to do this bump in 15.6? If not, I'd prefer to not bump it there.

KirillOsenkov:
It's been almost 2 months between what we have and their final version, so who knows how much we're missing.

KirillOsenkov:
I don't think we'll ever need this. We can always reopen if we need this.

PR #4108, opened 2018-03-09
Attempt to fix test that fails only on VSTS

nosami:


KirillOsenkov:
Looks good! Please get this in, because this creates quite a lot of noise. Thanks!

nosami:
thanks @KirillOsenkov!

PR #4137, opened 2018-03-13
[Ide] Fix a crash happening on a null callback post/send

Therzok:
Something is queueing a null callback on the UI thread. Stop that from happening, and log a stacktrace on the bg thread doing the queueing so we know what is causing it

Fixse VSTS #581086 - NRE exception in MD.Ide. DispatchService


sgmunn:
looks ok to me.

PR #4143, opened 2018-03-13
[Native Shell] Start splitting the docking components

iainx:
Splitting the Docking components into interfaces with the UI components hidden from the world. Doesn't have a Cocoa backend yet, but I'd like to get the interfaces in so I don't need to keep rebasing them.


iainx:
I'm thinking now is a good time to get it into master for the 15.8 timeframe, so that we have time to see if there are regressions

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockBarItem.cs
```diff
@@ -439,15 +534,15 @@ protected override bool OnEnterNotifyEvent (Gdk.EventCrossing evnt)
...
 			} else {
-				AutoHide (false);
+				parentItem.AutoHide (false);
 			}
 		}
 		
 		protected override bool OnLeaveNotifyEvent (Gdk.EventCrossing evnt)
 		{
-			ScheduleAutoHide (true);
+			parentItem.ScheduleAutoHide (true);
```
slluis: The code that calls AutoShow/AutoHide/ScheduleAutoHide should probably also be part of the control logic in DockBarItem

slluis:
Yes, I think it is a good time. It will take a bit of time to review, it is a huge patch.

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -1087,6 +1088,18 @@ internal override void RestartIde (bool reopenWorkspace)
 			proc.StartInfo = psi;
 			proc.Start ();
 		}
+
+		internal override T CreateNativeImplementation<T> (params object[] p)
+		{
+			var assembly = Assembly.GetExecutingAssembly ();
+			foreach (var type in assembly.GetTypes ()) {
```
Therzok: We should do something smarter here, if we intend to keep it. Doing this every time we want to create an instance is expensive.

We should have a mapping for each platform service and keep the constructor we want to use and invoke that directly.

Worst case, if we do end up using this as the main way to create native implementations, we could rely on something like [FastActivator](https://github.com/mono/gtk-sharp/blob/gtk-sharp-2-12-branch/glib/FastActivator.cs) I wrote in Gtk# to avoid slow reflection

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockItemToolbar.cs
```diff
@@ -165,127 +243,112 @@ void Add (Control control, bool fill, int padding, int index)
...
-				frame.Visible = value;
-			}
-		}
-		
-		public bool Sensitive {
-			get { return frame.Sensitive; }
-			set { frame.Sensitive = value; }
+
+		public Control[] ToolbarItems {
+			get { return box.Children.Select (child => (Control)child).ToArray (); }
```
Therzok: `box.Children.Cast<Control> ().ToArray()`.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.DockNotebook/TabStrip.cs
```diff
@@ -404,7 +405,11 @@ public PadTitleWindow (DockFrame frame, DockItem draggedItem) : base (Gtk.Window
 			{
 				SkipTaskbarHint = true;
 				Decorated = false;
-				TransientFor = (Gtk.Window)frame.Toplevel;
+				var controlWidget = frame.Control as Widget;
+				if (controlWidget == null) {
+					LoggingService.LogWarning ("Toolkit mismatch");
```
Therzok: `throw` here?

main/src/addins/MacPlatform/MacPlatform.cs
iainx: I agree, this was just something I was using to test with, and at the moment isn't used. I'll remove it and when make something better when necessary

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockItemToolbar.cs
iainx: 👍

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.DockNotebook/TabStrip.cs
iainx: Yup, from old code before I had `ToolkitMismatchException`

Fixed in all instances

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockBarItem.cs
iainx: Done. I'm sure there's other places like that as well that I missed

PR #4157, opened 2018-03-14
Report the version of Xamarin.Mac that's used at runtime by the IDE

Therzok:
Fixes VSTS #577285 Report version of Xamarin.Mac that's used at runtime

### DIFF
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -156,7 +156,50 @@ namespace MonoDevelop.MacIntegration
 
 			Xwt.Toolkit.CurrentEngine.RegisterBackend<IExtendedTitleBarWindowBackend,ExtendedTitleBarWindowBackend> ();
 			Xwt.Toolkit.CurrentEngine.RegisterBackend<IExtendedTitleBarDialogBackend,ExtendedTitleBarDialogBackend> ();
+
+			var description = XamMacBuildInfo.Value;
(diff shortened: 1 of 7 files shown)


main/external/Xamarin.Mac.version
```diff
@@ -0,0 +1 @@
+4.4.0.36
```
slluis: What is this for?

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -156,8 +156,42 @@ public MacPlatformService ()
...
+			return line.Split (':') [1].Trim ();
+		}
+
+		static Lazy<string> XamMacBuildInfo = new Lazy<string> (() => {
+			const string buildInfoResource = "Xamarin.Mac.buildinfo";
+			var asm = System.Reflection.Assembly.GetExecutingAssembly ();
+
+			string version, hash, branch;
+
+			using (var stream = asm.GetManifestResourceStream (buildInfoResource))
```
slluis: This method should be better protected against crashes, that may happen if the format of buildinfo changes for some reason.

main/external/Xamarin.Mac.version
Therzok: Remnant.

main/src/addins/MacPlatform/MacPlatform.cs
Therzok: 👍

Therzok:
Fixed

PR #4158, opened 2018-03-14
Report the version of Xamarin.Mac that's used at runtime by the IDE

Therzok:
Fixes VSTS #577285 Report version of Xamarin.Mac that's used at runtime


Therzok:
Updated with same as master.

PR #4164, opened 2018-03-14
Fixes VSTS 581661: [Feedback] /* */ Comments slow or non responsive

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/581661

Seems that a span change detection was missing completely - that this
worked in former versions must've been a side effect. I took the
chance to ensure that highlighting is really done in background which
looks ok on my systems.


DavidKarlas:
I think this `Task.Run` should be at https://github.com/mono/monodevelop/blob/ddb4355/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/RoslynClassificationHighlighting.cs#L137 but before calling `Task.Run` on `document.GetSemanticModelAsync ()` `document.TryGetSemanticModel` should be called, but maybe I'm wrong...

mkrueger:
I want to avoid to create too many tasks for that - there are multiple calls and that could be costly. This is more resource efficient.

And it is just one implementation of the highlighting engine - there are several others…

PR #4174, opened 2018-03-15
[Tests] Improve project capability tests

mrward:
Register the MonoDevelop.Core.Tests.Addin in the project capability
tests and unregister it afterwards. This is instead of registering
the addin using an .addins file generated in the local-config folder
at build time and then deleting it after the test is run.


slluis:
Please update to latest master.

mrward:
Rebased on origin/master

PR #4175, opened 2018-03-15
[Ide] Use monospace font for commands

rodrmoya:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
```diff
@@ -73,6 +73,14 @@ class BuildOutputNode : TreePosition
 		public virtual int ErrorCount { get; set; }
 		public virtual int WarningCount { get; set; }
 
+		public virtual bool IsCommandLine {
+			get {
+				return NodeType == BuildOutputNodeType.Message && Parent.NodeType == BuildOutputNodeType.Task &&
+					                                  (Parent.Message == "Csc" || Parent.Message == "AL" ||
```
Therzok: These should be some string constants.

netonjm: If this is not going to change in time, we can set the value, instead check in every draw

PR #4179, opened 2018-03-15
[d15-7][Mac] Compile monostub as 2 variants, one with gui, one without

Therzok:


Therzok:
Rebased on top of latest d15-7 because it was failling compat check.

Issue #4173, opened 2018-03-15
C# sources with `None` build action are not editable


atsushieno:
Sometimes we add C# sources with build action `None ` (e.g. to become some kind of resources). The still need to be editable, but MonoDevelop fails to make it so. Cursors don't move when we switch to editor for such files.

Repro:
- Create a console project.
- Set build action for `Program.cs` from `Compile` to `None`.
- Close the editor.
- Open the file again.
- Try to edit it.

Checked with monodevelop 036b03b.

> VS bug [#583940](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/583940)

mrward:
Reproduced on VS Mac 7.5 build 843 (d15-7).

VS Mac 7.4 build 1033 (d15-6) is OK.

The IDE log has:

```
ERROR [2018-03-15 09:22:44Z]: Error while updating the indentation engine
System.NullReferenceException: Object reference not set to an instance of an object
  at MonoDevelop.CSharp.Formatting.CSharpTextEditorIndentation.SafeUpdateIndentEngine (System.Int32 offset) [0x00000] in /Users/builder/data/lanes/5965/0f413f3d/source/monodevelop/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Formatting/CSharpTextEditorIndentation.cs:84 
ERROR [2018-03-15 09:22:44Z]: Error in text editor extension chain
System.NullReferenceException: Object reference not set to an instance of an object
  at MonoDevelop.CSharp.Formatting.CSharpTextEditorIndentation.KeyPress (MonoDevelop.Ide.Editor.Extension.KeyDescriptor descriptor) [0x002c0] in /Users/builder/data/lanes/5965/0f413f3d/source/monodevelop/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Formatting/CSharpTextEditorIndentation.cs:436 
  at MonoDevelop.Ide.Editor.Extension.TextEditorExtension.KeyPress (MonoDevelop.Ide.Editor.Extension.KeyDescriptor descriptor) [0x00000] in /Users/builder/data/lanes/5965/0f413f3d/source/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/TextEditorExtension.cs:83 
[...]

PR #4198, opened 2018-03-16
Fixes issue #4173 C# sources with `None` build action are not editable

mkrueger:
https://github.com/mono/monodevelop/issues/4173

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp/DocumentContextExtension.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp/DocumentContextExtension.cs
@@ -1,0 +1,52 @@
+//
+// DocumentContextExtension.cs
+//
+// Author:
+//       Mike Krüger <mikkrg@microsoft.com>
+//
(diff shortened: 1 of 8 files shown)


main/src/addins/CSharpBinding/MonoDevelop.CSharp/DocumentContextExtension.cs
```diff
@@ -0,0 +1,52 @@
...
+namespace MonoDevelop.CSharp
+{
+	static class DocumentContextExtension
+	{
+		public static async Task<OptionSet> GetOptionsAsync (this DocumentContext ctx, CancellationToken cancellationToken = default (CancellationToken))
+		{
+			if (ctx == null)
+				throw new ArgumentNullException (nameof (ctx));
+			if (ctx.AnalysisDocument != null) {
+				var result = await ctx.AnalysisDocument.GetOptionsAsync ();
```
DavidKarlas: I think we want `.ConfigureAwait(false);` here or calls with `WaitAndGetResult` can deadlock.

PR #4206, opened 2018-03-16
[SourceEditor] Implemented other apporach to fix the update issue.

mkrueger:
This way it's a bit less complex.


Therzok:
?w=1 looks better.

PR #4184, opened 2018-03-15
[15.6] Fixes VSTS 581661: [Feedback] /* */ Comments slow or non responsive

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/581661

Seems that a span change detection was missing completely - that this
worked in former versions must've been a side effect. I took the
chance to ensure that highlighting is really done in background which
looks ok on my systems.

PR #4207, opened 2018-03-16
[SourceEditor] Implemented other apporach to fix the update issue.

mkrueger:
This way it's a bit less complex.

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.cs
@@ -1358,14 +1358,30 @@ namespace Mono.TextEditor
 			var token = cacheSrc.Token;
 			var task = doc.SyntaxMode.GetHighlightedLineAsync (line, token);
 			if (task.IsCompleted) {
+				if (result != null && ShouldUpdateSpan (result, task.Result))
+					textEditor.QueueDraw ();
 				cachedLines [lineNumber] = task.Result;
 				return Tuple.Create (TrimChunks (task.Result.Segments, offset - line.Offset, length), true);
 			}
 			task.ContinueWith (t => {
 				cachedLines [lineNumber] = t.Result;
-				Document.CommitLineUpdate (lineNumber); // Required for highlighting updates
+				if (result != null && ShouldUpdateSpan (result, t.Result)) {
+					textEditor.QueueDraw ();
+				} else {
+					Document.CommitLineUpdate (line);
+				}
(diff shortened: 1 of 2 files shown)


Therzok:
?w=1 looks better. LGTM.

slluis:
Which issue is this fixing?

mkrueger:
Highlighting update - the old fix broke webtools highlighting implementation.

In any case it's just a work around - we need a roslyn callback.

mkrueger:
Basically reverts https://github.com/mono/monodevelop/commit/099391c3df9fc4e4fbe5c3aa6f5320336d6572bd

And implements another approach. But as said the parser should call the ui. Because of an update bug that used to work in older versions where we switched to the roslyn classifiers - there the editor often updated too much. With the drawing update optimization that has become visible. 

I need some feedback from the roslyn team how they handle that  the classifier API has no form of callbacks. The workaround is not the best solution for that - the parser knows about state changes.

mkrueger:
It's already in the 15.6 PR:

https://github.com/mono/monodevelop/pull/4184

However having some days to try it out would be nice

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.cs
```diff
@@ -1356,39 +1356,32 @@ public ChunkDescriptor (TextDocument doc, DocumentLine line, int offset, int len
...
-					if (!LinesAreEqual(highlightedCurLine, updatedCurLine)) {
-						cachedLines = cachedLines.SetItem (curLineNumber, updatedCurLine);
-						await Runtime.RunInMainThread (delegate {
-							Document.CommitMultipleLineUpdate (lineNumber, lineNumber + (int)(textEditor.Allocation.Height / textEditor.LineHeight) + 1);
-						});
-						return;
-					}
+			var task = doc.SyntaxMode.GetHighlightedLineAsync (line, token);
+			if (task.IsCompleted) {
+				if (result != null && ShouldUpdateSpan (result, task.Result))
```
Therzok:
```
       /// <summary>
        /// Gets whether this <see cref="Task">Task</see> has completed.
        /// </summary>
        /// <remarks>
        /// <see cref="IsCompleted"/> will return true when the Task is in one of the three
        /// final states: <see cref="System.Threading.Tasks.TaskStatus.RanToCompletion">RanToCompletion</see>,
        /// <see cref="System.Threading.Tasks.TaskStatus.Faulted">Faulted</see>, or
        /// <see cref="System.Threading.Tasks.TaskStatus.Canceled">Canceled</see>.
        /// </remarks>
```

This means we should check whether the task is faulted or cancelled before running this.

PR #4208, opened 2018-03-16
[Ide] Fix file template API break

mrward:
Fixed SingleFileDescription's public methods being removed.
Ensure that code that uses the obsolete non-async methods still
work by adding back the original code for these methods.


Therzok:
Closing as this is going in via d15-7 merge.

PR #4211, opened 2018-03-16
[BuildOutput] Default to native backend on Mac

sevoku:


netonjm:
Before merging this PR, I would wait to fix performance issues in Xamarin.Mac backend. This needs to be improved and tested.

Therzok:
Retargeted against master.

PR #4221, opened 2018-03-17
[Mac] Fix Add Files dialog being broken on High Sierra

Therzok:
This uses the same fix as a4bbb0989b9c9767766e6b16a83f0c07c406d0c2
to fix this dialog being broken on High Sierra. Don't layout
after setting the accessory view

Fixes VSTS #585364 Add File/[Options] UI is not shown


Therzok:
Also moved the file filter at the bottom to be consistent with the open files dilaog.

screen shot 2018-03-17 at 16 23 19
screen shot 2018-03-17 at 16 23 33

PR #4232, opened 2018-03-19
[CI] Add VSTS build branch name to IDE build info

Therzok:


Therzok:
Works!

screen shot 2018-03-19 at 20 44 12

PR #4250, opened 2018-03-20
[Core] Improve .NET Core project load times for project with project files

mrward:
Fixes VSTS #524077 - Performance: Opening .NET Core project with many
files that are not excluded

Two parts two to this change:

 1. Improve re-evaluation performance
 2. Improve initial project load time

Timinges for .NET Core console project that has a node_modules directory
containing 17150 files:

  - Initial project load time down from ~70-80 seconds down to ~20 seconds.
  - Re-evaluation time from ~110 seconds down to ~20 seconds.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
@@ -477,7 +477,7 @@ namespace MonoDevelop.Projects.MSBuild
 
 		object readLock = new object ();
 
-		internal MSBuildProjectInstanceInfo LoadNativeInstance ()
+		internal MSBuildProjectInstanceInfo LoadNativeInstance (bool evaluateItems)
 		{
 			lock (readLock) {
 				var supportsMSBuild = UseMSBuildEngine && GetGlobalPropertyGroup ().GetValue ("UseMSBuildEngine", true);
@@ -514,8 +514,10 @@ namespace MonoDevelop.Projects.MSBuild
 						};
 						var xml = SaveToString (ctx);
 
-						foreach (var it in GetAllItems ())
-							it.EvaluatedItemCount = 0;
+						if (evaluateItems) {
+							foreach (var it in GetAllItems ())
+								it.EvaluatedItemCount = 0;
+						}
 
 						nativeProjectInfo.Project = e.LoadProject (this, xml, FileName);
 					} catch (Exception ex) {
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -2914,17 +2914,41 @@ namespace MonoDevelop.Projects
 			if (loadedItems != null)
 				loadedItems.Clear ();
 
+			HashSet<ProjectItem> unusedItems = null;
+			Dictionary<string, ProjectItem> lookupItems = null;
+			List<ProjectItem> newItems = null;
+			if (IsReevaluating) {
+				unusedItems = new HashSet<ProjectItem> (Items);
+				lookupItems = new Dictionary<string, ProjectItem> ();
+				newItems = new List<ProjectItem> ();
+
+				// Improve ReadItem performance by creating a dictionary of items that can be
+				// searched faster than using Items.FirstOrDefault. Building this dictionary takes ~15ms
+				foreach (var it in Items) {
+					if (it.BackingItem != null && it.BackingEvalItem != null) {
+						lookupItems.Add (GetProjectItemLookupKey (it.BackingEvalItem), it);
+					}
+				}
+			}
+
 			var localItems = new List<ProjectItem> ();
 			foreach (var buildItem in msproject.EvaluatedItemsIgnoringCondition) {
 				if (buildItem.IsImported && !ProjectExtension.OnGetSupportsImportedItem (buildItem))
 					continue;
 				if (BuildAction.ReserverIdeActions.Contains (buildItem.Name))
 					continue;
-				ProjectItem it = ReadItem (buildItem);
-				if (it == null)
+				var result = ReadItem (buildItem, lookupItems);
+				if (result.Item == null)
 					continue;
-				it.Flags = flags;
-				localItems.Add (it);
+
+				result.Item.Flags = flags;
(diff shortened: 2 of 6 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
```diff
@@ -477,7 +477,7 @@ public MSBuildProjectInstance CreateInstance ()
 
 		object readLock = new object ();
 
-		internal MSBuildProjectInstanceInfo LoadNativeInstance ()
+		internal MSBuildProjectInstanceInfo LoadNativeInstance (bool evaluateItems)
```
Therzok: Nice

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -2914,29 +2914,58 @@ internal void LoadProjectItems (MSBuildProject msproject, ProjectItemFlags flags
 			if (loadedItems != null)
 				loadedItems.Clear ();
 
+			HashSet<ProjectItem> unusedItems = null;
+			Dictionary<string, ProjectItem> lookupItems = null;
+			List<ProjectItem> newItems = null;
```
Therzok: We're going to pass newItems and unusedItems to immutable APIs. The API is faster if we use ImmutableList here. (they just move a node, rather than constructing a tree). Maybe we can use an `ImmutableList.Builder` for newItems at least?

Therzok: (ImmutableHashSet does not trigger a hot path on immutable list addrange)

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -2914,29 +2914,58 @@ internal void LoadProjectItems (MSBuildProject msproject, ProjectItemFlags flags
...
 			if (IsReevaluating)
-				Items.SetItems (localItems);
+				Items.SetItems (localItems, newItems, unusedItems);
 			else
 				Items.AddRange (localItems);
 		}
 
+		static string GetProjectItemLookupKey (IMSBuildItemEvaluated item)
+		{
+			return $"{item.Name}-{item.Include}";
```
Therzok: Do we really want all these allocations? Can't we have a valuetuple key?

mrward: So just use `(string Name, string Include)` as the key? I assume that will have some magic GetHashCode and Equals implementation that uses both properties. Sounds fine to me 👍

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
mrward: I can try some different things and take some timings. 👍

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
Therzok: Yeah, ValueTuple does that.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
Therzok: Btw, passing in the ImmutableList.Builder works too

mrward: Looking at the code the loadedItems will be passed to ImmutableList.AddRange. The other collections, newItems and unusedItems, are not used with any ImmutableList api - they are just used to raise events. With 17000 items and loadedItems being a List<ProjectItem> the AddRange call is pretty fast ~1ms.

I could try changing List<ProjectItem> loadedItems to be an ImmutableList Builder to see if that makes a difference.

Therzok: I was thinking also in terms of allocations, as the loadedItems list would create `ProjectItem[]` allocations on resize, only to end up creating N ImmutableList.Node after AddRange is called.

Having an ImmutableList.Builder would remove those allocations and would make the AddRange call an assignment. Not a lot, but worthy to do, as this code has to deal with an unbounded amount of items

Therzok:
```
1) TestSimpleLogging(Debug,"LogDebug") (MonoDevelop.Core.LoggingServiceTests.TestSimpleLogging(Debug,"LogDebug"))
     Expected string length 21 but was 75. Strings differ at index 0.
  Expected: "This is a log message"
  But was:  "[1052 (+3)] Registering assemblies for framework .NETFramewor..."
  -----------^

  at MonoDevelop.Core.LoggingServiceTests.AssertLastMessageEqual (System.String message) [0x00000] in /Users/builder/data/lanes/6003/5f4fd871/source/monodevelop/main/tests/MonoDevelop.Core.Tests/MonoDevelop.Core/LoggingServiceTests.cs:75 
  at MonoDevelop.Core.LoggingServiceTests.AssertLastMessageEqual (System.String message, MonoDevelop.Core.Logging.LogLevel level) [0x00037] in /Users/builder/data/lanes/6003/5f4fd871/source/monodevelop/main/tests/MonoDevelop.Core.Tests/MonoDevelop.Core/LoggingServiceTests.cs:70 
  at MonoDevelop.Core.LoggingServiceTests.TestSimpleLogging (MonoDevelop.Core.Logging.LogLevel level, System.String methodName) [0x0009a] in /Users/builder/data/lanes/6003/5f4fd871/source/monodevelop/main/tests/MonoDevelop.Core.Tests/MonoDevelop.Core/LoggingServiceTests.cs:125 
  at (wrapper managed-to-native) System.Reflection.MonoMethod.InternalInvoke(System.Reflection.MonoMethod,object,object[],System.Exception&)
  at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00032] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-02/external/bockbuild/builds/mono-x64/mcs/class/corlib/System.Reflection/MonoMethod.cs:305 
```

This is the test failure on wrench. Seems unrelated, apparently logging service got something else written into it.

Issue #4257, opened 2018-03-21
IDE crash - null reference in CSharpTextEditorIndentation


mrward:
No idea how to reproduce this. I created a Forms project using a local master build and was opening a .cs file, maybe closing it.

``` 
FATAL ERROR [2018-03-21 10:43:47Z]: An unhandled exception has occured. Terminating Visual Studio? True
System.NullReferenceException: Object reference not set to an instance of an object
  at MonoDevelop.CSharp.Formatting.CSharpTextEditorIndentation+<>c__DisplayClass13_0.<HandleTextOptionsChanged>b__0 () [0x00044] in monodevelop/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Formatting/CSharpTextEditorIndentation.cs:161 
  at MonoDevelop.Core.Runtime+<>c.<RunInMainThread>b__39_0 (System.Object state) [0x00015] in monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs:337 
--- End of stack trace from previous location where exception was thrown ---
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () [0x0000c] in /Users/builder/jenkins/workspace/build-package-osx-mono/2017-12/external/bockbuild/builds/mono-x64/mcs/class/referencesource/mscorlib/system/runtime/exceptionservices/exceptionservicescommon.cs:152 
  at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task) [0x00037] in /Users/builder/jenkins/workspace/build-package-osx-mono/2017-12/external/bockbuild/builds/mono-x64/mcs/class/referencesource/mscorlib/system/runtime/compilerservices/TaskAwaiter.cs:187 
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task) [0x00028] in /Users/builder/jenkins/workspace/build-package-osx-mono/2017-12/external/bockbuild/builds/mono-x64/mcs/class/referencesource/mscorlib/system/runtime/compilerservices/TaskAwaiter.cs:156 
[...]

PR #4260, opened 2018-03-21
Fixes issue #4257 IDE crash - null reference in

mkrueger:
CSharpTextEditorIndentation

That prevents that this code runs on a disposed extension.


mrward:
Build error on VSTS:

MonoDevelop.CSharp.Formatting/CSharpTextEditorIndentation.cs(174,24): error CS0136: A local or parameter named 'e' cannot be declared in this scope because that name is used in an enclosing local scope to define a local or parameter

PR #4263, opened 2018-03-21
Fix VSTS 581051 - Value cannot be null. Parameter name: context

KirillOsenkov:
Fix VSTS 581051 - Value cannot be null. Parameter name: context

Fixes https://developercommunity.visualstudio.com/content/problem/210744/value-cannot-be-null-1.html

When opening a .json file (such as appsettings.json in an ASP.NET Core Web app) you get a fatal error message: Value cannot be null. Parameter name: context.

We think this is because of a race condition, where if someone first requests the JoinableTaskContext from a background thread, the importing constructor of JoinableTaskContextHost runs and returns, because we do not await the task spawned by the constructor. And so we grab the value of JoinableTaskContext property, which is still null, before it is being assigned by the UI thread task that we've started in the constructor.

The solution is to expose the MainThread from the runtime and forcibly initialize the JoinableTaskContext to the already known values of UI thread and SyncContext.

Stack:
Parameter name: context
  at Microsoft.VisualStudio.Utilities.JoinableTaskHelper..ctor (Microsoft.VisualStudio.Threading.JoinableTaskContext context) [0x00039] in C:\vs-editor-api\src\Text\Def\Internal\TextData\JoinableTaskHelper.cs:31 
  at Microsoft.VisualStudio.Text.Tagging.Implementation.TagAggregator`1[T]..ctor (Microsoft.VisualStudio.Text.Tagging.Implementation.TagAggregatorFactoryService factory, Microsoft.VisualStudio.Text.Editor.ITextView textView, Microsoft.VisualStudio.Text.Projection.IBufferGraph bufferGraph, Microsoft.VisualStudio.Text.Tagging.TagAggregatorOptions options) [0x00023] in C:\vs-editor-api\src\Text\Impl\TagAggregator\TagAggregator.cs:52

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
@@ -409,6 +409,11 @@ namespace MonoDevelop.Core
 		}
 
 		/// <summary>
+		/// The main UI thread of the application. Needed to initialize the JoinableTaskContext.
+		/// </summary>
+		public static Thread MainThread => mainThread;
+
+		/// <summary>
 		/// Returns true if current thread is GUI thread.
 		/// </summary>
 		public static bool IsMainThread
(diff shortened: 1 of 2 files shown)


Therzok:
@sgmunn found it!

KirillOsenkov:
Builds are green - @slluis Ashley mentioned they are probably going to do a SR for this change. Should we merge this into d15-6?

KirillOsenkov:
I'll defer to you to actually merge this - it says I'm not authorized :)

sgmunn:
@KirillOsenkov I can merge it once the qa-pending label is removed.

PR #4264, opened 2018-03-21
Fix VSTS 581051 - Value cannot be null. Parameter name: context

KirillOsenkov:
Fixes https://developercommunity.visualstudio.com/content/problem/210744/value-cannot-be-null-1.html

When opening a .json file (such as appsettings.json in an ASP.NET Core Web app) you get a fatal error message: Value cannot be null. Parameter name: context.

We think this is because of a race condition, where if someone first requests the JoinableTaskContext from a background thread, the importing constructor of JoinableTaskContextHost runs and returns, because we do not await the task spawned by the constructor. And so we grab the value of JoinableTaskContext property, which is still null, before it is being assigned by the UI thread task that we've started in the constructor.

The solution is to expose the MainThread from the runtime and forcibly initialize the JoinableTaskContext to the already known values of UI thread and SyncContext.

Stack:
Parameter name: context
  at Microsoft.VisualStudio.Utilities.JoinableTaskHelper..ctor (Microsoft.VisualStudio.Threading.JoinableTaskContext context) [0x00039] in C:\vs-editor-api\src\Text\Def\Internal\TextData\JoinableTaskHelper.cs:31
  at Microsoft.VisualStudio.Text.Tagging.Implementation.TagAggregator`1[T]..ctor (Microsoft.VisualStudio.Text.Tagging.Implementation.TagAggregatorFactoryService factory, Microsoft.VisualStudio.Text.Editor.ITextView textView, Microsoft.VisualStudio.Text.Projection.IBufferGraph bufferGraph, Microsoft.VisualStudio.Text.Tagging.TagAggregatorOptions options) [0x00023] in C:\vs-editor-api\src\Text\Impl\TagAggregator\TagAggregator.cs:52

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
@@ -417,6 +417,11 @@ namespace MonoDevelop.Core
 		}
 
 		/// <summary>
+		/// The main UI thread of the application. Needed to initialize the JoinableTaskContext.
+		/// </summary>
+		public static Thread MainThread => mainThread;
+
+		/// <summary>
 		/// Returns true if current thread is GUI thread.
 		/// </summary>
 		public static bool IsMainThread
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/JoinableTaskContextHost.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/JoinableTaskContextHost.cs
@@ -43,10 +43,7 @@ namespace MonoDevelop.Ide.Composition
 		[ImportingConstructor]
 		public JoinableTaskContextHost ()
 		{
-			Runtime.RunInMainThread (() => {
-				var joinableTaskContext = new JoinableTaskContext ();
-				this.JoinableTaskContext = joinableTaskContext;
-			});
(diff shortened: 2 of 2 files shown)


GLjackyvaswani:
Have verified the bug with the latest build. Issue is working fine.

Build Info:

Microsoft Visual Studio Enterprise 2017 for Mac (Preview)
Version 7.4.1 (build 49)

Xamarin.Mac
Version: 4.2.1.28 (Visual Studio Enterprise)
Xamarin.iOS
Version: 11.8.1.28 (Visual Studio Enterprise)
Xamarin.Android
Version: 8.2.0.16 (Visual Studio Enterprise)

Screen Cast: https://www.screencast.com/t/BfXrYEXZMF

Gist Link: https://gist.github.com/mohakbarokar/9676ab7dbdb8b6f6a52d24a62d469c0c

GLjackyvaswani:
I can confirm that this issue is not reproducible on below build.
Hence marking it as QA Approved.

Build Info:

Microsoft Visual Studio Enterprise 2017 for Mac (Preview)
Version 7.5 Preview (7.5 build 992)

Xamarin.Android
Version: 8.2.0.16 (Visual Studio Enterprise)
Xamarin.Mac
Version: 4.2.1.28 (Visual Studio Enterprise)
Xamarin.iOS
Version: 11.10.0.32 (Visual Studio Enterprise)


Screen Cast: https://screencast.com/t/ZPgxubGL

Gist Link: https://gist.github.com/neerajdeshmukh/1102c353eb4567004c2be5e76f7d9081

PR #4277, opened 2018-03-22
Backport d15-7 via rebase

Therzok:

PR #4282, opened 2018-03-22
Fix 541368: MSTest not supported in .NET Framework library project

DavidKarlas:
Problem was very obscure to debug because problem was 2 processes away... Basically when running locally VSfM ResultDirectory was set to `/Path/To/Current/Path/Of/Process/TestResults/` but when VSfM is started via `Finder` or `open /Application/Visual Studio.app/`, `Directory.GetCurrentDirectory()` returns empty, resulting in `/TestResults/` folder and test runner doesn't have permissions to access that folder, hence it failed. Now from VSfM always set ResultDirectory to project `obj/TestResults` folder.


Therzok:
@monojenkins backport master

Therzok:
@DavidKarlas please open PRs against master and backport to release branches using monojenkins to avoid a lot of work done like https://github.com/mono/monodevelop/pull/4277

PR #4285, opened 2018-03-22
[DotNetCore] Fix TargetFramework short name format changing on save

mrward:
Adding a file to a .NET Core project that targets Tizen 4.0 would
result in the TargetFramework changing from tizen40 to tizen4.0
Now the original framework identifier name is not modified and if
the version of the framework changes then the version will be
dotted or just numbers based on the format that was originally used
in the project.

Fixes VSTS 578016 - Saving Sdk style project should not change the
TargetFramework format.


mrward:
@monojenkins backport d15-7

Therzok:
Test failure unrelated.

Issue #3870, opened 2018-02-15
No correct way to host multiple `TextEditor` instance in a ViewContent


garuma:
Although it's doable to introduce an intermediary widget (e.g. `Gtk.Alignment`) as the `ViewContent.Control` and swap `TextEditor` instances as a child of that container, there is no way to inform the rest of the IDE that the internal content has changed and thus *everything* that depends on calling `ViewContent.GetContent` needs to be re-called again.

For instance:
 - Pathbar doesn't work
 - Document events are not re-registered
 - Extensions are not re-applied correctly
 - ...

There should be an event on `ViewContent` or an extra interface added to signify that a `ViewContent` can be the host of multiple editors.

> VS bug #569963

mhutch:
What's the use case for this?

garuma:
The Android designer represents multiple layout files in a unified single view

mkrueger:
Can you try to call StartReparseThread () ? Or re attach the project 
AttachToProject (null);AttachToProject (oldProject);

Extensions are per text editor so that should be no problem. The current design for Document is a 1:1 representation it's not changeable in the current form - but re attaching the project or reparsing should update everything.

garuma:
@mkrueger this doesn't seem to work because the `if` line in http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs,749 doesn't allow `null` to be passed in (the `adhocProject` field is null in that case causing the check there to pass and nothing to happen).

`StartReparseThread` is internal so we can't call it directly and it seems like it wouldn't cause our text editor extensions to be registered.

mkrueger:
give yourself an ivt - internal API can be used by us.

xamarin-release-manager:
Fixed in version 7.6.0.158 (master)

Author: Jérémie Laval
Commit: 4fa26d58109595d0e960207a02a20a964bf9c21c (mono/monodevelop)

garuma:
@mkrueger sorry I hadn't seen you updated your comment.

I just tried calling `document.UpdateParseDocument` and it doesn't work. The path bar is not visible in the second text editor instance (to be clear when I said the path bar didn't work I meant that the **widget** itself wasn't getting registered not that it was displaying wrong data).

[...]

PR #4288, opened 2018-03-22
[Ide] Add an internal method to allow reloading extensions attached to a document

garuma:
Context: https://github.com/mono/monodevelop/issues/3870

This commits adds an internal method that we can call to reload the extension chain and allow the pathbar widget to be updated when a view content changes the underlying text editor instances it uses (and thus alter the value that can be returned from GetContent).

Note that it's likely still profitable to add a ContentChanged event of some sort in the future to handle the use case described in the issue but that can be done on a longer term effort to support those. In the meantime that method is an acceptable workaround for the subset of features we care about.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
@@ -673,7 +673,19 @@ namespace MonoDevelop.Ide.Gui
 
 			if (window is SdiWorkspaceWindow)
 				((SdiWorkspaceWindow)window).AttachToPathedDocument (GetContent<MonoDevelop.Ide.Gui.Content.IPathedDocument> ());
+		}
 
+		/* In some situation, a document can be associated with an editor
+		 * outside of the normal document opening process. This happens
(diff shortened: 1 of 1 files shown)


garuma:
@slluis I agree but is that something that can be targeted for 15.8? Otherwise we will need the workaround

garuma:
@slluis should we commit this as a workaround for now or should I go back to using reflection?

mhutch:
@garuma FYI @DavidKarlas is working on something

mkrueger:
Even if it is a hack - that's why it is internal. We already have support for view content embedding that's the subviews are for.
However if for some use case that's not suitable there needs to be a way to control the update - which is the case here. 
The only issue here is that the view content changes the editor that's attached.

ATM Document-Editor is a 1:1 - so changing it to 1:n is the only thing that's required. Doing that would require to add a notify change method to Document. And that's basically what this patch does.

decriptor:
@DavidKarlas is there a branch with this work? Can we link to it from here for documentation purposes?

Therzok:
@decriptor https://github.com/mono/monodevelop/pull/4443

Therzok:
New document controller model should handle this already.

PR #4290, opened 2018-03-23
Fixes VSTS 581785: Not getting Intellisense for completing an iOS

mkrueger:
event on UIButton

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/581785

Needs to be removed if the completion engine is changed in that
context.


slluis:
@monojenkins backport d15-6

slluis:
@mkrueger can you create a PR for 15.6?

PR #4292, opened 2018-03-23
Fix 581070: [Feedback] setting breakpoint on non existing c# function…

DavidKarlas:
…s breaks debugging - "Debugger operation failed". Synchronous operation cancelled!

Problem was that IDE sent empty/null path for breakpoint to .NET Core debug adapter which froze in that case, hence I added `!string.IsNullOrEmpty(b.Filepath)` which fixes the problem, I also added support for FunctionBreakpoints, but later figured .NET Core adapter doesn't support that yet...


mrward:
What happens if you do send a function breakpoint request to the VS Code debugger when it does not support that? Just wondering if having that code is going to break anything :)

DavidKarlas:
I tested... looks like it's ignoring, nothing happens I put this is, so it just works in future when they support...

DavidKarlas:
@monojenkins backport d15-7

PR #4295, opened 2018-03-23
Fixes VSTS 564610: code completion is broken

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/564610

Work around for injecting protocol completion into the override
context.


slluis:
Can we have unit tests for this?

mkrueger:
Hard for that layer - it's a hack :/

slluis:
Then maybe it can be tested at a higher layer?

mkrueger:
fortunately there were some old c# completion text editor extension tests.

PR #4297, opened 2018-03-23
Fixes VSTS 586405: [Feedback] No intellisense for #if preprocessor di…

mkrueger:
…rective

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/586405


main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/CompletionProvider/PreProcessorCompletionProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+
+		public override async Task ProvideCompletionsAsync (CompletionContext context)
+		{
+			var document = context.Document;
+			var position = context.Position;
+			var cancellationToken = context.CancellationToken;
+
+			var model = await document.GetSemanticModelForSpanAsync (new TextSpan (position, 0), cancellationToken).ConfigureAwait (false);
+
+			var workspace = document.Project.Solution.Workspace;
```
Therzok: Actually, should we not take preprocessor values from all the configurations of this project into account?

PR #4302, opened 2018-03-23
Fix 581032: Debugger session not properly exited

DavidKarlas:
Main issue here was that when Attaching to process and user Quiting IDE, it didn't ask user for confirmation, we usually do this when debugging specific project when unloading that project, but we can't do that when attached...
Also modified  IdeApp.Exiting event execution to fix possible issue with different handlers setting Cancel=true, Cancel=false same as 6873f00665b1885a8095454de148075ee678fd68

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
@@ -106,7 +106,21 @@ namespace MonoDevelop.Debugger
 				// Refresh the evaluators list
 				evaluators = null;
 			});
+			IdeApp.Exiting += IdeApp_Exiting;
 		}
+
+		static void IdeApp_Exiting (object sender, ExitEventArgs args)
+		{
+			if (!IsDebugging)
+				return;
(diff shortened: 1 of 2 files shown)


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
```diff
@@ -106,8 +106,22 @@ static DebuggingService ()
...
 			});
+			IdeApp.Exiting += IdeApp_Exiting;
 		}
 
+		static void IdeApp_Exiting (object sender, ExitEventArgs args)
+		{
+			if (!IsDebugging)
+				return;
+			if (MessageService.Confirm (GettextCatalog.GetString (
+						"Debugging is currently running and will have to be stopped. Do you want to stop debugging?"),
```
mrward: Maybe we should re-word the first part of this question? Maybe something like "The debugger is currently running and will have to be stopped.  Do you want to stop debugging?"

Quitting the IDE when an application is being debugged the message is: "The project 'foo' is currently running and will have to be stopped. Do you want to continue closing it?"

The wording 'Debugging is currently running' seems odd to me.

DavidKarlas:
@monojenkins backport d15-7

PR #4303, opened 2018-03-23
[Core] Fix wildcards being expanded on saving a project

mrward:
Saving a project with a file wildcard that had a Link that used
MSBuild item metadata, as shown below, would result in the wildcard
being removed and replaced with MSBuild items for each file included
by the wildcard.

```
<Compile Include="**\*.cs" Exclude="obj\**">
  <Link>%(RecursiveDir)%(Filename).cs</Link>
</Compile>
```

The problem was that the evaluated Link property value was being
compared with the unevaluated value, which did not match, resulting
in the wildcard being replaced with individual MSBuild items for
each file. Now if the property value contains '%' a comparison is
made against the evaluated value when checking if the project item
has changed.

Fixes VSTS #569295 - Projects that includes files using wildcards
will expand (and remove) wildcards upon save

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -3728,13 +3728,25 @@ namespace MonoDevelop.Projects
 		{
 			// Compare only metadata, since item name and include can't change
 
+			MSBuildEvaluationContext context = null;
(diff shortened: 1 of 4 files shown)


mrward:
@monojenkins backport d15-7

slluis:
Let's merge for now, but I don't like much this fix. Evaluation of items should be done only during the project evaluation, not afterwards.

PR #4321, opened 2018-03-26
[BuildOutput] Implement navigation shortcuts

rodrmoya:


netonjm:
If we press Shift key with some files selected and we open the contextual right button, Copy must act over all selected files

PR #4325, opened 2018-03-26
[Core] Fix duplicate key restore error if duplicate files in project

mrward:
If a .NET Core project has a duplicate MSBuild item or an MSBuild
item that is already implicitly added then the NuGet restore would
fail with an error similar to:

An item with the same key has already been added. Key: (None, dupe.txt)

Now duplicate MSBuild items are handled when the project is
re-evaluated.

Fixes VSTS #591092 - Duplicate key error when loading project with
duplicate items


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -2983,12 +2996,22 @@ void InitFormatProperties ()
...
 		{
 			if (IsReevaluating) {
 				// If this item already exists in the current collection of items, reuse it
-				if (lookupItems.TryGetValue (GetProjectItemLookupKey (buildItem), out ProjectItem eit)) {
-					if (ItemsAreEqual (buildItem, eit.BackingEvalItem) || CheckProjectReferenceItemsAreEqual (buildItem, eit)) {
+				(ProjectItem Item, ImmutableList<ProjectItem>.Builder List) existingItem;
+				if (lookupItems.TryGetValue (GetProjectItemLookupKey (buildItem), out existingItem)) {
+					ProjectItem eit = null;
+					if (existingItem.List != null) {
+						eit = existingItem.List.FirstOrDefault (it => ItemsAreEqual (buildItem, it));
```
Therzok: From what I see, this method is called in loops. Avoid lambda captures here

mrward:
@monojenkins backport d15-7

PR #4353, opened 2018-03-28
[Core] Fix references not saved on installing NuGet packages

mrward:
The fix applied in commit c459d3c
now means that Project.ReevaluateProject can be called at the same
time as NuGet packages are being installed if those NuGet packages
contains MSBuild imports. When creating a new Xamarin.Forms project
references are added to the Android project at the same time that
the ReevaluateProject method is called, indirectly when the type
system calls GetSourceFilesAsync and causes the CoreCompileDependsOn
targets to be run. During the call to MSBuildProject.EvaluateAsync,
which the ReevaluateProject method awaits on, the UI thread can be
used to add a new reference to the project. If the project is not
saved the MSBuildProject is not updated so the LoadProjectItems
was removing the references added to the project since they do not
exist in the evaluated items. When the project is then saved by the
NuGet addin the reference is not in the .csproj file.

Now items added to the project whilst MSBuildProject.EvaluateAsync
is called in ReevaluteProject are captured and LoadProjectItems
takes these into account to ensure they are not lost.

Fixes VSTS #586393 Getting multiple errors on building Forms App

Also handle the case where uninstalling a NuGet package may result in a reference not being removed.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -2326,6 +2326,12 @@ namespace MonoDevelop.Projects
 					throw new InvalidOperationException (it.GetType ().Name + " already belongs to a project");
 				it.Project = this;
 			}
+
+			if (monitorItemsModifiedDuringReevaluation) {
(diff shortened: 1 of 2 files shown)


mrward:
@monojenkins backport d15-7

PR #4355, opened 2018-03-28
[BuildOutput] Humanize target names

rodrmoya:


Therzok:
The approach is wrong here:
a) The same string is humanized multiple times
b) Inconsistent humanize calls (once before adding to string pool, once after - the after one removes the benefits of the string cache completely).

The right approach is probably providing a wrapper on top of TextKeyedCache, as we are now transforming the string: https://github.com/dotnet/roslyn/blob/872202543f4be3033d39d0737426b9663a444c05/src/Compilers/Core/Portable/InternalUtilities/TextKeyedCache.cs#L16

PR #4361, opened 2018-03-29
Fixes VSTS 590492: [Feedback] Intellisense in Shared Projects seems b…

mkrueger:
…roken in latest Visual Studio for Mac

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/590492

The GetAst infrastructure is incompatible with roslyn and broke in the shared project context - so I removed the usages there.
It has only historically been used for nrefactory.

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Refactoring/RefactoryCommands.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Refactoring/RefactoryCommands.cs
@@ -56,18 +56,15 @@ namespace MonoDevelop.CSharp.Refactoring
 				del ();
 		}
 
-		protected override void Update (CommandArrayInfo ainfo)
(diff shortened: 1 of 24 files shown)


main/src/addins/CSharpBinding/MonoDevelop.CSharp.Refactoring/RefactoryCommands.cs
```diff
@@ -56,12 +56,12 @@ protected override void Run (object dataItem)
 				del ();
 		}
 
-		protected override void Update (CommandArrayInfo ainfo)
+		protected override async void Update (CommandArrayInfo ainfo)
```
slluis: Please override `UpdateAsync()`

Therzok: Good catch!

Issue #4356, opened 2018-03-28
Error in C# options when opening project


mhutch:
The following error happens in the terminal when i open a project:

```
ERROR [2018-03-28 12:26:05Z]: An unhandled exception has occured. Terminating MonoDevelop? False
System.NullReferenceException: Object reference not set to an instance of an object
  at MonoDevelop.CSharp.DocumentContextExtension+<GetOptionsAsync>d__0.MoveNext () [0x000fa] in /Users/mikayla/code/monodevelop/main/src/addins/CSharpBinding/MonoDevelop.CSharp/DocumentContextExtension.cs:47 
--- End of stack trace from previous location where exception was thrown ---
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () [0x0000c] in /Users/builder/jenkins/workspace/build-package-osx-mono/2017-12/external/bockbuild/builds/mono-x64/mcs/class/referencesource/mscorlib/system/runtime/exceptionservices/exceptionservicescommon.cs:152 
  at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task) [0x00037] in /Users/builder/jenkins/workspace/build-package-osx-mono/2017-12/external/bockbuild/builds/mono-x64/mcs/class/referencesource/mscorlib/system/runtime/compilerservices/TaskAwaiter.cs:187 
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task) [0x00028] in /Users/builder/jenkins/workspace/build-package-osx-mono/2017-12/external/bockbuild/builds/mono-x64/mcs/class/referencesource/mscorlib/system/runtime/compilerservices/TaskAwaiter.cs:156 
  at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd (System.Threading.Tasks.Task task) [0x00008] in /Users/builder/jenkins/workspace/build-package-osx-mono/2017-12/external/bockbuild/builds/mono-x64/mcs/class/referencesource/mscorlib/system/runtime/compilerservices/TaskAwaiter.cs:128 
[...]

PR #4363, opened 2018-03-29
Fixes issue #4356 Error in C# options when opening project

mkrueger:


mkrueger:
@monojenkins backport d15-7

PR #4372, opened 2018-03-29
Fixes VSTS 591384: Text marker grays out code after repeated

mkrueger:
comment/uncomment; gets cleared after clicking on an affected line

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/591384

Was caused by a change in text marker representation. After removing
them the marker Offsets become invalid.


mkrueger:
@monojenkins backport d15-7

Issue #4216, opened 2018-03-16
IWKNavigationDelegate has two interface methods, only show one code completion method


dalexsoto:
_From @LeeCenY on March 16, 2018 10:28_

### Steps to Reproduce

1.   mvWebView.NavigationDelegate = this;
2.   IWKNavigationDelegate
3.  decidePolicy
4.  decidePolicy

### Expected Behavior

[Export("webView:decidePolicyForNavigationResponse:decisionHandler:")]
[Export("webView:decidePolicyForNavigationAction:decisionHandler:")]

```
[Export("webView:decidePolicyForNavigationResponse:decisionHandler:")]
public void DecidePolicy(WKWebView webView, WKNavigationResponse navigationResponse, Action<WKNavigationResponsePolicy> decisionHandler)
{
   throw new System.NotImplementedException();
}
```
```
[Export("webView:decidePolicyForNavigationAction:decisionHandler:")]
public void DecidePolicy(WKWebView webView, WKNavigationAction navigationAction, Action<WKNavigationActionPolicy> decisionHandler)
{
    throw new System.NotImplementedException();
}
```
### Actual Behavior

[Export("webView:decidePolicyForNavigationResponse:decisionHandler:")]
[Export("webView:decidePolicyForNavigationResponse:decisionHandler:")]

```
[Export("webView:decidePolicyForNavigationResponse:decisionHandler:")]
public void DecidePolicy(WKWebView webView, WKNavigationResponse navigationResponse, Action<WKNavigationResponsePolicy> decisionHandler)
{
   throw new System.NotImplementedException();
}
```
```
[Export("webView:decidePolicyForNavigationResponse:decisionHandler:")]
public void DecidePolicy(WKWebView webView, WKNavigationResponse navigationResponse, Action<WKNavigationResponsePolicy> decisionHandler)
{
   throw new System.NotImplementedException();
}
```

### Environment

```
=== Visual Studio Community 2017 for Mac ===

Version 7.4 (build 1033)
Installation UUID: 891dc29f-4210-4ba9-a91b-9564b6fea41d
Runtime:
	Mono 5.8.0.129 (2017-10/835a98d8559) (64-bit)
	GTK+ 2.24.23 (Raleigh theme)

	Package version: 508000129

=== NuGet ===

版本: 4.3.1.4445

=== .NET Core ===

运行时: /usr/local/share/dotnet/dotnet
运行时版本:
	2.0.5
	2.0.0
[...]

PR #4387, opened 2018-03-29
Supply SortText to preserve protocol methods in the completion list

olegoid:
This PR is an attempt to fix the following issues:
* https://github.com/xamarin/xamarin-macios/issues/3762
* https://devdiv.visualstudio.com/DevDiv/_workitems/edit/585126
* https://github.com/mono/monodevelop/issues/4216

Essentially the problem boils down to iOS/Mac protocol method overloads not showing up in the completion list.

The fix description:

When we generate completion items in `ProtocolMemberCompletionProvider`, methods with the same name (but different signature) get similar completion info. Specifically, they get the same `SortText` and `Span`.

These two properties are then used by Roslyn to remove duplicates when merging multiple completion contexts. Refer to this code for more details:
https://github.com/dotnet/roslyn/blob/master/src/Features/Core/Portable/Completion/CompletionServiceWithProviders.cs#L360

If two completion items have the same `DisplayName`, `SortText`, and `Span` the completion service will consider removing one of those items from the list. This is precisely the root cause of https://github.com/xamarin/xamarin-macios/issues/3762

By changing the `SortText` property of protocol methods we will help the completion service to distinct similar items and remove only duplicates that have the same name, span, and signature.

How it worked before:
categorycompletion

How it works now:
categorycompletion

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/CompletionProvider/ProtocolMemberCompletionProvider.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/CompletionProvider/ProtocolMemberCompletionProvider.cs
@@ -157,7 +157,7 @@ namespace MonoDevelop.CSharp.Completion.Provider
 				pDict = pDict.Add ("DeclarationBegin", declarationBegin.ToString());
 
 				var tags = ImmutableArray<string>.Empty.Add ("NewMethod");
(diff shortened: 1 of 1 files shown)


Therzok:
@monojenkins backport d15-7

Therzok:
@monojenkins backport d15-6

Therzok:
@olegoid can you please backport to 15.6 and 15.7?

PR #4398, opened 2018-03-30
[Core] Log addin load traces to help diagnose eager loading of addins

Therzok:
https://gist.github.com/Therzok/542dcbfdedc2bac5fe2d4f6e8af7e746

Fixes VSTS #591639 Log addin load stacktraces to have a baseline of where an addin is loaded


Therzok:
Adding support in mono-addins to say whether the addin load came as a direct load or as a dependency load.

Therzok:
TODO:
* [ ] Make mono-addins report which extension point is causing the load
* [ ] Report that when loading

Issue #4383, opened 2018-03-29
IDE focus-in/out does a lot of IO


Therzok:
We should leverage the fact that we have a FileSystemWatcher to replace Workbench.SaveFileStatus/CheckFileStatus.

```
0.05%   <SaveFileStatus>b__0  •  5,838 ms  •  22 calls  •  MonoDevelop.Ide.Gui.Workbench+<>c__DisplayClass125_0.<SaveFileStatus>b__0
  0.05%   Start  •  5,838 ms  •  22 calls  •  System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start(TStateMachine&)
    0.05%   MoveNext  •  5,838 ms  •  22 calls  •  MonoDevelop.Ide.Gui.Workbench+<>c__DisplayClass125_0+<<SaveFileStatus>b__0>d.MoveNext
       0.04%   GetLastWriteTimeUtc  •  5,557 ms  •  288,838 calls  •  System.IO.File.GetLastWriteTimeUtc(String)
       0.00%   15 functions hidden  •  129 ms total  •  1733248 calls total
     0.00%   2 functions hidden  •  0 ms total  •  44 calls total
   0.00%   3 functions hidden  •  0 ms total  •  66 calls total
```

```
0.04%   <CheckFileStatus>b__126_0  •  5,134 ms  •  23 calls  •  MonoDevelop.Ide.Gui.Workbench.<CheckFileStatus>b__126_0
  0.04%   Start  •  5,134 ms  •  23 calls  •  System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start(TStateMachine&)
    0.04%   MoveNext  •  5,134 ms  •  23 calls  •  MonoDevelop.Ide.Gui.Workbench+<<CheckFileStatus>b__126_0>d.MoveNext
       0.04%   GetLastWriteTimeUtc  •  4,933 ms  •  288,838 calls  •  System.IO.File.GetLastWriteTimeUtc(String)
       0.00%   15 functions hidden  •  92 ms total  •  1444476 calls total
     0.00%   2 functions hidden  •  0 ms total  •  44 calls total
   0.00%   3 functions hidden  •  0 ms total  •  66 calls total
```

> VS bug [#593841](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/593841)

mhutch:
Ah. It was closed without any justification.

xamarin-release-manager:
[...]

PR #4400, opened 2018-03-30
[IDE] Use a FileSystemWatcher to take changes made while unfocused

iainx:
Track file changes made while unfocused using a FileSystemWatcher instead of
stating every file for the modification time when the application loses focus
and then again when it regains focus

Fixes GH #4383

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
@@ -1273,73 +1273,83 @@ namespace MonoDevelop.Ide.Gui
 			workbench.UnlockActiveWindowChangeEvent ();
 		}
 
-		List<FileData> fileStatus;
-		SemaphoreSlim fileStatusLock = new SemaphoreSlim (1, 1);
-		// http://msdn.microsoft.com/en-us/library/system.io.file.getlastwritetimeutc(v=vs.110).aspx
-		static DateTime NonExistentFile = new DateTime(1601, 1, 1);
+		List <FileSystemWatcher> watchers;
+		// The FileSystemWatcher will report all files changed in the WorkspaceItem's base directory
+		// but we only care about the files returned from GetKnownFiles
+		HashSet<string> filesWeCareAbout;
+
 		internal void SaveFileStatus ()
 		{
-//			DateTime t = DateTime.Now;
-			List<FilePath> files = new List<FilePath> (GetKnownFiles ());
-			fileStatus = new List<FileData> (files.Count);
-//			Console.WriteLine ("SaveFileStatus(0) " + (DateTime.Now - t).TotalMilliseconds + "ms " + files.Count);
-			
-			Task.Run (async delegate {
-//				t = DateTime.Now;
-				try {
-					await fileStatusLock.WaitAsync ().ConfigureAwait (false);
-					if (fileStatus == null)
-						return;
-					foreach (FilePath file in files) {
-						try {
-							DateTime ft = File.GetLastWriteTimeUtc (file);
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
```diff
@@ -1273,73 +1273,83 @@ internal void UnlockActiveWindowChangeEvent ()
...
-					}
-				} finally {
-					fileStatusLock.Release ();
-				}
-//				Console.WriteLine ("SaveFileStatus " + (DateTime.Now - t).TotalMilliseconds + "ms " + fileStatus.Count);
-			});
+			watchers = new List<FileSystemWatcher> ();
+
+			// Take each workspace item and watch its base directory for any file changes
+			foreach (WorkspaceItem item in IdeApp.Workspace.Items) {
```
Therzok: We shouldn't create and get rid of a file watcher every time we focus in/out.

We should probably iterate the solution and solution items and find the base common dir. If all paths are children of the common dir, just monitor the common dir.

Otherwise make multiple file watchers for all the paths: i.e. solution A has a project in `$(slndir)/../project`, we monitor `$slndir` and `$(slndir)/../project`.

In focus in/out should just toggle enable raising events.

Therzok: From slack:
```
iain holmes [7:24 PM]
I can move the watcher to the WorkspaceItem
```

Which is an even better idea!

Therzok:
So, it seems like the mono corefx FSW is only available in master atm.

I think the simplest solution we have for now is to split MonoDevelop.FSW into a different assembly and give IVT to Core and IDE so we don't expose this as API. (cc @mrward )

slluis:
This is not the right approach. We need to use MonoDevelop.FSW inside the project model or FileSystemService. The check for changes on focus-in/out was a workaround for the lack of efficient FSW, but it is not the right solution now that we have good FSW.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
iainx: In fact, if we rewrite `WorkspaceItem.FileStatusTracker` to use `FileSystemWatcher` like it sort of was planned (all the necessary code is commented out) then we could do away with the `FocusIn`/`FocusOut` tracking altogether, no?

iainx:
@slluis yes, that's what I was thinking. If I was to rewrite DesktopItem.FileStatusTracker to use the good FSW, that would remove the need for the focus-in/out workaround.

Would that work instead of putting it into the project model?

mkrueger:
The focus in/focus out approach was for a non working FSW  - when using for example VIM the FSW didn't throw a file change notification - was very strange. Some editors/file changes did some did not.

VS.NET does the file update in their project system AFAIK.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
```diff
@@ -1273,73 +1273,83 @@ internal void UnlockActiveWindowChangeEvent ()
...
+				var watcher = new FileSystemWatcher (item.BaseDirectory);
```
mkrueger: Projects can contain files that are outside of their base directory.

mkrueger:
That patch doesn't work for files that are not in a project. The old approach was just a work around but was correct in that use case.

Therzok:
@mrward is this fixed in file-watcher-integration PR?

mrward:
@Therzok - Yes this is fixed on the file-watcher-integration branch. I guess we can close this.

PR #4410, opened 2018-04-02
[VSTS] Bring in support for a local resolve directory for gtk

Therzok:
Instead of installing support libraries, we just download a zip
with the files and unpack it in the repo.

This way we can resolve assemblies from a local directory


Therzok:
Interesting, windows build passed.

Issue #4242, opened 2018-03-20
F# function keyword is highlighted incorrectly after pipe symbol


nosami:
<img width="186" alt="screen shot 2018-03-20 at 11 57 49 am" src="https://user-images.githubusercontent.com/667194/37653222-026be5f6-2c36-11e8-8c69-56da79f844e2.png">

> VS bug [#586435](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/586435)

xamarin-release-manager:
> [sync] [VS-17] Comment by **xamarinc**



Fixed in version 8.6.0.802 (master)



Pull Request #4421 merged by: Lluis Sanchez  
 Author: mono  
 Commit: 9ee8bd62bd89374cf8ce73253ffd08a802663b34 (xamarin/monodevelop)

PR #4421, opened 2018-04-03
Fix function highlighting. Fixes VSTS #586435 (#4242)

nosami:


nosami:
@monojenkins backport d15-7

Issue #4376, opened 2018-03-29
OnIMProcessedKeyPressEvent triggering highlighting


Therzok:
```
0.31%   OnIMProcessedKeyPressEvent  •  1,984 ms  •  39 calls  •  MonoDevelop.SourceEditor.ExtensibleTextEditor.OnIMProcessedKeyPressEvent(Key, UInt32, ModifierType)
  0.31%   OnIMProcessedKeyPressEventInternal  •  1,984 ms  •  39 calls  •  MonoDevelop.SourceEditor.ExtensibleTextEditor.OnIMProcessedKeyPressEventInternal(Key, UInt32, ModifierType)
    0.30%   ExtensionKeyPress  •  1,981 ms  •  39 calls  •  MonoDevelop.SourceEditor.ExtensibleTextEditor.ExtensionKeyPress(Key, UInt32, ModifierType)
      0.30%   KeyPress  •  1,981 ms  •  39 calls  •  MonoDevelop.DocFood.DocFoodTextEditorExtension.KeyPress(KeyDescriptor)
        0.30%   KeyPress  •  1,981 ms  •  897 calls  •  MonoDevelop.Ide.Editor.Extension.TextEditorExtension.KeyPress(KeyDescriptor)
          0.18%   KeyPress  •  1,147 ms  •  39 calls  •  MonoDevelop.CSharp.Completion.CSharpCompletionTextEditorExtension.KeyPress(KeyDescriptor)
            0.18%   KeyPress  •  1,147 ms  •  39 calls  •  MonoDevelop.Ide.Editor.Extension.CompletionTextEditorExtension.KeyPress(KeyDescriptor)
              0.16%   get_CurrentCodeCompletionContext  •  1,052 ms  •  68 calls  •  MonoDevelop.SourceEditor.SourceEditorView.get_CurrentCodeCompletionContext
                0.16%   CreateCodeCompletionContext  •  1,052 ms  •  68 calls  •  MonoDevelop.SourceEditor.SourceEditorView.CreateCodeCompletionContext(Int32)
                  0.16%   LocationToPoint  •  1,051 ms  •  68 calls  •  Mono.TextEditor.MonoTextEditor.LocationToPoint(DocumentLocation)
                    0.16%   LocationToPoint  •  1,051 ms  •  68 calls  •  Mono.TextEditor.TextViewMargin.LocationToPoint(DocumentLocation)
[...]

PR #4423, opened 2018-04-03
Fixes issue #4376 OnIMProcessedKeyPressEvent triggering highlighting.

mkrueger:
The costly coordinate calculation operation can be postponed until
showing the window.

### DIFF
--- a/main/src/addins/CSharpBinding/AspNet/RazorCSharpCompletionBuilder.cs
+++ b/main/src/addins/CSharpBinding/AspNet/RazorCSharpCompletionBuilder.cs
@@ -173,15 +173,12 @@ namespace MonoDevelop.CSharp.Completion
 		{
 			var savedCtx = realDocumentContext.GetContent<ICompletionWidget> ().CreateCodeCompletionContext (
 				realEditor.CaretOffset + triggerOffset - docInfo.CaretPosition);
-			var result = new CodeCompletionContext ();
+			
+			var result = new MonoDevelop.Ide.Editor.Projection.ProjectedCompletionExtension.ProjectedContext (savedCtx);
 			result.TriggerOffset = triggerOffset;
 			var loc = docInfo.UnderlyingDocument.Editor.OffsetToLocation (triggerOffset);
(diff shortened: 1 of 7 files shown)


mkrueger:
Ignore them ? In which way they're ignored ?

slluis:
Those properties are not being assigned a value anymore, so code that uses those properties will stop working.

Therzok:
Also, this bit of code is still causing double creation of the completion context:
https://github.com/mono/monodevelop/blob/c53726f31a769d65c4c7956e7ccd81d6c1f7c95f/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/CompletionTextEditorExtension.cs#L339

It can happen once unconditionally in one of the `if`s above, and once here.

mkrueger:
This properties are assigned by some code and get used in that cases. The context is only used for the communication between the text editor extension and the completion ui and not by a 3rd party consumer that pops up a 3rd party UI on that.

Addins can still use these properties and they are working. What you describe is that another addin is using the C# completion text editor extension to use the completion popup coordinates for something.

mkrueger:
@Therzok: Double creation is not the issue - this just sets now the caret position.
The real issue is that the costly coordinate calculation is running when it shouldn't be. And that's what I fixed.

Therzok:
Oh, yes, definitely, I know that the perf problem comes from x/y calculation.

I was thinking maybe we could reuse the context, but it makes sense to re-create it as other extensions can modify the caret offset

mkrueger:
Yes caret offsets can change but coordinate calculation needs only to be done once - there is just one time the coordinates are required. And completion context change is used while it's open etc. so calculating always these coordinates is wrong.

The completion context is pretty old code it was even in monodevelop before I started working on it :). AFAIR it was not inherited from SharpDevelop btw.

Therzok:
@slluis can you please re-review?

mkrueger:
Fixed & rebased to current master.

Therzok:
@slluis awaiting your approval

PR #4442, opened 2018-04-04
D15 7 disable static registrar

alanmcgovern:


Therzok:
Approved from me. Still investigating why it's breaking.

PR #4443, opened 2018-04-04
Split content view

DavidKarlas:
Change on designer side is here: https://github.com/xamarin/designer/pull/767
This two PRs change the way Xamarin.Forms Previewer operates with TextEditor, before this PRs Previewer AddIn created PreviewerContentView which had EditorWidget embedded inside HPaned widget and optionally on side when Preview button was pressed Preview appeared. There were multiple bugs with this approach because editor was just Widget not actual EditorContentView which meant Previewer had to take care of Loading/Saving file content setting IsDirty state, beside this problems and bugs, Cmd+-/Cmd++ didn't work for TextEditor, Pathbar for code part of .xaml not appearing...
This also fixes most of https://devdiv.visualstudio.com/DevDiv/_workitems/edit/558175 (need to sync Pathbar and Toolbar heights).
Old view:
image

New in Preview:
image

New in Source:
image

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PathBar.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PathBar.cs
@@ -256,6 +256,11 @@ namespace MonoDevelop.Components
 			}
 		}
 
+		public void HandlePathChange (object sender, MonoDevelop.Ide.Gui.Content.DocumentPathChangedEventArgs args)
+		{ 
+			SetPath (((Ide.Gui.Content.IPathedDocument)sender).CurrentPath);
+		}
+
 		public void SetPath (PathEntry[] path)
 		{
 			if (ArrSame (this.leftPath, path))
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Tabstrip.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Tabstrip.cs
@@ -352,7 +352,7 @@ namespace MonoDevelop.Components
 			get { return Label == "|"; }
 		}
 		
-		public object Tag {
+		public BaseViewContent[] ViewsSelection {
 			get;
 			set;
 		}
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/BaseViewContent.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/BaseViewContent.cs
@@ -159,24 +159,6 @@ namespace MonoDevelop.Ide.Gui
 		/// </summary>
 		/// <value>The display binding used to create this view.</value>
 		public IDisplayBinding Binding { get; internal set; }
-
-		public virtual IEnumerable<TabPageInfo> GetTabPageInfos() {
-			return new [] { new TabPageInfo (TabPageLabel, TabAccessibilityDescription, new [] { this }) };
-		}
-	}
-
-	public class TabPageInfo
-	{
-		public TabPageInfo (string label, string accessibilityDescription, BaseViewContent [] viewsSelection)
-		{
-			Label = label;
-			AccessibilityDescription = accessibilityDescription;
-			ViewsSelection = viewsSelection;
-		}
-
-		public string Label { get; set; }
-		public string AccessibilityDescription { get; set; }
-		public BaseViewContent [] ViewsSelection { get; set; }
 	}
 
 	public enum ProjectReloadCapability
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/BaseViewContent.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/BaseViewContent.cs
@@ -159,6 +159,24 @@ namespace MonoDevelop.Ide.Gui
 		/// </summary>
 		/// <value>The display binding used to create this view.</value>
 		public IDisplayBinding Binding { get; internal set; }
+
+		public virtual IEnumerable<TabPageInfo> GetTabPageInfos() {
+			return new [] { new TabPageInfo (TabPageLabel, TabAccessibilityDescription, new [] { this }) };
+		}
+	}
+
+	public class TabPageInfo
+	{
+		public TabPageInfo (string label, string accessibilityDescription, BaseViewContent [] viewsSelection)
+		{
+			Label = label;
+			AccessibilityDescription = accessibilityDescription;
+			ViewsSelection = viewsSelection;
+		}
+
+		public string Label { get; set; }
+		public string AccessibilityDescription { get; set; }
+		public BaseViewContent [] ViewsSelection { get; set; }
 	}
 
 	public enum ProjectReloadCapability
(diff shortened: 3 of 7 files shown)


DavidKarlas:
This also fixes https://devdiv.visualstudio.com/DevDiv/_queries/edit/550287

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/BaseViewContent.cs
```diff
@@ -159,6 +159,24 @@ protected virtual void OnSetProject (Project project)
 		/// </summary>
 		/// <value>The display binding used to create this view.</value>
 		public IDisplayBinding Binding { get; internal set; }
+
+		public virtual IEnumerable<TabPageInfo> GetTabPageInfos() {
```
mhutch: Can we get some doc comments on this API? :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PathBar.cs
```diff
@@ -256,6 +256,11 @@ void DisposeProxies()
 			}
 		}
 
+		public void HandlePathChange (object sender, MonoDevelop.Ide.Gui.Content.DocumentPathChangedEventArgs args)
```
Therzok: Does this need to be public?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/BaseViewContent.cs
```diff
@@ -159,6 +159,24 @@ protected virtual void OnSetProject (Project project)
...
+	public class TabPageInfo
+	{
+		public TabPageInfo (string label, string accessibilityDescription, BaseViewContent [] viewsSelection)
+		{
+			Label = label;
+			AccessibilityDescription = accessibilityDescription;
+			ViewsSelection = viewsSelection;
+		}
+
+		public string Label { get; set; }
```
Therzok: Public setters?

garuma:
Couple of comments from an implementer point of view:

- The `GetTabPageInfos` API is not super straightforward to understand, I'd rather have an interface implemented by my view content (or some other overridable property) to say that the ViewContent is happy to participate in splitting and tell which spot it prefers to be (left or right).
- Currently it seems like for initial displaying only the XF previewer use case is hardcoded with the source buffer being shown first. For the Android Designer I'd like to continue starting up with the design surface and not the text buffer or start directly in split mode.
- In general with attachable view content, a bunch of nice things like getting `Load` called and properties setup for us now disappear. Ideally I would like to retain the same behavior than before for my "main" ViewContent.
- Because of the way document loading proceed, the open document is only registered after everything has initialized. This mean I cannot access the text buffer indirectly via `TextFileProvider.Instance.GetTextEditorData` until `WrapDocument` is called at the end of the process. This breaks one of my use case.

Beside those issue and my buffer problem in https://github.com/mono/monodevelop/issues/3870 it looks like it works.

slluis:
Before we go further with this I'd like to spend some time thinking about the best experience for split and alternative views. We can schedule some work for 15.8 once that is more clear. Meanwhile, if the designer needs a workaround, let's merge it for 15.7.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/BaseViewContent.cs
DavidKarlas: I was planning to get back to write this, but later forgot and wanted to get code out for review asap, Done

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/PathBar.cs
DavidKarlas: Yes, because its accessed from another class, but notice, class is not public, so this is not public API outside IDE...

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/BaseViewContent.cs
DavidKarlas: Removed in new PR

garuma:
@DavidKarlas I still feel like it's not quite natural with the setup but as long as it works I don't mind.

Your patch compared to mine also highlights other issues due to the fact that we lost the spot of the main ContentView (beside losing `Load` and property settings like `Project`): we lost for instance the callback to `Save` (because we need to save other things that just the single buffer), the indication that we are not a `IsFile` tab  (we are multiple files), .... basically all things that we naturally override/use from `ViewContent`/`BaseViewContent` but are now either not set or ignored by the IDE.

In general as I mentioned before those seems to be more attachable ViewContent problems in general that they are not quite the same than main ViewContent. All of that makes me wonder if in the case of the Android Designer it wouldn't be better if the roles were still flipped like today (i.e. we are the main view content and the source editor is the attachable one). That would need more thinking though (what @slluis suggested), the current API and approach are completely fine with the XF Previewer though so they can definitely use it like it is now.

mhutch:
@garuma I think you could have the designer be the primary `ViewContent` and then have an `IAttachableDisplayBinding` that creates an `TextEditorViewContent` instance

You might need to subclass it and maybe make a couple private members protected so you can wire things up correctly to the designer, but I think it'd be pretty straightforward.

garuma:
Yep, that's what we do today but it seemed like having the text editor not be the primary content view was causing issues that @DavidKarlas was trying to solve with this work in the first place. If I'm not correct then I'm happy to continue doing it like that.

mhutch:
@garuma i think the issues were that the ViewContent was embedded and not wired up correctly for things like GetContent, WorkbenchWindow, focus events, toolbar, etc. In this new model it would be properly initialized and parented by the Document like any other ViewContent.

garuma:
@slluis @DavidKarlas what is the current status of this PR? Can we still get something done in time for 15.8?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Tabstrip.cs
```diff
@@ -352,7 +352,7 @@ class Tab : IDisposable
 			get { return Label == "|"; }
 		}
 		
-		public object Tag {
```
Therzok: Can we preserve this? Would make it a NOP API change.

alanmcgovern:
I take it we can't get this until d16 now?

alanmcgovern:
What's the status of this?

alanmcgovern:
Any update on this?

decriptor:
It has been another month. I suppose it is time to bump the issue again!

sgmunn:
@slluis after talking with David I believe that this addresses a dev comm issue that has been raised - 
https://developercommunity.visualstudio.com/content/problem/201007/replace-in-files-does-not-update-xaml-files-that-a.html

What is the plan to merge this, if any?

alanmcgovern:
I spoke to Lluis  directly last week. This Should be on the current sprint, if not it should be on the next sprint :)

alanmcgovern:
Actually to put some more info on this: There are some aspects of the approach that need to be changed and we're just waiting on Lluis to provide the details of that so we can action it and land the patch :)

olegoid:
+1 for landing this patch as it fixes some bugs assigned to me.

DavidKarlas:
Replaced by https://github.com/mono/monodevelop/pull/7158

Issue #4459, opened 2018-04-05
F# tests shouldn't inherit the Nuget.config from MonoDevelop


KirillOsenkov:
The MonoDevelop NuGet.config has some unreliable package sources (such as myget).

Ideally we should isolate the tests from unnecessary sources. We should investigate how to override or remove the package sources that come from MonoDevelop's root NuGet.config and just leave the relevant ones.

> VS bug [#595393](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/595393)

KirillOsenkov:
Potentially relevant test code: https://github.com/mono/monodevelop/blob/master/main/external/fsharpbinding/MonoDevelop.FSharp.Tests/TemplateTests.fs#L153-L157

nosami:
Fixed by https://github.com/mono/monodevelop/pull/4460

xamarin-release-manager:
> [sync] [VS-10] Comment by **xamarinc**



Fixed in version 8.6.0.802 (master)



Pull Request #4460 merged by: Jason Imison  
 Author: mono  
 Commit: 51600c0f7a72f9cc66706a6bbe9d51130da42115 (xamarin/monodevelop)

PR #4460, opened 2018-04-05
Faster test package restore. Fixes VSTS #595393 (#4459)

nosami:


Therzok:
Maybe we should also make this configuration for the rest of the repo and only define sources if they're needed. This might speed up builds by quite a bit.

PR #4465, opened 2018-04-06
[Mac] Ensure we support Copy(NSZone) for our NSTextFieldCell subclasses

alanmcgovern:
This ensures that if AppKit clones the native cell we will be able to
create a new C# wrapper for it.

One known cause of this is when calculating the BaselineOffsetFromBottom.
This issue also affected the designers.

This is being added as a preventative as we may not actually be
triggering an API which clones the native object, but it's better to
be safe than sorry!

### DIFF
--- a/main/src/addins/MacPlatform/MainToolbar/VerticallyCenteredTextFieldCell.cs
+++ b/main/src/addins/MacPlatform/MainToolbar/VerticallyCenteredTextFieldCell.cs
@@ -26,6 +26,7 @@
 using System;
 using AppKit;
 using CoreGraphics;
+using Foundation;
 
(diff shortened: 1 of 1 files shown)


Therzok:
We also need to apply a similar patch here:
https://github.com/mono/xwt/blob/289d4470a1017e7d1bd652248d17de8dc24cdf21/Xwt.XamMac/Xwt.Mac/TextEntryBackend.cs#L342
https://github.com/mono/xwt/blob/6948e605e2e620ffb505b3d25b3d8f2dbf2e721e/Xwt.XamMac/Xwt.Mac/LabelBackend.cs#L270

Therzok:
@slluis can we merge this?

PR #4466, opened 2018-04-06
Make unit tests relocatable

slluis:
This is necessary so that tests can be run against a VS for Mac bundle, outside the source tree. 

15.7 port of https://github.com/mono/monodevelop/pull/4270


slluis:
To be merged with https://github.com/xamarin/md-addins/pull/3060

Issue #4452, opened 2018-04-04
[C#] Go to Definition NRE when trying to navigate (possibly while solution is reloading)


Therzok:
```
ERROR [2018-04-04 21:47:22Z]: Error while executing command: Refactory Operations
System.AggregateException: One or more errors occurred. ---> System.NullReferenceException: Object reference not set to an instance of an object
  at ICSharpCode.NRefactory6.CSharp.Features.GotoDefinition.GoToDefinitionService+<FindSymbolAsync>d__6.MoveNext () [0x0001a] in /Users/therzok/Work/md/monodevelop/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Features/GotoDefinition/GotoDefinitionService.cs:62
   --- End of inner exception stack trace ---
  at System.Threading.Tasks.Task.ThrowIfExceptional (System.Boolean includeTaskCanceledExceptions) [0x00013] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-02/external/bockbuild/builds/mono-x64/mcs/class/referencesource/mscorlib/system/threading/Tasks/Task.cs:2166
  at System.Threading.Tasks.Task`1[TResult].GetResultCore (System.Boolean waitCompletionNotification) [0x0002b] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-02/external/bockbuild/builds/mono-x64/mcs/class/referencesource/mscorlib/system/threading/Tasks/Future.cs:562
  at System.Threading.Tasks.Task`1[TResult].get_Result () [0x00000] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-02/external/bockbuild/builds/mono-x64/mcs/class/referencesource/mscorlib/system/threading/Tasks/Future.cs:532
  at ICSharpCode.NRefactory6.CSharp.Features.GotoDefinition.GoToDefinitionService.TryGoToDefinitionInternal (Microsoft.CodeAnalysis.Document document, System.Int32 position, System.Threading.CancellationToken cancellationToken) [0x00001] in /Users/therzok/Work/md/monodevelop/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Features/GotoDefinition/GotoDefinitionService.cs:97
[...]

PR #4473, opened 2018-04-09
Fixes issue #4452 [C#] Go to Definition NRE when trying to navigate

mkrueger:
(possibly while solution is reloading)

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Features/GotoDefinition/GotoDefinitionService.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Features/GotoDefinition/GotoDefinitionService.cs
@@ -59,6 +59,8 @@ namespace ICSharpCode.NRefactory6.CSharp.Features.GotoDefinition
 
 		public static async Task<ISymbol> FindSymbolAsync (Document document, int position, CancellationToken cancellationToken)
 		{
+			if (document == null)
+				return null;
 			var workspace = document.Project.Solution.Workspace;
 
 			var semanticModel = await document.GetSemanticModelAsync (cancellationToken).ConfigureAwait (false);


mkrueger:
@monojenkins backport d15-7

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Features/GotoDefinition/GotoDefinitionService.cs
```diff
@@ -59,6 +59,8 @@ static ISymbol FindRelatedExplicitlyDeclaredSymbol (ISymbol symbol, Compilation
 
 		public static async Task<ISymbol> FindSymbolAsync (Document document, int position, CancellationToken cancellationToken)
 		{
+			if (document == null)
+				return null;
```
knocte: @mkrueger did you mean Task.FromResult(null) instead of null here?

Therzok: It's async, it'll be wrapped.

knocte: you mean that `await x()` works if x returns a null Task?

Therzok: It's not a null task. Inside:
```
async Task<X> GetX()
{
    return null;
}
```

is transformed into:
```
Task<X> GetX()
{
    return Task.FromResult<object>(null);
}
```

So `await GetX()` returns null.

slluis: It is not returning a null task, it is returning a task that with null ISymbol result.

knocte: ah right, my bad! sorry for the noise

Issue #4381, opened 2018-03-29
MetadataReferenceCache implementation does IO on the UI thread


Therzok:
```
0.44%   <RunInMainThread>b__39_0  •  2,885 ms  •  8,957 calls  •  MonoDevelop.Core.Runtime+<>c.<RunInMainThread>b__39_0(Object)
  0.33%   CheckForChanges  •  2,145 ms  •  132 calls  •  MonoDevelop.Ide.TypeSystem.MetadataReferenceCache.CheckForChanges
    0.33%   CheckForChange  •  2,120 ms  •  61,512 calls  •  MonoDevelop.Ide.TypeSystem.MetadataReferenceCache+MetadataReferenceCacheItem.CheckForChange
       0.32%   GetLastWriteTimeUtc  •  2,104 ms  •  61,512 calls  •  System.IO.File.GetLastWriteTimeUtc(String)
       0.00%   op_Inequality  •  3 ms  •  61,512 calls  •  System.DateTime.op_Inequality(DateTime, DateTime)
     0.00%   6 functions hidden  •  8 ms total  •  123684 calls total
   0.11%   10 functions hidden  •  735 ms total  •  17782 calls total
```

This should be changed to use FSW

Therzok:
Other than that, it seems to be the cause of our duplicate references in the csproj files:
https://gist.github.com/Therzok/ed70ee188f05433c01ac89f4071cdf94

Therzok:
About https://github.com/mono/monodevelop/issues/4381#issuecomment-378595116, seems like there's a roslyn fix.

https://github.com/dotnet/roslyn/pull/26438/files

xamarin-release-manager:
> [sync] [VS-58] Comment by **xamarinc**



Fixed in version 8.6.0.802 (master)



Pull Request #4683 merged by: Matt Ward  
 Author: mono  
 Commit: 724ab15995b7aa0cdca47fefebbc8fc30a23040c (xamarin/monodevelop)

PR #4478, opened 2018-04-09
Fix queued sheets

iainx:
[Mac] Correctly queue modal sheets

If you change the current branch outside of Monodevelop and certain files change then `ProjectOperations.AllowSave` will be called and it will display a sheet. Sometimes however, if multiple files change, the first sheet is shown, and the second is shown at the same time as a dialog.

The sheet is shown in `MacAlertDialogHandler` using `BeginSheet`
`BeginSheet` starts a modal sheet, but does it without changing the Runloop. To
block for a response we need to call `NSApplication.RunModalForWindow`.

It seems like when `RunModalForWindow` is called on a sheet, it does not stop processing the mainloop because when `AllowSave` still creates a second sheet. This sheet is then queued for showing once the first sheet is closed.

According to the documentation

> If it is not already visible, the window is centered on the screen using the value in its center method and made visible

Because the second sheet has been queued for display, `RunModalForWindow` decides to show it, and it appears as a dialog. This time, because `RunModalForWindow` has shown a dialog, the runloop is changed and processing stops showing any more dialogs until this one is shown.

So instead of calling `RunModalForWindow` with the new `NSAlert` calling it with
the main application window changes the runloop and stops processing until the sheet is closed.

> Fixes VSTS #937831

### DIFF
--- a/main/src/addins/MacPlatform/Dialogs/MacAlertDialogHandler.cs
+++ b/main/src/addins/MacPlatform/Dialogs/MacAlertDialogHandler.cs
@@ -218,7 +218,12 @@ namespace MonoDevelop.MacIntegration
 							NSApplication.SharedApplication.StopModal ();
 						});
 
-						NSApplication.SharedApplication.RunModalForWindow (alert.Window);
+						// pass parent and not alert so that the Runloop will change
+						// and processing will stop until the sheet is closed.
+						// If we pass alert, then it will run until a second alert is created
+						// which will be shown as a dialog and then the runloop changes and
+						// processing stops
+						NSApplication.SharedApplication.RunModalForWindow (parent);
 					}
 				}
 


iainx:
This is an example of the bug

screen shot 2018-04-09 at 3 10 57 pm

(NB: This patch doesn't fix the other problem that the whole process that calls `AllowSave` can spam the user with multiple (IMO) useless dialogs. I've had about 30 dialogs shown for one branch change in some cases - this patch simply streamlines them so that all appear correctly as sheets one after the other, which might be an issue somewhere else)

Therzok:
The underlying cause of the multiple dialogs: https://github.com/mono/monodevelop/issues/4381#issuecomment-378595116

Therzok:
Can we rebase and merge this?

Issue #4494, opened 2018-04-10
Memory retention because of TextEditorExtension


Therzok:
Indentation text editor extension caches the document option set, which causes the whole document state to be preserved for a document.
<img width="1654" alt="screen shot 2018-04-10 at 15 16 18" src="https://user-images.githubusercontent.com/109974/38556463-6e51c0e0-3cd2-11e8-8d66-c25296ef08ea.png">

PR #4497, opened 2018-04-10
Fixes issue #4494 Memory retention because of TextEditorExtension

mkrueger:
Try it out if it really fixes the issue.

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Formatting/CSharpTextEditorIndentation.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Formatting/CSharpTextEditorIndentation.cs
@@ -95,11 +95,12 @@ namespace MonoDevelop.CSharp.Formatting
 			}
 		}
 
-		void RunFormatter (MonoDevelop.Ide.Editor.DocumentLocation location)
+		async void RunFormatter (MonoDevelop.Ide.Editor.DocumentLocation location)
 		{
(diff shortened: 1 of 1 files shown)


main/src/addins/CSharpBinding/MonoDevelop.CSharp.Formatting/CSharpTextEditorIndentation.cs
```diff
@@ -95,11 +95,12 @@ internal void SafeUpdateIndentEngine (int offset)
 			}
 		}
 
-		void RunFormatter (MonoDevelop.Ide.Editor.DocumentLocation location)
+		async void RunFormatter (MonoDevelop.Ide.Editor.DocumentLocation location)
```
Therzok: Method is unused.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Formatting/CSharpTextEditorIndentation.cs
```diff
@@ -560,7 +561,7 @@ public override bool KeyPress (KeyDescriptor descriptor)
 			return result;
 		}
 
-		void HandleOnTheFlyFormatting (KeyDescriptor descriptor)
+		async void HandleOnTheFlyFormatting (KeyDescriptor descriptor)
```
Therzok: This is not an event handler, should not be async void.

Therzok:
Problem is fixed, but don't think this is the right fix (making methods async void, that is)

Therzok:
Going to close this for now, discussing with roslyn folks.

PR #4510, opened 2018-04-11
Test remove timeouts

nosami:
Removes timeouts from F# Interactive tests and kills processes when done.

One of these tests would fail roughly one in twenty times.


nosami:
@monojenkins backport d15-7

PR #4516, opened 2018-04-11
[Core] Fix files not re-opened on opening a solution

mrward:
If the solution used a custom BaseDirectory then the user preferences
were saved inside the custom base directory and then not loaded on
re-opening the solution. A custom base directory can be configured
in solution preferences - General - Main Settings and then changing
the Root Directory. This then adds a section to the solution file:

GlobalSection(MonoDevelopProperties) = preSolution
	BaseDirectory = ..
EndGlobalSection

Now the user preferences are always stored in a .vs directory
relative to the solution file itself instead of using the Solution's
BaseDirectory.

Fixes VSTS #598545 When opening a solution, last files are not
re-opened correctly


mrward:
@monojenkins backport d15-7

PR #4524, opened 2018-04-11
Fix VSTS 586152: Preserve documents on project reload

KirillOsenkov:
DotNetProject.OnReevaluateProject causes a LoadProject call where we attempt to incrementally preserve documents from the previous project. The CreateDocuments call incorrectly passed the current projectData to GetOrCreateDocumentId, and the document was not found there and was recreated from scratch.

Simple fix is to pass the oldProjectData so we preserve the DocumentId and other info from the previous project data.

Symptoms included lack of classification when a .cshtml document was opened on project open.


KirillOsenkov:
Note that line 606 was correct already, only the non-.cs documents were impacted.

KirillOsenkov:
@monojenkins backport master

PR #4526, opened 2018-04-11
[Mac] Remove AssemblyLoad hack

Therzok:
This seems not to be needed. The original report where Macaque would break
the top menu seems to not reproduce anymore.

Fixes VSTS #598766


Therzok:
@monojenkins backport d15-7

Issue #4521, opened 2018-04-11
[Build] Improve build speeds (ResolveAssemblyReferences not using cache file)


Therzok:
Currently, it seems like ResolveAssemblyReferences is not using the cache file, as the inputs are not the same.

Work by examining the buildlog for both MonoDevelop and md-addins to not cause invalidation and improve build speeds.

> VS bug [#599108](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/599108)

Therzok:
http://source.roslyn.io/#BuildBoss/StructuredLoggerCheckerUtil.cs,93f7b519d87a3b83

These analyzers are the kind we want to integrate in our build system.

Therzok:
Filed as https://github.com/mono/mono/issues/9657

PR #4528, opened 2018-04-11
Build speed improvements

Therzok:
This is the first part of fixing the build so it's snappy.

Right now, projects cause ResolveAssemblyReferences to be re-evaluated a lot of times and this just works towards fixing NOP builds to not compile anything.

Will add a CI based test that verifies that `make` does not contain targets we should be skipping.

TODO:

* [ ] Integrate with CI so we assert that things don't get rebuilt - maybe integrate a binlog parser and some msbuild task that automates this process
* [x] ~Fix md-addins~ - Will do in a later step to unblock this
* [x] Figure out why RAR takes a long time - Seems like nugets which contain references to shims happen to cause issues with resolving assemblies from the GAC. System.Runtime versions from nugets do not match the versions in the Mono GAC, so mono ends up probing a lot of directories trying to find the right versions.
* [ ] Remove tests local copies of MD assemblies, they should be loaded from mdtool/vstool
* [ ] Revert monodoc change for now.

Edit:
# What is this mastodon PR?

This PR tries to fix some common problems we have in how our projects reference eachother and nugets, avoiding issues with incremental builds.

It tries to consolidate nuget versions into props files and only MonoDevelop.Core has to do the NuGet restore step.

It also fixes issues with mismatched versions by adding some binding redirects.

Unifies assembly references and only local copies from where it should be.

# What do we gain?

Better incrementality.

We removed the system dependency on monodoc, we now get it from nuget, but it's a bit hacky:
https://github.com/mono/api-doc-tools/issues/229
https://github.com/mono/api-doc-tools/issues/228
https://github.com/mono/api-doc-tools/issues/227

# There are some missing changes

I reverted the changes in F# until we have a consensus about switching from Paket to this model.

I haven't opened PRs yet to mono-addins, debugger-libs and NRefactory

# Current issues

PackageReference would solve a lot of our issues, and we'd only need to have versions defined, and not manual references to different assemblies, right now, I used what nuget added to the csproj files.

### DIFF
--- a/main/external/Makefile.am
+++ b/main/external/Makefile.am
@@ -2,33 +2,36 @@ TARBALL_PATH=../../tarballs/external
 XAMMAC_PATH=/Library/Frameworks/Xamarin.Mac.framework/Versions/Current
 XAMMAC_LIB_PATH=$(XAMMAC_PATH)/lib
 XAMMAC_PDB=$(XAMMAC_LIB_PATH)/@MAC_ARCHITECTURE@/full/Xamarin.Mac.pdb
-all:
-
-clean:
 
 if ENABLE_MACPLATFORM
 all: Xamarin.Mac.dll Xamarin.Mac.registrar.full.a libxammac.dylib Xamarin.Mac.buildinfo
 	$(MAKE) -C monomac/src
 
 Xamarin.Mac.dll: $(XAMMAC_LIB_PATH)/@MAC_ARCHITECTURE@/full/Xamarin.Mac.dll
-	cp -p $< $@
-	if [[ -f $<.mdb ]]; then cp -p $<.mdb $@.mdb; fi;
-	if [[ -f $(XAMMAC_PDB) ]]; then cp -p $(XAMMAC_PDB) Xamarin.Mac.pdb; fi;
+	cp $< $@
+	if [[ -f $<.mdb ]]; then cp $<.mdb $@.mdb; fi;
+	if [[ -f $(XAMMAC_PDB) ]]; then cp $(XAMMAC_PDB) Xamarin.Mac.pdb; fi;
 
 Xamarin.Mac.registrar.full.a: $(XAMMAC_LIB_PATH)/mmp/Xamarin.Mac.registrar.full.a
-	cp -p $< $@
+	cp $< $@
 
 libxammac.dylib: $(XAMMAC_LIB_PATH)/libxammac.dylib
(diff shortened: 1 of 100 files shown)


main/external/Makefile.am
```diff
@@ -53,4 +56,3 @@ dist-clean:
 		fi \
 	done
 
-.PHONY: Xamarin.Mac.dll
```
Therzok: This was causing Xamarin.Mac to always be copied from `make`.

Therzok:
Working towards this: https://github.com/mono/monodevelop/issues/4521

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.csproj
```diff
@@ -190,7 +190,6 @@
     <Reference Include="System.Runtime.Serialization" />
     <Reference Include="System.Threading.Tasks.Dataflow, Version=4.6.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
       <HintPath>..\..\..\packages\System.Threading.Tasks.Dataflow.4.7.0\lib\netstandard1.0\System.Threading.Tasks.Dataflow.dll</HintPath>
-      <Private>True</Private>
```
KirillOsenkov: >True [](start = 15, length = 4)

Maybe an explicit False just in case?

Therzok: Sure! But it's not appearing as double right in binlog anymore

Therzok:
So, we can't get a SharpZipLib that works for monodoc and AndroidSDK manager. Maybe we can get upstream monodoc to work with the stripped version.

KirillOsenkov:
Looking great so far! :+1:

Therzok:
I'm considering reverting the monodoc change for now and file new issues for migrating monodoc and mono.posix gac usage to nuget.

Therzok:
Temporary monodoc solution: create a locally nupkg that we publish via xampubbot or myget.

Will redo binding redirect fixes based on what Windows reports and I'll file bugs to mono/msbuild about the inconsistencies on mac.

Therzok:
I'll split this PR into atomic PRs. Better than reviewing one massive PR

Therzok:
Closing. We already had a migration to PackageReference. If we want to fix rebuilding again, we need to redo the PR completely.

PR #4530, opened 2018-04-12
Invalidate cached line upon TextArea.RedrawLine

ToddGrun:
DO NOT MERGE. This PR is strictly to demonstrate a fix I have locally that fixes some colorization issues.

The commented out code in TagBasedSyntaxHighlighting is what I was doing if I wanted to keep this as a tactical fix just for VS classifications (and the textarea.cs change wouldn't be necessary). There is a more widespread issue in colorization outside razor that I can reproduce in c#.

C# repro:
 Create a new class.
 In a different file, reference that class (notice it colorizes correctly)
 Go back to new class file and modify the class name.
 Go back to referencing file and notice that it still colorizes the old class name as if it were an existing type.

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
@@ -1084,6 +1084,9 @@ namespace Mono.TextEditor
 		{
 			if (isDisposed || logicalLine > LineCount || logicalLine < DocumentLocation.MinLine)
 				return;
+
+			textViewMargin.RemoveCachedLine(logicalLine);
+
 			double y = LineToY (logicalLine) - this.textEditorData.VAdjustment.Value;
 			double h = GetLineHeight (logicalLine);
 
--- a/main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
@@ -142,9 +142,9 @@ namespace Microsoft.VisualStudio.Platform
                 int startLineIndex = this.textDocument.OffsetToLineNumber (args.ChangeSpan.Start);
                 int endLineIndex = this.textDocument.OffsetToLineNumber (args.ChangeSpan.End);
 
-                IEnumerable<IDocumentLine> documentLines = this.textDocument.GetLinesBetween (startLineIndex, endLineIndex);
-                foreach(IDocumentLine documentLine in documentLines)
+                for (int curLineIndex = startLineIndex; curLineIndex <= endLineIndex; curLineIndex++)
                 {
+                    IDocumentLine documentLine = this.textDocument.GetLine(curLineIndex);
                     handler(this, new LineEventArgs(documentLine));
                 }
             }


KirillOsenkov:
It would be nice to check this in or a variant of this one to fix the various classification issues we've been seeing.

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
```diff
@@ -1084,6 +1084,9 @@ internal void RedrawLine (int logicalLine)
 		{
 			if (isDisposed || logicalLine > LineCount || logicalLine < DocumentLocation.MinLine)
 				return;
+
+			textViewMargin.RemoveCachedLine(logicalLine);
```
mkrueger: This lowers drawing performance for some operations - where do you need that ?

main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
```diff
@@ -141,10 +142,13 @@ private void OnClassificationChanged(object sender, ClassificationChangedEventAr
...
 
-                IEnumerable<IDocumentLine> documentLines = this.textDocument.GetLinesBetween (startLineIndex, endLineIndex);
-                foreach(IDocumentLine documentLine in documentLines)
+                for(int curLineIndex = startLineIndex; curLineIndex <= endLineIndex; curLineIndex++)
                 {
+                    IDocumentLine documentLine = this.textDocument.GetLine(curLineIndex);
                     handler(this, new LineEventArgs(documentLine));
+
+                    //var view = (this.textDocument as TextEditor).GetContent<SourceEditorView>();
+                    //view.TextEditor.TextViewMargin.RemoveCachedLine(curLineIndex);
```
mkrueger: The old version is more efficient it was O(log n  + endLine-startLine)  now it's O((endLine-startLine)* log n)

line has a NextLine property which is more efficient so getting  startLine and using .NextLine on it will maybe be an option ?

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
ToddGrun: For me, it's when the tagbasedsyntaxhighlighter calls down into this. The C# need for something like this is in the original comment.

main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
ToddGrun: The old version didn't work. It would only work for lines *between* startLine and endLine. I want to include those lines. Actually, I see both the TextDocument.GetLine and DocumentLine.NextLine call into the textsnapshot to get the line from the line number. Not seeing where a perf difference would be.

ToddGrun: Also, the GetLinesBetween implementation essentially does the same thing I do with slightly different endpoint conditions.

mkrueger: Ah forget that comment that was my implementation - but still that's a bad implementation - we do not need to wonder about editor slowness when we're using index access all over the place in a tree based data structure.

At least I hope that the line representation is tree based…

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
mkrueger: Shouldn't screw up perf. too badly RedrawLine is only called in very few operations.

ToddGrun:
So, what's the consensus here? Should I go ahead and clean this up a bit keeping the RedrawLine change? What about RedrawLines?

mkrueger:
Just keep the redraw line change. It's ok.

KirillOsenkov:
@slluis I think this is ready to go in to d15-7. Please merge if builds passing.

KirillOsenkov:
@monojenkins backport master

PR #4531, opened 2018-04-12
[529478][Editor] Fix search marker fallback color

sevoku:
This is a possible fix for less visible search result markers. If the search result highlighting color of the current editor theme is not visible enough, a default marker color will be used instead.

TODO:
* Final marker color (same for dark/light IDE?)
* Needs a better color visibility detection than brightness comparison?

NOTE: making the theme color brighter based on the current background is too complex. We have simple (brightness based) calculation in [`SearchResultWidget`](https://github.com/mono/monodevelop/blob/fix529478-editor-search-marker-fallback-color/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/SearchResultWidget.cs#L318), but it tends to calculate black/white colors for most themes, which doesn't look really good.


vancura:
I have some more changes on this branch, hence restoring.

vancura:
Ok, GitHub failed here. I thought that if I restore a PR branch and push there, that the PR will be updated, it was not. Hence, closing this, I'll prepare another set of PRs from master and 15.7.

PR #4533, opened 2018-04-12
Fix 597220: No IntelliSense for pages deriving from a custom ContentPage

DavidKarlas:
This is just MonoDevelop part of fix, what I did here is update XmlName/Root and Tag states to support having `< xmlns="formsNamespace"` and parse this as element with attributes, because until now this was not element element yet(because it doesn't have name) so attributes also were not parsed, resulting in Forms extension not being aware of referenced namespaces, hence code completion not working...

I included also 2 unrelated fixes in SourceEditorView.cs and SimpleBracketMatcher.cs which deal with file edge cases that I found while working on this.


DavidKarlas:
@monojenkins backport master

PR #4488, opened 2018-04-10
D15 6 fix592988

mkrueger:
Test case:

Code should not flicker black while typing.

The /* should update next lines.

/* 
Code <- should be commented.

The old fix broke a work around that fixed the /* updating.

Fixes VSTS #594738.

PR #4534, opened 2018-04-12
Master fix multispan highlighting

mkrueger:
forward port of https://github.com/mono/monodevelop/pull/4488

That adds a better solution for the problem + unit tests.


mkrueger:
@monojenkins backport d15-7

PR #4540, opened 2018-04-12
Quote executagble paths in <Exec /> in MonoDevelop.Core.csproj

directhex:
This fixes build failures on Windows with users whose MonoDevelop git
checkout is in a path with a space in it.


Therzok:
Triggered a build.

PR #4546, opened 2018-04-13
Fixes VSTS 597271: [Feedback] Getting "The type initializer for

mkrueger:
'Mono.TextEditor.TextViewMargin' threw an exception." errors from VS
Community (Mac) on macOS 10.13.5 Beta

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/597271

Seems to be a bug in the macOS beta but it can be handled.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/OSXEditor.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/OSXEditor.cs
@@ -34,24 +34,30 @@ namespace MonoDevelop.Ide.Editor
 {
 	class OSXEditor
 	{
+		static bool hasLoaded;
 		static Xwt.Drawing.Image image;
+		static object loadLock = new object ();
 
 		public static Xwt.Drawing.Image IBeamCursorImage {
 			get {
-				if (image != null)
-					return image;
-				try {
-					var cacheFileName = Path.Combine(UserProfile.Current.CacheDir, "MacCursorImage.tiff");
-					if (!File.Exists(cacheFileName)) {
-							NSCursor.IBeamCursor.Image.AsTiff().Save(cacheFileName, true);
+				lock (loadLock) {
+					if (hasLoaded)
+						return image;
+					try {
+						var cacheFileName = Path.Combine (UserProfile.Current.CacheDir, "MacCursorImage.tiff");
+						if (!File.Exists (cacheFileName)) {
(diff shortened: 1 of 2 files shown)


mkrueger:
@monojenkins backport d15-7

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/OSXEditor.cs
```diff
@@ -40,13 +40,19 @@ class OSXEditor
...
+					if (!File.Exists(cacheFileName)) {
+							NSCursor.IBeamCursor.Image.AsTiff().Save(cacheFileName, true);
+					}
+					var img = Xwt.Drawing.Image.FromFile(cacheFileName);
+					var size = NSCursor.IBeamCursor.Image.Size;
+					image = img.WithSize(size.Width, size.Height);
+					return image;
+				} catch (Exception e) {
+					LoggingService.LogError("Error while getting IBeam cursor image.", e);
+					return null;
```
slluis: If this fails, it will try to load the image over and over, and will log an error every time. It shouldn't try to load the image again.
In any case it would be interesting to know why is it failing.

mkrueger: This is loaded in a static constructor in the editor how does a static constructor fail over and over ?

Therzok: You're not setting `image` to anything.

slluis: If for some reason this call is moved somewhere else it can cause trouble. It is not hard to guard against repeated failures, so let's do it.

Therzok:
I was this failing in AspNet unit tests locally because it was constructed from a background thread. Maybe we should move it out of a static constructor?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/OSXEditor.cs
mkrueger: that's true but it shouldn't be called more than once in any case - because it loads an image from disk.

mkrueger:
TextView should always be constructed in the UI thread. It's not thread safe.

Issue #4544, opened 2018-04-12
Dragging a Forms XAML page up the compile order in FX doesn't move the .xaml.cs file


jimbobbennett:
__To reproduce__
* Create a new blank Xamarin.Forms app in F#. This will have an existing 'Hello World' page.
* Add a new Content Page (XAML) called NewPage
* Drag the NewPage.xaml file above the HelloWorld.xaml in the solution explorer
* The solution explorer shows the NewPage.xaml and nested NewPage.xaml.fs files in the right place
* Reference the NewPage from HelloWorld.fs - the compiler complains.

If you check the .fsproj file, the NewPage.xaml file is above the HelloWorld.fs page, but NewPage.xaml.fs is still at the bottom of the compile order. You have to manually move it up.

> VS bug [#599769](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/599769)

xamarin-release-manager:
> [sync] [VS-12] Comment by **xamarinc**



Fixed in version 8.6.0.802 (master)



Pull Request #4552 merged by: Jason Imison  
 Author: mono  
 Commit: c6a5590bb2b09821d305ae4826997f5d7ef2d016 (xamarin/monodevelop)

PR #4552, opened 2018-04-13
Allow XAML files and codebehind files to be moved as a single unit

nosami:
Fixes VSTS #599769 (#4544)


nosami:
@monojenkins backport d15-7

PR #4555, opened 2018-04-13
Fix 591345: [Updater] "Updates have been downloaded and are ready to …

DavidKarlas:
…install" still shown after updating to Alpha

Problem was with changes around MonoDevelop.Ide.AddinUpdater.UpdateChannel/UpdateLevel migration where `idx` was not set for each channel resulting in UpdateChannel.Equals returning wrong values causing https://github.com/xamarin/md-addins/blob/5b4c8d9/Xamarin.Ide/Xamarin.Ide/Xamarin.Ide.Updater/UpdatesDialog.cs#L278 to not execute and update list of updatable items.


slluis:
@DavidKarlas please make sure this is fixed in master too

DavidKarlas:
@monojenkins backport master

PR #4562, opened 2018-04-13
[Templates] Added ItemTemplate.GetImage() for loading themed images

jstedfast:
This is needed by the new Azure Functions templates so that the
"New Function" dialog can render theme-appropriate icons based
on the icons shipped with the Azure templates.


slluis:
Nice fix!!

PR #4563, opened 2018-04-13
[Core] Fix MSBuild items added for globs with link metadata

mrward:
A project containing the following MSBuild file globs would have
extra MSBuild items added when the project file was saved.

```
<ItemGroup>
  <Compile Include="..\**\*.cs">
    <Link>%(RecursiveDir)%(Filename)%(Extension)</Link>
  </Compile>
</ItemGroup>
```

On saving the project Compile Update items would be added for each
file included by the file glob.

```
<ItemGroup>
  <Compile Update="..\Test\Class1.cs">
    <Link>Class1.cs</Link>
  </Compile>
</ItemGroup>
```

Fixes VSTS #527721 - Overriding globs

Also re-worked the fix for VSTS #569295 - Projects that includes files using wildcards will expand (and remove) wildcards upon save. Now no evaluation of the MSBuild property is done on saving the project.


mrward:
@monojenkins backport d15-7

PR #4565, opened 2018-04-13
[d15-7][Templates] Added ItemTemplate.GetImage() for loading themed images

jstedfast:
This is needed by the new Azure Functions templates so that the
"New Function" dialog can render theme-appropriate icons based
on the icons shipped with the Azure templates.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/ItemTemplate.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/ItemTemplate.cs
@@ -45,17 +45,12 @@ namespace MonoDevelop.Ide.Templates
 
 		public virtual Stream GetStream (string path)
 		{
-			return null;
+			return File.OpenRead (path);
 		}
 
 		public virtual Image GetImage (string path)
 		{
-			var stream = GetStream (path);
-
-			if (stream == null)
-				return null;
-
-			return Image.FromStream (stream);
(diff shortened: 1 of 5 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/ItemTemplate.cs
```diff
@@ -45,5 +47,15 @@ public virtual Stream GetStream (string path)
...
 		}
+
+		public virtual Image GetImage (string path)
+		{
+			var stream = GetStream (path);
+
+			if (stream == null)
+				return null;
+
+			return Image.FromStream (stream);
```
slluis: Why not just use `Image.FromFile(path)`? Otherwise, alternative styles of the image won't be loaded.

slluis: Why not just use `Image.FromFile(path)`?

jstedfast: Ok, so the problem is that GetStream()'s default impl returns null, but if it returned `File.OpenRead(path)`, I think it would be reasonable to use Image.FromFile().

slluis: Yes, I mean for the default implementation.

Issue #4393, opened 2018-03-29
Error in text editor


mhutch:
Version 7.4 build 985

This error looks like it's written by a WriteLine rather than being reported to the logging system.

May be related to #4392 as it was immediately after it in the log.

```
error while DiffTracker.TrackDocument_TextChanging:System.ArgumentOutOfRangeException: 0 <= length, index(69)+count <= 91
Parameter name: count
Actual value was 31.
  at Mono.TextEditor.Utils.CompressingTreeList`1[T].RemoveRange (System.Int32 index, System.Int32 count) [0x0016a] in /Users/builder/data/lanes/5810/0bed5a33/source/monodevelop/main/src/core/Mono.TextEditor.Shared/Mono.TextEditor.Utils/CompressingTreeList.cs:246 
  at Mono.TextEditor.DiffTracker.TrackDocument_TextChanging (System.Object sender, MonoDevelop.Core.Text.TextChangeEventArgs e) [0x00060] in /Users/builder/data/lanes/5810/0bed5a33/source/monodevelop/main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Document/DiffTracker.cs:80 
```

> VS bug [#593845](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/593845)

mhutch:
I still see a bunch of writelines in difftracker... we should convert them to logging so they show up in fault telemetry.

xamarin-release-manager:
Fixed in version 7.6.0.557 (master)

Pull Request #4573 merged by: Lluis Sanchez
Author: mono
Commit: 0a9d5590b5e68857ef98b96b3b70ca69c01cbd9b (mono/monodevelop)

xamarin-release-manager:
Fixed in version 7.6.0.557 (master)

Pull Request #4573 merged by: Lluis Sanchez
Author: mono
Commit: 0a9d5590b5e68857ef98b96b3b70ca69c01cbd9b (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #4573 merged by: Lluis Sanchez
Author: mono
Commit: 0a9d5590b5e68857ef98b96b3b70ca69c01cbd9b (xamarin/monodevelop)

PR #4573, opened 2018-04-16
Fixes issue #4393 Error in text editor The original error should be

mkrueger:
already fixed (the line numbers indicate it). However the text editor
errors should be printed to the log and not the console. This had
historical reasons because the editor was stand alone and had no
connection to the Core/Ide and no own logging capabilities.

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor.Theatrics/BounceFadePopupWindow.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor.Theatrics/BounceFadePopupWindow.cs
@@ -27,6 +27,7 @@
 
 using System;
 using Gdk;
+using MonoDevelop.Core;
 
 namespace Mono.TextEditor.Theatrics
 {
@@ -349,7 +350,7 @@ namespace Mono.TextEditor.Theatrics
 					Draw (cr, userspaceArea);
 				}
 			} catch (Exception e) {
-				Console.WriteLine ("Exception in animation:" + e);
+				LoggingService.LogError ("Exception in animation:" + e);
 			}
 			return true;
 		}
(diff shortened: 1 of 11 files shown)


main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor.Theatrics/BounceFadePopupWindow.cs
```diff
@@ -349,7 +350,7 @@ protected override bool OnExposeEvent (EventExpose evnt)
 					Draw (cr, userspaceArea);
 				}
 			} catch (Exception e) {
-				Console.WriteLine ("Exception in animation:" + e);
+				LoggingService.LogError ("Exception in animation:" + e);
```
slluis: The exception should be passed as an argument to LogError, not concatenated to the message.

slluis: Same for others below.

main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Actions/ClipboardActions.cs
```diff
@@ -214,7 +214,7 @@ static CopyOperation ()
 							try {
 								copyData = pasteHandler.GetCopyData (segment.Offset, segment.Length);
 							} catch (Exception e) {
-								Console.WriteLine ("Exception while getting copy data:" + e);
+								LoggingService.LogError ("Exc eption while getting copy data:" + e);
```
slluis: "Exc eption" -> "Exception"

main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/TextEditorOptions.cs
```diff
@@ -396,7 +397,7 @@ protected void DisposeFont ()
 					try {
 						font = Pango.FontDescription.FromString (FontName);
 					} catch {
-						Console.WriteLine ("Could not load font: {0}", FontName);
+						LoggingService.LogError ("Could not load font: {0}", FontName);
```
Therzok: Maybe also log the exception here?

PR #4574, opened 2018-04-16
Logpoint icons

vancura:
Some new icons for the future logpoint feature:

run

(Only highlighted are new).


slluis:
Isn't logpoint the same as tracepoint?
Also, why logpoint is round as command icon and square as gutter icon?

vancura:
> Isn't logpoint the same as tracepoint?

Well, I can't answer this one. Tracepoint icons were in the repo since ever, but I am not sure they were used at all.

> Why logpoint is round as command icon and square as gutter icon?

First, we need to see if we can use tracepoint icons for logpoints. Then we can kill one or the other, and unify the gutter icons.

slluis:
AFAIK, tracepoints are logpoints are the same. When you edit a breakpoint and you change the action from break to print text, then the icon is changed to tracepoint everywhere.

vancura:
Hm, ok. Then we don't need this PR :)

PR #4579, opened 2018-04-16
Fixes VSTS 591324: Multiple Popup save dialog when close using

mkrueger:
keyboard shortcut multiple times

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/591324

This method isn't called anywhere else.


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileCommands.cs
```diff
@@ -148,15 +148,22 @@ protected override void Run ()
 	// MonoDevelop.Ide.Commands.FileCommands.CloseAllFiles
 	public class CloseAllFilesHandler : CommandHandler
 	{
+		static bool isRunning;
+
 		protected override void Run ()
```
Therzok: `override async Task Run()`

PR #4586, opened 2018-04-17
[Ide] Switched to roslyn folding infrastructure.

mkrueger:


main/src/addins/CSharpBinding/MonoDevelop.CSharp.Parser/CSharpParsedDocument.cs
```diff
@@ -306,181 +306,9 @@ public override void VisitThrowStatement (Microsoft.CodeAnalysis.CSharp.Syntax.T
 			}
 		}
 
-		IReadOnlyList<FoldingRegion> foldings;
-		SemaphoreSlim foldingsSemaphore = new SemaphoreSlim (1, 1);
+		static readonly Task<IReadOnlyList<FoldingRegion>> foldings  = Task.FromResult((IReadOnlyList<FoldingRegion>)new FoldingRegion[0]);
 
-		public override Task<IReadOnlyList<FoldingRegion>> GetFoldingsAsync (CancellationToken cancellationToken = default(CancellationToken))
```
Therzok: Maybe we should also remove the GetCommentsAsync contents?

mkrueger: GetCommentsAsync has nothing to do with the folding infrastructure.
But it's unused AFAIK so safe to remove but as said it's unused so there is no huge need for that either…

Issue #4566, opened 2018-04-13
Enabling code folding by default


mhutch:
Developers often ask for code folding to be supported. We have this feature, but since it's off by default (unlike any other editor or IDE with this feature I can think of), they assume we don't have it and don't even look for the feature.

We should enable this by default (including for existing users).

Additionally, we should reduce the visual noise by following what VS Code does:

* the current version of the fold margin is only displayed when hovering over it
* when not hovering, the fold margin only shows [+] expand markers.

> VS bug [#600192](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/600192)

mkrueger:
I remember that I once implemented an on hover solution ... but no one wanted it at that time :/.

slluis:
@migueldeicaza ok to change the default behavior?

Therzok:
http://source.roslyn.io/#Microsoft.CodeAnalysis.EditorFeatures.Wpf/Structure/VisualStudio15StructureTaggerProvider.cs,99d968331621305b

They seem to provide a tagger for IBlockTag. We should consider switching to it.

Therzok:
Here is the code used for outlining: https://github.com/dotnet/roslyn/tree/1ead9e9c4dab6dc5092c1690b560dd2f113a236c/src/EditorFeatures/Core.Wpf/Structure

nosami:
Would it also be possible to stop the folds from expanding when you move the caret across a fold? This makes folds almost pointless IMO (unless you only ever move the caret using the mouse)

mkrueger:
The folding translation has nothing to do with the default behavior - do not mix up things up. 

For C# the parsed document should be unused in any case btw.

Therzok:
@mkrueger yeah, it's currently used, at least for the path selector.

mhutch:
@nosami  you should file a separate issue for that

mkrueger:
@Therzok : I didn't mean it is unused now - I work on the parsed document usage from time to time to remove it. It's not a super urgent thing to do but the goal is not longer using the parsed document infrastructure for c#. Roslyn is a complete replacement for all the parsed document features.

However we'll need it for xml or other languages for a while because roslyn is not really extensible enough to provide everything yet. So it's a poor mans solution for the other use cases we have.

xamarin-release-manager:
Fixed in version 7.6.0.683 (master)

Pull Request #4587 merged by: Mikayla Hutchinson
Author: mono
[...]

PR #4587, opened 2018-04-17
Fixes issue #4566 Enabling code folding by default

mkrueger:
https://github.com/mono/monodevelop/issues/4566

Should match the VS.NET default options.

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor.Theatrics/Stage.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor.Theatrics/Stage.cs
@@ -80,14 +80,14 @@ namespace Mono.TextEditor.Theatrics
             }
         }
 
-        public Actor<T> Add (T target, uint duration)
+		public Actor<T> Add (T target, uint duration, double percent = 1.0)
         {
             lock (this) {
                 if (Contains (target)) {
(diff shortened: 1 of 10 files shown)


mhutch:
Will this change it for existing users?

Also this change was contingent on the hover behavior.

mhutch:
(also, any chance we could do the fade in / fade out?)

slluis:
Can you make a screencast of the new behavior?

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor.Theatrics/Stage.cs
```diff
@@ -80,14 +80,14 @@ public Actor<T> Add (T target)
             }
         }
 
-        public Actor<T> Add (T target, uint duration)
+        public Actor<T> Add (T target, uint duration, double percantage = 1.0)
```
Therzok: typo: percentage

mkrueger:
When it's merged and shipped it will affect users which installed that version.

iainx:
Just a heads up, this has broken accessibility both keyboard and VoiceOver, because the fold markers need to be drawn if the fold margin is focused by the keyboard, or if VoiceOver is operational.

slluis:
@iainx can you please file a bug report for this?

iainx:
@slluis done

Issue #4518, opened 2018-04-11
Surface tracepoints on the editor


migueldeicaza:
We have a lot of power behind the IDE, but it is not always convenient to use.

For example, recently VSCode introduce a nice UI for adding logging statements to a running program and it was received with a lot of euphoria, the UI is simple, it is a contextual menu on top of a breakpoint:

![image](https://user-images.githubusercontent.com/36863/38621295-975899b8-3d6e-11e8-8976-f45023728b07.png)

With an inline editing experience:

![image](https://user-images.githubusercontent.com/36863/38621335-ace4c176-3d6e-11e8-91e7-7da73b1c228b.png)

We have the same capability, it is just not well known that it exists, and it is difficult to get to.   By not being as easy to use, it removes the incentive to use for quick tasks:

![image](https://user-images.githubusercontent.com/36863/38621186-524e3d6e-3d6e-11e8-9577-685a837dacc3.png)

We should better surface this capability.

mhutch:
This is a simplified version of what VS does. I think it would make a lot of sense to base our UX on that.

Short version -

* There's a streamlined inline (peek) editor for breakpoint properties
* A little toolbar appears when you move near a breakpoint icon in the margin that makes it super easy to open the properties
* There's an "Add Tracepoint" context menu command in the editor that creates a breakpoint, configures it as a tracepoint, and opens the breakpoint editor with the focus in the train e expression field

PR #4588, opened 2018-04-17
[Debugger] Add logpoints to the icon margin

iainx:
A log point is a breakpoint that just prints a message and continues. Add
support for them to the icon margin.

Requires https://github.com/mono/debugger-libs/pull/182

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.addin.xml
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.addin.xml
@@ -126,6 +126,9 @@
 				defaultHandler = "MonoDevelop.Debugger.NewCatchpointHandler"
 				_label = "New Exception Catchpoint"
 				icon = "md-catchpoint-new" />
+		<Command id = "MonoDevelop.Debugger.DebugCommands.ToggleLogpoint"
+				 defaultHandler = "MonoDevelop.Debugger.ToggleLogpointHandler"
+				 _label = "Toggle Logpoint" />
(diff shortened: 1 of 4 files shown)


mhutch:
How are these different from tracepoints?

mhutch:
Some context:
https://github.com/Microsoft/vscode/issues/45128
https://github.com/Microsoft/vscode-node-debug2/issues/163

It looks like VS Code is calling tracepoints "logpoints". However, since VS [also calls them tracepoints](https://blogs.msdn.microsoft.com/zainnab/2010/02/07/setting-a-tracepoint-in-source-code/), I think we should stick with the existing name for now.

I can see value in having a "new tracepoint" command (like VS) that opens up the breakpoint editor already configured as a tracepoint. But I'm not sure about "toggle tracepoint" - we can't simply toggle a tracepoint on, we need to ask for an expression, which means showing a UI, and toggle off is already covered by the breakpoint toggle command.

Note also that in VS the breakpoint menu is shown in the editor, not the margin - because breakpoints can have a column offset too when there are multiple statements on one line.

I wonder whether we should investigate making it easier to change breakpoint properties after creation with something like the floating toolbar and inline access breakpoint properties like VS: https://blogs.msdn.microsoft.com/devops/2014/10/06/new-breakpoint-configuration-experience-in-visual-studio-2015.

Therzok:
See https://github.com/mono/monodevelop/issues/4518

mhutch:
Yup. Definitely agree making this functionality .or discoverable and usable is good, just think it needs some design work.

jstedfast:
Based on the comments, it sounds like this should be closed? Or?

iainx:
Yeah, it was a quick proof of concept, so closed

PR #4589, opened 2018-04-17
[DotNetCore] Handle .NET Core 2.1 SDK not installed

mrward:
If a .NET Core 2.1 project is opened and .NET Core SDK 2.1.300 or
above is not installed then a dialog is shown allowing the SDK to
be downloaded. The project in the Solution window will show an error
icon with the same information about .NET Core 2.1 not being
installed. Currently the download link goes to the preview 2 page
- https://www.microsoft.com/net/download/dotnet-core/sdk-2.1.300-preview2
the url is shortened and can be redirected later without changing
the code once .NET Core 2.1 SDK final is released.

Fixes VSTS #573785 - Prompt to install .NET Core 2.1 SDK if .NET Core
2.1 project opened


mrward:
@monojenkins backport d15-7

PR #4592, opened 2018-04-17
Remove xbuild engine

mhutch:
xbuild is deprecated and msbuild support is solid, there's no reason to support xbuild any longer.

Getting rid of this baggage allows us to depend on things that only work in MSBuild.


mhutch:
@slluis ping?

slluis:
Good to go once the conflicts are fixed.

mhutch:
@slluis what conflicts?

Issue #4502, opened 2018-04-10
[MSBuild] Possible cause of slow MSBuild GetSourceFilesAsync evaluation


Therzok:
Investigating msbuild evaluation performance, I've come across a few details that might be slowing down actual MSBuild evaluation:

1. https://github.com/mono/monodevelop/issues/4501 - We're not using Design Time targets which make use of caches
2. All C# projects get invalidated 'in memory' by property changes on load - https://github.com/mono/monodevelop/blob/master/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CSharpProjectExtension.cs#L63

We should track which properties are modified on solution load and not modify in memory.

What happens here:
1. We set the `modifiedInMemory` flag to true, even for properties which are not serialized to the csproj - i.e. StockIcon - https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs#L405
2. If we have that flag, we'll end up querying UnsavedContent - http://source.monodevelop.com/#MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.Formats.MSBuild/BuildEngine.cs,86
3. That means we load the project from memory, instead of from disk, possibly messing up with the timestamps that msbuild uses internally for caches
4. we do `Directory.CreateDirectory` unconditionally here - https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.Formats.MSBuild/ProjectBuilder.v4.0.cs#L178
5. We do ProjectRootElement.Create without passing in a project collection, thus not reusing caches.

> VS bug [#597807](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/597807)

xamarin-release-manager:
> [sync] [VS-49] Comment by **xamarinc**



Fixed in version 8.6.0.802 (master)



Author: therzok  
 Commit: f5580cdf4300574a4ca344359fa850c2b9b4acc2 (xamarin/monodevelop)

Issue #4501, opened 2018-04-10
[MSBuild] Use actual design time build targets


Therzok:
See: https://github.com/Microsoft/msbuild/wiki/MSBuild-Tips-&-Tricks#visual-studio-design-time-intellisense-builds

We should replace CoreCompileDependsOn and co with these values

mrward:
Was investigating the design time MSBuild targets to see if I could work out how Visual Studio 2017 was getting information about the generated NuGet files for - [VSTS 636061](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/636061)

Was interesting to see that for .NET Core projects a set of targets is called in one go, as described in the MSBuild tips and tricks. For non sdk style projects that use PackageReferences it seemed that Visual Studio 2017 was calling just the Compile target, but not generating any assemblies. Did not investigate much further than that - just opening the solution into the Visual Studio 2017 and seeing what design time builds were logged. I added some notes in a doc attached to the VSTS bug report with more detail.

Therzok:
https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets

Therzok:
While DesignTime does call the actual target, the search paths are different:
https://source.dot.net/#MSBuildFiles/C/ProgramFiles(x86)/MicrosoftVisualStudio/2017/Enterprise/MSBuild/15.0/Bin_/Microsoft.Common.CurrentVersion.targets,2617
vs
https://source.dot.net/#MSBuildFiles/C/ProgramFiles(x86)/MicrosoftVisualStudio/2017/Enterprise/MSBuild/15.0/Bin_/Microsoft.Common.CurrentVersion.targets,565

That means the design time targets don't actually traverse the GAC folders - which on Mono are pretty huge.

There's also other configuration properties we can use, depending on how we call them, we could possibly do a recursive search for referenced items, etc.

Therzok:
RAR is now using [design time builds](https://github.com/mono/monodevelop/pull/7848/files#diff-bf46248a3a265d1dc44bf13c9a8d9e05R1123)

but there are a few others, like `CompileDesignTime` not using it.

PR #4593, opened 2018-04-17
Switch Assembly References to be resolved using Design Time targets

Therzok:
This PR aims to optimize MSBuild builders actual evaluation and target running:

* https://github.com/mono/monodevelop/issues/4502
* https://github.com/mono/monodevelop/issues/4501

The first commit aims to fix root element caching:

* Don't use a custom project collection, but use the actual global collection provided by msbuild. We won't use multiple collections in the same builder anyway.

The second commit adds root caching for unsaved content files:

* This adds unsaved content elements cached into the root project collection, and changes the reader to not preserve formatting and ignore dtd, similar to how msbuild does it.

The third commit removes the `ReevaluateIfNecessary()` calls that are not needed. Creating a ProjectInstance re-evaluates the project if it needs to: http://source.dot.net/#Microsoft.Build/Definition/Project.cs,2604

The fourth commit a simple way to initialize a evaluation context for design time builds.

WIP from here:
The fifth and so on switch actual GetReferencedAssemblies calls to use design time targets, instead of actual targets to resolve files to compile and so on.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
@@ -1026,13 +1026,13 @@ namespace MonoDevelop.Projects
 
 				var monitor = new ProgressMonitor ();
 
-				var context = new TargetEvaluationContext ();
-				context.ItemsToEvaluate.Add ("ReferencePath");
+				var context = new TargetEvaluationContext (designTime: true);
+				context.ItemsToEvaluate.Add ("_ReferencesFromRAR");
 				context.BuilderQueue = BuilderQueue.ShortOperations;
 				context.LoadReferencedProjects = false;
 				context.LogVerbosity = MSBuildVerbosity.Quiet;
 
-				var result = await RunTarget (monitor, "ResolveAssemblyReferences", configuration, context);
+				var result = await RunTarget (monitor, "ResolveAssemblyReferencesDesignTime", configuration, context);
 
 				refs = result.Items.Select (i => new AssemblyReference (i.Include, i.Metadata)).ToList ();
 
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
@@ -1028,11 +1028,14 @@ namespace MonoDevelop.Projects
 
 				var context = new TargetEvaluationContext (designTime: true);
 				context.ItemsToEvaluate.Add ("_ReferencesFromRAR");
+				context.ItemsToEvaluate.Add ("_ProjectReferencesFromRAR");
+				context.ItemsToEvaluate.Add ("_ResolvedNativeProjectReferencePaths");
+
 				context.BuilderQueue = BuilderQueue.ShortOperations;
 				context.LoadReferencedProjects = false;
 				context.LogVerbosity = MSBuildVerbosity.Quiet;
 
-				var result = await RunTarget (monitor, "ResolveAssemblyReferencesDesignTime", configuration, context);
+				var result = await RunTarget (monitor, "ResolveAssemblyReferencesDesignTime;ResolveProjectReferencesDesignTime", configuration, context);
 
 				refs = result.Items.Select (i => new AssemblyReference (i.Include, i.Metadata)).ToList ();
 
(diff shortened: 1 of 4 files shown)


main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.Formats.MSBuild/ProjectBuilder.v4.0.cs
```diff
@@ -134,7 +133,6 @@ public ProjectBuilder (BuildEngine buildEngine, ProjectCollection engine, string
 							project.RemoveGlobalProperty (p.Key);
 						foreach (var p in originalGlobalProperties)
 							project.SetGlobalProperty (p.Key, p.Value);
-						project.ReevaluateIfNecessary ();
```
mrward: Does this work without re-evaluating? I seem to remember having to do this to support global properties - https://github.com/mono/monodevelop/commit/4a6c1a980aa8f2e12cb21c3a305669806eccbdd0 - although there is a test for this so if that passes then I guess we are good 😄

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
```diff
@@ -1026,13 +1026,16 @@ internal protected virtual Task<List<AssemblyReference>> OnGetFacadeAssemblies (
...
+				context.ItemsToEvaluate.Add ("_ReferencesFromRAR");
+				context.ItemsToEvaluate.Add ("_ProjectReferencesFromRAR");
+				context.ItemsToEvaluate.Add ("_ResolvedNativeProjectReferencePaths");
+
 				context.BuilderQueue = BuilderQueue.ShortOperations;
 				context.LoadReferencedProjects = false;
 				context.LogVerbosity = MSBuildVerbosity.Quiet;
 
-				var result = await RunTarget (monitor, "ResolveAssemblyReferences", configuration, context);
+				var result = await RunTarget (monitor, "ResolveAssemblyReferencesDesignTime;ResolveProjectReferencesDesignTime", configuration, context);
```
mrward: Be good to see if this now means Xamarin.Mac full works without having to use a custom OnGetFacadeAssemblies implementation. Then some code in md-addins can be removed.

mrward: Maybe the Facades logic is still needed.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.Formats.MSBuild/ProjectBuilder.v4.0.cs
Therzok: That's why I queued everything to build and made commits atomic so I see if it works. From what I've seen, the first thing that it does after creating a ProjectInstance is re-evaluating.

slluis:
Please explain what's this PR about in the description.

Therzok:
Forgot to add do not merge. I wanted to open a PR to queue builds and see how the PR works.

I'll update the PR description when it's ready

Therzok:
https://github.com/dotnet/project-system/blob/master/docs/design-time-builds.md for reference

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
slluis: The documentation you referenced in the comments says that those targets are for C#/VB. It is not a feature that will work for all languages, so using it should be opt-in.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/TargetEvaluationContext.cs
```diff
@@ -33,11 +33,19 @@ public class TargetEvaluationContext: ProjectOperationContext
...
+		}
+
+		public TargetEvaluationContext (bool designTime)
 		{
 			PropertiesToEvaluate = new HashSet<string> ();
 			ItemsToEvaluate = new HashSet<string> ();
 			LogVerbosity = MSBuildProjectService.DefaultMSBuildVerbosity;
+
+			if (designTime) {
+				GlobalProperties.SetValue ("DesignTimeBuild", "true");
```
slluis: I'd like the logic for configuring a design-time build to not be implemented in TargetEvaluationContext. This class is for running any kind of target for any language, and different languages may have different ways of configuring it. This property can be set after creating the TargetEvaluationContext instance [here](https://github.com/mono/monodevelop/pull/4593/files#diff-bf46248a3a265d1dc44bf13c9a8d9e05R1029).

Therzok: :+1:

Therzok:
@slluis please take a look at the first commits in this PR.

slluis:
@Therzok The first two commits look ok.

Therzok:
Closing. RAR design time targets fixed as part of TFM PR.

PR #4594, opened 2018-04-17
Remove assembly folder support

mhutch:
The preferences panel was confusing as it did not work for msbuild based projects, which is pretty much everything that anyone uses.


mhutch:
(failing build is an network timeout, the other builds and tests passed so it's fine)

PR #4595, opened 2018-04-17
Update sln format

mhutch:
From 2012 onwards, the solution format is unchanged, so there's no need to define a 2017 format.

The ToolsVersion defined by the format is irrelevant except to new project creation with the old template system. The build engine will always be the latest MSBuild version.

Change the comment added to new files to match VS2017.

### DIFF
--- a/main/tests/test-projects/solution-custom-data/no-custom-data.sln
+++ b/main/tests/test-projects/solution-custom-data/no-custom-data.sln
@@ -1,6 +1,6 @@
 
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2012
+# Visual Studio 15
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 	EndGlobalSection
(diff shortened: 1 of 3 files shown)


main/tests/test-projects/solution-custom-data/no-custom-data.sln
```diff
@@ -1,6 +1,6 @@
 
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2012
+# Visual Studio 15
 Global
```
mrward: This does not look right to me. It looks like the ProductDescription for the MSBuildFileFormat is using the wrong value if the solution file being generated contains `Visual Studio 15`. It is now `Visual Studio 15` - I think it should be change to `Visual Studio 2012` which is what the MSBuildFileFormatVS12 had originally.

mrward: Visual Studio 2017 uses 'Visual Studio 15' in the solution file so this change is all good. 👍

mhutch: It's just a comment, the file can still be opened in VS2012. It just makes new solutions more similar what VS2017 generates.

PR #4598, opened 2018-04-17
Improve AutoInsertBracketHandler extension filtering

mhutch:
There's now a first pass filter using the file extension and mimetype, avoiding loading assemblies/types until needed. More specific matches are preferred over less specific matches, so there's less need to worry about node ordering.

The extension point now respects conditions on the editor context, and automatically refreshes when the mimetypes or the context change.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/AutoInsertBracketTextEditorExtension.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/AutoInsertBracketTextEditorExtension.cs
@@ -25,35 +25,46 @@
 // THE SOFTWARE.
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using Mono.Addins;
+using MonoDevelop.Ide.Extensions;
 
 namespace MonoDevelop.Ide.Editor.Extension
 {
 	public abstract class AutoInsertBracketHandler
 	{
-		public abstract bool CanHandle (TextEditor editor);
-
+		public virtual bool CanHandle (TextEditor editor) => true;
 		public abstract bool Handle (TextEditor editor, DocumentContext ctx, KeyDescriptor descriptor);
 	}
 
 	class AutoInsertBracketTextEditorExtension : TextEditorExtension
 	{
-		static List<AutoInsertBracketHandler> allHandlers = new List<AutoInsertBracketHandler> ();
+		const string extensionPoint = "/MonoDevelop/Ide/AutoInsertBracketHandler";
+		List<AutoInsertBracketHandler> handlers;
 
-		public AutoInsertBracketTextEditorExtension ()
+		protected override void Initialize ()
 		{
-			AddinManager.AddExtensionNodeHandler ("/MonoDevelop/Ide/AutoInsertBracketHandler", delegate (object sender, ExtensionNodeEventArgs args) {
-				var newHandler = (AutoInsertBracketHandler)args.ExtensionObject;
-				switch (args.Change) {
-				case ExtensionChange.Add:
-					allHandlers.Add (newHandler);
-					break;
-				case ExtensionChange.Remove:
-					allHandlers.Remove (newHandler);
-					break;
-				}
-			});
+			base.Initialize ();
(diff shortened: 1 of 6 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/AutoInsertBracketTextEditorExtension.cs
```diff
@@ -25,48 +25,104 @@
...
-				}
-			});
+			base.Initialize ();
+
+			//whenever the mimetype or the extension context has changed, refilter handlers
+			Editor.MimeTypeChanged += ContextChanged;
+			Editor.ExtensionContext.ExtensionChanged += ContextChanged;
+		}
+
+		void ContextChanged (object sender, EventArgs e)
```
ToddGrun: This must always get fired when a document is opened?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/AutoInsertBracketTextEditorExtension.cs
```diff
@@ -25,48 +25,104 @@
...
+		{
+			handlers = new List<AutoInsertBracketHandler> (
```
ToddGrun: Would it make sense to defer creating this until inside keypress where the AutoInsertMatchingBracket option has been checked?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/AutoInsertBracketTextEditorExtension.cs
mhutch: Nah, initialize should call it. Good catch.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/AutoInsertBracketTextEditorExtension.cs
mhutch: Maybe, yeah. A Lazy<T> or something, and null it out when the context changes.

mhutch: It's not cheap, but not expensive either. I guess it's a toss up as to whether incurring the cost on load (especially when restoring lots of docs on solution open) or on first keystroke is a worse ux.

mhutch: True... I was more wondering about the case where we reopen a solution and restore a bunch of tabs. But I guess that's a more generic problem that we could solve by deferring document initialization until the tab is focused.

ToddGrun: My concern here was that we shouldn't pay the price of to initialize the handlers if you've got the option turned off. You could still do it on initialization if you like if the option is turned on.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/AutoInsertBracketTextEditorExtension.cs
```diff
@@ -25,48 +25,104 @@
...
+			base.Dispose ();
 		}
 
 		public override bool KeyPress (KeyDescriptor descriptor)
 		{
 			var result = base.KeyPress (descriptor);
 
 			if (DefaultSourceEditorOptions.Instance.AutoInsertMatchingBracket && !Editor.IsSomethingSelected) {
-				var handler = allHandlers.FirstOrDefault(h => h.CanHandle (Editor));
+				var handler = handlers.FirstOrDefault (h => h.CanHandle (Editor));
```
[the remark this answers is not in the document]
Therzok: Also, the lambda captures. Avoid that at all costs.

PR #4599, opened 2018-04-18
[TextEditor] Handled line cache removal in redraw methods.

mkrueger:
This way the caller can choose not to remove the line cache - removing
the line cache is always more costly and most of the time not needed.


mkrueger:
Added the change - but still it's not that much of an issue. We're talking here of an internal method that is called 1 time in the whole IDE and that handler already removed the line cache. 

But y it makes sense to move that down. I added an optional parameter for the line cache removal. In the cases RedrawLine is called inside monodevelop the line cache removal doesn't make much sense.

PR #4606, opened 2018-04-18
Fixes VSTS 593638: [Feedback] Visual Studio MAC 7.5 tooltip does not …

mkrueger:
…disappear

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/593638


slluis:
I don't understand the fix. What are tooltip providers supposed to implement in TakeMouseControl? I don't see any real implementation, only the wrapper implementation that just calls the same for the wrapped.

mkrueger:
No there is no implementation for that but it's in theory required for an interactive tooltip. Would maybe be interesting for the interactive debug tooltips. But that's nothing that is used now. 

The editor tooltips where that bug was about are not interactive anymore we had that some time ago. But when the mouse is over an interactive tooltip it should do the same than the text area.

There are 2 options there:

1) Mouse leaves to the text editor -> leave tooltip open.
2) Mouse leaves to another compontent -> close tooltips.

And option 2) is where that method comes into play. However I don't know if that is a good ui decision or not - I'm not sure about it. So atm there is no behavior change beside the editor tooltips where the behavior is clearer.

PR #4615, opened 2018-04-18
Fix VSTS 586112: ignore TextChange events with no Changes

KirillOsenkov:
We are already checking for args.Changes == null, but we're not checking for args.Changes.Count == 0. As a result we were adding UndoOperations where changes list was empty.

The symptom was then in Undo when this code was throwing ArgumentOutOfRange:
((UndoOperation)op).Changes[0].OldPosition;

because it assumed that the Changes list is non-empty.


KirillOsenkov:
I should add that the empty Changes list is possible when committing Completion.

KirillOsenkov:
@monojenkins backport master

PR #4620, opened 2018-04-19
[WIP] Performance test suite

slluis:
This is initial work to create a performance test suite we can use to detect regressions. 

Performance tests are implemented as regular unit tests. There is a new 'PerfTool' tool that loads a test results file and detects regressions by comparing it to a baseline. The tool generates a regular test results file with failures for those tests that regressed in time.

The tests can be unit tests running inside the IDE, or UI tests remotely controlling the IDE. Right now the main focus is on testing performance of the most basic operations: startup time, solution load time, build time and editor performance. We need to measure the performance of the real user experience, so using UI tests is the best option for testing those operations.

The current implementation provides some NUnit extensions that will help writing the tests:

* The `[Benchmark]` attribute can be applied to a unit test, and can be used to specify the time variance tolerance. A tolerance of 0.1 means that the test won't fail if the resulting time us up to 10% slower than baseline. For example: https://github.com/mono/monodevelop/blob/c6e0362890d409d436d2486b7d2b732ec62233ee/main/tests/performance/MonoDevelop.Ide.PerfTests/Test.cs#L39
* The `Benchmark.SetTime()` method can be used to report the time it took to run the test. If not called, the time reported by NUnit will be used. Performance tests usually require some preparation before the real test starts, so `SetTime()` should be used for most tests. For example: https://github.com/mono/monodevelop/blob/c6e0362890d409d436d2486b7d2b732ec62233ee/main/tests/performance/MonoDevelop.Ide.PerfTests/Test.cs#L44

This is an overview of the changes included:

* New 'performance' folder created inside 'test'. This will contain performance test projects.
* New MonoDevelop.PerformanceTesting project. It will contain performance testing logic that we need to use in all performance tests.
* New PerfTool project, which implements the test results processing tool.
* New MonoDevelop.UserInterfaceTesting project, which contains UI testing infrastructure we need to run some performance tests.
* New MonoDevelop.Ide.PerfTests project with a dummy test.

TODO:

* Create at least the startup time test in MonoDevelop.Ide.PerfTests.
* Implement msbuild target that can be invoked in CI to execute the tests.
* Implement msbuild target for updating the test result baseline.
* Configure the Performance Test environment in VSTS to run the tests.


main/tests/MacPlatform.Tests/CarbonTests.cs
```diff
@@ -47,7 +47,7 @@ public void TestProcessName ()
 		public void TestGestalt ()
 		{
 			int majorVersion = Carbon.Gestalt ("sys1");
-			Assert.AreEqual (majorVersion, 10);
+			Assert.AreEqual (majorVersion, 11, "Something is wrong\t");
```
Therzok: Maybe we should be using this: https://developer.apple.com/documentation/foundation/nsprocessinfo/1410906-operatingsystemversion?language=objc

main/tests/MonoDevelop.Core.Tests/MonoDevelop.Core.Tests.csproj
```diff
@@ -7,7 +7,7 @@
     <OutputType>Library</OutputType>
     <RootNamespace>MonoDevelop.Core.Tests</RootNamespace>
     <AssemblyName>MonoDevelop.Core.Tests</AssemblyName>
-    <TestRunnerCommand>..\..\build\bin\mdtool.exe</TestRunnerCommand>
+    <TestRunnerCommand>/Users/lluis/Desktop/Visual Studio.app/Contents/Resources/lib/monodevelop/bin/vstool.exe</TestRunnerCommand>
```
Therzok: Why bundle and not local?

slluis: This change needs to be reverted, I was doing some local testing.

main/tests/ui/MonoDevelop.UserInterfaceTesting/MonoDevelop.UserInterfaceTesting/UITestBase.cs
```diff
@@ -0,0 +1,285 @@
...
+			}
+			else if (!string.IsNullOrEmpty (mdBinPath)){
+				Console.WriteLine ("[UITEST] Installed Visual Studio not found. Falling back to default behavior.");
+				MonoDevelopBinPath = mdBinPath;
+			} else {
+				Console.WriteLine ("[UITEST] Trying to work out where MonoDevelop is installed.");
+
+				var binDir = Path.GetDirectoryName (typeof(AutoTestClientSession).Assembly.Location);
+				Console.WriteLine ($"[UITEST] Starting with {binDir}");
+				var md = Path.Combine (binDir, "MonoDevelop");
```
Therzok: Shouldn't this be relocatable? Don't we want these tests to be able to be run against a bundle?

Therzok: Sorry, by relocatable I meant configurable. Bundles can have different main launcher names.

main/tests/performance/MonoDevelop.Ide.PerfTests/MonoDevelop.Ide.PerfTests.csproj
```diff
@@ -30,14 +30,14 @@
...
   </ItemGroup>
-  <ItemGroup>
-    <Compile Include="TestStartup.cs" />
-  </ItemGroup>
   <ItemGroup>
     <None Include="packages.config" />
     <None Include="Baseline_MonoDevelop.Ide.PerfTests.xml">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Include="TestSolutions\ExampleFormsSolution.zip">
```
Therzok: We shouldn't add binaries to the repo.

main/tests/performance/PerfTool/Program.cs
```diff
@@ -79,12 +87,57 @@ static int GenerateResults (string baseFile, string inputFile, string resultsFil
...
 
+		static int UpdateBaseLine (string baseFile, string inputFile, string outputFile, string resultName)
+		{
+			var baseTestSuite = new TestSuiteResult ();
+			baseTestSuite.Read (baseFile);
+
+			var inputTestSuite = new TestSuiteResult ();
+			inputTestSuite.Read (inputFile);
+
+			inputTestSuite.RegisterPerformanceRegressions (baseTestSuite, out List<TestCase> regressions, out List<TestCase> improvements);
```
slluis: This method can be used for creating the new baseline:
https://github.com/mono/monodevelop/blob/e12b4e83cd600274e651b599696d29be39a1b931/main/tests/performance/PerfTool/TestSuiteResult.cs#L103

So it should be enough to call `inputTestSuite.UpgradeToBaseline (baseTestSuite)` (not tested).

slluis:
@iainx something broke a test in MacPlatform: https://devdiv.visualstudio.com/DevDiv/_release?releaseId=128936&_a=release-contribution-tab-ms.vss-test-web.test-result-in-release-management

Therzok:
OS X 11 is not out yet, not sure why we're asserting major version = 11.

iainx:
@Therzok yeah, dunno why that was changed. Fixed now

PR #4625, opened 2018-04-19
[NuGet] Restore PackageReference project after target framework changed

mrward:
When the target framework version in project options is changed and
the project uses PackageReferences a restore is now run after the
project file is saved. This ensures the PackageReferences work with
the new target framework and re-generates the project.assets.json file
to ensure the build is using the correct references. Without this
changing the target framework to use latest for an Android project
can result in the build failing since the project.assets.json file
is out of sync with the project until a restore was run again.

Fixes VSTS #603274 - PackageReference project not restored after
target framework changed


mrward:
@monojenkins backport d15-7

PR #4634, opened 2018-04-20
Fix VSTS 602848: Project Reload doesn't work

KirillOsenkov:
In this commit https://github.com/mono/monodevelop/commit/88afb8bc3257cde9e500fcbb193615c851eb7e68#diff-a93bf3423c7fe66c6f8e07d45a997a72 we have regressed the Unload and Reload behavior where manually reloading a project would result in errors.

The problem is that we didn't find an oldProject because the ReplacedItem was an UnloadedSolutionItem and not the original project.

This change attempts to recover the actual project that was previously unloaded and passes it to LoadProject, so that LoadProject can properly diff against it.

Also adding a call to Workspace.OnProjectRemoved to remove a Roslyn project from the workspace when the MonoDevelop project was removed.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
@@ -1495,6 +1495,8 @@ namespace MonoDevelop.Ide.TypeSystem
 				MetadataReferenceCache.RemoveReferences (id);
 
 				UnloadMonoProject (project);
+
+				OnProjectRemoved (id);
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


KirillOsenkov:
It is important to get this in, because the Unload/Reload projects is very broken.

There are two scenarios in Unload/Reload: the actual manual project Unload and Reload and the editing of .csproj which triggers a reload. I've manually verified that both scenarios work now.

slluis:
I don't understand why this is necessary. If the project was manually unloaded, then it should have been removed from the workspace, so there will be no oldProject against which to diff.

KirillOsenkov:
That’s what I think should happen ideally. However right now we cannot differentiate between two scenarios: project unload manually (and later reload) and quick reload after .csproj has changed. We want to fully unload in the first case, but not in the second.

The way it currently works is we don’t remove the project from the Workspace on reload (I know it’s wrong). This fix at least makes this kind of sort of work.

Without this fix, manual Unload/Reload is very broken in d15-7. We absolutely need this fix.

KirillOsenkov:
@monojenkins backport master

Issue #4471, opened 2018-04-09
[Linux] Add support for XFCE's xfce4-terminal


aaronfranke:
https://github.com/mono/monodevelop/blob/master/main/src/addins/GnomePlatform/GnomePlatform.cs#L307

Currently, MonoDevelop checks for GNOME, MATE, and KDE terminals, but not XFCE's terminal. On XFCE, MonoDevelop opens the fallback terminal, `xterm`.

I suggest we add `xfce4-terminal` to this list, as it's preferable to `xterm`.

And, perhaps, we could tweak the fallback option. On my system (Xubuntu) there's an environment variable called `COLORTERM` which is set to `xfce4-terminal`. Perhaps we should try to use this?

There's also `x-terminal-emulator` which points to `xfce4-terminal` on my system.

directhex:
The big problem with the terminal support is we need to pass arguments to the terminal (to tell it what to run, set the working directory, and the window title) - and there is no common way to do it. There's no "generic" command-line interface to make a terminal do that, so we have per-terminal code.

aaronfranke:
Well, good point. In terms of `xfce4-terminal` specifically, you can use `-e` to tell it what to execute, `-T` to set title, and `--working-directory=` to set the working directory. Hopefully this helps.

directhex:
I'm not gonna lie: I don't have time to do this. But it's _super easy_ to add. Copy-paste these methods:

https://github.com/mono/monodevelop/blob/master/main/src/addins/GnomePlatform/GnomePlatform.cs#L234-L246
https://github.com/mono/monodevelop/blob/master/main/src/addins/GnomePlatform/GnomePlatform.cs#L266-L268

and insert a copy of https://github.com/mono/monodevelop/blob/master/main/src/addins/GnomePlatform/GnomePlatform.cs#L313-L317 into the same if-else block.

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Author: Aaron Franke
Commit: f30931405f104349b79feb52929599127fec4048 (xamarin/monodevelop)

PR #4645, opened 2018-04-20
[Linux] Add support for XFCE's xfce4-terminal

aaronfranke:
Should fix https://github.com/mono/monodevelop/issues/4471, I think everything's right, please check my code. Thanks @directhex for the help.


directhex:
I'm not sure how useful the VSTS test build would be, since it doesn't CI the GnomePlatform code, but I'm building locally

directhex:
I can't test it working, as master is broken in other ways on Linux right now, but it builds fine...

PR #4647, opened 2018-04-20
[Linux] Temporarily disable persistent storage on linux

Therzok:
Seems like the .so file cannot be loaded by `ld`. Need to investigate further.


Therzok:
Yup

directhex:
This workaround is included in 7.5.0.1062-0xamarin11 debs, to work around crash on project open.

mkrueger:
I use that already locally - works on linux.

Issue #4667, opened 2018-04-25
[TextEditor] SmartTag marker position wrong


mkrueger:
Smart tag markers are painted on top of the line where it should be placed below. 

Seems to be a side effect of the vs.net line data structures are no longer comparable.

> VS bug [#606069](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/606069)

xamarin-release-manager:
Fixed in version 7.6.0.647 (master)

Pull Request #4668 merged by: Mike Krüger
Author: mono
Commit: 942e82e654c68d6dce9c74bbb3a7fa3ccf773f79 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 7.5.0.1233 (d15-7)

Pull Request #4669 merged by: Lluis Sanchez
Author: mono
Commit: 0b2196fa38551a5bc5b486bf95d5b750f51ff6dc (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #4668 merged by: Mike Kr??ger
Author: mono
Commit: 942e82e654c68d6dce9c74bbb3a7fa3ccf773f79 (xamarin/monodevelop)

PR #4668, opened 2018-04-25
Fixes issue #4667 [TextEditor] SmartTag marker position wrong.

mkrueger:
Line data structures are no longer comparable.


mkrueger:
@monojenkins backport d15-7

mkrueger:
It's a regression.

PR #4675, opened 2018-04-25
Improve startup time by optimizing MEF composition

Therzok:
This changeset aims to improve MEF Composition in a few ways:
a) Adding logging for when something on the UI thread queries MEF before it's loaded.
b) Fixes offenders
c) Adds timer infrastructure around MEF composition and refactors code
d) Implements a rudimentary MEF cache which decides whether to invalidate the MEF cache based on assembly timestamp changes
e) Reliability around MEF cache usage, so cache cannot become corrupt
f) Makes the cache logic unit testable

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
@@ -41,6 +41,9 @@ namespace MonoDevelop.Ide.Composition
 {
 	public partial class CompositionManager
 	{
+		/// <summary>
+		/// Used by tests to inject changes into the cache data in memory so it asserts whether caching can be used.
+		/// </summary>
 		internal interface ICachingFaultInjector
 		{
 			void FaultAssemblyInfo (MefControlCacheAssemblyInfo info);
@@ -119,10 +122,10 @@ namespace MonoDevelop.Ide.Composition
 					// Validate that the assemblies match and we have the same time stamps on them.
 					var currentAssemblies = new HashSet<string> (Assemblies.Select (asm => asm.Location));
 					foreach (var assemblyInfo in controlCache.AssemblyInfos) {
+						cachingFaultInjector?.FaultAssemblyInfo (assemblyInfo);
 						if (!currentAssemblies.Contains (assemblyInfo.Location))
 							return false;
 
-						cachingFaultInjector?.FaultAssemblyInfo (assemblyInfo);
 						if (File.GetLastWriteTimeUtc (assemblyInfo.Location) != assemblyInfo.LastWriteTimeUtc)
 							return false;
 					}
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
@@ -1,0 +1,160 @@
+﻿//
+// CompositionManager.Caching.cs
+//
+// Author:
+//       Marius Ungureanu <maungu@microsoft.com>
+//
+// Copyright (c) 2018 
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+using System.Xml.Serialization;
+using Microsoft.VisualStudio.Composition;
+using Mono.Addins;
+using MonoDevelop.Core;
+using MonoDevelop.Core.AddIns;
+using MonoDevelop.Core.Instrumentation;
+
+namespace MonoDevelop.Ide.Composition
+{
+	public partial class CompositionManager
+	{
+		internal class Caching
+		{
+			Task saveTask;
+			public HashSet<Assembly> Assemblies { get; }
+			readonly string mefCacheFile;
+			readonly string mefCacheControlFile;
+
+			public Caching (HashSet<Assembly> assemblies, Func<string, string> getCacheFilePath = null)
+			{
+				Assemblies = assemblies;
+
+				getCacheFilePath = getCacheFilePath ?? (file => Path.Combine (AddinManager.CurrentAddin.PrivateDataPath, file));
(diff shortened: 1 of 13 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
```diff
@@ -0,0 +1,189 @@
...
+						}
+					} catch (Exception ex) {
+						LoggingService.LogError ("Could not deserialize MEF cache control", ex);
+						DeleteFiles ();
+						return false;
+					}
+
+					// Short-circuit on number of assemblies change
+					if (controlCache.AssemblyInfos.Length != Assemblies.Count)
+						return false;
```
alanmcgovern: What if we didn't add/remove, but we just changed. For example we side loaded in an updated designer.

alanmcgovern: Perhaps we should rely solely on timestamps?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
```diff
@@ -0,0 +1,189 @@
...
+			}
+
+			internal Task Write (RuntimeComposition runtimeComposition, CachedComposition cacheManager)
+			{
+				IdeApp.Exiting += IdeApp_Exiting;
+
+				return saveTask = Task.Run (async () => {
+					try {
+						await WriteMefCache (runtimeComposition, cacheManager);
+						IdeApp.Exiting -= IdeApp_Exiting;
```
alanmcgovern: potential multi-threaded access to both `IdeApp.Exiting` and also `saveTask`. Will this be a problem? What about concurrent saves?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
```diff
@@ -106,83 +113,123 @@ internal CompositionManager ()
...
-				// even if the composition has some errors. TODO: re-enable this.
-				//configuration.ThrowOnErrors ();
-			}
+		{
+			var assemblies = ReadAssembliesFromAddins ();
+			var caching = new Caching (assemblies);
+
+			// Try to use cached MEF data
+			if (caching.CanUse ()) {
+				RuntimeComposition = await TryCreateRuntimeCompositionFromCache (caching);
```
alanmcgovern: Do we need to try/cache this, or is this guaranteed to return null instead of throwing an exception if the cache is invalid/broken/missing?

main/tests/Ide.Tests/MonoDevelop.Ide.Composition/CompositionManager.CachingTests.cs
```diff
@@ -0,0 +1,168 @@
...
+			var cacheManager = new CachedComposition ();
+
+			await caching.Write (composition, cacheManager);
+
+			File.WriteAllText (caching.MefCacheFile, "corrupted");
+			Assert.IsNull (await CompositionManager.TryCreateRuntimeCompositionFromCache (caching), "Cache was able to be constructed from corrupted cache");
+
+			Assert.IsFalse (File.Exists (caching.MefCacheFile), "Cache was not deleted on corruption");
+			Assert.IsFalse (File.Exists (caching.MefCacheControlFile), "Cache control was not deleted on corruption");
+		}
```
alanmcgovern: 💯

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
Therzok: See below. It try/catches itself.

main/tests/Ide.Tests/MonoDevelop.Ide.Composition/CompositionManager.CachingTests.cs
```diff
@@ -0,0 +1,168 @@
...
+			var composition = await CompositionManager.CreateRuntimeCompositionFromDiscovery (caching);
+			var cacheManager = new CachedComposition ();
+
+			await caching.Write (composition, cacheManager);
+
+			File.WriteAllText (caching.MefCacheControlFile, "corrupted");
+
+			Assert.AreEqual (false, caching.CanUse ());
+			Assert.IsFalse (File.Exists (caching.MefCacheFile), "Cache was not deleted on corruption");
+			Assert.IsFalse (File.Exists (caching.MefCacheControlFile), "Cache control was not deleted on corruption");
```
alanmcgovern: 💯

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
alanmcgovern: This is nicely covered by tests :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
Therzok: This is only done once. If we race here, we hage cache corruption verification everywhere else

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
Therzok: Yes, but that wouldn't change the code below which verifies stamps. We check if the cache contains the given assembly path (the one we had loaded at composition time), then the timestamp. So it's fine either way. This is a fast path to ensure composition is redone in case stuff is unloaded/loaded.

alanmcgovern: yeah, i misread the method :p Whoop!

Therzok:
I can't reproduce the failures locally

Therzok:
Okay, so it seems like creating the composition with a cache in unit tests causes test failures. Unit tests should either be creating a composition without cache or use a special cache folder.

Therzok:
Should be ready for review pass: @slluis

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
```diff
@@ -0,0 +1,202 @@
+﻿//
+// CompositionManager.Caching.cs
+//
+// Author:
+//       Marius Ungureanu <maungu@microsoft.com>
+//
+// Copyright (c) 2018
```
mkrueger: You should really get your company string fixed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
```diff
@@ -0,0 +1,202 @@
...
+			{
+				try {
+					File.Delete (MefCacheFile);
+				} catch {
+				}
+
+				try {
+					File.Delete (MefCacheControlFile);
+				} catch {
+				}
```
mkrueger: Log exceptions maybe ?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
```diff
@@ -0,0 +1,202 @@
...
+
+			void IdeApp_Exiting (object sender, ExitEventArgs args)
+			{
+				// As of the time this code was written, serializing the cache takes 200ms.
+				// Maybe show a dialog and progress bar here that we're closing after save.
+				// We cannot cancel the save, vs-mef doesn't use the cancellation tokens in the API.
+				saveTask?.Wait ();
+			}
+
+			internal Stream OpenCacheStream () => File.Open (MefCacheFile, FileMode.Open);
```
mkrueger: What's in the exception case here ? File system operation. In general there are many file operations in that file that could use a try…catch we should harden our code for that.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
Therzok: Maybe?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
Therzok: This is guarded in all the callers. Will fix that in a future iteration though.

PR #4683, opened 2018-04-26
Fix 593840: MetadataReferenceCache implementation does IO on the UI thread

DavidKarlas:
Also moved from Timer checking to FileSystemWatcher to observe referenced .dlls changes, added support for removing references and implemented IMetadataService.


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MetadataReferenceCache.cs
```diff
@@ -212,6 +238,51 @@ protected override string GetDocumentationForSymbol (string documentationMemberI
...
+				public event EventHandler UpdatedOnDisk;
+
+				FileSystemWatcher watcher;
+
+				public string FilePath { get; }
+
+				public FileChangeTracker (string filePath)
+				{
+					FilePath = filePath;
+					watcher = new FileSystemWatcher (Path.GetDirectoryName (filePath), Path.GetFileName (filePath));
```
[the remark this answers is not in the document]
mrward: Unfortunately right now the native file system watcher does not seem to generate change events correctly. The Mono/.NET file watcher works as expected. Although in this case things may just work since it has event handlers for all possible events.

mkrueger (dismissed):
The fsw should work most of the time - anyways we need that 100% working in mono. 

Some things in the VS.NET world rely on that. So if it do esn't work in some cases we should fix it in mono.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MetadataReferenceCache.cs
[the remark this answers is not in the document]
mrward: Adding to @Therzok suggestions:
c) Move the FileChangeTracker class to Core. Then it can use the native file system watcher and the IDE can use the FileChangeTracker class from Core.
d) Wait for Mono 5.14.

slluis:
If the watcher problem is fixed in Mono 5.14, let's set it as required dependency. Dismissing approval until this is fixed.

Therzok:
@slluis that means we can't merge it until the next release after 15.8. As mono master has not even been branched yet.

slluis:
@Therzok Mono 5.14 is branched, it's the 2018-04 branch, which goes to stable in 15.8.

Therzok:
:+1: from me then

Therzok:
I'm taking this over while David is on holidays. I've already written support for smarter caching by reusing roslyn code, which uses mmap'd files rather than pure IO (this allows for metadata references to be constructed from pointer offsets rather than reading the full file).

Therzok:
So, I did some research around corefx file watchers in mono, and it seems like 5.14 is returning empty filesystemwatcher handle

Therzok (dismissed):
Requesting changes until I land vs roslyn import and figure out why Mono FSW is broken

Therzok:
See https://github.com/mono/mono/issues/8712

Therzok:
Going green soon

mkrueger (approved):

Therzok:
Found a bug.

Therzok:
Need to rewrite the file watching code to handle most common paths via normal watching (workspace, runtime assemblies) and create watchers for out of tree assembly folders.

Therzok:
Reminder: Keep track of https://github.com/dotnet/roslyn/pull/28162

Therzok:
Reminder for myself tomorrow, to look into providing perf-sensitive APIs around PathTreeNode without probing for paths, and registrations (i.e. metadata references) avoid tree lookups (which is basically `O(logn)`)

Therzok:
Track lazy fsw subscription: https://github.com/dotnet/roslyn/pull/28475/files

Therzok:
https://github.com/dotnet/roslyn/pull/28490/files

Therzok:
This PR is ready for testing now!

The main fixes is:
a) We should not be refreshing assemblies when focusing out/in the IDE, causing minor chokes while typing
b) We should test that when referencing an assembly by dll (not project) in a multi project setup:

```
project A
class MainClass
{
    public static void Main ()
    {
        MyOtherClass.Do();
    }
}
```

```
project B
public class MyOtherClass
{
    public static void Do() {}
}
```

where A references B by assembly (not project). Build solution, change B to be:
```
project B
public class MyOtherClass
{
    //public static void Do() {}
}
```

Build just B, see that A has build errors in the IDE

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -1493,7 +1560,9 @@ protected override void ApplyMetadataReferenceRemoved (ProjectId projectId, Meta
 		{
 			var mdProject = GetMonoProject (projectId) as MonoDevelop.Projects.DotNetProject;
 			var path = GetMetadataPath (metadataReference);
-			if (mdProject == null || path == null)
+			if (path == null)
+				return;
+			if (mdProject == null)
```
mkrueger: I find the old version easier to read :)

Therzok: I'm not sure how this got in. I'll fix this.

main/tests/Ide.Tests/MonoDevelop.Ide.TypeSystem/MonoDevelopMetadataReferenceManager.cs
```diff
@@ -0,0 +1,261 @@
...
+
+			return metadata;
+		}
+
+		void GetStorageInfoFromTemporaryStorage (FileKey moduleFileKey, out ITemporaryStreamStorage storage, out Stream stream, out IntPtr pImage)
+		{
+			int size;
+			using (var copyStream = SerializableBytes.CreateWritableStream ()) {
+				// open a file and let it go as soon as possible
+				using (var fileStream = FileUtilities.OpenRead (moduleFileKey.FullPath)) {
```
mkrueger: Can you put that in try…catch file operations may file for some reason - as a general pattern - just to have some info in the log.

mkrueger (approved):

main/tests/Ide.Tests/MonoDevelop.Ide.TypeSystem/MonoDevelopMetadataReferenceManager.cs
Therzok: This is caught in `MonoDevelopMetadataReference.Snapshot.GetMetadataImpl`. If we catch here, the error will not bubble when it should.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -556,7 +610,21 @@ public DocumentId GetDocumentId (string name)
...
+			}
+
+			public void Dispose ()
+			{
+				if (!workspaceRef.TryGetTarget (out var workspace))
+					return;
+
+				lock (workspace.updatingProjectDataLock) {
+					MonoDevelop.Projects.FileWatcherService.WatchDirectories (this, null).Ignore ();
+					lock (metadataReferences) {
```
mrward: Is this needed? It does not seem that the ProjectData adds itself to the file watcher.

Therzok:
Rebased and all fixed!

mrward (approved):
Tested out b) described in comment - 
https://github.com/mono/monodevelop/pull/4683#issuecomment-406575296 - and all is good with a direct assembly reference and rebuilding the other project.

Therzok:

sgmunn:
@Therzok don't forget to backport to 7.7 as well

PR #4684, opened 2018-04-26
Fixes VSTS 592103: Invalid "Convert anonymous group to method group" …

mkrueger:
…replacement

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/592103


mkrueger:
Seems that some refactoring essentials updates were not in yet.

PR #4685, opened 2018-04-26
[Build] Make builds deterministic

Therzok:
http://blog.paranoidcoding.com/2016/04/05/deterministic-builds-in-roslyn.html


mhutch:
FYI this is set already https://github.com/mono/monodevelop/blob/b8eac0a121160f827e846e10cf3f55ce6e02e688/main/msbuild/MonoDevelop.BeforeCommon.targets#L18

PR #4692, opened 2018-04-27
Fix concurrency issues in RemoteProcessConnection

slluis:
When aborting pending messages, do it outside the message waiters lock,
since that can cause new messages to be posted from the thread that has
the lock, modify the messageWaiters collection, and break everything.

Also make sure no new messages are queued after the connction has
been disposed.

Fixes VSTS bug #580790 - The stable monodevelop version crashes very frequently.
Fixes VSTS bug #605244 - Fatal error on opening solution created in windows on mac


slluis:
I'll try to add some unit tests.

slluis:
Test added.

slluis:
@monojenkins backport d15-7

PR #4694, opened 2018-04-27
Master issue4693

mkrueger:


mkrueger:
The typo change is unrelated but not really worth for an own issue.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/ParsedDocument.cs
```diff
@@ -246,7 +247,7 @@ public static IEnumerable<FoldingRegion> ToFolds (this IReadOnlyList<Comment> co
 					
 					string txt;
 					if (endOffset > startOffset) {
-						txt = "/* " + SubstrEllipsize (comment.Text, startOffset, endOffset - startOffset) + " */";
+						txt = "/* " + GetFirstLine (comment.Text) + " ...";
 					} else {
 						txt = "/* */";
```
slluis: Isn't this specific to c#? there shouldn't be c# specific code in MonoDevelop.Ide.TypeSystem

PR #4698, opened 2018-04-30
[BuildOutput] Use new set of icons and fix issues with solution/proje…

rodrmoya:
…ct icons


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputTreeCellView.cs
```diff
@@ -514,7 +503,7 @@ protected override Size OnGetRequiredSize (SizeConstraint widthConstraint)
 			TextLayout layout = status.GetUnconstrainedLayout ();
 			layout.Width = maxLayoutWidth;
 			var textSize = layout.GetSize ();
-			var height = Math.Max (textSize.Height + 2 * status.LayoutYPadding, ImageSize);
+			var height = Math.Max (textSize.Height + 2 * status.LayoutYPadding, 20);
```
netonjm: ImageSize = 20

https://github.com/mono/monodevelop/blob/558c51b51260fa2360db2529589d2aabe7b87756/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputTreeCellView.cs#L210

PR #4703, opened 2018-05-02
Fix VSTS 609859 - access the correct snapshot on text change

KirillOsenkov:
TextViewMargin.SpanUpdateListener was listening to Document_TextChanging, but calling Layout code that assumed that the currentSnapshot is the new snapshot.

At the time TextChanging is called the currentSnapshot has not been updated yet and still points to the old snapshot. If the new code tries to apply the new changes against the old snapshot they will get corrupt data.

Instead, as a tactical fix for 15.7 introduce a new internal event, TextChangedHighPriority, which is analogous to the same-named event of VS Editor, which is raised immediately after the currentSnapshot has been updated and it is what the layout code expects.

Long term, the right fix would be to move off TextDocument.TextChanging/TextChanged events entirely and listen to the appropriate events of ITextBuffer.


KirillOsenkov:
@monojenkins backport master

slluis:
This change has caused several failures in the unit tests. Unless there is a simple and obvious fix, we'll have to revert.

Issue #4695, opened 2018-04-27
Attempt to access UI from background thread when event handler shown in intellisense window


mrward:
VS Mac 7.5.0.1236

```
Error while adding overload : [CompletionData: Icon=md-newmethod, DisplayText=FileService_FileRemoved, Description=, CompletionText=FileService_FileRemoved, DisplayFlags=None]
System.InvalidOperationException: Operation not supported in background thread
  at MonoDevelop.Core.Runtime.AssertMainThread () [0x00012] in /Users/builder/data/lanes/5965/e52c63d9/source/monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs:443 
  at MonoDevelop.Ide.Editor.TextEditor.get_IndentationTracker () [0x00000] in /Users/builder/data/lanes/5965/e52c63d9/source/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditor.cs:1529 
  at MonoDevelop.CSharp.Completion.Provider.DelegateCompletionProvider.GetInsertText (System.Collections.Immutable.ImmutableDictionary`2[TKey,TValue] properties, System.String& beforeText, System.String& afterText, System.String& newMethod) [0x00021] in /Users/builder/data/lanes/5965/e52c63d9/source/monodevelop/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/CompletionProvider/DelegateCompletionProvider.cs:411 
  at MonoDevelop.CSharp.Completion.Provider.DelegateCompletionProvider.GetTextChangeAsync (Microsoft.CodeAnalysis.Completion.CompletionItem selectedItem, System.Nullable`1[T] ch, System.Threading.CancellationToken cancellationToken) [0x00000] in /Users/builder/data/lanes/5965/e52c63d9/source/monodevelop/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/CompletionProvider/DelegateCompletionProvider.cs:400 
  at Microsoft.CodeAnalysis.Completion.CommonCompletionProvider.GetTextChangeAsync (Microsoft.CodeAnalysis.Document document, Microsoft.CodeAnalysis.Completion.CompletionItem selectedItem, System.Nullable`1[T] ch, System.Threading.CancellationToken cancellationToken) [0x00000] in /_/src/Features/Core/Portable/Completion/CommonCompletionProvider.cs:93 
[...]

PR #4707, opened 2018-05-03
Fixes issue #4695 Attempt to access UI from background thread when

mkrueger:
event handler shown in intellisense window


mkrueger:
@monojenkins backport d15-7

PR #4714, opened 2018-05-03
Fix VSTS 609859.

KirillOsenkov:
All Web editor classifiers (HTML, CSS, Json) expect to only be called on the TextSnapshot which is the CurrentSnapshot of the buffer they're attached to.

Avoid calling into those classifiers in a moment when TextBuffer.CurrentSnapshot has been updated to V2, but the snapshot passed is still V1.

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
@@ -49,11 +49,8 @@ namespace Microsoft.VisualStudio.Platform
 
         public async Task<HighlightedLine> GetHighlightedLineAsync(IDocumentLine line, CancellationToken cancellationToken)
         {
-            //TODO verify that the snapshot line from this.textBuffer is equivalent to the document line converted to a snapshotline.
-            //Possibly take in a TextDataModel as a parameter and verify the buffers are appropriate.
-            //ITextSnapshotLine snapshotLine = (line as Mono.TextEditor.TextDocument.DocumentLineFromTextSnapshotLine)?.Line;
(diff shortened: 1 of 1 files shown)


KirillOsenkov:
We should merge either this PR or #4717, but not both. We should evaluate both approaches and choose the one that has less risk and better quality.

mkrueger:
Hard to tell which one to take. 

Why not both  ? :)

GLjackyvaswani:
We have verified the both the bugs in below builds by performing following scenario.
As both the scenarios are being satisfied,We are marking this bot the bugs as QA Approved.

Build Taken From:
https://devdiv.visualstudio.com/DevDiv/_build/index?buildId=1660687&tab=artifacts&_a=summary

Build Gist:
https://gist.github.com/tanmay-bhattacharjee/e124be290145116da6c3c112d147fb92

Steps Performed:
Scenario #1:
1. In MonoDevelop create a new HTML file
2. Ctrl+A
3. Delete

Screencast:  https://screencast.com/t/64wfc1HmT
---------------------------------------------------
Scenario #2:
1. Create a new C# class library.
2. Create a new project.json file.
3. Paste in this incomplete json into the project.json file.

"dependencies": {
"Flurl": "2.7.1",
"Flurl.Http": "2.3.1",
"MvvmCross": "5.7.0",
"MvvmCross.Binding": "5.7.0",
"MvvmCross.Core": "5.7.0",
"MvvmCross.Platform": "5.7.0",
"MvvmCross.StarterPack": "5.7.0",
"Newtonsoft.Json": "11.0.2"
},

"frameworks": {
".NETPortable,Version=v4.5,Profile=Profile111": {}
}
4. Try to type
Screenast: https://www.screencast.com/t/pch2jgVnuq

Kindly let us know if we have missed something while the validation of bug, can re-construct testing again if required.

PR #4717, opened 2018-05-03
Fix VSTS 609859.

KirillOsenkov:
Todd came up with another approach to fixing 690859. Basically we need to make sure that the line highlight recalculation happens outside of the core Edit transaction. To achieve this we move it to the TextBuffer event handlers. This way this code doesn't run in the intermittent state when the snapshot doesn't match TextBuffer.CurrentSnapshot.

We're also filling the cache of lines with all visible lines uniformly.

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.SpanUpdateListener.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.SpanUpdateListener.cs
@@ -47,6 +47,7 @@ using System.Collections.Immutable;
 using System.Threading;
 using MonoDevelop.Ide;
 using System.Threading.Tasks;
+using Microsoft.VisualStudio.Text;
 
 namespace Mono.TextEditor
 {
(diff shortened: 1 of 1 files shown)


KirillOsenkov:
We should merge either this PR or https://github.com/mono/monodevelop/pull/4714, but not both. We should evaluate both approaches and choose the one that has less risk and better quality.

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.SpanUpdateListener.cs
```diff
@@ -100,7 +102,7 @@ void Document_TextChanged (object sender, TextChangeEventArgs e)
 
 			bool UpdateLineHighlight (int lineNumber, HighlightedLine oldLine, HighlightedLine newLine)
 			{
-				if (oldLine != null && ShouldUpdateSpan (oldLine, newLine)) {
+				if (oldLine == null || ShouldUpdateSpan (oldLine, newLine)) {
```
mkrueger: if oldLine == null redraw shouldn't be needed because that line wasn't drawn so it can't be invalid. But I don't think that'll ever happen on text buffer change therefore it's not an issue.

mkrueger:
So I am correct that TextChanging/TextChanged events is broken for webtools ?

At least they're not balanced - I think this code isn't the only case where this is needed.

slluis:
Some tests are failing.

KirillOsenkov:
Yes, the two failing tests indicate a potential real problem with this approach. We'll find a more comprehensive fix for 15.8.

Issue #4709, opened 2018-05-03
Right-click -> FindReferences when used in an interface, returns too many useless results


knocte:
Steps to reproduce (testcase which definitely leads to the bug, not sure if it can be simpler):
1. Clone https://github.com/metacosa/nbitcoin
2. Open it in VisualStudioForMac.
3. Go to Transaction.fs file, to the TxOut constructor that takes arguments `(Money value, IDestination destination)`.
4. Right-click on the `IDestination` word, -> Find References.

Expected results:
It should show just results where this interface is used, e.g. as a type declaration, or the classes that implement the interface, etc.

Current results:
It shows too many results, e.g. when all types that inherit from IDestination are used, such as `txbuilder.Send(new Key().ScriptPubKey, Money.Coins(0.4m));` (Key implements IDestination).

Version information:
```
Visual Studio Community 2017 for Mac
Version 7.4.3 (build 10)
Installation UUID: fd06ff65-69de-47ae-8b99-ddf257032b7c
Runtime:
	Mono 5.8.1.0 (2017-10/6bf3922f3fd) (64-bit)
	GTK+ 2.24.23 (Raleigh theme)

	Package version: 508010000

NuGet
Version: 4.3.1.4445

.NET Core
Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	2.0.5
	2.0.0
SDK: /usr/local/share/dotnet/sdk/2.1.4/Sdks
SDK Versions:
	2.1.4
	2.0.0
MSBuild SDKs: /Library/Frameworks/Mono.framework/Versions/5.8.1/lib/mono/msbuild/15.0/bin/Sdks

Xamarin.Profiler
Version: 1.6.1
Location: /Applications/Xamarin Profiler.app/Contents/MacOS/Xamarin Profiler

Apple Developer Tools
Xcode 9.3 (14154)
Build 9E145

Xamarin.Mac
Version: 4.2.1.29 (Visual Studio Community)

Xamarin.iOS
Version: 11.9.1.24 (Visual Studio Community)
Hash: f62de472
Branch: xcode9.3
Build date: 2018-03-29 19:30:53-0400

Xamarin.Android
Version: 8.2.0.16 (Visual Studio Community)
Android SDK: /Users/andres/Library/Developer/Xamarin/android-sdk-macosx
	Supported Android versions:
		4.4 (API level 19)
		7.1 (API level 25)

SDK Tools Version: 26.1.1
[...]

PR #4732, opened 2018-05-07
Fixes issue #4709 Right-click -> FindReferences when used in an

mkrueger:
interface, returns too many useless results For some reason the find &
find all refs calls were swapped. Find all should find the whole
familiy where find just should find the specific symbol and it's
usages.


Therzok:
Can we unit test this?

mkrueger:
It's just a call to the roslyn API - I would prefer getting the roslyn commands working directly rather than testing that old code that's going away soon. We had unit tests for the old find system but testing the roslyn API is more or less pointless IMO.

Therzok:
@mkrueger we're not testing the roslyn API, we're testing that we're integrating it right.

PR #4733, opened 2018-05-07
Fixes VSTS 610190: Collapsed Regions Auto Expand

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/610190

GetFoldingContaining is end inclusive which doesn't work here. the
auto formatter may make a text change right after the collapsed fold
which unfolds the region.


slluis:
@monojenkins backport d15-7

Issue #4734, opened 2018-05-07
[Progress Monitoring] No way to use the synchronization context to write to the log


Therzok:
`ProgressMonitor.Log` should be abstracted with the containing SynchronizationContext.

Currently, users of ProgressMonitor can either:
a) write the log on the wrong synchronization context by doing `monitor.Log.WriteLine`
b) write the log on the wrong synchronization context because every progress monitor can set their own.

Consider abstracting the log into a ProgressMonitorLog that yields the calls to the right synchronization context - i.e. GUI for dialogs

> VS bug [#612113](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/612113)

slluis:
Progress monitors should write log through the sync context they have assigned, like they do for other methods. Not sure what you mean by ProgressMonitorLog.

Therzok:
Ah, so you mean `ProgressLogWriter` should be also keeping a hold of the SynchronizationContext and use that when doing actual writing?

Therzok:
Because, `monitor.Log` is accessed directly in many places and `LogTextWriter` does not use the synchronization context

slluis:
The progress monitor subscribes events from `monitor.Log`, so it can handle those write events in the correct context. LogTextWriter doesn't need to be aware of it.

Therzok:
Fixed via https://github.com/mono/monodevelop/pull/4735

xamarin-release-manager:
> [sync] [VS-17] Comment by **xamarinc**



Fixed in version 8.6.0.802 (master)



Author: therzok  
 Commit: 0ba4c961ddb304de605857b76417c9fa6aa5bdd9 (xamarin/monodevelop)

PR #4735, opened 2018-05-07
[Git] Fix monitor writes not being done on the UI thread

Therzok:
Fixes VSTS #611145 - Git apply stash - Gtk warning - operation done on background thread


slluis:
In general progress monitors should be thread safe.

Therzok:
Pushed another attempt, that also fixes #4734

Therzok:
Hmmm, I can unit test this, give me a bit

Therzok:
Added unit tests and also made overrides for all the methods that `TextWriter` routes into:
https://github.com/mono/mono/blob/master/mcs/class/referencesource/mscorlib/system/io/textwriter.cs#L49-L94

The current problem is that `Dispose` is not called on the right context.

`public virtual void Dispose` does not allow for this.

Consider breaking API when possible to make it `protected virtual void OnDisposed`, so we can call it on the right sync context.

Therzok:
Added unit tests. @slluis please take another look

Issue #4716, opened 2018-05-03
TextChanged line highlighting logic needs to be revised


KirillOsenkov:
Here:
http://source.monodevelop.com/#MonoDevelop.SourceEditor/Mono.TextEditor/Gui/TextViewMargin.SpanUpdateListener.cs,be4fd928cb39a11f

It seems that for every text change we get the Line that contains the starting point of the change, and add the line to the collection.

If two changes are on the same line it seems that the line will be formatted and added twice.

Also it seems that if a change spans multiple lines, then only the first line will be laid out and added to the list.

> VS bug [#611065](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/611065)

PR #4738, opened 2018-05-07
Fixes issue #4716 TextChanged line highlighting logic needs to be

mkrueger:
revised

The span update logic is in fact very specific to the roslyn
classification highlighting breaking the webtools highlighting tagger.
So moving that to the roslyn highlighting will fix any interference
issues. And fixed some issues with the span update listener Kirill
found.


KirillOsenkov:
@toddgrun @davidkarlas

mkrueger:
We should take #4740

Issue #3770, opened 2018-02-04
GetSemanticModelAsync should not be called from UI Thread


DavidKarlas:
https://github.com/mono/monodevelop/blob/ddb4355/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/RoslynClassificationHighlighting.cs#L137

Found this issue via UIThread hangs monitor...

> VS bug #569967

mkrueger:
It's not called from ui thread - it's called async with await. I'm not sure if the hangs monitor is the issue here.

DavidKarlas:
It is called from UI thread from gtk `OnExposeEvent` method:
<img width="1664" alt="image" src="https://user-images.githubusercontent.com/774791/36531554-992fa928-17be-11e8-8898-338ad621282d.png">

Even if it's called with `await`, there is no guarantee(or even promise) it won't execute on same thread.

mkrueger:
It's executed on the same thread when the semantic model is up2date and cached.
Creating always a task would be too much overhead. I'll discuss that with the roslyn guys.

mkrueger:
We call it always on UI thread like that - if that it would be a problem we would block all the time. So I doubt it's an issue.

mkrueger:
ATM I don't see a way to fix that - and no issue here. If you think you can improve that - try it out. 
IMO it would just cause redraw flickers and not improve the UI in any way. That call can't block the UI thread and the semantic info is needed for the classifications since the roslyn classifiers are not 2 step like our old system was :(.

DavidKarlas:
This was fixed with https://github.com/mono/monodevelop/pull/4740

PR #4740, opened 2018-05-07
Tag based classification for C#

DavidKarlas:
This PR introduces 2 big changes:
 * Switches from MonoDevelopSourceTextContainer to `TextBuffer.AsTextContainer()` which enables some roslyn functionalities(Tagger among others)
 * Switches to VSEditor Tagger classification instead direct Roslyn one which brings huge performance improvement when typing

I will need to figure out how to optimize rendering logic, to not redraw lines if nothing changed, like VS2017 does...


DavidKarlas:
Forgot to mention this fixes https://github.com/mono/monodevelop/issues/3770

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/MdTextViewLineCollection.cs
```diff
@@ -84,7 +89,21 @@ internal void RemoveLinesAfter (int lineNumber)
 
 		public ITextViewLine LastVisibleLine => this.LastOrDefault ();
 
-		public SnapshotSpan FormattedSpan => new SnapshotSpan (this [0].Start, this.Last ().EndIncludingLineBreak);
+		public SnapshotSpan FormattedSpan {
+			get {
+				if (this.Count == 0)
+					return new SnapshotSpan (textEditor.TextSnapshot, 0, 0);
+				var start = this [0].Start;
+				var end = this.Last ().EndIncludingLineBreak;
```
Therzok: Please avoid linq. It usually does not get inlined.

mkrueger: If you never use linq you don't do anything wrong.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -824,12 +824,12 @@ TextDocument InternalInformDocumentOpen (DocumentId documentId, TextEditor edito
 			if (project == null)
 				return null;
 			TextDocument document = project.GetDocument (documentId) ?? project.GetAdditionalDocument (documentId);
-			if (document == null || openDocuments.Any (d => d.Id == documentId)) {
+			if (document == null || openDocuments.ContainsKey(documentId)) {
 				return document;
 			}
-			var monoDevelopSourceTextContainer = new MonoDevelopSourceTextContainer (this, documentId, editor);
+			var monoDevelopSourceTextContainer = editor.TextView.TextBuffer.AsTextContainer ();
```
mkrueger: Now we're starting to benefit from the vs.net editor work - nice :)

main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
```diff
@@ -24,276 +24,364 @@
...
+				//Which orders different classifications by priority...
+				//This is needed so Semantical classification for Interface or Class is greater then Identifier from Syntactical classifeer
+				if (scope != null && (baseType.Classification == PredefinedClassificationTypeNames.Identifier || baseType.Classification == PredefinedClassificationTypeNames.Keyword))
+					continue;
+				if (classificationMap != null && classificationMap.TryGetValue (baseType.Classification, out var mappedScope))
+					scope = mappedScope;
+				var styleName = GetStyleNameFromClassificationName (baseType.Classification);
+				if (styleName == null)
+					continue;
+				scope = new ScopeStack (styleName);
```
mkrueger: This doesn't look right the classification map already contains the scope and then you're using GetStyleNameFromClassificationName to produce a new scope ? 

GetStyleNameFromClassificationName has to go and the missing entries need to be added to the classificationMap.

If an entry is missing add the default case from GetStyleNameFromClassificationName to the ClassificationMap. So the ScopeStack objects get re-used instead of re-created every time.

main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
```diff
@@ -24,276 +24,364 @@
...
+			}
+			return classificationTypeToScopeCache [classificationType] = scope ?? new ScopeStack (EditorThemeColors.Foreground);
```
mkrueger: Do not create a new ScopeStack (EditorThemeColors.Foreground) more than once.

main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
```diff
@@ -24,276 +24,364 @@
...
+		}
+
+		Dictionary<IClassificationType, ScopeStack> classificationTypeToScopeCache = new Dictionary<IClassificationType, ScopeStack> ();
+
+		private ScopeStack GetScopeNameFromClassificationType (IClassificationType classificationType)
+		{
+			if (classificationTypeToScopeCache.TryGetValue (classificationType, out var cachedScope))
+				return cachedScope;
+			ScopeStack scope = null;
+			foreach (var baseType in classificationType.BaseTypes.Concat (classificationType)) {
```
mkrueger: Wouldn't it be more performant to start with the classificationType and then check the base types ?

main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
```diff
@@ -24,276 +24,364 @@
...
+					}
+				}
+			}
+		}
+
+		Dictionary<IClassificationType, ScopeStack> classificationTypeToScopeCache = new Dictionary<IClassificationType, ScopeStack> ();
+
+		private ScopeStack GetScopeNameFromClassificationType (IClassificationType classificationType)
+		{
+			if (classificationTypeToScopeCache.TryGetValue (classificationType, out var cachedScope))
```
mkrueger: classificationTypeToScopeCache is an immutable dictionary. Let's say you've a vb.net and c# file open.
Is the IClassificationType keyword the same for vb.net and c#  ? If so it would return a wrong value for one of these files.

DavidKarlas:
I updated PR with `However this breaks span updates. /* & @". VS.NET taggers need to do that somehow.` fixed.

main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
DavidKarlas: This object and cache is created per ITextView so I don't expect this to be an issue.

main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
DavidKarlas: About performance, we are caching this so it's not really problem, plus we have to check all anyway.
Base types are in alphabetic order, priorty is handled differently in VSEditor, see https://github.com/dotnet/roslyn/blob/88d1bd1/src/EditorFeatures/Core.Wpf/Classification/ClassificationTypeFormatDefinitions.cs
I'm now handling BaseTypes of BaseTypes...

Logging non-found classifcaiton is imo not an option because some classifications are created on the fly when merging multiple classifications...

main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
DavidKarlas: Fixed.

main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
DavidKarlas: I fixed this now by returning of mapping one is found, also, all of this is cached so it shouldn't be a problem.

slluis:
Tell me more, what is that huge performance improvement?

DavidKarlas:
Often times when I see laggy typing I profile it and it turns out to be problem like seen at #3770.
Problem is that we call `await document.GetSemanticModelAsync ()` on UI thread every time user types new character(since we want to render), most times this is incrementally parsed and most things cached hence it's fast to return, but sometimes it can take few hundred milliseconds or even seconds...
https://github.com/mono/monodevelop/blob/b315922e16628a3c6c35ea3ee0a4fe04f674e9ec/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/RoslynClassificationHighlighting.cs#L199

The way new logic work is same as in VS Windows, it uses SnapshotSpan.TranslateTo to translate older classification to latest textbuffer so it's "instantly" renders best approximation and later calls `classifier.ClassificationChanged` when newer classification is available instead of blocking UI thread to figure out all classifications(like we did before)...

mkrueger:
Still missing the ISemanticChangeNotificationService usage.

However testing semantic updates seem to work for some reason. I think that the code analysis may do the updates since it takes a while for highlighting to correct. But it works so I approve.

One thing: remove the tests for the span updater I already did that. And we should consider this for d15-7 because it's a better fix to my span updater move. The tag based highlighter needs to work anyways.

DavidKarlas:
`ISemanticChangeNotificationService` is used here: source.roslyn.io/#Microsoft.CodeAnalysis.EditorFeatures/Implementation/Classification/SemanticClassificationViewTaggerProvider.cs and that is notifying us via `classifier.ClassificationChanged`.

Tnx for heads ups about Unit tests, I noticed locally few minutes ago that it doesn't compile... I removed them now.

PR #4741, opened 2018-05-07
Enable MEF caching and refactorings

Therzok:
This PR is groundwork to enable MEF caching and refactors unit test logic for reusability.

The unit tests (code issue/action) now pass due to the tests being placed on the C# level.

If we want to unit test a C# MEF component, we should be testing in CSharpBinding.Tests (or any addin depending on it)

Previously they used to live on MonoDevelop.Refactoring layer.

Also, refactor the test editor extension code for reusability and migrate some users to that. Makes for less code and more maintainable test scaling.


Therzok:
Investigating current failures. Todo comments is still on IDE level, not C#, and I made a typo in the TestEditorExtensionData code which causes infinite looping. Will fix.

Therzok:
Ready for review now.

Issue #4746, opened 2018-05-07
Multiple "var to explicit type" actions


mhutch:
The fix menu shows multiple actions/fixes to convert var to explicit types.

<img width="639" alt="screen shot 2018-05-07 at 2 42 07 pm" src="https://user-images.githubusercontent.com/183285/39726934-9596bd66-5205-11e8-97e9-e89077c41083.png">

> VS bug [#612396](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/612396)

Therzok:
The first one is a quick fix, the second one is a refactoring.

I'm not sure why refactor is not grouped.

Therzok:
And they're both from roslyn:
https://github.com/dotnet/roslyn/blob/master/src/Features/CSharp/Portable/CodeRefactorings/UseExplicitOrImplicitType/UseExplicitTypeCodeRefactoringProvider.cs#L17
https://github.com/dotnet/roslyn/blob/master/src/Features/CSharp/Portable/TypeStyle/UseExplicitTypeCodeFixProvider.cs

mhutch:
Nope, "To explicit type" is [RefactoringEssentials]( https://github.com/icsharpcode/RefactoringEssentials/blob/master/RefactoringEssentials/CSharp/CodeRefactorings/Synced/ReplaceVarWithExplicitTypeCodeRefactoringProvider.cs#L58)

Roslyn is "Use explicit type".

The screenshot above is a fix, not a refactoring. Roslyn appears to disable the refactoring when the analyzer fix would make it redundant.

You can see here with the analyzer disabled, both the RE and Roslyn refactorings show up:

<img width="381" alt="screen shot 2018-05-07 at 3 06 02 pm" src="https://user-images.githubusercontent.com/183285/39727606-44dd1566-5208-11e8-889f-9b89f09165c6.png">

Therzok:
Okay then, we can remove the dupe in RE.

xamarin-release-manager:
Fixed in version 7.6.0.730 (master)

Pull Request #4750 merged by: Mike Krüger
Author: mono
Commit: d404b39167435f903225f22b452a1828c25b0a3b (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #4750 merged by: Mike Kr??ger
Author: mono
Commit: d404b39167435f903225f22b452a1828c25b0a3b (xamarin/monodevelop)

PR #4750, opened 2018-05-08
Fixes issue #4746 Multiple "var to explicit type" actions

mkrueger:
Explict -> var was a dupe as well.


Therzok:
Build failed. Something else needs removing.

Issue #4743, opened 2018-05-07
Squiggles do not scale with editor


mhutch:
As you can see in this example, the suggestion squiggle under the word _var_ is almost invisible.

<img width="427" alt="screen shot 2018-05-07 at 2 28 26 pm" src="https://user-images.githubusercontent.com/183285/39726158-0240f7b8-5203-11e8-8983-19073a5be801.png">

> VS bug [#612541](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/612541)

xamarin-release-manager:
Fixed in version 7.6.0.772 (master)

Pull Request #4753 merged by: Lluis Sanchez
Author: mono
Commit: d5c9ded3bc2b40d2dfef719e637249e45321b479 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #4753 merged by: Lluis Sanchez
Author: mono
Commit: d5c9ded3bc2b40d2dfef719e637249e45321b479 (xamarin/monodevelop)

PR #4753, opened 2018-05-08
Fixes issue #4743 Squiggles do not scale with editor

mkrueger:


slluis:
The squiggle grows too much when zooming in, and disappears when zooming out. Did anyone test this?

kapture 2018-05-10 at 17 43 26

mkrueger:
These were not part  of the patch or bug report. I just searched for the ones that didn't take the zoom level.
And the wave lines do.

PR #4758, opened 2018-05-09
[Ide] Fix duplicate references added to project

mrward:
When a project referenced another project and the referenced assembly
changed a Reference MSBuild item would be added to the project. This
would result in unwanted Reference MSBuild items being added since
the ProjectReference was all that was needed.

Changed the MetadataReferenceCache so it no longer calls TryApplyChanges
on the MonoDevelopWorkspace which would result in the project being
modified. Now OnMetadataReferenceRemoved and OnMetadataReferenceAdded
are called which just updates the metadata so code completion is
up to date based on the latest assemblies.

Fixes VSTS #613601 - Duplicate references added on rebuilding project

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MetadataReferenceCache.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MetadataReferenceCache.cs
@@ -133,13 +133,8 @@ namespace MonoDevelop.Ide.TypeSystem
 						foreach (var solution in IdeApp.Workspace.GetAllSolutions ()) {
 							var workspace = TypeSystemService.GetWorkspace (solution);
 							foreach (var projId in InUseBy) {
-								while (true) {
-									var project = workspace.CurrentSolution.GetProject (projId);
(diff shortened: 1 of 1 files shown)


mrward:
I am fine with #4683 replacing this. I made the fix on master first and tested it there instead of d15-7.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MetadataReferenceCache.cs
```diff
@@ -148,13 +143,8 @@ public void CheckForChange ()
...
 							var workspace = TypeSystemService.GetWorkspace (solution);
 							foreach (var projId in InUseBy) {
-								while (true) {
-									var project = workspace.CurrentSolution.GetProject (projId);
-									if (project == null)
-										break;
-									if (workspace.TryApplyChanges (project.AddMetadataReference (Reference).Solution))
-										break;
-								}
+								if (workspace.CurrentSolution.ContainsProject (projId))
```
Therzok: Invert this if

PR #4760, opened 2018-05-09
Fixes convert tabs to spaces option being ignored

slluis:
Fixes VSTS bug #613507
Fixes VSTS bug #606153


slluis:
@monojenkins backport d15-7

PR #4777, opened 2018-05-11
[Ide] Improve DnD handling

### LINKED ISSUE #4403: Drag and Drop into welcome page does not work anymore
The welcome page used to support drag and drop of files into the welcome page, and that would load the files in the IDE.

My test case was a `.dll` file.

> VS bug [#593849](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/593849)

sevoku:
This fixes several DnD handling issues:

* Bring back Workbench (incl. welcome screen) drop support (fixes #4403, partially reverts 67b943d6d58daf7ab9a5b7b5931804e1487f6bd4)
* Support opening multiple solutions
* Allow dropping solution folders (will open first solution file in the dropped folder)
* Use only internal viewers to open dropped files
* Don't load projects dropped into the documents notebook

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.DockNotebook/DockNotebook.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.DockNotebook/DockNotebook.cs
@@ -254,7 +254,7 @@ namespace MonoDevelop.Components.DockNotebook
 						if (Services.ProjectService.IsWorkspaceItemFile (filePath))
 							IdeApp.Workspace.OpenWorkspaceItem(filePath);
 						else
-							IdeApp.Workbench.OpenDocument (filePath, null, -1, -1, MonoDevelop.Ide.Gui.OpenDocumentOptions.Default, null, null, this);
+							IdeApp.Workbench.OpenDocument (filePath, null, -1, -1, MonoDevelop.Ide.Gui.OpenDocumentOptions.DefaultInternal, null, null, this);
 					} catch (Exception e) {
 						MonoDevelop.Core.LoggingService.LogError ("unable to open file {0} exception was :\n{1}", file, e.ToString());
 					}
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.DockNotebook/DockNotebook.cs
```diff
@@ -249,12 +249,10 @@ void OnDragDataReceived (object o, Gtk.DragDataReceivedArgs args)
 				string file = individualFile.Trim ();
 				if (file.StartsWith ("file://")) {
 					var filePath = new FilePath (file);
-
+					if (filePath.IsNullOrEmpty || filePath.IsDirectory) // skip directories and empty paths
+						continue;
 					try {
-						if (Services.ProjectService.IsWorkspaceItemFile (filePath))
```
slluis: I'd like to keep the old behavior here. If a user drops a solution file to the IDE, it is very likely the intention to open the solution, not edit the solution file. On the other hand, if the project file is being dropped, it is very likely the intention is to edit the project file (notice that `IsWorkspaceItemFile` will return true for solutions, but false for projects).

sevoku: It's not IDE, it's only dropping into the document notebook tabbar, IDE (toolbar and anywhere outside of the editor, + welcome screen) will just open the solution (not in the editor). If I drop something into the tabbar, which is only this area:
grafik
I expect it to be opened in the editor.

sevoku: Actually many other areas don't treat solution files correctly at all. For example the solution pad should open a dropped solution, but it doesn't anything with it. I'll fix it...

slluis: kapture 2018-05-14 at 12 28 11

I want to keep this behavior.

sevoku: Yeah, got it, I missed the case when no document is open. Fixed with: 5eb56ba90b3dbdccd6a836fd569fc60fe32d58cd

slluis:
Approved by QA.

PR #4782, opened 2018-05-12
Fix initial output message in F# interactive

nosami:
Fixes VSTS #615186

We now send the filename of the active document along with the user
input on each invocation.

This also means that __SOURCE_FILE__ works now too.


nosami:
Build is failing, but no clue as to why

image

PR #4785, opened 2018-05-13
[WebReferences] Fix background thread used when adding file to project

mrward:
Adding a .NET 2.0 Web Services web reference would add a file to
the project using a background thread. For ASP.NET Core MVC projects
that use a .cshtml file this would result in an error message:

   The web reference could not be added
   'DotNetProject_Modified' must be called on the foreground thread.

Fixes VSTS #615584 - Cannot add web reference to ASP.NET Core MVC
project


mrward:
@monojenkins backport d15-7

mrward:
Note that .NET Core 2.0 projects do not support the System.Web.Services namespace used by the .NET 2.0 web reference. WCF service references seem to compile if you install say the System.ServiceModel.Http NuGet package into your .NET Core 2.0 project. So presumably the web references dialog should enable WCF instead of .NET 2.0 web references for .NET Core projects.

PR #4787, opened 2018-05-14
Fix 528282: Immediate window completion only contains locals

DavidKarlas:
This commit adds new `MonoDevelop.Debugger.IDebuggerCompletionProvider` interface that can be implemented on EditorExtensions which allows each language extension to provide enhanced code completion items for debugger.
So users of debugger code completion (Immidiate pad, Expresson evaluator, Watchpad...) now use new `DebuggingService.GetCompletionDataAsync` method which first attempts to use `IDebuggerCompletionProvider` implementation and if that fails, falls back to old way of using debugger locals.
C# implementation is mimicking what VS does, which is, add `;` around expression at end of where execution stopped(also handles special cases like when we stop at end of method on `}` to place it before `}`) and calls Roslyn code completion with updated solution.

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/CSharpCompletionTextEditorExtension.IDebuggerCompletionProvider.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/CSharpCompletionTextEditorExtension.IDebuggerCompletionProvider.cs
@@ -1,0 +1,198 @@
+﻿//
+// CSharpCompletionTextEditorExtension.IDebuggerCompletionProvider.cs
+//
+// Author:
+//       David Karlaš <david.karlas@microsoft.com>
+//
+// Copyright (c) 2018 Microsoft Corp
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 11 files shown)


DavidKarlas:
Added more details to PR/commit

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/CSharpCompletionTextEditorExtension.IDebuggerCompletionProvider.cs
```diff
@@ -0,0 +1,198 @@
...
+					result |= Mono.Debugging.Client.ObjectValueFlags.Namespace;
+					break;
+				case CompletionTags.ExtensionMethod:
+				case CompletionTags.Method:
+					result |= Mono.Debugging.Client.ObjectValueFlags.Method;
+					break;
+				case CompletionTags.Parameter:
+					result |= Mono.Debugging.Client.ObjectValueFlags.Parameter;
+					break;
+				case CompletionTags.Property:
```
Therzok: maybe add a default case or a unit test for this asserting we use all values?

PR #4793, opened 2018-05-14
Fix classification in Razor and compat check

KirillOsenkov:


KirillOsenkov:
Only merge this if the master of md-addins is still at bc4153d5a153696f466797033ec3113fed004238

PR #4795, opened 2018-05-15
Fix 591905: VSTest adapter delays solution load

DavidKarlas:
This change moves rebuilding unit tests from UI thread to Threadpool and handles cancelation if newer requests comes in.

### DIFF
--- a/main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestService.cs
+++ b/main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestService.cs
@@ -350,22 +350,38 @@ namespace MonoDevelop.UnitTesting
 			return false;
 		}
 
-		static void RebuildTests ()
+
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestService.cs
```diff
@@ -350,22 +350,37 @@ static bool IsSolutionGroupPresent (Solution sol, IEnumerable<UnitTest> tests)
...
+						UnitTest t = BuildTest (it);
+						if (t != null)
+							list.Add (t);
+					}
+				}, token);
+				if (token.IsCancellationRequested)
+					return;
+				rootTests = list.ToArray ();
+				NotifyTestSuiteChanged ();
+			} catch (Exception ex) {
```
Therzok: Maybe a catch for a `OperationCancelledException`?

PR #4797, opened 2018-05-15
Fix 593833: Showing a document does MEF work on the UI thread

DavidKarlas:
`HostDiagnosticAnalyzerPackage` was fully created when `MonoDevelopWorkspaceDiagnosticAnalyzerProviderService` was requested which is requested from UI thread by `ICodeFixService` which is requested by `CodeActionEditorExtension`. To fix this I moved creation to Threadpool and changed `Options` field that was requested from multiple sources inside MonoDevelop to `GetOptionsAsync` so it returns options only once they are loaded. Notice that `GetHostDiagnosticAnalyzerPackages` is called from threadpool, so even if creating `HostDiagnosticAnalyzerPackage` takes long time it won't block UI.

Fixes VSTS 593833.

### DIFF
--- a/main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/AnalysisCommands.cs
+++ b/main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/AnalysisCommands.cs
@@ -34,6 +34,7 @@ using MonoDevelop.CodeIssues;
 using MonoDevelop.Components.Commands;
 using MonoDevelop.Ide.Gui.Dialogs;
 using Microsoft.CodeAnalysis.Diagnostics;
+using System.Threading.Tasks;
 
 namespace MonoDevelop.AnalysisCore
 {
@@ -44,10 +45,10 @@ namespace MonoDevelop.AnalysisCore
 
 	class ExportRulesHandler : CommandHandler
 	{
-		static MonoDevelopWorkspaceDiagnosticAnalyzerProviderService.OptionsTable options =
-			((MonoDevelopWorkspaceDiagnosticAnalyzerProviderService)Ide.Composition.CompositionManager.GetExportedValue<IWorkspaceDiagnosticAnalyzerProviderService> ()).Options; 
+		static Task<MonoDevelopWorkspaceDiagnosticAnalyzerProviderService.OptionsTable> optionsTask =
+			                                                                 ((MonoDevelopWorkspaceDiagnosticAnalyzerProviderService)Ide.Composition.CompositionManager.GetExportedValue<IWorkspaceDiagnosticAnalyzerProviderService> ()).GetOptionsAsync(); 
 
-		protected override void Run ()
+		protected override async void Run ()
 		{
 			var lang = "text/x-csharp";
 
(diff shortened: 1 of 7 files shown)


main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/AnalysisCommands.cs
```diff
@@ -44,10 +45,10 @@ public enum AnalysisCommands
 
 	class ExportRulesHandler : CommandHandler
 	{
-		static MonoDevelopWorkspaceDiagnosticAnalyzerProviderService.OptionsTable options =
-			((MonoDevelopWorkspaceDiagnosticAnalyzerProviderService)Ide.Composition.CompositionManager.GetExportedValue<IWorkspaceDiagnosticAnalyzerProviderService> ()).Options; 
+		static Task<MonoDevelopWorkspaceDiagnosticAnalyzerProviderService.OptionsTable> optionsTask =
```
Therzok: We should stop saving MEF values into static values, as that will not allow us to recompose.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/MonoDevelopWorkspaceDiagnosticAnalyzerProviderService.cs
```diff
@@ -44,16 +45,16 @@ partial class MonoDevelopWorkspaceDiagnosticAnalyzerProviderService : IWorkspace
 
 		const bool ClrHeapEnabled = false;
 
-		internal OptionsTable Options = new OptionsTable ();
-		readonly ImmutableArray<HostDiagnosticAnalyzerPackage> hostDiagnosticAnalyzerInfo;
+		private TaskCompletionSource<OptionsTable> optionsCompletionSource = new TaskCompletionSource<OptionsTable> ();
```
Therzok: readonly

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/MonoDevelopWorkspaceDiagnosticAnalyzerProviderService.cs
```diff
@@ -99,11 +101,12 @@ ImmutableArray<HostDiagnosticAnalyzerPackage> CreateHostDiagnosticAnalyzerPackag
 
 					// Figure out a way to disable E&C analyzers.
 					assemblies.Add (asm.Location);
-					Options.ProcessAssembly (asm);
+					options.ProcessAssembly (asm);
 				} catch (Exception e) {
 					LoggingService.LogError ("Error while loading diagnostics in " + asm.FullName, e);
 				}
 			}
+			optionsCompletionSource.SetResult (options);
```
Therzok: In case a source already has its result set, it'll fail.

Let's use a `TrySetResult` and on failure log that we got double called from roslyn.

DavidKarlas: Why? This method is only called once.

Therzok: ah, the task is started in the constructor. 👍

PR #4801, opened 2018-05-15
[Ide] Fix command route for document tab context menu commands

sevoku:
Fixes VSTS #588821


Therzok:
Cc @iainx for accessibility behaviour

PR #4802, opened 2018-05-15
[Tests] Introduce new IdeUnitTests.dll

Therzok:
This PR adds a new assembly that implements some TestBase functionality that can be reused by addins' test suites.

It also adds TypeForwarders for whatever users of MonoDevelop.Ide.Tests.dll just in case, so we don't break ABI.


Therzok:
This allows for running tests from a non-deployed tests collection from md-addins

Therzok:
@slluis PTAL

slluis:
A more solid solution would be to change the logic in UnitTests.dll that assumes tests run on the MD tests directory, then that assembly can be local copied when necessary.

Therzok:
@slluis repurposed this PR.

It seems like the way to go is for each out of tree test assembly to define its own test-projects directory to use as a test roots dir.

slluis:
Tests failed...

Therzok:
Had forgotten to add the test project.

Therzok:
Looks good to go!

Therzok:
Rebased.

PR #4806, opened 2018-05-15
Fix mimeTypes attribute on CSharpAutoInsertBracketHandler.

KirillOsenkov:
It was being loaded for all mime types. Now the mimeTypes will be respected.

Bug https://github.com/mono/mono-addins/issues/116 is a suggestion to add a warning to prevent such errors in the future.


slluis:
Is there any reported bug that this fixes?

KirillOsenkov:
No, I’ve discovered this as I was fixing the “”” issue, but it doesn’t fix it alone.

PR #4807, opened 2018-05-16
Fixes VSTS Bug 615849: Automatic matching brace completion deletes `{`

mkrueger:
when `}` is deleted

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/615849


DavidKarlas:
Will this work correctly also in case I use Delete(not Backspace) key?

slluis:
> Will this work correctly also in case I use Delete(not Backspace) key?

It looks like so.

mkrueger:
If you can reproduce any error with the delete key just open an issue.

Issue #4691, opened 2018-04-27
Create new event handler method wants to put method at start of class


mrward:
VS Mac 7.5.0.1236.

# To repro

1) Create a large class. To verify this I created a single view iOS project and opened the AppDelegate.cs - The class should have many lines so the 'class' header definition is not visible when you are at the end of the class.
2) Go to the end of the class.
3) Create an event at the end of the class.
    public event EventHandler Foo;
4) Create a method after this event handler and then type in:
    this.Foo += 
5) Select Handler_Foo in the intellisense window that opens.

![eventhandlermethodinsertion2](https://user-images.githubusercontent.com/372361/39354945-b9f00a64-4a03-11e8-8af7-5e09eb8c6914.gif)

# Expected result

The insertion point for the event handler method is near where you are in the code. Above or below the method you are in.

# Actual result

The insertion point is at the start of the class.
If the class header is not visible the insertion point is not shown since the text editor does not seem to scroll to line.
This moves you away from where you are coding so you lose context.

> VS bug [#607580](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/607580)

xamarin-release-manager:
Fixed in version 7.6.0.835 (master)

Pull Request #4813 merged by: Mike Krüger
Author: mono
Commit: e32d088cbf8837b7065bb9fcefb5f1aae29007f0 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #4813 merged by: Mike Kr??ger
Author: mono
Commit: e32d088cbf8837b7065bb9fcefb5f1aae29007f0 (xamarin/monodevelop)

PR #4813, opened 2018-05-16
Fixes issue #4691 Create new event handler method wants to put method

mkrueger:
at start of class.


main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SourceEditorView.cs
```diff
@@ -2680,7 +2680,23 @@ void ITextEditorImpl.StartInsertionMode (InsertionModeOptions insertionModeOptio
...
+					var point = mode.InsertionPoints [i];
+					if (point.Location < caretLocation) {
+						mode.CurIndex = i;
+					} else {
+						mode.CurIndex = i;
+						break;
+					}
+ 				}
+
+			}
```
mrward: Can mode.InsertionPoints ever be empty? Just wondering since mode.CurIndex would not be set in that case and would be -1. Not sure if this could happen and what would happen if mode.CurIndex -1.

Therzok: Yes, it can be empty.

mkrueger: At that point it can't

PR #4814, opened 2018-05-16
Fixes VSTS 558577: Setting Place System directives first cannot be tu…

mkrueger:
…rned off

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/617333

This option is not supported anymore.


mrward:
VS 2017 supports this. Just wondering why VS Mac cannot support it?

In VS 2017 you can go to options - Text Editor - C# - Advanced and there is a `Place 'System' directives first when sorting usings` check box. It is off by default.

mhutch:
This can be controlled from editorconfig, so there's definitely a Roslyn option we can map to

mkrueger:
In any case it is >not< a formatting option.

Therzok:
This looks better.

Issue #4818, opened 2018-05-16
Double click text selection is often blocked by new tooltip


mrward:
Double clicking to select some text often seems to result in the tooltip being clicked and not the text itself.

![doubleclickpreventedbytooltip](https://user-images.githubusercontent.com/372361/40120137-6d6c9672-5916-11e8-8db5-86dd4eb34226.gif)

I think the tooltip appears too close to the caret. The original tooltip was below the line. You cannot click the original tooltip window since it disappears when you try to move the caret over it.

> VS bug [#617309](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/617309)

xamarin-release-manager:
Fixed in version 7.6.0.831 (master)

Pull Request #4826 merged by: Kirill Osenkov
Author: mono/dev/kirillo
Commit: a8e3893b363adc3020909191951d5b2e90a36459 (mono/monodevelop)

mrward:
Doesn't this problem still happen for those files that use this new tooltip, such as with web files such as .css? The fix applied just hides the tooltip for C# files.

For example, if you create a new ASP.NET Core MVC project, then open the site.css and mis-type a css property:

```
/* Wrapping element */
/* Set some basic padding to keep content from hitting the edges */
.body-content {
	padding-left: 15px;
	paddinng-right: 15px;
}
```

Then if you hover over the mis-typed paddinng-right the new tooltip blocks the click

KirillOsenkov:
Yes, we do need to fix the actual problem, my PR was just to stop the bleeding in the common case.

mkrueger:
https://github.com/mono/monodevelop/pull/4977 could maybe fix the problem it removes the old style tooltips completely.

xamarin-release-manager:
Fixed in version 7.8.0.951 (master)

Pull Request #6075 merged by: Lluis Sanchez
Author: mono
Commit: cff04dc9f05da5e8246d88ed275416c8e8ec7e42 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.0.0.471 (release-8.0)

Pull Request #6075 merged by: Lluis Sanchez
Author: mono
Commit: cff04dc9f05da5e8246d88ed275416c8e8ec7e42 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #6075 merged by: Lluis Sanchez
Author: mono
[...]

Issue #4816, opened 2018-05-16
Tooltip does not disappear on typing or moving to another item


mrward:
VS Mac 7.6.0.818 (master)

The new tooltips, which are different to the original tooltips, do not disappear on typing or moving to another item on the same line. The original tooltips work OK - they disappear on typing or moving to something else on the same line. This makes typing difficult at times since you cannot see what you are typing - you have to move the caret to dismiss the tooltip and then go back to writing some code.

Screen recording:

![tooltipnotdisappear2](https://user-images.githubusercontent.com/372361/40117488-0a3c6f34-590f-11e8-8825-268e5cc620da.gif)

> VS bug [#617304](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/617304)

xamarin-release-manager:
Fixed in version 7.6.0.831 (master)

Pull Request #4826 merged by: Kirill Osenkov
Author: mono/dev/kirillo
Commit: a8e3893b363adc3020909191951d5b2e90a36459 (mono/monodevelop)

KirillOsenkov:
Yes, we haven't bumped WebToolingAddin yet. I'll keep the bug open to check after we do so.

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #4826 merged by: Kirill Osenkov
Author: mono/dev/kirillo
Commit: a8e3893b363adc3020909191951d5b2e90a36459 (xamarin/monodevelop)

PR #4826, opened 2018-05-16
Fix https://github.com/mono/monodevelop/issues/4816 (disable Roslyn Quick Info)

KirillOsenkov:
Disable the Roslyn-based Quick Info for now for C#, because it conflicts with the MonoDevelop one.

We will reenable it later as part of moving to VS Editor.

Also fixes https://github.com/mono/monodevelop/issues/4818


mhutch:
Maybe we should have an editorconfig in the VSEditor directory using the VS codestyle to prevent mismatched style like this?

KirillOsenkov:
I'm OK either way, maybe we should reformat all these files and add copyright headers in a separate commit?

PR #4841, opened 2018-05-18
Fixes VSTS 616981: [Feedback] VSMac Freezes When Saving A file

mkrueger:
Can't repro the issue but the diff tracker relied on text
changing/text changed pairs which is kind of broken in the current
editor. Reworked it so that only the text changed event is used &
handled the exception from the log.


neerajdeshmukh:
Verified on build configuration Visual Studio Enterprise 2017 for Mac Version 7.5.2 (build 14).
It is working as expected. Hence approved for Merging. 
Build Gist: https://gist.github.com/mohakbarokar/e4df413656312e2e411009ba77ab4da2

PR #4855, opened 2018-05-21
Spinner speed increase

vancura:
Spinners frames now take 40ms instead of previous 60ms – and some were even at default 200ms.


vancura:
Sorry, I didn't see it before. Everything should be fixed now.

PR #4858, opened 2018-05-21
[PerfMetrics] Add platform related metadata

iainx:
This adds some missing platform related telemetry to the start up event.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/620314.


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
```diff
@@ -679,6 +679,16 @@ static IdeCustomizer LoadBrandingCustomizer ()
...
 
+			var platform = DesktopService.PlatformTelemetry ();
+			metadata ["TimeFromMachineStart"] = platform.TimeSinceMachineStart.ToString ();
+			metadata ["CpuArchitecture"] = platform.CpuArchitecture;
+			metadata ["CpuFamily"] = platform.CpuFamily.ToString ();
+			metadata ["CpuFrequency"] = platform.CpuFrequency.ToString ();
+			metadata ["HardDriveTotalVolumeSize"] = platform.HardDriveTotalVolumeSize.ToString ();
+			metadata ["HardDriveFreeVolumeSize"] = platform.HardDriveFreeVolumeSize.ToString ();
+			metadata ["RamTotal"] = platform.RamTotal.ToString ();
+			metadata ["HadDriveOsMediaType"] = platform.HardDriveOsMediaType.ToString ();
```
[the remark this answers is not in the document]
mrward: Just wondering if the startup event is where this all this data should go. VS on Windows does not put all this data in its vs/core/perf/startup event. Looking at the VS startup event I can only see TimeSinceMachineStart being set here. Should we be looking at where VS on Windows puts this data and do something similar?

mrward:
To get this data into the telemetry there also needs to be a change in md-addins in the CounterMapping class so the new properties are mapped across.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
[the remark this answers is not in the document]
iainx: Yeah, we should, I asked what event most of this data went onto, but haven't had an answer yet

main/src/addins/MacPlatform/MacTelemetryDetails.cs
```diff
@@ -0,0 +1,258 @@
...
+			var attrs = NSFileManager.DefaultManager.GetFileSystemAttributes ("/");
+			size = attrs.Size;
+			freeSize = attrs.FreeSize;
+
+			osType = GetMediaType ("/");
+		}
+
+		public TimeSpan TimeSinceMachineStart => TimeSpan.FromMilliseconds (NSProcessInfo.ProcessInfo.SystemUptime);
+
+		public TimeSpan TimeSinceLogin => TimeSpan.Zero;
```
Therzok: `CTL_KERN `, `KERN_BOOTTIME`?

Therzok: `kern.boottime:`

iainx: surely that's what `NSProcessInfo.ProcessInfo.SystemUptime` is using?
TimeSinceLogin has problems on macOS, because the utmp log is only accessible with elevated privileges and doesn't just record users logging into the computer, but also any Terminal.app that is opened as a login terminal. So if I open a new terminal before starting VSM, the TimeSinceLogin could be counted as a few seconds.

Therzok: @iainx users don't usually start VSM from a terminal

iainx: @Therzok no, you misunderstand me, if they open a terminal, that will be counted as a login.

Therzok:
```
int main()
{
    struct lastlogx ll = {}, *ll2;
    printf ("Initial: %lds\n", ll.ll_tv.tv_sec);

    getlastlogxbyname("therzok", &ll);
    printf ("%lds\n", ll.ll_tv.tv_sec);

    ll2 = getlastlogxbyname("therzok", NULL);
    printf ("%lds\n", ll2->ll_tv.tv_sec);
    free (ll2);
}
```

this worked for me.

iainx: It works on my machine with SIP enabled.

main/src/addins/MacPlatform/MacTelemetryDetails.cs
```diff
@@ -0,0 +1,275 @@
...
+				CFReleaseInternal (cfRef);
+		}
+
+		/*
+		 * It appears that getlastlogxbyname only works if you have elevated permissions
+		 * but it also doesn't distinguish between user login into the system and user opening a new login terminal
+		 */
+		static DateTime GetLoginTime ()
+		{
+			var llHandle = getlastlogxbyname (Environment.UserName, IntPtr.Zero);
```
Therzok: Allocate the LastLogX struct on the stack

```
var ll = new LastLogX();
if (IntPtr.Zero == getlastlogxbyname (Environment.UserName, ref ll))
    return DateTime.MinValue;

// Rest of the code
```

That way we can avoid allocating native memory.

main/src/addins/MacPlatform/MacTelemetryDetails.cs
```diff
@@ -0,0 +1,275 @@
...
+			if (llHandle == IntPtr.Zero) {
+				// getlastlogxbyname doesn't work if SIP is disabled
+				return DateTime.MinValue;
+			}
+
+			var ll = Marshal.PtrToStructure<LastLogX> (llHandle);
+
+			var epoch = new DateTime (1970, 1, 1, 0, 0, 0, 0, DateTimeKind.Local);
```
Therzok: DateTimeOffset.FromUnixTime

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
```diff
@@ -90,6 +90,24 @@ public static class IdeApp
 			}
 		}
 
+		static EventHandler startupCompleted;
+		public static event EventHandler StartupCompleted {
+			add {
+				Runtime.RunInMainThread (() => {
```
Therzok: I would also add that we should handle this the same way we handle `Initialized`. But it can be done in a later PR.

Issue #4868, opened 2018-05-22
Text editor misrenders when jumping to a line of source


alanmcgovern:
Use quicksearch to jump to a line of source and then the screen corrupts. Using command-a to select all text forces a re-render and fixes the problem: http://recordit.co/zQJUx4ZX0a

This happens across a lot of files.

> VS bug [#620304](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/620304)

xamarin-release-manager:
Fixed in version 7.6.0.866 (master)

Author: Mike Krüger
Commit: 627e657551a63d22b723d687afaaea3e35a56d93 (mono/monodevelop)

PR #4869, opened 2018-05-22
Fixes issue #4868 Text editor misrenders when jumping to a line of

mkrueger:
source


mkrueger:
Didn't think that the folding change produced so many little side effects :/

PR #4870, opened 2018-05-22
[Core] Cache evaluated source files

mrward:
If the configuration has not changed and the project has not been
modified then the files returned from Project.GetSourceFilesAsync
will be returned from a cached result. This prevents the project
being re-evaluated each time this method is called.

Currently repeated calls to GetSourceFilesAsync only occur when
the project has files that need to be generated. The project's
PerformGeneratorAsync method calls GetProjectFileTimestamps which
then calls GetSourceFilesAsync. Note that the GetProjectFileTimestamps
will most likely be removed with the file watcher integration since
it is only used to programmatically generate FileService.FileChanged
events which would be automatically generated by the file watcher.

Fixes VSTS #620341 - Cache evaluated source files.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -542,12 +542,8 @@ namespace MonoDevelop.Projects
 			// pre-load the results with the current list of files in the project
 			var results = new List<ProjectFile> ();
 
-			var buildActions = GetBuildActions ().Where (a => a != "Folder" && a != "--").ToArray ();
-
-			var config = configuration != null ? GetConfiguration (configuration) : null;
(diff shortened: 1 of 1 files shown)


Therzok:
Nice!

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -557,6 +553,49 @@ public Task<ProjectFile[]> GetSourceFilesAsync (ConfigurationSelector configurat
 			return results.ToArray ();
 		}
 
+		object evaluatedSourceFilesLock = new object ();
+		string evaluatedSourceFilesConfiguration;
+		TaskCompletionSource<ImmutableList<ProjectFile>.Builder> evaluatedSourceFilesTask;
```
Therzok: Could've been immutable array

PR #4878, opened 2018-05-23
[Ide] Fix roslyn option querying of CodeStyleOption/NamingPreferences…

Therzok:
…Style in case of default values

With custom serializers, we need to take into account that null might not be the default value.

In that case, set null accordingly and default values where it should.

Fixes startup log spew of failing to deserialize CodeStyle


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices.Options/MonoDevelopGlobalOptionPersister.cs
```diff
@@ -74,9 +74,17 @@ internal MonoDevelopGlobalOptionPersister (IGlobalOptionService globalOptionServ
...
-			};
+			PropertyService.PropertyChanged += OnPropertyChanged;
+		}
+
+		void OnPropertyChanged (object sender, Core.PropertyChangedEventArgs args)
+		{
+			Refresh (args.Key, args.NewValue);
+		}
+
+		void IDisposable.Dispose ()
```
Therzok: This is needed because unit tests work on the same PropertyService instance.

PR #4887, opened 2018-05-24
Fixes VSTS 621570: [Feedback] TODO doesn't highlight (worked before

mkrueger:
Update)

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/621570

Tried to add a tagger but it was a very complex approach and
interfered with the c# tagger - overall this is way more efficient
than re evaluating the syntax tree again and works for any tag based
language not just for one.


DavidKarlas:
Imo, this needs to be added to Roslyn tagger or maybe via custom MonoDevelop tagger... Definitely doesn't belong into this class.

KirillOsenkov:
I agree with David, in my opinion this class should only be concerned with mapping the classification spans to MonoDevelop highlighted lines. All language-specific logic should be elsewhere.

That said, I don't have a good alternative solution off the top of my head, so maybe this is OK for now?

I'm meeting Dustin today, I'll talk with him and see what he thinks.

mkrueger:
As said taggers are fairly complex and would simply be a performance issue. The correct solution would be to allow roslyn to be extended - but that's atm not possible.

The good thing at that solution is that it works for >any< language - how does that work with taggers ?

PR #4890, opened 2018-05-24
[BuildOutput] Add basic telemetry counters

rodrmoya:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -392,15 +392,17 @@ public Task<BuildOutputNode> FirstMatch (string pattern)
 			currentMatchIndex = -1;
 
 			return Task.Run (() => {
-				if (!string.IsNullOrEmpty (pattern)) {
-					// Perform search
-					foreach (var root in rootNodes) {
-						root.Search (currentSearchMatches, currentSearchPattern);
-					}
+				using (Counters.SearchBuildLog.BeginTiming ()) {
```
Therzok: This shouldn't live here. It should be done on textentry focus in

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/MSBuildOutputProcessor.cs
```diff
@@ -60,14 +60,15 @@ public override void Process ()
...
 
-			try {
-				binlogReader.Replay (FileName);
-			}
-			catch (Exception ex) {
-				LoggingService.LogError ($"Can't process {FileName}: {ex.ToString ()}");
+				try {
+					binlogReader.Replay (FileName);
+				} catch (Exception ex) {
+					LoggingService.LogError ($"Can't process {FileName}: {ex.ToString ()}");
```
Therzok: This exception shouldn't live in logs only.

PR #4891, opened 2018-05-24
Small project loading optimization

slluis:
When loading a solution set the stored active configuration before adding
the solution to the workspace. In this way roslyn will use the correct
configuration to do the initial analysis, instead of trying to load with the
default configuration and later load again when stored preferences are
restored.

Fixes VSTS #603762 - LoadProject is called twice on solution load due
to configuration change


Therzok:
Can we unit test this?

slluis:
Added unit test.

Therzok:
Perfect!

PR #4894, opened 2018-05-24
[Templates] Provide a way for Wizards to change their number of pages

jstedfast:
Needed for the Azure Functions wizard because the number of pages
is dynamic depending on which function is selected.

(Also needed to handle if the user hits Back and selects a different
function template)

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/TemplateWizardProvider.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/TemplateWizardProvider.cs
@@ -35,6 +35,7 @@ namespace MonoDevelop.Ide.Templates
 {
 	class TemplateWizardProvider
 	{
+		TemplateWizard currentWizard;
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/TemplateWizardProvider.cs
```diff
@@ -43,7 +44,19 @@ class TemplateWizardProvider
 		public bool IsFirstPage { get; private set; }
 		public bool IsLastPage { get; private set; }
 		public int CurrentPageNumber { get; private set; }
-		public TemplateWizard CurrentWizard { get; private set; }
+
+		public TemplateWizard CurrentWizard {
+			get { return currentWizard; }
+			private set {
+				if (currentWizard != null) {
+					currentWizard.TotalPagesChanged -= OnTotalPagesChanged;
```
Therzok: Where is this unsubscribed? It can cause ref cycles.

PR #4895, opened 2018-05-24
[TextEditor] Check commit string for null or empty.

bratsche:


slluis:
Is there any bug related to this?

bratsche:
@slluis I ran into a NRE while I was debugging issues with Korean input because it tried to commit when there was no preedit string. I ended up removing that change from my current tree, so I'm not sure if the NRE can still be hit but I thought this still seemed like a decent safety precaution just in case.

PR #4899, opened 2018-05-25
[Ide] Optimize roslyn in case of memory pressure

Therzok:
This PR adds the following:
* Deterministically sets the roslyn logger when we initialize the threading data
* Added a memory monitor API in DesktopService which currently just notifies whether the memory status has changed.
* Removes the full solution analysis old option and better mirror how VS handles it (ClosedFileDiagnostic persisted, FullSolutionAnalysis at runtime)
* Disable full solution analysis (background solution crawler) and flush MRU caches when memory pressure is notified by the OS


Therzok:
I really wanted to have an integration test for this, but all the solutions require either:
a) installing a privileged helper tool for the test suite which means having an app bundle for the test runner
b) root
c) trashing the OS for a while

slluis:
Which functionality we lose when full analysis is disabled? Maybe we should warn the user when that happens?

Therzok:
Yeah, that was my next step, showing an info on the status bar that we disabled it due to memory pressure.

We lose analyzers being run in closed files (diagnostics in closed files - we don't surface yet, todo comments in closed files)

Therzok:
@vancura looping you in:

VSWindows has this UX for this mechanism: http://go.microsoft.com/fwlink/?LinkID=799402&clcid=0x409

We currently cannot replicate it. What should we tell the user here?

vancura:
I wanted to file a UI request for the information ticker, but to be honest, I didn't have time.

For now, could you write something in error colors into the status area, and we'll add the ticker later?

slluis:
I don't like the proposed UI for this, let's use a notification icon instead of a status message. Also, please don't add markup support to the status area. If we need to mark a message as important, we'll add a property for it.

vancura:
Yes, please don't create new UI. We need to do it right via a proper UX research.

Therzok:
@slluis In that case, an informational icon with "Advanced features were disabled to improve performance" and clicking it would open the options dialog?

vancura:
I'll design it next week.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/Gui/AnalysisOptionsPanel.cs
```diff
@@ -60,7 +61,7 @@ class AnalysisOptionsWidget : VBox
 		CheckButton enabledFullCheck;
 		CheckButton enabledTest;
 
-		public AnalysisOptionsWidget ()
+		public AnalysisOptionsWidget (string languageName)
```
iainx: Is `languageName` needed?

iainx:
MacPlatform stuff looks good. `NSProcessInfo` has notifications about the thermal state of the machine as well, it might be good to stop the background crawler if the machine gets too hot?

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/Gui/AnalysisOptionsPanel.cs
Therzok: Right now, no. But most of the roslyn options are per language. So it'd make sense to have this extensible

Therzok:
Well, the thing is:
a) background crawler finishes at some point
b) we might never even get it running to the end if we turn it off when the fans start working

We could experiment with that, memory pressure is something that can make the IDE slow down to a crawl.

Fwiw, when macOS starts compressing your app's memory, it starts being really slow.

iainx:
Well, I assume that when we start the crawler again it doesn't start from the beginning again? So even if we spin it down and up as the fans go crazy it will get to the end.

I guess it depends whether the heat is being caused by us, or by some external system.

Therzok:
> Well, I assume that when we start the crawler again it doesn't start from the beginning again? So even if we spin it down and up as the fans go crazy it will get to the end.

it does.

slluis:
What is the UI? can you show screenshots?

Therzok:
screen shot 2018-05-30 at 18 59 58
screen shot 2018-05-30 at 19 00 25

I think these are the latest versions

Therzok:
This is under the gtk backend. Waiting on https://github.com/mono/mono/pull/8951 for a mac backend.

Therzok:
screen shot 2018-05-30 at 18 59 58
screen shot 2018-05-30 at 19 00 25

Latest versions above. The `Re-enable` button was changed to `Restore`.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices/MonoDevelopInfoBarService.cs
```diff
@@ -0,0 +1,119 @@
...
+using MonoDevelop.Core;
+using MonoDevelop.Ide.Gui.Components;
+using Roslyn.Utilities;
+
+namespace MonoDevelop.Ide.RoslynServices
+{
+	[ExportWorkspaceService(typeof(IInfoBarService), layer: ServiceLayer.Host), Shared]
+    sealed class MonoDevelopInfoBarService : ForegroundThreadAffinitizedObject, IInfoBarService
+    {
+        readonly IForegroundNotificationService _foregroundNotificationService;
```
slluis: Spaces?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -116,25 +115,70 @@ internal MonoDevelopWorkspace (MonoDevelop.Projects.Solution solution) : base (H
...
+			IdeApp.Preferences.EnableSourceAnalysis.Changed += OnEnableSourceAnalysisChanged;
+
+			// TODO: Unhack C# here when monodevelop workspace supports more than C#
+			IdeApp.Preferences.Roslyn.CSharp.SolutionCrawlerClosedFileDiagnostic.Changed += OnEnableFullSourceAnalysisChanged;
 
 			foreach (var factory in AddinManager.GetExtensionObjects<Microsoft.CodeAnalysis.Options.IDocumentOptionsProviderFactory>("/MonoDevelop/Ide/TypeService/OptionProviders"))
 				Services.GetRequiredService<Microsoft.CodeAnalysis.Options.IOptionService> ().RegisterDocumentOptionsProvider (factory.Create (this));
 
+			if (solution != null)
+				DesktopService.MemoryMonitor.StatusChanged += OnMemoryStatusChanged;
```
slluis: Shouldn't we also check the current memory status? if memory status is low since the beginning and that doesn't change, the event won't be raised.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/InfoBar.cs
```diff
@@ -0,0 +1,231 @@
...
+	{
+
+	}
+
+	class XwtInfoBar : Widget
+	{
+		static Image closeImage = Image.FromResource ("pad-close-9.png");
+		static Image closeImageInactive = Image.FromResource ("pad-close-9.png").WithAlpha (0.5);
+
+		class Styles
```
slluis: Please define the colors in https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Styles.cs, and let that class handle the theme switch.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
Therzok: Getting the current memory status is more complicated. I'm taking a look at how to pinvoke and get kernel info via `task_vm_info_data_t`, as the dispatch source is only for notifications.

Therzok: Ah, we can do that simply by querying sysctl:
`sysctl kern.memorystatus_vm_pressure_level`

iainx: We should share the sysctl pinvoke API in MacPlatform

Therzok:
Ready to merge!

Therzok:
@slluis PTAL

PR #4904, opened 2018-05-25
Fixes missing semantic highlighting after switch to TagBasedSyntaxHightligthing after scrolling

DavidKarlas:
Problem was that Semantic highligthing was generated only for text visible on screen + few lines before and after
But since MonoTextEditor didn't report `LayoutChanged` events SemanticTagger didn't know it needs to update classifications for new area of file hence it didn't work
With this change we notify hence it works...

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
@@ -2232,7 +2232,7 @@ namespace Mono.TextEditor
 				foreach (var drawer in margin.MarginDrawer)
 					drawer.Draw (cr, cairoRectangle);
 			}
-			
+			LayoutChanged?.Invoke (this, EventArgs.Empty);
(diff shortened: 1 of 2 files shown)


main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
```diff
@@ -3578,6 +3578,7 @@ public override GLib.GType ChildType ()
 		}
 		
 		internal List<MonoTextEditor.EditorContainerChild> containerChildren = new List<MonoTextEditor.EditorContainerChild> ();
+		internal event System.Action LayoutChanged;
```
mkrueger: Please use EventHandler - I know it's not needed but it's a pattern.

DavidKarlas:
Fixed to use EventHandler

garuma:
My member declarations and invocations are still not being classified correctly even after this PR.

slluis:
@garuma yes, this PR doesn't fix the declaration vs invocation issue, it fixes a rendering issue

PR #4912, opened 2018-05-25
[Roslyn] Implement a background compiler that also holds a strong ref to the most recently used compilation

Therzok:


Therzok:
Need to fix the unit tests. The breakage came in from foregroundthreaddata not being deterministically set in unit tests.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/ProcessWrapper.cs
```diff
@@ -102,11 +102,9 @@ public void WaitForOutput ()
 					// Ignore
 				}
 
-				lock (lockObj) {
```
DavidKarlas: ManualResetEvent is thread-safe but, that if `endEventOut.Close();` is called before `endEventOut=null` is called. Before this could not happen but now, it can happen and `ObjectDisposedException` will be thrown: https://msdn.microsoft.com/en-us/library/system.threading.eventwaithandle.set(v=vs.90).aspx
My thinking is... Do:
```
try {
   endEventOut?.Set ();
} catch(ObjectDisposedException) {}
```
Notice I used `?.` instead `if` check which is more thread-safe.

Therzok: Well, the reason I added this was because the tests were hanging due to thread.abort() being called from the unit test runner, and the WaitOne() being done in a lock before Set() was being done, so the hold was being held.

Maybe the actual fix is to keep the lock for set/set to null, and not the waitone.

DavidKarlas:
@Therzok can we add `LoggingService.LogWarning("Low memory disabling caching of type system");` so in case users complain about slow type system we can check logs if this is the reason...

Therzok:
Adding it as part of memory monitor PR

Therzok:
Also, might be worth adding a loginfo when the user toggles it.

Therzok:
Pushed thread safety fix.

PR #4913, opened 2018-05-25
[TextEditor] Ensure preedit is committed before selecting left/right

bratsche:
Before we shift-left or shift-right to select, ensure that the preedit
buffer is committed or the selection can be in the wrong place and
mess things up.

Fixes VSTS #617555

Also commit the preedit string on mouse click. This fixes issues with
text selection when you try to select the last visible character when
there is a preedit string (described in VSTS #617559) and also another
bug I observed that isn't in VSTS where if you click elsewhere in the
text editor while there is a preedit string then text gets moved in a way
that looks very jarring and unpredictable to the user.

Fixes VSTS #617559


bratsche:
QA approved

PR #4916, opened 2018-05-26
Include line break in TryGetSnapshotPoint.

KirillOsenkov:


KirillOsenkov:
This method should work even if the location requested is after the line break at the end of the line.

PR #4924, opened 2018-05-28
Fix VSTS 605711: Incorrect XAML completion for </ContentPage.Re

DavidKarlas:
Added `.` to ignored commit characters for closing element completion. Reason for this is that `.` can be part of element name...
Also moved `...` in front of element name because:
  a) it's more correct since other elements are inserted before
  b) because general code completion logic has weird rule to "ignore" `IsCommitCharacter` if `DisplayText` ends with `.`
Also found bug with `InsertCompletionText` that was attempting to fix caret position but failed, but also failed at inserting text by not removing existing text.


slluis:
@mkrueger can you review?

slluis:
@DavidKarlas The changes are causing some test failures.

mkrueger:
Unit tests would've been nice.

slluis:
@DavidKarlas can you please fix the unit tests?

PR #4933, opened 2018-05-29
[PerfMetrics] Track the experienced times for key presses

iainx:
The length of time a user experiences for a keypress is the time taken from
keypress to the key being visible on screen. Adjust TextEditorKeyPressTiming to
handle multiple keys being pressed before the expose event happens

Fixes VSTS #620320


main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
```diff
@@ -1239,62 +1247,76 @@ void SetCursor (Gdk.Cursor cursor)
...
-				Gdk.ModifierType mod;
-				KeyboardShortcut[] accels;
-				GtkWorkarounds.MapKeys (evt, out key, out mod, out accels);
-				//HACK: we never call base.OnKeyPressEvent, so implement the popup key manually
-				if (key == Gdk.Key.Menu || (key == Gdk.Key.F10 && mod.HasFlag (ModifierType.ShiftMask))) {
-					OnPopupMenu ();
-					return true;
-				}
+				keyPressTimings.StartTimer ();
+				return HandleTextKey (evt);
```
mrward: Is there any possibility that we could call StartTimer, then have an exception occur somewhere so EndTimer is never called? Or is OnExposeEvent always called?

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextEditorKeyPressTimings.cs
```diff
@@ -39,11 +42,34 @@ class TextEditorKeyPressTimings
 		TimeSpan? firstTime;
 		int count;
 
+		// The length of time it takes to process a key is the time
+		// from the key being pressed to the character being drawn on screen
+		// As the expose event is on a fixed cycle, it's possible that multiple characters
+		// can be pressed before the expose event happens again.
+		// We need to keep a track of keys that have been pressed and are waiting to be drawn on screen
+		//
+		List<long> activeCounts = new List<long> ();
```
Therzok: How big will this list get?

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.csproj
```diff
@@ -129,7 +129,6 @@
       <Private>False</Private>
     </Reference>
     <Reference Include="Microsoft.VisualStudio.Text.Implementation">
-      <SpecificVersion>False</SpecificVersion>
       <HintPath>..\..\..\build\bin\Microsoft.VisualStudio.Text.Implementation.dll</HintPath>
```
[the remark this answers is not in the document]
Therzok: Not a big problem, but we should consolidate nuget packages.

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextEditorKeyPressTimings.cs
```diff
@@ -53,33 +79,44 @@ public void StartTimer ()
...
+					AddTime (duration);
+				}
 
-			totalTime += duration;
-			count++;
+				activeCounts.Clear ();
+			} else {
+				// Some keypresses do not trigger a draw event, so we process them once
+				// they are finished and remove them from the activeCounts list
+				var tick = activeCounts.Last ();
```
Therzok: No LINQ extensions here please. We have a list.

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextEditorKeyPressTimings.cs
```diff
@@ -53,33 +79,44 @@ public void StartTimer ()
...
-			}
+			if (complete) {
+				stopwatch.Stop ();
+				stopwatch.Reset ();
 
-			if (!firstTime.HasValue) {
-				firstTime = duration;
-			}
+				foreach (var ticks in activeCounts) {
+					var duration = new TimeSpan (durationTicks - ticks);
```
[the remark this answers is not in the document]
iainx: Yeah, I wasn't sure how much overhead `Elapsed` would add over `ElapsedTicks` so I went with ticks, which is the wrong thing

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
iainx: `OnExposeEvent` will always be called, at most 60times a second. If an exception somehow prevents one from being triggered there'll be another one along shortly after, and if there's not lots of other things are going to break.

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextEditorKeyPressTimings.cs
iainx: it'll get cleared every time an expose event happens, so it'll have as many keys as you can type between the event. Although, I've seen the speed you type, so probably about 30 :)

If the UI thread bogs down, then the expose event happens less frequently, but it still shouldn't get very long.

Therzok: Can we keep a pre-allocated array and a count integer which we can reuse?

I really don't like the memory pressure we cause on typing just for this (we end up creating lots of small underlying arrays, especially since the list is not pre-allocated).

Therzok: What I mean here is have an array per timings instance and reuse it by just resetting the count value.

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextEditorKeyPressTimings.cs
```diff
@@ -27,21 +27,60 @@
...
 
-		public void StartTimer ()
+		// The length of time it takes to process a key is the time
+		// from the key being pressed to the character being drawn on screen
+		// As the expose event is on a fixed cycle, it's possible that multiple characters
+		// can be pressed before the expose event happens again.
+		// We need to keep a track of keys that have been pressed and are waiting to be drawn on screen
+		//
+
+		const int numberOfCountSpaces = 100;
```
Therzok: Reuse this below in `activeCounts`

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextEditorKeyPressTimings.cs
```diff
@@ -53,33 +92,47 @@ public void StartTimer ()
...
-				firstTime = duration;
+			var telemetry = DesktopService.PlatformTelemetry ();
+			if (telemetry == null) {
+				activeCountIndex = 0;
+				return;
 			}
 
-			totalTime += duration;
-			count++;
+			var sinceStartup = (long)DesktopService.PlatformTelemetry ().TimeSinceMachineStart.TotalMilliseconds;
```
mrward: Can we measure what the overhead is of creating a new PlatformTelemetry class each time has on the key press? Stopwatch overhead is minimal. It looks like the PlatformTelemetry creates a bunch of things and calls a set of APIs, whilst we are only trying to get the TimeSinceMachineStart.

mrward: Here it seems we are creating PlatformTelemetry twice? Should we not be using the telemetry instance just created above?

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextEditorKeyPressTimings.cs
```diff
@@ -27,21 +27,60 @@
...
+using MonoDevelop.SourceEditor;
+using MonoDevelop.Ide;
 
-namespace MonoDevelop.SourceEditor
+namespace Mono.TextEditor
 {
 	class TextEditorKeyPressTimings
 	{
-		Stopwatch stopwatch = new Stopwatch ();
+		//Stopwatch stopwatch = new Stopwatch ();
```
mrward: If this is not being used the commented out line should be removed.

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextEditorKeyPressTimings.cs
```diff
@@ -53,33 +92,47 @@ public void StartTimer ()
 		/// which can take ~800ms for the first ever key press in the text editor for
 		/// the current IDE session.
 		/// </summary>
-		public void EndTimer ()
+		public void EndTimer (bool complete = false)
```
mrward: This method has a comment which has timings that apply to using the Stopwatch.

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextEditorKeyPressTimings.cs
iainx: `PlatformTelemetry` is a singleton, created once at startup.

iainx: but yes, should just reuse that variable

slluis: If it is a singleton, better expose it as a property instead of a method.

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -1144,14 +1144,11 @@ internal override void RestartIde (bool reopenWorkspace)
 			proc.Start ();
 		}
 
-		MacTelemetryDetails details;
-		internal override IPlatformTelemetryDetails PlatformTelemetryDetails ()
-		{
-			if (details == null) {
-				details = MacTelemetryDetails.CreateTelemetryDetails ();
+		Lazy<MacTelemetryDetails> details = new Lazy<MacTelemetryDetails> (() => MacTelemetryDetails.CreateTelemetryDetails ());
+		internal override IPlatformTelemetryDetails PlatformTelemetryDetails {
```
Therzok: This is a nitpick, but I'd rather have the `PlatformTelemetryDetails` property be implemented in DesktopService:
```
readonly Lazy<IPlatformTelemetryDetails> platformTelemetryDetails = new Lazy<IPlatformTelemetryDetails> (() => CreatePlatformTelemetryDetails ());
IPlatformTelemetryDetails PlatformTelemetryDetails => platformTelemetryDetails.Value
```

iainx: Yeah, was thinking that it should be moved up to DesktopService

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/PlatformService.cs
```diff
@@ -586,9 +586,10 @@ internal virtual string GetNativeRuntimeDescription ()
 			return null;
 		}
 
-		internal virtual IPlatformTelemetryDetails PlatformTelemetryDetails ()
-		{
-			return null;
+		internal virtual IPlatformTelemetryDetails PlatformTelemetryDetails {
```
Therzok: Now that there is a lazy in DesktopService, make this a method `CreatePlatformTelemetryDetails`, so implementations don't do the wrong thing.

Therzok:
md-addins build failed.

slluis:
@iainx We should report the key press timings only if there were more than 500 keystrokes.

iainx:
@slluis I understood it that we reported all the keystrokes and they were filtered on the 500/week at the query side.

iainx:
@monojenkins backport 7.6

iainx:

mohakbarokar:
@iainx We tested this PR on Build Taken From: https://devdiv.visualstudio.com/DevDiv/Default/_build/results?buildId=1783476&view=logs

Build Gist: https://gist.github.com/mohakbarokar/d58d0831e2044cc97dc73295eaf5ced0

& We are facing Blocker Bug https://devdiv.visualstudio.com/DevDiv/XamarinVS/_workitems/edit/637410

Marking this PR as **QA-Failed**

mohakbarokar:
Verified on Visual Studio Enterprise 2017 for Mac (Preview) Version 7.6 Preview (7.6 build 1814).
It is working fine. Hence approved for merge.
Build Gist: https://gist.github.com/mohakbarokar/e19d9e7a7dcc6f7da8a5dcad95ebc455

PR #4943, opened 2018-05-30
Add binding redirects for SQLite binaries to match the latest version.

KirillOsenkov:


KirillOsenkov:
Hmm, I wonder why BinaryCompatChecker didn't complain about this.

Therzok:
Because SQLitePCLRaw started using assembly versioning in 0.11

KirillOsenkov:
Yeah, but I'd expect BinaryCompatChecker to fail as soon as you updated to 1.1.10. Maybe I don't check for this case?? I'll investigate separately.

KirillOsenkov:
Argh!! Found a problem in BinaryCompatChecker!! It should have caught this, but has a bug!

Issue #3493, opened 2017-12-12
Load syntax schemes on demand for each file type


Therzok:
Right now, syntax highlighting loads all syntax schemes eagerly on first request:
https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/SyntaxHighlightingService.cs#L460

> VS bug #568061

Therzok:
```
1.42%   TextEditorDisplayBinding..cctor  •  1,267 ms  •  1 call  •  MonoDevelop.Ide.Editor.TextEditorDisplayBinding..cctor
  1.42%   InitSourceEditor  •  1,267 ms  •  1 call  •  MonoDevelop.Ide.Editor.TextEditorDisplayBinding.InitSourceEditor
    1.42%   LoadCustomStylesAndModes  •  1,263 ms  •  1 call  •  MonoDevelop.Ide.Editor.TextEditorDisplayBinding.LoadCustomStylesAndModes
      1.42%   SyntaxHighlightingService..cctor  •  1,262 ms  •  1 call  •  MonoDevelop.Ide.Editor.Highlighting.SyntaxHighlightingService..cctor
        1.20%   LoadStylesAndModes  •  1,065 ms  •  2 calls  •  MonoDevelop.Ide.Editor.Highlighting.SyntaxHighlightingService.LoadStylesAndModes(Assembly)
          ►1.19%   LoadFile  •  1,060 ms  •  2,842 calls  •  MonoDevelop.Ide.Editor.Highlighting.SyntaxHighlightingService.LoadFile(LanguageBundle, String, Func[Stream], Func[IStreamProvider])
          ►0.00%   SyntaxHighlightingService+<>c__DisplayClass31_1..ctor  •  1 ms  •  2,842 calls  •  MonoDevelop.Ide.Editor.Highlighting.SyntaxHighlightingService+<>c__DisplayClass31_1..ctor
           0.00%   GetManifestResourceNames  •  0 ms  •  2 calls  •  System.Reflection.RuntimeAssembly.GetManifestResourceNames
           0.00%   CtorClosed  •  0 ms  •  5,684 calls  •  System.MulticastDelegate.CtorClosed(Object, IntPtr)
          ►0.00%   OnAssemblyLoadEvent  •  0 ms  •  1 call  •  System.AppDomain.OnAssemblyLoadEvent(RuntimeAssembly)
[...]

PR #4949, opened 2018-05-30
Issue #3493 Load syntax schemes on demand for each file type

mkrueger:
Brings loading to <60ms on my machine. Unfortunately some files need to be loaded to the end. The text mate format is not really well thought out. The sublime text 3 format is way better :).


mkrueger:
Note: we should wait with the merge for the branch - we need some time to test it internally. In theory it should be covered with the unit test suite.

mkrueger:
unit test failure unrelated - rebased to latest master today.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.TextMate/TextMateBundleOptionsPanelWidget.cs
```diff
@@ -90,8 +90,10 @@ void AddLanguageBundle (object sender, EventArgs e)
 			if (success) {
 				var bundle = SyntaxHighlightingService.LoadStyleOrMode (SyntaxHighlightingService.userThemeBundle, newFileName) as LanguageBundle;
 				if (bundle != null) {
-					foreach (var h in bundle.Highlightings)
-						h.PrepareMatches ();
+					foreach (var h in bundle.Highlightings) {
+						var def = h as SyntaxHighlightingDefinition;
```
Therzok:
```
if (h is SyntaxHighlightingDefinition def)
    def.PrepareMatches ();
```

mkrueger: 2 LOC are 2 LOC  - I would even argue my version is better because it's not increasing the indent level.

Therzok: Yeah, was just setting a standard, but as said, LGTM

PR #4952, opened 2018-05-30
Xml schema update

mhutch:
Fixes [VSTS #551305](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/551305) - Incorrect XSLT validation error for xsl:variable.


slluis:
Are there any changes worth mentioning in release notes?

mhutch:
In theory it should fix VSTS 551305, though I didn't specifically test that.

mhutch:
I tested that example and got a validation error, looks like Mono doesn't like the XSD schema.

mhutch:
Turns out the error was because .net doesn't support 1.1 XSD schemas, so I downgraded the schema.

Added a test for the issue reported in VSTS 551305. I'd like to add a test for the XSLT transform command in case I regressed it. 

Also, XSD-based intellisense seems pretty badly broken, but that was an issue already.

mhutch:
@mrward @slluis ok, tests added and a couple more bugs fixed :)

main/src/addins/Xml/MonoDevelop.Xml.csproj
```diff
@@ -63,6 +64,9 @@
       <HintPath>..\..\..\build\bin\System.Collections.Immutable.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <Reference Include="System.ValueTuple">
+      <HintPath>..\..\..\packages\System.ValueTuple.4.4.0\lib\net461\System.ValueTuple.dll</HintPath>
+    </Reference>
```
mrward: This seems to be missing a Private false. I guess this is what is causing the duplicate check failures and causing the VSTS build to fail.

mhutch: Good catch, fixed!

It's been about 10 years since I looked at it too :|

mhutch:
@mrward the remaining issue appears to have been present in the repo already

mrward:
Yep, just the binary compat failure which I believe has been fixed on master.

PR #4957, opened 2018-05-30
Fix VSTS 623602: SnapshotSpan is on a different ITextSnapshot than th…

KirillOsenkov:
…is SnapshotSpan.

After a text change only the lines that were re-rendered were moved to the latest snapshot. This resulted in some lines being on an older snapshot than others.

Now we forcibly advance all lines to the latest snapshot after every change.


KirillOsenkov:
Fixed Todd's suggestion to subscribe directly from within MdTextViewLines. Had to make sure the initialization order is correct.

mkrueger:
+1

PR #4958, opened 2018-05-31
Fix wrong `ProjectCacheService` being returned resulting in caching n…

DavidKarlas:
…ot working properly


Therzok:
Ouch, good catch!

Issue #4950, opened 2018-05-30
Error navigating to type in assembly browser


mhutch:
Attempting to got to definition on the type `Enumerable.Union` failed with the following error. 

```
ERROR [2018-05-30 00:25:28Z]: Assembly browser: Can't find: M:System.Linq.Enumerable.Union``1(System.Collections.Generic.IEnumerable{``0}).
```

This was preceded by a huge number of errors looking very similar to the following.

```
ERROR [2018-05-30 00:25:28Z]: Exception while creating generic instance type id string System.Func`3<TFirst,TSecond,TResult>
System.ArgumentException: Value does not fall within the expected range.
  at Mono.Cecil.Rocks.TypeReferenceRocks.MakeGenericInstanceType (Mono.Cecil.TypeReference self, Mono.Cecil.TypeReference[] arguments) [0x00020] in <bbd684a8611f4ae984cf8cd45f9b22b3>:0 
  at MonoDevelop.AssemblyBrowser.UnresolvedIdStringProvider.AppendTypeName (System.Text.StringBuilder sb, Mono.Cecil.TypeReference type, System.Boolean explicitInterfaceImpl, System.Int32 outerTypeParameterCount) [0x001cf] in /Users/mikayla/code/monodevelop/main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/UnresolvedIdStringProvider.cs:161 
```

> VS bug [#624054](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/624054)

xamarin-release-manager:
Fixed in version 7.6.0.1475 (master)

Author: Mike Krüger
Commit: e187f5bf70215205a618e4f87b473d488d8fb30e (mono/monodevelop)

PR #4959, opened 2018-05-31
Fixes issue #4950 Error navigating to type in assembly browser

mkrueger:
After fixing & unit testing the IdStringProvider I found out that it
is just redundant since cecil actually can generate proper ID strings…
the id string provider was a left over from nrefactory days…


Therzok:
Nice!

PR #4975, opened 2018-06-01
Properly disable new QuickInfo for all scenarios for now.

KirillOsenkov:


minestarks:
@KirillOsenkov That's fine, we do have a TypeScript `IAsyncQuickInfoSession` in our 15.8 drop, but I implemented a `TooltipProvider` to call it directly and display the tooltip as a short term solution.

KirillOsenkov:
@DavidKarlas @slluis can we merge this? This disables another tooltip on Squiggles that comes from Roslyn.

KirillOsenkov:
This will be more necessary with Roslyn 2.9 because they've added more stuff to the tooltip.

PR #4983, opened 2018-06-01
Fix slow code completion caused by AdditionalDocuments

DavidKarlas:
It's not really clear what is purpose of `AdditionalDocuments`. I can't find any use case in Roslyn source code(or MonoDevelop). I also debugged VS Windows and checked on MonoDevelop.Ide.csproj and Wpf application from template, neither had any `AdditionalDocuments`. Problem was at https://github.com/dotnet/roslyn/blob/5b113ae/src/Workspaces/Core/Portable/Workspace/Solution/ProjectState.cs#L202 which iterates over all documents and additionalDocuments and checks for latest version to see if compilation needs to be recalculated. In case of MonoDevelop.Ide.csproj documents were ~1k and additionalDocuments were ~2.8k(.png and other files).
"C#: Got completions" went from  ~1second to ~40ms.
I suspect also other cases where Semantic Model is requested will be faster.

I also added `DebuggerDisplay` for `TmerTraceList` this is result:
image

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimeCounter.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimeCounter.cs
@@ -210,21 +210,18 @@ namespace MonoDevelop.Core.Instrumentation
 
 		string DebuggingText {
 			get {
-				if (FirstTrace == null)
-					return string.Empty;
 				var stringBuilder = new StringBuilder ();
-				ToString (stringBuilder, FirstTrace, null);
+				var current = FirstTrace;
+				TimerTrace previous = null;
+				while (current != null) {
+					stringBuilder.Append (previous == null ? "N/A" : (current.Timestamp - previous.Timestamp).ToString (@"ss\.fff"));
+					stringBuilder.Append (" : ");
+					stringBuilder.AppendLine (current.Message);
+					previous = current;
+					current = current.Next;
+				}
 				return stringBuilder.ToString ();
 			}
-		}
-
-		void ToString (StringBuilder stringBuilder, TimerTrace current, TimerTrace previous)
-		{
-			stringBuilder.Append (previous == null ? "N/A" : (current.Timestamp - previous.Timestamp).ToString (@"ss\.fff"));
-			stringBuilder.Append (" : ");
-			stringBuilder.AppendLine (current.Message);
-			if (current.Next != null)
-				ToString (stringBuilder, current.Next, current);
 		}
 	}
 	
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimeCounter.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimeCounter.cs
@@ -28,6 +28,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Threading;
+using System.Text;
 
 namespace MonoDevelop.Core.Instrumentation
 {
@@ -196,6 +197,7 @@ namespace MonoDevelop.Core.Instrumentation
 	}
 	
 	[Serializable]
+	[DebuggerDisplay ("{DebuggingText}")]
 	class TimerTraceList
 	{
 		public TimerTrace FirstTrace;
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimeCounter.cs
```diff
@@ -205,6 +207,25 @@ class TimerTraceList
...
+				return stringBuilder.ToString ();
+			}
+		}
+
+		void ToString (StringBuilder stringBuilder, TimerTrace current, TimerTrace previous)
+		{
+			stringBuilder.Append (previous == null ? "N/A" : (current.Timestamp - previous.Timestamp).ToString (@"ss\.fff"));
+			stringBuilder.Append (" : ");
+			stringBuilder.AppendLine (current.Message);
+			if (current.Next != null)
```
Therzok: Would prefer a while loop instead of recursion.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -649,11 +651,6 @@ static bool CanGenerateAnalysisContextForNonCompileable (MonoDevelop.Projects.Pr
 						continue;
 					documents.Add (CreateDocumentInfo (solutionData, p.Name, projectData, f, sck));
 				} else {
-					var id = projectData.GetOrCreateDocumentId (f.Name, oldProjectData);
-					if (!duplicates.Add (id))
-						continue;
-					additionalDocuments.Add (CreateDocumentInfo (solutionData, p.Name, projectData, f, sck));
```
Therzok: We do need to add additional documents, it's just that our mechanism to choose which is bad.

Normal documents are the ones in the `Compile` item group, AdditionalDocuments are in the `AdditionalFiles` itemgroup:
http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces.MSBuild/MSBuild/Constants/ItemNames.cs,7

Therzok: i.e. xaml files go here.

DavidKarlas: .xaml files as AdditionalDocuments are used in VS Windows specific part of Roslyn, and we have this solved differently, so imo, for now, lets just completely remove all AdditionalDocuments and add it when needed, hence I added comment and not remove everything related to AdditionalDocuments...

Therzok: Yeah, was just mentioning. Given we'll probably want to reuse roslyn vswin xaml support, we might have to deal with this.

Therzok: But why can't we just grab `AdditionalFiles` from the `CoreCompileDependsOn` msbuild query we do?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -633,6 +633,8 @@ static bool CanGenerateAnalysisContextForNonCompileable (MonoDevelop.Projects.Pr
 		Tuple<List<DocumentInfo>, List<DocumentInfo>> CreateDocuments (ProjectData projectData, MonoDevelop.Projects.Project p, CancellationToken token, MonoDevelop.Projects.ProjectFile [] sourceFiles, ProjectData oldProjectData)
 		{
 			var documents = new List<DocumentInfo> ();
+			// We don' add additionalDocuments anymore because they were causing slowdown of compilation generation
```
KirillOsenkov: don't - t is missing

KirillOsenkov:
Yes, we should not be blindly adding all random files as additional documents. We should be intentional about adding the ones we actually need. VS for Windows doesn't add all files.

slluis:
Maybe we should only add files known to be text files?

KirillOsenkov:
Not even worth it for text files. There's literally no reason for any additional documents to be added right now. It's just a load on the system.

PR #4990, opened 2018-06-04
Fixes VSTS Bug 625320: [Feedback] Code completion in anonymous method…

mkrueger:
…s broken when Intellisense - Show Import Items turned on

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/625320


mkrueger:
the compatibility check seems to fail in master :/

PR #5002, opened 2018-06-05
[CounterMetadata] Add a nicer way to handle property setting for meta…

Therzok:
…data

Also, a minor cleanup in the code

Fixes VSTS #626931 - Typed counter metadata is more complex that it needs to be


Therzok:

PR #5006, opened 2018-06-05
[Roslyn] Implement IDocumentSupportsFeatureService

iainx:


DavidKarlas:
@iainx can you just confirm `== null` is intentional?

Therzok:
@DavidKarlas yes, it is. If the document is registered, it's a contained language (i.e. razor)

@slluis none yet opened for it, as it isn't used yet.

The idea is that VSWindows disables the features specified in the service for Razor documents, as that's not supported in the commanding layer that will be integrated from editor features in the future.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices/MonoDevelopDocumentSupportsFeatureService.cs
```diff
@@ -0,0 +1,60 @@
...
+using MonoDevelop.Ide.TypeSystem;
+
+namespace MonoDevelop.Ide.RoslynServices
+{
+	[ExportWorkspaceService (typeof(IDocumentSupportsFeatureService), ServiceLayer.Host), Shared]
+	internal sealed class MonoDevelopDocumentSupportsFeatureService : IDocumentSupportsFeatureService
+	{
+		public bool SupportsCodeFixes (Document document)
+		{
+			return (MonoDevelopHostDocumentRegistration.FromDocument (document) == null);
```
Therzok: To ensure clarity, maybe we should add a helper method like:
```
static bool IsContainedDocument (Document document)
{
    return MonoDevelopHostDocumentRegistration.FromDocument (document) == null;
}
```

and the service becomes:
```
...
public bool SupportsNavigationToAnyPosition (Document document) => !IsContainedDocument (document);
...
```

iainx:
@Therzok updated

PR #5008, opened 2018-06-05
Fix warnings

mhutch:
Fixes >400 warnings. There's 84 left.


main/src/addins/Xml/Editor/XmlParsedDocument.cs
```diff
@@ -43,49 +43,41 @@ public XmlParsedDocument (string fileName) : base (fileName)
...
 		{
-			return System.Threading.Tasks.Task.FromResult((IReadOnlyList<FoldingRegion>)Foldings.ToList ());
+			return Task.FromResult((IReadOnlyList<FoldingRegion>)Foldings.ToList ());
 		}
 
 		public IEnumerable<FoldingRegion> Foldings {
 			get {
 				if (XDocument == null)
 					yield break;
-				foreach (var region in GetCommentsAsync().Result.ToFolds ())
```
Therzok: Are we supposed to do that?

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Parser/CSharpParsedDocument.cs
```diff
@@ -330,8 +330,12 @@ static bool SkipError (bool isAdhocProject, string errorId)
 		{
 			if (Ide.IdeApp.Preferences.EnableSourceAnalysis)
 				return emptyErrors;
-			
+
+			// FIXME: remove this fallback, error squiggles should always be handled via the source analysis mechanism
```
Therzok: The only reason we still have this is to slow live errors even if source analysis is disabled. I think we should not, but VS does not allow source analysis to be disabled.

Might be worth completely removing it, then we can get rid of parsed document completely

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Refactoring/CodeGenerationService.cs
```diff
@@ -166,11 +166,12 @@ static class CodeGenerationService
 			root = await document.GetSyntaxRootAsync (cancellationToken).ConfigureAwait (false);
 
 			var node = root.GetAnnotatedNodes (insertedMemberAnnotation).Single ();
+			var model = await doc.AnalysisDocument.GetSemanticModelAsync (cancellationToken);
 
-			Application.Invoke (async (o, args) => {
+			Application.Invoke ((o, args) => {
```
Therzok: nice

main/src/addins/CSharpBinding/MonoDevelop.Ide.Completion.Presentation/MyRoslynCompletionData.cs
```diff
@@ -56,7 +56,7 @@ public abstract class MyRoslynCompletionData : CompletionData
 
 		public CompletionItem CompletionItem { get; private set; }
 
-		public CompletionItemRules Rules { get { return CompletionItem.Rules; } }
+		public override CompletionItemRules Rules { get { return CompletionItem.Rules; } }
```
Therzok: nice catch

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring/InsertionPointService.cs
```diff
@@ -27,19 +27,20 @@
...
+using MonoDevelop.Ide.Editor;
 using MonoDevelop.Ide.TypeSystem;
 
 namespace MonoDevelop.Refactoring
 {
 	public static class InsertionPointService
 	{
-		public static List<InsertionPoint> GetInsertionPoints (IReadonlyTextDocument data, MonoDevelop.Ide.TypeSystem.ParsedDocument parsedDocument, ITypeSymbol type, int part)
+		[Obsolete ("Use overload that takes a SemanticModel")]
+		public static List<InsertionPoint> GetInsertionPoints (IReadonlyTextDocument data, ParsedDocument parsedDocument, ITypeSymbol type, int position)
```
Therzok: So, this code change is going to break non-roslyn languages (xml).

Therzok: Xml does not have a semantic model.

main/src/addins/MonoDevelop.SourceEditor2/VSEditor/Text/Impl/ToolTipAdornment/ViewElementFactoryService.cs
```diff
@@ -63,7 +63,7 @@ internal sealed class ViewElementFactoryService : IViewElementFactoryService
         }
 
         private IEnumerable<Lazy<IViewElementFactory, IViewElementFactoryMetadata>> OrderedFactories
-            => this.orderedFactories ?? (this.orderedFactories ?? Orderer.Order(this.unorderedViewFactories));
+            => this.orderedFactories ?? (this.orderedFactories = Orderer.Order(this.unorderedViewFactories));
```
Therzok: Nice.

main/src/addins/MonoDevelop.SourceEditor2/VSEditor/Language/Impl/Intellisense/SignatureHelp/DefaultSignatureHelpPresenterSurfaceElement.cs
```diff
@@ -149,7 +149,6 @@ void ShowTooltipInfo()
             Content.QueueForReallocate();
         }
 
-        public event EventHandler MouseLeftButtonDown;
```
Therzok: I'd prefer having pragma warning here. This is new code, it might end up being used.

Therzok:
`Task.Ignore()` will mask the warnings and mark them as hidden technical debt on some cases.

We need a way to make a distinction between intentionally ignored tasks and ones which get ignored just because the API evolved and were not migrated.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring/InsertionPointService.cs
mhutch: XML doesn't use the refactoring insertion points.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Parser/CSharpParsedDocument.cs
mhutch: Surely we could get live errors by just disabling all the analyzers when analysis is turned off?

main/src/addins/CSharpBinding/MonoDevelop.Ide.Completion.Presentation/MyRoslynCompletionData.cs
mhutch: <3 warnings

main/src/addins/MonoDevelop.SourceEditor2/VSEditor/Language/Impl/Intellisense/SignatureHelp/DefaultSignatureHelpPresenterSurfaceElement.cs
mhutch: If we need it, we can add it back, surely?

main/src/addins/Xml/Editor/XmlParsedDocument.cs
Therzok: Okay, then if we set that expectation, sure.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Parser/CSharpParsedDocument.cs
Therzok: Sooooo, this is more complex.

Compiler errors/warnings are enabled as incremental analyzers only. And they're toggled by default in the MDWorkspace, but if analyzers are disabled, it won't work.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring/InsertionPointService.cs
Therzok: Still, we're limiting all the languages which don't use roslyn.

mhutch: InsertionPointService internally depends on Roslyn anyways and is only used from the C# binding. If we have a need for a more generic version, we can implement that when we have a better idea of what we need :)

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Parser/CSharpParsedDocument.cs
mhutch: Right but we could disable all analyzers that aren't compiler errors/warnings but enable the analyzer mechanism surely?

Therzok: We could make it so incremental analyzers work, but that also enables normal analyzers.

http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces/SolutionCrawler/ISolutionCrawlerRegistrationService.cs,15

mhutch:
@Therzok it wasn't obvious to me that any of the ones here _should_ be awaited. Any specific concerns?

mhutch:
Current status - fixed > 400 warnings.

There's 78 left. Of these:

* 15 are fallout from obsoleting highlighting colors. It's not obvious to me whether the calling code can be deleted or whether the warnings should be suppressed /cc @KirillOsenkov @mkrueger 
* 20 are conflicts with imported editor types /cc @KirillOsenkov
* 16 are reference version conflicts. Not sure how to fix these.
* 4 are fallout from obsoleting ClrVersion
* 8 are broken pieces from the incomplete port of the old webforms/razor editors to roslyn

main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement.Gui/PackageCellView.cs
```diff
@@ -276,7 +276,7 @@ Point GetPackageImageLocation (Size imageSize, Rectangle cellArea)
 			return new Point (width, height);
 		}
 
-		protected override Size OnGetRequiredSize ()
+		protected override Size OnGetRequiredSize (SizeConstraint widthConstraint)
```
Therzok: what.

main/src/addins/AspNet/Razor/RazorCSharpParser.cs
```diff
@@ -95,40 +91,43 @@ public override bool CanGenerateProjection (string mimeType, string buildAction,
...
-				var line = doc.GetLineByOffset (idx);
-				var offset = line.NextLine.Offset + map.Value.StartGeneratedColumn - 1;
-
-				var seg = new ProjectedSegment (map.Value.StartOffset.Value, offset, map.Value.CodeLength);
-				segments.Add (seg);
-			}
+				return Task.FromResult ((IReadOnlyList<Projection>)new Projection[0]);
+
+			return Task.Run (() => {
+				var doc = TextEditorFactory.CreateNewDocument (new StringTextSource (code), razorDocument.PageInfo.ParsedDocument.FileName, "text/x-csharp");
```
Therzok: I don't think this is thread safe.

main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement.Gui/PackageCellView.cs
mhutch: Obsolete overload.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeFormatting/CodeFormatter.cs
```diff
@@ -53,13 +53,11 @@ internal CodeFormatter (string mimeType, AbstractCodeFormatter formatter)
 			this.formatter = formatter;
 		}
 
-		[Obsolete ("Use roslyn instead")]
```
mkrueger: Was there for a purpose - y the 13 callers are not migrated which is bad. 

There won't be a general purpose formatter anymore that's not part of the core - instead each language binding needs to provide that functionality on it's own.

The API here is not suited for formatting. Only in the basic case.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Parser/CSharpParsedDocument.cs
mkrueger: We should just remove the source analysis disable function or make it purely a visiblet hing (CSxxx IDs can always be shown)

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring/InsertionPointService.cs
mkrueger: Doesn't make sense for non C# atm since the roslyn refactorings are >roslyn< refactorings :).
The InsertionMode is always there for non roslyn languages to use. We're not going to provide real generic things from beside: just plug it in. Providing general APIs which at the end can't be used for more than  2 languages don't make sense.

That's what language bindings are for. (btw. roslyn is just a gigantic language binding for .NET based languagesr

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeFormatting/CodeFormatter.cs
mhutch: Sure, this API has issues. But we do need an extensibility point into which language services can plug formatters for use by the "new file" and "new project" commands. You can't just obsolete a necessary API without a replacement.

mhutch:
@mkrueger I figured that might be contentious so I left it for a later commit :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewFileDialog.cs
```diff
@@ -543,10 +543,10 @@ protected void CheckedChange (object sender, EventArgs e)
 				}
 
 				if (project != null)
-					IdeApp.ProjectOperations.SaveAsync (project);
+					IdeApp.ProjectOperations.SaveAsync (project).Ignore ();
```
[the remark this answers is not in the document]
mhutch: Nah, file template creation shouldn't block on project save.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.OptionPanels/D152AccessibilityPanel.cs
```diff
@@ -128,7 +128,7 @@ void ShowQuitOption (object sender, EventArgs args)
 		void RestartClicked (object sender, System.EventArgs e)
 		{
 			Save ();
-			IdeApp.Restart (true);
+			IdeApp.Restart (true).Ignore ();
```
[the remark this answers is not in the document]
mhutch: Sure, we could turn this event handler into `async void`, though I don't see much point :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditorViewContent.cs
```diff
@@ -84,7 +84,7 @@ public TextEditorViewContent (TextEditor textEditor, ITextEditorImpl textEditorI
 			if (ContentName != textEditorImpl.ContentName && !string.IsNullOrEmpty (textEditorImpl.ContentName))
 				AutoSave.RemoveAutoSaveFile (textEditorImpl.ContentName);
 			if (textEditorImpl.ContentName != null && textEditorImpl.ContentName != this.ContentName) {
-				EditorConfigService.RemoveEditConfigContext (textEditorImpl.ContentName);
+				EditorConfigService.RemoveEditConfigContext (textEditorImpl.ContentName).Ignore ();
```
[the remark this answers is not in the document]
mhutch: I don't think so. This is just disposing the stale context, no need to delay applying the new context

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring/InsertionPointService.cs
mhutch: @mkrueger we should really move Roslyn dependencies out of MonoDevelop.Ide...

main/src/addins/AspNet/Razor/RazorCSharpParser.cs
mhutch: @Therzok you mean `CreateNewDocument`? It just news up a document object.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewFileDialog.cs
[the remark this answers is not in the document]
Therzok: What if you open the file before it's actually saved? You'll get a document that's not attached to a project

main/src/addins/AspNet/Razor/RazorCSharpParser.cs
Therzok: I meant the `razorDocument.PageInfo.ParsedDocument.FileName`. `ParsedDocument` can change.

PR #5010, opened 2018-06-05
Check in BinaryChecker config and baseline.

KirillOsenkov:


KirillOsenkov:
The plan is to move the baseline to the MD repo, so that updating the baseline can be done without the md-addins dance.

Therzok:
👍

PR #5018, opened 2018-06-07
Build output fixes

rodrmoya:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -594,6 +594,10 @@ void ProcessLogs (bool showDiagnostics)
 						// Expand root nodes and nodes with errors
 						ExpandErrorOrWarningsNodes (treeView, buildOutputDataSource, false);
 						processingCompletion.TrySetResult (null);
+
+						FileNameChanged?.Invoke (this, filePathLocation.IsEmpty ?
```
Therzok: How do you dissociate between an actual path and a temporary title?

PR #5026, opened 2018-06-11
Do more work on background thread. Fixes VSTS #631120

nosami:


main/external/fsharpbinding/MonoDevelop.FSharpBinding/FSharpHighlightUsagesExtension.fs
```diff
@@ -41,34 +40,35 @@ type HighlightUsagesExtension() =
...
+                    try
+                        match resolveResult with
+                        | Some(fsSymbolName, references) ->
+                            LoggingService.logDebug "HighlightUsagesExtension: GetReferences starting on %s" (x.DocumentContext.Name |> IO.Path.GetFileName)
+                            //TODO: Can we use the DisplayName from the symbol rather than the last element in ident islands?
+                            // If we could then we could remove the Parsing.findLongIdents in GetUsesOfSymbolAtLocationInFile.
+                            references
+                            |> Seq.map (fun symbolUse ->
+                                            let start, finish = Symbol.trimSymbolRegion symbolUse fsSymbolName
+                                            let startOffset = x.Editor.LocationToOffset (start.Line, start.Column+1)
```
Therzok: Editor usage should be done on UI thread.

mrward: I thought that too. However that code seems to use a readonly text document. So is that not thread safe?

Therzok: Where is it using a RO Document? You can safely get a text snapshot on the UI thread and work on that.

mrward: It uses the IReadonlyTextDocument interface which will be implemented on the TextEditor and the text editor should not be used from a background thread.

nosami: If I need to switch to UI thread to run x.Editor.LocationToOffset, then it's probably not worth merging this PR at all. The thread context switches probably outweigh the cost of running this code on the UI thread.

mrward: You could, as Marius suggested, call TextEditor.CreateDocumentSnapshot and use that IReadonlyTextDocument in the background task.

PR #5027, opened 2018-06-11
[ExtensionChain] Fix default insertion position setting

Therzok:
Previously, this was done of the temporary extension chain, but it was removed in a cleanup.


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
```diff
@@ -396,6 +396,8 @@ void InitializeExtensionChain ()
 			extensionChain.SetDefaultInsertionPosition (defaultExts.FirstOrDefault ());
```
slluis: This can be removed now.

PR #5040, opened 2018-06-12
File watcher integration

mrward:
Use native file watcher to monitor solutions and individual files opened in the IDE.

Fixes VSTS #593841 - IDE focus-in/out does a lot of IO
Fixes VSTS #619697 - UI freeze after opening solution
Fixes VSTS #619707 - Git pull results in inconsistent project status


main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/Commands.cs
```diff
@@ -173,7 +173,9 @@ protected override void Run ()
 						}
 						finally {
 							monitor.Dispose ();
-							statusTracker.Dispose ();
+							Runtime.RunInMainThread (delegate {
+								FileService.ThawEvents ();
```
slluis: We can just do the RunInMainThread call inside ThawEvents() if necessary.

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
```diff
@@ -154,12 +154,6 @@ bool RefreshFile (string path, CheckoutNotifyFlags flags)
 		{
 			FilePath fp = RootRepository.FromGitPath (path);
 			Gtk.Application.Invoke ((o, args) => {
-				if (IdeApp.IsInitialized) {
-					MonoDevelop.Ide.Gui.Document doc = IdeApp.Workbench.GetDocument (fp);
-					if (doc != null)
-						doc.Reload ();
```
slluis: I think we still need this. If the user executes an operation that results in a file being updated, that file should be automatically reloaded in the editor if it is already open, it shouldn't notify the user that the file has changed.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/RevertCommand.cs
```diff
@@ -60,15 +60,6 @@ protected override void Run ()
 					list[0].Repository.Revert (list.Paths, true, Monitor);
 				
 				Gtk.Application.Invoke ((o, args) => {
-					foreach (VersionControlItem item in items) {
-						if (!item.IsDirectory) {
-							FileService.NotifyFileChanged (item.Path);
-							// Reload reverted files
-							Document doc = IdeApp.Workbench.GetDocument (item.Path);
-							if (doc != null && System.IO.File.Exists (item.Path))
-								doc.Reload ();
```
slluis: Same as my previous comment about Reload. The goal here is to avoid prompting the user about file change.

mrward:
Updated the pull request.
- FileService.ThawEvents now handles being called from a non-UI thread.
- Added back code to the Git addin that reloads files when they are reverted.

Issue #5025, opened 2018-06-11
Text loses indentation when typing


mrward:
When typing on a new line sometimes the code is moved all the way to the left and loses the indentation. Not sure exactly how to reproduce this. In the screencast below I was typing in some text into a new C# library project, then using backspace to remove it, this was done a few times and then sometimes the text gets shifted all the way to the left.

![texteditorindent](https://user-images.githubusercontent.com/372361/41226372-35702272-6d69-11e8-8c9a-17230214438e.gif)

- VS Mac 7.6.0.1544 (master Wrench build).
  - Unable to reproduce this on VS Mac 7.5.2.

xamarin-release-manager:
Fixed in version 7.6.0.1755 (release-7.6)

Pull Request #5043 merged by: Lluis Sanchez
Author: mono
Commit: 79d590f33e9242dd14b3d5cc920aa27716602eeb (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #5043 merged by: Lluis Sanchez
Author: mono
Commit: 79d590f33e9242dd14b3d5cc920aa27716602eeb (xamarin/monodevelop)

PR #5043, opened 2018-06-12
Fixes issue #5025 Text loses indentation when typing

mkrueger:
The virtual indent behavior was quite broken in the completion window
case. This should stop virtual indent to interfere with completion and
indent should be corrected after the completion session ends.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionController.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionController.cs
@@ -224,6 +224,8 @@ namespace MonoDevelop.Ide.CodeCompletion
 
 		public void HideWindow ()
 		{
+			if (CompletionWidget is ITextEditorImpl impl)
+				impl.FixVirtualIndentation ();
 			HideDeclarationView ();
 			view.Hide ();
 			ReleaseObjects ();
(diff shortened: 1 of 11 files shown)


mkrueger:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionController.cs
```diff
@@ -224,6 +224,8 @@ public bool ShowListWindow (ICompletionDataList list, CodeCompletionContext comp
 
 		public void HideWindow ()
 		{
+			if (CompletionWidget is ITextEditorImpl impl)
+				impl.FixVirtualIndentation ();
```
slluis: This call is out of place. The code completion controller shouldn't have to care about indentation.

mkrueger: Use case: indentation should correct when window hides.

Where should that be placed if not the control ?

slluis: Text editor deciding to fix indentation when completion window closes -> Yes
Completion window special casing editor to fix indentation when it closes -> No, not his business.

It is a matter of correctly setting responsibilities. The completion window controller controls the behavior of the completion window, not the behavior of the editor.

slluis:
@mrward can you please verify that this PR fixes the issue?

mrward:
Looks good to me. I cannot reproduce the problem I was seeing before 👍

PR #5048, opened 2018-06-13
[BuildOutput] Unsubscribe from events on Dispose

rodrmoya:
Fixes VSTS #625411 - Sometimes the build output does not update after building again


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -639,6 +644,7 @@ public Control CreatePathWidget (int index)
 		protected override void Dispose (bool disposing)
 		{
 			if (disposing) {
+				BuildOutput.OutputChanged -= OnOutputChanged;
```
Therzok: We should unsubscribe all of these even if not disposing.

PR #5053, opened 2018-06-13
[MSBuild] Use the global collection when evaluating projets

Therzok:
Optimizes MSBuild evaluation by reducing the total number of evaluations and reusing the same project collection when evaluating.


Therzok:
This PR contains commits from https://github.com/mono/monodevelop/pull/4593 which touch the actual MSBuild builder.

It should help with not re-evaluating a project multiple times are multiple phases

slluis:
@Therzok why aren't the ReevaluateIfNecessary calls necessary?

Therzok:
@slluis `CreateProjectInstance()` already does it

http://source.dot.net/#Microsoft.Build/Definition/Project.cs,2606

See last commit's message.

PR #5055, opened 2018-06-14
[Core] Fix editor errors when .NET Standard assembly referenced

mrward:
With a Xamarin.iOS project that uses an assembly that is compiled
for .NET Standard, such as the assembly used in the
System.Collections.Immutable NuGet package, the facade assemblies
were not being included in the list of referenced assemblies given
to the type system. This then resulted in the text editor showing
errors even though the project could be compiled without succesfully.
The error displayed was similar to:

    The type 'ValueType' is defined in an assembly that is not
    referenced. You must add a reference to assembly 'netstandard,
    Version=2.0.0.0, Culture=neutral, PublicKeytoken=cc7b1dffcd2ddd51'.

A short term fix has been added here. A check is made to determine if
the assembly referencing netstandard and if so the facade assemblies
are included. Previously only a check was made for the assembly
referencing System.Runtime. Ideally the information would be found by
using MSBuild.

Fixes VSTS #632353 - Editor doesn't work with .NET Standard 2 NuGets

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/SystemAssemblyService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/SystemAssemblyService.cs
@@ -460,6 +460,35 @@ namespace MonoDevelop.Core.Assemblies
 			return false;
 		}
 
+		static Dictionary<string, bool> facadeReferenceDict = new Dictionary<string, bool> ();
+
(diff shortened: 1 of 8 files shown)


slluis:
I guess ContainsReferenceToSystemRuntimeAsync can be deprecated?

mrward:
Yes I guess ContainsReferenceToSystemRuntimeAsync can be deprecated. If it was not public I would have removed it.

PR #5056, opened 2018-06-14
[BuildOutput] Add a spinner when processing .binlog files

rodrmoya:
Fixes VSTS #629161 - Log viewer does not show any loading progress

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
@@ -60,6 +60,7 @@ namespace MonoDevelop.Ide.BuildOutputView
 		Label resultInformLabel;
 		BuildOutputDataSearch currentSearch;
 		BuildOutputTreeCellView cellView;
+		MDSpinner loadingSpinner;
 
 		public string ViewContentName { get; private set; }
 		public BuildOutput BuildOutput { get; private set; }
@@ -213,6 +214,11 @@ namespace MonoDevelop.Ide.BuildOutputView
 			cellView.ExpandErrors += (s, e) => ExpandErrorOrWarningsNodes (treeView, false);
 
 			PackStart (treeView, expand: true, fill: true);
+
+			loadingSpinner = new MDSpinner (Gtk.IconSize.Button) {
+				Visible = false
+			};
+			PackStart (loadingSpinner, expand: true, vpos: WidgetPlacement.Center, hpos: WidgetPlacement.Center);
 		}
 
 		static void ExpandErrorOrWarningsNodes (TreeView treeView, bool warnings)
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -582,6 +596,8 @@ void ProcessLogs (bool showDiagnostics)
 			IsDirty = true;
 
 			Task.Run (async () => {
+				await SetSpinnerVisibility (true);
```
Therzok: Can't we just synchronously set this before we Task.Run()? Why do we want to switch context?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -213,6 +214,11 @@ void Initialize (DocumentToolbar toolbar)
 			cellView.ExpandErrors += (s, e) => ExpandErrorOrWarningsNodes (treeView, false);
 
 			PackStart (treeView, expand: true, fill: true);
+
+			loadingSpinner = new Spinner {
```
sevoku: Better use [`MonoDevelop.Components.MDSpinner`](https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/MDSpinner.cs#L34), it has a slightly different design and we use it everywhere else.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
netonjm: +1

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
```diff
@@ -570,6 +576,14 @@ static void ExpandChildrenWithErrorsOrWarnings (TreeView tree, BuildOutputDataSo
 			}
 		}
 
+		Task SetSpinnerVisibility (bool visible)
```
netonjm: This could be a normal method returning void

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
rodrmoya: This method can be called from any thread, and we need to, indeed, context switch, as some operations are UI and others really need to be ran on a background thread

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputWidget.cs
rodrmoya: ok, changed, looks better indeed

PR #5062, opened 2018-06-14
Merp upload mode

sgmunn:
using this PR to try and get a build...


sgmunn:
we do not need this anymore

PR #5077, opened 2018-06-15
[DotNetCore] Fix default build action for new file

mrward:
Adding a new .html file to the wwwroot folder of an ASP.NET Core
project would add the file as a None item instead of a Content
item. This would result in the .html file not being used when
publishing the project. If you use use 'dotnet publish' the
publish directory would not contain the .html file. ASP.NET Core
projects have different build actions for files based on where they
are added. An .html file in the root directory would be a None item
whilst a .html file in the wwwroot directory would be a Content item
by default. To fix this the default build action for a file is
determined from the file glob information available from the .NET
Core SDK.

Fixes VSTS #627227 - Publish the project, content is not the same
as the localhost page.


mrward:

PR #5079, opened 2018-06-15
[Core] Fix .xaml.cs intellisense in new Forms .NET Standard project

mrward:
Creating a new Xamarin.Forms .NET Standard project and then
modifing the .xaml to add new named UI items would result in no
intellisense in the .xaml.cs file until the solution was closed
and re-opened. The problem was that the .xaml and .xaml.cs files
were being removed from the ProjectFileCollection's files lookup
list when the project was re-evaluated. On re-evaluation after
the NuGet restore is first done the old MSBuild items for the
.xaml and .xaml.cs file have the wrong metadata, so they need
to be removed, whilst new MSBuild items with the updated metadata
need to be added. To fix this when re-evaluating the items are now
removed first and then added to prevent the newly added items being
removed incorrectly since they have the same filenames.

Fixes VSTS #615578 - Intellisense does not work for auto-generated
field via Xamarin.Forms XAML


mrward:

PR #5084, opened 2018-06-16
Remove dead code

mhutch:


decriptor:
Wow, strongwind... I haven't heard that in 10 years?

PR #5088, opened 2018-06-16
Replace ArrayList with List<> for memory and performance reasons

Therzok:


Therzok:
Agreed with @mkrueger . There should be some way to write roslyn refactorings for these kind of changes.

iainx:
@slluis Changes made

PR #5091, opened 2018-06-18
[NuGet] Fix updating PackageReference removes metadata

mrward:
On updating a NuGet PackageReference the old PackageReference was
removed from the project and then a new PackageReference was added.
This resulted in any custom MSBuild properties associated with the
PackageReference being removed. Now on updating a NuGet package the
version of the existing PackageReference is updated.

Fixes VSTS #568994 - Updating PackageReference removes metadata


mrward:

PR #5095, opened 2018-06-18
[Core] Fix UI hang when building ASP.NET Core 2.1 Angular project

mrward:
Creating a new ASP.NET Core 2.1 Angular project from the command line,
then opening it into the IDE, and building it would result in the
IDE hanging.

On building an ASP.NET Core 2.1 Angular project (also React.js and
Redux) will attempt to run an npm install. This results in thousands
of files being created in the node_modules directory. These files
being created resulted in the UI thread being used to check to
see if the file needed to be automatically added to the project
and resulted in a hang of the IDE. This check has been moved off
the UI thread when files are created which prevents the UI hang.

Also adding some code to ignore temporary files which are
generated by the text editor on saving (e.g. '.#MyClass.cs) which
broke TextFileUtility_WriteText_FileNotRemovedAndAddedBackToProject
in the tests. These temporary files were being added to the project
and breaking the test.

Fixes VSTS #632347 - UI hang when building ASP.NET Core 2.1 Angular
project

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -4244,16 +4244,22 @@ namespace MonoDevelop.Projects
 
 		void OnFileCreated (object sender, FileSystemEventArgs e)
 		{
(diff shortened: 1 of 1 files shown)


mrward:

PR #5102, opened 2018-06-19
Fix #4822: Fixing code comment command for all cases

### LINKED ISSUE #4822: Toggle line comment menu no longer exists
VS Mac 7.6.0.818 (master)

Right clicking a selection of code in the text editor and the Toggle line comment menu item is not available. So it is not currently possible to comment out a section of code without typing in the comments yourself.

> VS bug [#617345](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/617345)

DavidKarlas:
Problem was that in case of start/end of line where Roslyn didn’t report any specific highlighting we used defaultScope, problem with that was that after switch from RoslynClassifier to TagBased one I used Foreground which didn’t stack on top of “sources.cs”. So now `defaultScopeStack` is set correctly, and to better match with RoslynClassifier I also re-introduced userScope.
Also includes unit tests


DavidKarlas:

Issue #5105, opened 2018-06-19
sln file must map configurations even if they're not built


mhutch:
When projects are excluded from the build in certain configurations, the solution should still include configuration mappings for those projects. If it doesn't, then when the solution is opened in VS, VS will add configurations that _do_ build those projects, which can resulting in things breaking.

mhutch:
Actually, it's worse than that. They mapped configurations must exist or VS will reset them to something that does exist - AND set them to build.

PR #5107, opened 2018-06-20
Fix solution for building on Windows

mhutch:
Fixes an error where several projects were built in VS when they should not have been (see issue #5105)


main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.DevCertInstaller/MonoDevelop.AspNetCore.DevCertInstaller.csproj
```diff
@@ -18,15 +18,13 @@
...
-    <PlatformTarget>x64</PlatformTarget>
   </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|x64' ">
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <Optimize>true</Optimize>
     <OutputPath>..\..\..\..\build\bin</OutputPath>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <ExternalConsole>true</ExternalConsole>
-    <PlatformTarget>x64</PlatformTarget>
```
mrward: This change seems to have broken the ASP.NET dev certificate app. It crashes on startup now.

```
  at <unknown> <0xffffffff>
  at (wrapper managed-to-native) Security.Authorization.AuthorizationCreate (Security.AuthorizationItemSet*,Security.AuthorizationItemSet*,Security.AuthorizationFlags,intptr&) <0x00012>
  at Security.Authorization.Create (Security.AuthorizationParameters,Security.AuthorizationEnvironment,Security.AuthorizationFlags) [0x001f8] in <1d35452d95a94ac5bbd68a29ad6a434d>:0
  at MonoDevelop.AspNetCore.DevCertInstaller.MainClass.Run (string,string,string) [0x00031] in <bde88a041aeb46609477112e419aea87>:0
  at MonoDevelop.AspNetCore.DevCertInstaller.MainClass.Main (string[]) [0x00031] in <bde88a041aeb46609477112e419aea87>:0
  at (wrapper runtime-invoke) <Module>.runtime_invoke_int_object (object,intptr,intptr,intptr) [0x00054] in <bde88a041aeb46609477112e419aea87>:0
```

Can we change this part back to use x64?

mrward: Currently it is using ProcessService.StartProcess which I guess is running it with mono32 causing it crash. I guess the better solution would be to force the console app to run with mono64. I will take a look.

mrward: Looks like the code was using the MonoRuntime to get the mono version and that returns mono32. Made a change to always run the console app with mono64 - https://github.com/mono/monodevelop/pull/5120

mhutch: Huh looking at `GetMonoExecutableForAssembly` it looks to me like it should return `mono` in this case, not `mono32`.

Anyways, restoring the ` <PlatformTarget>x64</PlatformTarget>` seems like a perfectly good fix, we can do that without the `Platform=x64` which caused the mapping problems.

mrward: Debugging the code and it returned mono32. The MainModuleAttributes peflags that Cecil returned were: ILOnly | Required32Bit | Prefer32Bit

mhutch: Ah. Apparently Prefer32Bit is true by default for .netframework exes: https://github.com/Microsoft/msbuild/blob/94c9c974c2a2d9666fa3fa90c13055c6db9e2a0b/src/Tasks/Microsoft.NETFramework.CurrentVersion.props#L100

mhutch: So putting back `<PlatformTarget>x64</PlatformTarget>` seems appropriate in this case

PR #5111, opened 2018-06-20
Fixes VSTS Bug 631500: [Feedback] EditorConfig doesn't work for VS for  Mac using .NET Core Rules

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/631500


mkrueger:
Test instructions:

This commit removes the opetion panel text editor -> source analysis -> c# -> code rules 
And replaces that with the VS.NET code style panel. 

Which is mostly the same but with some different options. We lose the ability to set the notification level for >all< code rules but it's possible to set them for most. We gain the ability to set more advanced settings and alter code generation styles.

To check: 

Check if the options work  - the preview window contains the test cases. 
Check if the options persist after IDE close.
Check if it's really the same as VS.NET.

Check if the ,editconfig styles are really taken.

My test case was:

.editconfig:
`
[*.cs]
csharp_style_pattern_matching_over_as_with_null_check = true:warning
`
C# file:
`
using System.Threading.Tasks;

class TestClass
{
	public async Task FooBar(object o)
	{
        var m = o as TestClass;
        if (m != null)
            await m.FooBar(o);
	}
}
`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices.Options/MonoDevelopGlobalOptionPersister.cs
```diff
@@ -127,6 +127,18 @@ public bool TryPersist (OptionKey optionKey, object value)
...
+				var propertyName = optionKey.GetPropertyName ();
+				if (propertyName == null) // empty storage location
+					return false;
+				MonitorChanges (storageKey, optionKey);
+				try {
+					var serializedValue = Serialize (value, optionKey.Option.Type);
+					PropertyService.Set (propertyName, serializedValue);
+				} catch (Exception ex) {
+					LoggingService.LogError ($"Failed to serialize key: {storageKey} type: {optionKey.Option.Type}", ex);
+				}
```
mkrueger: @Therzok : The question is here if that's enough even for the property case - IMO the code above is obsolete and that should always work.
But I didn't change it because I wasn't 100% sure. However for really using the option service persistance of all proerties is needed.

Now we should be able to use just the roslyn options for everything :). Maybe adding configuration properties for them - I prefer our approach.

Therzok: A lot of code here is duplicated, please fix that.

We can persist those which have keys, but preferred approach is having ConfigurationProperties surfaced. This is okay for now.

Therzok: On the property case, we have the following problem, documented in the roslyn options doc:

PropertyService.Set does not update a ConfigurationProperty:
https://paper.dropbox.com/doc/Roslyn-integration-API-discussion-nDLFhzV6ZrQBQgHEPWX4W#:uid=055698422129842044857557&h2=Configuration-Property-issues

That's why the Configuration updaters exist, as a workaround to the broken property service/configurationproperty behaviour.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.Rename/SymbolRenamedCodeActionOperationFactoryWorkspaceService.cs
```diff
@@ -0,0 +1,98 @@
...
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeActions.WorkspaceServices;
+using Microsoft.CodeAnalysis.Editor;
+using Microsoft.CodeAnalysis.Host.Mef;
+using MonoDevelop.Core;
+
+namespace MonoDevelop.Refactoring.Rename
+{
+	[ExportWorkspaceService (typeof (ISymbolRenamedCodeActionOperationFactoryWorkspaceService), ServiceLayer.Host), Shared]
+	sealed class SymbolRenamedCodeActionOperationFactoryWorkspaceService : ISymbolRenamedCodeActionOperationFactoryWorkspaceService
```
Therzok: We should file a bug against roslyn that this should be moved into an upper layer:
http://source.roslyn.io/#Microsoft.VisualStudio.LanguageServices/Implementation/Workspace/VisualStudioSymbolRenamedCodeActionOperationFactoryWorkspaceService.cs,19

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.Options/Util.cs
```diff
@@ -0,0 +1,59 @@
...
+				var idx = title.IndexOf (filter, StringComparison.OrdinalIgnoreCase);
+				if (idx >= 0) {
+					string color;
+					if (IdeTheme.UserInterfaceTheme == MonoDevelop.Ide.Theme.Light) {
+						color = "yellow";
+					} else {
+						color = "#666600";
+					}
+
+					title =
```
Therzok: Use a stringbuilder please.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.OptionProvider/StyleViewModel.cs
```diff
@@ -0,0 +1,1287 @@
...
+			CodeStyleItems.Add (new BooleanCodeStyleOptionViewModel (CSharpCodeStyleOptions.PreferPatternMatchingOverIsWithCastCheck, GettextCatalog.GetString ("Prefer pattern matching over 'is' with 'cast' check"), s_preferPatternMatchingOverIsWithCastCheck, s_preferPatternMatchingOverIsWithCastCheck, this, optionSet, expressionPreferencesGroupTitle));
+			CodeStyleItems.Add (new BooleanCodeStyleOptionViewModel (CSharpCodeStyleOptions.PreferPatternMatchingOverAsWithNullCheck, GettextCatalog.GetString("Prefer pattern matching over 'as' with 'null' check"), s_preferPatternMatchingOverAsWithNullCheck, s_preferPatternMatchingOverAsWithNullCheck, this, optionSet, expressionPreferencesGroupTitle));
+			//CodeStyleItems.Add (new BooleanCodeStyleOptionViewModel (CodeStyleOptions.PreferConditionalExpressionOverAssignment, GettextCatalog.GetString("Prefer conditional expression over 'if' with assignments"), s_preferConditionalExpressionOverIfWithAssignments, s_preferConditionalExpressionOverIfWithAssignments, this, optionSet, expressionPreferencesGroupTitle));
```
Therzok: Why disabled?

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.CodeIssues/CodeDiagnosticDescriptor.cs
```diff
@@ -92,46 +79,11 @@ class CodeDiagnosticDescriptor
...
 				foreach (var diagnostic in GetProvider ().SupportedDiagnostics) {
-					if (GetIsEnabled (diagnostic))
-						return true;
+					if (!diagnostic.IsEnabledByDefault)
+						return false;
 				}
-				return false;
+				return true;
 			}
-			set {
```
Therzok: Make the setter a NOP and obsolete.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.CodeIssues/CodeDiagnosticDescriptor.cs
```diff
@@ -65,23 +65,10 @@ class CodeDiagnosticDescriptor
...
 				foreach (var diagnostic in GetProvider ().SupportedDiagnostics) {
-					if (!result.HasValue)
-						result = GetSeverity(diagnostic);
-					if (result != GetSeverity(diagnostic))
-						return null;
-				}
-				return result;
-			}
-
-			set {
```
Therzok: Make the setter a NOP and obsolete.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.Options/Util.cs
mkrueger: makes only sense in an undetermined concatination case not in a fixed size one.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.OptionProvider/StyleViewModel.cs
mkrueger: Compile error.
PreferConditionalExpressionOverAssignment wasn't there - maybe the next version.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.Options/Util.cs
Therzok: Ah, we have a string array, so we'll do one string allocation in the end. Guess the problem relies on how the Markup API is exposed.

Therzok: Although, we still should be able to solve this in a smarter way, by only doing one `Markup.EscapeText` and inserting the coloring spans there.

Therzok: So that means, EscapeText on the filter and search for that:

```
var escaped = Markup.EscapeText (title);
if filter is not empty
    new stringbuilder() // we can preallocate the builder
    sb.Add (escaped, 0, idx);
    sb.Add ("<span bgcolor");
    sb.Add (color);
    sb.Add (">");
    sb.Add (title, idx, filter.Length);
    sb.Add ("</span>);
    sb.Add (title, idx + filter.Length);
    escaped = sb.ToString()
}

title = escaped
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices.Options/MonoDevelopGlobalOptionPersister.cs
mkrueger: Doesn't concern me much - because we've optionset (roslyn side) and configuration properties (monodevelop side). When using roslyn options we should use option sets for now - all the roslyn code we use does.

So ConfigurationProperties are a legacey thing. However would be nice if we can use configuration properties somehow but for the roslyn configuration properties we should base them on the IOptionService and not touch our propertyservice infrastructure anymore.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.Options/Util.cs
mkrueger: It's a cnp helper method which is not a time critical btw. 
As said it's a constant append case and you're optimizing on the wrong end here. StringBuilder is not superior in all append cases in that one it simply doesn't matter.

For example - the compiler can easily use stringbuilders here if it's better in any way - so this is not an optimization case in any way.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices.Options/MonoDevelopGlobalOptionPersister.cs
Therzok: Roslyn properties are persisted via the property service. It should concern you that someone can set the value via PropertyService.Set and you do not get updates for the value itself in IOptionService.

That's why this infrastructure that wraps roslyn options into ConfigurationProperty was made in the first place

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.CodeIssues/CodeDiagnosticDescriptor.cs
mkrueger: No public API - so we can change it.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.CodeIssues/CodeDiagnosticDescriptor.cs
mkrueger: No public API - so we can change it.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.Options/Util.cs
Therzok: Roslyn does not optimize a lot of things...

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices.Options/MonoDevelopGlobalOptionPersister.cs
mkrueger: No because roslyn properties are another "world" - it should never be used using the property service at least I won't. Sooner or later there will be no 'property service'.

We'll be getting more and more code that will use IOptionService and nothing else - the PropertyService is a legacy thing.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.Options/Util.cs
mkrueger: In any case a O(1) operation.

Therzok: Where did you get that from? Concat, especially with non-compile-time constants is O(n). The substrings are also O(n).

In the case of the values passed to concat not being a `string[]`, it does `string.Concat()` on 2 elements at a time.

In this case, it's a matter of allocations, not CPU, as we create 3 strings for substrings, another 3 for escaped text and one for the concatenation, rather than just one string.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices.Options/MonoDevelopGlobalOptionPersister.cs
Therzok: Until IOptionService is part of Platform, not Roslyn, it's not a good idea to use the roslyn component for everything.

If we had a decent split between IDE and Roslyn, there would be no way for code which wasn't using roslyn to use options.

Imo, IOptionService is roslyn and should be used to interact with roslyn, not replace PropertyService.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.Options/Util.cs
mkrueger: That's computer science. It's O(1) because it's a constant set of operations there is no variable N here.
O(7) == O(1)

Do not optimize what a compiler can do esp. not in critical cases. Other than that use the string builder pool for string builders :).
In general optimizations reduce readability of code and are only worth it in areas where they actually do something. The method was cut & pasted from a class that has gone so the code worked every rework of such code that is running and tested need to be justified and unjustified optimizations are no good excuse to maybe introduce errors.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices.Options/MonoDevelopGlobalOptionPersister.cs
mkrueger: I don't say use it for everything >now< but we're getting more and more code that uses the IOptionService so that needs to be working. I say use IOptionService for all roslyn options. 
Doesn't make sense to mix it.

The PropertyService should only be for general things - every C# specific option should be IOptionService since language backends are roslyn world.

Long term we'll be using some OptionService like approach for the VS.NET interopability. Basically all our public API surface should be as close as possiblie to VS.NET and I don't see why the option system should be an exception to that.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.Options/Util.cs
Therzok: I'm not talking computationally, I'm talking allocation-wise. We're allocating too much junk, when we could not. String manipulation has some tricky APIs to get right.

In this case, we create a ton of junk strings. The 3 substrings create 3 temporary junk strings.
EscapeText creates a temporary string.

Also, C# is not really an optimizing compiler. It lets the runtime do the optimizing.

Regardless, I'll fix it in a future PR if it proves to be an issue.

Therzok: http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MarkupUtilities.cs,43

is what we could use here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices.Options/MonoDevelopGlobalOptionPersister.cs
```diff
@@ -127,6 +127,18 @@ public bool TryPersist (OptionKey optionKey, object value)
...
 				}
+			} else {
+				// Non property case
+				var propertyName = optionKey.GetPropertyName ();
+				if (propertyName == null) // empty storage location
+					return false;
+				MonitorChanges (storageKey, optionKey);
+				try {
+					var serializedValue = Serialize (value, optionKey.Option.Type);
+					PropertyService.Set (propertyName, serializedValue);
```
Therzok: This is missing passing the default value.

Therzok: Possible cause of https://github.com/mono/monodevelop/issues/5179 ?

mkrueger: Handled the default value but I don't think it's required.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices.Options/MonoDevelopGlobalOptionPersister.cs
```diff
@@ -127,6 +127,26 @@ public bool TryPersist (OptionKey optionKey, object value)
...
 					LoggingService.LogError ($"Failed to serialize key: {storageKey} type: {optionKey.Option.Type}", ex);
 				}
+			} else {
+				// Non property case
+				var propertyName = optionKey.GetPropertyName ();
+				if (propertyName == null) // empty storage location
+					return false;
+				MonitorChanges (storageKey, optionKey);
+				try {
+					if (optionKey.Option.DefaultValue != null) {
```
Therzok: Still does not handle not serializing null:
https://github.com/mono/monodevelop/commit/57430b3893163cd81801b62da0626a518fc06e44#diff-d2dfde948eb0d602559066d1e14302b4R137

mkrueger: btw. you don't handle that in your code above as well.

mkrueger: That case is handled:

serializedValue = Serialize (value, optionKey.Option.Type);

returns null if value == null so it's the same thing…

mhutch:
For reference here's the VS panel:

image

I see a few differences - VS has a bar about editorconfig at the top, and the table header says "Preferences" instead of "Property". Its also has icons for severity.

Additionally, VS strips away the surrounding boilerplate, and itspreview uses semantic highlighting.

image

image

slluis:
For the record, this PR brings us closer to how VS works, and fixes several issues with our .editorconfig support. However, we are also losing some functionality. None of the rules implemented in the Refactoring Essentials package supports options, so it won't be possible to set the severity or suppress them using the UI. It is also not possible to configure them using a .ruleset file. The only option is to suppress them using a per-project suppress file (it's one of the options shown when right clicking on an issue). This is going to affect users which changed the preferences in the Source Analysis. Those preferences will be lost.

Solving this problem is not trivial, since it will require changes in the Refactoring Essentials package. We don't have an estimation of the cost, we'll do some research on this. In any case we have to decide wether we want to marge this PR as is and fix the linked issue (about which we have several reports), or wait and implement the whole solution.

slluis:
@jmatthiesen see my last comment, forgot to CC you.

mkrueger:
Didn't know about the ruleset files - y that'll work it's a roslyn way of doing things.

We should support rulesets - however rulesets are not part of the UI story. In short: we should do what vs.net does and support what they do, provide UIs with the same options etc.

Therzok:
@mhutch going in and adding options for every RefactoringEssentials analyzer is going to be tedious. Rulesets are independent, you just work on the IDE, this is the simplest way to properly support analyzer configuration.

mkrueger:
I prefer editorconfig as well and having 2 ways of configuring the same thing is a bit awkward design. But we should support what VS.NET supports - it is that simple. :)

RefactoringEssentials needs maintainance badly - we should decide here how we're treating that.

GLjackyvaswani:
Have tested it with the build provided on the https://github.com/mono/monodevelop/pull/5111,
Issue still reproducible, not all the patterns are working.

we have logged 2 new defects for the same. Hence we have marked the bug as QA-failed"

New Defect logged:
BUG 656467 - Editor rules defined under .editorconfig file are not getting applied automatically when user write the code on editor.
BUG 656471 - Quick fix > Use pattern matching is not displayed every time.

Build taken: https://dl.internalx.com/artifacts/devdiv-1883705/VisualStudioForMac-Preview-7.7.0.707.dmg

mhutch:
@mkrueger @jmatthiesen I think we should consider moving Refactoring Essentials to an extension

mkrueger:
656467 - unrelated - requires the indentation & formatting change branch.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.OptionProvider/CSharpDocumentOptionsProvider.cs
```diff
@@ -98,23 +98,20 @@ public DocumentOptions (OptionSet optionSet, ICodingConventionsSnapshot codingCo
...
-
 				if (codingConventionsSnapshot != null) {
 					var editorConfigPersistence = option.Option.StorageLocations.OfType<IEditorConfigStorageLocation> ().SingleOrDefault ();
 					if (editorConfigPersistence != null) {
 						var allRawConventions = codingConventionsSnapshot.AllRawConventions;
 						try {
 							var underlyingOption = underlyingOptions.GetOption (option);
-							if (editorConfigPersistence.TryGetOption (underlyingOption, allRawConventions, option.Option.Type, out value))
-								return true;
+							return editorConfigPersistence.TryGetOption (underlyingOption, allRawConventions, option.Option.Type, out value);
```
Therzok: Why did this change?

Therzok:
This PR has no tests at all 😕

PR #5114, opened 2018-06-20
[Core] Allow loading of SDK style projects without TargetFramework  

mrward:
Handle the missing main property group when loading and saving a
project. Previously a null reference would occur in several places
on loading.

Fixes VSTS #567738 - NullReferenceException loading sdk-style project
with no TargetFramework


mrward:

PR #5116, opened 2018-06-20
Fixes VSTS 634797: Pasted text indents with extra whitespace

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/634797


mkrueger:
I"ve a branch which uses IEditorFormattingService excusively for the typing behavior unfortunately it's not working like VS.NET in some cases - it needs a bit more VS.NET editor infrastructure I guess. 
However works very well in the cnp case.

mrward:

PR #5124, opened 2018-06-20
[Core] Detect circular dependencies between MSBuild imports

mrward:
If project has an import which caused a circular dependency a
stackoverflow exception would be thrown. Circular dependencies are
now detected and an error is reported to the user indicating the
problem.

Fixes VSTS #569959 - attempt to load a project with conflicting
custom SDKs results in StackOverflowException

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
@@ -1220,6 +1220,8 @@ namespace MonoDevelop.Projects.MSBuild
 			if (!File.Exists (file))
 				return;
 
+			context.AddImport (file);
+
 			var pref = LoadProject (context, file);
 			project.ReferencedProjects.Add (pref);
 
@@ -1244,6 +1246,8 @@ namespace MonoDevelop.Projects.MSBuild
 				
 				project.Properties [p.Key] = p.Value;
 			}
+
+			context.RemoveImport (file);
 		}
 
 		void Evaluate (ProjectInfo project, MSBuildEvaluationContext context, MSBuildChoose choose, bool evalItems)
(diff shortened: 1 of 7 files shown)


mrward:

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildEvaluationContext.cs
```diff
@@ -1064,5 +1068,20 @@ public void Dump ()
 			foreach (var v in allProps.OrderBy (s => s))
 				Log.LogMessage (string.Format ($"{v,-30} = {GetPropertyValue (v)}"));
 		}
+
+		/// <summary>
+		/// Registers imports to check for circular dependencies. Once the import has been
+		/// evaluated it should be removed using RemoveImport.
+		/// </summary>
+		public void AddImport (string file)
```
Therzok: The pattern I usually do for things which should be scoped is having the API modeled like:
```
public IDisposable RegisterImport (string file)
{
    // code here
    return new Registration (file);
}

class Registration : IDisposable
{
    public Registration (string file) { // blah }
    public void Dispose () => unregister;
}
```

Therzok: The reasoning for that is that you can wrap the return value in a using statement, and ensure that you don't need to `try { } finally {}` to unregister.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
```diff
@@ -1244,6 +1246,8 @@ void ImportFile (ProjectInfo project, MSBuildEvaluationContext context, MSBuildI
 				
 				project.Properties [p.Key] = p.Value;
 			}
+
+			context.RemoveImport (file);
```
Therzok: Code before this can throw

slluis: I guess it doesn't matter. If there is an exception during evaluation, the evaluation is stopped and the context discarded.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildEvaluationContext.cs
mrward: Not sure it is worth adding a new class here. If the evaluation fails then everything stops. The context is just a temporary object created for the evaluation.

PR #5129, opened 2018-06-21
[Ide] Fix build not being done when file is unsaved

mrward:
With an unsaved file open in the text editor, running the project
would not trigger a build after the file was saved. The problem
was that the file watcher file change event happened after the
FastCheckNeedsBuild was checked so no build occurred. To fix this
the Document now uses the FileService.NotifyFileChanged to force
a file notification to occur on saving the file so the
FastCheckNeedsBuild is updated before checking if a build is needed.


mrward:

Issue #5100, opened 2018-06-19
Debugger tooltip not always displayed


mrward:
VS Mac 7.6 Preview (build 1771) - VSTS build.

Hitting a breakpoint and hovering over a variable from the right shows a different tooltip (normal text editor tooltip) than if you hover over a variable from the left hand side (debugger tooltip). In both cases the debugger tooltip I was expecting the debugger tooltip to be displayed.

![exceptiontooltip](https://user-images.githubusercontent.com/372361/41592111-80cd8728-73b3-11e8-9b51-e2d8898bb73e.gif)

VS Mac 7.5.2 seems OK.

> VS bug [#635513](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/635513)

xamarin-release-manager:
Fixed in version 7.7.0.389 (master)

Pull Request #5132 merged by: Lluis Sanchez
Author: mono
Commit: 7273f7b933effb3ef87142488bf8ffd4457b2e57 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 7.6.0.1860 (release-7.6)

Pull Request #5236 merged by: Lluis Sanchez
Author: mono
Commit: 226e7a935cbb435e8a02f5c33a26045332430821 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #5132 merged by: Lluis Sanchez
Author: mono
Commit: 7273f7b933effb3ef87142488bf8ffd4457b2e57 (xamarin/monodevelop)

xamarin-release-manager:
> [sync] [VS-28] Comment by **xamarinc**



Fixed in version 8.6.0.802 (master)



Author: Mike Kr??ger  
 Commit: 51f8edcd50052bc3af7afb4adca6bbd93d12b07b (xamarin/monodevelop)

PR #5132, opened 2018-06-21
Fixes issue #5100 Debugger tooltip not always displayed

mkrueger:
We shouldn't have to implement that service - it's not using anything special. So I request to move it.

https://github.com/dotnet/roslyn/issues/28041

For now copying the code 1:1 works.


main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/CSharpCompletionTextEditorExtension.cs
```diff
@@ -757,7 +757,18 @@ public Task<MonoDevelop.Ide.CodeCompletion.ParameterHintingResult> InternalHandl
 
 		async Task<DebugDataTipInfo> IDebuggerExpressionResolver.ResolveExpressionAsync (IReadonlyTextDocument editor, DocumentContext doc, int offset, CancellationToken cancellationToken)
 		{
-			return await Resolver.DebuggerExpressionResolver.ResolveAsync (editor, doc, offset, cancellationToken).ConfigureAwait (false);
+			var analysisDocument = doc.AnalysisDocument;
+			if (analysisDocument == null)
+				return default (DebugDataTipInfo);
+			var debugInfoService = doc.RoslynWorkspace.Services.GetLanguageServices (LanguageNames.CSharp).GetService<Microsoft.CodeAnalysis.Editor.Implementation.Debugging.ILanguageDebugInfoService> ();
```
Therzok: We could use this: http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces/Shared/Extensions/DocumentExtensions.cs,23

slluis:
@mrward can you verify the fix?

mrward:
Partially fixed. Hovering over exception variables inside the catch statement all work as expected.

If you hover over the catch line near the right hand side of the 'ex' you see the local variable tooltip instead of the debugger tooltip. If you hover over the left hand side you see the debugger tooltip. In VS Mac 7.5 the debugger tooltip is shown unless you hover over the ')' at the end of the catch. Not tried VS 2017 here.

debuggertooltip2

mrward:
VS 2017's behaviour seems to the same as VS Mac 7.5.

slluis:
@mkrueger did you change something? What's the status of this?

mkrueger:
Should work IMO

slluis:

slluis:
@mkrueger can you backport to 7.6? thanks.

Issue #5161, opened 2018-06-22
CopyCommentsFromBaseCodeRefactoringProvider.ComputeRefactoringsAsync XML parse exception


KirillOsenkov:
1. Start MonoDevelop, open new C# console app
2. Place caret on class name

Exception bar is shown: XmlException there are multiple root elements

![image](https://user-images.githubusercontent.com/679326/41752289-7bc5477e-757b-11e8-9990-be1896c86e30.png)

```
 	System.Xml.dll!System.Xml.XmlTextReaderImpl.Throw Line 2501	C#
 	System.Xml.dll!System.Xml.XmlTextReaderImpl.Throw Line 2477	C#
 	System.Xml.dll!System.Xml.XmlTextReaderImpl.ParseDocumentContent Line 3644	C#
 	System.Xml.dll!System.Xml.XmlTextReaderImpl.Read Line 1166	C#
 	System.Xml.Linq.dll!System.Xml.Linq.XContainer.ReadContentFrom Line 3083	C#
 	System.Xml.Linq.dll!System.Xml.Linq.XContainer.ReadContentFrom Line 3193	C#
 	System.Xml.Linq.dll!System.Xml.Linq.XDocument.Load Line 5730	C#
 	System.Xml.Linq.dll!System.Xml.Linq.XDocument.Parse Line 5779	C#
 	System.Xml.Linq.dll!System.Xml.Linq.XDocument.Parse Line 5751	C#
>	RefactoringEssentials.dll!RefactoringEssentials.CSharp.CodeRefactorings.CopyCommentsFromBaseCodeRefactoringProvider.ComputeRefactoringsAsync Line 45	C#
 	Microsoft.CodeAnalysis.Features.dll!Microsoft.CodeAnalysis.CodeRefactorings.CodeRefactoringService.GetRefactoringFromProviderAsync Line 142	C#
 	Microsoft.CodeAnalysis.Features.dll!Microsoft.CodeAnalysis.CodeRefactorings.CodeRefactoringService.GetRefactoringsAsync.AnonymousMethod__0 Line 105	C#

```

XML snippet in question:
```
      <summary>Supports all classes in the .NET Framework class hierarchy and provides low-level services to derived classes. This is the ultimate base class of all classes in the .NET Framework; it is the root of the type hierarchy.To browse the .NET Framework source code for this type, see the Reference Source.</summary>
      <filterpriority>1</filterpriority>
```

[...]

PR #5164, opened 2018-06-22
Fixes issue CopyCommentsFromBaseCodeRefactoringProvider.ComputeRefactoringsAsync XML parse exception #5161

mkrueger:
Fixes issue #5161 CopyCommentsFromBaseCodeRefactoringProvider.ComputeRefactoringsAsync XML parse exception


Therzok:
Prefer `string.Join(string, string[])` overload by using a different XML reader.

mkrueger:
Best thing would've been indexOf('>') .... lastIndexOf ('<') - y I also found it a bit weird what they did - for stripping the root node I woudn't have used any xml reader or join operations.

KirillOsenkov:
I wish there was a way to avoid a first-chance exception for malformed XML. It happens pretty often unfortunately. I wonder if RefactoringEssentials has access to the MonoDevelop's Xml Parser? I'd guess not? And it's probably not worth introducing a new dependency just to fix this (such as https://github.com/KirillOsenkov/XmlParser).

mkrueger:
Refactoring essentials has an own very capable xml parser.

PR #5167, opened 2018-06-22
[Telemetry] Protect the GetLoginTime

iainx:
Wrap GetLoginTime and related code in a try catch so that if there's an issue
the details still get set up correctly

Fixes VSTS #637410
FIxes VSTS #637958

### DIFF
--- a/main/src/addins/MacPlatform/MacTelemetryDetails.cs
+++ b/main/src/addins/MacPlatform/MacTelemetryDetails.cs
@@ -66,10 +66,10 @@ namespace MacPlatform
 
 			try {
 				var login = GetLoginTime ();
-				if (login != DateTime.MinValue) {
-					var epoch = new DateTime (1970, 1, 1, 0, 0, 0, 0, DateTimeKind.Utc);
+				if (login != DateTimeOffset.MinValue) {
+					var epoch = new DateTimeOffset (1970, 1, 1, 0, 0, 0, 0, TimeSpan.Zero);
 
-					var timeSinceEpoch = DateTime.UtcNow - epoch;
+					var timeSinceEpoch = DateTimeOffset.UtcNow - epoch;
 					var loginSinceEpoch = login - epoch;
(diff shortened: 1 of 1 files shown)


iainx:
the crash was the exception. Normally the MacTelemetryDetails is created at startup, but when the exception happens it doesn't get created and typing keeps trying to create it and the exception happened again and crashed.

slluis:
So the exception is `System.ArgumentOutOfRangeException: The UTC time represented when the offset is applied must be between year 0 and 10,000.`. Can't we fix that?

iainx:
We could, if I could find out why it was returning a broken time, but I've not been able to reproduce it. @Therzok was able to, but hasn't for a while.

iainx:
This is more of a bandaid to stop fatal crashes than a full fix

main/src/addins/MacPlatform/MacTelemetryDetails.cs
```diff
@@ -63,15 +64,19 @@ internal static MacTelemetryDetails CreateTelemetryDetails ()
 
 			result.osType = GetMediaType ("/");
 
-			var login = GetLoginTime ();
-			if (login != DateTime.MinValue) {
-				var epoch = new DateTime (1970, 1, 1, 0, 0, 0, 0, DateTimeKind.Utc);
+			try {
+				var login = GetLoginTime ();
+				if (login != DateTime.MinValue) {
```
Therzok: I think just using `DateTimeOffset` everywhere instead of `DateTime` will fix the problem.

Therzok: Which is what I had locally when I said I couldn't reproduce.

iainx: Yeah, It might, but I'd like to know what was causing the original version to fail before just going with a change to `DateTimeOffset` because it seems like the value should fit between the years 0 and 10,000 whether or not it's a `DateTime` or a `DateTimeOffset`

slluis:

PR #5185, opened 2018-06-25
[PerformanceDiagnostics] Add UI hang detection for telemetry

mrward:
The UI thread monitor console application now has a way to check for
the IDE's UI thread hanging without profiling the process. It will
create a file to indicate a hang is occurring. This file will be
used by the IDE when reporting the previous shutdown reason.

Needs md-addins changes.

Fixes VSTS #619945

Not implemented:

 - [x] UI Thread console app connection retry. If the connection back to the IDE is lost the console app currently exits. I think it should retry the connection until the IDE process exits. The IDE will terminate the console app itself.

### DIFF
--- a/main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/UIThreadMonitor.cs
+++ b/main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/UIThreadMonitor.cs
@@ -20,7 +20,19 @@ namespace PerformanceDiagnosticsAddIn
 	{
 		public static UIThreadMonitor Instance { get; } = new UIThreadMonitor ();
 
-		UIThreadMonitor () { }
+		UIThreadMonitor ()
+		{
+			IdeApp.Exited += IdeAppExited;
+		}
+
+		void IdeAppExited (object sender, EventArgs e)
+		{
+			try {
+				Instance.Stop ();
+			} catch (Exception ex) {
+				LoggingService.LogError ("UIThreadMonitor stop error.", ex);
+			}
+		}
 
 		Thread tcpLoopThread;
 		Thread dumpsReaderThread;
--- a/main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/UIThreadMonitor.cs
+++ b/main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/UIThreadMonitor.cs
@@ -144,7 +144,7 @@ namespace PerformanceDiagnosticsAddIn
 				dumpsReaderThread = new Thread (new ParameterizedThreadStart (DumpsReader));
 				dumpsReaderThread.IsBackground = true;
 				dumpsReaderThread.Start (process);
-				PumpErrorStream (process).Ignore ();
+				Task.Run (() => PumpErrorStream (process)).Ignore ();
 			}
 		}
 
--- a/main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/UIThreadMonitor.cs
+++ b/main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/UIThreadMonitor.cs
@@ -11,6 +11,7 @@ using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using System.Globalization;
 using System.Collections.Generic;
+using System.Text;
 using System.Threading.Tasks;
 
 namespace PerformanceDiagnosticsAddIn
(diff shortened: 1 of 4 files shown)


main/src/addins/PerformanceDiagnostics/UIThreadMonitorDaemon/Program.cs
```diff
@@ -41,37 +42,95 @@ public class MainClass
...
-					StartCollectingStacks ();
-					if (!responseEvent.WaitOne (10000))
-						Console.Error.WriteLine ($"No response({seq}) in 10sec");
-					else
+					if (sample)
+						StartCollectingStacks ();
+
+					int waitResult = WaitHandle.WaitAny (events, 10000);
+					if (waitResult == 0) // Got a response.
 						Console.Error.WriteLine ($"Response({seq}) in {sw.Elapsed}");
```
slluis: It is not necessary to send messages through Console.Error when not sampling.

main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/UIThreadMonitor.cs
```diff
@@ -86,61 +106,87 @@ public void Profile (int seconds)
...
-			dumpsReaderThread.Start ();
 
-			pumpErrorThread = new Thread (new ThreadStart (PumpErrorStream));//We need to read this...
+			if (IsSampling) {
+				dumpsReaderThread = new Thread (new ParameterizedThreadStart (DumpsReader));
+				dumpsReaderThread.IsBackground = true;
+				dumpsReaderThread.Start (process);
+			}
+
+			pumpErrorThread = new Thread (new ParameterizedThreadStart (PumpErrorStream));//We need to read this...
```
slluis: Instead of starting a thread just for pumping the error stream, start an async method that uses ReadLineAsync() to read the error messages in a loop.

main/src/addins/PerformanceDiagnostics/UIThreadMonitorDaemon/Program.cs
```diff
@@ -108,13 +167,13 @@ static void Loop (object portObj)
...
 				socket.Send (new byte [1] { ++seq });
 				responseEvent.Reset ();
 				sentEvent.Set ();
 				var readBytes = socket.Receive (response, 1, SocketFlags.None);
 				if (readBytes != 1) {
-					disonnected = true;
+					disconnected = true;
 					disconnectEvent.Set ();
 					return;
 				}
```
slluis: This issue: https://github.com/mono/monodevelop/blob/cfa8c4ddeea3e720dbf8b8e6a283e8083f7cd09d/main/src/addins/PerformanceDiagnostics/UIThreadMonitorDaemon/Program.cs#L181

should not be handled with an exception, since it will crash the monitor process.

slluis:
Using a command handler for starting the monitor is not very nice. Why not just make UIThreadMonitor public and call the Start method at startup?

mrward:
Currently the CrashDetector in md-addins starts the thread monitor and passes the session id information to it. So we could have the CrashDetector directly using the PerformanceDiagnostics addin's ThreadMonitor.

main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/UIThreadMonitor.cs
```diff
@@ -86,61 +108,87 @@ public void Profile (int seconds)
...
-
-			pumpErrorThread = new Thread (new ThreadStart (PumpErrorStream));//We need to read this...
-			pumpErrorThread.IsBackground = true;
-			pumpErrorThread.Start ();
+
+			if (IsSampling) {
+				dumpsReaderThread = new Thread (new ParameterizedThreadStart (DumpsReader));
+				dumpsReaderThread.IsBackground = true;
+				dumpsReaderThread.Start (process);
+				PumpErrorStream (process).Ignore ();
```
Therzok: I'd rather do a `Task.Run()` here.

We cannot guarantee that the underlying implementation will not do a blocking sync call on `ReadLineAsync()` (`TextReader` implementation does `Task.FromResult(ReadLine())`, so we'd effectively block until the first task is returned.

mrward: I was thinking about using a Task.Run initially 😃

main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/UIThreadMonitor.cs
```diff
@@ -16,7 +16,7 @@
 
 namespace PerformanceDiagnosticsAddIn
 {
-	class UIThreadMonitor
+	public class UIThreadMonitor
```
Therzok: Why public?

PR #5193, opened 2018-06-25
Avoid churn in the sln file on Windows

mhutch:


Therzok:
We lose [this](https://github.com/mono/monodevelop/blob/master/main/src/addins/MonoDevelop.Gettext/MonoDevelop.Gettext.NodeBuilders/TranslationProjectNodeBuilder.cs) capability, but I don't think it's a big issue.

Maybe we should also include the translation files in the `<None>` category?

mhutch:
We don't need to add them as `None`, the `GettextTranslation` items show up in the solution tree anyways.

PR #5194, opened 2018-06-26
Simplify resource ids

mhutch:
Simplify resource id definitions using a definition group.

### DIFF
--- a/main/external/fsharpbinding/MonoDevelop.FSharp.Gui/MonoDevelop.FSharp.Gui.csproj
+++ b/main/external/fsharpbinding/MonoDevelop.FSharp.Gui/MonoDevelop.FSharp.Gui.csproj
@@ -46,7 +46,6 @@
   </ItemGroup>
   <ItemGroup>
     <EmbeddedResource Include="gtk-gui\gui.stetic">
-      <LogicalName>gui.stetic</LogicalName>
       <SubType>Designer</SubType>
     </EmbeddedResource>
   </ItemGroup>
--- a/main/msbuild/MonoDevelop.AfterCommon.props
+++ b/main/msbuild/MonoDevelop.AfterCommon.props
@@ -29,4 +29,10 @@
 		<AssemblyName>$(MSBuildProjectName)</AssemblyName>
 		<RootNamespace>$(MSBuildProjectName)</RootNamespace>
 	</PropertyGroup>
+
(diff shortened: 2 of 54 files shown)


main/msbuild/MonoDevelop.AfterCommon.props
```diff
@@ -29,4 +29,10 @@
 		<AssemblyName>$(MSBuildProjectName)</AssemblyName>
 		<RootNamespace>$(MSBuildProjectName)</RootNamespace>
 	</PropertyGroup>
+
+	<ItemDefinitionGroup Condition="'$(ShortEmbeddedResourceIDs)'!='False'">
+		<EmbeddedResource>
```
Therzok: Nice!

What about cases where the subtype is set?

mhutch: That should still work! This just adds metadata.

Also, I'm not sure what exactly the subtype is for - VS adds it in some cases, but it isn't used by MSBuild at all.

main/external/fsharpbinding/MonoDevelop.FSharp.Gui/MonoDevelop.FSharp.Gui.csproj
```diff
@@ -45,10 +45,7 @@
     <Reference Include="glib-sharp, Version=2.12.0.0, Culture=neutral, PublicKeyToken=35e10195dab3c99f" />
   </ItemGroup>
   <ItemGroup>
-    <EmbeddedResource Include="gtk-gui\gui.stetic">
-      <LogicalName>gui.stetic</LogicalName>
-      <SubType>Designer</SubType>
```
Therzok: @mhutch you deleted subtype here.

Therzok: I guess it's used as a viewer hint?

mhutch: It's back :)

slluis:
I'm not sure VSMac can handle this. Does roundtripping work?
Also, VSMac will add the resource name when new resources are added, unless we change the naming policy.

mhutch:
@slluis I confirmed roundtripping works, and update the policy

PR #5195, opened 2018-06-26
Fix window launch glitch

mhutch:
Fix glitches on Windows where the window would resize immediately after startup. Also cleans up some dead code.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
@@ -80,7 +80,6 @@ namespace MonoDevelop.Ide.Gui
 
 		bool closeAll;
 
-		bool fullscreen;
 		Rectangle normalBounds = new Rectangle(0, 0, MinimumWidth, MinimumHeight);
 		
 		Gtk.Container rootWidget;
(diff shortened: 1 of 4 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
```diff
@@ -673,7 +670,6 @@ public Properties GetStoredMemento (ViewContent content)
 						Iconify ();
 					}
 					//GdkWindow.State = memento.WindowState;
-					FullScreen = memento.FullScreen;
```
slluis: Can't we make this work on Mac?

Therzok: It should work on mac.

Therzok: Also, agreed, maybe we should not do this on windows only.

mhutch:
Is restoring fullscreen state reasonable on Mac?

The code was broken already because it didn't save the fullscreen state, it only restored it.

slluis:
Xcode saves and restores full screen mode, so I guess it is reasonable to do it.

mhutch:
Ah ok! I didn't have a Mac handy to check.

PR #5196, opened 2018-06-26
[StructuredOutput] Fixes rounding rendering duration time

netonjm:
This PR fixes rendendering in diagnostic and normal mode

Removed the old format in diagnostic 00:02.748 to use

2.341s rounding 1 decimal in case de normal and 3 decimal
in case of diagnostic.

Example: 0.94 displays as 0.9, 0.95 displays as 1.0

Before | After
------------ | -------------
00:00.001 | .001s
00:02.341 | 2.341s
1:50.211 | 1m 50.001s

Fixes Bug #636828
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/636828

Fixes Bug #636830
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/636830

screen shot 2018-07-16 at 20 36 46

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
@@ -372,24 +372,27 @@ namespace MonoDevelop.Ide.BuildOutputView
 			return null;
 		}
 
+		const int NormalRoundPrecision = 1;
+		const int DiagnosticRoundPrecision = 3;
 		public static string GetDurationAsString (this BuildOutputNode node, bool includeDiagnostics)
 		{
 			var duration = node.EndTime.Subtract (node.StartTime);
-			if (includeDiagnostics) {
-				if (duration.TotalHours >= 1) {
-					return string.Format ("{0,12}", duration.ToString (@"hh\:mm\:ss\.fff"));
-				} 
-				return string.Format ("{0,12}", duration.ToString (@"mm\:ss\.fff"));
-			} 
-
 			if (duration.TotalHours >= 1) {
 				return string.Format ("{0,7}", GettextCatalog.GetString ("{0}h {1}m", duration.Hours.ToString(), duration.Minutes.ToString ("00")));
-			} 
+			}
 
 			if (duration.TotalMinutes >= 1) {
 				return string.Format ("{0,7}", GettextCatalog.GetString ("{0}m {1}s", duration.Minutes.ToString(), duration.Seconds.ToString ("00")));
-			} 
-			return string.Format ("{0,7}", GettextCatalog.GetString ("{0}s", duration.Seconds.ToString ()));
+			}
+
+			var precision = includeDiagnostics ? DiagnosticRoundPrecision : NormalRoundPrecision;
+			var value = Math.Round ((duration.Seconds + duration.Milliseconds / 1000d), precision);
+
+			//We don't want print 0 values
+			if (value == 0) {
+				return null;
+			}
+			return value.ToString ("F" + precision) + "s";
 		}
 
 		static void ToString (this BuildOutputNode node, bool includeChildren, StringBuilder result, string margin)
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
@@ -374,6 +374,9 @@ namespace MonoDevelop.Ide.BuildOutputView
 
 		const int NormalRoundPrecision = 1;
 		const int DiagnosticRoundPrecision = 3;
+		const string NormalRoundPrecisioFormat = "{0:F1}s";
+		const string DiagnosticRoundPrecisionFormat = "{0:F3}s";
+
 		public static string GetDurationAsString (this BuildOutputNode node, bool includeDiagnostics)
 		{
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
```diff
@@ -388,8 +388,8 @@ public static string GetDurationAsString (this BuildOutputNode node, bool includ
 
 			if (duration.TotalMinutes >= 1) {
 				return string.Format ("{0,7}", GettextCatalog.GetString ("{0}m {1}s", duration.Minutes.ToString(), duration.Seconds.ToString ("00")));
-			} 
-			return string.Format ("{0,7}", GettextCatalog.GetString ("{0}s", duration.Seconds.ToString ()));
+			}
+			return duration.Seconds != 0 ? string.Format ("{0,7}", GettextCatalog.GetString ("{0}s", duration.Seconds.ToString ())) : "";
```
Therzok: Can we not have a ternary here? Makes it hard to follow.

Also, I would shortcircuit this and put it at the top.

```
if (duration.TotalSeconds == 0)
    return string.Empty;
```

rodrmoya:
We already tried this, not showing times for items that lasted < 1s, so not sure we want to re-add it, as it was quite confusing, according to Maryanne

Therzok:
This only hides for 0s, not < 1s.

netonjm:
We don't work with decimal values in seconds, then <1 in our case is 0

rodrmoya:
I mean hiding time, not just < 1s, 0s, just hiding time for some rows and not for others was what was confusing, but anyway, let's see what Maryanne says

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
```diff
@@ -388,7 +388,12 @@ public static string GetDurationAsString (this BuildOutputNode node, bool includ
...
 			if (duration.TotalMinutes >= 1) {
 				return string.Format ("{0,7}", GettextCatalog.GetString ("{0}m {1}s", duration.Minutes.ToString(), duration.Seconds.ToString ("00")));
-			} 
+			}
+
+			if (duration.TotalSeconds < 1) {
+				return string.Empty;
+			}
+
 			return string.Format ("{0,7}", GettextCatalog.GetString ("{0}s", duration.Seconds.ToString ()));
```
mhutch: FYI the ToString() here is redundant.

mhutch: Also, this is wrong. It should round, not truncate.

Therzok: The `ToString()` is not redundant. It avoids a box of the seconds field.

mhutch:
There's a lot of format strings that could be cached here. Also, why do the normal and diagnostic views use different formats/codepaths? Diagnostic view is not rounding.

netonjm:
It was introduced because this Task: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/571671

mohakbarokar:
Verified on Visual Studio Enterprise 2017 for Mac (Preview) Version 7.7 Preview (7.7 build 315).
636828 Issue is still reproducible. Kindly find screencast. Hence marked issue as Reproduced.
Build Gist: https://gist.github.com/mohakbarokar/bd36754a29f604a811bce2c20f793b8f
Screencast: https://www.screencast.com/t/6PVS0jBCv
Build Taken From: https://github.com/mono/monodevelop/pull/5196/commits

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
```diff
@@ -376,10 +376,11 @@ public static string GetDurationAsString (this BuildOutputNode node, bool includ
 		{
 			var duration = node.EndTime.Subtract (node.StartTime);
 			if (includeDiagnostics) {
+				const int DiagnosticRoundPrecision = 3;
```
Therzok: `const string diagnosticRoundPrecision = "fff"`

This will make all the string concatenations below not allocate, for either the precision string or the format string.

netonjm:
PR is ready to review

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutputNode.cs
```diff
@@ -372,24 +372,27 @@ public static BuildOutputNode SearchFirstNode (this BuildOutputNode buildOutputN
...
+			}
+
+			var precision = includeDiagnostics ? DiagnosticRoundPrecision : NormalRoundPrecision;
+			var value = Math.Round ((duration.Seconds + duration.Milliseconds / 1000d), precision);
+
+			//We don't want print 0 values
+			if (value == 0) {
+				return null;
+			}
+			return value.ToString ("F" + precision) + "s";
```
Therzok: Still allocates more than needed.

```
const string NormalRoundPrecision = "{0:F1}s";
const string DiagnosticRoundPrecision = "{0:F3}s";

...

var precision = includeDiagnostics ? DiagnosticRoundPrecision : NormalRoundPrecision;
return string.Format (precision, value);
```

slluis:

PR #5199, opened 2018-06-26
[Roslyn] Fix show stacktrace not showing the exception

### LINKED ISSUE #5170: Show stack trace in exception bar does not show display any information
Using the repro in #5169 you should be able to generate the exception bar message. Clicking on the Show Stack Trace does not seem to do anything.

If there is nothing to be displayed then the Show Stack Trace should not be visible. Not sure if this is the case here.

VS Mac 7.6.0.1797

> VS bug [#637462](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/637462)

Therzok:
Fixes #5170 - Show stack trace in exception bar does not show display any information


slluis:

PR #5206, opened 2018-06-26
Alt enter timing

iainx:


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
```diff
@@ -1017,35 +1024,50 @@ public static void Stop ()
...
 		{
+			var metadata = new DebuggerActionMetadata {
+				Type = DebuggerActionMetadata.ActionType.StepInto
+			};
+			Counters.DebuggerAction.BeginTiming (metadata);
 			Runtime.AssertMainThread ();
 
 			if (!IsDebugging || !IsPaused || CheckIsBusy ())
 				return;
 			currentSession.Session.StepLine ();
```
Therzok: I think we need to begin timing before this line?

Therzok: Otherwise, we'll never `EndTiming` if the condition above exits the method.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/RootWorkspace.cs
```diff
@@ -532,11 +533,21 @@ public Task<bool> OpenWorkspaceItemInternal (FilePath file, bool closeCurrent, b
...
+				Counters.TimeToCode.Trace ("Started opening solution");
+			}
 			try {
 				var oper = BackgroundLoadWorkspace (monitor, file, loadPreferences, reloading, metadata, timer);
 				return await oper;
 			} finally {
 				timer.End ();
+
+				if (!calculatedTimeToCode) {
+					Counters.TimeToCode.EndTiming ();
```
Therzok: Should time to code include the time it takes to load the typesystem?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
```diff
@@ -677,6 +677,12 @@ public static void ShowDisassembly ()
 
 		internal static ProcessAsyncOperation InternalRun (ExecutionCommand cmd, DebuggerEngine factory, OperationConsole c)
 		{
+			// Start assuming success, update on failure
+			var metadata = new DebuggerStartMetadata {
+				Result = Core.Instrumentation.CounterResult.Success
+			};
+			Counters.DebuggerStart.BeginTiming (metadata);
```
slluis: The debugger can handle several sessions at once, so you can't rely on calling EndTiming() on the static Counters.DebuggerStart. Instead, store the TimeCounter returned by BeginTiming into the SessionManager created below in this method.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.CodeActions/CodeActionEditorExtension.cs
```diff
@@ -363,15 +364,26 @@ void CurrentSmartTagPopup ()
...
+					metadata.TriggeredBySmartTag = true;
+					currentSmartTagBegin = Editor.CaretOffset;
+					PopupQuickFixMenu (null, null);
+					return;
+				}
+
+				metadata.TriggeredBySmartTag = false;
+
+				CancelSmartTagPopupTimeout ();
+				PopupQuickFixMenu (null, menu => { });
```
slluis: `PopupQuickFixMenu` is an async method. You should end timing once the async method has completed. In any case, maybe it is better to measure the time inside `PopupQuickFixMenu`.

slluis: You also need to measure quick fix menu time here: http://source.monodevelop.com/#MonoDevelop.Refactoring/MonoDevelop.Refactoring/QuickFixMenuHandler.cs,47

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/RootWorkspace.cs
```diff
@@ -532,11 +533,21 @@ public Task<bool> OpenWorkspaceItemInternal (FilePath file, bool closeCurrent, b
 			IdeApp.Workbench.LockGui ();
 			metadata = GetOpenWorkspaceItemMetadata (metadata);
 			ITimeTracker timer = Counters.OpenWorkspaceItemTimer.BeginTiming (metadata);
+
+			// Add a trace here to indicate how long the user allowed the app to sit before starting the solution load process
```
slluis: What we want to measure is the time it takes from double clicking on a solution or file in Finder until the solution or file is loaded. So this metric should not be reported if the user starts VSMac and then opens a solution.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/RootWorkspace.cs
iainx: No. It was time until you could open a file and type

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/RootWorkspace.cs
```diff
@@ -532,11 +532,25 @@ public Task<bool> OpenWorkspaceItemInternal (FilePath file, bool closeCurrent, b
...
+				// only update the counter if we need to track the TTC
+				if (IdeApp.NeedToSubmitTimeToCode && IdeApp.ReportTimeToCode) {
+					Counters.TimeToCode.Inc ("SolutionLoaded", ttcMetadata);
+					IdeApp.ReportTimeToCode = false;
+				}
+
+				// If we've loaded a solution from inside the application, we don't need to report TimeToCode
+				IdeApp.NeedToSubmitTimeToCode = false;
+				IdeApp.ReportTimeToCode = false;
+
```
slluis: This logic does not belong to here. In IdeStartup you can subscribe the `IdeApp.Workspace.FirstWorkspaceItemRestored` event and do the above in the handler.

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -598,6 +598,7 @@ void GlobalSetup ()
 				ApplicationEvents.OpenDocuments += delegate (object sender, ApplicationDocumentEventArgs e) {
 					//OpenFiles may pump the mainloop, but can't do that from an AppleEvent, so use a brief timeout
 					GLib.Timeout.Add (10, delegate {
+						IdeApp.ReportTimeToCode = true;
```
slluis: What if the user starts the IDE, and after a while it double clicks on a file in Finder? Won't that cause TimeToCode to be incorrectly reported?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
```diff
@@ -961,14 +996,22 @@ static void OnDisableConditionalCompilation (DocumentEventArgs e)
...
 				if (PausedEvent != null)
 					PausedEvent (null, EventArgs.Empty);
 				NotifyLocationChanged ();
 				IdeApp.Workbench.GrabDesktopFocus ();
+
+			}).ContinueWith ((arg) => {
+				// PausedEventHandlers may queue additional UI events that can cause a freeze.
+				// Ensure those UI events have completed before we stop tracking the time.
+				Runtime.RunInMainThread (() => {
+					sessionManager.ActionTimeTracker.Dispose ();
```
slluis: You have to check for `sessionManager.TrackActionTelemetry` here, otherwise `sessionManager.ActionTimeTracker` will be null and this call will crash.

main/src/addins/MacPlatform/MacPlatform.cs
slluis: Remove this?

slluis: Ignore my last comment.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebugExecutionHandlerFactory.cs
```diff
@@ -74,6 +74,8 @@ private void TargetReady(object sender, EventArgs e)
 		{
 			session.TargetReady -= TargetReady;
 			ProcessId = (int)(session.GetProcesses().FirstOrDefault()?.Id ?? 0);
+
+			Counters.DebuggerStart.EndTiming ();
```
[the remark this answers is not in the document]
slluis: Remove this?

iainx:
I think the TimeToCode was a bug in the md-addins patch, it should be fixed now, just waiting for a build to try to test. Looking at the debugger issue

PR #5210, opened 2018-06-27
[StructuredOutput] Creates views for each nodeview type with managed draw logic

netonjm:
This PR makes easy work with each type of view creating subtypes of views and sending to it the draw handling


Therzok:
What was the benefit of this?

netonjm:
this was a initial step to create different types of row views in structured build output and each view manage it's own logic

PR #5225, opened 2018-06-29
Fixes Bug 627039: Editor occasionally becomes unresponsive to user

mkrueger:
input

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/627039

Can't reproduce the issue however that patch ensures that the edit
operation in the main buffer works - a projection buffer could not be
updated but will be corrected on the next parse event.


mkrueger:

PR #5247, opened 2018-07-01
[Roslyn] Fix serialization and deserialization issues

Therzok:
Punts at serialization/deserialization issues, both in implementation and being out of sync with roslyn. See respective commits for issues they fix.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Properties.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Properties.cs
@@ -50,10 +50,10 @@ namespace MonoDevelop.Core
 			}
 		}
 
-		T Convert<T> (object o)
+		T Convert<T> (object o, Type type)
 		{
-			TypeConverter converter = GetConverter (typeof(T));
-			
+			TypeConverter converter = GetConverter (type);
+
 			if (o is string) {
 				try {
 					return (T)converter.ConvertFromInvariantString (o.ToString ());
@@ -86,6 +86,18 @@ namespace MonoDevelop.Core
 			}
 			return converter;
 		}
+
+		public object Get (string property, object defaultValue, Type type)
+		{
+			if (!defaultValues.ContainsKey (property))
+				defaultValues = defaultValues.SetItem (property, defaultValue);
+
+			if (GetPropertyValue (property, out object value, type))
+				return value;
+			properties = properties.SetItem (property, defaultValue);
+			return defaultValue;
+		}
+
 		
 		public T Get<T> (string property, T defaultValue)
 		{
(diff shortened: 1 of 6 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Core/Properties.cs
```diff
@@ -106,16 +118,30 @@ public T Get<T> (string property)
...
 				return false;
 			}
 
-			if (o is T t) {
-				val = t;
-				return true;
+			if (type != null) {
+				if (type.IsInstanceOfType (o)) {
+					val = (T)o;
+					return true;
```
mrward: Maybe this it is just me but reading this method on its own seems confusing. This cast will work since if you look at where the GetPropertyValue method is used when the type parameter is passed the T type will be an object so the (T)o cast will work. Just looking at the method on its own it just seems like T and Type could be incompatible and this cast would not work.

Also seems confusing having a generic method that also takes a non-generic Type as a parameter.

I can see this to allow this method to be re-used when the property is being read as an object.

I sort of what to indicate that T should be a base class of Type type. Something like the following which I don't think would work if the Type type parameter is null.

```
bool GetPropertyValue<T, U> (string property, out T val, U type = null)
where U : T
```

Or maybe we just live with a bit of duplication and have one GetPropertyValue that is generic and one that returns an object and takes a Type?

Therzok: That works too, I just thought of a way to not duplicate code. I guess it's not clear enough. I'll refactor.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices.Options/OptionsExtensions.cs
```diff
@@ -33,6 +34,20 @@ namespace MonoDevelop.Ide.RoslynServices.Options
...
+				yield break;
+
+			foreach (var storageLocation in optionKey.Option.StorageLocations) {
+				if (storageLocation is RoamingProfileStorageLocation roamingLocation)
+					yield return roamingLocation.GetKeyNameForLanguage (optionKey.Language);
+				if (storageLocation is LocalUserProfileStorageLocation userLocation)
+					yield return userLocation.KeyName;
+			}
+		}
+
```
mkrueger: Looks like code duplication of GetPropertyName - can you reuse GetPropertyName here ?

main/src/core/MonoDevelop.Core/MonoDevelop.Core/Properties.cs
Therzok: Fixed it by keeping the `T` at API level, but using objects internally. Makes more sense, since the underlying APIs all work on objects.

Therzok:
Fixed both issues.

Therzok:
@monojenkins backport release-7 y

Therzok:

Issue #5162, opened 2018-06-22
Renaming method should not trigger removal of generic type arguments


atsushieno:
A fun discovery of refactoring operation issue. Try to rename `Apply<T>()` method below:

```
using System;

namespace RenameGenericMethodParameterInference
{
	class MainClass
	{
		static T Apply<T> (int arg) => default (T);

		static void TakeInt (int i) { }

		public static void Main (string [] args)
		{
			TakeInt (Apply<int> (0));
			TakeInt (Apply<int> (0));
		}
	}
}
```

What happens: The generic instance argument `<int>` from call to `Apply<T>()` is removed, and results in compilation error:

```
/home/atsushi/Desktop/RenameGenericMethodParameterInference/RenameGenericMethodParameterInference/Program.cs(13,13):
Error CS0411: The type arguments for method 'MainClass.Foo<T>(int)' cannot be inferred from the usage.
Try specifying the type arguments explicitly. (CS0411) (RenameGenericMethodParameterInference)
```

It may be combined with the refactoring operation (redundant code detection is buggy) too.

> VS bug [#641295](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/641295)

atsushieno:
- Mono JIT compiler version 5.17.0 (master/507464a36ba 2018年  6月 18日 月曜日 02:31:18 JST)
- monodevelop: 4bfed8b

xamarin-release-manager:
Fixed in version 7.7.0.406 (master)

Pull Request #5250 merged by: Mike Krüger
Author: mono
Commit: 3b671ed77cc89ac762dc73b2c96f6174db38dc0e (mono/monodevelop)

xamarin-release-manager:
Fixed in version 7.6.0.1869 (release-7.6)

Pull Request #5251 merged by: Lluis Sanchez
Author: mono
Commit: d8b4f4b2437cb4ec942e40bf71a07398a4a1734f (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #5250 merged by: Mike Kr??ger
Author: mono
Commit: 3b671ed77cc89ac762dc73b2c96f6174db38dc0e (xamarin/monodevelop)

PR #5250, opened 2018-07-02
Fixes issue #5162 Renaming method should not trigger removal of  generic type arguments

mkrueger:


mkrueger:

PR #5256, opened 2018-07-02
[FSW] Fix resource starvation by using a helper class

Therzok:
This PR should be reviewed commit by commit.

The idea is that it uses a helper tree that mimicks the real filesystem tree to prune the maximum amount of directories being watched.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
@@ -191,9 +191,6 @@ namespace MonoDevelop.Projects
 			if (set.Count > 0) {
 				monitoredDirectories [id] = set;
 				foreach (var path in set) {
-					if (path.IsNullOrEmpty)
-						continue;
-
 					tree.AddNode (path, id);
 				}
 
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
@@ -26,6 +26,7 @@
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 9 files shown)


Therzok:
Huh, will take a look at the tests

mrward:
I think the problem is that the old WatchDirectories used to always remove null or empty directories. Now if a HashSet is passed to WatchDirectories this does not happen. Workbench passes a HashSet to the WatchDirectories and this can include null or empty paths for a document even though the Document indicates it is a file. I guess maybe we should have a test for that.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
```diff
@@ -26,180 +26,181 @@
...
+				return modified;
+
+			// Apply new ones if we have any
+			var set = new HashSet<FilePath> (directories.Where(x => !x.IsNullOrEmpty));
+
+			if (set.Count > 0) {
+				monitoredDirectories [id] = set;
+				foreach (var path in set) {
+					if (path.IsNullOrEmpty)
+						continue;
```
mrward: Will this line be reached? It looks to me that the set is created each time and IsNullOrEmpty items are removed before this so the continue will not be called.

Therzok:
With a build from this PR, I don't see the IDE running out of handles when going in idle.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
Therzok: Left-over. Will remove.

Therzok:
Some FSW is not unregistered and causing deadlocks on shutdown. will fix.

Therzok:
@slluis PTAL, should be solid now.

Therzok:

PR #5268, opened 2018-07-03
Fixes VSTS Bug 640356: IntelliSense/Type recognition breaks in editor

mkrueger:
Note: It's just one of numerous exceptions in the log.


slluis:

PR #5269, opened 2018-07-03
[Ide] Addin startup logging and lazy loading

mrward:
Needs mono-addins change.

- AddinManager has a new AddinAssembliesLoaded event which is logged in debug mode.
- Fixes VSTS #640897 - Mime Types loaded on startup
- Fixes VSTS #640899 - FileService loads file system extensions on startup
- Fixes VSTS #641263 - LocationService loads addins on startup
- Fixes VSTS #591639 - Log addin load stacktraces to have a baseline of where an addin is loaded

With the above changes addins are still being loaded by the CommandManager (due to string translations), pad NodeBuilders and startup handlers. So startup is doing less but the addin assemblies are still being loaded.


mrward:

Issue #5235, opened 2018-06-29
assembly browser does not show public members if target Visibility is "Only public members"


atsushieno:
The assembly browser does not show public members if target Visibility is "Only public members".

Open some library on the assembly browser and see what's shown there. The attached sshot is an example (from desktop System.dll):

![screenshot from 2018-06-29 20-27-42](https://user-images.githubusercontent.com/53929/42090082-e40b050c-7bda-11e8-8990-f48d00791d3b.png)

If I choose "All members" then it shows public members too, but they are shown like nonpublic:

![screenshot from 2018-06-29 20-29-28](https://user-images.githubusercontent.com/53929/42090138-251a4d1e-7bdb-11e8-904e-5ae6ffeb641f.png)

Versions:

- monodevelop: master/9406afb4
- mono: master/3eabde79db1

> VS bug [#640892](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/640892)

xamarin-release-manager:
Fixed in version 7.7.0.420 (master)

Pull Request #5270 merged by: Lluis Sanchez
Author: mono
Commit: 3e5a9012ff14886561274c6c09800e2ab616ff82 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 7.6.0.1877 (release-7.6)

Pull Request #5271 merged by: Lluis Sanchez
Author: mono
Commit: 2171d8871d5d87ff6fad774a7c9853e44e41c1f6 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #5270 merged by: Lluis Sanchez
Author: mono
Commit: 3e5a9012ff14886561274c6c09800e2ab616ff82 (xamarin/monodevelop)

PR #5270, opened 2018-07-03
Fixes issue #5235 assembly browser does not show public members if

mkrueger:
target Visibility is "Only public members"

Cecil has fortunately easy to use properties to check the
accessiblity.


mkrueger:

PR #5277, opened 2018-07-03
[Ide] Fix renaming a class asks to reload the file

mrward:
Right clicking a MyClass definition, selecting Refactoring - Rename,
specifying a new name for the class and opting in to rename the file
at the same time would result in the text editor showing a message
indicating the file has changed outside the IDE and allows the file
to be reloaded.

The problem was that the refactoring renames the file and changes
the last saved date of the file. When the rename occurs the file
watcher generates a change event for the new file and the last saved
time of the document was not updated in the DocumentRegistry so the
text editor believed the file had been changed outside the IDE. To
handle this when the document's filename has been changed the last
saved date is updated in the DocumentRegistry since a rename could
affect this date.

Fixes VSTS #637706 - Renaming a class asks to reload file
Fixes VSTS #642585 - Document.DisposeDocument subscribes to an event


mrward:

Therzok:
Ugh, the tests failed because the document was not disposed.

PR #5285, opened 2018-07-03
[Ide.Tests] Ensure Document is disposed

mrward:
The new DocumentReloadTests class added was not disposing the
Document it created which could cause tests to fail.


mrward:

PR #5291, opened 2018-07-04
Fix syntax highlighting not showing for some languages (TypeScript).

KirillOsenkov:
When quickly scanning the TextMate Json file for fileTypes, properly parse the file type regardless whether it starts with a dot or not.

Fix VSTS 642869

### DIFF
--- a/main/tests/Ide.Tests/MonoDevelop.Ide.Editor/SyntaxHighlightingTest_TextMate.cs
+++ b/main/tests/Ide.Tests/MonoDevelop.Ide.Editor/SyntaxHighlightingTest_TextMate.cs
@@ -147,9 +147,9 @@ test ""f\t"" this bar
 		[Test]
 		public void TestTryScanJSonStyle ()
 		{
-			var stream = new MemoryStream ();
-			var writer = new StreamWriter (stream);
-			writer.Write (@"{
+			using (var stream = new MemoryStream ())
+			using (var writer = new StreamWriter (stream)) {
+				writer.Write (@"{
 	""information_for_contributors"": [
 		""This file has been converted from https://github.com/Microsoft/TypeScript-TmLanguage/blob/master/TypeScript.tmLanguage"",
 		""If you want to provide a fix or improvement, please create a pull request against the original repository."",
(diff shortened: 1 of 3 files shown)


KirillOsenkov:
FYI @minestarks

KirillOsenkov:
@mkrueger any idea why the test `TestCDATASection` fails?

```
expected:<span foreground="#e5da73">&lt;![CDATA[ test]]&gt;</span>
But got:<span foreground="#eeeeec">&lt;![CDATA[ test]]&gt;</span>
String lengths are both 57. Strings differ at index 20.
Expected: "<span foreground="#e5da73">&lt;![CDATA[ test]]&gt;</span>"
But was: "<span foreground="#eeeeec">&lt;![CDATA[ test]]&gt;</span>"
-------------------------------^
```

I've seen this test fail previously I think. Looks like `#e5da73` comes from the Dark Theme, and `#eeeeec` is a gray comment?

The test passes locally (on Windows) so I'm not sure how to investigate.

KirillOsenkov:
@slluis we have identified a recent regression (introduced with https://github.com/mono/monodevelop/pull/4949) where we lose classification for all `.ts` files.

This is a pretty bad regression. It would be nice for this PR to be fixed and merged into both master and release-7.6 so that it fixes it in the preview we're about to release.

I'll be out on Wednesday so it would be great if @mkrueger and you could land this together if you have time. It would be super nice to avoid breaking classification for a lot of our file formats.

This not only impacts TypeScript, but all languages whose .tmLanguage files don't contain a `.` in the file extension in the `fileTypes` inside the .json file. I haven't made a list but from cursory glance it's extensive.

main/tests/Ide.Tests/MonoDevelop.Ide.Editor/SyntaxHighlightingTest_TextMate.cs
```diff
@@ -143,5 +143,31 @@ public void TestPushPop ()
 			var h = TextMateFormat.ReadHighlighting (new MemoryStream (Encoding.UTF8.GetBytes (highlighting)));
 			SyntaxHighlightingTest.RunHighlightingTest (h, test);
 		}
+
+		[Test]
+		public void TestTryScanJSonStyle ()
+		{
+			var stream = new MemoryStream ();
```
Therzok: Dispose of the resources you use.

KirillOsenkov: I thought MemoryStream doesn't need disposing? It's just a wrapper around a byte[] array?

slluis:

PR #5295, opened 2018-07-04
[Core] Improve error information on project load failure

mrward:
Include the filename in the error message when a project or an
imported MSBuild project cannot be loaded. This will help locate the
problem. If there is invalid xml, for example, previously you would
see an error about the xml being invalid but there would be no
information about what file has the invalid xml.

Fixes VSTS #640783 - Failure while evaluating MSBuild project

Before:

before

After:

after

In the above screenshots the project filename is displayed. Here you could say that the information is not required in this case since the project file has the error. However if the project is in an imported MSBuild project file somewhere else, say Xamarin.iOS.targets, the error message would indicate which file needs to be looked at.

### DIFF
--- a/main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/MSBuildProjectTests.cs
+++ b/main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/MSBuildProjectTests.cs
@@ -1633,6 +1633,25 @@ namespace MonoDevelop.Projects
 				MSBuildProjectService.UnregisterGlobalPropertyProvider (prov2);
 			}
 		}
+
+		/// <summary>
+		/// Tests that the MSBuildProject.Load method includes information about the file
+		/// being loaded when there was an error. This makes it easier to fix problems when
+		/// the project file or imported file fails to load in the IDE.
+		/// </summary>
+		[Test]
+		public void LoadInvalidXml_ExceptionContainsFileBeingLoaded ()
+		{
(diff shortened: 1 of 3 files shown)


mrward:

main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/MSBuildProjectTests.cs
```diff
@@ -1633,6 +1633,25 @@ public void MultipleGlobalPropertyProvider ()
...
+		/// Tests that the MSBuildProject.Load method includes information about the file
+		/// being loaded when there was an error. This makes it easier to fix problems when
+		/// the project file or imported file fails to load in the IDE.
+		/// </summary>
+		[Test]
+		public void LoadInvalidXml_ExceptionContainsFileBeingLoaded ()
+		{
+			string directory = Util.CreateTmpDir ("MSBuildProjectLoadInvalidXml");
+			string fileName = Path.Combine (directory, "MSBuildProjectLoadInvalidXml.csproj");
+			File.WriteAllText (fileName, "<Project></Project>\n</Project>");
```
Therzok: Can we also test imported files?

mrward: MSBuildProject.Load is used to load imported project files as well as load the main project file. So I only wrote a test for Load.

mrward: But I can add one.

PR #5302, opened 2018-07-04
[Instrumentation] Fix possible crash when converting null to value type

Therzok:
If a new CounterMetadata is created and it is a value type, by querying
the property without setting it first, we would crash because
Convert.ToType does not handle that scenario


Therzok:

PR #5303, opened 2018-07-04
[VCS] Fix exception propagation from VersionControlService

Therzok:
See individual commits. Prevents bricking of IDE that happens due to exception propagation from VersionControlService into users.

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlService.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlService.cs
@@ -245,8 +245,12 @@ namespace MonoDevelop.VersionControl
 					}
 				}
 			}
-			return detectedVCS == null ? null : detectedVCS.GetRepositoryReference (bestMatch, id);
-
+			try {
+				return detectedVCS?.GetRepositoryReference (bestMatch, id);
+			} catch (Exception e) {
+				LoggingService.LogError ($"Could not query {detectedVCS.Name} repository reference", e);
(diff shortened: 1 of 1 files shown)


Therzok:
@slluis we might also want this in 7.5, as it completely bricks the IDE.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlService.cs
```diff
@@ -212,16 +212,15 @@ public static Repository GetRepository (WorkspaceObject entry)
...
 				repo.AddRef ();
-				if (!referenceCache.TryGetValue (repo, out rref)) {
+				if (!referenceCache.TryGetValue (repo, out var rref)) {
 					rref = new InternalRepositoryReference (repo);
 					referenceCache [repo] = rref;
 				}
+				entry.ExtendedProperties [typeof (InternalRepositoryReference)] = rref;
 			}
-			entry.ExtendedProperties [typeof(InternalRepositoryReference)] = rref;
-
```
mrward: This code is a bit different now. Now if GetRepositoryReference returns null the entry.ExtendedProperties [typeof(InternalRepositoryReference)] would not be set to null. It seems that this code cleanup has changed the logic slightly. Not sure if this is a problem or not.

Therzok: Good point, I'll change it back.

Therzok:
Removed the logic change.

Therzok:

PR #5310, opened 2018-07-05
[VSTest] Fix potential UI hang

mrward:
On opening a solution containing a .NET Core test can sometimes
result in a UI hang. The IDE log shows a background thread and the
UI thread both stuck awaiting test discovery:

var discoveredTests = await VsTestDiscoveryAdapter.Instance.DiscoverTestsAsync (Project);

VsTestProjectTestSuite/<OnCreateTests>d__12.MoveNext
in MonoDevelop.UnitTesting.VsTest/VsTestProjectTestSuite.cs:95

Not been able to reproduce this problem directly by opening a
solution. However by modifying the code so a background thread
calls the VsTestDiscoveryAdapter.Instance just before the UI thread
tries to call it then the UI can hang. The problem here is that
the static Instance variable creates a new instance of the
VsTestDiscoveryAdapter. The VsTestDiscoveryAdapter's constructor
created a new progress monitor. The ProgressMonitorManager class
used is a GuiSyncObject so it will try to dispatch the call to
the UI thread. This blocks the background thread. Now the UI thread
attempts to access the VsTestDiscoveryAdapter.Instance and is blocked
resulting in a UI thread hang.

A small fix has been applied here to move the progress monitor
creation out of the constructor and into the DiscoveryTestsAsync
method. This prevents the UI thread hang.

Fixes VSTS #641134 - VS for mac freezes when opening solution

Opened another VSTS bug [643297](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/643297) with some other thread safety problems with the test adapter.

### DIFF
--- a/main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/VsTestDiscoveryAdapter.cs
+++ b/main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/VsTestDiscoveryAdapter.cs
@@ -43,17 +43,32 @@ namespace MonoDevelop.UnitTesting.VsTest
 	class VsTestDiscoveryAdapter : VsTestAdapter
 	{
 		ProgressMonitor monitor;
-		Pad pad;
-		public VsTestDiscoveryAdapter ()
+
+		/// <summary>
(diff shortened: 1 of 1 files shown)


mrward:

PR #5313, opened 2018-07-05
[Core] Optimize ExtensionChain logic

Therzok:
This PR is a consolidated effort to prevent rechaining of extension chains where not needed and making the logic it follows reliable.

This PR touches users of ExtensionChain and ExtensionChain itself.

ExtensionChain is modified to do the following things:
* Dispose of the placeholder extensions at the beginning of the chain, as we cannot guarantee that a ChainedExtension's implementation will not hold onto resources indefinitely if not (one case where Dispose is overriden is in the MSBuild Extension).
* Provide new API that allows for batch modification of an extension, which only rechains on the disposal of the batch modification context.
* Fixed a bug where if insertAfter/insertBefore were not found, we would insert at the -1 position (thus, throw an exception). We now insert before the default position.
* Modify the placeholder extensions to not rechain unless needed.

The logic of `unless needed` is as follows. Given an Extension Chain and the type we want to look for an extension for, we check if the new chain was had a node inserted before or in place of the node we picked as the next implementation. Otherwise, it's just doing reflection checks to get to the same result.

Combined with the batch modifier, this reduces the amount of times an extension chain is init and also the number of times we do reflection on the types.

Users of ExtensionChain are modified to use operations which cause the least amount of rechains possible. That is, instead of manually disposing extensions in a chain, it disposes the chain completely, removing the rechaining on every disposal and they make use of the new API which allows batch modification of the ExtensionChain.

This PR also adds tests around how many times an extension is initialized and how the whole logic works, some docs around how the code works and some comments on what could be possible future improvements.

Fixes VSTS #643291 - ExtensionChain does a lot of unnecessary operations

Supersedes https://github.com/mono/monodevelop/pull/2491

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ExtensionChain.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ExtensionChain.cs
@@ -31,8 +31,8 @@ namespace MonoDevelop.Projects
 {
 	public class ExtensionChain
 	{
-		Dictionary<Type,ChainedExtension> chains = new Dictionary<Type, ChainedExtension> ();
-		ChainedExtension[] extensions;
+		Dictionary<Type,ChainedExtensionSentinel> chains = new Dictionary<Type, ChainedExtensionSentinel> ();
+		ChainedExtension [] extensions;
 		ChainedExtension defaultInsertBefore;
 		BatchModifier batchModifier;
 
@@ -49,13 +49,12 @@ namespace MonoDevelop.Projects
 
 		public T GetExtension<T> () where T:ChainedExtension, new()
 		{
-			ChainedExtension e;
-			if (!chains.TryGetValue (typeof(T), out e)) {
-				e = new T ();
-				e.InitChain (this, ChainedExtension.FindNextImplementation<T> (extensions[0]));
+			if (!chains.TryGetValue (typeof(T), out ChainedExtensionSentinel e)) {
+				e = new ChainedExtensionSentinel (new T());
+				e.Update (this, typeof (T), -1);
 				chains [typeof(T)] = e;
 			}
-			return (T)e;
+			return (T)e.Extension;
 		}
 
 		internal void SetDefaultInsertionPosition (ChainedExtension insertBefore)
@@ -92,8 +91,10 @@ namespace MonoDevelop.Projects
 				extensions [n] = extensions [n - 1];
 			extensions [index] = ext;
 
-			if (batchModifier == null)
-				Rechain ();
+			if (batchModifier != null)
+				batchModifier.UpdateFirstIndex (index);
+			else
+				Rechain (index);
 		}
 
 		internal void RemoveExtension (ChainedExtension ext)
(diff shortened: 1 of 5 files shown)


Therzok:
CI shows yellow, but seems like they finished?

slluis:
This PR is adding complexity to the extension chain model. If the goal of the PR is to reduce the amount of reflection we do, why not just cache the reflection check results? that would be a more simple solution.

Therzok:
Caching won't work with assembly unloading, since the caches are per types, the only real win we'd get with a cache would be a static once.

Invalidation would be hard, since extension chains are explicitly for the assembly load/unload. So we might have bigger problems then.

slluis:
Assembly unloading? we don't unload assemblies.

slluis:
I know this PR is orthogonal to caching reflection checks. My question is, if we implement the cache, do we need this PR at all?

My problem with this PR is that it adds complexity and makes the code harder to understand.

Therzok:
Yes, we do, rechaining also resizes the extension chain array, which happens a lot of times for each project

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ExtensionChain.cs
```diff
@@ -78,42 +81,120 @@ internal void AddExtension (ChainedExtension ext, ChainedExtension insertAfter =
...
 			if (extensions == null)
 				return;
 
-			extensions = extensions.Where (e => e != ext).ToArray ();
-			Rechain ();
+			int index = extensions.Length;
+			extensions = extensions.Where ((e, eindex) => {
+				bool shouldRemove = e == ext;
+				if (shouldRemove)
+					index = Math.Min (index, eindex);
```
slluis: This can just be `index=eindex`

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ExtensionChain.cs
```diff
@@ -78,42 +81,120 @@ internal void AddExtension (ChainedExtension ext, ChainedExtension insertAfter =
...
 			for (int n = extensions.Length - 1; n > index; n--)
 				extensions [n] = extensions [n - 1];
 			extensions [index] = ext;
-			Rechain ();
+
+			if (batchModifier != null)
+				batchModifier.UpdateFirstIndex (index);
+			else
+				Rechain (index);
 		}
```
slluis: The above check for batchModifier can be moved inside `Rechain()` this make the code more readable. If you do that, `BatchModifier.Dispose()` will have to nullify chain.batchModifier before calling `Rechain()`

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ExtensionChain.cs
Therzok: Well, we're looking for the minimum value. I think the `Where` eindex is increasing from `0` -> `extensions.Length -1`, so setting it to eindex will choose the _last_, since we don't have any safeguard against duplicate extensions.

Might be overthinking here, but will validate.

slluis: Yeah, the thing is that there can't be duplicate extensions, since that would cause chain loops.

slluis (dismissed):
See comments.

Therzok:
Changes applied

slluis (approved):

PR #5318, opened 2018-07-05
[VCS] Fix Review and Commit selection persisting through context menu

Therzok:
The fix is two fold, once the current info needed to be passed
to the command update handlers used by the context menu.

Then, there was a bug surrounding select all and clicking in the tree,
which would make nodes in the tree unselectable.

Fixes VSTS #621178 - [Feedback] Source Control selection in "Tools -> Options -> Source Control" is lost when opening context menu.


Therzok:

Issue #5184, opened 2018-06-25
Intermittent crashing when building Xamarin Forms Control Gallery


PureWeen:
I was testing some changes to the control galleries project file by deleting out this custom command

https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.ControlGallery.Android/Xamarin.Forms.ControlGallery.Android.csproj#L59

And every few builds VS for Mac would crash.

My steps were basically
-edit project file
-close vs for Mac
-delete bin/obj folders in the https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.ControlGallery.Android
-open solution back up
-build Xamarin.Forms.ControlGallery.Android.csproj
-The build takes a bit and sometimes it would just cause VS to crash

I've attached the log from the crash.
Version 7.6 Preview (7.6 build 1773)

[Ide.2018-06-22__13-56-14.log.zip](https://github.com/mono/monodevelop/files/2133915/Ide.2018-06-22__13-56-14.log.zip)

> VS bug [#644694](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/644694)

mrward:
Does this still happen if the IoT addin is disabled or uninstalled?

Most of the IDE errors before the crash seem to be related to having Xamarin.IoT 7.5 installed.

```
MSBuild property evaluation failed: [Microsoft.Build.Utilities.ToolLocationHelper]::GetPathToStandardLibraries($(TargetFrameworkIdentifier), $(TargetFrameworkVersion), $(TargetFrameworkProfile), $(PlatformTarget), $(TargetFrameworkRootPath))
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.TypeInitializationException: The type initializer for 'Microsoft.Build.Collections.MSBuildNameIgnoreCaseComparer' threw an exception. ---> System.EntryPointNotFoundException: GetSystemInfo
  at (wrapper managed-to-native) 
```

Crash:

```
Stacktrace:

  at <unknown> <0xffffffff>
  at (wrapper managed-to-native) System.Runtime.InteropServices.GCHandle.CheckCurrentDomain (int) [0x00007] in <30f82f22c5df4a87974ae23bf2702f8f>:0
[...]

PR #5323, opened 2018-07-05
[MSBuild] Fix fatal exception on cancelling an MSBuild target being run

mrward:
Cancelling an MSBuild target that was being run, which was not a build,
would cause a fatal unhandled exception in the MSBuild build host.

```
The operation cannot be completed because BeginBuild has not yet been
called.
  at Microsoft.Build.Shared.ErrorUtilities.ThrowInvalidOperation (System.String resourceName, System.Object[] args)
 in /_/src/Shared/ErrorUtilities.cs:304
  at Microsoft.Build.Execution.BuildManager.ErrorIfState (Microsoft.Build.Execution.BuildManager+BuildManagerState disallowedState, System.String exceptionResouorce)
 in /_/src/Build/BackEnd/BuildManager/BuildManager.cs:1195
  at Microsoft.Build.Execution.BuildManager.EndBuild ()
 in /_/src/Build/BackEnd/BuildManager/BuildManager.cs:571
at MonoDevelop.Projects.MSBuild.BuildEngine.STARunner ()
 in MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild/BuildEngine.Shared.cs:333
```

So Project.RunTarget is called with a cancellation token and that token
is cancelled.

The problem was that the cancel request caused the worker thread to be
aborted. Code in the ThreadAbortException catch block would then try
to call EndBuild on the DefaultBuildManager. When not running a build
the DefaultBuildManager's BeginBuild is not called. The EndBuild
method has some state checks which throw an exception if a build is not
being run.

Now a check is made to see if a build operation was started before
trying to call EndBuild when the thread is aborted.

Fixes VSTS #643113 - MSBuild broken due to commit 1c63942

### DIFF
--- a/main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild/BuildEngine.cs
+++ b/main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild/BuildEngine.cs
@@ -143,16 +143,19 @@ namespace MonoDevelop.Projects.MSBuild
 		{
 			// End the MSBuild build session started in BeginBuildOperation
 
-			RunSTA (this, delegate {
-				engine.RemoveGlobalProperty ("CurrentSolutionConfigurationContents");
-				BuildOperationStarted = false;
-				BuildManager.DefaultBuildManager.EndBuild ();
+			RunSTA (this, EndBuildOperationNonSTA);
+		}
 
-				// Dispose the loggers. This will flush pending output.
-				loggerAdapter.Dispose ();
-				loggerAdapter = null;
-				sessionLogWriter = null;
-			});
+		void EndBuildOperationNonSTA ()
+		{
+			engine.RemoveGlobalProperty ("CurrentSolutionConfigurationContents");
+			BuildOperationStarted = false;
+			BuildManager.DefaultBuildManager.EndBuild ();
+
+			// Dispose the loggers. This will flush pending output.
+			loggerAdapter.Dispose ();
+			loggerAdapter = null;
+			sessionLogWriter = null;
 		}
 
 		public MSBuildLoggerAdapter StartProjectSessionBuild (IEngineLogWriter logWriter)
@@ -179,7 +182,7 @@ namespace MonoDevelop.Projects.MSBuild
 			if (buildEngine.BuildOperationStarted) {
 				// Try to end the build here, to workaround a finalizer crash in zstream in mono
 				// https://github.com/mono/mono/issues/9142
-				BuildManager.DefaultBuildManager.EndBuild ();
+				buildEngine.EndBuildOperationNonSTA ();
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


mrward:

Therzok:
Nice.

Therzok:
Ugh, where's the test suite run?

Therzok:
When we added the structured output pad, we requested a binlog logger. The binlog logger is tied to the build operation, so if we don't explicitly end the build, the logger won't get cancelled.

In some cases (i.e. test runs), cancellation could happen before the builder exits, so the binlog logger would get finalized, essentially causing this native crash: https://github.com/mono/mono/issues/9142

mrward:
Going to look at changing this to call EndBuildOperation if the main build thread is aborted.

Also going to look at this Android build error which has the same Mono crash and may be the same problem -https://github.com/mono/monodevelop/issues/5184 This Android build problem can be reproduced locally.

mrward:
Seems that calling EndBuildOperation during the thread abort causes the build host to hang so the build never finishes and you cannot stop it. Calling the DefaultBuildManager.EndBuild does not hang the build host.

Looking at the code I suspect this is because EndBuildOperation calls RunSTA which will try to lock the threadLock but that will be locked by the STARunner thread which will be handling the thread abort exception.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.Projects.MSBuild/BuildEngine.cs
```diff
@@ -143,16 +143,19 @@ void EndBuildOperation ()
...
+		}
 
-				// Dispose the loggers. This will flush pending output.
-				loggerAdapter.Dispose ();
-				loggerAdapter = null;
-				sessionLogWriter = null;
-			});
+		void EndBuildOperationNonSTA ()
+		{
+			engine.RemoveGlobalProperty ("CurrentSolutionConfigurationContents");
```
Therzok: Can we add a `Debug.Assert (MonitorIsEntered (threadLock))` here?

mrward: Yeah - let me add that and test it.

slluis:
So, EndBuild needs to be called when BeginBuild has been called. BeginBuild is not called when running a target, so this PR adds a check so that if BeginBuild has not been called, it doesn't call EndBuild. I get that.

What I don't understand is why that EndBuild call on thread abort was necessary at all. When a build starts the BeginBuild method is called by MD, and when the build ends, the EndBuild method should be called also by MD, no matter if the build was cancelled or not. This is done [here](http://source.monodevelop.com/#MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildSolutionExtension.cs,66). Isn't this called when a build is cancelled?

mrward:
Yes EndBuildOperation is called when the build is cancelled in the IDE.

The Mono crash happens with the Android project mentioned in #5184 without having to cancel the build.

Therzok:
Further analysis:

* Found a simple repro for mono - https://github.com/mono/mono/issues/9142#issuecomment-404171551
* I still haven't figured out why EndBuildOperation is _not_ being called. There's multiple things to take into account, such as process killing forcing the builder to go into finalization.

mrward:
Closing this. Replaced by Marius's fix - https://github.com/mono/monodevelop/pull/5379

PR #5324, opened 2018-07-05
[Roslyn] Disable ClosedFileDiagnostic for C# by default

Therzok:
This might improve CPU consumptions while loading the roslyn project.

This is disabled by default for C# in VSWindows, so we should do the same.

Fixes VSTS #643384 - Roslyn Closed File Diagnostic is toggled on by default for C#


Therzok:

PR #5327, opened 2018-07-05
Surface lazy methods for CompositionManager

Therzok:
This API is useful for cases where you want to lazily query a service which you use later on.


Therzok:

PR #5334, opened 2018-07-05
[Xml] Implement Expand/Shrink selection for XML

mhutch:
Implement the expand selections/shrink selection command for XML files.

### DIFF
--- a/main/src/addins/Xml/Editor/XmlExpandSelectionHandler.cs
+++ b/main/src/addins/Xml/Editor/XmlExpandSelectionHandler.cs
@@ -1,0 +1,384 @@
+﻿//
+// Copyright (c) Microsoft Corp
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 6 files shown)


main/src/addins/Xml/Editor/XmlExpandSelectionHandler.cs
```diff
@@ -0,0 +1,384 @@
...
+			Name,
+			Content,
+			WithClosing,
+			Document,
+			Attributes
+		}
+
+		class XmlExpandSelectionAnnotation
+		{
+			Stack<(int, SelectionLevel)> expansions = new Stack<(int, SelectionLevel)> ();
```
Therzok: How big can this stack get and/or how often is it changed?

Might be better with a non-array-backed stack? (i.e. ImmutableStack)

main/src/addins/Xml/Editor/XmlExpandSelectionHandler.cs
```diff
@@ -0,0 +1,384 @@
...
+				}
+				var current = NodePath [Index];
+				switch (Level) {
+				case SelectionLevel.Self:
+					return current.Region;
+				case SelectionLevel.WithClosing:
+					var element = (XElement)current;
+					return new DocumentRegion (element.Region.Begin, element.ClosingTag.Region.End);
+				case SelectionLevel.Name:
+					return current.TryGetNameRegion ().Value;
```
Therzok: `current.TryGetNameRegion ()?.Value`

mhutch: It must have a value for SelectionLevel.Name to be true

main/src/addins/Xml/Editor/XmlExpandSelectionHandler.cs
mhutch: The maximum size will be about double the depth in the AST that the "Expand selection" command is executed. It grows when the "Expand selection" command is executed, shrinks when the "shrink selection" command is executed, and is discarded when the caret moves.

In theory it could be eliminated entirely as it's possible to compute the values on demand but the computation is nontrivial and I wanted to reduce risk of bugs caused by getting "shrink" calculations different or out of sync from "expand" calculations.

Therzok: Then, we could use an immutablestack. The advantage is no arrays allocated, nodes are linked together like linked lists

mhutch: But then we have a bunch of nodes allocated... Sure they're smaller but this isn't used much so it probably doesn't matter.

Therzok: Yes, but the net impact changes. When resizing a stack, you do an O(n) allocation, whilst when using ImmutableStack, you just add a node

mhutch: https://github.com/mono/monodevelop/pull/5334

PR #5346, opened 2018-07-09
[Xml] Switch Stack to ImmutableStack

mhutch:
Reduce array allocations. Belated review feedback from https://github.com/mono/monodevelop/pull/5334


mhutch:
Merging as the failing test is unrelated

PR #5347, opened 2018-07-10
[Telemetry] Add solution extension for fetching platform-specific metadata

bratsche:
This adds a way to fetch platform-specific metadata for extra `XS.Solution.Perf.ProjectBuild.*` properties.

This is part of VSTS #598635

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectExtension.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectExtension.cs
@@ -302,10 +302,6 @@ namespace MonoDevelop.Projects
 
 		#endregion
 
-		internal protected virtual ProjectEventMetadata UpdateProjectEventMetadata (ConfigurationSelector configurationSelector, ProjectEventMetadata metadata)
-		{
-			return next.UpdateProjectEventMetadata (configurationSelector, metadata);
-		}
 	}
 }
 
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectExtension.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectExtension.cs
@@ -301,6 +301,11 @@ namespace MonoDevelop.Projects
 		}
 
 		#endregion
+
+		internal protected virtual ProjectEventMetadata UpdateProjectEventMetadata (ConfigurationSelector configurationSelector, ProjectEventMetadata metadata)
+		{
+			return next.UpdateProjectEventMetadata (configurationSelector, metadata);
+		}
 	}
 }
 
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectExtension.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectExtension.cs
@@ -301,7 +301,6 @@ namespace MonoDevelop.Projects
 		}
 
 		#endregion
-
 	}
 }
 
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionItem.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionItem.cs
@@ -887,7 +887,7 @@ namespace MonoDevelop.Projects
 
 		public ProjectEventMetadata CreateProjectEventMetadata (ConfigurationSelector configurationSelector)
 		{
-			return ItemExtension.OnGetProjectEventMetadata (configurationSelector, new ProjectEventMetadata ());
+			return ItemExtension.OnGetProjectEventMetadata (configurationSelector);
 		}
 
 		[Obsolete ("Use OnGetProjectEventMetadata (ProjectEventMetadata) instead")]
@@ -895,11 +895,7 @@ namespace MonoDevelop.Projects
 		{
 		}
 
-		protected virtual void OnGetProjectEventMetadata (ProjectEventMetadata metadata)
-		{
-		}
-
-		protected virtual ProjectEventMetadata OnGetProjectEventMetadata (ConfigurationSelector configurationSelector, ProjectEventMetadata origMetadata)
+		protected virtual ProjectEventMetadata OnGetProjectEventMetadata (ConfigurationSelector configurationSelector)
 		{
 			string id = null;
 			if (configurationSelector != null) {
@@ -909,9 +905,7 @@ namespace MonoDevelop.Projects
 				}
 			}
 
-			var metadata = new ProjectEventMetadata (id);
-			OnGetProjectEventMetadata (metadata);
-			return metadata;
+			return new ProjectEventMetadata (id);
 		}
 
 		/// <summary>
@@ -1607,9 +1601,10 @@ namespace MonoDevelop.Projects
 				return Item.OnGetReferencedItems (configuration);
 			}
 
-			protected internal override ProjectEventMetadata OnGetProjectEventMetadata (ConfigurationSelector configurationSelector, ProjectEventMetadata metadata)
+			protected internal override ProjectEventMetadata OnGetProjectEventMetadata (ConfigurationSelector configurationSelector)
 			{
-				return Item.OnGetProjectEventMetadata (configurationSelector, metadata);
+				var metadata = Item.OnGetProjectEventMetadata (configurationSelector);
+				return new ProjectEventMetadata (metadata);
 			}
 
 			internal protected override void OnSetFormat (MSBuildFileFormat format)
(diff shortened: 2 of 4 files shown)


iainx (approved):

mrward (approved):
Should we have a OnGetProjectEventMetadata method added to the SolutionItemExtension? The default behaviour would be to call SolutionItem.OnGetProjectEventMetadata. That seems to match the current pattern used in the SolutionItemExtension.

mrward (changes requested):
Meant to comment not approve it.

slluis (changes requested):
I agree with Matt.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectExtension.cs
```diff
@@ -301,6 +301,7 @@ internal protected virtual void OnFileRenamedInProject (ProjectFileRenamedEventA
 		}
 
 		#endregion
+
 	}
```
mrward: Not sure what this new line is for :)

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionItem.cs
```diff
@@ -896,8 +896,7 @@ public ProjectEventMetadata CreateProjectEventMetadata (ConfigurationSelector co
 			}
 
 			var metadata = new ProjectEventMetadata (id);
-			OnGetProjectEventMetadata (metadata);
-			return metadata;
+			return itemExtension.OnGetProjectEventMetadata (metadata, configurationSelector);
```
mrward: I was thinking more along the lines of the SolutionItemExtension.OnProjectEventMetadata method calling into SolutionItem which would create the ProjectEventMetadata. So CreateProjectEventMetadata would be just:
```
public ProjectEventMetadata CreateProjectEventMetadata (ConfigurationSelector configurationSelector)
{
	return ItemExtension.OnGetProjectEventMetadata (configurationSelector);
}
```

Then the SolutionItemExtension would at the end call back into a non-public method on the SolutionItem which would do the creation:

```
internal ProjectEventMetadata OnGetProjectEventMetadata (ConfigurationSelector configurationSelector)
{
	string id = null;
	if (configurationSelector != null) {
		var slnConfig = configurationSelector as SolutionConfigurationSelector;
		if (slnConfig != null) {
			id = slnConfig.Id;
		}
	}

	return new ProjectEventMetadata (id);
}
```

The SolutionItem's GetReferencedItems method has a similar pattern to this.

mrward: Also the protected OnGetProjectEventMetadata (ProjectEventMetadata metadata) should be called at some point here in the OnGetProjectEventMetadata method.

bratsche: @mrward I'm still struggling to get this to work again with these changes. And I don't really understand that `protected OnGetProjectEventMetadata (ProjectEventMetadata metadata)` call there before it returns. Is that what I should be overriding to add the platform-specific metadata? I don't see anywhere to override it from. As far as I understand the places we can override in the addins are deriving from the extensions, not from project/solutionitem.

bratsche:
@mrward I think this is pretty close to what you suggested now.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionItem.cs
```diff
@@ -886,6 +886,20 @@ protected virtual Task<BuildResult> OnClean (ProgressMonitor monitor, Configurat
...
+		[Obsolete ("Use OnGetProjectEventMetadata (ProjectEventMetadata) instead")]
+		protected virtual void OnGetProjectEventMetadata (IDictionary<string, string> metadata)
+		{
+		}
+
+		protected virtual void OnGetProjectEventMetadata (ProjectEventMetadata metadata)
+		{
+		}
+
+		protected virtual ProjectEventMetadata OnGetProjectEventMetadata (ConfigurationSelector configurationSelector, ProjectEventMetadata origMetadata)
```
mrward: So I was more thinking the creation of the project event metadata would be done on the SolutionItem's OnGetProjectEventMetadata. Currently it is being passed in and returned.

The other SolutionItemExtension methods take something as a parameter and then return something else.

`internal protected virtual DateTime OnGetLastBuildTime (ConfigurationSelector configuration)`

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionItem.cs
```diff
@@ -1601,6 +1601,12 @@ internal protected override IEnumerable<SolutionItem> OnGetReferencedItems (Conf
 				return Item.OnGetReferencedItems (configuration);
 			}
 
+			protected internal override ProjectEventMetadata OnGetProjectEventMetadata (ConfigurationSelector configurationSelector)
+			{
+				var metadata = Item.OnGetProjectEventMetadata (configurationSelector);
+				return new ProjectEventMetadata (metadata);
```
mrward: This can just return Item.OnGetProjectEventMetadata. No need to create the ProjectEventMetadata again here.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionItemExtension.cs
```diff
@@ -88,6 +88,12 @@ internal protected virtual void OnInitializeFromTemplate (ProjectCreateInformati
 			next.OnInitializeFromTemplate (projectCreateInfo, template);
 		}
 
+		internal protected virtual ProjectEventMetadata OnGetProjectEventMetadata (ConfigurationSelector configurationSelector)
+		{
+			var metadata = next.OnGetProjectEventMetadata (configurationSelector);
+			return new ProjectEventMetadata (metadata);
```
mrward: Similar here - no need to create a new ProjectEventMetadata class just return it from the next.OnGetProjectEventMetadata call.

mrward:
OnEndLoad seems to still be calling the obsolete GetProjectEventMetadata. It is not part of your change but I am not sure that is correct. Shouldn't it be calling the new CreateProjectEventMetadata?

bratsche:
@iainx ^ Is that intended to still be calling the obsolete `GetProjectEventMetadata()` from `OnEndLoad()` still?

mrward:
Another thing I am wondering about, which is not part of your change, is how the deprecated methods work. If I am using the old deprecated methods, either overriding one or calling the public one, would we expect the information captured using the new methods to be available here and vice versa?

It is probably not a problem for us since we would just make sure the non-obsolete methods are called. Just wondering. @iainx

iainx:
I think it should be calling both, calling the new one first and if that returns null calling the obsolete one. It's awkward because we can't stop calling the old obsolete one because it's a API break.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionItemExtension.cs
bratsche: If we don't do this then they don't ever get combined though, right? The whole point of all this was to combine the platform-specific properties into the project properties.

bratsche:
It used to pass the `ProjectEventMetadata` down the chain and in the extension it would do something like `new IPhoneProjectEventMetadata (metadata)`, but now I don't think it's combining the metadata anymore since I changed `OnGetProjectEventMetadata()` to just take a `ConfigurationSelector` parameter.

mrward (commented):
This is looking good now with the extension and project bits.

Not sure if we want to address the other problems here or not.

mrward:
In your project extension, such as the IPhone one, you need to call the base class method to get the metadata, then add the new metadata. That is what is missing on that side. You can still do `new IPhoneProjectEventMetadata (metadata)` just call the base class method to get the metadata.

bratsche:
@mrward Thanks!

bratsche:
@mrward I think I'd rather have other issues handled in a separate PR, if that's okay with you and @iainx, so I can just get the platform-specific stuff done.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionItem.cs
```diff
@@ -1824,6 +1829,19 @@ public ProjectEventMetadata ()
 		{
 		}
 
+		public ProjectEventMetadata (CounterMetadata metadata)
+			: base (metadata)
+		{
+		}
+
+		public ProjectEventMetadata (CounterMetadata metadata, string configurationId)
```
mrward: This constructor does not seem to be used anywhere so I think it can be removed.

mrward (approved):

slluis (approved):
Nice!

bratsche:

PR #5351, opened 2018-07-10
[Ide][Mac] Fix parent window handling in MacAlertDialogHandler

sevoku:
This PR fixes alert parent handling if the parent is a native (non Gtk) window, `GtkQuartz.FocusWindow(Gtk.Window)` fails in this case.

### DIFF
--- a/main/src/addins/MacPlatform/Dialogs/MacAlertDialogHandler.cs
+++ b/main/src/addins/MacPlatform/Dialogs/MacAlertDialogHandler.cs
@@ -174,15 +174,20 @@ namespace MonoDevelop.MacIntegration
 
 				int response = -1000;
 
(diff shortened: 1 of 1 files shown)


olegoid:
@sevoku, do we have a bug for that? or, what goal are we trying to achieve with this PR?

sevoku:
@olegoid sorry for the late response, I missed the GH notification mail. It can't be tested right now, because we have no native windows using `MessageService`. I have tested it for my use case with a native window and it works. Hence I need a review/approve that it doesn't break anything else (especially your use case in the preferences).

sevoku:
rebased to resolve conflicts and fix build

olegoid:
@sevoku it looks like dialog queues still work (at least with light theme)

sevoku:
Thanks @olegoid!

PR #5352, opened 2018-07-10
[Ide][Mac] Fix closed window handling in IdeTheme

sevoku:
With HighSierra [NSNotificationCenter observer handling has been
redeisgned](https://developer.apple.com/library/archive/releasenotes/Foundation/RN-Foundation/index.html#10_11NotificationCenter) and now observers are removed automatically during finalization.
However the actual NSWindow.WillCloseNotification might be raised
after the window has been finalized and removing the observer manually
would result in a hard native crash, hence we need to treat
window closed winows differently depending on the macOS version.


sevoku:
Oh, looks like I've found another issue... please don't merge for now!

PR #5355, opened 2018-07-10
Update reference to md-addins

mhutch:
Remove accidentally bundled preview IoT extension


mhutch:
Looks like it didn't fix the test failure, but merging anyways as the preview extension should not have been included.

Issue #4377, opened 2018-03-29
WebReference addin slows down folder building


Therzok:
For every folder item we have, we try to build the WebReferences/WCF node. We should use a bound item collection to read, rather than trying to get al the reference items on every query

```
0.10%   BuildChildNodes  •  651 ms  •  2 calls  •  MonoDevelop.Ide.Gui.Pads.ProjectPad.FolderNodeBuilder.BuildChildNodes(ITreeBuilder, Object)
  0.09%   AddChildren  •  611 ms  •  4 calls  •  MonoDevelop.Ide.Gui.Components.ExtensibleTreeView+TreeBuilder.AddChildren(IEnumerable)
    0.07%   GetAttributes  •  459 ms  •  135 calls  •  MonoDevelop.Ide.Gui.Components.ExtensibleTreeView+TreeBuilder.GetAttributes(ITreeBuilder, NodeBuilder[], Object)
      0.07%   GetNodeAttributes  •  454 ms  •  65 calls  •  MonoDevelop.WebReferences.NodeBuilders.ProjectFolderNodeBuilderExtension.GetNodeAttributes(ITreeNavigator, Object, NodeAttributes&)
        0.07%   MoveNext  •  454 ms  •  65 calls  •  MonoDevelop.WebReferences.WebReferencesService+<GetWebReferenceItems>d__4.MoveNext
          0.05%   MoveNext  •  306 ms  •  65 calls  •  MonoDevelop.WebReferences.WS.WebServiceEngineWS+<GetReferenceItems>d__2.MoveNext
            ►0.02%   Any  •  152 ms  •  65 calls  •  System.Linq.Enumerable.Any(IEnumerable[TSource])
            ►0.02%   MoveNext  •  146 ms  •  65 calls  •  MonoDevelop.Projects.ProjectItemCollection+<GetAll>d__7`2.MoveNext
            ►0.00%   get_BaseDirectory  •  5 ms  •  65 calls  •  MonoDevelop.Projects.SolutionFolderItem.get_BaseDirectory
             0.00%   Exists  •  3 ms  •  65 calls  •  System.IO.Directory.Exists(String)
            ►0.00%   Combine  •  0 ms  •  65 calls  •  MonoDevelop.Core.FilePath.Combine(String)
            ►0.00%   GetAll  •  0 ms  •  130 calls  •  MonoDevelop.Projects.ProjectItemCollection`1.GetAll
[...]

PR #5366, opened 2018-07-11
Fixes issue #4377 WebReference addin slows down folder building

mkrueger:


Therzok:
This works for me, but the proper solution would be to use a bound ItemCollection to the main item collection, like here:
http://source.monodevelop.com/#MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs,88 http://source.monodevelop.com/#MonoDevelop.AddinMaker/AddinProjectFlavor.cs,32

But that means having WebReferenceItem deriving from ProjectItem?

main/src/addins/MonoDevelop.WebReferences/MonoDevelop.WebReferences/WebReferencesProjectExtension.cs
```diff
@@ -0,0 +1,74 @@
...
+namespace MonoDevelop.WebReferences
+{
+	[ExportProjectModelExtension]
+	class WebReferencesProjectExtension : DotNetProjectExtension
+	{
+		protected override void OnItemReady ()
+		{
+			base.OnItemReady ();
+			Project.ProjectItemAdded += ItemChanged;
+			Project.ProjectItemRemoved += ItemChanged;
```
slluis: Instead of subscribing the ProjectItemAdded/ProjectItemRemoved events, you can just override OnItemsAdded and OnItemsRemoved, which are defined by ProjectExtension.

slluis (changes requested):

Therzok:
So, my initial proposed solution in the github issue wanted to make use of:
http://source.monodevelop.com/#MonoDevelop.WebReferences/MonoDevelop.WebReferences.WS/WebServiceEngineWS.cs,59
http://source.monodevelop.com/#MonoDevelop.WebReferences/MonoDevelop.WebReferences.WCF/WebServiceEngineWCF.cs,89

Using these to ProjectItem kinds and using `Bind<>` to create the collections.

mkrueger:
Would require some larger structural changes, updating etc. I just made a minimal cache to fix that issue. But you're right that this approach is probably cleaner. 

But I think my patch is less code needed than getting the Bind way working. It would need some restructuring and a project extension as well - or at least refactoring the WebServiceEnigne to be one.

main/src/addins/MonoDevelop.WebReferences/MonoDevelop.WebReferences/WebReferencesProjectExtension.cs
```diff
@@ -0,0 +1,77 @@
...
+			if (webReferenceItemsWcf == null)
+				webReferenceItemsWcf = WebReferencesService.WcfEngine.GetReferenceItems (Project).ToList ();
+			return webReferenceItemsWcf;
+		}
+
+		List<WebReferenceItem> webReferenceItemsWS;
+		internal IEnumerable<WebReferenceItem> GetWebReferenceItemsWS ()
+		{
+			if (webReferenceItemsWS == null)
+				webReferenceItemsWS = WebReferencesService.WcfEngine.GetReferenceItems (Project).ToList ();
```
slluis: Should be WebReferencesService.WsEngine

slluis:
I tend to agree with Mike. The PR fixes the problem, we don't need to expend much time refactoring all that code right now.

Therzok (approved):

mkrueger:
Cut&Paste == evil - ty :)

slluis (approved):

Therzok:

PR #5377, opened 2018-07-11
[BuildOutput] Use Array.IndexOf instead of LINQ's .Contains

rodrmoya:


rodrmoya:

PR #5378, opened 2018-07-11
[DotNetCore] Fix project not restored on target framework being changed

mrward:
Changing the target framework of a .NET Core project in project
options would not trigger a NuGet package restore. After doing this
if the project file was edited in the text editor and saved the
packages would not be restored. This was because the file watcher
integration changed the logic around file reloads. The .NET Core addin
was setting the Project.NeedsReload flag to true and then calling
FileService.NotifyFileChanged to force a file reload of the project.
This no longer works since the file watcher is not responsible for
triggering the reload based on file events that happen outside the
IDE. To force a reload in the IDE there are Reload methods available
on the Project itself. To solve this problem this old logic has been
removed and now when the target framework of the project is changed,
after the project is saved, a NuGet package restore is run. Before the
restore is run the project needs to be re-evaluated to ensure the
target framework and any implicit PackageReferences are correct when
the restore is run. Without this re-evaluation the old  target
framework and old PackageReferences were being used.

Fixes VSTS #635592 - .NET Core project not restored after changing
project properties


mrward:

PR #5386, opened 2018-07-11
[MD] Add in-house analyzer example

Therzok:
This PR implements a localization analyzer which looks for calls
to GetString which are not being passed a string literal.

In that case, we report an error.

TODO:
* Figure out a way to integrate these analyzers as part of the build
* Figure out a way to decouple them from MonoDevelop.CSharpBinding (could go into Refactoring)
* Code fixes
* Look for other types which could benefit from this

List of other potential analyzers, in no order of importance / difficulty:
* Avoid Application.Invoke and use Runtime.RunInMainThread instead
* Avoid GLib.Idle where possible
* Analyzer for BCL optimizations - https://github.com/xamarin/md-addins/commit/36c57f54015d3ca9881774baffc3082c05fe92da https://github.com/mono/monodevelop/commit/6e566e63dbdce849309e9d2eab0437e6ad700a80 https://github.com/mono/monodevelop/commit/db21a1560f9b36d26323e53d923e0f2434274559
* Analyzer for Editor APIs - https://github.com/mono/monodevelop/commit/e452de482a13088da22d54037589e2d1fa57b7a1
* Avoid structs without IEquatable<T>
* Avoid .NET 1.1 collections to avoid enumerator allocation
* Always set a `Name` property to widgets for UI test capabilities
* ~Do not override Gtk methods that don't start with `On`. i.e. Destroy -> OnDestroyed~
* Async handler registration to sync event
* GettextCatalog.GetString in loop should be cached
* [PerformanceSensitive] attribute should be used to mark optimized methods and should prevent adding allocations


Therzok:
Screenshot of how it looks / what it does.

screen shot 2018-07-11 at 20 05 40

mhutch:
Awesome! Would love to have more of these, I'm, sure there's a whole bunch of bad patterns we could detect.

IMO this belongs in a separate project that can be referenced from the shared msbuild targets as an analyzer reference. That way we have have our analyzers run on build/CI, and have analyzers match the source code version rather than editor version. That said, we'll have to implement analyzer references...

Therzok:
Added unit tests for the analyzer.

Therzok:
Implemented the gtk destroy analyzer. It seems the only one off which is surfaced in the API.

Therzok:
@slluis what do you think?

slluis:
Maybe we could consume the analyzers through a NuGet?

slluis:
In any case I love the idea!

Therzok:
Closing in favor of a new repository that will be published to nuget for us to consume. https://github.com/Therzok/MonoDevelop.Analyzers

PR #5397, opened 2018-07-12
Fix crash in MEF cache deserialization

mhutch:
Fixes a crash that happened on my machine because for some reason the MEF cache control deserialized as null. Guessing it was because of a format change?

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
@@ -196,13 +196,17 @@ namespace MonoDevelop.Ide.Composition
 		[Serializable]
 		class MefControlCache
 		{
+			[JsonRequired]
 			public MefControlCacheAssemblyInfo [] AssemblyInfos;
 		}
 
 		[Serializable]
 		internal class MefControlCacheAssemblyInfo
 		{
+			[JsonRequired]
 			public string Location;
+
+			[JsonRequired]
 			public DateTime LastWriteTimeUtc;
 		}
 	}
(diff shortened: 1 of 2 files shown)


Therzok:
Hmmm, we should be able to unit test this. What were the control file contents?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
```diff
@@ -117,6 +115,13 @@ internal bool CanUse ()
 						return false;
 					}
 
+					//this can return null (if the cache format changed?). clean up and start over.
```
Therzok: I guess the saveTask was cancelled and it only partly serialized. Not sure how to control that better though.

mhutch:
Unfortunately I forgot to save it :( I did look at the file though, and it was not empty.

mhutch:
What would you want to test in this case?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs
Therzok: So, what I suggest here is:

* Set all the members and classes we have as required via `[Required]`. That will cause json to throw in case we don't have values.
* Add some tests with mef-cache-control containing just `{}`, and empty file, and partial json

Here is a sample JSON to create partial data:
https://gist.github.com/Therzok/fd0fae2c79bf586a8bc43f219c350e0c

Therzok:
So, what I suggest here is:

* Set all the members and classes we have as required via `[Required]`. That will cause json to throw in case we don't have values.
* Add some tests with mef-cache-control containing just `{}`, and empty file, and partial json

Here is a sample JSON to create partial data:
https://gist.github.com/Therzok/fd0fae2c79bf586a8bc43f219c350e0c

Issue #5279, opened 2018-07-03
Text Editor outdents to hard left instead of going back an indent level


praeclarum:
When you hit Return in the F# editor it moves the cursor the next line and the same column. Great. However, the virtual space system is broken and if you outdent, it takes you to the beginning of the line.

This is very frustrating when trying to write new functions or else blocks, etc.

This is a regression! The editor used to behave itself.

**Current Behavior**

If you hit backspace, or unindent (Cmd+[) then it takes you to the beginning of the line. 

**Expected Bahavior**

It should outdent the cursor as the C# and text editors do.

I have included a video. 

![fsharpbadoutdent](https://user-images.githubusercontent.com/323548/42232000-262a3178-7ebb-11e8-8045-0aae90751e65.gif)

> VS bug [#642673](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/642673)

nosami:
This behaviour happens in C# files too, but only when unindenting _immediately_ after inserting a newline. As soon as you indent once or more, unindenting works as expected.

![outdentcsharp](https://user-images.githubusercontent.com/667194/42449342-394c4b62-8378-11e8-9ec7-4f38c8bad866.gif)

praeclarum:
Yes it seems to be a bug with the virtual space system.

xamarin-release-manager:
Fixed in version 7.7.0.592 (master)

Author: Mike Krüger
Commit: 71afefd7f06aa29fde7daa77b50036ba0f66c916 (mono/monodevelop)

PR #5398, opened 2018-07-12
Fixes issue #5279 Text Editor outdents to hard left instead of going

mkrueger:
back an indent level

### DIFF
--- a/main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Actions/MiscActions.cs
+++ b/main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Actions/MiscActions.cs
@@ -128,10 +128,10 @@ namespace Mono.TextEditor
 			}
 			var line = data.Document.GetLine (data.Caret.Line);
 			if (line != null) {
-				if (line.Length == 0 && data.Caret.Column > 1) {
-					data.Caret.Column = 1;
-				} else {
-					data.Document.ApplyTextChanges (new [] { RemoveTabInLine (data, line) });	
+				using (var undo = data.OpenUndoGroup ()) {
+					data.EnsureCaretIsNotVirtual ();
+					data.Document.ApplyTextChanges (new [] { RemoveTabInLine (data, line) });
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.TextEditor.Tests/MonoDevelop.TextEditor.Tests.csproj
```diff
@@ -70,7 +70,6 @@
     <ProjectReference Include="..\..\addins\MonoDevelop.SourceEditor2\MonoDevelop.SourceEditor.csproj">
       <Project>{F8F92AA4-A376-4679-A9D4-60E7B7FBF477}</Project>
       <Name>MonoDevelop.SourceEditor</Name>
-      <Private>False</Private>
```
Therzok: Why remove this?

mkrueger:
Note: With local copy flag the tests don't run in the editor the source editor dll and it's deps need to be copied to the test directory. So if it's not breaking on the build bot it's a change by purpose for beeing able to run tests inside monodevelop properly.

main/src/core/MonoDevelop.TextEditor.Tests/MonoDevelop.TextEditor.Tests.csproj
mkrueger: With local copy flag the tests don't run in the editor the source editor dll and it's deps need to be copied to the test directory. So if it's not breaking on the build bot it's a change by purpose for beeing able to run tests inside monodevelop properly.

Therzok: The thing is, GuiUnit resolves all the deps from the actual mono-addins initialization.

If we need a local copy, the addin is busted (no right dependencies/runtime imports) and we're only loading the right deps by chance

mkrueger:
Y I see the build bot has problems - but in any case writing tests this way is not really possible so the local copy flag needs to be set/unset every time.

PR #5402, opened 2018-07-12
Unify on MonoDevelop/Ide/Composition extension point.

KirillOsenkov:
There were two legacy extension points: MefHostServices and PlatformMefHostServices. There is no need to differentiate.

I will remove the two legacy extension points in 16.0 after we make sure nobody external uses them.


KirillOsenkov:
We don't need this in release-7.6.

PR #5403, opened 2018-07-12
Normalize whitespace in VSEditor files.

KirillOsenkov:
Use tabs, LF (\n) line breaks, MonoDevelop formatting options, remove BOM.

This change is whitespace only.


mkrueger:
+1 I always hated that.

But IMO we've a problem regarding to the formatting - if we take VS or members using VS to change our code should use the monodevelop formatting style. If I open a PR against roslyn it's required to follow everything 100% and we're taking code with all formatting & naming rules.

For example the underscore prefix for private members is no monodevelop style. 

As an option we could switch monodevelop to the VS style - would make things easier in the future however that would basically destroy much of the blame history unless we support whitespace change skipping in our version control backend.

 /discuss

mhutch:
Why not put an `.editorconfig` file in the `VSEditor` directory to give it VS style?

KirillOsenkov:
Regarding the BOM - I'm OK either way, I just think no BOM UTF-8 is a good thing by convention? 

Let me try the .editorconfig approach in a separate PR.

KirillOsenkov:
Just done some research about BOM - this thread has convincing arguments for BOM and Roslyn uses BOM as well:

But maybe Mikayla is right and we should just go with an .editorconfig file for VSEditor folder. Here's the PR: 
https://github.com/mono/monodevelop/pull/5418

mhutch:
Yup, assuming UTF8 without a BOM is great in theory... unless you have to work with Windows

mkrueger:
UTF8 without BOM doesn't look good in theory :). If you ask me every file that can have a BOM should have one.

But even if we put a .editorconfig file there it's messy to mix different formatting styles inside the same project.

KirillOsenkov:
OK agreed to have a BOM everywhere.

So shall we abandon this PR then? I've just committed https://github.com/mono/monodevelop/pull/5418, which "legalizes" the formatting used by the VSEditor folder.

PR #5405, opened 2018-07-13
Fixes VSTS Bug 515119: Accessibility:MAS36:KB: Unable to see mouse ov…

mkrueger:
…er content when using the keyboard

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/515119

We now have a keyboad command for showing tooltips however the code segment tooltips were not showable by that. I changed the command so that the code segment tooltips can be queried when placing the caret before the folding.

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.cs
@@ -670,7 +670,7 @@ namespace Mono.TextEditor
 
 		void DisposeGCs ()
 		{
-			ShowTooltip (TextSegment.Invalid, Gdk.Rectangle.Zero);
+			ShowCodeSegmentPreviewTooltip (TextSegment.Invalid, Gdk.Rectangle.Zero);
 		}
 
 		int underlinePosition, underLineThickness;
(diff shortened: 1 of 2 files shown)


main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextViewMargin.cs
```diff
@@ -2523,15 +2523,15 @@ internal void OpenCodeSegmentEditor ()
...
+		internal void ShowCodeSegmentPreviewTooltip (ISegment segment, Rectangle hintRectangle, uint timeout = 650)
 		{
 			if (previewWindow != null && previewWindow.Segment.Equals (segment))
 				return;
 			CancelCodeSegmentTooltip ();
 			HideCodeSegmentPreviewWindow ();
 			if (segment.IsInvalid () || segment.Length == 0)
 				return;
-			codeSegmentTooltipTimeoutId = GLib.Timeout.Add (650, delegate {
+			codeSegmentTooltipTimeoutId = GLib.Timeout.Add (timeout, delegate {
```
Therzok: Looking around, this variable doesn't exist?

mkrueger: look at the parameter list and you'll see it :). Keyboard toggle doesn't need timeouts they should appear on key press.

slluis:
@iainx can you review and test?

PR #5407, opened 2018-07-13
[Editor] Ensure preedit committed before all selections actions

bratsche:
This is a follow-up to https://github.com/mono/monodevelop/pull/5083

That PR fixed the issue when holding shift and pressing an arrow key
to change the selection when a multi-key language such as Korean is
being used, but the issue persists when changing the selection in
other ways. For example, Cmd-Shift-Left arrow key will cause the
selection to move back as expected, but it will drop the last character
of the selection (and then when you unselect the buffer will be
broken in totally unexpected ways).

This PR committs the preedit for the necessary action, plus all other
selection actions to be on the safe side.

Fixes VSTS #635558

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/ExtensibleTextEditor.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/ExtensibleTextEditor.cs
@@ -730,11 +730,6 @@ namespace MonoDevelop.SourceEditor
 			info.Bypass = HasFocus == false;
 		}
 
-		void EnsurePreeditCommitted()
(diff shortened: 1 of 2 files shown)


Therzok:
One other solution I can think of would be committing the preedit string here: http://source.monodevelop.com/#MonoDevelop.SourceEditor/SelectionActions.cs,78

slluis:
@bratsche maybe you can just call `TextEditorData.Parent.CommitPreedit()`. @mkrueger what do you think?

bratsche:
@slluis I totally overlooked `Parent`. Yes, that works. :)

Therzok:
Works for me!

slluis:

PR #5409, opened 2018-07-13
Custom projects in build output

rodrmoya:
> [Fixes VSTS #623973](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/623973)


slluis (changes requested):
This needs to be solved in SolutionItem, so that any subclass can generate output without changes.

rodrmoya:
hmm, override OnBuild in SolutionItem you mean? Doing so just does nothing, as subclasses seem to override that method, so the base class one is never called. So, @slluis you mean having all subclasses call the base method?

Issue #3869, opened 2018-02-15
"[TestCaseSource ("xzy")]" isn't seen as a test in the text editor. No dot next to it.


decriptor:
Methods with the `[Test]` or `[TestCase ()]` attribute on it have a dot in the margin that allows you to execute a test. However, a method with the `[TestCaseSource()]` attribute doesn't have a dot in the margin.
<img width="294" alt="screenshot 2018-02-15 13 44 17" src="https://user-images.githubusercontent.com/49821/36280053-59f9e13c-1256-11e8-8866-f4e4c65d2c12.png">

> VS bug #569964

decriptor:
TestCaseSource docs -> https://github.com/nunit/docs/wiki/TestCaseSource-Attribute

xamarin-release-manager:
Fixed in version 7.7.0.628 (master)

Pull Request #5412 merged by: Lluis Sanchez
Author: mono
Commit: 43a85f8419405146f239dd96b345d0eb9012369b (mono/monodevelop)

PR #5412, opened 2018-07-16
Fixes issue #3869 "[TestCaseSource ("xzy")]" isn't seen as a test in

mkrueger:
the text editor. No dot next to it.

TODO: Merge interfaces in the next AP break.

### DIFF
--- a/main/src/addins/MonoDevelop.UnitTesting/Services/AbstractUnitTestEditorExtension.cs
+++ b/main/src/addins/MonoDevelop.UnitTesting/Services/AbstractUnitTestEditorExtension.cs
@@ -386,5 +386,17 @@ namespace MonoDevelop.UnitTesting
 		/// <value>The ignore test method attribute marker.</value>
 		string IgnoreTestClassAttributeMarker { get; }
 	}
+
+	/// <summary>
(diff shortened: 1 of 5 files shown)


main/src/addins/MonoDevelop.UnitTesting/Services/AbstractUnitTestEditorExtension.cs
```diff
@@ -386,5 +386,17 @@ public interface IUnitTestMarkers
 		/// <value>The ignore test method attribute marker.</value>
 		string IgnoreTestClassAttributeMarker { get; }
 	}
+
+	/// <summary>
+	/// TODO: Merge with IUnitTestMarkers - possible replace it with an abstract class in next API break.
+	/// </summary>
+	public interface IUnitTestMarkers2 : IUnitTestMarkers
```
Therzok: Can't we make this interface internal?

mkrueger: Would still be an API change. I'm not sure why that's an extensible thing to begin with.

Therzok:
Also, unit tests?

Issue #4110, opened 2018-03-09
Inconvenient behavior of Ctrl+Right and Ctrl+Left


jkpubsrc:
Pressing Ctrl plus Cursor-Right or Cursor-Left will let you skip too much code in the edit component. MonoDevelop exhibits a quite strange behavior here compared to other IDEs and general text editors.

Here is a more detailed analysis of the situation describing stops that occur if pressing Ctrl+Right:

```
foo ( ) [ ] { }  bar
```

* MonoDevelop:
    * after "foo"
    * after "bar"
* Pluma:
    * after "foo"
    * after "bar"
* Notepad (Mono):
    * before "("
    * before "["
    * before "{"
    * before "bar"
* VisualStudioCode:
    * after "foo"
    * after "("
    * after ")"
    * after "["
    * after "]"
    * after "{"
    * after "}"
    * after "bar"
* Github text edit control in FireFox:
    * after "foo"
    * before ")"
    * before "["
    * before "]"
    * before "{"
    * before "}"
    * after "bar"
* vi
    * before "("
    * before ")"
    * before "["
    * before "]"
    * before "{"
    * before "}"
    * before "bar"
* NetBeans:
    * after "foo"
    * before "("
    * before ")"
    * before "["
    * before "]"
    * before "{"
    * before "}"
    * before "bar"
    * after "bar"
* LibreOffice Writer:
    * before "("
    * before ")"
    * before "["
    * before "]"
    * before "{"
    * before "}"
    * before "bar"
    * after "bar"

As you can see MonoDevelop behaves more than a regular text editor than as an IDE. We software developers have special needs in regard to backets and other delimiters. Even LibreOffice is more convenient in that regard.

This is extremely annoying if you want to mark something by pressing Shift as well. If you have comment lines starting with "//" - even multiple comment lines - it can happen that a single press of Ctrl+Right will pass through many many source code lines. You can NOT mark a single line comment conveniently.

I'd personally expect a Ctrl+Right to somehow ...

* stop before "//" which mark the beginning of a line comment (as I'd like to mark them sometimes which can't be done no as easily as in other IDEs)
* stop before "("
* stop before "{"
* stop before "."
* stop before any word.

For example Visual Studio Code seems to stop whenever some character is followed by a whitespace. If no whitespaces occur stops are performed before "[", before "(", after "}" and before ".". This might be debatable but it's much more convenient than the behavior in MonoDevelop where there is almost no stop.

[...]

PR #5420, opened 2018-07-17
Fixes issue #4110 Inconvenient behavior of Ctrl+Right and Ctrl+Left

mkrueger:
In theory we could switch to all vs.net editor operations however
we've many mac behavior commands built in so we would break some mac
specific behavior. But it's ok for ctrl+left/right there are no mac
specifics here.

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/MonoTextEditor.ITextView.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/MonoTextEditor.ITextView.cs
@@ -40,6 +40,8 @@ using System.Diagnostics;
 using MonoDevelop.Ide;
 using Microsoft.VisualStudio.Text.Classification;
 using System.Threading;
+using Microsoft.VisualStudio.Text.Operations.Implementation;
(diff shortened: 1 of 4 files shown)


main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/MonoTextEditor.ITextView.cs
```diff
@@ -349,8 +353,8 @@ public void DisplayTextLineContainingBufferPosition (SnapshotPoint bufferPositio
 		}
 
 		public SnapshotSpan GetTextElementSpan (SnapshotPoint point)
-		{
-			throw new NotImplementedException ();
+		{
+			return new SnapshotSpan (point, point);
```
Therzok: Shouldn't we be looking for a word start/end at a given point?

mkrueger:
@Therzok : It's not what this method does. For the common case it's just returning the extent of 1 character. However I added the support of multibyte characters there and implemented the GetTextElementSpan for the line collection and line.

One thing it's currently missing is the support of non printable characters but that's not in caret move actions either. But multibyte should now work.

PR #5421, opened 2018-07-17
[Core] Refresh type system when configuration mapping changes

mrward:
Changing a solution configuration mapping for a project in the
Solution Options, such as changing a project so its Debug|x86
configuration is mapped to Debug|AnyCPU, would not refresh the type
system. If the project had conditional code based on the configuration
then this would not be highlighted correctly in the text editor until
the solution was closed and re-opened.

When the solution configurations are updated by the Solution Options
dialog the project configurations are updated. When these project
configurations updated they now raise the Project's Modified event
and pass the CompilerParameters hint text. This then causes the
type system to be refreshed for the project.

Problems with this change:

Project.Modified event is fired once for every configuration when
it is updated - this will trigger the type system to start and then
cancel a project reload.

Fixes VSTS #589733 - Configuration mapping change for a project does
not notify modifications


Therzok (approved):

mrward:

Issue #4390, opened 2018-03-29
Cmd does not light up navigation for symbol under mouse


mhutch:
If I have the symbol on a mouse and hold down cmd, the navigation link link does not light up for the symbol under the mouse. If I move the mouse, it lights up.

> VS bug [#593843](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/593843)

xamarin-release-manager:
Fixed in version 7.7.0.716 (master)

Pull Request #5436 merged by: Mike Krüger
Author: mono
Commit: 0f8f19f9ac14d8513448ddae2ca13aacb2443807 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #5436 merged by: Mike Kr??ger
Author: mono
Commit: 0f8f19f9ac14d8513448ddae2ca13aacb2443807 (xamarin/monodevelop)

Issue #4389, opened 2018-03-29
Cmd-click navigation sometimes gets stuck


mhutch:
Sometimes when I use Cmd-click to navigate, the editor gets "stuck" in navigation mode after I let go of Cmd: the underlines keep lighting up, and clicking on a symbol navigates. This is extremely annoying as I can no longer select text. If I switch to another editor tab and back it starts working again.

> VS bug [#593842](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/593842)

mkrueger:
I've problems reproducing that - is there any pattern to trigger that behavior ?

mhutch:
I haven't found a repro but i still see this from time to time.

decriptor:
I don't a repro, but I do notice issues with it from time to time when working on the `xamarin/designer` Xamarin.AndroidDesigner solution.

xamarin-release-manager:
Fixed in version 7.7.0.716 (master)

Pull Request #5436 merged by: Mike Krüger
Author: mono
Commit: 0f8f19f9ac14d8513448ddae2ca13aacb2443807 (mono/monodevelop)

nosami:
I just saw this bug in the release notes for 7.7. I mentioned this bug to Mike before. I used to be able to repro this every time by holding cmd down in the editor so that the underline would show, then without letting go, hold Cmd+Tab to tab away to any other app and then back again.

I can confirm that it's now fixed.

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #5436 merged by: Mike Kr??ger
Author: mono
Commit: 0f8f19f9ac14d8513448ddae2ca13aacb2443807 (xamarin/monodevelop)

PR #5436, opened 2018-07-18
Fixes issue #4390 Cmd does not light up navigation for symbol under

mkrueger:
mouse Fixes issue #4389 Cmd-click navigation sometimes gets stuck

Improved the reliablility of teh cmd navigation. There were duplicate
hover segments and some possible issues with the event system that may
have caused #4389.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/AbstractNavigationExtension.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/AbstractNavigationExtension.cs
@@ -39,7 +39,7 @@ namespace MonoDevelop.Ide.Editor.Extension
 	{
 		uint timerId;
 		CancellationTokenSource src = new CancellationTokenSource ();
-
+		
 		#region Key handling
 		static bool linksShown;
 
@@ -160,6 +160,7 @@ namespace MonoDevelop.Ide.Editor.Extension
 				ShowLinks ();
 		}
 
+		object markerLock = new object ();
 		List<ITextSegmentMarker> markers = new List<ITextSegmentMarker> ();
 		List<IDocumentLine> visibleLines = new List<IDocumentLine> ();
 		void ShowLinks ()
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/AbstractNavigationExtension.cs
```diff
@@ -202,18 +203,24 @@ void ShowLinks ()
...
 					try {
-						foreach (var segment in segments) {
-							if (token.IsCancellationRequested) {
-								return;
+						lock (markerLock) {
+							foreach (var segment in segments) {
+								if (token.IsCancellationRequested) {
+									return;
+								}
+								if (markers.Any (m => m.Offset == segment.Offset && m.Length == segment.Length))
```
Therzok: We should avoid LINQ here if possible, it's a lambda with a capture in a loop, on the UI thread.

Otherwise, LGTM.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/AbstractNavigationExtension.cs
```diff
@@ -202,18 +203,24 @@ void ShowLinks ()
...
-				if (segments == null || token.IsCancellationRequested)
+				if (segments == null || token.IsCancellationRequested) {
 					return;
+				}
 				await Runtime.RunInMainThread(delegate {
 					try {
-						foreach (var segment in segments) {
-							if (token.IsCancellationRequested) {
-								return;
+						lock (markerLock) {
```
Therzok: Can we use a non-blocking lock here? A `SemaphoreSlim(1, 1)`? Would be great to not hang the UI.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Extension/AbstractNavigationExtension.cs
Therzok: It's not about the collection size. For every segment, you're going to allocate a lambda with a capture. This is also on the UI thread.

ganorkargaurav:
Verified on Visual Studio Enterprise 2017 for Mac (Preview) Version 7.7 Preview (7.7 build 606).
Marked as QA-Approved.
Build taken from: https://github.com/mono/monodevelop/pull/5436/commits
Screencast link: http://www.screencast.com/t/CZmLWMYOp2w

**Approved for Merging**

PR #5440, opened 2018-07-18
[MEF] Change the API around GetExports a bit so we don't hit the comp…

Therzok:
…osition until needed

The API is not exactly the same as the MEF one, but it achieves the same thing.

Once exports are still lazily evaluated.

Fixes VSTS #650169 - MEF is queried before idle is hit


Therzok:

PR #5445, opened 2018-07-18
[NuGet] Ensure generated files are available for code completion

mrward:
The LibLog NuGet package has contentFiles that are processed by
MSBuild and converted into .cs files that are implicitly included
in the project. These files are generated into the obj folder,
for example:

    obj/Debug/netstandard2.0/NuGet/SomeID/LibLog/5.0.0/ILog.cs

These types defined inside these generated files can be referenced by
code in the project and the project will compile but the text editor
would show an error.

These generated files are not created or returned by running the
CoreCompileDependsOn target. So to fix this the NuGet addin has
a project extension that overrides the OnGetSourceFiles method and
modifies the targets being called, when the CoreCompileDependsOn
target is evaluated, so that NuGet specific targets are also run to
generate the files and returns them to the type system.

Fixes VSTS #636061 - NuGet source transformations not recognized by
intellisense


mrward:

mrward:
Going to look at simplifying the NuGet project extension logic:

- Just override OnRunTarget.
- First check that Compile items are being evaluated.
- Then reproduce the logic in [Project](https://github.com/mono/monodevelop/blob/0be2d6c56b1ef6a1126b902fb2c8ec36bbdb590c/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs#L614-L622) to get the CoreCompileDependsOn targets and compare that with the targets being run in OnRunTarget.
- Then check if the project is a .NET Core project or the project has PackageReferences.
- Then modified the targets and evaluation context properties as needed.

PR #5446, opened 2018-07-18
Fix #5427: Unclassified spans in editor default to `entity.name.class`

### LINKED ISSUE #5427: Unclassified spans in editor default to `entity.name.class`
https://github.com/mono/monodevelop/pull/5102 introduced the behavior where the unclassified parts of the editor get the type name classification.

For webtools, HTML doesn't classify the text inside elements. It used to be black, but now it is the color of type names.

> VS bug [#649777](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/649777)

DavidKarlas:
I used `userScope` before because thats RoslynSyntaxHighligher(what we used for coloring C# code had), having `Foregroud` as default makes much more sense imo...


KirillOsenkov:

Issue #4751, opened 2018-05-08
[ProgressMonitors] Dispose is not called on the right sync context


Therzok:
Dispose should be called on the right synchronization context, but it is not possible to do so at the moment.

https://github.com/mono/monodevelop/blob/bbc80493465ff7c7e32b51c3d4b25ed751859610/main/src/core/MonoDevelop.Core/MonoDevelop.Core/ProgressMonitor.cs#L134

Dispose context is controlled by the caller, so we can't marshal to any other sync context.

Consider breaking API and changing to:
```
public void Dispose ()
{
...
}

protected virtual void OnDisposed()
{
...
}
```

And using the synchronization context. This is currently a bad case for UI progress monitors disposed from bg threads.

> VS bug [#612543](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/612543)

xamarin-release-manager:
Fixed in version 7.7.0.690 (master)

Author: therzok
Commit: fdaef70c331ffe5580ce9d88b9aba4a3aaa1f348 (mono/monodevelop)

Therzok:
Shouldn't have been closed.

xamarin-release-manager:
> [sync] [VS-18] Comment by **xamarinc**



Fixed in version 8.0.0.778 (release-8.0)



Author: Marius Ungureanu  
 Commit: 276c2bcdc505d249d48f953d6233a3b7a1f1229e (mono/monodevelop)

xamarin-release-manager:
> [sync] [VS-20] Comment by **xamarinc**



Fixed in version 8.0.0.1001 (master)



Author: Marius Ungureanu  
 Commit: 276c2bcdc505d249d48f953d6233a3b7a1f1229e (mono/monodevelop)

xamarin-release-manager:
> [sync] [VS-21] Comment by **xamarinc**



Fixed in version 8.0.0.1738 (release-8.0)



Author: Marius Ungureanu  
 Commit: 276c2bcdc505d249d48f953d6233a3b7a1f1229e (xamarin/monodevelop)

xamarin-release-manager:
> [sync] [VS-25] Comment by **xamarinc**



Fixed in version 8.6.0.802 (master)



Author: Marius Ungureanu  
 Commit: 276c2bcdc505d249d48f953d6233a3b7a1f1229e (xamarin/monodevelop)

PR #5447, opened 2018-07-18
[VCS] Fixup more UI thread semantics and make Wakeup use the UI context

Therzok:
This was a race condition. the UI lock-up would happen when dispose
would finish before the EndTask did.

The reason here is that the Dispose() method on ProgressMonitor
does not take into account the monitor's synchronization context.

See https://github.com/mono/monodevelop/issues/4751

Second attempt at:
Fixes VSTS #641132 - [Feedback] Visual Studio Mac checkout functionality freezes entire program
Fixes VSTS #619788 - Error when cloning new Git repository locks IDE main menu

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/CheckoutCommand.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/CheckoutCommand.cs
@@ -53,15 +53,22 @@ namespace MonoDevelop.VersionControl
 				);
 			}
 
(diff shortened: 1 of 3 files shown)


Therzok:
@slluis I don't like the fix, but there's nothing we can do until we fix the ProgressMonitor API.

What happens is:
* Dispose is context-less
* Dispose happens first for follower monitors in an aggregated monitor
* Then an OnCompleted callback is queued for the UI thread which happens after the monitor disposes

Why it doesn't fail in the current state, I have no clue. The dialog appears then disappears for Matt, doesn't for me.

Therzok:

PR #5473, opened 2018-07-20
[Ide] Don't show welcome page when opening a specific item

Therzok:
Showing the welcome page delays loading the IDE (querying the welcome page
service and so on).

Fixes VSTS #652347 - Welcome page should not show when loading a solution from finder

Still working on the UX, trying to get the welcome page to show while the workspace is loading, but also trying to prevent a race.

screen shot 2018-07-20 at 23 33 44

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
@@ -205,7 +205,9 @@ namespace MonoDevelop.Ide
 			}
 		}
 
-		public static void Initialize (ProgressMonitor monitor)
+		public static void Initialize (ProgressMonitor monitor) => Initialize (monitor, showWelcomePage: true);
+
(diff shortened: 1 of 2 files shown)


Therzok:
Trying to make it so the welcome page displays after the workspace load triggers, so we don't have empty UI while it's loading.

Therzok:
So, given current limitations, I'm removing do-not-merge. This can go in if @vancura ok's the UX.

The upside is that unity project loads are faster and snappier by some margin, compared to the UX where we show the welcome page first, but the screen shows up as 'empty'

Therzok:
@mrward works for you?

mrward:
Test failures are unrelated.

Therzok:
@mrward That is not a use-case I considered for this. We can file a new bug tracking that use-case.

Therzok:
Filed as https://github.com/mono/monodevelop/issues/5553

Therzok:

sgmunn:

Therzok:

PR #5484, opened 2018-07-23
[StatusService] Separate the status from the statusbar

iainx:
Create a StatusService which keeps track of multiple status messages and how they change. The Statusbar simply becomes one of many possible listeners to this StatusService.

Obsoletes #3100


iainx:
Rebased to master and moved it to the mono repo so VSTS will do its thing

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/StatusBarContextHandler.cs
```diff
@@ -25,72 +25,99 @@
...
 
-		internal StatusBar StatusBar {
-			get;
-			private set;
+		readonly List<StatusMessageContext> activeContexts = new List<StatusMessageContext> ();
+
+		public StatusBarContextHandler ()
+		{
+			StatusService.ContextAdded += StatusServiceContextAdded;
+			StatusService.ContextRemoved += StatusServiceContextRemoved;
```
Therzok: StatusService is static. If we don't unsubscribe the event handlers, we'll leak all the context handlers.

I guess the solution is to make the context handler idisposable

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.StatusService/StatusService.cs
```diff
@@ -0,0 +1,106 @@
...
+// THE SOFTWARE.
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Timers;
+using MonoDevelop.Core;
+
+namespace MonoDevelop.Ide.Status
+{
+	public static class StatusService
```
Therzok: Can we make it so this is mockable? Would help a lot with testing.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.StatusService/StatusMessageContext.cs
```diff
@@ -0,0 +1,224 @@
...
+using System;
+using System.Threading;
+using MonoDevelop.Core;
+using MonoDevelop.Ide.Gui;
+
+using StockIcons = MonoDevelop.Ide.Gui.Stock;
+
+namespace MonoDevelop.Ide.Status
+{
+	public class StatusMessageContext : IDisposable
```
Therzok: Anyone can dispose the MainContext with this API.

Maybe have a base class that doesn't implement IDisposable and use that for MainContext?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.StatusService/StatusService.cs
```diff
@@ -0,0 +1,106 @@
...
+			var ctx = new StatusMessageContext ();
+			contexts.Add (ctx);
+
+			OnContextAdded (ctx);
+			return ctx;
+		}
+
+		public static StatusBarIcon ShowStatusIcon (Xwt.Drawing.Image pixbuf)
+		{
+			return IdeApp.Workbench.StatusBar.ShowStatusIcon (pixbuf);
```
Therzok: We need to ensure Workbench is initialized here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.StatusService/StatusService.cs
```diff
@@ -0,0 +1,140 @@
...
+
+namespace MonoDevelop.Ide.Status
+{
+	public interface IStatusService
+	{
+		event EventHandler<StatusServiceContextEventArgs> ContextAdded;
+		event EventHandler<StatusServiceContextEventArgs> ContextRemoved;
+		event EventHandler<StatusServiceStatusImageChangedArgs> StatusImageChanged;
+
+		StatusMessageContext MainContext { get; }
```
Therzok: We seem to have lost documentation on these properties and methods.

PR #5486, opened 2018-07-23
Remove Stetic Gtk#2 GUI designer

directhex:
Stetic is what sold me on MonoDevelop in the first place, more than a decade ago. A high quality GUI designer, fully integratwd into the IDE, on Linux, was bliss - as compared to the state of the art of `glade-2` XML files at the time.

However.

In 2018, it is effectively dead weight for us.

Visual Studio for Mac is not a Gtk# IDE, and does not benefit from Gtk# app authoring support. And, crucially, Stetic depends heavily on the internals of libglade, which no longer exists in Gtk+ 3.0+.

In order to move forwards with porting MonoDevelop to Gtk#3, Stetic in its current incarnation needs to go away. All the internals need rewriting, and we cannot justify the engineer time required to do that ourselves. So, like our C/C++ support, remove it & focus on the fundamentals.

This removal removes support for creating new Gtk# projects (as the templates are in MonoDevelop.GtkCore), but it can still open/compile existing Gtk# projects fine.


Therzok:
This is okay by me. @slluis ?

PR #5497, opened 2018-07-24
[Core] Fix project run configurations not used when running solution

mrward:
Creating a new .NET Core console project, editing the project run
configuration to use extra command line arguments or to not use the
external console, then building and running the solution would result
in the project run configuration not being used. So no extra
arguments would be passed to the console project, and the external
console would still be used. This could be fixed by closing and
re-opening the solution.

The problem was that when the project is re-evaluated when it is
initially created its run configurations are cleared. The solution's
StartupConfiguration has a run configuration that was still using
the original project run configuration that was no longer used. So
changes to the project run configuration had no affect when running
the solution since the two project run configurations were not the
same object. Closing and re-opening the solution fixed this since
the run configuration in the .csproj.user file is re-used when the
project is re-evaluated so both the solution run configuration and
the project run configuration refer to the same object. To fix this,
on re-evaluating the project, if the solution's run configuration
referred to a project run configuration which has been removed then
the solution's startup configuration is refreshed.

Note that there is a similar problem with multiple solution run
configurations that can occur which is not addressed by this fix.

Fixes VSTS #602834 - Unchecking Use External Console/passing command
line args has no affect for .NET Core project

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -4103,6 +4103,8 @@ namespace MonoDevelop.Projects
 					var oldCapabilities = new HashSet<string> (projectCapabilities);
 					bool oldSupportsExecute = SupportsExecute ();
 
+					var solutionStartupProjectRunConfig = GetSolutionStartupProjectRunConfigurationForThisProject ();
+
 					try {
 						IsReevaluating = true;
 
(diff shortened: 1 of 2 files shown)


Therzok (approved):

ganorkargaurav:
Verified on Visual Studio Enterprise 2017 for Mac (Preview) Version 7.7 Preview (7.7 build 731). 
It is working fine. No external window opened. Output shown in Application Output window.
QA-Approved.

Steps Performed:
1) Create a new .NET Core 1.1 Console project. 
2) Run - Start Debugging.
3) Opens external console window as expected.
4) Open project options - Run - Configurations - Default and uncheck Use External Console. OK the dialog.
5) Run - Start Debugging.

Screencast: https://screencast.com/t/XFHtitdH
Build Gist: https://gist.github.com/parasAmbhore/2575ede3f31d37ed4b88f17708743736

**Approved for Merging**

mrward:

PR #5507, opened 2018-07-24
[Ide] Fix no files added to new project 

mrward:
Creating a Xamarin.Mac project inside a directory with other existing
projects could cause the project to be created without any files added
to the project but the files were created on disk.

The problem is that the ProjectDescriptor.InitializeItem method was
an async void method and when the Info.plist was created it called
into the EditorConfigService.GetEditorConfigContext when then called
codingConventionsManager.GetConventionContextAsync which presumably
was using a Task.Run in certain cases. This meant the UI thread was
unblocked and New Project dialog would then save the project before
all the files were created. Resulting in the files not being added
to the project file.

To fix this the InitializeItem method now returns a Task that is
awaited. One problem here that is not fixed is that there is a
TemplateInitializationCallback when adding a new project to an
existing solution that may run after the project is saved in a similar
way as for a new project. The TemplateInitializationCallback calls
the project descriptor's InitializeItem method. Note that this delay
does not seem to happen in practice. Changing the
TemplateInitializationCallback to be Task based is a more complicated
change since most of this API is public so has been left for now.
Note that with a Task.Run and delay added to the project descriptor's
InitializeItem and adding a new project to a solution the files are
added to the project and displayed in the Solution pad but the project
is not saved. This happens with or without the fix applied here.

Fixes VSTS #645418 - The project is empty after creation

### DIFF
--- a/main/tests/Ide.Tests/ProjectTemplateTests.cs
+++ b/main/tests/Ide.Tests/ProjectTemplateTests.cs
@@ -102,7 +102,22 @@ namespace MonoDevelop.Ide
 
 			var sharedAssetsProject = template.CreateProjects (solution.RootFolder, cinfo)
 				.OfType<SharedAssetsProject> ().Single ();
-			var myclassFile = sharedAssetsProject.Files.First (f => f.FilePath.FileName == "MyClass.cs");
+
+			// Template initialization is now asynchronous so we need to wait and unblock the UI thread
+			// to ensure the template finishes adding files to the project.
+			const int timeout = 2000; // ms
+			int waitedTime = 0;
+			int delayTime = 100;
+			ProjectFile myclassFile = null;
+			while (myclassFile == null) {
+				myclassFile = sharedAssetsProject.Files.FirstOrDefault (f => f.FilePath.FileName == "MyClass.cs");
+				if (myclassFile == null) {
+					if (waitedTime >= timeout)
+						Assert.Fail ("Timed out waiting for file to be added to shared project.");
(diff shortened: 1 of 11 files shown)


mrward:
One of the ide tests is failing which I need to investigate. I did run them locally at one point and I thought they were working.

main/tests/Ide.Tests/ProjectTemplateTests.cs
```diff
@@ -102,7 +102,22 @@ public override void TearDown ()
...
-			var myclassFile = sharedAssetsProject.Files.First (f => f.FilePath.FileName == "MyClass.cs");
+
+			// Template initialization is now asynchronous so we need to wait and unblock the UI thread
+			// to ensure the template finishes adding files to the project.
+			const int timeout = 2000; // ms
+			int waitedTime = 0;
+			int delayTime = 100;
+			ProjectFile myclassFile = null;
+			while (myclassFile == null) {
+				myclassFile = sharedAssetsProject.Files.FirstOrDefault (f => f.FilePath.FileName == "MyClass.cs");
```
Therzok: Not as part of this PR, but maybe we could have a utility method that does polling for the test suite:
```
async Task PollWait (Func<bool> assertion, int timeout, int delayBetweenPolls)
{
     int waitedTime = 0;
     while (assertion() == false) {
           if (waitedTime >= timeout)
               Assert.fail();
           await Task.Delay (delayTime);
           waitedTime += delayTime;
     }
}
```

mrward:

Issue #4996, opened 2018-06-04
Show breakpoints and caret location in scrollbar


mhutch:
It would be really nice to show breakpoints and caret location in the scrollbar. These are significant features that I often want to go looking for after scrolling away.

> VS bug [#643116](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/643116)

xamarin-release-manager:
Fixed in version 7.8.0.61 (master)

Author: Mike Krüger
Commit: 9ca1749ab0135f3ca43a1181a0567690c4314e73 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 7.8.0.126 (release-7.7)

Author: Mike Krüger
Commit: 9ca1749ab0135f3ca43a1181a0567690c4314e73 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 7.7.0.931 (release-7.7)

Author: Mike Krüger
Commit: 9ca1749ab0135f3ca43a1181a0567690c4314e73 (mono/monodevelop)

xamarin-release-manager:
> [sync] [VS-31] Comment by **xamarinc**



Fixed in version 8.6.0.802 (master)



Author: Mike Kr??ger  
 Commit: 9ca1749ab0135f3ca43a1181a0567690c4314e73 (xamarin/monodevelop)

PR #5509, opened 2018-07-25
Fixes issue #4996 Show breakpoints and caret location in scrollbar

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/643116

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.QuickTasks/QuickTaskOverviewMode.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.QuickTasks/QuickTaskOverviewMode.cs
@@ -726,11 +726,11 @@ namespace MonoDevelop.SourceEditor.QuickTasks
 			if (TextEditor.EditorTheme == null)
 				return;
 			drawnCaretLine = TextEditor.Caret.Line;
-			double y = GetYPosition (drawnCaretLine);
+			int y = (int)GetYPosition (drawnCaretLine);
 
 			cr.SetSourceColor (SyntaxHighlightingService.GetColor (TextEditor.EditorTheme, EditorThemeColors.Foreground));
-			var w = Allocation.Width * 0.618;
-			cr.Rectangle (0.5 + Allocation.Width - w, y - 1.5, w, 2);
+			var w = Math.Floor (Allocation.Width * 0.618);
+			cr.Rectangle (Allocation.Width - w, y - 1, w, 2);
 			cr.Fill ();
 		}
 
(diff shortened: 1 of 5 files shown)


mkrueger:
We may need to tweak the look - that's the 1st version :).

Therzok:
cc @vancura

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.QuickTasks/QuickTaskOverviewMode.cs
```diff
@@ -170,6 +172,11 @@ public QuickTaskOverviewMode (QuickTaskStrip parent)
 			CanFocus = true;
 		}
 
+		void BreakpointsChanged (object sender, Mono.Debugging.Client.BreakEventArgs e)
+		{
+			QueueDraw ();
```
Therzok: Can't we just redraw where the breakpoint will be?

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.QuickTasks/QuickTaskOverviewMode.cs
```diff
@@ -701,19 +713,42 @@ int YToLine (double y)
 			return TextEditor.GetTextEditorData ().VisualToLogicalLine ((int)line);
 		}
 
+
+		void Caret_PositionChanged (object sender, CaretLocationEventArgs e)
+		{
+			if (drawnCaretLine != TextEditor.Caret.Line)
```
Therzok: Same here, can't we just draw what we need to draw?

vancura:
Nice, thanks!

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.QuickTasks/QuickTaskOverviewMode.cs
mkrueger: Not really worth on breakpoint change. It redraws on analyzer change way more frequently.

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.QuickTasks/QuickTaskOverviewMode.cs
mkrueger: Overdraws things would require background buffering - but here it may be worth it (not for the breakpoints change btw). 

Depends on how fast the drawing is - at least it only repaints on line change.

PR #5517, opened 2018-07-26
Fixes issue #4110 Inconvenient behavior of Ctrl+Right and Ctrl+Left

mkrueger:
Behavior on mac/linux seems to differ a bit.


Therzok:
I don't like the `WaitAndGetResult()`

PR #5529, opened 2018-07-27
[Ide] Fix null ref logged by telemetry when creating NuGet project

mrward:
The language name for a NuGet packaging project is not set, since
it does not have a language. This was causing a null reference
exception to be logged by the telemetry when it attempted to call
ToString() on the value.

Fixes VSTS #655218 - Null reference exception logged on creating a
NuGet packaging project


mrward:

PR #5531, opened 2018-07-27
[PerformanceDiagnostics] Fix UI thread hang app not starting

mrward:
The UI thread hang console app was being started without quoting
the path to the executable so it was failing to start when inside
a path with a space, such as from within the 'Visual Studio.app'
bundle.

Fixes VSTS #655338 - Exception on quit


mrward:

PR #5561, opened 2018-08-01
Fix VSTS 656752 (Classification doesn't update)

KirillOsenkov:
Fix VSTS 645599 (ArgumentOutOfRange in Razor)

Update MdTextViewLineCollection to the latest editor Snapshot when buffer changes. We've had a number of issues in TagBasedSyntaxHighlighting where some lines were partially translated to the latest snapshot, but created a discrepancy because a SnapshotSpan would be tracked to latest, but a DocumentLine wouldn't get translated.

Now we re-add the lines we want to reuse against the latest snapshot, and the changed lines would be added later by the layout engine.

Removing the incorrect fix to translate lines as it didn't properly translate the DocumentLine of an MdTextViewLine.


KirillOsenkov:

KirillOsenkov:

PR #5568, opened 2018-08-02
Bump guiunit to a version which reports exceptions as exit code 1

Therzok:
Do not merge, will want to bump guiunit in master once we fix android tests running.

Using it as a test run to see if android tests now fail


Therzok:
Already bumped.

PR #5594, opened 2018-08-03
Always set the LangVersion to latest for C#

Therzok:
This allows us to use newer features like default literal,
new valuetype ref semantics, private protected, stackalloc arrays


mhutch:
I wonder whether an explicit value is better so it's clear which features can be used on the currently supported toolchains? That way the IDE/build will catch accidental use of preview language features early rather than waiting till it hits CI.

Therzok:
The reason I set `latest` is that I wanted defining the langversion specifically for all languages (C#, VB), but I guess I can do that

Therzok:

PR #5597, opened 2018-08-05
Extend support for Sdk definition in .NET Core

mathbourgeois:
Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=58564 (I couldn't find an equivalent on the Github issues)

MsBuild has 3 methods of defining an SDK:
- Sdk attribute on the Project node
- Sdk node as a child of the Project node
- Sdk attribute on any Import. This permits explicit imports

Currently, MonoDevelop only supports the first one. Adding support for the other two cases, including not doing any implicit importing if the project already does so.

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
@@ -78,7 +78,7 @@ namespace MonoDevelop.DotNetCore
 		/// </summary>
 		bool IsSdkProject (DotNetProject project)
 		{
-			return project.MSBuildProject.GetProjectSdksName() != null;
+			return project.MSBuildProject.GetReferencedSDKs().Length > 0;
 		}
 
 		protected override bool OnGetCanReferenceProject (DotNetProject targetProject, out string reason)
@@ -406,7 +406,7 @@ namespace MonoDevelop.DotNetCore
 
 		protected override void OnBeginLoad ()
 		{
-			dotNetCoreMSBuildProject.Sdk = Project.MSBuildProject.GetProjectSdksName();
+			dotNetCoreMSBuildProject.Sdk = Project.MSBuildProject.Sdk;
 			base.OnBeginLoad ();
 		}
 
@@ -416,7 +416,7 @@ namespace MonoDevelop.DotNetCore
 
 		protected bool IsWebProject (DotNetProject project)
 		{
-			return (project.MSBuildProject.GetProjectSdksName()?.IndexOf ("Microsoft.NET.Sdk.Web", System.StringComparison.OrdinalIgnoreCase) ?? -1) != -1;
+			return (project.MSBuildProject.GetReferencedSDKs()?.FirstOrDefault(x => x?.IndexOf ("Microsoft.NET.Sdk.Web", System.StringComparison.OrdinalIgnoreCase) != -1) != null);
 		}
 
 		public bool IsWeb {
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
@@ -78,7 +78,7 @@ namespace MonoDevelop.DotNetCore
 		/// </summary>
 		bool IsSdkProject (DotNetProject project)
 		{
-			return project.MSBuildProject.Sdk != null;
+			return project.MSBuildProject.GetProjectSdksName() != null;
 		}
 
 		protected override bool OnGetCanReferenceProject (DotNetProject targetProject, out string reason)
@@ -406,7 +406,7 @@ namespace MonoDevelop.DotNetCore
 
 		protected override void OnBeginLoad ()
 		{
-			dotNetCoreMSBuildProject.Sdk = Project.MSBuildProject.Sdk;
+			dotNetCoreMSBuildProject.Sdk = Project.MSBuildProject.GetProjectSdksName();
 			base.OnBeginLoad ();
 		}
 
@@ -416,7 +416,7 @@ namespace MonoDevelop.DotNetCore
 
 		protected bool IsWebProject (DotNetProject project)
 		{
-			return (project.MSBuildProject.Sdk?.IndexOf ("Microsoft.NET.Sdk.Web", System.StringComparison.OrdinalIgnoreCase) ?? -1) != -1;
+			return (project.MSBuildProject.GetProjectSdksName()?.IndexOf ("Microsoft.NET.Sdk.Web", System.StringComparison.OrdinalIgnoreCase) ?? -1) != -1;
 		}
 
 		public bool IsWeb {
--- a/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/DotNetCoreNuGetProject.cs
+++ b/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/DotNetCoreNuGetProject.cs
@@ -98,7 +98,7 @@ namespace MonoDevelop.PackageManagement
 
 		public static bool CanCreate (DotNetProject project)
 		{
-			return project.MSBuildProject.Sdk != null;
+			return project.MSBuildProject.GetProjectSdksName() != null;
 		}
 
 		public static NuGetProject Create (DotNetProject project)
--- a/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/DotNetCoreNuGetProject.cs
+++ b/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/DotNetCoreNuGetProject.cs
@@ -98,7 +98,7 @@ namespace MonoDevelop.PackageManagement
 
 		public static bool CanCreate (DotNetProject project)
 		{
-			return project.MSBuildProject.GetProjectSdksName() != null;
+			return project.MSBuildProject.GetReferencedSDKs().Length > 0;
 		}
 
 		public static NuGetProject Create (DotNetProject project)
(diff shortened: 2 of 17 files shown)


mrward:
Maybe SdkWithImplicitImports could be replaced with a bool includeImplicitImports flag on the GetReferencedSdks method? We would need probably to store the sdk information and a flag whether it is an implicit import or not. I now understand why SdkWithImplicitImports was added here after reading the code again.

mathbourgeois:
For the tests, I entirely agree. I can add them to DotNetCoreProjectTests, it seems like the logical spot to add them. I'm not 100% sure about the coverage we should expect out of those tests though. Obviously, we should see if we can load a case with the Sdk node, and one with the explicit imports. Do we want to check for weirder cases, like a mix with different sdks? I'm not 100% sure how it should react to something like a Project Sdk attribute for an sdk, followed by explicit imports of those same sdk, so I'm not 100% sure where we should go.

For the properties, I agree that it's a bit messy. I agree with your solution for the GetReferencedSdks. I'll see how much I can reduce the complexity of the interface, it's indeed not very clear what each case does (at that point, settings the Sdk for the project should probably not be a property setter anyway, since it behaves now much more like a function, potentially removing nodes in the tree.

I'll try to work on this a little more this week, not sure when exactly I'll have the time though...

mrward:
Yeah I would just look at adding tests for loading a project with an Sdk element and a Import/@Sdk element. I would not be worried about some mix of things for now. We can check how Visual Studio on Windows reacts to some strange mix of sdk imports.

The test projects can be added in a subdirectory here - https://github.com/mono/monodevelop/tree/master/main/tests/test-projects

mathbourgeois:
I've cleaned up the api based on your recommandations and added some basic load tests. for all formats. Only thing I'm not 100% sure is the name to receive a string defining the sdk(s) for the whole project, no matter where they are defined : I named it GetProjectSdksName, which I don't think is perfect, but I couldn't find any better name off the top of my head. I'll be glad to change it if you have a better idea :smile:

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
```diff
@@ -983,17 +1106,41 @@ public XmlElement GetMonoDevelopProjectExtension (string section)
...
-		public string[] GetReferencedSDKs ()
-		{
-			if (!string.IsNullOrEmpty (Sdk)) {
-				if (sdkArray == null)
-					sdkArray = Sdk.Split (new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
-				return sdkArray;
-			}
-			else
-				return Array.Empty<string> ();
+		public SdkDefinition[] GetReferencedSDKs ()
```
mrward: Unfortunately this is a public API break so we cannot change this to return SdkDefinitions. This would still need to return a string array.

I think GetReferencedSdks should return all referenced sdks and then we have a separate GetReferencedSdks method with a parameter that allows only the implicitly imported sdks to be returned, since the project model needs to import them itself.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -78,7 +78,7 @@ protected override bool OnGetSupportsFramework (TargetFramework framework)
 		/// </summary>
 		bool IsSdkProject (DotNetProject project)
 		{
-			return project.MSBuildProject.Sdk != null;
+			return project.MSBuildProject.GetProjectSdksName() != null;
 		}
```
mrward: This could just use the GetReferencedSdks method.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -406,7 +406,7 @@ static string GetDotNetCoreSdkRequiredBuildErrorMessage (bool isUnsupportedVersi
 
 		protected override void OnBeginLoad ()
 		{
-			dotNetCoreMSBuildProject.Sdk = Project.MSBuildProject.Sdk;
+			dotNetCoreMSBuildProject.Sdk = Project.MSBuildProject.GetProjectSdksName();
```
mrward: The dotNetCoreMSBuildProject.Sdk should still be the Project.MSBuildProject.Sdk value.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -416,7 +416,7 @@ protected override void OnBeginLoad ()
 
 		protected bool IsWebProject (DotNetProject project)
 		{
-			return (project.MSBuildProject.Sdk?.IndexOf ("Microsoft.NET.Sdk.Web", System.StringComparison.OrdinalIgnoreCase) ?? -1) != -1;
+			return (project.MSBuildProject.GetProjectSdksName()?.IndexOf ("Microsoft.NET.Sdk.Web", System.StringComparison.OrdinalIgnoreCase) ?? -1) != -1;
```
mrward: I think this should change to just check the GetReferencedSdks instead of needing a new GetProjectSdksName method.

main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/DotNetCoreNuGetProject.cs
```diff
@@ -98,7 +98,7 @@ public DotNetCoreNuGetProject (DotNetProject project)
 
 		public static bool CanCreate (DotNetProject project)
 		{
-			return project.MSBuildProject.Sdk != null;
+			return project.MSBuildProject.GetProjectSdksName() != null;
 		}
```
mrward: This should just check the project has referenced sdks so it could just the GetReferencedSdks method instead of introducing a new method.

main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/DotNetProjectExtensions.cs
```diff
@@ -171,7 +171,7 @@ public static IEnumerable<string> GetDotNetCoreTargetFrameworks (this Project pr
 
 		public static bool IsDotNetCoreProject (this Project project)
 		{
-			return project.MSBuildProject.Sdk != null;
+			return project.MSBuildProject.GetProjectSdksName() != null;
 		}
```
mrward: Same here.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
```diff
@@ -554,19 +561,135 @@ public string ToolsVersion
 			}
 		}
 
+		public class SdkDefinition
+		{
```
mrward: Not sure this need to be public if GetReferencedSdks will be returning a string array as before. If it needs to be public this should be in its own file.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
```diff
@@ -983,17 +1106,41 @@ public XmlElement GetMonoDevelopProjectExtension (string section)
...
+				GenerateSdkArray ();
+			}
+			return sdkArray;
 		}
 
+		/// <summary>
+		/// Returns an Sdk attribute name defining all Sdks used in the project (implicit & explicit)
+		/// </summary>
+		public string GetProjectSdksName()
+		{
```
mrward: Not sure we need this. The information is available from GetReferencedSdks.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
```diff
@@ -983,17 +1106,41 @@ public XmlElement GetMonoDevelopProjectExtension (string section)
...
+			}
+			return fullSdkName;
+		}
+
+		/// <summary>
+		/// Sets the Sdk attribute for the current project. This will clear every other previous definition
+		/// and will set an explicit sdk attribute on the project node.
+		/// </summary>
+		public void SetProjectSdk(string sdk)
+		{
```
mrward: I do not think we need this.

main/tests/Ide.Tests/MonoDevelop.Ide.Projects/DotNetCoreProjectTests.cs
```diff
@@ -60,5 +60,38 @@ public async Task MoveResourceFilesToFolder ()
...
+		public async Task LoadDotNetCoreProjectWithSdkImports ()
+		{
+			FilePath projFile = Util.GetSampleProject ("DotNetCoreSdkFormat", "DotNetCoreImportSdk", "DotNetCoreImportSdk.csproj");
+
+			using (var p = (DotNetProject)await Services.ProjectService.ReadSolutionItem (Util.GetMonitor (), projFile)) {
+				Assert.AreEqual (p.MSBuildProject.GetProjectSdksName (), "Microsoft.NET.Sdk");
+				Assert.AreEqual (p.MSBuildProject.GetReferencedSDKs ().Where (x => x.Implicit).Count (), 0);
+			}
+		}
 	}
```
mrward: 👍

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
mrward: On the MSBuildProject think we can get away with the following new public methods:

    // Returns all sdks (explicit or implicit) - this is kind of obsolete due to the new method below.
    public string[] GetReferencedSdks ()

    public SdkDefinition[] GetSdks ()

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
```diff
@@ -554,19 +561,135 @@ public string ToolsVersion
...
+			}
+
+			public bool Implicit {
+				get;
+				private set;
+			}
+
+			public string QualifiedName {
+				get {
+					return !string.IsNullOrEmpty(Version) ? $"{Name}/{Version}" : Name;
```
Therzok: Qualified name doesn't take into account that name and version can be null. Are we sure that nameAndVersion in the constructor _can_ be null?

PR #5604, opened 2018-08-07
Fixes VSTS Bug 659996: [Feedback] VS Mac 2017 - Pick members to

mkrueger:
override for a class window doesn't render options correctly

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/659996

Can't reproduce what's seen in the screenshot howover this test case
fails for me which is fixed by that patch.


Therzok:

Therzok:

PR #5610, opened 2018-08-07
[Telemetry] Fix editor keypress telemetry to not report overflowed va…

Therzok:
…lues

TimeSinceMachineStart was a long value, while gdk event time is
an uint32 value which would overflow and reset from 0.

To prevent this issue, we're now requesting a long value by default
which we can pass from native APIs.

Current gdk event is translated from an NSEvent, so we can assume
GdkEvent.Current is translated from the actual current NSEvent.

Fixes VSTS #659150 - xs/editor/textcommandtiming/typechar has very large values in GeoCountryRegionIso == 'US'

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextEditorKeyPressTimings.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextEditorKeyPressTimings.cs
@@ -64,7 +64,7 @@ namespace Mono.TextEditor
 			8, 16, 32, 64, 128, 256, 512, 1024
 		);
 
-		public void StartTimer (Gdk.EventKey evt)
+		public void StartTimer (long eventTime)
 		{
 			if (activeCountIndex == numberOfCountSpaces) {
 				// just drop these events now
@@ -75,7 +75,7 @@ namespace Mono.TextEditor
 			droppedEvents = 0;
 
 			// evt.Time is the time in ms since system startup (on macOS at least)
-			activeCounts[activeCountIndex++] = evt.Time;
+			activeCounts[activeCountIndex++] = eventTime;
 		}
 
 		int CalculateBucket (TimeSpan duration)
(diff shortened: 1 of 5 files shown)


Therzok:
@sgmunn

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextEditorKeyPressTimings.cs
```diff
@@ -127,6 +127,8 @@ public void EndTimer (bool complete = false)
 				return;
 			}
 
+			// Gdk key events are wrapped to uint32, so if we use longs here, we will get keypresses that
+			// seemingly last for days.
 			var sinceStartup = (long)telemetry.TimeSinceMachineStart.TotalMilliseconds;
```
mrward: Might be a dumb question but why do we use telemetry.TimeSinceMachineStart.TotalMilliseconds here and then a different way to get this value in the OnKeyPressEvent method?

sgmunn:
https://cl.ly/2h2n1R0V3G3F

opened a project, typed some text, closed the project. Still get large numbers that don't make sense to me.

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextEditorKeyPressTimings.cs
Therzok: The current event does not change, so the timestamp will be the same, I guess? I could try to get the current event's timestamp here too.

mrward: It uses `TimeSpan.FromSeconds (NSProcessInfo.ProcessInfo.SystemUptime)` which should update. But anyway - not really a problem with this change :)

Therzok: Ah, you meant using TimeSpan.FromSeconds there? Sure, could do.

sgmunn:
screen shot 2018-08-08 at 11 42 58 am

@Therzok 👍

Therzok:

Therzok:

PR #5620, opened 2018-08-08
Revert "[Ide] Don't show welcome page when opening a specific item"

Therzok:
This reverts commit 74477489ef87a0d7fd3dca1605e6b304a576ef71.


slluis:
Why?

Therzok:
Different number of issues. Flakiness in behaviour (i.e. with this patch, opening slns from unity editor doesn't work anymore).

There's multiple scenarios we need to cover:
* opening via terminal - which the patch we're reverting fixed
* opening via finder/open - aka launchservices - which is harder to fix, since the open document event comes after we show the welcome page
* switching solutions

And for better UX, we need to have a way to safely show the welcome page after we start loading in the background.

It needs a bit of rework alltogether.

PR #5622, opened 2018-08-08
[Ide] Log addin startup handlers

Therzok:
These force load the addin, so log these to track in the future

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/643441


mrward:
I think `args.ExtensionNode.Id` returns a random number/id for the startup handlers since they have no id - they are all type extensions. I added logging recently just to check how long each handler was taking and had to do something like `var typeExt = args.ExtensionNode as TypeExtensionNode;` and then log the `typeExt.TypeName` to get a nice name displayed.

Also not sure do we need to log the stack trace? Will it not it be the same on startup for each handler?

Therzok:
Stacktraces are not needed, no, it was just muscle reflex of logging them.

Yeah, TypeExtensionNode casts make sense. We could probably also log the command codon ID

Therzok:
Up for review again! cc @mrward @slluis

PR #5629, opened 2018-08-08
Make MdTextViewLineCollection recreate the non-modified lines and del…

Therzok:
…ete the modified ones. This appears to fix several of the exceptions seen during editing/scrolling.

Saw this branch had no PR, so I opened one


mrward:
Do we want this? There seems to have been a different change made on master in this part of the code.

KirillOsenkov:
@Therzok I was about to prepare a PR with this change, but didn't get to it yet. /cc @ToddGrun who made the actual change.

KirillOsenkov:
Ah. This is not the change we want. The bug you're experiencing is https://devdiv.visualstudio.com/DevDiv/_workitems/edit/657666 and the fix is in dev/toddgrun/release-7.6/GetTextViewLineContainingBufferPosition. 

I'll try to get to it today or tomorrow.

PR #5630, opened 2018-08-09
Fix #4969

### LINKED ISSUE #4969: XML intellisense is very annoying
*Repro*

1. Create new xslt file
2. Try to type `<xsl:stylesheet`

*Actual result*

the `:` commits the first completion result (`<xsl:if`) then is entered as a char and you end up with `<xsl:if:stylesheet`

*Expected result*

You get what you typed

This is a regression. i'm not sure from what version, but I'm certain that years ago I made sure that characters would not commit intellisense if they matched an entry in the list.

Maybe we should fix this like we did for XAML: not insert the colons into the list, and instead add intellisense for the portion after the colon.

> VS bug [#624635](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/624635)

DavidKarlas:
This fixes #4969 and also adds unit test


parasAmbhore:
I can confirm that this issue is not reproducible when tested with the following build:
https://dl.internalx.com/artifacts/devdiv-1932754/VisualStudioForMac-Preview-7.8.0.48.dmg

Verification Screen-Cast:
https://www.screencast.com/t/GuqBjhmJyN

**QA Approved.**

slluis:
@DavidKarlas any idea why tests may be failing? It seems unrelated, but I ran the tests twice and they are still failing.

DavidKarlas:
Because changes in `TestViewContent.cs`, some C# unit tests start picking up different formatting rules and start failing, debugged once and couldn't figure out what is happening... Will try again to figure out how to fix this.

slluis:

slluis:

Issue #5590, opened 2018-08-03
Alt-shift-left/right no longer extends selection


mhutch:
Version 7.7 build 768

Rather than extending the selection by a word, alt-shift-left/right now _moves_ the selection by a word 

![2018-08-03 15_57_22](https://user-images.githubusercontent.com/183285/43663021-f0d4aaec-9735-11e8-9113-3143f8ab0241.gif)

> VS bug [#659183](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/659183)

mhutch:
In comments it's even worse, it moves the selection by a single character within words.

![2018-08-03 16_04_21](https://user-images.githubusercontent.com/183285/43663278-e6126ed6-9736-11e8-8333-bdf623fb6174.gif)

mkrueger:
Interesting - works on linux but not on os x

mhutch:
Can we add some tests for this? It's pretty fundamental.

mhutch:
I gave 5bb533efb6a9e7daeb5ad341d3c2c8b1bd283b56 a spin and it's even worse.

Alt-shift-left/right extends the selection now, but it doesn't *start* a selection if nothing is selected.

And word navigation in comments skips over *entire sentences*.

mkrueger:
hm, we had tests for it but I don't know how the vs.net commands should behave. But what's weird is that it doesn't work on os x but on linux it works - I can't find any difference there.

Maybe I should switch back to our custom implementation.

xamarin-release-manager:
Fixed in version 7.8.0.82 (master)

Pull Request #5650 merged by: Mikayla Hutchinson
Author: mono
Commit: 23fb81854741266de8af1e87863bfbe7537e3bcc (mono/monodevelop)

xamarin-release-manager:
Fixed in version 7.8.0.126 (release-7.7)

Pull Request #5650 merged by: Mikayla Hutchinson
Author: mono
Commit: 23fb81854741266de8af1e87863bfbe7537e3bcc (mono/monodevelop)

xamarin-release-manager:
Fixed in version 7.7.0.931 (release-7.7)

Pull Request #5650 merged by: Mikayla Hutchinson
Author: mono
Commit: 23fb81854741266de8af1e87863bfbe7537e3bcc (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

[...]

PR #5650, opened 2018-08-10
Fixes issue #5590 Alt-shift-left/right no longer extends selection

mkrueger:


Therzok:
Can we add some tests?

mhutch:
@KirillOsenkov any ideas? Looks like Roslyn's AbstractTextStructureNavigatorProvider.TextStructureNavigator.cs is getting used here

KirillOsenkov:
Adding @grabalon who investigated this recently.

KirillOsenkov:
@grabalon just stopped by and we discussed this. Basically if you're at the start of a comment, Roslyn reports the whole comment as a "word". Hence the behavior you're seeing. 

He has an internal PR (https://devdiv.visualstudio.com/DevDiv/Editor/_git/VS-Platform/pullrequest/135348?_a=overview) where he works around it by looking at the next character, and then asking for the word extent of that.

If you look into his PR, look in SelectionTransformer.cs change.

mhutch:
@KirillOsenkov I can see from the code that's the intended behavior. However, right now it skips from _anywhere within the comment_ to the end, so something is broken.

mkrueger:
The behavior on linux is different it works there - so it must be something in the os x implementation causing that. However I can't find any difference :/

PR #5671, opened 2018-08-10
Add MonoDevelop.MonoDroid InternalsVisibleTo in MonoDevelop.Ide

bratsche:
This will be needed in order to use InfoBars by https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems/edit/646093


bratsche:
@slluis I'm not sure what you mean. I want to use `Workbench.ShowInfoBar()` but it's marked as `internal`. I didn't see a public way of doing this.

bratsche:
@slluis The API I want to use is marked `internal`. I asked @Therzok about it and he suggested I add `InternalsVisibleTo`. Should we make it public instead?

slluis:
If we need to use it in an add-in it should be public, unless there is a very good reason not to.

Therzok:
I didn't want to open up the API when I first wrote it.

I guess the API can look like this, if we want to make it public:
```
	public interface IInfoBarHost
	{
		Control AddInfoBar (InfoBarOptions options);
	}

 	public class InfoBarOptions
    	{
		public InfoBarItem[] Items { get; set; }

		public InfoBarOptions(string description)
                 {
                 }
	}
```

And we add `IInfoBarHost` to `Workbench` which would route into `DefaultWorkbench`’s `IInfoBarHost` implementation.

bratsche:
Closing in favor of https://github.com/mono/monodevelop/pull/5877

Therzok:
See https://github.com/mono/monodevelop/pull/5878

PR #5681, opened 2018-08-13
Attempt to Fix VSTS 662566.

KirillOsenkov:
We suspect that while the completion window is painted, it is being closed from a background thread, thus causing an assert in Cairo that brings down the process.

If we schedule the Close event to be on the UI thread, it is no longer possible for it to interrupt painting.

Fixes VSTS 662457.


KirillOsenkov:
@sayedihashimi this is an attempt to fix the crash while editing .cshtml.

KirillOsenkov:
Abandoning because this approach didn't work.

KirillOsenkov:
Obsoleted by https://github.com/mono/monodevelop/pull/5690.

PR #5682, opened 2018-08-13
Temp workaround for VSTS 662639.

KirillOsenkov:
The first call may fail with the ArgumentException coming from Mono due to ConditionalWeakTable
reentrancy, however by that time the table will have already been populated.
Subsequent call should not hit this problem and should succeed.

TODO: remove when https://github.com/dotnet/roslyn/issues/28256 is fixed.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/662639.

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.Ide.Completion.Presentation/ExtensionMethods.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.Ide.Completion.Presentation/ExtensionMethods.cs
@@ -1,0 +1,31 @@
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Text;
+using Microsoft.VisualStudio.Text;
+
+namespace MonoDevelop.Ide.Completion.Presentation
+{
+	public static class ExtensionMethods
+	{
+		/// <summary>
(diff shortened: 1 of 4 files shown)


main/src/addins/CSharpBinding/MonoDevelop.Ide.Completion.Presentation/ExtensionMethods.cs
```diff
@@ -0,0 +1,31 @@
...
+		/// TODO: remove when https://github.com/dotnet/roslyn/issues/28256 is fixed.
+		/// </summary>
+		public static Document GetOpenDocumentInCurrentContextWithChangesSafe (this ITextSnapshot textSnapshot)
+		{
+			Document document = null;
+			try {
+				document = textSnapshot.GetOpenDocumentInCurrentContextWithChanges ();
+			} catch {
+				try {
+					document = textSnapshot.GetOpenDocumentInCurrentContextWithChanges ();
```
slluis: This second call should not be in a try/catch. If it fails it will return null and will NRE later on.

KirillOsenkov:
I've tested this pretty extensively and this fixes the Roslyn issue. 

Bonus: this seems to also have fixed https://devdiv.visualstudio.com/DevDiv/_workitems/edit/645709 (completion tooltips staying open).

PR #5690, opened 2018-08-14
Fix VSTS 662457.

KirillOsenkov:
Avoid a crash caused by repositioning the completion window while it's being repainted.


KirillOsenkov:
Sayed confirmed this fixes the issue.

slluis:
@monojenkins backport master

PR #5363, opened 2018-07-10
[NuGet] Rewrite options panel in XWT

mhutch:
WIP

XWT rewrite of the panel:

image

PR #5692, opened 2018-08-15
[C#] Fix 35 obsoletion warnings

mhutch:


mhutch:
(cherry-picked from https://github.com/mono/monodevelop/pull/5363)

PR #5700, opened 2018-08-16
[Core] Invalidate framework cache when Mono upgraded

mhutch:
Fixes assembly resolution in certain codepaths that don't yet depend on MSBuild.


slluis (approved):

mrward:

PR #5708, opened 2018-08-17
[Core] Fix incorrect build location with imported config based property

mrward:
With a project that has an import that defines a property based on
the configuration the build location of the assembly would be
incorrect.

Project file:

    <Project ...>
      <Import Project="Imported.props" />
      <PropertyGroup>
        <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
        <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
        <OutputPath>$(OutputBase)\Common</OutputPath>
      </PropertyGroup>

Imported.props:

     <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
      <PropertyGroup>
        <TargetDir>target</TargetDir>
        <OutputBase>$([System.IO.Path]::Combine($(TargetDir), $(Configuration)))</OutputBase>
      </PropertyGroup>
    </Project>

When building the above project, with the debug configuration
selected, the assembly would be built into the directory
'target/Common' instead of 'target/Debug/Common'. To fix this the
configuration is now always set for the project in the remote project
builder.

Fixes VSTS #540728 - OutputBase No Longer Includes Build Configuration

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/ProjectDescriptor.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/ProjectDescriptor.cs
@@ -257,6 +257,11 @@ namespace MonoDevelop.Ide.Templates
 			string dir = ReplaceParameters (directory, substitution, projectCreateInformation);
 			projectCreateInformation.ProjectBasePath = Path.Combine (projectCreateInformation.SolutionPath, dir);
 
+			// Ensure that any relative path (e.g. "./Droid") used by 'directory' for the project template is removed since this
+			// can cause a problem with the remote msbuild host trying to load the project multiple times into the build engine
+			// due to a mismatch in the filename being used.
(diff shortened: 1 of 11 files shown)


mrward:
Also seems to be a similar problem with Platform. Will open a separate bug for that.

slluis:
Can you please check if this change has any impact in performance? IIRC the comparison was added to avoid modifying the project at every build, so a repeated build would net require an re-evalualuation of the project.

mohakbarokar:
This bug is tested on Build : Visual Studio Enterprise 2017 for Mac (Preview) Version 7.8 Preview (7.8 build 102)

Build Gist: https://gist.github.com/mohakbarokar/7715b086861b34c6acdc4a877ace0e10

Build Taken from: https://github.com/mono/monodevelop/pull/5708

Screencast Link: https://www.screencast.com/t/ONVBkd1xl

& the bug is found to be fixed. 

VSfM is building  assembly into the target/Debug/Common folder. We can merge this PR

mrward:
All the tests are passing now.

Issue #5712, opened 2018-08-17
Unable to load SDK-style projects that lack `Sdk` attribute


sandyarmstrong:
The following project loads fine in VS 15.7.4, but fails to load in VSmac 7.6 build 2086:

```xml
<!-- Notice no Sdk="Microsoft.NET.Sdk". This prevents the project from loading
     in VSmac, though it loads fine in VSwin. Avoiding the SDK attribute and
     manually doing SDK props/targets importing is often suggested as a
     workaround for bugs like https://github.com/NuGet/Home/issues/5894. -->
<Project>
    <!-- This is implicit with Project@Sdk="Microsoft.NET.Sdk" -->
    <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />

    <PropertyGroup>
        <TargetFramework>net472</TargetFramework>
    </PropertyGroup>

    <!-- This is implicit with Project@Sdk="Microsoft.NET.Sdk" -->
    <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
</Project>
```

The following error appears:

![image](https://user-images.githubusercontent.com/155076/44281554-0cee0480-a20d-11e8-94e4-06bff8b23414.png)

In the log:

```
ERROR [2018-08-17 10:41:57Z]: Error while trying to load the project /Users/sandy/xam-git/simple/simple.csproj
MonoDevelop.Projects.MSBuild.UnknownSolutionItemTypeException: Unknown solution item type
  at MonoDevelop.Projects.ProjectService+<>c__DisplayClass25_0.<ReadSolutionItem>b__0 () [0x000ca] in /Users/vsts/agent/2.136.1/work/1/s/monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectService.cs:125 
  at MonoDevelop.Core.Runtime+<>c__42`1[T].<RunInMainThread>b__42_0 (System.Object state) [0x00046] in /Users/vsts/agent/2.136.1/work/1/s/monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs:393 
```

mrward:
This is a known limitation - explicit imports are not as yet supported - https://bugzilla.xamarin.com/show_bug.cgi?id=58564 (VSTS 522315)

There is a [pull request](https://github.com/mono/monodevelop/pull/5597) with a fix that I need to review/test.

mrward:
This should now be fixed.

[...]

PR #5713, opened 2018-08-19
[DotNetCore] Support explicit .NET Core Sdk imports

mrward:
Local monodevelop branch with some tweaks of the pull request #5597 to verify the tests.

Adds support for:

 - `<Import Sdk="Microsoft.NET.Sdk" Project="Sdk.props" />`
- `<Import Sdk="Microsoft.NET.Sdk" Project="Sdk.targets" />`
 - `<Sdk Name="Microsoft.NET.Sdk" Version="1.0.0" />`
 
Fixes VSTS #522315 - Support explicit .NET Core Sdk imports
Fixes github #5712 - Unable to load SDK-style projects that lack `Sdk` attribute

Looking at the [spec](https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-use-project-sdk) and the [MSBuild unit tests](https://github.com/Microsoft/msbuild/blob/7b3a457540ab87cf02f6d41ce6095e6ecbbb6ea7/src/Build.UnitTests/Evaluation/ProjectSdkImplicitImport_Tests.cs) it looks like the logic we currently have for Import Sdk elements needs changing. Import elements can define a Version. Also Sdk and Import elements both support a MinimumVersion attribute. Currently the Import/Sdk attribute is parsed as though it has could contain an Sdk and a Version part which looks like it is incorrect. The Project/Sdk attribute does support defining both the sdk, version and minimum version. We could fix this in a separate pull request.


slluis (approved):

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -416,7 +416,7 @@ protected override void OnBeginLoad ()
 
 		protected bool IsWebProject (DotNetProject project)
 		{
-			return (project.MSBuildProject.Sdk?.IndexOf ("Microsoft.NET.Sdk.Web", System.StringComparison.OrdinalIgnoreCase) ?? -1) != -1;
+			return (project.MSBuildProject.GetReferencedSDKs ()?.FirstOrDefault (x => x?.IndexOf ("Microsoft.NET.Sdk.Web", System.StringComparison.OrdinalIgnoreCase) != -1) != null);
```
Therzok: can `x` be null here?

Therzok (approved):

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
mrward: Not with the current implementation. GetReferencesSdks should always at least return an empty array and should never return any null items. So I think we can get rid of the null ? operator for the entire line of code.

PR #5728, opened 2018-08-21
[Ide] Fix project not being run when it has warnings

mrward:
With "Run project if build completed with warnings" set to true in
Preferences a project would not be run if the build had warnings.
The preferences setting had the opposite effect since it was being
treated as though it indicated the run should be cancelled when it
was true.

Fixes VSTS #667156 - [VSM] Project is not built and started to run
after Toolbar -> Run hit


slluis:

PR #5730, opened 2018-08-21
[Icons] New zoom icons

vancura:
Here's new set of zoom icons:

zoom

As you can see they're much simpler now (I removed the lens symbol). For that reason, we still need to test if the icons are still readable enough sporting this minimalism. Hence, please don't merge yet.


vancura:
Closing, this didn't work.

PR #5735, opened 2018-08-22
Fix VSTS 645709: Completion description tooltip sometimes stays open

KirillOsenkov:
Our theory is that when the async method returned it was checking the wrong cancellation token - there may be multiple async operations in flight, but only one cancellation token. So if we cancel one cancellation token, but another operation comes back, it would create a new Window.

I've refactored the code to ensure that only one window can be open at a time.


KirillOsenkov:
@sayedihashimi we have a build

KirillOsenkov:

sayedihashimi:
I tried the build and I was not able to reproduce the tooltip issue.

PR #5736, opened 2018-08-22
Fixes VSTS Bug 650534: [Feedback] High CPU usage viewing "Search in

mkrueger:
Solution" results, UI delays and unresponsiveness

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/650534

The old implementation created a huge number of text editor objects
which is not efficient anymore. Instead now one text editor object is
re-used. However the search result data is cached querying the text
editor is expensive.


mkrueger:

PR #5744, opened 2018-08-22
[MSBuild] Enable usage of analyzers from `<Analyzer>` itemgroup

Therzok:
This patch aims to implement one part of analyzers from msbuild, and that is including items from the `<Analyzer>` itemgroup into the build.

We just pass those values to Roslyn, and it handles the rest!

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -81,6 +81,8 @@ namespace MonoDevelop.Projects
 
 		IEnumerable<string> loadedAvailableItemNames = ImmutableList<string>.Empty;
 
+		CachingCoreCompileEvaluator<ProjectFile> compileEvaluator;
+
 		protected Project ()
 		{
 			runConfigurations = new RunConfigurationCollection (this);
@@ -89,6 +91,8 @@ namespace MonoDevelop.Projects
 			files = new ProjectFileCollection ();
 			Items.Bind (files);
 			DependencyResolutionEnabled = true;
+
+			compileEvaluator = new CachingCoreCompileEvaluator<ProjectFile> (this, "Compile", CreateProjectFile);
         }
 
 		public ProjectItemCollection Items {
@@ -499,7 +503,7 @@ namespace MonoDevelop.Projects
 			results.AddRange (evaluatedItems);
 
 			// add in any compile items that we discover from running the CoreCompile dependencies
-			var evaluatedCompileItems = await GetCompileItemsFromCoreCompileDependenciesAsync (monitor, configuration);
+			var evaluatedCompileItems = await compileEvaluator.GetItemsFromCoreCompileDependenciesAsync (monitor, configuration);
 			var addedItems = evaluatedCompileItems.Where (i => results.All (pi => pi.FilePath != i.FilePath)).ToList ();
 			results.AddRange (addedItems);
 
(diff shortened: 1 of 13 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -81,6 +81,9 @@ public class Project : SolutionItem
 
 		IEnumerable<string> loadedAvailableItemNames = ImmutableList<string>.Empty;
 
+		CachingCoreCompileEvaluator<ProjectFile> compileEvaluator;
+		CachingCoreCompileEvaluator<FilePath> analyzerEvaluator;
```
Therzok: @mrward I wonder if we should have only one CachingCoreCompileEvaluator, and query both Compile and Analyzer items in there and just return cached results.

Therzok: The reason is that we might invoke the actual targets twice, delaying the typesystem load.

Therzok:
@slluis @mrward ping for review

This enables analyzers for users which include `<Analyzer>` MSBuild items in the IDE. That msbuild itemgroup enables analyzers for builds and IDE usage.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -461,6 +465,32 @@ public Task<TargetEvaluationResult> PerformGeneratorAsync (ProgressMonitor monit
 			return this.RunTarget (monitor, generatorTarget, configuration);
 		}
 
+		/// <summary>
+		/// Gets the analyzer files that are included in the project, including any that are added by `CoreCompileDependsOn`
+		/// </summary>
+		public Task<FilePath []> GetAnalyzerFilesAsync (ConfigurationSelector configuration)
+		{
+			if (sourceProject == null)
+				return Task.FromResult (Array.Empty<FilePath> ());
```
mkrueger: Maybe a little improvement: TaskUtil.Default<FilePath[]> ();

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -487,6 +517,15 @@ public Task<ProjectFile[]> GetSourceFilesAsync (ConfigurationSelector configurat
 			return ProjectExtension.OnGetSourceFiles (monitor, configuration);
 		}
 
+		/// <summary>
+		/// Gets the analyzer files that are included in the project, including any that are added by `CoreCompileDependsOn`
+		/// </summary>
+		protected virtual async Task<FilePath[]> OnGetAnalyzerFiles (ProgressMonitor monitor, ConfigurationSelector configuration)
+		{
+			var coreCompileResult = await compileEvaluator.GetItemsFromCoreCompileDependenciesAsync (this, monitor, configuration);
+			return coreCompileResult.AnalyzerFiles.ToArray ();
```
mkrueger: Does the return type really require to be an array ? Othwise we can just save the ToArray call here.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -561,100 +601,137 @@ void OnMSBuildProjectImportChanged (object sender, EventArgs args)
...
+				if (startTask) {
+					var coreCompileDependsOn = project.sourceProject.EvaluatedProperties.GetValue<string> ("CoreCompileDependsOn");
+
+					if (string.IsNullOrEmpty (coreCompileDependsOn)) {
+						currentTask.SetResult (CoreCompileEvaluationResult.Empty);
+						return currentTask.Task.Result;
+					}
+
+					var result = CoreCompileEvaluationResult.Empty;
+					var dependsList = string.Join (";", coreCompileDependsOn.Split (new [] { ";" }, StringSplitOptions.RemoveEmptyEntries).Select (s => s.Trim ()).Where (s => s.Length > 0));
```
mkrueger: I'm not sure how performance critical this code is but the operation above could be way faster using a string builders and custom methods for checking.
For example the Select /Where is basically just a very costly string.IsNullOrWhiteSpace check.

In general LINQ calls very expensive as well as string manipulations.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -561,100 +601,137 @@ void OnMSBuildProjectImportChanged (object sender, EventArgs args)
...
+					try {
+						// evaluate the Compile targets
+						var ctx = new TargetEvaluationContext ();
+						ctx.ItemsToEvaluate.Add ("Compile");
+						ctx.ItemsToEvaluate.Add ("Analyzer");
+						ctx.LoadReferencedProjects = false;
+						ctx.BuilderQueue = BuilderQueue.ShortOperations;
+						ctx.LogVerbosity = MSBuildVerbosity.Quiet;
```
mkrueger: Can the context be reused ?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -561,100 +601,137 @@ void OnMSBuildProjectImportChanged (object sender, EventArgs args)
...
-				} catch (Exception ex) {
-					LoggingService.LogInternalError (string.Format ("Error running target {0}", dependsList), ex);
+					currentTask.SetResult (result);
 				}
-				currentTask.SetResult (result ?? new ProjectFile [0]);
-			}
 
-			return await currentTask.Task;
-		}
+				return await currentTask.Task;
```
mkrueger: await needed ?

mkrueger (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
Therzok: We're in Core, can't access IDE from here.

mkrueger:
In general I've seen you're prefixing member fields with underscore - I know in some MS projects it's the case however it's not the monodevelop programming style. We should maybe discuss that and change our guidelines there.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
Therzok: If the API were ImmutableArray, sure, otherwise people could just modify the values. (I thought about it too)

Therzok: I did array for consistency, could probably make it ImmutableArray.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
Therzok: Will modify it, it can be considered performance critical, but after this code is run, the IDE will just wait for the msbuild builder to return results.

Still, good to optimize it, I didn't look at it closely when I moved the code around.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
Therzok: Never seems to be reused across the IDE, but valid point. It's 'immutable' data mostly.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
Therzok: It's an async method, yeah, it is.

Therzok:
Those underscore members are copy pasted from roslyn. I kept the naming conventions so we don't have conflicts/massive changes when we reimport from roslyn.

http://source.roslyn.io/#Microsoft.VisualStudio.LanguageServices/Implementation/TaskList/HostDiagnosticUpdateSource.cs,17

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
mkrueger: y same on my side when moving code around but awful constructs like that is something we should take out in the project system. String manipulations are very expensive

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
mkrueger: That leaves the question why TaskUtil isn't in core ... but that's not part of this PR :).

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
Therzok: ReadOnlySpan is in now, going to see if I can use any tricks there.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
Therzok: I'll go with `ImmutableArray`. Interface returns are okay-ish, but iterating them via foreach is slower.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
Therzok: Filed https://github.com/mono/monodevelop/issues/5931

Therzok:
All comments have been handled. Ready to merge.

mrward (approved):

Therzok:

PR #5748, opened 2018-08-23
Fixes VSTS Bug 669434: AddImportCompletionData Throws Exception

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/669434

With the provided information it's not possible to see what's going
on.
But I reviewed that method and hardened it against exceptions and
prevented some possible null reference cases.
At least the import completion no longer prevents completion from
showing up and will give better error information in case of failure.


slluis:

PR #5750, opened 2018-08-23
[Ide] Fix text editor errors after creating new .NET Core project

mrward:
- Fixes type system not reloading solution when project modified
 - Fixes text editor errors on adding .NET Core project to solution
 - Fix text editor errors after creating new .NET Core project

Occasionally on creating a new .NET Core project the text editor
would show errors such as "Predefined type 'System.Object' is not
defined or imported." as though the project had no references. This
can also happen on adding a new .NET Core project to an existing
solution. Both of these seem to be caused by the type system being
out of sync since it did not always process project modify events
when they occurred.

Fixes VSTS #635423 - Error after creating new .NET Core project


slluis:
@DavidKarlas @mkrueger can you review please?

PR #5758, opened 2018-08-24
Fixes vsts Bug 669030: [Feedback] VSMac 7.6 UI Blitz after pasting c#

mkrueger:
code from docs sample

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/669030

Without beeing able to reprodue the issue I can only guard against the
span out of range case. Getting no refactorings at a position
shouldn't produce visible crashes.


Therzok:
I think @DavidKarlas 's analysis is good, and we should do that.

Issue #5661, opened 2018-08-10
RoslynCompletionData: Can't find item type 'ExtensionMethod,Public'


KirillOsenkov:
Saw this in the logs, decided to file an issue just in case:

```
WARNING [2018-08-03 12:33:43Z]: RoslynCompletionData: Can't find item type 'ExtensionMethod,Public'
```

> VS bug [#662527](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/662527)

Therzok:
Seems like we need to sync with http://source.roslyn.io/#Microsoft.CodeAnalysis.Features/Completion/GlyphTags.cs,10

mkrueger:
Must eb a very old verison - ExtensionMethod should work. Can you reproduce it somehow ?

KirillOsenkov:
This roslynCompletionTypeTable has the ExtensionMethod:
http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/RoslynCompletionData.cs,3022781113baecd8

but this one doesn't:
http://source.monodevelop.com/#MonoDevelop.CSharpBinding/MonoDevelop.Ide.Completion.Presentation/RoslynCompletionPresenterSession.View.cs,6914a15f006706d6

Should we unify them somehow?

KirillOsenkov:
Also check for `Intrinsic`:
`RoslynCompletionData: Can't find item type 'Intrinsic'`

mkrueger:
Y they should be unified. The code shouldn't have been duplicated to begin with. 
However I remember that I removed the code duplication at some point but I don't know in which branch  - seems to have not been merged or overwritten :(

xamarin-release-manager:
Fixed in version 7.8.0.247 (master)

Pull Request #5760 merged by: Lluis Sanchez
Author: mono
Commit: 2a4d92262d07519cb7706d6f31ddece3f53f4412 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 7.7.0.1140 (release-7.7)

Pull Request #5760 merged by: Lluis Sanchez
Author: mono
Commit: 2a4d92262d07519cb7706d6f31ddece3f53f4412 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #5760 merged by: Lluis Sanchez
Author: mono
[...]

PR #5760, opened 2018-08-24
Fixes issue #5661 RoslynCompletionData: Can't find item type

mkrueger:
'ExtensionMethod,Public'


main/src/addins/CSharpBinding/MonoDevelop.Ide.Completion.Presentation/RoslynCompletionPresenterSession.View.cs
```diff
@@ -371,7 +371,7 @@ protected override bool OnExposeEvent (Gdk.EventExpose args)
 					//	layout.Attributes = attrList;
 					//}
 
-					Xwt.Drawing.Image icon = ImageService.GetIcon (GetIcon (item));
+					Xwt.Drawing.Image icon = ImageService.GetIcon (RoslynCompletionData.GetIcon (item, "text/csharp"));
```
Therzok: Can we add the comment back we had before?

```
//TODO: Todd, can you sprinkle some magic on this to do
//textView.GetTextBufferFromSpan(triggerSpan).GetMimeType() instead fixed text/csharp?
```

PR #5762, opened 2018-08-24
[Core] Fix properties not being saved.

mrward:
If the property value was a Properties instance then the values
would not be saved if the original property did not exist in the
saved properties file. This was because the Properties class was
checking the property values against its list of default values and
not serializing them if there is a match. This check did not work
for values which are Properties since they are the default value
if the property did not originally exist. This resulted in the
value not being saved.

Changed the code to special case Properties being used as a value so
a check is made to see if it contains only default values instead
of not being saved.

Fixes VSTS #669533 - Editor font settings cannot be saved.


mrward:

Therzok (approved):

slluis (approved):

PR #5769, opened 2018-08-24
[Ide] Fix NRE in CommentTasksView

sevoku:
When notifying about comment tasks changes, some cached
tasks might have become invalid. Let the CommentTasksProvider
skip tasks with invalid document or project,
resulting in a null CommentTaskChange.

Fixes VSTS #670781


sevoku:

PR #5772, opened 2018-08-25
Add IsBuildingForExecution MSBuild property

mhutch:
This can be used by builds to distinguish between a standalone build and a build that is being performed as a prerequisite for execution.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Solution.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Solution.cs
@@ -966,7 +966,7 @@ namespace MonoDevelop.Projects
 			try {
 				operationStarted = await BeginBuildOperation (monitor, configuration, operationContext);
 
-				return result = await RunParallelBuildOperation (monitor, configuration, slnConf, sortedItems, (ProgressMonitor m, SolutionItem item) => {
+				return result = await RunParallelBuildOperation (monitor, configuration, slnConf, sortedItems, operationContext, (ProgressMonitor m, SolutionItem item) => {
 					return item.Clean (m, configuration, operationContext);
 				}, false, false);
 			} finally {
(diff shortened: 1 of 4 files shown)


Therzok (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Solution.cs
```diff
@@ -1047,7 +1048,7 @@ public async Task<BuildResult> BuildItems (ProgressMonitor monitor, Configuratio
 
 				var myStatus = buildStatus[item];
 
-				if (skipUnmodified && item is Project p && !p.FastCheckNeedsBuild (configuration)) {
+				if (skipUnmodified && item is Project p && !p.FastCheckNeedsBuild (configuration, (operationContext as TargetEvaluationContext) ?? new TargetEvaluationContext ())) {
```
slluis: Better:
`(operationContext as TargetEvaluationContext) ?? new TargetEvaluationContext (operationContext)`

slluis:
A more simple solution would be to just add a IsBuildingForExecution boolean property to OperationContext.

Also, what are the changes in LogView.cs for? they look unrelated.

mhutch:
@slluis The LogViewPad was crashing while running the unit tests, because the test creates two build outputs in rapid succession, triggering the race.

I put the property in MSBuild so it would be available to MSBuild targets - the use case is a project type (Docker) which is able to optimize things by launching the project as part of the build, so this information would need to be available to the targets. That said, I guess we could special case it and have a property that later maps down to MSBuild.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Solution.cs
mhutch: Oh, good catch, didn't notice the ctor would accept null

slluis (approved):

mhutch:

Issue #5360, opened 2018-07-10
Not obvious which "fix all" is which


mhutch:
In the following screenshot, there's nothing linking each of the "fix all" fixes to the actual fix description, they just have codes. I *guess* it's same order but it's not clear.

![image](https://user-images.githubusercontent.com/183285/42534268-4bd8bc56-845a-11e8-9669-c2bbe9dd6bb9.png)

> VS bug [#645791](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/645791)

mhutch:
VS handles this by displaying the codes in the fix preview

![image](https://user-images.githubusercontent.com/183285/42534680-59f58322-845b-11e8-87f7-4d8dcf9618b1.png)

mkrueger:
We can't do a preview in the context menu unfortunately :(.
However would be possible in a custom popup dialog like in VS.

xamarin-release-manager:
Fixed in version 7.7.0.1228 (release-7.7)

Pull Request #5776 merged by: Mike Krüger
Author: mono
Commit: cf3cd8a1639c614bb122feebeda0b5cd560b3c3e (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #5776 merged by: Mike Kr??ger
Author: mono
Commit: cf3cd8a1639c614bb122feebeda0b5cd560b3c3e (xamarin/monodevelop)

PR #5776, opened 2018-08-27
Fixes issue #5360 Not obvious which "fix all" is which The fix all

mkrueger:
text is generated by roslyn but it's possible to add a header item for
the fix all items. OTOH the items would become to large when using the
full title in one line.
Best solution would probably be to just move the code fixes & issues
out of the context menu to a custom popup dialog. But that's a user
impact that's beyond the scope of the bug report.


slluis:
@mkrueger can you attach a screenshot of how the menu looks like after the change?

mkrueger:
There is a bug on os x (not on linux) - the item that has enabled == false is sensitive on the os x context menu implementation. I investigated that this morning but didn't find the cause yet.

I've 1-2 other more high priority items on the todo list.

mkrueger:
image

PR #5781, opened 2018-08-28
Fix 668995: [Feedback] Xaml Syntax Highlighting Has Reverted in 7.6 (2190)

DavidKarlas:
In 7.6 some logic was changed in SyntaxHighligthingService.cs to do `return name != null && scopeName != null;`(before it just returned `true`) since `xaml.json` has `scopeName` on 2nd line it failed... Hence now we read for new format also 1st and 2nd line.
Also some logic around regex changed so I made few changes to xaml.json to support new regex rules.


KirillOsenkov:
Longer term we should probably rewrite this in terms of File.ReadAllLines() and Enumerable.Concat(), but it's probably better not to do it in 7.6. 

So this seems OK.

mkrueger:
the idea is to not need to read the whole file here.

slluis:
@monojenkins backport master

KirillOsenkov:
Good point, then we could use `File.ReadLines`, which is lazy.

PR #5783, opened 2018-08-28
[Ide] Safe cast to MDW in CommentTaskProvider

Therzok:
Can throw for Razor Workspaces
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/672552


KirillOsenkov:
You could almost get away with this, if not for the definite assignment rules:

image

Therzok:
There's 1338 opened at csharplang that could fix that

Therzok:
Which makes me think that `args.Workspace aint MonoDevelopWorkspace ws` is cleaner than `isnot`

PR #5784, opened 2018-08-28
[Core] Fix NRE in writing ProjectReference of SharedAssetProject

Therzok:
Fixes VSTS #656296 - [Feedback] VSfM Version 7.6 Preview (7.6 build 2003) crashes while building solution

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectReference.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectReference.cs
@@ -288,7 +288,7 @@ namespace MonoDevelop.Projects
 		{
 			// If the project is not supported, don't try to update any metadata of the property,
 			// just leave what was read
-			if (OwnerProject.IsUnsupportedProject)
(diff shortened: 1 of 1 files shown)


mrward:
This fixes the null reference exception in the IDE log. The bug itself seems to be the remote msbuild host crash that is fixed here - https://github.com/mono/monodevelop/pull/5780 - so this is to fix the null ref exception but not the remote host crash? Was confused since I could not see how this could fix the msbuild remote host crash. Just wondering if a comment on the bug report is needed?

Therzok:
Yeah, it does not fix the builder crash, but we should be able to recover here in case of a crash (null owner project). The other code surrounding this does do a null check for owner project.

mrward (approved):
OK. 👍 

I think this was confusing QA since they were expecting it to fix the remote msbuild host crash since it links to that bug.

Therzok:
I'm wondering whether this would mask bugs though.

Therzok:
Closing. Seems to be caused by builder carsh.

PR #5791, opened 2018-08-29
[Ide][Performance] Send feedback on a background thread

sevoku:
Fixes VSTS #668256


slluis:
Do we use the feedback service at all? if not let's just remove it.

Therzok:
Stub and obsolete?

PR #5796, opened 2018-08-29
Separate file saving logic from DoBeforeCompileAction

olegoid:
Needed to fix VSTS: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/669201

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
@@ -1707,16 +1707,21 @@ namespace MonoDevelop.Ide
 
 		async Task<BuildResult> DoBeforeCompileAction ()
 		{
+			return await ApplySavePolicy ();
+		}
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
```diff
@@ -1707,18 +1707,23 @@ async Task SaveAllFiles (BuildResult result)
...
-			switch (action) {
-			case BeforeCompileAction.Nothing: break;
-			case BeforeCompileAction.PromptForSave: await PromptForSave (result); break;
-			case BeforeCompileAction.SaveAllFiles: await SaveAllFiles (result); break;
-			default: System.Diagnostics.Debug.Assert (false); break;
-			}
-			
-			return result;
-		}
+			return await ApplySavePolicy ();
```
Therzok: Can we drop the async in `DoBeforeCompileAction` and just `return ApplySavePolicy ();`?

PR #5797, opened 2018-08-29
Bring back CompletionWindowManager.WordCompleted event

KirillOsenkov:
It is used by WebTools.

Longer term there's the bug tracking figuring out whether we can get rid of this, but for now we need this back: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/673887


Therzok:
Why didn't compat check fail when removing this?

KirillOsenkov:
The actual event wasn't removed, it was just obsoleted and became inert (never raised). But it was still there.

KirillOsenkov:
It was removed here: 
https://github.com/mono/monodevelop/commit/912d9cc00c91be860a30c8ff33789f7667d7dbc1

PR #5798, opened 2018-08-30
Increase test timeout.

KirillOsenkov:
Extract constant.


KirillOsenkov:
We'll want it in release-7.7 as well (if it passes and is approved) because 7.7 builds are impacted by this a lot.

PR #5800, opened 2018-08-30
Fixes VSTS User Story 586682: Undo refactor should undo all changes

mkrueger:
from refactoring

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/586682

Global undo & redo is quite complex and a feature that could lead to
data loss. Integrating it with the local undo stack like VS.NET does
is atm problematic. But using global undo/redo operations do work.


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/GlobalUndoService.WorkspaceUndoTransaction.cs
```diff
@@ -0,0 +1,270 @@
...
+			string description;
+
+			public string Description => description;
+			List<Change> changes = new List<Change> ();
+			Dictionary<string, ITextSource> originalDocuments = new Dictionary<string, ITextSource> ();
+
+			// indicate whether undo transaction is currently active
+			bool transactionAlive;
+
+			public WorkspaceUndoTransaction (Workspace workspace, GlobalUndoService globalUndoService, string description)
```
Therzok: Roslyn's transaction is a ForegroundThreadAffinitizedObject, that means we need to assert that we're constructing this on the UI thread.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/GlobalUndoService.WorkspaceUndoTransaction.cs
```diff
@@ -0,0 +1,270 @@
...
+				this.description = description;
+			}
+
+			public void AddDocument (DocumentId id)
+			{
+				// nothing
+			}
+
+			public void Commit ()
+			{
```
Therzok: This should be done on the UI thread.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/GlobalUndoService.WorkspaceUndoTransaction.cs
```diff
@@ -0,0 +1,270 @@
...
+				}
+				changes.Add (new TextReplaceChange (this, fileName, offset, removedChars, insertedText));
+			}
+
+			public void RenameFile (string oldName, string newName)
+			{
+				changes.Add (new RenameChange (oldName, newName));
+			}
+
+			public void Dispose ()
```
Therzok: Check that this is done on the UI thread.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/GlobalUndoService.WorkspaceUndoTransaction.cs
```diff
@@ -0,0 +1,270 @@
...
+				} catch (OperationCanceledException) {
+					CloseOpenDocuments ();
+					return;
+				}
+
+				// reset documents
+				foreach (var doc in originalDocuments) {
+					var data = GetTextEditorData (doc.Key);
+					if (doc.Value?.Version != null && data.Version.BelongsToSameDocumentAs (doc.Value.Version)) {
+						foreach (var change in doc.Value.Version.GetChangesTo (data.Version).ToList()) {
```
Therzok: Why do we `ToList` here?

slluis:
Mike, we need unit tests for this. Also, test instructions for QA.

It seems like this PR only adds global undo support for rename operations, is that right?

slluis:
We can't have a command specific for global undo. Can't we just push a special global undo step to the editor's undo stack after a refactoring operation?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/GlobalUndoService.cs
```diff
@@ -0,0 +1,157 @@
...
+		protected override void Run ()
+		{
+			if (GlobalUndoHandler.NagScreen())
+				GlobalUndoService.Redo ();
+		}
+	}
+
+	interface IMonoDevelopUndoTransaction : IWorkspaceGlobalUndoTransaction
+	{
+		void TextReplace (string fileName, int offset, int removedChars, string insertedText);
```
KirillOsenkov: I would design this differently. To me it seems like it would be good to group each action's Undo and Redo logic together in a concept called Action, and have two independent Actions, such as TextReplace and RenameFile.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/GlobalUndoService.cs
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	/// <summary>
+	/// A service that provide a way to undo operations applied to the workspace
+	/// </summary>
+	[ExportWorkspaceService (typeof (IGlobalUndoService), ServiceLayer.Host), Shared]
+	partial class GlobalUndoService : IGlobalUndoService
+	{
+		static Stack<WorkspaceUndoTransaction> undoStack = new Stack<WorkspaceUndoTransaction> ();
+		static Stack<WorkspaceUndoTransaction> redoStack = new Stack<WorkspaceUndoTransaction> ();
```
KirillOsenkov: This feels like this could be a generic IDE service, not related to Roslyn or workspaces. It feels like it should be an independent entity that anyone can use.

KirillOsenkov:
If we're designing an Undo system why not do it in a way that makes it independent of workspaces and general purpose?

See for example https://github.com/KirillOsenkov/Undo. Not to say we should take this (or copy the code), but it feels like actions should group the Undo and Redo logic together.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.Rename/RenameItemDialog.cs
```diff
@@ -206,43 +209,57 @@ async void OnOKClicked (object sender, EventArgs e)
 			var properties = Properties;
 			((Widget)this).Destroy ();
 			try {
-				var changes = await this.rename (properties);
-				ProgressMonitor monitor = IdeApp.Workbench.ProgressMonitors.GetBackgroundProgressMonitor (Title, null);
-
+				var workspace = TypeSystemService.Workspace;
+				using (var undoTransaction = workspace.OpenGlobalUndoTransaction (GettextCatalog.GetString ("Rename"))) {
+					IList<Change> changes;
+					if (rename != null) {
```
KirillOsenkov: Why are there two codepaths here? (rename == null and rename != null)?

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.Rename/RenameItemDialog.cs
```diff
@@ -206,43 +209,57 @@ async void OnOKClicked (object sender, EventArgs e)
...
+						changes = await this.rename (properties);
+					} else {
+						var newSolution = await renameRefactoring.GetNewSolution (symbol, properties);
```
KirillOsenkov: This code is duplicated elsewhere - is there a chance to unify it?

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.Rename/RenameItemDialog.cs
```diff
@@ -206,43 +209,57 @@ async void OnOKClicked (object sender, EventArgs e)
...
+					}
+					ProgressMonitor monitor = IdeApp.Workbench.ProgressMonitors.GetBackgroundProgressMonitor (Title, null);
 
-				if (ChangedDocuments != null) {
-					AlertButton result = null;
-					var msg = new QuestionMessage ();
-					msg.Buttons.Add (AlertButton.MakeWriteable);
-					msg.Buttons.Add (AlertButton.Cancel);
-					msg.AllowApplyToAll = true;
+					if (ChangedDocuments != null) {
```
KirillOsenkov: It would be nice to extract a separate method whose sole purpose is to go through the files and ask the user if any file is read-only. Otherwise this functionality is tangled into this method.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring/Change.cs
```diff
@@ -130,7 +136,12 @@ public override void PerformChange (ProgressMonitor monitor, RefactoringOptions
 				textEditorData.ReplaceText (Offset, RemovedChars, InsertedText);
 			}
 		}
-		
+
+		internal override void CommitTransaction (IMonoDevelopUndoTransaction transaction)
+		{
+			transaction.TextReplace (FileName, Offset, RemovedChars, InsertedText);
```
KirillOsenkov: Not sure why this redirection is needed, because the logic to apply change is right above this method - PerformChange. The code is harder to understand when the control is bouncing back and forth between entities.

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SourceEditorView.cs
```diff
@@ -3437,6 +3437,11 @@ Task<ScopeStack> ITextEditorImpl.GetScopeStackAsync (int offset, CancellationTok
 			return TextEditor.GetLineHeight (line);
 		}
 
+		void ITextEditorImpl.ResetQuickDiff ()
+		{
+			TextEditor.Document.DiffTracker.Reset ();
```
KirillOsenkov: Maybe add a comment on why this is needed?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/InternalExtensionAPI/ITextEditorImpl.cs
```diff
@@ -234,5 +234,7 @@ interface ITextEditorImpl : IDisposable
 		Task<ScopeStack> GetScopeStackAsync (int offset, CancellationToken cancellationToken);
 
 		double GetLineHeight (int line);
+
+		void ResetQuickDiff ();
```
KirillOsenkov: This feels oddly specific. Is there another way to reset the Quick Diff at just the right time, without expanding this interface?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/GlobalUndoService.WorkspaceUndoTransaction.cs
```diff
@@ -0,0 +1,270 @@
...
+using MonoDevelop.Ide.Editor;
+using MonoDevelop.Core.Text;
+using System.Linq;
+using System.IO;
+
+namespace MonoDevelop.Ide.TypeSystem
+{
+	partial class GlobalUndoService
+	{
+		abstract class Change
```
KirillOsenkov: Why are we creating a second family of "Change" classes? Shouldn't we unify on one?

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.Rename/RenameItemDialog.cs
mkrueger: the other one is used by F#. It's more generic - the C# one is more sepecific.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/GlobalUndoService.WorkspaceUndoTransaction.cs
mkrueger: Not without API break.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/InternalExtensionAPI/ITextEditorImpl.cs
mkrueger: It is oddly specific.

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/SourceEditorView.cs
mkrueger: Will go away with the new editor in a few weeks. It resets the quick diff state. The method name is very specific.

mkrueger:
@slluis: "We can't have a command specific for global undo. Can't we just push a special global undo step to the editor's undo stack after a refactoring operation?"

Why can't we have one ? VS.NET has one. I tried several approaches atm mixing that with the editors stack is not easily possible and may do harmful things. In fact global undo is a very, very dangerous operation. I think I would even prefer not having any form of global undo.

In any case a gloabl undo needs a global undo stack in conjunction.

@KirillOsenkov : "If we're designing an Undo system why not do it in a way that makes it independent of workspaces and general purpose?"

It is - however we need to implement the roslyn interfaces. But basically it can take anything. There is nothing to prevent implementing a second or third approach. But as said roslyn inferfaces need to be implemented.

Basically this approach is a hack for the specific case. We need to get the VS.NET undo stack running to get a better integration.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/GlobalUndoService.cs
mkrueger: Basically that's a hack - we're getting the undo system from vs.net with the new editor then we can use the vs.net implementation of the global undo service.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/GlobalUndoService.cs
mkrueger: Basically it is not related to roslyn - you're looking at implementation details there.

main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/Document/TextDocument.cs
```diff
@@ -53,55 +54,51 @@ class TextDocument : ITextDocument, IDisposable
...
+		bool lineEndingMismatch;
+
+		//HACK ImmutableText buffer;
+		//HACK readonly ILineSplitter splitter;
+
+		ISyntaxHighlighting syntaxMode = null;
+
+		//HACK TextSourceVersionProvider versionProvider = new TextSourceVersionProvider ();
+
+		bool readOnly;
```
KirillOsenkov: >  [](start = 6, length = 1)

The file got formatted (likely accidentally). Would be nice to have the formatting changes separately.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring.Rename/RenameItemDialog.cs
KirillOsenkov: Would you mind adding more comments in places like these? So that the readers later understand what these all are for?

PR #5806, opened 2018-08-30
Fixes VSTS Bug 673555: Brace and Indent completion is all kinds of

mkrueger:
broken.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/673555

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/MonoTextEditor.ITextView.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/MonoTextEditor.ITextView.cs
@@ -554,7 +554,8 @@ namespace Mono.TextEditor
 			// We need to manually synchronize our caret with what MultiSelectionBroker thinks the caret is.
 			// The other direction happens when we move our caret.
 			if (TextCaret.Position.VirtualBufferPosition != MultiSelectionBroker.PrimarySelection.InsertionPoint) {
-				TextCaret.MoveTo (MultiSelectionBroker.PrimarySelection.InsertionPoint);
+				// This sync breaks on type formatting/desired column calculation and some other things.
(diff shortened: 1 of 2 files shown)


mkrueger:
Would be possible to use the vs.net caret however that requires way more infrastructural changes. 
For now disabling this line seems to fix the typing behavior.

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/MonoTextEditor.ITextView.cs
```diff
@@ -553,8 +553,9 @@ private void OnMultiSelectionSessionChanged (object sender, EventArgs e)
 			// The MultiSelectionBroker API has been updated, but currently in VSMac we still have a separate Caret concept.
 			// We need to manually synchronize our caret with what MultiSelectionBroker thinks the caret is.
 			// The other direction happens when we move our caret.
-			if (TextCaret.Position.VirtualBufferPosition != MultiSelectionBroker.PrimarySelection.InsertionPoint) {
-				TextCaret.MoveTo (MultiSelectionBroker.PrimarySelection.InsertionPoint);
+			if (TextCaret.Position.VirtualBufferPosition != MultiSelectionBroker.PrimarySelection.InsertionPoint) {
+				// This sync breaks on type formatting/desired column calculation and some other things.
+				//	TextCaret.MoveTo (MultiSelectionBroker.PrimarySelection.InsertionPoint);
```
slluis: This code was added very recently here: https://github.com/mono/monodevelop/pull/5789. @KirillOsenkov we have a conflict.

KirillOsenkov: I've filed https://devdiv.visualstudio.com/DevDiv/_workitems/edit/674296 for myself to find a proper solution here.

PR #5814, opened 2018-08-30
Fix VSTS 674296 - fix sync with MultiSelectionBroker.

KirillOsenkov:
There was a problem in the logic where we synchronized the caret with the latest position of the MultiSelectionBroker.

We should first set the caret to the new position, then synchronize the MultiSelectionBroker to it. The ordering was wrong, causing smart indent not working.


KirillOsenkov:
This together with https://github.com/mono/monodevelop/pull/5815 should fix a few Editor issues. Without https://github.com/mono/monodevelop/pull/5815 things are still broken, so if you test this PR alone it will still be buggy.

slluis:
@mkrueger can you review?

mkrueger:
Looks good - the formatting changes were too long in a separate branch they broke badly wtih the multi selection broker changes a few days ago :/.

KirillOsenkov:
I'm happy we got all of these changes in finally!

PR #5815, opened 2018-08-30
Fix usage of ISmartIndent MEF component.

KirillOsenkov:
We had a first-chance MEF composition exception because we were expecting a single ISmartIndentProvider, but getting 5 instead. The right way to do this is to get an ISmartIndentationService and use that instead. This service will aggregate various ISmartIndentProviders and use the right one for the right view.

There were some editor problems because of this - smart indent not working, etc.


slluis:
@mkrueger can you review?

PR #5820, opened 2018-08-31
Fix VSTS 669018. Tab to complete IntelliSense not working in .cshtml …

KirillOsenkov:
…files.

We need to instantiate EditorOperations, because it in turn will register the UndoHistory
for the buffer via:
https://github.com/KirillOsenkov/vs-editor-api/blob/d06adf1581eb8e16242c8b6eabc7ba13ceaf0d54/src/Text/Impl/EditorOperations/EditorOperations.cs#L108

Without Undo History Roslyn Completion bails via:
https://github.com/dotnet/roslyn/blob/a107b43dcad83cf79addd47a9919590c7366d130/src/EditorFeatures/Core/Implementation/IntelliSense/Completion/Controller_Commit.cs#L66

See https://devdiv.visualstudio.com/DevDiv/_workitems/edit/669018 for details.

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/MonoTextEditor.ITextView.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/MonoTextEditor.ITextView.cs
@@ -72,7 +72,8 @@ namespace Mono.TextEditor
 		{
 			get {
 				if (editorOperations == null) {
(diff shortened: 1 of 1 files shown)


KirillOsenkov:
I regressed this when updating to Editor 15.8:
https://github.com/mono/monodevelop/commit/5a6e365a2b2d8c3105e02126dc726a55d603f3ec#diff-518b01f5b69b5cb7ff53a2ad4e8abccbR76

PR #5823, opened 2018-08-31
Fixes VSTS Bug 634581: JSON IntelliSense inserts three quotes where

mkrueger:
two are needed under certain circumstances

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/634581

Basically that removes the default auto insert bracket handler for the
.json case. However the question is if it should be disabled for any
language. However atm it's only useful in the html/xml case. But
whitelisting all html/xml mime types is hard and even for xml it needs
to be improved a lot.

Mid term development approach: Remove that and rely on text mate
properties for open/closing brace insertion vs code has that. Remove
extensibility of that feature completly. We've a roslyn based approach
for that. (Language extensions could always provide own approaches)

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/DefaultAutoInsertBracketHandler.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/DefaultAutoInsertBracketHandler.cs
@@ -24,11 +24,14 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 using System;
+using System.Collections.Generic;
 using System.Threading;
 using Mono.TextEditor.Highlighting;
+using MonoDevelop.Core.Text;
 using MonoDevelop.Ide;
 using MonoDevelop.Ide.Editor;
 using MonoDevelop.Ide.Editor.Extension;
+using System.Linq;
 
 namespace MonoDevelop.SourceEditor
 {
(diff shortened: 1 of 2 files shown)


main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/DefaultAutoInsertBracketHandler.cs
```diff
@@ -37,25 +40,28 @@ class DefaultAutoInsertBracketHandler : AutoInsertBracketHandler
...
+			"text/fsharp", "text/x-csharp", "text/x-json"
+		};
+
 		public override bool Handle (TextEditor editor, DocumentContext ctx, KeyDescriptor descriptor)
 		{
-			if (descriptor.KeyChar == '\'' && editor.MimeType == "text/fsharp")
+			if (descriptor.KeyChar == '\'')
 				return false;
-			if (editor.MimeType == "text/x-csharp")
+			if (excludedMimeTypes.Contains(editor.MimeType))
```
Therzok: This is a LINQ extension. Prefer IndexOf

Therzok: Pushed a commit.

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/DefaultAutoInsertBracketHandler.cs
```diff
@@ -37,25 +40,28 @@ class DefaultAutoInsertBracketHandler : AutoInsertBracketHandler
...
 		const string closingBrackets = "}])'\"";
 
+		static readonly string[] excludedMimeTypes = {
+			"text/fsharp", "text/x-csharp", "text/x-json"
+		};
+
 		public override bool Handle (TextEditor editor, DocumentContext ctx, KeyDescriptor descriptor)
 		{
-			if (descriptor.KeyChar == '\'' && editor.MimeType == "text/fsharp")
+			if (descriptor.KeyChar == '\'')
```
ToddGrun: Seems like this would always ignore single quotes, not intended right?

KirillOsenkov:
Turns out this PR didn't fix the issue. I should have tried before I merged. Turns out the DefaultBraceCompletion wasn't even a problem since WebTools provides our own BraceCompletion that supercedes the default one. 

See related fix from Mikayla here:
https://github.com/mono/monodevelop/pull/4598

The real fix is in https://github.com/mono/monodevelop/pull/5831.

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor/DefaultAutoInsertBracketHandler.cs
KirillOsenkov: good catch!

PR #5830, opened 2018-08-31
Resolve System.IO.FileSystem and friends from 4.7.1 Facades.

KirillOsenkov:
Though this solves the problem of copying the System.IO.FileSystem.dll to output it may be a problem because now the Facades for 4.7.1 (targeting pack) needs to be installed on the machine to build MonoDevelop and I'm not sure this is acceptable. 

Also I'm not sure how it will build on a Mac because I don't know how Facades work there.


KirillOsenkov:
Ideally we'd resolve these from NuGet, but I haven't found these versions of the .dlls on NuGet.

PR #5831, opened 2018-08-31
Fix VSTS 634581 (triple """ in Json)

KirillOsenkov:
There was a special case where if the char is punctuation and the current completion list item ends in that char then it used to commit completion.

Now we detect the situation where if we're in a Json editor and the completion is one of three Json special completions " ", { } or [ ] and we press either " [ or { then we hide completion, send the char through to the editor and don't commit.

This fixes the problem because it lets Brace Completion commit the second character and lets user overtype it if they type it manually.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionController.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeCompletion/CompletionController.cs
@@ -999,7 +999,12 @@ namespace MonoDevelop.Ide.CodeCompletion
 					if (selectedItem < 0 || (dataList != null && selectedItem >= dataList.Count)) {
 						return KeyActions.CloseWindow;
 					}
-					if (dataList [selectedItem].DisplayText.EndsWith (descriptor.KeyChar.ToString (), StringComparison.Ordinal)) {
+					var displayText = dataList[selectedItem].DisplayText;
+					// Workaround for Json bug https://devdiv.visualstudio.com/DevDiv/_workitems/edit/634581
+					// If the completion item is " " or { } or [ ] and the char is " or { or [ then don't commit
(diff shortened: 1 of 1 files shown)


KirillOsenkov:
@monojenkins backport master

mkrueger:
Approving however I still feel it's something that should be done in the brace completion. Brace completion needs to be done by the language backends. 

But y that may work in that special case but may not work for different language (json editor isn't detected there - right ?). However all that code there is going to be replaced with the new editor therefore it doesn't matter that much.

slluis:
I also don't like this. Couldn't this be fixed by adding a new "DisableCommit" property to CompletionData that is set to True by the Json extension?

parasAmbhore:
I can confirm that the issue in not reproducible when tested using shared steps.
Hence marking it as QA Approved.

Build taken from: https://dl.internalx.com/artifacts/devdiv-1996979/VisualStudioForMac-7.6.4.5.dmg

Old Build with Bug: https://gist.github.com/p-lad/b688302dfe1af95f8ca4d7ebab25035f

Screen Cast with Bug: https://screencast.com/t/t3ObacCLpx

Build Gist with Fix:  https://gist.github.com/p-lad/72878ca24366d7cf6ffc2d7c69ae8460

Screen Cast with Fix: https://screencast.com/t/S7aUbDIa

Please refer both the screencast and suggest if we need to check anything else apart from this scenario.


Thanks

KirillOsenkov:
@slluis @mkrueger I agree this is ugly :) However this is a safe, minimal, tactical fix for release-7.6 that fixes the bug and gets the job done.

Keep in mind that the bug is caused by the code right next to this fix, where if the completion item name ends with the character and the character is punctuation then we special case for "param:". That logic is C# specific and has no place in the controller. So basically my hack is refining the previous hack :)

Additionally, I think the WebTools behavior here may be slightly wrong too - on Windows completion doesn't disappear after you've typed " (and it should, because at that point completion has fulfilled its role and has become useless). We can't disable commit for these items because we do want them to commit on Tab and Enter, but we don't want them to commit on ". We could design more complex logic with per-item commit characters, but as Mike says, this whole thing is going away. 

I humbly request that this is merged as is ;) In the interest of getting the fix out to the users asap.

PR #5834, opened 2018-09-02
Fixing bug with copying build configuration

Jammer77:
If project has:
  ```
<PropertyGroup>
    <PreBuildEvent>some command</PreBuildEvent>
  </PropertyGroup>
```
MonoDevelop adding new build config after it and this project can't build any more. 
My fix changes this situation, and adding new configuration in the right way.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
@@ -802,8 +802,14 @@ namespace MonoDevelop.Projects.MSBuild
 			}
 			if (!added) {
 				if (insertAtEnd) {
-					var last = ChildNodes.FindLastIndex (g => g is MSBuildPropertyGroup);
-					if (last != -1) {
+
+                    var last = ChildNodes.FindLastIndex(g => g is MSBuildPropertyGroup &&
+                                         (g as MSBuildPropertyGroup).GetChildren()
+                                                                    .OfType<MSBuildProperty>()
+                                                                    .All(prop => prop.Name != "PreBuildEvent" &&
+                                                                                 prop.Name != "PostBuildEvent"));
+
+                    if (last != -1) {
 						ChildNodes = ChildNodes.Insert (last + 1, group);
 						added = true;
 					}

### CODE REVIEW

**Comment by slluis** [comment]
The order of the property groups should not matter in this case. I tried adding that property group to a project and then adding a configuration and the project built fine for me. Which error are you getting?

**Comment by Jammer77** [comment]
This is error i've  gotten.  I'm not just copying  configuration. I doing it in the project with preBuild  action.

> Library/Frameworks/Mono.framework/Versions/5.16.0/lib/mono/msbuild/15.0/bin/Microsoft.Common.CurrentVersion.targets(5,5): Error: The OutputPath property is not set for project 'ColorPicker.csproj'.  Please check to make sure that you have specified a valid combination of Configuration and Platform for this project.  Configuration='DebugMac'  Platform='AnyCPU'.  This error may also appear if some other project is trying to follow a project-to-project reference to this project, this project has been unloaded or is not included in the solution, and the referencing project does not build using the same or an equivalent Configuration or Platform. (ColorPicker)

**Comment by Jammer77** [comment]
For my solution i fixed it manually, but i investigated behavior of VSforWindows  and it write copied configuration right after configurations, but not after preBuild actions in the end of the xml.

**Comment by Jammer77** [comment]
image

**Comment by Jammer77** [comment]
image

**Comment by mrward** [comment]
1. Can you post the entire contents of the project file (.csproj)?
2. Does VS on Windows work with the Pre/PostBuildEvents at the start where VS Mac adds them?

**Comment by Jammer77** [comment]
[ColorPicker.csproj.zip](https://github.com/mono/monodevelop/files/2344756/ColorPicker.csproj.zip)

**Comment by Therzok** [comment]
Can we have a failing unit test that passes after this fix is applied?

**Review by Therzok — DISMISSED** [review]
Needs tests

**Comment by Jammer77** [comment]
Sorry, I have no time for test writing. If it's need it can implement somebody from the teem.

File: main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs [hunk 1]
```diff
@@ -801,9 +801,15 @@ public void AddPropertyGroup (MSBuildPropertyGroup group, bool insertAtEnd = tru
 				}
 			}
 			if (!added) {
-				if (insertAtEnd) {
-					var last = ChildNodes.FindLastIndex (g => g is MSBuildPropertyGroup);
-					if (last != -1) {
+				if (insertAtEnd) {
+
+                    var last = ChildNodes.FindLastIndex(g => g is MSBuildPropertyGroup &&
```
Therzok: This is a huge perf hit, as we will look up every property in every property group now.

Jammer77: Sure. 2-4 group property with 8-10 items inside, and we call this method only we have a new one configuration.

Jammer77:
So. Can i merge this issue? This is big pain. If you add new configure in the solution, you have to edit all projects manually about 1 day.

slluis (changes requested):
This can't be merged as is. The ordering of the property groups shouldn't matter in this case. We need a better diagnostic of the problem.

PR #5837, opened 2018-09-03
Fixes VSTS Bug 669421: CheckProjectIsInCurrentSolution Throws

mkrueger:
Exception

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/669421

It's unknown in which use case that happens but it shouldn't crash the
IDE. If a reference update doesn't reference to the current solution
some race is going on with project updates.
Skipping that event in that case is ok. 

In any case this patch catches exceptions in that update event so it no longer crashes the IDE.

As side effect that should also fix the IDE crash in:

Fixes Bug 669418: CheckProjectHasMetadataReference Exception

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectData.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectData.cs
@@ -29,6 +29,7 @@ using System.Collections.Immutable;
 using System.Threading;
 using Microsoft.CodeAnalysis;
 using MonoDevelop.Core;
+using ICSharpCode.NRefactory.MonoCSharp;
 
 namespace MonoDevelop.Ide.TypeSystem
 {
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectData.cs
```diff
@@ -64,8 +65,9 @@ void OnMetadataReferenceUpdated (object sender, EventArgs args)
 				lock (workspace.updatingProjectDataLock) {
 					if (!RemoveMetadataReference_NoLock (reference, workspace))
 						return;
+					if (!workspace.CurrentSolution.ContainsProject (projectId))
```
Therzok: Hmmm, something else might need to be under `updatingProjectData` lock

mkrueger: Without use case it's hard to tell what's going on there.

Therzok: Yeah, I don't see how the event handler would trigger here in a project that's not contained. We remove the data from roslyn's workspace before we trigger a project reload event, so I'm not sure what's going on.

Therzok: Found the root cause.

Therzok: It seems like there are races going on. https://github.com/mono/monodevelop/pull/5872/commits/8a665f7617fbff5ebc5f559d0d490ded38a96d5b might help a bit with this exception.

I'm not sure this exception happens in 7.7+. I refactored the workspace layer and I fixed quite a couple of races around our lock usage.

Therzok: @slluis ^

PR #5844, opened 2018-09-04
[Ide] Remove deprecated FeedbackService

sevoku:
As discussed in #5791, this PR removes the whole `FeedbackService`, which has not been used for years now. 

> Fixes VSTS [#668256](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/668256)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Extensions/ServiceUrlExtensionNode.cs
```diff
@@ -29,6 +29,7 @@
 
 namespace MonoDevelop.Ide.Extensions
 {
+	[Obsolete]
 	public class ServiceUrlExtensionNode : ExtensionNode
```
sevoku: Only this class is public, adding `Obsolete` attribute to not hard break the API.

slluis: Let's remove it too. It should have never been public.

slluis:

Therzok:
Rebased on top of current master to fix the build.

PR #5855, opened 2018-09-05
Updated reference to xamarin/md-addins@d396efb

xamarin-release-manager:
Updated reference to xamarin/md-addins@d396efb

Fixes designer https://devdiv.visualstudio.com/DevDiv/_workitems/edit/665173


Aguilex:
@slluis can you please help me figure out why this PR build failed? I wonder if it has anything to do with the Azure outage in the South Central US region (many of our systems/services are hosted in that region).

PR #5867, opened 2018-09-06
Fix binding redirects for reference assemblies

KirillOsenkov:
Looks like now that we're targeting 4.7.1 the binding redirects are no longer necessary and we don't need to copy the dlls to output either. The simplest option appears to work.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
@@ -144,14 +144,8 @@
     <Reference Include="System.IO.Compression">
       <HintPath>..\..\..\packages\System.IO.Compression.4.3.0\lib\net46\System.IO.Compression.dll</HintPath>
     </Reference>
-    <Reference Include="System.IO.FileSystem, Version=4.0.3.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
-      <Private>true</Private>
-      <SpecificVersion>true</SpecificVersion>
-    </Reference>
-    <Reference Include="System.IO.FileSystem.Primitives, Version=4.0.3.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
-      <Private>true</Private>
-      <SpecificVersion>true</SpecificVersion>
-    </Reference>
+    <Reference Include="System.IO.FileSystem" />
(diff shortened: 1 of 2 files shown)


KirillOsenkov:
I've been seeing very weird assembly loading issues on windows where fusion would tell me "Can't load reference assemblies for execution". Basically it makes sense since we were taking the reference assemblies from the Facades directory and copying them to output. Now that we had a binding redirect to the exact version that is in the output, it was taking precedence over loading an older version from the GAC.

I'm guessing that without the binding redirect and without the assembly in the output directory the version is loaded from the GAC, even if it doesn't exactly match the version that Roslyn is referencing.

Therzok (approved):

mrward:
Seeing some type load exceptions with this branch, the same as master, when I open an iOS project.

```
ERROR [2018-09-06 18:43:18Z]: Roslyn fatal exception
System.AggregateException: One or more errors occurred. ---> System.TypeLoadException: Could not resolve type with token 010000dc (from typeref, class/assembly System.Security.Cryptography.IncrementalHash, System.Security.Cryptography.Algorithms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a)
  at Microsoft.CodeAnalysis.Checksum.Create (Microsoft.CodeAnalysis.Serialization.WellKnownSynchronizationKind kind, Roslyn.Utilities.IObjectWritable object) [0x00024] in /_/src/Workspaces/Core/Portable/Workspace/Solution/Checksum_Factory.cs:57 
  at Microsoft.CodeAnalysis.ProjectInfo+ProjectAttributes.Microsoft.CodeAnalysis.IChecksummedObject.get_Checksum () [0x0000e] in /_/src/Workspaces/Core/Portable/Workspace/Solution/ProjectInfo.cs:535 
  at Microsoft.CodeAnalysis.Serialization.SerializerService.CreateChecksum (System.Object value, System.Threading.CancellationToken cancellationToken) [0x00028] in /_/src/Workspaces/Core/Portable/Execution/SerializerService.cs:56 
  at Microsoft.CodeAnalysis.ProjectState+<ComputeChecksumsAsync>d__114.MoveNext () [0x000c5] in /_/src/Workspaces/Core/Portable/Workspace/Solution/ProjectState_Checksum.cs:41 
```

PR #5869, opened 2018-09-06
Fix #5779: [VSfM] Wrong sections visually commented out with preprocessor directives

### LINKED ISSUE #5779: [VSfM] Wrong sections visually commented out with preprocessor directives
Version: Version 7.8 Preview (7.8 build 129)

It looks like the logic of what section to show is being inverted

Here is a video of the effect. You'll notice when I have it set to iOS the MOBILE section is visually commented out but when I hover my mouse over the class it shows that it's "Active"

Here's the same block of code on VS for Windows

> VS bug [#672064](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/672064)

DavidKarlas:
When opening document/switching between documents `InformDocumentOpen` is called which calls `InternalInformDocumentOpen` 1st on documentId with correct ProjectId(by correct I mean selected project, and then other projects) which results in Roslyn setting last project(so "not correct") as editing one... Which results in wrong highlighting... VS Window uses `OnDocumentContextUpdated` to update document when switching between projects, hence to keep changes to minimal and not rely on order of opening and hope that that will do right thing, we just call `OnDocumentContextUpdated` at end of opening sequence.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
@@ -387,6 +387,7 @@ namespace MonoDevelop.Ide.TypeSystem
 					InternalInformDocumentOpen (linkedDoc, editor, context);
 				}
 			}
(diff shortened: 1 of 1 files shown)


DavidKarlas:

PR #5877, opened 2018-09-06
[InfoBars] Make IInfoBarHost public

bratsche:
This is a replacement for https://github.com/mono/monodevelop/pull/5671


bratsche:
Marius also was doing this, so I'll close mine. https://github.com/mono/monodevelop/pull/5878

PR #5879, opened 2018-09-06
Fix VSTS 677075: JS triple "" inserted instead of two

KirillOsenkov:
This was a MimeType inserted into MonoDevelop so that a unit-test passes without WebTools installed. However this has some pretty drastic consequences - a completely different mime type chain is created for .js and .ts files, bypassing the base x-vs content type. As a result, the WebTooling VSAutoInsertBracketHandler was not included in the chain, and counterintuitively, this resulted in 3 quotes being inserted, not 2.

This regressed with this commit:
https://github.com/mono/monodevelop/commit/39002a65cc594bf0b436f1d2c17f9770ea23984a
https://github.com/mono/monodevelop/pull/5734


KirillOsenkov:
@monojenkins backport master

slluis:

PR #5883, opened 2018-09-07
[Ide] MetadataReferenceCache now removes the right snapshots

Therzok:
Previously, if the same metadata reference was reused between
multiple projects, we would fall in the problem of a snapshot
being updated on one metadata reference update, that means that
future metadata reference updates would try to remove the new one
from their store.

Send over the old and current snapshots to all event handler subscribers
to ensure we're removing and adding the right ones.

Fixes VSTS #669418 - CheckProjectHasMetadataReference Exception


Therzok:
@slluis

PR #5885, opened 2018-09-07
[Ide] Improve roslyn logging infrastructure

Therzok:
Move FatalError registration into RoslynService initialization.

Add a new RoslynFileLogger class that writes everything roslyn reports
into its own log file.

Keep the old one for database exceptions so we catch them in the ide log

Fixes VSTS #679463 - Log everything that roslyn reports to the log file


Therzok:
We might want this to happen behind a feature toggle. A few minutes of usage on a really small solution ended up with 6MB of logging.

Therzok:
Should be okay now.

Therzok:
Rebased, will merge once green.

PR #5893, opened 2018-09-08
Report an error and avoid exception when snippet version is unsupported.

KirillOsenkov:
I've noticed another case when reading a corrupted/unsupported snippet would throw and lose all other snippets.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeTemplates/CodeTemplateService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeTemplates/CodeTemplateService.cs
@@ -312,8 +312,10 @@ namespace MonoDevelop.Ide.CodeTemplates
 						switch (reader.LocalName) {
 						case Node:
 							string fileVersion = reader.GetAttribute (VersionAttribute);
-							if (fileVersion != Version) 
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.CodeTemplates/CodeTemplateService.cs
```diff
@@ -312,8 +312,10 @@ static List<CodeTemplate> LoadTemplates (XmlReader reader, string filePathForDeb
 						switch (reader.LocalName) {
 						case Node:
 							string fileVersion = reader.GetAttribute (VersionAttribute);
-							if (fileVersion != Version) 
-								return null;
+							if (fileVersion != Version) {
```
Therzok: Can't we just ignore this node?

Therzok: Ah, hm, this is in a CodeTemplates node, might be okay.

KirillOsenkov:
Ping @mkrueger - minor fix

KirillOsenkov:
For some reason it still wants @mkrueger's approval.

PR #5894, opened 2018-09-09
Fix find base type if it is generic instance.

Gankov:


Therzok:
The only thing I can think of is that it _might_ fail when there's both a generic type `MyType<T>` and a non-generic one `MyType`.

Gankov:
About conflict with non-generic same class name, i did not test, but i think it will work, because generic types has names like ```MyType`1<T>``` and without <> like ```MyType`1```, non-generic will named `MyType`.

Therzok:
Build starting soon, will merge as soon as it's green.

Issue #5873, opened 2018-09-06
Exception when using syntax highlighting in VS for Mac 7.6


jimbobbennett:
I have an add-in for VS for Mac that uses the cucumber syntax highlighting bundle for Sublime. This was working fine until the VS for Mac 7.6 update. Since then an exception is thrown for the syntax highlighting:

ERROR [2018-08-24 16:00:07Z]:
System.NullReferenceException: Object reference not set to an instance of an object
at MonoDevelop.Ide.Editor.Highlighting.SyntaxHighlightingService+<>c__DisplayClass30_1.b__1 () [0x00000] in /Users/vsts/agent/2.138.6/work/1/s/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/SyntaxHighlightingService.cs:308
at MonoDevelop.Ide.Editor.Highlighting.AbstractSyntaxHighlightingDefinitionProvider.GetSyntaxHighlightingDefinition () [0x0000f] in /Users/vsts/agent/2.138.6/work/1/s/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/ISyntaxHighlightingDefinitionProvider.cs:78
at MonoDevelop.Ide.Editor.Highlighting.SyntaxHighlightingService.GetSyntaxHighlightingDefinitionByName (MonoDevelop.Core.FilePath fileName) [0x000df] in /Users/vsts/agent/2.138.6/work/1/s/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/SyntaxHighlightingService.cs:715
at MonoDevelop.Ide.Editor.Highlighting.SyntaxHighlightingService.GetSyntaxHighlightingDefinition (MonoDevelop.Core.FilePath fileName, System.String mimeType) [0x00009] in /Users/vsts/agent/2.138.6/work/1/s/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/SyntaxHighlightingService.cs:748

The original issue raised in the add-in repo is here: https://github.com/straighteight/SpecFlow-VS-Mac-Integration/issues/12

The latest master branch from that repo reproduces this issue. Grab the code and debug it, creating a new spec flow project and you will see the crash. Removing the syntax highlighting bundle from the add-in fixes the crash.

[...]

PR #5897, opened 2018-09-10
Fixes issue #5873 Exception when using syntax highlighting in VS for

mkrueger:
Mac 7.6

It contained a .json file with less than 3 lines so the JSon scanning
didn't work. Also fixed that a single loading error prevented to add a
whole bundle.


mkrueger:

Issue #4805, opened 2018-05-15
No tooltips on toolbar


mhutch:
There are no tooltips on the toolbar for the run/build button, the run configuration selector, the build configuration selector, or the device selector.

I would propose something like this:

Build
Builds the projects in the active solution build configuration.

Debug
Debugs the project or projects in the active run configuration. Builds the projects in the active solution build configuration if necessary.

Run Configuration
A project or named set of projects and execution options that should be launched when running or debugging the solution.

Solution Build Configuration
A named set of projects and their configurations to be built when when building the solution.

Target Device
The device on which to deploy and launch the projects when running or debugging.

slluis:
In case a project can't be executed (so the toolbar shows the build icon instead of run), the tooltip should explain why execution is not available.

sevoku:
@slluis I've added just the tooltips (+a11y on Mac) in #5899, but it doesn't address https://github.com/mono/monodevelop/issues/4805#issuecomment-389476529. This could be handled by the [`MainToolbarController`](https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/MainToolbarController.cs#L802) which would need to update the toolbar through a new [`IRunButtonView.RunButtonDescription`](https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/IMainToolbarView.cs#L57).
However it'd need an API change, and I'd propose that we use the chance and rethink that API in general. For example, we could reuse the `CommandInfo` which is used internally and also align its metadata with what we show in the toolbar (more strings – espacially for a11y – are hardcoded in the mac platform addin).

Therzok:
The only thing we could possibly do to prevent future API breaks is make `IRunButtonView` only have a `RunButtonOptions` class which we can then extend in the future.

xamarin-release-manager:
Fixed in version 7.8.0.456 (master)

Pull Request #5899 merged by: Lluis Sanchez
Author: mono
Commit: a207e408aabef1eab061b222642ba1670bc5eecb (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #5899 merged by: Lluis Sanchez
Author: mono
Commit: a207e408aabef1eab061b222642ba1670bc5eecb (xamarin/monodevelop)

xamarin-release-manager:
[...]

PR #5899, opened 2018-09-10
Add missing main toolbar tooltips

sevoku:
As proposed in #4805 , this PR adds tooltips to the "Run" button and "Configuration" selectors (for Mac and Gtk) in the main toolbar. Additionally, the tooltip text is used for `AccessibilityHelp` on Mac.

Fixes issue #4805


slluis:

PR #5900, opened 2018-09-10
Fixes VSTS Bug 677772: NullReferenceException in get_IsFile

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/677772

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
@@ -215,7 +215,7 @@ namespace MonoDevelop.Ide.Gui
 		}
 
 		public bool IsFile {
-			get { return Window.ViewContent.IsFile; }
+			get { return Window?.ViewContent.IsFile == true; }
 		}
 		
 		public bool IsDirty {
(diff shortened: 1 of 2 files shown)


mkrueger:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
```diff
@@ -215,7 +215,7 @@ void OnFileNameChanged ()
 		}
 
 		public bool IsFile {
-			get { return Window.ViewContent.IsFile; }
+			get { return Window?.ViewContent.IsFile == true; }
```
Therzok: Why the null check?

Therzok: We already changed it so CheckFileChange does the null check before checking SkipView, so this might not be needed.

mkrueger: Because it may break in other IsFile calls.
Checking that for null is not uncommon in other properties/methods of Document.cs so it's more in line with the rest.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DocumentRegistry.cs
```diff
@@ -109,7 +109,17 @@ static void CheckFileChange (string fileName)
 			}
 
 			var changedViews = new List<DocumentInfo> ();
-			foreach (var view in openFiles) {
+			for (var i = 0; i < openFiles.Count; i++) {
+				var view = openFiles[i];
+
+				// Remove files that have already been closed/moved.
+				if (view.Document.Window?.ViewContent == null) {
```
slluis: Why do we need this? Files should already be removed from the list when a document is closed.

mkrueger:
It would've hardened against the issue. Telemetry bugs don't provide the information required to fix it.

PR #5902, opened 2018-09-10
[Ide] Track unhandled exceptions

iainx:
Tracks fatal unhandled exceptions in a counter

Fixes #503250


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Services.cs
```diff
@@ -161,5 +161,13 @@ class TimeToCodeMetadata : CounterMetadata
 			set => SetProperty (value);
 		}
 	}
+
+	public class UnhandledExceptionMetadata : CounterMetadata
```
mrward: Can we make this private? The rest are not public.

iainx: We need to access it inside md-addins, which isn't InternalsVisibleTo

Therzok: https://github.com/xamarin/md-addins/pull/3566/files#diff-402a6f9c6b2fc36fc92f859be5d3e13aR469

Therzok: You're not using it from md-addins though.

Therzok: Also, Xamarin.Ide is in IVT https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.csproj#L4268

Therzok: Ah, Xamarin.Ide.IdeCustomization is not.

mrward: It looks like we use the Exception from the metadata so we do not need the counter itself to be public.

iainx: All very good points, well made. I think I had planned to use it from md-addins, but then did it the property name method.

slluis:

PR #4863, opened 2018-05-21
[Roslyn] Integration configuration property mapping between the IDE and roslyn

Therzok:

PR #5906, opened 2018-09-10
Fix VSTS 669803: using vstool with Razor

KirillOsenkov:
vstool/mdtool was not setting the Runtime.MainSynchronizationContext, which in turn didn't set the MainTaskScheduler. Razor addin relies on the MainTaskScheduler being set. Repro: vstool build Foo.sln failed with ArgumentNullException.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
@@ -313,7 +313,7 @@ namespace MonoDevelop.Core
 			}
 			set {
 				if (mainSynchronizationContext != null && value != null)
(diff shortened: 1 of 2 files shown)


KirillOsenkov:

KirillOsenkov:

Therzok:
The reason I didn't do this change initially is that other vstool tools set their own MainTaskScheduler - i.e. BuildTool. Those need to be removed, as setting it will throw.

Therzok:
And if run-md-tests set it, other applications might.

Therzok (approved):

KirillOsenkov:
See also: https://github.com/mono/monodevelop/pull/4863

KirillOsenkov:

PR #5916, opened 2018-09-11
Bump RE to bring in some dupe analyzer/refactoring removal

Therzok:


Therzok:

mrward:
- Use new keyword - does not offer all the options that VS Mac 7.6 does. VS Mac 7.6 offers: Hide base member and Add 'new' modifier. Only seeing: Hide base member.
- Replace if with return - Not available.
- Convert to computed property - does not exist. Convert to full property generates different code. Are we removing convert to computed property?
- Unused code - OK
- Local variable not used - OK
- Convert to hex - OK
- Convert to dec - OK
- Invert if - OK
- Field can be made readonly - OK
- Use autoproperty - OK
- Convert if to conditional ternary - OK
- Convert to null coalescing - OK

Therzok:
> Use new keyword - does not offer all the options that VS Mac 7.6 does. VS Mac 7.6 offers: Hide base member and Add 'new' modifier. Only seeing: Hide base member.

don't these generate the same code?

> Replace if with return - Not available.

That should be: use conditional expression

> Convert to computed property - does not exist. Convert to full property generates different code. Are we removing convert to computed property?

That just generated a get/set property with both cases throwing a NotImplementedException.

mrward:
- Use new keyword - OK - Add 'new' modifier does the same thing as Hide base member so both menu items are not required.
- Use conditional expression - OK
- Convert to computed property - OK - removed.

PR #5927, opened 2018-09-11
[MSBuild] Build the MSBuild builder as Any CPU

Therzok:
Allowing it to be 64bit on mac

Fixes VSTS #681973 - MSBuildBuilder.exe is being run as 32bit


mhutch:
I hope it's safe to assume all tasks are 64-bit safe :)

Issue #5813, opened 2018-08-30
Editor IntelliSense/Code Generation Bugs


NickSpag:
VS for Mac: 7.6.1
macOS: 10.13.4
Log file: [MSFT Internal link](https://microsoft-my.sharepoint.com/:u:/p/nispagn/ETSaJz9eSn9DneXlgagB-a8BZ_t7NFPy4_4eDugJfedG9w?e=Ajb545) 

I looked around a bit at the issues, couldn't see anything related to these. The last few releases have had some intellisense and editor issues. I've documented the ones I can reproduce and remember below.

#### Deleting Adjacent Word
If you accept an intellisense recommendation while your cursor is at the beginning of another word, it deletes that word.

Gif of the behavior here: https://media.giphy.com/media/cRNeCtloiDs2qIzFBM/giphy.gif

#### Intellisense Doesn't Exit After Deleting All Text
Maybe this is normal behavior and I just noticed it, but seems counter-intuitive. Trigger an intellisense dialogue, and then delete all the letters so you're back to blank, and the intellisense dialogue is still open.

### Code Generation Intellisense
Use the "Create new method" intellisense recommendation in an Event Handler assignment. If you then change the method name or otherwise create an error in that assignment, clicking on the error (or selecting the text with the error) spawns the "Create new method" dialog once for each click, or multiple times when selecting/using arrow keys. You then have to escape out of that "stack" of dialogs before doing anything else. This can also happen as you're typing when the intellisense dialog is still open, if you previously left that dialog by arrowing to a new line and returned to it (since the uncompleted text will now have an error, presumably) and it starts creating dialogs behind the intellisense window. Bit less consistent to reproduce, once it happens though, it's a mess. I believe it leaves some sort of error artifact because text around it starts getting the red arrow underline. 

Note: None of this happens when you use the right click generate method, just if the method was originally created with the intellisense recommendation.

Gif of the core bug behavior here: https://media.giphy.com/media/61Z3DfMKdGWUN93sIp/giphy.gif

Hope it helps.

> VS bug [#674334](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/674334)

mkrueger:
The adjacent word behavior is something we can't change atm. This is from VS.NET and is used for changing the word to the one from intellisense. 

I agree its too over eager in that use case but with do what I want features it's always problematic.

The second one is a bug from us we can change.

xamarin-release-manager:
[...]

PR #5970, opened 2018-09-14
Fixes issue #5813 Editor IntelliSense/Code Generation Bugs.

mkrueger:
This ensures that insertion mode is only called for the currently
opened document in the real editor and not for a preview. However the
bug described seems already to be fixed with one of the recent roslyn
updates but guarding against that case improves this code.


slluis:

PR #5975, opened 2018-09-15
[MacPlatform] Fix credential provider dialog tabbing and focus

mrward:
The credential provider dialog did not focus the username text field
and tabbing did not move around the text fields and buttons.

Fixes VSTS #603815 - Credential provider dialog: tabbing and initial
focus

### DIFF
--- a/main/src/addins/MacPlatform/MacProxyCredentialProvider.cs
+++ b/main/src/addins/MacPlatform/MacProxyCredentialProvider.cs
@@ -156,24 +156,10 @@ namespace MonoDevelop.MacIntegration
 					view.AddSubview (passwordInput);
 
 					alert.AccessoryView = view;
+					alert.Window.WeakDelegate = new PasswordAlertWindowDelegate (usernameInput, passwordInput, cancelButton, okButton);
 					alert.Window.InitialFirstResponder = usernameInput;
-
-					EventHandler handler = (o, e) => {
-						// The NSAlert defines the keyviewloop after it is displayed so the tab order is defined
-						// here otherwise once the focus is on the OK and Cancel buttons it is not possible to tab
-						// to the username and password NSTextFields.
-						usernameInput.NextKeyView = passwordInput;
-						passwordInput.NextKeyView = cancelButton;
-						okButton.NextKeyView = usernameInput;
-					};
-
-					try {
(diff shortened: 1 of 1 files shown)


main/src/addins/MacPlatform/MacProxyCredentialProvider.cs
```diff
@@ -155,9 +156,24 @@ static ICredentials GetCredentialsFromUser (Uri uri, IWebProxy proxy, Credential
 					view.AddSubview (passwordInput);
 
 					alert.AccessoryView = view;
+					alert.Window.InitialFirstResponder = usernameInput;
+
+					EventHandler handler = (o, e) => {
```
Therzok: Can't we just set the alert windows' Delegate? Keep reference to the fields with `WeakReference` and that way we don't have to deal with event handlers

Therzok: I'll take over this, Matt is on FTO.

Therzok:
Tried Matt's patch without any changes, can't seem to tab between the fields and the buttons.

Therzok:
I tried a local build with High Sierra installed. I'll try a dmg.

mrward:
The fix is setting the first responder to the username text field and setting the key view order after the NSAlert is displayed. If there is a better way to do the second part then that is great.

Setting the key view order after the NSAlert is displayed was based this github source code - https://github.com/AzureAD/azure-activedirectory-library-for-objc/pull/989/files - however here they use a sheet but we are using RunModal. Setting the key view order before the NSAlert is displayed would half work - you could tab from the text box down to the buttons but once on the buttons you could not tab back to the text fields.

Therzok:
This is what I see locally. Tab doesn't work to jump from text field to button
kapture 2018-09-17 at 14 55 07

mrward:
My guess is that you do not have All controls enabled in Mac's Keyboard Settings.

keyboardsettings

You cannot tab to the buttons if All controls is not selected. I think the default is Text boxes and lists only. I can reproduce what you are seeing if I select Text boxes and lists.

I would guess enabling All controls will allow your delegate code to work.

Therzok:
Works with both event handler and delegate versions! 👍 Added verified

slluis:

slluis:

PR #5990, opened 2018-09-17
Primitive IWaitIndicator implementation

Therzok:
This is a left-over from when I thought fix all required this. This is not needed until we go full stack code actions implemented in roslyn (which is out of scope for initial fix all dialog support)


Therzok:
Will probably depend on ProgressDialogMonitor being done like https://devdiv.visualstudio.com/DevDiv/_workitems/edit/677555

Therzok:
Obsolete

PR #6046, opened 2018-09-20
Revert "[Ide] Fix race crash in log pad"

Therzok:
This reverts commit 92241d43d03221a075fe390de50bbec69aaea884.

This fixes the output pad not having anything written in it,
and instances not being reused

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs
@@ -25,19 +25,19 @@
 // THE SOFTWARE.
 
 using System;
-using System.Collections.Generic;
-using System.IO;
-using System.Linq;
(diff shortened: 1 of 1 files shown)


Therzok:
cc @mhutch

mrward:
I think a fix here is to move the padCookie assignment after the Clear call in the constructor. Right now it looks like the cookie is always invalid since in nearly all cases Clear is called in the constructor.

https://github.com/mono/monodevelop/blob/92241d43d03221a075fe390de50bbec69aaea884/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs#L869-L874

clearConsole is normally true. I think it is only false for some cases in the NuGet addin. Right now it looks like the padCookie is incremented on Clear so nothing will be written. I think this also affects build output.

Therzok:
Let me try that!

Therzok:
There's also a `base.OnCompleted` call that's been removed. I'll add that back.

PR #6047, opened 2018-09-20
[Output Pad] Fix nothing being written and monitors not being reused

Therzok:
The cookie was set before clear was called, making it not write anything.


Therzok:

PR #6050, opened 2018-09-20
Revert "Simplify references to Gtk assemblies."

alanmcgovern:
This reverts commit ab5e55090188f27a54ae6acf57df552f64814314.
This has two major issues. The first is that by bundling the managed
assemblies we also bundle the .dll.config files, and those .dll.config
files are *not* intended to be bundled. They contain the full path
to the installed files in the *Mono.framework* directory. For
example:
	`<dllmap dll="libglade-2.0-0.dll" target="/Library/Frameworks/Mono.framework/Versions/5.14.0/lib/libglade-2.0.0.dylib"/>`

Secondly, we cannot bundle the managed assemblies without also bundling
the native assemblies. They are strongly tied together so we have to load
a matching pair from the system or a matching pair from our app bundle.

This is also true for Mono.Posix.dll as it relies on libmonoposix.dylib.


Therzok (approved):

KirillOsenkov:
Let me quickly prepare a simpler fix.

KirillOsenkov:
We no longer need this (hopefully!)

PR #6051, opened 2018-09-20
[Telemetry] Add the correct StartupType values for the startup event

iainx:
fixes VSTS #678313


main/src/core/MonoDevelop.Ide/options/MonoDevelopProperties.xml
```diff
@@ -3,6 +3,7 @@
   <Properties>
     <Property key="MonoDevelop.Core.FirstRun" value="True" />
     <Property key="MonoDevelop.Core.Gui.Dialog.TipOfTheDayView.ShowTipsAtStartup" value="False" />
+	<Property key="MonoDevelop.Core.RestartRequested" value="False" />
```
Therzok: Spacing is off

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
```diff
@@ -842,13 +848,33 @@ static IdeCustomizer LoadBrandingCustomizer ()
...
 		{
 			var assetType = StartupAssetType.FromStartupInfo (startupInfo);
+			StartupType startupType = StartupType.Normal;
+
+			if (startupInfo.Restarted && !IdeApp.IsInitialRunAfterUpgrade) {
+				startupType = StartupType.ConfigurationChange; // Assume a restart without upgrading was the result of a config change
+			} else if (IdeApp.IsInitialRun) {
+				startupType = StartupType.FirstLaunch;
+			} else if (Debugger.IsAttached) {
+				startupType = StartupType.LaunchedAsDebugger;
```
slluis: I don't think LaunchedAsDebugger applies to VSMac. On Windows this happens when an application crashes and the debug button is clicked on the crash dialog. This is unrelated to the VSMac process being debugged or not.

Therzok: We could be debugging a bundle via `Debug Application`.

Therzok: might be worth taking these out as they skew results

iainx: I think I meant `DebuggerPresent` not `LaunchedAsDebugger`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
```diff
@@ -842,13 +848,33 @@ static IdeCustomizer LoadBrandingCustomizer ()
...
+		}
+
 		static StartupMetadata GetStartupMetadata (StartupInfo startupInfo, IPlatformTelemetryDetails platformDetails, Dictionary<string, long> timings)
 		{
 			var assetType = StartupAssetType.FromStartupInfo (startupInfo);
+			StartupType startupType = StartupType.Normal;
+
+			if (startupInfo.Restarted && !IdeApp.IsInitialRunAfterUpgrade) {
+				startupType = StartupType.ConfigurationChange; // Assume a restart without upgrading was the result of a config change
+			} else if (IdeApp.IsInitialRun) {
```
Therzok: Should we consider doing:
```
if (IdeApp.IsInitialRunAfterUpgrade) {
    startupType = StartupType.FirstLaunch;
} else if (IdeApp.IsInitialRun) {
    startupType = StartupType.FirstLaunchSetup;
}
```

iainx: No, `FirstLaunch` is only ever sent once, on the initial launch

Therzok: We pay the cost of creating the MEF cache after upgrade, so we need to notify that (otherwise, we'll just have a trend of startup being higher on release and slowly normalizing the the older version.

iainx:

iainx:

PR #6062, opened 2018-09-21
Add Jose Medrano and myself as owners of the build output code

rodrmoya:


CODEOWNERS
```diff
@@ -26,10 +27,11 @@
 /main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles @mkrueger @sevoku
 /main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Execution @slluis
 /main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components @slluis @sevoku
+/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs @netonjm @rodrmoya
```
Therzok: What's the reasoning behind this?

rodrmoya: The build output depends heavily on this file, if a change is done on how the ProgressMonitor is fed, the build output breaks.

rodrmoya: yes

rodrmoya: I'm not sure though how to add tests, as we would need to build something, since ErrorListPad depends on the build engine to raise events to feed the BuildOutput's ProgressMonitor, so are there already BuildEngine tests available that we can add code to for testing ErrorListPad?

Therzok:
> I'm not sure though how to add tests, as we would need to build something, since ErrorListPad depends on the build engine to raise events to feed the BuildOutput's ProgressMonitor, so are there already BuildEngine tests available that we can add code to for testing ErrorListPad?

Abstract out the code that triggers the notifications into its own controller class, which you can instantiate in the error list pad.

Thus, you get rid of the dependency on the pad itself and only have a small component to depend on, which you can then unit test (abstract class API)

rodrmoya:
Well, right, that could be used to test the interaction between the ErrorListPad and the BuildOutput, but I was talking about testing ErrorListPad itself. We not only depend on the build events, but only on ErrorListPad passing the BuildOutput's ProgressMonitor to the builds, and a few other things. So, really a full integration test, I guess, is what we want.

Therzok:
Then the build event source can be abstracted too, and the logic to control both the error pad input and the build output can be tested.

Therzok:
I'm not against integration tests, I'd love having both unit tests and integration tests. If we can split logic into multiple testable components, it'll help us more in the long run.

PR #6064, opened 2018-09-22
Fix VSTS 674288.

KirillOsenkov:
Need to set RootWindow.Visible to true otherwise the editor may not be focused. The symptoms for this problem are: caret is in the Editor and not blinking. Clicking in the Editor doesn't fix it. Only fix is to move focus outside the app and bring the focus back.
> [Fixes VSTS #674288](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/674288)

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
@@ -126,6 +126,10 @@ namespace MonoDevelop.Ide.Gui
 			var memento = PropertyService.Get (workbenchMemento, new Properties ());
 			Counters.Initialization.Trace ("Setting memento");
 			workbench.Memento = memento;
+
+			// Very important: see https://github.com/mono/monodevelop/pull/6064
+			// Otherwise the editor may not be focused on IDE startup and can't be
+			// focused even by clicking with the mouse.
 			RootWindow.Visible = true;
 			Counters.Initialization.Trace ("Setting layout");
 			workbench.CurrentLayout = "Solution";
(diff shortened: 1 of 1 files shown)


KirillOsenkov:
This regressed in https://github.com/mono/monodevelop/pull/5195.

Therzok:
Note: this might be why, sometimes the main app cannot be focused on startup.

mhutch:
@therzok Do you mean the issue with the app sometimes not being focusable on startup if launched with `make run`? That's been around for years.

@kirillosenkov is there a reliable repro? This seems weird, the call should not be necessary and I'd be much happy if we understood why it was needed.

slluis:

slluis:

KirillOsenkov:
@mhutch looks like this is a multi-faceted issue, or maybe several issues actually.

My easy repro on Windows:
1. Start MonoDevelop
2. Click on a recent project
3. When that opens, it should open a recent file (if it doesn't, open a file and go to step 1)
4. When the file opens, the caret is not blinking and the editor is not focused. Focusing the editor with the mouse does not resolve the issue. Switching focus to another app and back does resolve the issue.

The issue Marius mentions might be a different issue.

Also look at the bug http://work.devdiv.io/674288 - I'm not sure if it's the same or different issue.

But my PR certainly fixes the Windows repro steps above.

KirillOsenkov:
I should add that I've spent 7 working days trying to build my own Gtk and Gtk-sharp on Windows, and I failed miserably! I've gotten them to build, but I couldn't get them to run together so I can debug the issue with symbols and such. So I gave up and bisected to see when this regressed, and found this fix.

KirillOsenkov:
@slluis the bug is not present in 7.6 so the patch didn't apply. We only need this in 7.7 and master.

KirillOsenkov:
@slluis let's get this in?

PR #6075, opened 2018-09-25
Fixes VSTS User Story 686778: Displaying Light Bulb & screwdriver

mkrueger:
icons

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/686778

> [Fixes  VSTS #693143](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/693143)

Fixes issue #4818 Double click text selection is often blocked by new tooltip 

Changed the result tooltips to the monodevelop style - it's no longer in the way of the mouse selection.


slluis:
**Some visual issues**:

* [ ] The lightbulb button has too much padding and it is missing the down arrow (see how it looks in VS for reference).
* [ ] The tooltip doesn't align properly with the lightbulb button.
* [ ] The tooltip is showing weird question marks and empty lines:

  captura de pantalla 2018-10-11 a les 0 08 27

* [ ] The menu hides the lightbulb button. It should be shown below it, not over:

  captura de pantalla 2018-10-11 a les 0 09 33

**Weird behavior**:

* [ ] Sometimes a lightbulb with empty tooltip is shown (see screencast below).
* [ ] Sometimes the tooltip seems to flicker and the shadow thickens. It looks like several tooltips are shown one over the other.
* [ ] I don't think the contextual lightbulb should be shown in the below example. In VS the drop down is shown only when hovering on a squiggle.
  captura de pantalla 2018-10-11 a les 0 09 33

* [ ] Some tooltips are never destroyed. I can reproduce it by just moving the mouse after the "}" in the first line:

  captura de pantalla 2018-10-11 a les 0 24 11

* [ ] Lightbulb shown when not necessary. In this case there is no error to fix, there is no squiggle, the contextual lightbulb should not be shown, only the one in the margin.

  captura de pantalla 2018-10-11 a les 0 28 04

* [ ] Lightbulb over a lightbulb:

  captura de pantalla 2018-10-11 a les 0 28 04

* [ ] Tooltip over lightbulb:

  captura de pantalla 2018-10-11 a les 0 28 04

mkrueger:
Point 2. question marks & empty lines is not caused by that branch. I wonder why nobody recognized that before.

slluis:
* [ ] Lightbulb disappers when moving mouse over arrow:

  kapture 2018-10-15 at 12 58 43

slluis:
* [ ] Clicking on the lightbulb or screwdriver on the left is unreliable, it often doesn't show the menu.

  kapture 2018-10-15 at 13 04 24

slluis:
@mkrueger I found a couple of issues, but it works much better!

Therzok:
Can we have folding markers on the left side of the lightbulb and the lightbulb draw over text? Seems like we're losing a lot of horizontal space.

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
```diff
@@ -3336,9 +3336,11 @@ public void HideTooltip (bool checkMouseOver = true)
 				//					if (x >= 0 && y >= 0 && x < w && y < h)
 				//						return;
 				//				}
-				currentTooltipProvider.DestroyTooltipWindow (tipWindow);
-				tipWindow = null;
-				tipItem = null;
+				var result = currentTooltipProvider.DestroyTooltipWindow (tipWindow);
+				if (result) {
```
slluis: This is a dangerous change. The tooltip provider shouldn't be able to always decide wether the tooltip can be destroyed or not. There are cases in which the tooltip has to be destroyed no matter what (for example, when text changes, when showing a new tooltip, or when disposing the document). A safer solution would be to add a new method `CanDestroyTooltip()` to `TooltiProvider`. This method would be called to check if the tooltip can be closed after an event such as mouse move, and `DestroyTooltipWindow()` would not be called if it returns false. In other cases (the ones I enumerated above and others), TextArea would end calling `DestroyTooltipWindow()` directly and forcing the destroy.

mkrueger: CanDestroyTooltip() does the same basically.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TooltipProvider.cs
```diff
@@ -197,7 +204,7 @@ protected virtual Xwt.Point CalculateWindowLocation (TextEditor editor, TooltipI
 			return new Xwt.Point (x, y);
 		}
 
-		public virtual bool DestroyTooltipWindow (Window tipWindow)
+		public virtual bool TryCloseTooltipWindow (Window tipWindow, TooltipCloseReason reason)
```
slluis: Please don't break the public api. `TryCloseTooltipWindow` can just call the old `DestroyTooltipWindow`.

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
slluis: You did not understand my proposal. For cases in which destroying the window can be optional (mouse move, etc), TextArea would call CanDestroyTooltip() to determine if the provider allows closing the window. For cases in which closing the tooltip is mandatory (text changes, new tooltip shown), TextArea would call DestroyTooltipWindow directly, without checking CanDestroyTooltip.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TooltipProvider.cs
mkrueger: It's no API break because the DestroyTooltipWindow method got introduced with this PR so it's free to change it.

slluis: oh, true

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/Gui/ResultTooltipProvider.FloatingQuickFixIconWidget.cs
```diff
@@ -0,0 +1,139 @@
...
+using System.Windows.Input;
+
+namespace MonoDevelop.AnalysisCore.Gui
+{
+	partial class ResultTooltipProvider
+	{
+		partial class FloatingQuickFixIconWidget : Gtk.Window
+		{
+			readonly CodeActionEditorExtension ext;
+			private readonly LanguageItemWindow window;
```
Therzok:
```suggestion
		readonly LanguageItemWindow window;
```

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/Gui/ResultTooltipProvider.FloatingQuickFixIconWidget.cs
```diff
@@ -0,0 +1,139 @@
...
+			readonly SourceEditorView sourceEditorView;
+			readonly CodeActionContainer fixes;
+			readonly Cairo.Point point;
+			private uint destroyTimeout;
```
Therzok:
```suggestion
			uint destroyTimeout;
```

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/Gui/ResultTooltipProvider.cs
```diff
@@ -97,13 +108,56 @@ public override Window CreateTooltipWindow (TextEditor editor, DocumentContext c
...
+			} catch (AggregateException ae) {
+				ae.Flatten ().Handle (aex => aex is OperationCanceledException);
+			} catch (OperationCanceledException) {
+			} catch (TargetInvocationException ex) {
+				if (!(ex.InnerException is OperationCanceledException))
+					throw;
+			}
+			if (tag == null)
+				return null;
+			Console.WriteLine (floatingWidgetShown);
```
Therzok:
```suggestion
```

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/Gui/ResultTooltipProvider.cs
```diff
@@ -97,13 +108,56 @@ public override Window CreateTooltipWindow (TextEditor editor, DocumentContext c
...
+			var tooltipInfo = new TaggedTooltipInformation<CodeActions.CodeActionContainer> {
+				SignatureMarkup = StringBuilderCache.ReturnAndFree (sb),
+				Tag = tag
+			};
+			return new TooltipItem (tooltipInfo, markerOffset, markerEndOffset - markerOffset);
+		}
+
+		public override Window CreateTooltipWindow (TextEditor editor, DocumentContext ctx, TooltipItem item, int offset, Xwt.ModifierKeys modifierState)
+		{
+			var result = item.Item as TooltipInformation;
```
Therzok: Either null check or hard cast here.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/Gui/ResultTooltipProvider.cs
```diff
@@ -115,6 +169,91 @@ public override void GetRequiredPosition (TextEditor editor, Window tipWindow, o
...
+
+		protected override Xwt.Point CalculateWindowLocation (TextEditor editor, TooltipItem item, Xwt.WindowFrame xwtWindow, int mouseX, int mouseY, Xwt.Point origin)
+		{
+			int w;
+			double xalign;
+			GetRequiredPosition (editor, xwtWindow, out w, out xalign);
+			w += 10;
+			var allocation = GetAllocation (editor);
+
+			var info = item.Item as TaggedTooltipInformation<CodeActions.CodeActionContainer>;
```
Therzok: nullcheck or hard cast here too.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.AnalysisCore/Gui/ResultTooltipProvider.cs
```diff
@@ -115,6 +169,91 @@ public override void GetRequiredPosition (TextEditor editor, Window tipWindow, o
...
+			if (y < geometry.Top)
+				y = geometry.Top;
+
+			return new Xwt.Point (x, y);
+		}
+
+		public override void ShowTooltipWindow (TextEditor editor, Components.Window tipWindow, TooltipItem item, Xwt.ModifierKeys modifierState, int mouseX, int mouseY)
+		{
+			base.ShowTooltipWindow (editor, tipWindow, item, modifierState, mouseX, mouseY);
+			var info = item.Item as TaggedTooltipInformation<CodeActions.CodeActionContainer>;
```
Therzok: null check or hard cast

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.CodeActions/CodeActionEditorExtension.cs
```diff
@@ -209,6 +205,40 @@ async void PopupQuickFixMenu (Gdk.EventButton evt, Action<CodeFixMenu> menuActio
 			}
 		}
 
+		internal void PopupQuickFixMenu (Gdk.EventButton evt, CodeActionContainer fixes,  Action<CodeFixMenu> menuAction, Xwt.Point? point = null)
+		{
+			using (Refactoring.Counters.FixesMenu.BeginTiming ("Show quick fixes menu")) {
```
Therzok: This already had computed quick fixes. We should differentiate between the two kind of counters.

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.CodeActions/CodeActionEditorExtension.cs
```diff
@@ -209,6 +205,40 @@ async void PopupQuickFixMenu (Gdk.EventButton evt, Action<CodeFixMenu> menuActio
 			}
 		}
 
+		internal void PopupQuickFixMenu (Gdk.EventButton evt, CodeActionContainer fixes,  Action<CodeFixMenu> menuAction, Xwt.Point? point = null)
```
Therzok: Also, this method is more or less copy pasted from above. Maybe we should reuse bits.

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
mkrueger: I moved that responsibility down to the tooltip provider.

slluis:
* [ ] Tooltip still not hidden in some cases:

kapture 2018-10-18 at 12 59 30

slluis:
* [ ] Margin icon cropped, and the menu is shown over the icon instead of below it:

kapture 2018-10-18 at 13 01 36

mkrueger:
Backport: https://github.com/mono/monodevelop/pull/6389

PR #6077, opened 2018-09-25
[Core] Evaluate multi-valued MSBuild search path

mrward:
MSBuild, that is included with Mono 5.16, has an MSBuild.dll.config
file that contains search path values that are multi-valued and
also need evaluating.

```
<searchPaths os="osx">
  <property name="MSBuildExtensionsPath" value="$(MSBuildExtensionsPathFallbackPathsOverride);/Library/Frameworks/Mono.framework/External/xbuild/" />
  <property name="MSBuildExtensionsPath32" value="$(MSBuildExtensionsPathFallbackPathsOverride);/Library/Frameworks/Mono.framework/External/xbuild/" />
  <property name="MSBuildExtensionsPath64" value="$(MSBuildExtensionsPathFallbackPathsOverride);/Library/Frameworks/Mono.framework/External/xbuild/" />
  <property name="VSToolsPath" value="/Library/Frameworks/Mono.framework/External/xbuild/Microsoft/VisualStudio/v$(VisualStudioVersion)" />
</searchPaths>
```

The property/values are used when looking for MSBuild imports. The
value was being used as the full path without being evaluated and
without being split into two paths. This resulted in imports that
are in the External/xbuild directory, such as Xamarin.iOS, to not
be found.

Fixes VSTS #682034 - "Tabbed Xamarin Forms App" fails to build due to
error "Your project.json doesn't have a runtimes section. You should
add '"runtimes": { "win": { } }' to your project.json and then re-run
NuGet restore."

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
@@ -306,8 +306,11 @@ namespace MonoDevelop.Projects.MSBuild
 							var pathOs = (string)searchPaths.Attribute ("os")?.Value;
 							if (!string.IsNullOrEmpty (pathOs) && pathOs != os)
 								continue;
-							foreach (var property in searchPaths.Elements ("property"))
-								list.Add (new ImportSearchPathExtensionNode { Property = property.Attribute ("name").Value, Path = property.Attribute ("value").Value });
+							foreach (var property in searchPaths.Elements ("property")) {
+								string propertyValue = property.Attribute ("value").Value ?? string.Empty;
+								foreach (var path in propertyValue.Split (';'))
+									list.Add (new ImportSearchPathExtensionNode { Property = property.Attribute ("name").Value, Path = path });
+							}
 						}
 					}
 				}
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
@@ -1120,7 +1120,10 @@ namespace MonoDevelop.Projects.MSBuild
 					if (import.Project.IndexOf (prop.MSBuildProperty, propertyStart, StringComparison.OrdinalIgnoreCase) == -1)
 						continue;
 
-					files = GetImportFiles (project, context, import, prop.Property, prop.Path, out resolvedSdksPath, out keepSearching);
+					string evaluatedPath = context.Evaluate (prop.Path);
+					if (string.IsNullOrEmpty (evaluatedPath))
+						continue; // Skip undefined property path.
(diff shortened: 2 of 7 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectService.cs
```diff
@@ -306,8 +306,11 @@ static List<ImportSearchPathExtensionNode> LoadDefaultProjectImportSearchPaths (
 							var pathOs = (string)searchPaths.Attribute ("os")?.Value;
 							if (!string.IsNullOrEmpty (pathOs) && pathOs != os)
 								continue;
-							foreach (var property in searchPaths.Elements ("property"))
-								list.Add (new ImportSearchPathExtensionNode { Property = property.Attribute ("name").Value, Path = property.Attribute ("value").Value });
+							foreach (var property in searchPaths.Elements ("property")) {
+								string propertyValue = property.Attribute ("value").Value ?? string.Empty;
+								foreach (var path in propertyValue.Split (';'))
```
Therzok: I'd use [Path.PathSeparator](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.pathseparator?redirectedfrom=MSDN&view=netframework-4.7.2)

Therzok: That way, non-Mac would work too.

mrward: Path.PathSeparator seems to be ':' on the Mac which breaks the test.

mrward: Looks like ';' being used everywhere is OK. It is an MSBuild separator not a platform specific path separator: 
https://github.com/mono/msbuild/commit/c4e70a612bbe5454e7ddc6189b9ea8dce01d17c8

Therzok: Yeah, that's the windows separator. It's going to be painful for Mac users which might have `;` in their file/directory names, since it's a valid unix file character.

Therzok:
```
➜  md ls ';'
;
```

mrward: MSBuild on Mono does not seem to handle ';' in the path in the MSBuild.dll.config. There is a comment there about not handling ';' as part of the path. Would end users be modifying the MSBuild.dll.config file typically? /cc @radical

radical:
This change to support multiple paths in the config file will be reverted as it was not accepted upstream and causes potential performance issues. The PR to merge the changes to master hasn't been merged yet, but this change will be removed then.

slluis:
So this PR is not necessary?

mrward:
Correct this PR is not required. The change seems to be reverted in mono/msbuild [2018-04](https://github.com/mono/msbuild/commit/d885d0a2c25026a0a6d637b23350de47a91b473d) but not master. Closing this.

PR #6078, opened 2018-09-25
[Core] Fix project in solution folder not reloaded

### LINKED ISSUE #6013: Modifying .csproj file that is inside SolutionFolder doesn't reload project
I suspect that this is regression caused by https://github.com/mono/monodevelop/pull/5040 more specifically https://github.com/mono/monodevelop/commit/8324211
Problem is that new code is using `solution.OnItemReloadRequiered` event which is not triggered for projects inside SolutionFolder, code causing this is at https://github.com/mono/monodevelop/blob/429ecb52e84709526e173d72b25d663f98db562f/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionFolder.cs#L909

Maybe problem is that `ItemReloadRequired` event is never used. And should be "Chained" until it's triggered on top SolutionFolder...

mrward:
When a project inside a solution folder was modified outside the IDE
it was not reloaded. The Solution's ItemReloadRequired event was not
fired for projects inside a solution folder so the workspace was not
reloading them.

Fixes VSTS #669787 - Switching branch outside of IDE does not trigger
project reload
Fixes #6013 - Modifying .csproj file that is inside SolutionFolder
doesn't reload project

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionFolder.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionFolder.cs
@@ -906,7 +906,7 @@ namespace MonoDevelop.Projects
 		
 		void OnItemReloadRequired (SolutionItemEventArgs e)
 		{
-			if (ParentFolder == null && ParentSolution != null)
+			if (ParentSolution != null)
 				ParentSolution.OnItemReloadRequired (e);
 			ItemReloadRequired?.Invoke (this, e);
 		}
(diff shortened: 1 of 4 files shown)


Therzok (approved):

mrward:

mrward:

mhutch:
The pattern `if (ParentFolder == null && ParentSolution != null)` seems pretty pervasive in this file. Is there any particular reason the change is needed only for this particular case or are other things broken too?

/cc @slluis

mrward:
The other events I think use the event handlers and raise them to their parents so I think they still work. https://github.com/mono/monodevelop/blob/c82f126b5cd5c3cbab4289dc42d62b3c760aa4ea/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SolutionFolder.cs#L272

We could do that for this particular case. Looks like I broke the pattern with this change.

PR #6079, opened 2018-09-25
Bump master version

MSylvia:


slluis:
We are not going to bump the version in master yet.

PR #6086, opened 2018-09-26
[Core] Fix css file added externally not shown in Solution window

mrward:
Adding a .css file, outside the IDE, into a wwwroot subdirectory
would not display the file in the Solution window. Reloading the
solution would show the file in the Solution window. The problem was
that the file glob, for files inside the wwwroot directory that
matches .css files, contained MSBuild metadata "CopyToPublishDirectory"
and was being ignored. Only file globs without any extra metadata
were supported.

    <Content Include="wwwroot\**" CopyToPublishDirectory="PreserveNewest" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />

If the project has opted in to use default metadata for new files
the file glob is no longer ignored if it has extra MSBuild metadata.

Fixes VSTS #660874 - Solution Pad not refreshing content when adding
files outside of the IDE

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -4446,7 +4446,10 @@ namespace MonoDevelop.Projects
 				return;
 			}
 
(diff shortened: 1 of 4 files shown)


slluis (approved):

mrward:

Therzok (approved):

PR #6088, opened 2018-09-26
FOR TESTING MERP ONLY - DO NOT MERGE

iainx:
Make unhandled exceptions inside threads fatal. This should cause merp to start


iainx:
Ok, with the correct version of Mono installed, this build will trigger Merp correctly

PR #6110, opened 2018-09-27
[Core] Fix file not reloaded when changed externally

mrward:
A file changed and saved multiple times in TextEdit.app on the Mac
would not be reloaded in the IDE. On the Mac the TextEdit.app renames
a file from a /.DocumentRevisions-V100/ subdirectory. Since this is
outside the monitored directories monitored by the solution the native
file watcher converts this rename event into a created event. To
support this the created event is now used to trigger the
FileService.FileChanged event so the file is reloaded.

Fixes VSTS #692174 - File not reloaded when changed externally


slluis (approved):

mrward:

mrward:

Issue #6014, opened 2018-09-18
[Regression] Unit-tests execution


garuma:
Two regressions appeared in recent masters when working with unit-testing:

- An "Application Output" pad is continuously created but no output is actually present:

<img width="2032" alt="screen shot 2018-09-18 at 9 24 42 am" src="https://user-images.githubusercontent.com/104105/45690653-e03b4e80-bb24-11e8-9166-dfed2683f4ff.png">

- An "Outdated Build" dialog is now shown anytime I try to execute a test despite having the `Settings > Build > Build projects before executing unit-tests` checked off.

<img width="616" alt="screen shot 2018-09-18 at 9 26 25 am" src="https://user-images.githubusercontent.com/104105/45690809-33150600-bb25-11e8-9764-f1a76a04ccfa.png">

> VS bug [#691668](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/691668)

Therzok:
`An "Application Output" pad is continuously created but no output is actually present:`

That's caused by https://devdiv.visualstudio.com/DevDiv/_workitems/edit/684712

Therzok:
I haven't seen it before this problem started exhibiting.

mkrueger:
Not related to unit testing - I get a new application output window on run as well.

mkrueger:
Please do not mix several issues into one bug report.

garuma:
Yep as @mrward said, I didn't realize at the time that the outdated build dialog also affected non-unit tests scenarios, I'm happy to re-open a different issue if that helps

xamarin-release-manager:
Fixed in version 7.8.0.633 (master)

Pull Request #6116 merged by: Lluis Sanchez
Author: mono
Commit: 6c0e4d7e0005500a0d9c0e50f40f739ec73a325d (mono/monodevelop)

xamarin-release-manager:
Fixed in version 7.7.0.1480 (release-7.7)

Pull Request #6117 merged by: Lluis Sanchez
Author: mono
Commit: ccb1533a56649c9194e09b5148e02d9e82125427 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.0.0.197 (release-8.0)

Pull Request #6116 merged by: Lluis Sanchez
Author: mono
Commit: 6c0e4d7e0005500a0d9c0e50f40f739ec73a325d (mono/monodevelop)

xamarin-release-manager:
[...]

PR #6116, opened 2018-09-27
Fixes issue #6014 [Regression] Unit-tests execution

mkrueger:
The multiple application output window got introduced by
65c5f17f86fc77f8add6c6090aa526c2cb2a6009 that introduced that an in
progress output window can't be re used. Ending the progress on
dispose as well fixes that issue for unit tests.


mkrueger:

PR #6119, opened 2018-09-27
[Core] Fix pinned projects not being saved

mrward:
If no projects were pinned on the welcome page when the IDE started,
pinning a project, then closing and re-opening the IDE, would result
in the project not being pinned on the welcome page. The problem
was that when there was no existing list of pinned projects stored
in the properties file a new list would be used as a default value.
On saving since the list object matched the default value it was
not saved in the properties file. To avoid this only value types,
strings and instances of the Properties class are supported as
default values.

Fixes VSTS #692381 - Pinned projects are no longer remaining pinned
after a restart


Therzok (approved):

mrward:

mrward:

mrward:
Forms failures are unrelated. These are also happening on the master branch.

Issue #6104, opened 2018-09-27
Invalid URI exception on opening file in root directory


mrward:
VS Mac: 7.7.0.1444
Git revision: 21c6b8485df5c695a53269ab999298dcba289e7c
Build date: 2018-09-26 15:59:10+00
Build branch: release-7.7
Xamarin extensions: 8c052293f9d925b55a5596022f1fd61e02601036

1. Open a project in VS Mac.
2. Open a file (not part of the project) that is in the root directory /

```
Error while getting coding conventions,
System.UriFormatException: Invalid URI: The format of the URI could not be determined.
  at System.Uri.CreateThis (System.String uri, System.Boolean dontEscape, System.UriKind uriKind) [0x0007d] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-04/external/bockbuild/builds/mono-x64/mcs/class/referencesource/System/net/System/UriExt.cs:56 
  at System.Uri..ctor (System.String uriString) [0x00014] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-04/external/bockbuild/builds/mono-x64/mcs/class/referencesource/System/net/System/URI.cs:408 
  at Microsoft.VisualStudio.CodingConventions.EditorConfig.EditorConfigCache.NormalizePath (System.String path) [0x00000] in <33418e5da6fc40388dc1fe4aee49e148>:0 
  at Microsoft.VisualStudio.CodingConventions.EditorConfig.EditorConfigCache+<GetConfigsAsync>d__14.MoveNext () [0x00036] in <33418e5da6fc40388dc1fe4aee49e148>:0 
--- End of stack trace from previous location where exception was thrown ---
  at Microsoft.VisualStudio.CodingConventions.EditorConfig.EditorConfigCache+<GetCodingConventionSnapshotAsync>d__8.MoveNext () [0x0009e] in <33418e5da6fc40388dc1fe4aee49e148>:0 
--- End of stack trace from previous location where exception was thrown ---
  at Microsoft.VisualStudio.CodingConventions.EditorConfig.EditorConfigCodingConventionsProvider+<GetCodingConventionsAsync>d__4.MoveNext () [0x00099] in <33418e5da6fc40388dc1fe4aee49e148>:0 
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #6122, opened 2018-09-28
Fixes issue #6104 Invalid URI exception on opening file in root

mkrueger:
directory The file watcher service change was a follow up bug occurred
on my system.


mkrueger:

mrward (commented):
Change looks good to me.

@Therzok Does the exception handler in the file watcher service look OK to you? If the file watcher decides the root directory is the one to be monitored and the permissions do not allow it then that would mean no directories would be monitored. Not sure we can do much about that though.

Therzok:
We could force pathtree to choose the segments after the root, but definitely nothing we can do about the root directory.

We should open a new issue to not forget about the workaround, and link it to both mono and the editorconfig library issues.

PR #6133, opened 2018-09-28
[Telemetry] Fix null reference when opening sln from command line 7.6

iainx:
Don't null the timer until after the time has been recorded

fixes VSTS #692948


iainx:
@mrward Yes, they're both called on the main thread.

PR #6136, opened 2018-09-28
Fixes VSTS Bug 691614: [Feedback] Sometimes pasting fails in XAML

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/691614


mkrueger:
I couldn't reproduce the original issue but I made the text view lines more stable. That should fix the issue.

PR #6137, opened 2018-09-28
[Editor] Add counter information and various data points

Therzok:
To help pin down possible slowdowns in the editor

Fixes VSTS 675548 - Research typing delays


iainx:
In a future revision of this patch, would it be useful to have individual margin drawing times as well?

Therzok:
It's on my list, might be added in this PR. Making it so TextEditorKeyPressTimings is accessible from the textarea, texteditorextensions and the margins so they can checkpoint on their own so we can further pin down what's taking up time.

Therzok:
@slluis ping?

Therzok:

Therzok:

PR #6138, opened 2018-09-28
[Accessibility] Update the accessibility folds on a 5second timer

iainx:
This should improve typing performance in big files

Fixes vsts #620363


slluis:

PR #6139, opened 2018-09-28
[DotNetCore] Update to correct .NET Core download URLs

rodrmoya:
aka.ms versions seem to not work anymore


rodrmoya:

mrward:
Can we not fix the aka.ms links? aka.ms links allow us to change the values if the destination urls change without having to do a new release.

mrward:
Also fixing the aka.ms links will fix the url for the current stable release.

rodrmoya:
ok, this is not needed then, closing

Issue #6142, opened 2018-09-28
Completion still commits on ENTER when an item is soft-selected


KirillOsenkov:
1. Open a C# file
2. Inside a method body, type `f` followed by `backspace`
3. You end up with completion list open and `for` is soft-selected (not filled, but just a border)
4. Press ENTER

**Expected:**
Completion should be dismissed without committing and ENTER should go through to the buffer

**Actual:** 
Completion is commited and `for` is inserted into the buffer

![softselecteditemcommits](https://user-images.githubusercontent.com/679326/46226837-5f453980-c312-11e8-9674-f9d296aa91a3.gif)

xamarin-release-manager:
Fixed in version 8.0.0.197 (release-8.0)

Pull Request #6148 merged by: Lluis Sanchez
Author: mono
Commit: c4b9856016139c07c6cf3ced83a7addf1dbbdedd (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #6148 merged by: Lluis Sanchez
Author: mono
Commit: c4b9856016139c07c6cf3ced83a7addf1dbbdedd (xamarin/monodevelop)

PR #6148, opened 2018-10-01
Fixes issue #6142 Completion still commits on ENTER when an item is

mkrueger:
soft-selected


mkrueger:

PR #6150, opened 2018-10-01
Bump debugger-libs

akoeplinger:
Brings in https://github.com/mono/debugger-libs/pull/196

/cc @slluis


akoeplinger:
Reopened as https://github.com/mono/monodevelop/pull/6182 so I get test results.

PR #6153, opened 2018-10-01
[Core] Reduce number of file watchers used with .NET Core projects  …

mrward:
A file watcher was being created for each SDK style project. When a
solution contains a lot of SDK style projects (e.g. OrchardCore
contains 126 projects) this can then cause Mono's networking stack
to stop working. NuGet restore, check for updates would fail with
errors similar to:

SecureChannelFailure (Unable to write data to the transport connection:
The socket is not connected.)

Since the FileWatcherService has a maximum number file watchers the
project can use the common file watchers instead of creating its own.
This reduces the number of file watchers. For the OrchardCore solution
this reduces the number of file watchers down from 127 to 1.

There are still other NuGet restore problems with the OrchardCore
solution, but this solves one part.

Fixes VSTS #678422 - Unable to restore packages in OrchardCore app


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -4425,13 +4422,59 @@ void OnFileDeleted (object sender, FileSystemEventArgs e)
...
+				}).Ignore ();
+				return;
+			}
+
+			OnDirectoryMovedOutOfProject (oldDirectory);
+		}
+
+		void OnDirectoryMovedIntoProject (FilePath newDirectory)
+		{
+			foreach (string file in Directory.GetFiles (newDirectory, "*", SearchOption.AllDirectories)) {
```
Therzok: `Directory.EnumerateFiles` here, as `GetFiles` is `EnumerateFiles ().ToArray()`

mrward: Yeah, forgot about EnumerateFiles 👍

main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/DotNetCoreFileWatcherTests.cs
```diff
@@ -72,6 +72,8 @@ async Task<DotNetProject> OpenProject (string solutionName = "DotNetCoreFileWatc
 
 			RunMSBuild ($"/t:Restore /p:RestoreDisableParallel=true \"{solution.FileName}\"");
 
+			await FileWatcherService.Add (solution);
```
Therzok: Where is the filewatcher remove for this? We might have tests hang if we don't remove it.

mrward: Happens when the Solution is disposed. No need for an explicit remove.

Therzok (approved):
Other than enumeration change, LGTM! 👍

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -4351,47 +4351,44 @@ void InitializeFileWatcher ()
...
 
-			watcher = new FSW.FileSystemWatcher (BaseDirectory);
-			watcher.IncludeSubdirectories = true;
-			watcher.NotifyFilter = NotifyFilters.FileName | NotifyFilters.DirectoryName;
-			watcher.Created += OnFileCreated;
-			watcher.Deleted += OnFileDeleted;
-			watcher.Renamed += OnFileRenamed;
-			watcher.Error += OnFileWatcherError;
-			watcher.EnableRaisingEvents = true;
+			// Use FileWatcherService for file created event since this does not run on the UI thread.
```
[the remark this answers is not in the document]
mrward: It can hang the IDE if there are lot of files being created. Seen this problem when doing an npm install with the project open in the IDE.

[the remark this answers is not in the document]
slluis: If we need to be able to handle file change events in the same thread that originated the event, let's add that capability to FileService. Having file change events in two classes is confusing.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/FileService.cs
```diff
@@ -845,6 +845,11 @@ static void OnFileChanged (FileEventArgs args)
 				}
 			}
 		}
+
+		/// <summary>
+		/// File watcher events - these are not fired on the UI thread.
+		/// </summary>
+		public static AsyncEvents AsyncEvents = new AsyncEvents ();
```
slluis: This should be a property.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/FileService.cs
```diff
@@ -966,5 +971,27 @@ public void Thaw ()
 		}
 	}
 
+	public class AsyncEvents
+	{
+		public event EventHandler<FileEventArgs> FileCreated;
+		public event EventHandler<FileEventArgs> FileRemoved;
+		public event EventHandler<FileCopyEventArgs> FileRenamed;
+
+		internal void NotifyFileCreated (FilePath fileName)
```
slluis: We don't need Notify* methods in this class. The idea is that the NotifyFileCreated method in FileService fires the FileService.AsyncEvents.FileCreated event from whatever thread is calling NotifyFileCreated, and it fires FileService.FileCreated in the UI thread.


 events defined in AsyncEvents are raised in the thread that notified then.

slluis (changes requested):

slluis (approved):

slluis:

PR #6154, opened 2018-10-01
[DotNetCore] Support .NETStandard 2.0 when new enough Mono is installed

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/694494

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Tests/MonoDevelop.DotNetCore.Tests/DotNetCoreProjectTemplateWizardTests.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Tests/MonoDevelop.DotNetCore.Tests/DotNetCoreProjectTemplateWizardTests.cs
@@ -121,7 +121,7 @@ namespace MonoDevelop.DotNetCore.Tests
 			CreateWizard ();
 			AddSupportedParameters ("NetStandard");
 			DotNetCoreRuntimesInstalled (new string[0]);
-			MonoRuntimeInfoExtensions.CurrentRuntimeVersion = new Version ("5.1.99");
+			MonoRuntimeInfoExtensions.CurrentRuntimeVersion = new Version ("5.4.0");
 
 			int pages = wizard.TotalPages;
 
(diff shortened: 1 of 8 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
```diff
@@ -116,8 +116,8 @@ internal static bool IsSupported (TargetFramework framework)
 			if (versions.Any (sdkVersion => IsSupported (projectFramework, projectFrameworkVersion, sdkVersion)))
 				return true;
 
-			// .NET Core 1.x is supported by the MSBuild .NET Core SDKs if they are installed with Mono.
-			if (projectFrameworkVersion.Major == 1)
+			// .NET Core <= 2.1 is supported by the MSBuild .NET Core SDKs if they are installed with Mono.
+			if (projectFrameworkVersion <= Version.Parse ("2.1"))
```
Therzok: Define a static readonly field for this, `new Version (2.1)`, no need to parse it from string every time.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/MonoRuntimeInfoExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+using System;
+using MonoDevelop.Core.Assemblies;
+
+namespace MonoDevelop.DotNetCore
+{
+	static class MonoRuntimeInfoExtensions
+	{
+		public static bool SupportsNetStandard20 (this MonoRuntimeInfo monoRuntime)
+		{
+			return monoRuntime.RuntimeVersion >= Version.Parse ("5.2.0");
```
Therzok: Same as above, no need to parse from string, `new Version (5, 2, 0)`, possibly cached.

mrward:
Can we add some new tests? There should be existing tests for DotNetCoreSdk, DotNetCoreProjectSupportedTargetFrameworks, and for DotNetCoreSdkInstalledCondition. Not sure how you want to do tests for the MonoRuntimeInfo monoRuntime part.

rodrmoya:
Indeed, this broke several tests

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Tests/MonoDevelop.DotNetCore.Tests/DotNetCoreProjectTemplateWizardTests.cs
```diff
@@ -120,6 +121,7 @@ public void NetStandard_NoRuntimesInstalled ()
 			CreateWizard ();
 			AddSupportedParameters ("NetStandard");
 			DotNetCoreRuntimesInstalled (new string[0]);
+			MonoRuntimeInfoExtensions.CurrentRuntimeVersion = new Version ("5.1.99");
```
mrward: Should we add a test to check .NET Standard 2.0 is available if Mono is at least 5.2? Or maybe change this to test just Mono 5.2 or later, so .NET Standard 2.0 is available, since that is what we will be using in the New Project dialog.

We have tests for the supported frameworks code so we could get just test what happens with a recent Mono in the New Project dialog which is probably more useful for us than checking an old Mono.

rodrmoya: ok, will change it to test Mono support, I guess that's better indeed

rodrmoya:

PR #6158, opened 2018-10-02
[Ide] Improve Pasting into Solution pad

sevoku:
When pasting data into the solution pad, it might happen that the currently selected node can not handle the data. This is for example the case if a file is selected, the user would expect the data to be pasted into the same folder.

This patch adds some logic to try pasing into the parent node, if the currently selected node can not handle it.

Fixes VSTS [#666943](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/666943)

NOTE: this depends on #6169


sevoku:

sevoku:
@slluis isn't it a kind of an opt-in now? If the node doesn't want to handle the data, the parent has the ability to decide in https://github.com/mono/monodevelop/pull/6158/files#diff-50d02569d1ec7f037ba42611f2066af6R1290 by checking the node and the data in `CanDropMultipleNodes`

mkrueger:
Rebased & added opt-in to master-vnext.

PR #6164, opened 2018-10-02
[Telemetry] Don't reset the startup timer

iainx:
Use a separate timer so that the total duration is reported correctly.

Fixes VSTS #695000.


iainx:

PR #6166, opened 2018-10-02
[NuGet] Fix Update menu enabled when project has no PackageReferences

mrward:
The Update menu was enabled if the project used PackageReferences but
had none in the project. Without any PackageReferences in the
project there is nothing to update. The check to see if the Update
menu is enabled has been changed to make sure the project has
PackageReferences in the project file, not just imported
PackageReferences. Also extended this logic to be used by the Update
NuGet Packages menu which is used to update packages for the solution.

Fixes VSTS #656994 - Update and restore button/menu should be disabled
for SDK NuGet for .NET core projects


mrward:

PR #6169, opened 2018-10-02
[F#] Fix CanDropNode so that it only works within F# projects

nosami:
Also check that drag node is not the same as drop node.

Fixes VSTS #694596


nosami:

PR #6175, opened 2018-10-03
Guard against project parent solution being null.

nosami:
Re: VSTS #694965


slluis:
@nosami can you resolve the conflicts?

nosami:
@slluis done

PR #6178, opened 2018-10-03
[Telemetry] Classify Build/Run events by project type

iainx:
Includes extra metadata on the buildanddeploy event:
 - ProjectID
 - ProjectType
 - ProjectFlavor
 - BuildTime

BuildTime is the length of time the build process took.

Fixes VSTS #675546

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimerCounter.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimerCounter.cs
@@ -240,7 +240,11 @@ namespace MonoDevelop.Core.Instrumentation
 
 		public CounterMetadata (CounterMetadata original)
 		{
-			this.properties = new Dictionary<string, object> (original.properties);
+			if (original != null) {
+				this.properties = new Dictionary<string, object> (original.properties);
+			} else {
+				this.properties = new Dictionary<string, object> ();
+			}
 		}
 
 		public void AddProperties (CounterMetadata original)
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimerCounter.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimerCounter.cs
@@ -235,12 +235,16 @@ namespace MonoDevelop.Core.Instrumentation
 
 		public CounterMetadata (IDictionary<string, object> properties)
 		{
-			this.properties = properties;
+			if (properties == null) {
+				this.properties = new Dictionary<string, object> ();
+			} else {
+				this.properties = properties;
+			}
 		}
 
 		public CounterMetadata (CounterMetadata original)
 		{
-			if (original != null) {
+			if (original != null && original.properties != null) {
 				this.properties = new Dictionary<string, object> (original.properties);
 			} else {
 				this.properties = new Dictionary<string, object> ();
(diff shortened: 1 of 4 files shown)


iainx:

iainx:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
```diff
@@ -1066,7 +1066,14 @@ public AsyncOperation Execute (IBuildTarget entry, ExecutionContext context, Con
...
 		{
-			var metadata = new CounterMetadata ();
+			var metadata = new BuildAndDeployMetadata ();
+
+			if (entry is Solution solution) {
+				var startupProject = solution.StartupItem;
+				metadata.ProjectID = startupProject.ItemId;
+				metadata.ProjectType = startupProject.TypeGuid;
+				metadata.ProjectFlavor = startupProject.TypeGuid;
+			}
```
mrward: Should we special case .NET Core projects here? I think Manish asked for this in the past.

C# sdk style projects use a C# project type guid so it does not distinguish them from classic C# projects. In md-addins we use `{9A19103F-16F7-4668-BE54-9A1E7A4F7556}` as a project kind guid property (not project type guid) for sdk style projects. This is sometimes used by VS 2017 for C# sdk style projects. Not sure if we want to use this instead of the existing type guid or perhaps add an extra property.

Sdk style projects can be determined by the `Project.MSBuildProject.GetReferencedSDKs ().Any ()`.

Also not sure ProjectFlavor is correct here. With other telemetry we are using the first part of FlavorGuid if it is available, otherwise falling back to the TypeGuid. I think that is what VS 2017 is doing when I looked at its telemetry but needs confirming.

`metadata.ProjectFlavor = FlavorGuids.FirstOrDefault () ?? TypeGuid;`

slluis:
Ugh, I commented on the wrong PR. Please see comments in https://github.com/mono/monodevelop/pull/6179.

iainx:
I had to move some of the metadata so the non-build related properties weren't just filled in at build time.

iainx:
And I think that fixes @mrward's concerns as well

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
```diff
@@ -1065,8 +1065,23 @@ public AsyncOperation Execute (IBuildTarget entry, ExecutionContext context, Con
...
 
 		async Task ExecuteAsync (IBuildTarget entry, ExecutionContext context, CancellationTokenSource cs, ConfigurationSelector configuration, RunConfiguration runConfiguration, bool buildBeforeExecuting)
-		{
-			var metadata = new CounterMetadata ();
+		{
+			ProjectEventMetadata eventMetadata = null;
+
+			if (entry is Solution solution) {
+				if (runConfiguration == null || runConfiguration is SingleItemSolutionRunConfiguration) {
+					var startupProject = solution.StartupItem;
```
slluis: You have to get the startup item from `((SingleItemSolutionRunConfiguration)runConfiguration).Item`, which may be different from the default `solution.StartupItem`.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimerCounter.cs
```diff
@@ -240,7 +240,11 @@ public CounterMetadata (IDictionary<string, object> properties)
 
 		public CounterMetadata (CounterMetadata original)
 		{
-			this.properties = new Dictionary<string, object> (original.properties);
+			if (original != null) {
```
Therzok: If we do this check, we should probably also check that `properties` is not null in the above constructor.

iainx:

iainx:

slluis (approved):

PR #6182, opened 2018-10-03
Bump debugger-libs

akoeplinger:
Brings in https://github.com/mono/debugger-libs/pull/196


akoeplinger:
Thanks! Please let me know if this needs to be backported to some branch that will be used with Mono 2018-08.

PR #6184, opened 2018-10-04
Update BinaryCompatBaseline.txt for Mono 2018-08.

KirillOsenkov:
Three new APIs were added to System.dll so we no longer expect them to be absent.

This is a companion PR to https://github.com/xamarin/md-addins/pull/3451. Ideally bump md-addins as part of this PR so that the whole transaction is atomic.


marek-safar:
Why do you guys do that? These APIs are private from assembly we'll never load on Mono.

KirillOsenkov:
It’s an automated tool to help us catch binary inconsistencies between VSMac and its dependencies, such as designer etc. Very helpful. The fact that it also looks in the framework assemblies is just a side effect. We could have an exclusion list I guess, but having a single baseline file is just easier.

akoeplinger:
@KirillOsenkov so it's expected that it looks at _private_ APIs as opposed to just public ones?

KirillOsenkov:
It makes sure all APIs that AssemblyB references from AssemblyA are actually present, via Cecil Resolve(). If Cecil says it expects a type or member in AssemblyA, we’re going to check.
We used to have MissingMethodExceptions, ReflectionTypeLoadExceptions etc.

marek-safar:
I'd probably exclude framework assemblies because in several cases they won't be actually used even if present in .exe directory

akoeplinger:
can we get this in? It's blocking the Mono 2018-08 integration :)

KirillOsenkov:
@slluis

PR #6185, opened 2018-10-04
Display project subcategories as children of categories

nosami:
So that Voiceover announces the relation between categories and
subcategories.

Fixes VSTS #648743


slluis:

PR #6199, opened 2018-10-04
[Core] Fix error adding files when project item uses Windows path

mrward:
When a project had a Reference item that used a Windows file path
the file watcher would cause an exception when it tried to determine
the root directories to monitor. This was because the FilePath's
ParentDirectory is empty when a Windows path is used on a non-Windows
operating system. This is now handled.

    <Reference Include="System.Net.Http, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
      <SpecificVersion>False</SpecificVersion>
      <HintPath>C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\Microsoft\Microsoft.NET.Build.Extensions\net461\ref\System.Net.Http.dll</HintPath>
    </Reference>

Fixes VSTS #697655 - Can't create new files for a solution


slluis (approved):

slluis:

slluis:

PR #6201, opened 2018-10-04
Prevent document switcher when modal is present

nosami:
Fixes VSTS #648193


nosami:
Closing as can't repro VSTS 648193 without the fix.

PR #6206, opened 2018-10-04
[RFC] Host CPU load counters

Therzok:
This PR has a few changes:
* Keeps track of the number of physical and logical CPU cores
* We now have a way on the mac host to query the current CPU load information
* Expose the API into IPlatformTelemetryDetails so it can be queried.

### DIFF
--- a/main/src/addins/MacPlatform/MacTelemetryDetails.cs
+++ b/main/src/addins/MacPlatform/MacTelemetryDetails.cs
@@ -39,7 +39,6 @@ namespace MacPlatform
 	{
 		int family;
 		int coreCount;
-		int threadCount;
 		long freq;
 		string arch;
 		ulong size;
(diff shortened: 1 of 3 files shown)


main/src/addins/MacPlatform/MacTelemetryDetails.cs
```diff
@@ -106,7 +110,9 @@ internal static MacTelemetryDetails CreateTelemetryDetails ()
 
 		public PlatformHardDriveMediaType HardDriveOsMediaType => osType;
 
+		public int CpuCoresCount => coreCount;
 
+		public int CpuThreadCount => threadCount;
```
iainx: `CpuCount` is the same as `CpuThreadCount`

slluis:
@Therzok can you please add a user story for this?

PR #5540, opened 2018-07-30
[A11y] Don't expand the sidebar items if using full keyboard control

iainx:
Fixes VSTS #646431

PR #6207, opened 2018-10-05
[a11y] Don't expand preference categories automatically

sevoku:
This is an alternative and more radical fix for #5540 and disables automatic expansion completely, Instead of disabling it just when accessibility tools are enabled.

It improves the usability when navigating with the keyboard, however, due to the underlying design of the dialog, it's visually slightly confusing if a category has no own panel assigned to it. In this case, the focus remains at the category node, but the selection must move to the first child in order to show the next panel.

preferences no autoexpand

TBH I'm not sure which behaviour is better. @alanjclark what do you think?

> Fixes VSTS [#646431](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/646431)

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/OptionsDialog.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/OptionsDialog.cs
@@ -598,11 +598,6 @@ namespace MonoDevelop.Ide.Gui.Dialogs
 					((Notebook)c).Page = 0;
 			}
 
-			if (!DesktopService.AccessibilityInUse) {
-				// Don't automatically expand trees if using accessibility
-				// as it can be confusing with screen readers
-				tree.ExpandToPath (store.GetPath (page.Iter));
-			}
 			tree.Selection.SelectIter (page.Iter);
 		}
 		


iainx:
TBH I think it's better to leave the behaviour as is unless using keyboard accessibility. The whole tree thing is weird anyway, and isn't helped with the Projects header that has a panel.

How about we make it so that when opening the dialog, all the top-level items are expanded, sub level items are only expanded if the user clicks on the expand arrow, and we remove the Projects panel, and either move that single Gtk related option to its own panel, or just ditch it, because it's a minor Gtk option?

alanjclark:
I think @iainx has a valid suggestion here.  Moving the Projects panel to a "Designer" (or such) child item should offer consistency and less confusing behaviour.

sevoku:
I don't think that opening all categories collapsed is the right approach to solve the navigation issue. Also if we rearrange some panels, we'd need to loop the design team in and in general I don't think that it makes much sense to spend too much time on this dialog. This is not the only issue we have here, there are many more (also accessibility) inside single panels and it's very old-fashioned.
There is a plan to replace this dialog with a new one wich a completely different navigation approach and it will be native. Let's merge #5540 to fix navigation when the special keyboard focus mode is enabled and keep the current behaviour in the regular case for now.

PR #6208, opened 2018-10-05
[Telemetry] Don't use the PropertyService before runtime is initialised

iainx:
Fixes a crash if we try to set the restart property


iainx:

iainx:

PR #6220, opened 2018-10-06
[DotNetCore] Add Templates for VB.Net Projects

Matthew-Collins:
Added Templates for VB.Net Projects including updated Tests.


mrward:
What language does this show in the drop down list in the New Project dialog? The existing VB.NET templates that MonoDevelop has use "VBNet". It looks like the .NET Core sdk VB.NET templates use "VB" which is inconsistent.

Also looks like there is currently no way to override the language in the .addin.xml file for templates that use the new templating engine.

rodrmoya:
Are there ASP.NET templates for VB also? If so, the ASP.NET addin would need to be updated as well, right?

rodrmoya:
hmm, no, there are no ASP.NET templates for VB.NET AFAICS, so ignore my comment

rodrmoya:
And yes, as @mrward said, this shows up templates' language as "VB", not "VBnet" like we do in the rest of the UI for other VB.NET templates

rodrmoya:
@Matthew-Collins I have included your fix in #6222 , so if you want to include new fixes/changes for this, please PR against that branch, ok?

I'll close this one then

Matthew-Collins:
@rodrmoya yeah great, thanks for your help!

Issue #5994, opened 2018-09-17
ArgumentNull: entity in SignatureMarkupCreator.cs:238


KirillOsenkov:
VSMac 7.6.5.20 

```
ERROR [2018-09-15 21:41:50Z]: Got exception while creating markup for :
System.ArgumentNullException: Value cannot be null.
Parameter name: entity
  at MonoDevelop.CSharp.SignatureMarkupCreator.GetMarkup (Microsoft.CodeAnalysis.ISymbol entity) [0x0013f] in /Users/vsts/agent/2.140.0/work/1/s/monodevelop/main/src/addins/CSharpBinding/MonoDevelop.CSharp/SignatureMarkupCreator.cs:238 
  at MonoDevelop.SourceEditor.LanguageItemTooltipProvider+<>c__DisplayClass8_0.<CreateTooltipInformation>b__0 () [0x00010] in /Users/vsts/agent/2.140.0/work/1/s/monodevelop/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Tooltips/LanguageItemTooltipProvider.cs:179 

```

xamarin-release-manager:
Fixed in version 7.8.0.733 (master)

Pull Request #6226 merged by: Lluis Sanchez
Author: mono
Commit: b10028718e7ef8a90f0bf8c23c17cc46afb60aac (mono/monodevelop)

xamarin-release-manager:
Fixed in version 7.7.0.1551 (release-7.7)

Pull Request #6227 merged by: Lluis Sanchez
Author: mono
Commit: 7b686f416711a80122212838189daad558898bb9 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #6226 merged by: Lluis Sanchez
Author: mono
Commit: b10028718e7ef8a90f0bf8c23c17cc46afb60aac (xamarin/monodevelop)

PR #6226, opened 2018-10-08
Fixes issue #5994 ArgumentNull: entity in

mkrueger:
SignatureMarkupCreator.cs:238 Shouldn't happen in monodevelop core
since entity is checked in c# ambience (the only point where this
method is called in monodevelop).


mkrueger:

Issue #6061, opened 2018-09-21
Add using writes the wrong reference down for a framework assembly


Therzok:
When adding a GAC dll reference, since we use reference assemblies for a given .net api, we need to tell Roslyn which assembly to resolve to.

The code fix would add:
`\Library\Frameworks\Mono.framework\Versions\$version\mono\lib\4.7-api\MyDll.dll`

What we need to do is import this service:
http://source.roslyn.io/#Microsoft.VisualStudio.LanguageServices/Implementation/ProjectSystem/MetadataReferences/VisualStudioFrameworkAssemblyPathResolverFactory.cs,18

> VS bug [#691667](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/691667)

xamarin-release-manager:
Fixed in version 7.8.0.889 (master)

Pull Request #6228 merged by: Marius Ungureanu
Author: mono
Commit: 7137ac5d73799076d66aa59f2f853153f3ed9eb8 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 7.7.0.1663 (release-7.7)

Pull Request #6359 merged by: Lluis Sanchez
Author: mono
Commit: da19e1fbc2b6acbe07cb68426def6b958c4c991f (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.0.0.471 (release-8.0)

Pull Request #6228 merged by: Marius Ungureanu
Author: mono
Commit: 7137ac5d73799076d66aa59f2f853153f3ed9eb8 (mono/monodevelop)

PR #6228, opened 2018-10-08
Fixes issue #6061 Add using writes the wrong reference down for a

mkrueger:
framework assembly

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopFrameworkAssemblyPathResolver.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopFrameworkAssemblyPathResolver.cs
@@ -61,7 +61,7 @@ namespace MonoDevelop.Ide.TypeSystem
 				if (!(workspace.GetMonoProject (projectId) is DotNetProject monoProject))
 					return null;
 
-				string assemblyFile = Runtime.SystemAssemblyService.DefaultAssemblyContext.GetAssemblyLocation (assemblyName, monoProject.TargetFramework);
+				string assemblyFile = monoProject.AssemblyContext.GetAssemblyLocation (assemblyName, monoProject.TargetFramework);
 				if (assemblyFile != null) {
 					//if (string.IsNullOrEmpty(fullyQualifiedName) || CanResolveType(ResolveAssembly (projectId, assemblyName), fullyQualifiedName))
 					return assemblyFile;
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopFrameworkAssemblyPathResolver.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopFrameworkAssemblyPathResolver.cs
@@ -33,64 +33,85 @@ using Microsoft.CodeAnalysis.Host;
 using Microsoft.CodeAnalysis.Host.Mef;
 using MonoDevelop.Core;
 using MonoDevelop.Core.Assemblies;
+using MonoDevelop.Projects;
 
 namespace MonoDevelop.Ide.TypeSystem
 {
-	[ExportWorkspaceService (typeof (IFrameworkAssemblyPathResolver), ServiceLayer.Host), Shared]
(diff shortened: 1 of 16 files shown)


mkrueger:
I'm not sure in which use case this service is used - @Therzok can you provide one ?

The VS.NET implementation loads reflection assemblies and checks if the type exists - even if the name is null. I would really like to avoid that  - looks like a double check. Roslyn has some way of discovering the type -> assembly name therefore I think this check assembly name -> type is not needed.

However if it is I would rather use cecil to look up the name - loading assemblies that way pollutes the app domain and loading may fail due to dependency/framework compatibility issues.

Any comments ?

Therzok:
> I'm not sure in which use case this service is used - @Therzok can you provide one ?

There's 1 scenario that's fixed by this patch, and that's searching reference assemblies without having them referenced in any project. Saw it used in the codeaction, assumed that it was the fix. We currently have that toggled off, sorry for misdiagnosing.

To fix the bug report scenario, we need to normalize the references at MonoDevelop workspace level and change the path when we encounter a GAC assembly:
http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces.MSBuild/MSBuild/ProjectFile/ProjectFile.cs,3a3a092b693fd0e7

My test project to repro the issue:
[testusing.zip](https://github.com/mono/monodevelop/files/2458581/testusing.zip)

Open `testconsaddusing` and trigger quickfix on `MessageQueueInstaller`. If that adds a relative reference to System.Messaging, the test should fail.

I've pushed a minor fixup to this branch, so the (currently) unused scenario is working

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopFrameworkAssemblyPathResolver.cs
```diff
@@ -0,0 +1,117 @@
...
+
+			public string ResolveAssemblyPath (ProjectId projectId, string assemblyName, string fullyQualifiedName = null)
+			{
+				if (workspace == null)
+					return null;
+
+				if (!(workspace.GetMonoProject (projectId) is DotNetProject monoProject))
+					return null;
+
+				string assemblyFile = Runtime.SystemAssemblyService.DefaultAssemblyContext.GetAssemblyLocation (assemblyName, monoProject.TargetFramework);
```
slluis: Shouldn't this just use the assembly context from the project? That is, `monoProject.AssemblyContext`

slluis:
Can we have unit tests for this based on Marius' example?

mkrueger:
We can test the service however I don't know how this example should help - it's just using a quick fix. Better is to test such services directly there is no need for throwing quick fixes in they may change or get removed or something.

Roslyn is nothing we can control.

slluis:
@Therzok can you please verify?

Therzok:
Issue is still not fixed, seeing relative path to `/Library` being written.

It seems like roslyn is not considering the assemblies as being reference assemblies, but like simple assembly references.

Trying to figure out what is causing them not to be seen as framework reference assemblies by Roslyn.

Therzok:
Discussed with roslyn folks: https://gitter.im/dotnet/roslyn?at=5bc4c2f2e65a634336e25bdf

Adding some code in workspace layer to handle this.

Therzok:
Currently working on adding tests. Two things don't work yet in this version:
* net461 project referencing System.Messaging, on adding reference to net471, it doesn't correct it to a package assembly
* gdk-sharp is a special case to handle. not sure how.

Therzok:
Added unit tests.

The one problem remaining is `glib-sharp` handling, as in, things which are in gac but not in a package.

I don't know how to solve that problem, especially without breaking AssemblyContext API or making it incredibly slow (IO + lots of allocations even on fast path).

slluis:
@Therzok is the glib-sharp issue a big problem? In which cases things won't work?

Therzok:
It writes a relative path to `/Library/Frameworks...` for assemblies which are not included in a SystemPackage.

I'm not sure it's worth pursuing a fix for that, I've only seen glib-sharp, gtk# and co in there.

slluis:
Ok then.

slluis:
Let's resolve the conflicts and merge.

Therzok:

PR #6229, opened 2018-10-08
[MacPlatform] Don't use obsolete Xamarin.Mac API

rodrmoya:


rodrmoya:
This is now ready, but keeps failing in F#/Forms tests, which seem unrelated

PR #6233, opened 2018-10-09
[Core] Fix race on saving projects

Therzok:
What happens here is that there's no locking happening at MSBuildPropertyGroup level
that would ensure that no read/write or write/write happens when accessing its property
values

Given that the project model is thread-safe for reading, but not thread-safe for writing,
we can workaround that by using an immutable dictionary, ensuring reading the values
at a given time gives the right snapshot.

Fixes VSTS #660905 - Setting the linker to "Link SDK" and saving the
changes in WatchKitAppExtension fails with error "Index was outside the
bounds of the array"
Fixes VSTS #687075 - “System.IndexOutOfRangeException”
exception on changing proprties in iOS Build Project Options

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildPropertyGroup.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildPropertyGroup.cs
@@ -36,7 +36,7 @@ namespace MonoDevelop.Projects.MSBuild
 {
 	public class MSBuildPropertyGroup: MSBuildElement, IMSBuildPropertySet, IMSBuildEvaluatedPropertyCollection
 	{
-		Dictionary<string,MSBuildProperty> properties = new Dictionary<string, MSBuildProperty> (StringComparer.OrdinalIgnoreCase);
+		ImmutableDictionary<string, MSBuildProperty> properties = ImmutableDictionary<string, MSBuildProperty>.Empty.WithComparers (StringComparer.OrdinalIgnoreCase);
 		internal List<MSBuildProperty> PropertiesAttributeOrder { get; } = new List<MSBuildProperty> ();
 		public MSBuildPropertyGroup ()
 		{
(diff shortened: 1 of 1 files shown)


Therzok:
@manish can we rerun automated uitests?

Therzok:
@slluis finding it pretty hard to crop up a test case that's deterministic.

With the previous version, I could pretty much repro it on my machine (not sure how to on slower machines).

```
using System;
using System.Threading.Tasks;
using NUnit.Framework;

namespace MonoDevelop.Projects.MSBuild
{
	[TestFixture]
	public class MSBuildPropertyGroupTests
	{
		[Test]
		public async Task IsThreadSafeReading ()
		{
			var group = new MSBuildPropertyGroup();
			var crit = new AsyncCriticalSection ();
			const int count = 5000;
			var tasks = new Task [count];

			for (int i = 0; i < count; ++i) {
				int _i = i;
				tasks [i] = Task.Run (async () => {
					if (_i % 2 == 0) {
						using (await crit.EnterAsync ()) {
							group.SetValue ("a", "a" + _i);
						}
					} else {
						group.GetValue ("a", "a" + _i);
					}
				});
			}

			await Task.WhenAll (tasks);
		}
	}
}
```

Therzok:
Funnily enough, after running that, the IDE becomes unresponsive, as a lot of FSW threads are now up.

slluis:
I don't see how this solves the problem. The problem happens when writing, not when reading, so this is a write/write concurrency issue. By using an immutable dictionary we just hide a concurrency case that should not happen.

Therzok:
There's a few problems at hand:

Dictionary enumeration is not thread safe, if any writer is writing. So, even if we have just 1 writer and 12 readers, it'll still throw an exception at some point or loop infinitely. So if we don't want to lock around the Dictionary, we have no choice but to use ImmutableDictionary.

The above stands because _anyone_ can request the global property group that's being written to and they can write to it on a completely different thread than the one writing the projects. This alone is a really hard problem, as none of the stacktraces we've had showed a different code-path.

Also, I just noticed `MSBuildProject.SetShared` is never called?

slluis:
I'm aware that this will fix the write/read concurrency problem, but this doesn't fix the write/write concurrency problem, which is what is reported in the referenced bugs.

Therzok:
cc @slluis 

Validated with UI tests, there is no write/write race from what I can see.

slluis (approved):

Therzok:

Therzok:

PR #6248, opened 2018-10-09
[Editor] Update highlighting code a bit so it takes a fast path on mi…

Therzok:
…smatch faster

It's more efficient to check whether there's a dot in the string rather than comparing strings

Saves around 60ms on UI rendering code out of 3500ms.


slluis:
@mkrueger can you review?

PR #6256, opened 2018-10-09
[WelcomePage] Show it when the app goes idle

Therzok:
Do not trigger showing the welcome page until the app is idle,
so as to not load the welcome page when a solution is passed to the app.

This improves startup time for solutions/documents loaded from finder

Fixes VSTS #652347 - Welcome page should not show when loading a solution from finder

### DIFF
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -600,7 +600,7 @@ namespace MonoDevelop.MacIntegration
 
 				ApplicationEvents.OpenDocuments += delegate (object sender, ApplicationDocumentEventArgs e) {
 					//OpenFiles may pump the mainloop, but can't do that from an AppleEvent, so use a brief timeout
-					GLib.Timeout.Add (10, delegate {
+					GLib.Timeout.Add (0, delegate {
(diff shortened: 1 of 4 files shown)


Therzok:
There is a really small flicker of the workbench until the welcome page shows now. I'm wondering if we should show nothing until the app is idle.

slluis:
We should avoid this flicker. Can you try delaying the show of the window?

Therzok:
I'll look into it.

slluis:
What I see now is that the app icon stops bouncing, stays idle for 2 or 3 seconds and then the window shows up. It should be bouncing until the window is shown.

Therzok:
Hmmm, need to figure out how to eliminate that, it might be the DispatchPendingEvents.

Therzok:
Fixed.

Therzok:
Fixed whitespace damage

slluis:
I don't see that bouncing pause any more. However, the original issue is not fixed. If I open a solution from finder the Welcome Page is still loaded.

Therzok:
Yes, it is, but it is loaded after the project starts loading. So that means we don't delay project load by showing the welcome page, we show it _after_ the project load started.

PR #6267, opened 2018-10-09
Prevent esc key propagation in Visualizer dialog

nosami:
Stop the escape key closing both the Value Visualizer dialog and the
Exception caught dialog

Fixes VSTS #648194

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.Visualizer/ValueVisualizerDialog.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.Visualizer/ValueVisualizerDialog.cs
@@ -29,6 +29,7 @@ using System.Linq;
 using System.Collections.Generic;
 using Mono.Debugging.Client;
 using Gtk;
+using Gdk;
 
 namespace MonoDevelop.Debugger.Viewers
 {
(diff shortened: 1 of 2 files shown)


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.Visualizer/ValueVisualizerDialog.cs
```diff
@@ -78,6 +79,20 @@ public void Show (ObjectValue val)
 			}
 		}
 
+		protected override bool OnKeyPressEvent (EventKey evnt)
+		{
+			if (evnt.Key == Gdk.Key.Escape) {
+				GLib.Timeout.Add (100, delegate {
```
nosami: Not sure why this is needed, but without it, the parent dialog window is closed

slluis: Why this delay? 100ms is quite a lot to react to a key press.

nosami: No reason for picking 100 in particular - any delay is ok here. What would be a good value? 10ms?

nosami: Hmm... I tried 10ms and the fix no longer works - I need to find another way.

nosami:

PR #6278, opened 2018-10-10
Fixes VSTS Bug 699450: [Feedback] Error creating file - NullReference…

mkrueger:
…Exception: Object reference not set to an instance of an object

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/699450


mkrueger:
May not be needed.

PR #6283, opened 2018-10-10
[Core] Add a simple race checker, that detects when lock contention happens on a resource

Therzok:
In non-async code, it can be used to monitor which threads are using a given resource.

It might be useful to debug unwanted races.

A LoggingRaceChecker is added by default.

The 2nd commit is there to be able to diagnose the msbuild property group race.


Therzok:
@manish do we have any results from running UI tests?

Therzok:
@slluis found some read/write issues in logs from manish using this.

Can we merge it so we can reuse it in case we have some concurrency issues in the future?

PR #6284, opened 2018-10-11
Fixes VSTS Bug 681385: Accessibility: VoiceOver: Editor Caret does not

mkrueger:
follow VoiceOver

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/681385

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
@@ -407,7 +407,7 @@ namespace Mono.TextEditor
 			this.textEditorData.SelectionChanged += TextEditorDataSelectionChanged;
 			this.textEditorData.UpdateAdjustmentsRequested += TextEditorDatahandleUpdateAdjustmentsRequested;
 			Document.DocumentUpdated += DocumentUpdatedHandler;
-			
+			Document.EndUndo += delegate { HandleAccessibilityCursorChange (); };
 			this.textEditorData.Options.Changed += OptionsChanged;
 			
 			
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
```diff
@@ -407,7 +407,7 @@ internal void Initialize (MonoTextEditor editor, TextDocument doc, ITextEditorOp
 			this.textEditorData.SelectionChanged += TextEditorDataSelectionChanged;
 			this.textEditorData.UpdateAdjustmentsRequested += TextEditorDatahandleUpdateAdjustmentsRequested;
 			Document.DocumentUpdated += DocumentUpdatedHandler;
-			
+			Document.EndUndo += delegate { HandleAccessibilityCursorChange (); };
```
slluis: Only called on EndUndo? shouldn't also be called when position or selection changes

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
```diff
@@ -571,7 +571,99 @@ void CaretPositionChanged (object sender, DocumentLocationEventArgs args)
...
-		
+		ISegment lastAccessiblitySelection = null;
+		ITextSourceVersion lastAccessiblityVersion;
+		DocumentLocation oldAccessibilityCursorLocation;
+		static readonly string selectedTextMessage = GettextCatalog.GetString (", selected");
+		static readonly string unselectedTextMessage = GettextCatalog.GetString (", unselected");
+
+		void HandleAccessibilityCursorChange ()
+		{
+			void NewLineAnnouncement () { ShowAccessibilityAnnouncement (GettextCatalog.GetString ("New line")); }
```
slluis: This method should do nothing if accessibility or voiceover is disabled.

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
mkrueger: Yes.

I need to test that.  It's called only on position / selection changes. However not on mouse selection yet but on keyboard commands.

main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/TextArea.cs
mkrueger: I need help from iain here atm it does no announcement - never.

Therzok:
Not sure we still want a11y fixes for the old editor.

mkrueger:
Closing - the old editor got replaced so there is no reason to work any longer on that.

PR #6288, opened 2018-10-11
Fixes VSTS Bug 701330: DTS: When adding Unit Test into the

mkrueger:
application, the application will crash with a stack trace that seems
to try to recursively add unit test over 85,000 times

https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems/edit/701330


mkrueger:

PR #6293, opened 2018-10-11
[Mac] Make stub launcher configure mono hybrid suspend

Therzok:
This commit tells mono to report any kind of hang in mono
due to gc hybrid suspend, as opposed to just hanging, as it currently does.

Usable to report bugs to runtime time team about hybrid suspend.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/703564

> [Fixes  VSTS #703564](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/703564)

### DIFF
--- a/main/build/MacOSX/Makefile.am
+++ b/main/build/MacOSX/Makefile.am
@@ -12,6 +12,10 @@ MONOSTUB_DEFINES=-DXM_REGISTRAR
 MONOSTUB_STATIC_LINK=$(EXTERNAL)/Xamarin.Mac.registrar.full.a
 EXTERN_C_XM_REGISTRAR_DEFINE=$(shell ./check-xm-extern.sh)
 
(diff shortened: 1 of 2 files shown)


Therzok:
@slluis Maybe it would be worth having this enabled in dogfood builds too. The problem is that Mono just hangs in case of a hybrid suspend bug, and with this patch, it aborts and reports useful information. Given how unusable VSMac is with the current mono, we should:
* Disable hybrid suspend for the next release
* Enable this in local/dogfood builds

See the issue below as an example output of this env var:
https://github.com/mono/mono/issues/11166

slluis:
Hmm, many test failures.

Therzok:
Rebased. Now that hybrid suspend is disabled, this is a NOP for now, but will be useful when we toggle it back on.

PR #6299, opened 2018-10-11
Allow template selection to work with tree

nosami:
Fixes VSTS #702771

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
@@ -53,7 +53,7 @@ namespace MonoDevelop.Ide.Projects
 
 			// Set up the list store so the test framework can work out the correct columns
 			SemanticModelAttribute modelAttr = new SemanticModelAttribute ("templateCategoriesListStore__Name", "templateCategoriesListStore__Icon", "templateCategoriesListStore__Category");
-			TypeDescriptor.AddAttributes (templateCategoriesListStore, modelAttr);
+			TypeDescriptor.AddAttributes (templateCategoriesTreeStore, modelAttr);
 			modelAttr = new SemanticModelAttribute ("templateListStore__Name", "templateListStore__Icon", "templateListStore__Template");
 			TypeDescriptor.AddAttributes (templatesTreeStore, modelAttr);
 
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -520,45 +521,60 @@ void SelectFirstSubTemplateCategory ()
...
+			do {
+				yield return iter;
+				if (model.IterHasChild (iter)) {
+					if (model.IterChildren (out TreeIter childIter, iter))
+						foreach (var node in WalkTree (model, childIter))
+							yield return node;
+				}
+
+			} while (model.IterNext (ref iter));
 		}
```
mrward: Would it make sense for WalkTree to do the GetIterFirst bit too if TreeIter iter is TreeIter.Zero? It seems that the code is doing GetIterFirst, then WalkTree in a few places.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -534,31 +536,45 @@ void SelectTemplateCategory (TemplateCategory category)
...
+			}
 		}
 
 		void SelectFirstTemplate ()
 		{
-			TreeIter iter = TreeIter.Zero;
-			if (templatesTreeStore.IterNthChild (out iter, 1)) {
-				templatesTreeView.Selection.SelectIter (iter);
+			// first child of first iter
+			if (templatesTreeStore.GetIterFromString (out TreeIter iter, "0:0")) {
```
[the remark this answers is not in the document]
nosami: No... first child of first iter

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
nosami: I don't think so, because this method is recursive.

mrward: I think Walk could handle it even though it is recursive. If the iter passed is TreeIter.Zero then you try to get the first iter for the model. For other cases it would pass a TreeIter that is not TreeIter.Zero.

Just an idea.

nosami: Ok, I'll do that. Might be a little better.

nosami: image

Any idea how to compare with TreeIter.Zero? :)

slluis:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
nosami: done

slluis:

nosami:

PR #6305, opened 2018-10-12
[NUnit] Workaround assembly loading errors

garuma:
This PR adds a workaround to be able to successfully execute the designer unit-tests.

The core of the problem is described in https://github.com/mono/monodevelop/commit/4a01d159b935a4d454d326a6f149b9959b8268ce as in the current logic for `IsSupportAssembly` doesn't take into account PackageReference binaries quite well since it cannot accurately determine with the availble information if such a reference will end up being copied to the output directory or not.

This causes a lot of assemblies to be loaded in the running process mistakenly which in our case cause AssemblyLoad errors to happen (we invoke a bunch of things like MSBuild which are disrupted by those spurious imports).

Note: I added an extra constructor parameter to `RemoteProcessConnection` as an optional parameter which keeps API but not ABI compatibility. Since this is a public class, do we care about this?

### DIFF
--- a/main/src/addins/MonoDevelop.UnitTesting.NUnit/MonoDevelop.UnitTesting.NUnit/ExternalTestRunner.cs
+++ b/main/src/addins/MonoDevelop.UnitTesting.NUnit/MonoDevelop.UnitTesting.NUnit/ExternalTestRunner.cs
@@ -48,7 +48,11 @@ namespace MonoDevelop.UnitTesting.NUnit.External
 		IRemoteEventListener listener;
 		readonly string assemblyDirectory;
 
-		public ExternalTestRunner (string assemblyDirectory = null)
+		public ExternalTestRunner ()
+		{
+		}
+
+		public ExternalTestRunner (string assemblyDirectory)
 		{
 			this.assemblyDirectory = assemblyDirectory;
 		}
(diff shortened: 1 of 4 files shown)


mrward:
I think we should add a second constructor overload for RemoteProcessConnection to avoid breaking binary compatibility. Presumably any addin that is using RemoteProcessConnection will break if the constructor is changed to take a new optional parameter.

main/src/addins/MonoDevelop.UnitTesting.NUnit/MonoDevelop.UnitTesting.NUnit/ExternalTestRunner.cs
```diff
@@ -46,15 +46,17 @@ class ExternalTestRunner: IDisposable
 	{
 		RemoteProcessConnection connection;
 		IRemoteEventListener listener;
+		readonly string assemblyDirectory;
 
-		public ExternalTestRunner ()
+		public ExternalTestRunner (string assemblyDirectory = null)
```
Therzok: Please avoid optional parameters and add a new constructor

garuma:
PR adjusted to add another constructor and keep ABI compat.

Therzok (approved):

slluis (approved):

slluis:
Ok, but we'll need a proper fix. https://devdiv.visualstudio.com/DevDiv/_workitems/edit/706645.

garuma:
Added a comment on the new work item for extra context

garuma:
The unit-tests seem to be timeout-ing in various places (F#, Forms addin) that *I think* are unrelated to the changes here.

Provided that's the case we need someone with write permission to merge.

mrward:
Yeah the timeout does not look to be specific to this pull request. I can see it happening elsewhere on other PRs.

PR #6313, opened 2018-10-15
Fixes VSTS Bug 701972: [Feedback] Mouse Cursor is always an arrow

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/701972

Cursors get disposed on destroy. The TextViewMargin tried to share the
same cursor objects (as only margin which changes the cursor).


slluis:
@mkrueger build failed...

slluis:

PR #6314, opened 2018-10-15
Fixes VSTS Bug 701970: [Feedback] "Show Invisible Characters" setting

mkrueger:
keeps being reset in 7.6

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/701970


Therzok:
Can we add a test that asserts that setting VersionControl's editor options does not change it for the IDE?

slluis:

PR #6331, opened 2018-10-15
[Ide] Properly handle symlinks at workbench level

Therzok:
When working with symlinked on disk files in a project, there's a
few things which were not handled correctly.

When a project references a symlink, we should keep the original
symlink information around so that it can be used when looking up
a for a Project containing a file or a file in general in a project.

Added in an OriginalFile property that is used to recover the original
file name, before resolving a symlink, that can be used by the typesystem
to figure out which project the file belongs to.

Fixes VSTS #700204 - Symlinked files do not have correct define constants

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
@@ -1082,7 +1082,8 @@ namespace MonoDevelop.Ide.Gui
 				if (project == null)
 					project = GetProjectContainingFile (openFileInfo.OriginalFileName);
 				openFileInfo.Project = project;
-			}
+			} else
+				project = openFileInfo.Project;
 			
 			if (openFileInfo.DisplayBinding != null) {
 				binding = viewBinding = openFileInfo.DisplayBinding;
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
@@ -1071,12 +1071,18 @@ namespace MonoDevelop.Ide.Gui
 			
 			IDisplayBinding binding = null;
 			IViewDisplayBinding viewBinding = null;
+			Project project = null;
 			if (openFileInfo.Project == null) {
 				// Set the project if one can be found. The project on the FileOpenInformation
 				// is used to add project metadata to the OpenDocumentTimer counter.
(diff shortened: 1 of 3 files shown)


Therzok:
Seems like this broke something in the forms tests (shared asset projects). Will investigate tomorrow.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
```diff
@@ -1068,12 +1068,18 @@ async Task<bool> RealOpenFile (ProgressMonitor monitor, FileOpenInformation open
...
 			IViewDisplayBinding viewBinding = null;
+			Project project = null;
 			if (openFileInfo.Project == null) {
 				// Set the project if one can be found. The project on the FileOpenInformation
 				// is used to add project metadata to the OpenDocumentTimer counter.
-				openFileInfo.Project = GetProjectContainingFile (fileName);
+				project = GetProjectContainingFile (fileName);
+
+				// In some cases, the file may be a symlinked file. We cannot find the resolved symlink path
+				// in the project, so we should try looking up the original file.
```
slluis: Why we cannot find the resolved symlink in the project? ProjectFile could keep the symlink and the resolved path, and GetProjectContainingFile() could check both when looking for a file.

Therzok: I have the following setup:
```
a.csproj:
* Symlink.cs -> ../Actual.cs
```

At some point, `fileName` is resolved to an actual file, so we'll try to look for `../Actual`. But good idea on the ProjectFile, will take a look into making it that way.

Therzok: I have a question about this comment.

If we make ProjectFile verify that the file it maps to is a symlink, we'd do IO for every file in the project just to validate it's a symlink.

The whole issue can be circumvented (like this PR does), by validating that the project contains either the symlink or the resolved file. Both lookups are correct.

Since ViewContent's work on the resolved symlink, we need to keep the original file information when opening.

What do you suggest doing?

Therzok: So, the main problem is that when editing files, we use the _actual file_ not the symlink. If the csproj contains a reference to the symlink, we'll never map the document (which refs the actual file) to the project containing it.

Therzok:
Taking a look at test failures.

slluis:
@Therzok any idea about the test failures?

slluis:
Looks good.

slluis:

PR #6332, opened 2018-10-15
[Roslyn] Add the option to toggle source analysis on from the quick f…

Therzok:
…ix menu

Fixes VSTS #598630 - Quick fix does not offer to add using statement if Source Analysis is disabled


Therzok:

PR #6366, opened 2018-10-17
Fixes VSTS Bug 705785: [Feedback] Search result tab has temporary

mkrueger:
files

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/705785

The user is right that search in whole solution shouldn't include
generated files. However it's a slight behavior change to old
versions. If users want to search in generated files for some reason
they should switch to 'search in directory' instead.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.cs
@@ -173,8 +173,12 @@ namespace MonoDevelop.Ide.FindInFiles
 							  (project, loop, providers) => {
 								  var conf = project.DefaultConfiguration?.Selector;
 
-								  foreach (ProjectFile file in project.GetSourceFilesAsync (conf).Result.Where (f => filterOptions.NameMatches (f.Name) && File.Exists (f.Name))) {
+								  foreach (ProjectFile file in project.GetSourceFilesAsync (conf).Result) {
 									  if ((file.Flags & ProjectItemFlags.Hidden) == ProjectItemFlags.Hidden)
+										  continue;
+									  if (!filterOptions.IncludeCodeBehind && file.Subtype == Subtype.Designer)
+										  continue;
+									  if (!filterOptions.NameMatches (file.Name))
 										  continue;
 									  if (!IdeServices.DesktopService.GetFileIsText (file.FilePath))
 										  continue;
@@ -229,8 +233,12 @@ namespace MonoDevelop.Ide.FindInFiles
 				monitor.Log.WriteLine (GettextCatalog.GetString ("Looking in project '{0}'", project.Name));
 				var alreadyVisited = new HashSet<string> ();
 				var conf = project.DefaultConfiguration?.Selector;
-				foreach (ProjectFile file in project.GetSourceFilesAsync (conf).Result.Where (f => filterOptions.NameMatches (f.Name) && File.Exists (f.Name))) {
+				foreach (ProjectFile file in project.GetSourceFilesAsync (conf).Result) {
 					if ((file.Flags & ProjectItemFlags.Hidden) == ProjectItemFlags.Hidden)
+						continue;
+					if (!filterOptions.IncludeCodeBehind && file.Subtype == Subtype.Designer)
+						continue;
+					if (!filterOptions.NameMatches (file.Name))
 						continue;
 					if (!IdeServices.DesktopService.GetFileIsText (file.Name))
 						continue;
--- a/main/src/addins/MonoDevelop.Autotools/MakefileData.cs
+++ b/main/src/addins/MonoDevelop.Autotools/MakefileData.cs
@@ -1426,7 +1426,7 @@ namespace MonoDevelop.Autotools
 
 			List<string> files = new List<string> ();
 			foreach (ProjectFile pf in OwnerProject.Files) {
-				if (pf.Subtype != Subtype.Code)
+				if (pf.Subtype != Subtype.Code && pf.Subtype != Subtype.Designer)
 					continue;
 				if (IsFileExcluded (pf.FilePath))
 					continue;
--- a/main/src/addins/MonoDevelop.Autotools/SimpleProjectMakefileHandler.cs
+++ b/main/src/addins/MonoDevelop.Autotools/SimpleProjectMakefileHandler.cs
@@ -152,7 +152,7 @@ namespace MonoDevelop.Autotools
 					{
 						case BuildAction.Compile:
 							
-							if ( projectFile.Subtype != Subtype.Code ) continue;
+							if (projectFile.Subtype != Subtype.Code && projectFile.Subtype != Subtype.Designer) continue;
 							files.AppendFormat ( "\\\n\t{0} ", MakefileData.ToMakefilePath (pfpath));
 							break;
 
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -729,13 +729,19 @@ namespace MonoDevelop.Projects
 			{
 				var analyzerList = new List<FilePath> ();
 				var sourceFilesList = new List<ProjectFile> ();
-
 				foreach (var item in items) {
 					var msbuildPath = MSBuildProjectService.FromMSBuildPath (project.sourceProject.BaseDirectory, item.Include);
 
-					if (item.Name == "Compile")
-						sourceFilesList.Add (new ProjectFile (msbuildPath, item.Name) { Project = project });
-					else if (item.Name == "Analyzer")
+					if (item.Name == "Compile") {
+						var projectFile = new ProjectFile (msbuildPath, item.Name) { Project = project };
+						const string subtype = "SubType";
+						if (item.Metadata.HasProperty (subtype)) {
+							var property = item.Metadata.GetProperty (subtype);
+							if (property.Value == "Designer")
+								projectFile.Subtype = Subtype.Designer;
+						}
+						sourceFilesList.Add (projectFile);
+					} else if (item.Name == "Analyzer")
 						analyzerList.Add (msbuildPath);
 				}
 
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -733,13 +733,16 @@ namespace MonoDevelop.Projects
 					var msbuildPath = MSBuildProjectService.FromMSBuildPath (project.sourceProject.BaseDirectory, item.Include);
 
 					if (item.Name == "Compile") {
-						var projectFile = new ProjectFile (msbuildPath, item.Name) { Project = project };
-						const string subtype = "SubType";
-						if (item.Metadata.HasProperty (subtype)) {
-							var property = item.Metadata.GetProperty (subtype);
+						var subtype = Subtype.Code;
+
+						const string subtypeKey = "SubType";
+						if (item.Metadata.HasProperty (subtypeKey)) {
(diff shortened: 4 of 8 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.cs
```diff
@@ -225,9 +227,11 @@ public override IEnumerable<FileProvider> GetFiles (ProgressMonitor monitor, Fil
 				monitor.Log.WriteLine (GettextCatalog.GetString ("Looking in project '{0}'", project.Name));
 				var alreadyVisited = new HashSet<string> ();
 				var conf = project.DefaultConfiguration?.Selector;
-				foreach (ProjectFile file in project.GetSourceFilesAsync (conf).Result.Where (f => filterOptions.NameMatches (f.Name) && File.Exists (f.Name))) {
+				foreach (var file in project.Files) {
 					if ((file.Flags & ProjectItemFlags.Hidden) == ProjectItemFlags.Hidden)
 						continue;
+					if (!filterOptions.NameMatches (file.Name) || !File.Exists (file.Name))
```
Therzok:
```suggestion
					if (!filterOptions.NameMatches (file.Name)
```

GetFileIsText below does File.Exists.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.cs
```diff
@@ -168,10 +170,11 @@ public override IEnumerable<FileProvider> GetFiles (ProgressMonitor monitor, Fil
 							  () => new List<FileProvider> (),
 							  (project, loop, providers) => {
 								  var conf = project.DefaultConfiguration?.Selector;
-
-								  foreach (ProjectFile file in project.GetSourceFilesAsync (conf).Result.Where (f => filterOptions.NameMatches (f.Name) && File.Exists (f.Name))) {
+								  foreach (var file in IncludeCodeBehind ? (IEnumerable<ProjectFile>)project.GetSourceFilesAsync (conf).Result : project.Files) {
 									  if ((file.Flags & ProjectItemFlags.Hidden) == ProjectItemFlags.Hidden)
 										  continue;
+									  if (!filterOptions.NameMatches (file.Name) || !File.Exists (file.Name))
```
Therzok:
```suggestion
									  if (!filterOptions.NameMatches (file.Name))
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
```diff
@@ -193,6 +194,11 @@ public static string FormatPatternToSelectionOption (string pattern, bool regex)
 			scopeStore.AppendValues (GettextCatalog.GetString ("Directories"));
 			scopeStore.AppendValues (GettextCatalog.GetString ("Current document"));
 			scopeStore.AppendValues (GettextCatalog.GetString ("Selection"));
+			if (workspace != null && workspace.GetAllSolutions ().Count () == 1) {
+				scopeStore.AppendValues (GettextCatalog.GetString ("Whole solution with code behind"));
+			} else {
+				scopeStore.AppendValues (GettextCatalog.GetString ("All solutions with code behind"));
+			}
```
slluis: Instead of a new combobox option, can you add an "Include code behind files" checkbox? In this way we can offer that option when the scope selection is workspace, solution or project.

mkrueger: Thought about that but I think that would clutter the UI more and I suspect that 99% of the ppl don't need that. It's a real power user feature. 

The bigger question is if we really, really need that ?!? I added it just in case.

slluis:
I'm re-thinking this fix. In the past we used to get the files from the project files collection and we changed it to GetSourceFilesAsync() to fix the bug https://bugzilla.xamarin.com/show_bug.cgi?id=35343. So using GetSourceFilesAsync() is not about searching on code behind files, it is about finding files imported by msbuild. So it is legit to always call GetSourceFilesAsync() to get the files. What we can do is just skip files in the obj directory.

Therzok:
Doesn't `Files` contain files from imports?

slluis:
No, Files only contains what is specified in the csproj.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
```diff
@@ -621,87 +637,89 @@ void CheckbuttonRecursivelyDestroyed (object sender, EventArgs e)
...
-					if (string.IsNullOrEmpty (item))
-						continue;
-					store.AppendValues (item);
-				}
-				entry.Entry.Text = items[0];
+			var includeCodeBehindItem = new CheckMenuItem (GettextCatalog.GetString ("_Include code behind files"));
+			includeCodeBehindItem.Active = regexSearch;
+			includeCodeBehindItem.DrawAsRadio = false;
+			includeCodeBehindItem.Toggled += delegate {
+				includeCodeBehind.Value = includeCodeBehindItem.Active;
```
Therzok: ref cycle here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
```diff
@@ -621,87 +637,89 @@ void CheckbuttonRecursivelyDestroyed (object sender, EventArgs e)
...
+			wholeWordsOnlyItem.Toggled += delegate {
+				wholeWordsOnly.Value = wholeWordsOnlyItem.Active;
+			};
+			searchEntryFind.Menu.Add (wholeWordsOnlyItem);
 
+			var regexSearchItem = new CheckMenuItem (GettextCatalog.GetString ("_Regex search"));
+			regexSearchItem.Active = regexSearch;
+			regexSearchItem.DrawAsRadio = false;
+			regexSearchItem.Toggled += delegate {
+				regexSearch.Value = regexSearchItem.Active;
```
Therzok: Ref cycle here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
```diff
@@ -621,87 +637,89 @@ void CheckbuttonRecursivelyDestroyed (object sender, EventArgs e)
...
-		{
-			var ec = new EntryCompletion ();
-/*			entry.Changed += delegate {
-				if (!entry.Entry.HasFocus)
-					entry.Entry.GrabFocus ();
+			var caseSensitiveItem = new CheckMenuItem (GettextCatalog.GetString ("_Case sensitive"));
+			caseSensitiveItem.Active = caseSensitive;
+			caseSensitiveItem.DrawAsRadio = false;
+			caseSensitiveItem.Toggled += delegate {
+				caseSensitive.Value = caseSensitiveItem.Active;
```
Therzok: Ref cycle here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
```diff
@@ -621,87 +637,89 @@ void CheckbuttonRecursivelyDestroyed (object sender, EventArgs e)
...
+			var recentSearches = new MenuItem (text);
+			recentSearches.Sensitive = false;
+			searchEntry.Menu.Add (recentSearches);
+
+			foreach (string item in history) {
+				if (item == searchEntry.Entry.Text)
+					continue;
+				var recentItem = new MenuItem (item);
+				recentItem.Name = item;
+				recentItem.Activated += delegate (object mySender, EventArgs myE) {
```
Therzok: ref cycle here.

Therzok: Lambda is assigned to `recentItems` event handler, which is ref'd from `this.searchEntry.Menu`, and lambda is a closure capturing `this`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
```diff
@@ -621,87 +637,89 @@ void CheckbuttonRecursivelyDestroyed (object sender, EventArgs e)
 		}
 
 		const char historySeparator = '\n';
+		const string searchHistoryKey = "MonoDevelop.FindReplaceDialogs.FindHistory";
+		const string replaceHistoryKey = "MonoDevelop.FindReplaceDialogs.ReplaceHistory";
+		static ConfigurationProperty<bool> caseSensitive = ConfigurationProperty.Create ("CaseSensitive", false);
```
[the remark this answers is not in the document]
mkrueger: Doesn't matter much since the dialog got redesigned.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
mkrueger: and where is the problem ? We're talking about UI code here.

mkrueger: The GC should be able to throw away cycles if there is no reference to the cycle.

 I'll try it out :). If that's really the case and this can't be handled we should switch the GC :)

Therzok: Gtk# doesn't have weak event handling. It's not a GC problem.

mkrueger: I checked that the finalizer gets called - 

`class CheckMenuItem2 : CheckMenuItem
		{
			string label;
			public CheckMenuItem2 (string label) : base(label)
			{
				this.label = label;
			}
			~CheckMenuItem2 ()
			{
				Console.WriteLine ("finalized !!!" + label);
			}
		}`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
mkrueger: Same as above -

Using 
```
class MenuItem2 : MenuItem
		{
			public MenuItem2 (string label) : base (label)
			{
			}

			~MenuItem2 ()
			{
				Console.WriteLine ("disposed!!!!");
			}
		}
```

Shows that finalizers get called.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
Therzok: The event handlers are not on the actual widget, but on the toggleref which holds the object. The managed object can go away, but not the toggleref.

Therzok (approved):

slluis (changes requested):
This PR derivated into a UI revamp that is out of scope of this bug, and which should be reviewed by the UX board. All we need to do is ignore files that are under the intermediate folder of a project, and I don't believe anymore that we need a new option for that.

mkrueger:
So that means the 1st implementation was correct ? 

But anyways let's review it by the UX board. It's improving the dialog a lot IMO.

mkrueger:
We should review it - it's enhancing the dialog.

mkrueger:
https://github.com/mono/monodevelop/pull/7458

PR #6379, opened 2018-10-18
[Core] Add binding redirects for NuGet assemblies

Therzok:
MonoDevelop.MSBuildEditor references nuget assemblies, so we need
to add binding redirects in order for that to work

Fixes VSTS #708240 - Missing binding redirects for NuGet.*


Therzok:

mrward:
Is there a plan to update the MSBuild editor to use the same NuGet version as VS Mac? In general the NuGet API breaks with new versions.

Therzok:
Saw this on sevo's machine, he upgraded from 7.6 to 7.7.

mrward:
If it works then all is good. Lots of things no longer compile in the NuGet addin when I upgrade the NuGet assembly version.

Therzok:
MSBuildEditor seems to build fine against MD with this version.

slluis:

PR #6383, opened 2018-10-18
Fixes VSTS Bug 705119: Code suggestion does wrong thing when ( is

mkrueger:
pressed after stray arrow key

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/705119


mkrueger:

PR #6387, opened 2018-10-18
[Core] Fix NuGet sdk resolver not being found in Mono 5.16

mrward:
Mono 5.16.0.173 includes MSBuild 16.0.40. More recent versions of
MSBuild allow an sdk resolver to use a manifest.xml file to define
the assembly location:

    <SdkResolver>
      <Path>..\..\Microsoft.Build.NuGetSdkResolver.dll</Path>
    </SdkResolver>

This was not supported and resulted in the NuGet sdk resolver not
being loaded. Any projects that use an MSBuild sdk from a NuGet
package no longer worked resolved and loaded the MSBuild sdk files.
To fix this the sdk resolution code has been updated based on the
latest MSBuild source code.

Fixes VSTS #706555 - Invalid configuration mapping when loading an
SDK-style project


Therzok (approved):

mrward:

slluis (approved):

PR #6396, opened 2018-10-18
Minor changes to warning and error text color and a better secondary selection text color

vancura:
This increases contrast mainly in Solution Pad:

- Warning and error popovers (better a11y); text here is now always black.
- Fixes a problem when a drawing a secondary (previously light blue) text on inactive (unselected) selection backgrounds; selected secondary text is now always white.


vancura:
Please don't merge just yet, I need to test something first.

vancura:
Everything is working fine now!

PR #6401, opened 2018-10-18
[Tests] Add ProjectTemplateTest helper class

rodrmoya:
And make .NET Core tests use it, for creating projects from templates
and building them.


rodrmoya:
This was added for allowing easily addition of ASP.NET Core tests, but I need it also for AzureFunctions tests, so I guess we can just merge this separately

PR #6405, opened 2018-10-19
Fixes VSTS User Story 700296: UnitTestService rebuilds tests multiple

mkrueger:
times on solution open

ProjectOperations_PackageReferencesModified already contained the
solution for rebuilding. Moved that up to RebuildTests.

### DIFF
--- a/main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestService.cs
+++ b/main/src/addins/MonoDevelop.UnitTesting/Services/UnitTestService.cs
@@ -49,7 +49,7 @@ namespace MonoDevelop.UnitTesting
 	public static class UnitTestService
 	{
 		static ArrayList providers = new ArrayList ();
-		static UnitTest[] rootTests;
+		static UnitTest[] rootTests = Array.Empty<UnitTest> ();
 
 		static UnitTestService ()
 		{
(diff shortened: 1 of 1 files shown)


mrward:
PackageReferenceAdded event is raised when a NuGet package is installed. For an existing NUnit Library project it will not be fired on loading the project. Testing master I only see the tests rebuilt once when the solution is opened in the workspace.

When creating a new NUnit library project the PackageReferenceAdded will be raised once - and you also get a project reference added event for the NUnit library assembly reference - and once for the workspace being opened event. The NuGet package install runs after the project is created/loaded.

mkrueger:
The events subscribed - it subscribes to item open/workbench changed as well. 
That's causing the rebuild event fired several times. With that change the monodevelop solution loads noticeable faster - so it's a problem.

slluis:
@mkrueger I see how this solves the solution load issue. My concern is that with this change, all test tree updates are going to be delayed by 5 seconds, in all use cases, and that's a regression in the user experience.

mkrueger:
I can lower it - works for me with 1s as well - however I know that most notebooks out there are bit slower than my i9. with the loading speed we've currently I didn't find 5s too bad.

However I'll change it to 2s.

mkrueger:

PR #6408, opened 2018-10-19
Fixes VSTS Bug 709557: [Feedback] Copy and paste sometimes fails on VS

mkrueger:
for mac

https://dev.azure.com/devdiv/DevDiv/_workitems/edit/709557

It's a race condition that sometimes happens. That does fix it - the
post formatting corrects the virtual indent.


mkrueger:

slluis:
@mkrueger one test is failing

PR #6410, opened 2018-10-19
Fixes VSTS Bug 709690: [Feedback] Unable to open Specflow feature fil…

mkrueger:
…e in VS2017 7.6 (for MAC)

https://dev.azure.com/devdiv/DevDiv/_workitems/edit/709690

On first glance it looks like a mono/.NET bug - but that solves it.


mkrueger:

slluis:
Not a bug in .net, it is a case of using a loop variable in a closure. The variable will have the value that was assigned to it when the closure is invoked, not the value when the closure was created.

PR #6413, opened 2018-10-19
[IDE] Fix a crash when closing a docknotebook tab

iainx:
The crash trace listed in the bug report doesn't say what the exception that caused the crash was, and I've
not been able to reproduce it, so I'll added a null check and a type check.

Fixes VSTS #706884

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
@@ -1354,8 +1354,14 @@ namespace MonoDevelop.Ide.Gui
 		
 		internal void CloseClicked (object o, TabEventArgs e)
 		{
-			if (e.Tab.Content != null && e.Tab.Content is SdiWorkspaceWindow sdiWorkspace) {
-				sdiWorkspace.CloseWindow (false, true).Ignore ();
+			if (e.Tab.Content != null) {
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
```diff
@@ -1354,7 +1354,9 @@ void CloseWindowEvent (object sender, WorkbenchWindowEventArgs e)
 		
 		internal void CloseClicked (object o, TabEventArgs e)
 		{
-			((SdiWorkspaceWindow)e.Tab.Content).CloseWindow (false, true).Ignore();
+			if (e.Tab.Content != null && e.Tab.Content is SdiWorkspaceWindow sdiWorkspace) {
```
Therzok: Can we log these two cases? Might be useful to know what went wrong.

Therzok:
Is the exception trace in the bug report all of it? I'm wondering if that means closeclicked ran on a bg thread.

iainx:
Yes, that's all of the CrashReport.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
iainx: Good idea, I noticed that while we treat `Tab.Content` as an `SdiWorkspaceWindow` everywhere, it's only enforced to be a `Gtk.Widget`.

slluis:

PR #6415, opened 2018-10-19
[DotNetCore] Handle implicit package reference versions

mrward:
.NET Core 2.2 preview 3 SDK now uses implicit package versions.
The package references loaded by the project model do not have the
version metadata directly applied. This results in the wrong version
being restored by NuGet. For a .NET Core 2.1 project this resulted in
a restore warning:

```
Project dependency Microsoft.NETCore.App does not contain an inclusive
lower bound. Include a lower bound in the dependency version to ensure
consistent restore results.
```

And a build error:

```
Error: NETSDK1061: The project was restored using Microsoft.NETCore.App
version 1.0.0, but with current settings, version 2.1.0 would be used
instead.
```

Added a workaround to find the implicit version and apply it to the
package reference so the correct version is restored by NuGet.

Fixes VSTS #707944 - After installing .net core 2.2 preview, .net
core 2.1 projects no longer build


mrward:

slluis:

PR #6417, opened 2018-10-19
[build] Don't use the API which has been obsoleted

alanmcgovern:
WarnAsError is on so we should make this change to avoid
breaking the build.

### DIFF
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -425,7 +425,7 @@ namespace MonoDevelop.MacIntegration
 				return;
 			var m = NSApplication.SharedApplication.MainMenu;
 			if (m != null) {
-				foreach (NSMenuItem item in m.ItemArray ()) {
+				foreach (NSMenuItem item in m.Items) {
 					var submenu = item.Submenu as MDMenu;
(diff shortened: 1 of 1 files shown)


mrward:
@rodrmoya opened a similar pull request - #6229

mrward:
I don't think this will build since md-addins still uses Xamarin.Mac 4.4.1. Rodrigo's pull request had an md-addins change to use a more recent Xamarin.Mac.

rodrmoya:
Right, it needs the bump to the newest XM in md-addins, but haven't been able to make my 2 PRs build correctly on VSTS, they always fail in forms tests :/

Therzok:
@rodrmoya have you rebased them!

rodrmoya:
yes, several times, just did, and the other that was failing (not related to XM bump) seems to work, but not this one and the related md-addins one, so I guess those tests are affected by Xcode10 upgrade now.

Will investigate further

alanmcgovern:
We're still blocked from easily compiling VSM :(

alanmcgovern:
I'll just close this in favour of the other one.

rodrmoya:
I just bumped the other one to latest master, let's see if tests run ok on VSTS now

PR #6424, opened 2018-10-19
Make sure MD.Refactoring builds on Windows.

KirillOsenkov:
Mono.Cairo.dll on Windows doesn't have Context.SetSourceColor() so we have to emulate it using an extension method so that MonoDevelop can build on Windows.


KirillOsenkov:

PR #6427, opened 2018-10-22
Fixes VSTS Bug 709244: [Feedback] Cannot add .bat (Windows Batch

mkrueger:
files) to projects - VS Mac hangs indefinitely

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/709244


mrward:
Can we fix the text editor not being available to open .bat files? If I right click a .bat file in the Solution window for VS Mac 7.7 I see - Assembly Browser, Hex Editor and Visual Studio Code.app. With this fix the default editor would be the Hex Editor I guess?

mkrueger:
It opens with source editor for me.

mrward:
I debugged the code a bit. The mime type I get for the .bat file is `application/x-msdownload`. This does not match any mime types defined so the open with menu shows VS Code and the Hex Editor:

openwithbatfile

slluis:
@mrward I see there is a check for application/x-msdownload. Is this fixed now?

mkrueger:

PR #6430, opened 2018-10-22
[Core] Fix MSBuild Remove items added when file deleted

mrward:
When a file was deleted from an SDK style project in the IDE sometimes
an MSBuild Remove item would be added to the project file even though
the file was deleted.

    <ItemGroup>
      <Compile Remove="Class1.cs" />
    </ItemGroup>

When a file is deleted in the IDE the following process occurs:

1. Ask for confirmation to delete the file or just remove it from
the project.
2. If delete is selected, then first remove the file from the
Project.Files collection.
3. Delete the file from disk using the FileService.DeleteFile.
4. Save the project file to disk.

If between 2 and 3 an MSBuild target is run for the project, such as
ResolveAssemblyReferences, then the in-memory MSBuildProject would
be updated to have an MSBuild Remove item for the file that was
in the process of being deleted since it was not in the Project.Files
collection but was still on disk. Then when the project was then
saved by the IDE after the file was deleted this new MSBuild Remove
item would be added to the project file when it should not have been.

To handle this the Project now tracks MSBuild Remove items that were
added but not saved to disk so these can be checked when the project
is saved again, and removed if the file has now been deleted. This
in memory approach was used to avoid removing MSBuild Remove items
from an existing project for things like generated files which may
not exist on loading the project.

Fixes VSTS #675480 - Files being added to ItemGroup.Compile[Remove]

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -1620,7 +1620,7 @@ namespace MonoDevelop.Projects
 
 			if (modifiedInMemory) {
 				modifiedInMemory = false;
-				string content = await WriteProjectAsync (new ProgressMonitor ());
+				string content = await WriteProjectAsync (new ProgressMonitor (), inMemoryOnly: true);
 				try {
 					await RemoteBuildEngineManager.RefreshProjectWithContent (FileName, content);
 				} catch {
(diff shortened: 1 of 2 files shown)


Therzok (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -3612,6 +3617,7 @@ internal virtual void SaveProjectItems (ProgressMonitor monitor, MSBuildProject
 							if (file == null || File.Exists (file.FilePath)) {
 								var removeItem = new MSBuildItem (removed.ItemName) { Remove = removed.Include };
 								msproject.AddItem (removeItem);
+								newMSBuildRemoveItems.Add ((removeItem, file.FilePath));
```
slluis: `file` can be null here, see `if` above.

slluis (changes requested):
Null check is missing.

mrward:
Updated with a null check.

slluis (approved):

PR #6443, opened 2018-10-23
[Ide] Show all files option is not remembered for solution

mrward:
No pad user preferences were being saved. This was because the
preferences were being requested from the PadWindow through its
GetMementoCapable method. This was generally returning null since the
PadWindow's content field is not set. Changed the Pad so it checks
the PadCodon's Content implements IMementoCapable instead. This
ensures pad user preferences are stored with the solution's user
preferences.

Fixes VSTS #699562 - Show All Files is not persisted

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Pad.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Pad.cs
@@ -143,8 +143,7 @@ namespace MonoDevelop.Ide.Gui
 		
 		internal IMementoCapable GetMementoCapable ()
 		{
-			PadWindow pw = (PadWindow) window;
-			return pw.GetMementoCapable ();
+			return content?.PadContent as IMementoCapable;
 		}
 		
 		public void Destroy ()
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Pad.cs
```diff
@@ -143,8 +143,7 @@ public void BringToFront (bool grabFocus)
 		
 		internal IMementoCapable GetMementoCapable ()
 		{
-			PadWindow pw = (PadWindow) window;
-			return pw.GetMementoCapable ();
+			return content?.PadContent as IMementoCapable;
```
Therzok: Can we unit test this?

mrward: The memento gets called indirectly via this code in the workbench - https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs#L1259 So a full integration test for this problem would need to load a solution into the workbench, wait for the LoadingUserPreferences to be processed and then check the tree view's option is enabled, maybe getting the node state from SaveTreeState and checking the option is set there.

Therzok: Ah, no, I just wanted a unit test asserting that the result of this would not be null for pads we care about.

PR #6449, opened 2018-10-23
[dotNetCore] Adds Survey base class and implements

jtorres:
Fixes VSTS #677824

NOTE: it's pending the location of `Survey` class

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
@@ -37,6 +37,8 @@ using MonoDevelop.PackageManagement.Commands;
 using MonoDevelop.Projects;
 using MonoDevelop.Projects.MSBuild;
 using MonoDevelop.Ide;
+using MonoDevelop.Ide.Gui.Components;
+using MonoDevelop.Core.Instrumentation;
 
 namespace MonoDevelop.DotNetCore
 {
@@ -44,15 +46,22 @@ namespace MonoDevelop.DotNetCore
 	public class DotNetCoreProjectExtension: DotNetProjectExtension
 	{
 		const string ShownDotNetCoreSdkInstalledExtendedPropertyName = "DotNetCore.ShownDotNetCoreSdkNotInstalledDialog";
+		int successfulBuildCount;
 
 		DotNetCoreMSBuildProject dotNetCoreMSBuildProject = new DotNetCoreMSBuildProject ();
 		DotNetCoreSdkPaths sdkPaths;
+
+		readonly Survey dotNetNsatHatsSurvey;
 
 		public DotNetCoreProjectExtension ()
 		{
 			DotNetCoreProjectReloadMonitor.Initialize ();
 			if (IsDotNetCoreSdk22Installed ())
 				DotNetProjectProxy.ModifyImplicitPackageReferenceVersion = ModifyImplicitPackageReference;
+
+			dotNetNsatHatsSurvey = new Survey ("DotNetCore.NSATHaTSSurvey",
+										"We'd love to hear about your .NET Core experience in Visual Studio for Mac",
+										"https://go.microsoft.com/fwlink/?linkid=2032986");
 		}
 
 		static bool IsDotNetCoreSdk22Installed ()
@@ -372,13 +381,34 @@ namespace MonoDevelop.DotNetCore
 			return base.OnClean (monitor, configuration, operationContext);
 		}
 
-		protected override Task<BuildResult> OnBuild (ProgressMonitor monitor, ConfigurationSelector configuration, OperationContext operationContext)
+		protected async override Task<BuildResult> OnBuild (ProgressMonitor monitor, ConfigurationSelector configuration, OperationContext operationContext)
 		{
-			BuildResult result = CheckCanRunCleanOrBuild ();
-			if (result != null) {
-				return Task.FromResult (result);
+			var result = CheckCanRunCleanOrBuild ();
+
+			if (result == null) {
+				result = await base.OnBuild (monitor, configuration, operationContext);
+
+				if (dotNetNsatHatsSurvey.IsAvailable) {
+					CheckSurveyPrompt (result.Failed);
+				}
 			}
-			return base.OnBuild (monitor, configuration, operationContext);
+			return result;
+		}
+
+		void CheckSurveyPrompt (bool buildFailed)
+		{
+#if DEBUG
+			const int buildTimesForPrompt = 1;
+#else
+			const int buildTimesForPrompt = 5;
+#endif
+
+			if (!buildFailed) {
+				successfulBuildCount++;
+			}
+			if (successfulBuildCount == buildTimesForPrompt) {
+				Runtime.RunInMainThread (() => dotNetNsatHatsSurvey.Prompt ()).Ignore ();
+			}
 		}
 
 		BuildResult CheckCanRunCleanOrBuild ()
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -37,22 +37,31 @@
...
+		readonly Survey dotNetNsatHatsSurvey;
+
 		public DotNetCoreProjectExtension ()
 		{
 			DotNetCoreProjectReloadMonitor.Initialize ();
 			if (IsDotNetCoreSdk22Installed ())
 				DotNetProjectProxy.ModifyImplicitPackageReferenceVersion = ModifyImplicitPackageReference;
+
+			dotNetNsatHatsSurvey = new Survey ("DotNetCore.NSATHaTSSurvey",
+										"We'd love to hear about your .NET Core experience in Visual Studio for Mac",
```
mrward: Should this string be translated?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -730,4 +760,106 @@ protected override string OnGetDefaultBuildAction (string fileName)
 			return base.OnGetDefaultBuildAction (fileName);
 		}
 	}
+
+	//TODO: move to MonoDevelop.Core.Instrumentation??
+	class Survey
+	{
```
mrward: This should be a separate file. Does not need to be in here.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -37,22 +37,31 @@
...
 
+		readonly Survey dotNetNsatHatsSurvey;
+
 		public DotNetCoreProjectExtension ()
 		{
 			DotNetCoreProjectReloadMonitor.Initialize ();
 			if (IsDotNetCoreSdk22Installed ())
 				DotNetProjectProxy.ModifyImplicitPackageReferenceVersion = ModifyImplicitPackageReference;
+
+			dotNetNsatHatsSurvey = new Survey ("DotNetCore.NSATHaTSSurvey",
```
mrward: Are we planning to do surveys for other project types? Just wondering if it makes more sense to have this in a separate project extension.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -37,22 +37,31 @@
...
+										"We'd love to hear about your .NET Core experience in Visual Studio for Mac",
+										"https://go.microsoft.com/fwlink/?linkid=2032986");
 		}
```
mrward: Should this use an aka.ms link? The link currently goes to a getting started with .NET Core page not a survey page.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -372,13 +381,34 @@ protected override Task<BuildResult> OnClean (ProgressMonitor monitor, Configura
...
+				result = await base.OnBuild (monitor, configuration, operationContext);
+
+				if (dotNetNsatHatsSurvey.IsAvailable) {
+					CheckSurveyPrompt (result.Failed);
+				}
+			}
+			return result;
+		}
+
+		void CheckSurveyPrompt (bool buildFailed)
```
mrward: Seems like this logic should be in the Survey class not here.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -730,4 +760,106 @@ protected override string OnGetDefaultBuildAction (string fileName)
...
+		{
+			OnDismissed ();
+			DisableSurvey ();
+		}
+
+		void TakeASurveyHandle ()
+		{
+			OnClicked ();
+			DesktopService.ShowUrl (Url);
+		}
```
mrward: If you take a survey, or at least click it, it looks like you will get prompted again in the next session. Is that the expected behaviour?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -730,4 +760,106 @@ protected override string OnGetDefaultBuildAction (string fileName)
...
+																					"DotNetCore.Surveys", id: $"{Key}.Dismissed");
+		}
+
+		public void Prompt ()
+		{
+			if (!IsAvailable)
+				return;
+				
+			var items = new List<InfoBarItem> ();
+			string msg = GettextCatalog.GetString (Description);
```
mrward: This will not work. The original string needs to be passed to GettextCatalog.GetString.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
jtorres: yep, actually I am asking here and in the PR where I should move it.. any clue?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
jtorres: yes it's in the Survey class

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
rodrmoya: Yes, we know at least of a survey following the results of this one

jtorres: not sure, we know there'll be more survey (in dotNetCore) not sure in other types though

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
jtorres: the link is not definitive, we are waiting for the final one... but I'll ask it again

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
jtorres: well, the condition is to build at least 5 times, so not since others survey might have different conditions not sure if there will be a good place, but I will take a look if I can use a Func<T,bool> or something like that

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
jtorres: ops... ok 👌

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
sayedihashimi: The FWLink will be updated once we have the survey in place. I’ve temporarily set it to a doc in the mean time.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
sayedihashimi: In the review we decided that when the user clicks on Take Survey that the browser will open and the survey prompt goes away. Also the user shouldn’t be promoted for this specific survey again.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -37,22 +37,31 @@
...
+
 		public DotNetCoreProjectExtension ()
 		{
 			DotNetCoreProjectReloadMonitor.Initialize ();
 			if (IsDotNetCoreSdk22Installed ())
 				DotNetProjectProxy.ModifyImplicitPackageReferenceVersion = ModifyImplicitPackageReference;
+
+			dotNetNsatHatsSurvey = new Survey ("DotNetCore.NSATHaTSSurvey",
+										"We'd love to hear about your .NET Core experience in Visual Studio for Mac",
+										"https://go.microsoft.com/fwlink/?linkid=2032986");
```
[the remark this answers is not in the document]
sayedihashimi: aka.ms is ideal when users need to type in the URL. In this case since the tooling opens the browser, a standard FWLink is fine.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
Therzok: `MonoDevelop.Core.Survey`? cc @slluis

PR #6461, opened 2018-10-25
[AspNetCore] Checks publish profiles directory 

jtorres:
- This prevents a DirectoryNotFound exception since we are always enumerating (and assuming) that such directory exists.


mrward:
We have the code `project.BaseDirectory.Combine ("Properties", "PublishProfiles");` in a few places, such as in the ProjectExtensions class. Should this path be defined method in one place? The ProjectExtensions class could extract this common code into one public method and re-used here.

rodrmoya:

PR #6498, opened 2018-10-30
[Core] Try using a hashset to add item results

Therzok:
10 seconds of CPU time is spent in the All LINQ call which does a lot
of work to figure out whether an item is in the list (with the result
from CoreCompileDependsOn running containing a lot of items).

Try using a HashSet which would make this an O(1), not O(n). This
should take down OnGetSourceFiles time on Main.sln by 14s of CPU time.

Fixes VSTS #719932 - OnGetSourceFilesAsync can be optimized

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -539,9 +539,7 @@ namespace MonoDevelop.Projects
 			var evaluatedCompileItems = coreCompileResult.SourceFiles;
 
 			var results = new HashSet<ProjectFile> (evaluatedItems, ProjectFileFilePathComparer.Instance);
-			foreach (var item in evaluatedCompileItems) {
-				results.Add (item);
-			}
+			results.UnionWith (evaluatedCompileItems);
 
 			return results.ToArray ();
 		}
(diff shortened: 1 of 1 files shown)


mrward (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -532,20 +532,29 @@ protected virtual async Task<ImmutableArray<FilePath>> OnGetAnalyzerFiles (Progr
...
 
 			// add in any compile items that we discover from running the CoreCompile dependencies
 			var coreCompileResult = await compileEvaluator.GetItemsFromCoreCompileDependenciesAsync (this, monitor, configuration);
 			var evaluatedCompileItems = coreCompileResult.SourceFiles;
-			var addedItems = evaluatedCompileItems.Where (i => results.All (pi => pi.FilePath != i.FilePath)).ToList ();
-			results.AddRange (addedItems);
+
+			var results = new HashSet<ProjectFile> (evaluatedItems, ProjectFileFilePathComparer.Instance);
+			foreach (var item in evaluatedCompileItems) {
+				results.Add (item);
```
slluis: maybe `results.UnionWith(evaluatedCompileItems)` ?

Therzok: It's doing the same thing under the cover:
https://source.dot.net/#System.Collections/System/Collections/Generic/HashSet.cs,71347988796dc52f

Therzok: But I'll change it.

slluis (approved):

PR #6509, opened 2018-10-31
Updated reference to xamarin/md-addins@da744b9

xamarin-release-manager:
Updated reference to xamarin/md-addins@da744b9


alanmcgovern:
Wil re-create with the xcode 10.1 support patches when they're ready! Our branch targeting d15-9 is green so we just merged it. final packages will be ready shortly.

PR #6518, opened 2018-11-01
[IDE] Wrap the Value getter in a try catch block

iainx:
Wrap the value getter in a try catch block to prevent the crash in 706892. Log the exception in case
we reproduce it.

Fixes VSTS #706892

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.PropertyGrid/PropertyEditorCell.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.PropertyGrid/PropertyEditorCell.cs
@@ -31,6 +31,7 @@ using System.ComponentModel;
 using Gdk;
 using Gtk;
 using MonoDevelop.Ide.Fonts;
+using MonoDevelop.Core;
 
 namespace MonoDevelop.Components.PropertyGrid
 {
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.PropertyGrid/PropertyEditorCell.cs
```diff
@@ -106,18 +107,25 @@ string GetNormalizedText (string s)
...
 		}
-		
+
 		public object Value {
-			get { return Instance != null ? Property.GetValue (Instance) : null; }
+			get {
+				try {
+					return Instance != null ? Property.GetValue (Instance) : null;
+				} catch (Exception e) {
+					LoggingService.LogInternalError ("Hit Bug 706892", e);
```
mrward: Should we log more information here such as the Instance's full type name? I am assuming the exception may not include this information about the type itself.

iainx: Hmm, you're right. The exception I saw triggered did include that information, but it would be good to have it just in case different exceptions can be hit

slluis (approved):

PR #6525, opened 2018-11-01
[Ide] Only subscribe file watchers for metadata references when needed

Therzok:
This should improve performance of filewatchers on metadata references,
but isn't sufficient on its own

Fixes VSTS #669772 - Lazier FSW service subscription for metadata references


Therzok:
Huh, one FSW is not disposed now.

Therzok:
I debugged a bit more after the PR. We behave differently from VSWin, as in, we use actual metadata references when requesting snapshots, thus FSWs are created always.

I tried migrating to their version, and we use many more FSWs in that scenario.

Closing this and the bug report.

PR #6526, opened 2018-11-01
[Core] Optimize the FileWatcherService algorithm so it does not updat…

Therzok:
…e when not needed

Fixes VSTS #643327 - PathTree should have registration modifications better

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
@@ -172,15 +172,15 @@ namespace MonoDevelop.Projects
 			bool modified = false;
 
 			if (monitoredDirectories.TryGetValue (id, out var oldDirectories)) {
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
```diff
@@ -153,22 +153,33 @@ static void RemoveWatcher_NoLock (FilePath directory)
...
 
 			bool modified = false;
 
 			if (monitoredDirectories.TryGetValue (id, out var oldDirectories)) {
+				if (set != null) {
+					var copy = new HashSet<FilePath> (oldDirectories);
+					// Remove the old ones which are not in the new set.
+					oldDirectories.ExceptWith (set);
+					// Only add those which are not in the old set.
+					set.ExceptWith (copy);
```
Therzok: Fix this: we can't mutate the set here, we need to put this in the dictionary and add tree nodes from another set.

slluis (approved):

PR #6528, opened 2018-11-01
Fix GetFocussableWidgets()

KirillOsenkov:
When a widget is not a container, c?.Children?.Length != 0 is true (because Length is `null` which is not 0), and so we were falling through to the `foreach` and null ref-ing on `c.Children`.

Also use pattern matching to simplify code a bit while I'm here.

Rename `GetFocussableWidgets` -> `GetFocusableWidgets`.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/SdiWorkspaceWindow.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/SdiWorkspaceWindow.cs
@@ -361,24 +361,21 @@ namespace MonoDevelop.Ide.Gui
 		
 		static IEnumerable<Gtk.Widget> GetFocussableWidgets (Gtk.Widget widget)
 		{
-			var c = widget as Container;
-
 			if (widget.CanFocus) {
 				yield return widget;
 			}
 
-			if (c != null) {
+			if (widget is Container c) {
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/SdiWorkspaceWindow.cs
```diff
@@ -359,42 +359,38 @@ static void DeepGrabFocus (Gtk.Widget widget)
...
-			var c = widget as Container;
-
 			if (widget.CanFocus) {
 				yield return widget;
 			}
 
-			if (c != null) {
-				foreach (var f in c.FocusChain.SelectMany (GetFocussableWidgets).Where (y => y != null)) {
+			if (widget is Container c) {
+				foreach (var f in c.FocusChain.SelectMany (GetFocusableWidgets).Where (y => y != null)) {
```
Therzok:
```suggestion
				foreach (var f in c.FocusChain.SelectMany (x => GetFocusableWidgets (x))) {
```

I don't see how we can return null from this. Also, Roslyn is better at generating code for the lambda variant, rather than method group, avoiding allocations by using a cached lambda.

slluis:
Does this fix any specific bug?

KirillOsenkov:
No, just something I noticed.

PR #6533, opened 2018-11-02
[Ide] Reset LogView cookie only when cleared for reusing

sevoku:
The `Cookie` should not be reset on a regular `Clear()` (i.e. Clear pad button clicked) but only when the pad is being cleared by a new consumer wanting to reuse an existing output pad.

Fixes VSTS #705935


sevoku:

PR #6556, opened 2018-11-06
[Editor] Fix a few inconsistencies in behaviour

Therzok:
Opening a document should not trigger a fold parsing with the C# old folding parser.

Refresh the token to be used as soon as possible to avoid a possible race where 2 quick fix requests are happening at the same time.

Remove an event handler that can cause a leak.


mkrueger:
How is it when opening files ? 
That should prevent the folding afer opening - can you verify that automatic folding of regions and comments doesn't cause an folding updates after opening ?

That's what the folding parser prevents. It's a very fast parser for just comments & regions - we've that for this purpose.  If behavior doesn't change I'm ok with it.

Therzok:
FirstTimeFoldUpdate unconditionally parses the document via the folding parser.

But, when calling UpdateFoldings, SkipFoldings flag is set, so it doesn't even use the result.

PR #6560, opened 2018-11-06
Bump vs-editor-api to 15.2.3-pre.

KirillOsenkov:
This is to fix a bug found by @garuma where we were attempting to call a native Windows API on a Mac.

https://github.com/KirillOsenkov/vs-editor-api/commit/9b4947b378c3ffc21c59b0197b5ee3cab3fd08d1

### DIFF
--- a/main/msbuild/ReferencesVSEditor.props
+++ b/main/msbuild/ReferencesVSEditor.props
@@ -34,7 +34,7 @@
       <Private>$(ReferencesVSEditorCopyToOutput)</Private>
     </Reference>
     <Reference Include="Microsoft.VisualStudio.Text.Implementation">
-      <HintPath>$(PackagesDirectory)\Microsoft.VisualStudio.Text.Implementation.15.2.2-pre\lib\net46\Microsoft.VisualStudio.Text.Implementation.dll</HintPath>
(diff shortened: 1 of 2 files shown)


garuma (approved):
The fix works on my end. I scouted the rest of the code but it looks like there was only that one pinvoke that was problematic

Therzok:
Why is there a try/catch rather than IsWindows check?

garuma:
Since that was the only spot that seemed to be Windows specific I didn't want to add yet another `IsWindows` check in that codebase when the exception catching seemed sufficient.

But I don't have strong feelings about this so whatever makes the most sense to others (as long as the problem is fixed).

KirillOsenkov:
Merging is blocked because we're currently impacted by the invalidated MyGet feed. Once that is fixed we can trigger a new build and merge this.

KirillOsenkov:
Since the entire vs-editor-api codebase should go away in few months we're OK with merging as is for the sake of expediency.

KirillOsenkov:
@Therzok you mentioned you're OK with this? :) Let's get this in?

Therzok (approved):

Issue #6559, opened 2018-11-06
Caret location in quick fix margin is incorrect


mrward:
VS Mac: 7.7.0.1829
Git revision: 2e0bc097db3ec4bff2bb76153a1d3c956c74a5e1
Build date: 2018-11-02 21:08:01+00
Build branch: release-7.7
Xamarin extensions: aa3663e75d1924e4fe293d8eab6f6f2d98ef1ebf

The caret location shown in the quick fix margin does not match the caret location in the text editor. In the screenshot below the caret is at the bottom of the file whilst the quick fix margin shows the caret about half way down.

<img width="737" alt="texteditorcaret-quick-fix" src="https://user-images.githubusercontent.com/372361/48077331-1f952b80-e1b5-11e8-9fa7-f10f1ad9ce35.png">

Also the caret location does not always update when moving around the document.

> VS bug [#724654](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/724654)

xamarin-release-manager:
Fixed in version 7.8.0.1122 (master)

Pull Request #6561 merged by: Lluis Sanchez
Author: mono
Commit: 2a2c38a755d3121473539c02c41b8a5712db3c91 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 7.8.0.1193 (release-7.8)

Pull Request #6561 merged by: Lluis Sanchez
Author: mono
Commit: 2a2c38a755d3121473539c02c41b8a5712db3c91 (mono/monodevelop)

PR #6561, opened 2018-11-06
Fixes issue #6559 Caret location in quick fix margin is incorrect

mkrueger:
QuickTaskMargin is drawed in a background buffer which blit shouldn't take much time anyways.


mkrueger:

PR #6573, opened 2018-11-08
More detailed error message when MonoDevelop failed to start.

KirillOsenkov:


Therzok:
There was a bug open about this

PR #6577, opened 2018-11-08
[DotNetCore.Tests] Fix old NuGet.Versioning being used by tests

mrward:
This was causing none of the NuGet addin tests to be run. In the
unit test output from running vstool would show an error about
failing to load NuGet.Versioning since the .NET Core tests were
copying the old assembly into the build/tests directory:

```
Could not load file or assembly 'NuGet.Versioning, Version=4.7.0.5,
Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its
dependencies.
```


rodrmoya:
Is that external/nuget-binary being used somewhere else? If not, maybe it should be removed?

mrward:
nuget-binary is still used in a few places. Would need some more work to remove it.

mrward:
Failing tests seem to be unrelated.

PR #6590, opened 2018-11-13
[Core] Fixes VSTS Bug 579048

jtorres:
- Checks if the user has entered or not the extension explicitly
- Detects when there is a suffix like Resources.es.resx

Fixes VSTS Bug 579048

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileTemplateTagsModifier.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileTemplateTagsModifier.cs
@@ -67,7 +67,7 @@ namespace MonoDevelop.Ide.Templates
 					identifier = string.Join (".", suffixes);
 				}
 				else {
-					identifier = CreateIdentifierName (identifier); ;
+					identifier = CreateIdentifierName (identifier);
 				}
 
 				tags ["Name"] = identifier;
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileTemplateTagsModifier.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileTemplateTagsModifier.cs
@@ -57,13 +57,19 @@ namespace MonoDevelop.Ide.Templates
 			//need an 'identifier' for tag substitution, e.g. class name or page name
 			//if not given an identifier, use fileName
 			if ((identifier == null) && (fileName != null))
-				identifier = Path.GetFileName (fileName);
+				identifier = Path.GetFileNameWithoutExtension (fileName);
 
 			if (identifier != null) {
-				//remove all extensions
-				while (Path.GetExtension (identifier).Length > 0)
-					identifier = Path.GetFileNameWithoutExtension (identifier);
-				identifier = CreateIdentifierName (identifier);
+				//check if there any suffix (like: Resources.xx.resx)
+				if (identifier.Contains (".")) {
+					var suffixes = identifier.Split ('.');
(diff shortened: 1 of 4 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/FileTemplateTagsModifier.cs
```diff
@@ -57,13 +57,19 @@ internal static void ModifyTags (SolutionFolderItem policyParent, Project projec
...
-					identifier = Path.GetFileNameWithoutExtension (identifier);
-				identifier = CreateIdentifierName (identifier);
+				//check if there any suffix (like: Resources.xx.resx)
+				if (identifier.Contains (".")) {
+					var suffixes = identifier.Split ('.');
+					suffixes [0] = CreateIdentifierName (suffixes [0]);
+					identifier = string.Join (".", suffixes);
+				}
+				else {
+					identifier = CreateIdentifierName (identifier); ;
```
Therzok: Nit, extra `;`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/SingleFileDescriptionTemplate.cs
```diff
@@ -362,7 +362,10 @@ public virtual string GetFileName (SolutionFolderItem policyParent, Project proj
 			
 			//substitute tags
 			if ((name != null) && (name.Length > 0)) {
-				var model = CombinedTagModel.GetTagModel (ProjectTagModel, policyParent, project, language, entryName ?? name, null);
+				//checks if the entryName contains the extension (i.e. added explicitly with the name)
+				if (entryName.EndsWith (Path.GetExtension (name), StringComparison.OrdinalIgnoreCase))
+					entryName = Path.GetFileNameWithoutExtension (entryName);
+				var model = CombinedTagModel.GetTagModel (ProjectTagModel, policyParent, project, language, entryName ?? name, fileName);
```
mrward: The original code did `entryName ?? name` before using the entryName field. Just wondering if this can be null. It seems that if defaultName is null/empty then it may be possible for entryName to be null?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/SingleFileDescriptionTemplate.cs
```diff
@@ -362,7 +362,10 @@ public virtual string GetFileName (SolutionFolderItem policyParent, Project proj
...
 				fileName = StringParserService.Parse (name, model);
 			}
```
mrward: How easy would it be to add a test for this? I do not think we have any existing tests in this area.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/SingleFileDescriptionTemplate.cs
jtorres: isn't this check enough? https://github.com/mono/monodevelop/pull/6590/files#diff-43f050ea5d4df866c61f60f071994104R358

slluis:
Yeah, it would be nice to have unit tests.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/SingleFileDescriptionTemplate.cs
jtorres: Added 👌

slluis:
👍

PR #6592, opened 2018-11-13
[Refactoring] Compute markup items on the UI thread

Therzok:
Tagging classifiers require to be run on the UI thread for them to work.

Move markup items on the UI thread, and only do simple text diffing on a background
thread

Fixes VSTS #677991 - Refactoring diff calculation needs to happen on the UI thread


Therzok:
@slluis do we still want to take this in or should I just close the bug?

slluis:
Let's take it.

PR #6593, opened 2018-11-13
[Ide] Don't add a hyphen if the first item is not a hyperlink

Therzok:
The hyphen should only be added when the item is a hyperlink
following the description

Fixes VSTS #723318 - InfoBar shows a hyphen between the text and the first item if that item is a link


slluis:

PR #6608, opened 2018-11-15
Revert "[Core] Fixes VSTS Bug 579048"

slluis:
Reverts mono/monodevelop#6590

The above PR seems to be caused a regression in unit tests, that for some reason VSTS is not reporting. Let's see if the revert fixes it.


slluis:
Still fails, so that's not it.

PR #6611, opened 2018-11-15
Revert RefactoringEssentials bump

slluis:
Try reverting RefactoringEssentials bump to see if that fixes the failing test.


slluis:
That was it. The reverted bump fixes the test failure.
/cc @Therzok @mkrueger

Therzok:
Link to failing build?

slluis:
https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=196572, just to give an example. The CSharp test has an error, even though it is not reported by VSTS. Also that test suite takes around 1h to run with the change, when usually it takes 3min.

PR #6614, opened 2018-11-16
Fixes VSTS Bug 729390: [Feedback] Display updates partially when

mkrueger:
commenting a collapsed method

Bug 729390: [Feedback] Display updates partially when commenting a
collapsed method


slluis:
@Therzok can you please review and verify?

Therzok:
Folded method is expanded, fold margin updates properly

PR #6619, opened 2018-11-16
[Git] Fix prompt over and over for keychain items

Therzok:
When requesting a keychain item, we need to append a null byte.
When adding one, we don't.

Fixes VSTS #729363 - Visual Studio for Mac prompts for Git Credentials every time "Update Solution" or "Push Changes" is called


Therzok:
To add tests for url with paths

Therzok:
Tests added.

slluis:

PR #6629, opened 2018-11-20
Rename ITextEditorFactoryService -> ITextEditorInitializationService.

KirillOsenkov:


KirillOsenkov:
This is needed so we can use the standard ITextEditorFactoryService from the editor. This interface was modified from the original to add a new factory method, but there's no reason this should be called ITextEditorFactoryService. I'm freeing up the name and also doing minor simplifications (some parameters were never used).

PR #6630, opened 2018-11-20
Use HttpClient instead of HttpWebRequest

mrward:
Fixes VSTS #601187 - Need to move to HttpClient from HttpWebRequest and use nativehandlers on MacOS

mono-addins change - https://github.com/mono/mono-addins/pull/127


main/src/addins/MacPlatform/BasicAuthenticationHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+
+			return Convert.ToBase64String (bytes);
+		}
+
+		static byte [] GetBytes (string str)
+		{
+			int i = str.Length;
+			byte [] result = new byte [i];
+			for (--i; i >= 0; i--)
+				result [i] = (byte)str [i];
```
Therzok: A string's character can contain multiple bytes. Do we only support ASCII?

main/src/addins/MacPlatform/Dialogs/HttpClientOptionsPanel.cs
```diff
@@ -0,0 +1,61 @@
...
+		public override void ApplyChanges ()
+		{
+			widget.ApplyChanges ();
+		}
+
+		public override Control CreatePanelWidget ()
+		{
+			if (control == null) {
+				widget = new HttpClientOptionsWidget ();
+				control = widget.ToGtkWidget ();
```
Therzok: `new XwtControl (widget)`

main/src/addins/MacPlatform/MacHttpMessageHandlerProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+				return CreateNSUrlSessionHandler (uri, settings);
+			}
+
+			return defaultProvider.CreateHttpMessageHandler (uri, settings);
+		}
+
+		HttpMessageHandler CreateNSUrlSessionHandler (Uri uri, HttpClientSettings settings)
+		{
+			var config = NSUrlSessionConfiguration.DefaultSessionConfiguration;
+			config.RequestCachePolicy = NSUrlRequestCachePolicy.ReloadIgnoringLocalCacheData;
```
Therzok: `NSURLRequestCachePolicy.ReloadRevalidatingCacheData`?

main/src/addins/MacPlatform/MacPlatformSettings.cs
```diff
@@ -0,0 +1,44 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+using MonoDevelop.Core;
+
+namespace MonoDevelop.MacIntegration
+{
+	static class MacPlatformSettings
+	{
+		static readonly string UseNSUrlSessionHandlerProperty = "MonoDevelop.MacIntegration.UseNSUrlSessionHandler";
```
Therzok: `ConfigurationProperty.Create ("MonoDevelop.MacIntegration.UseNSUrlSessionHandler", true)`?

main/src/addins/MacPlatform/NSUrlSessionCredentialsHandler.cs
```diff
@@ -0,0 +1,70 @@
...
+		/// <summary>
+		/// Not all WWW-Authenticate basic auth responses are handled by the NSUrlSessionHandler, such as those
+		/// from VSTS NuGet package sources, so an Authorization header is added to the request and re-sent
+		/// if basic auth credentials can be found.
+		/// </summary>
+		protected override async Task<HttpResponseMessage> SendAsync (HttpRequestMessage request, CancellationToken cancellationToken)
+		{
+			bool retry = false;
+			while (true) {
+				var response = await base.SendAsync (request, cancellationToken);
```
Therzok:
```suggestion
				var response = await base.SendAsync (request, cancellationToken).ConfigureAwait (false);
```

main/src/addins/MacPlatform/BasicAuthenticationHandler.cs
mrward: This was taken from Mono's source code - [BasicClient.cs](https://github.com/mono/mono/blob/master/mcs/class/System/System.Net/BasicClient.cs#L48-L56)

main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/MonoDevelopServerWarningLogHandler.cs
```diff
@@ -0,0 +1,32 @@
...
+		public MonoDevelopServerWarningLogHandler (HttpMessageHandler innerHandler)
+			: base (innerHandler)
+		{
+		}
+
+		protected override async Task<HttpResponseMessage> SendAsync (HttpRequestMessage request, CancellationToken cancellationToken)
+		{
+			var configuration = request.GetOrCreateConfiguration ();
+
+			var response = await base.SendAsync (request, cancellationToken);
```
Therzok:
```suggestion
			var response = await base.SendAsync (request, cancellationToken).ConfigureAwait (false);
```

main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/NuGetHttpSourceAuthenticationHandler.cs
```diff
@@ -0,0 +1,230 @@
...
+			while (true) {
+				// Clean up any previous responses
+				if (response != null) {
+					response.Dispose ();
+				}
+
+				// store the auth state before sending the request
+				var beforeLockVersion = credentials.Version;
+
+				response = await base.SendAsync (request, cancellationToken);
```
Therzok:
```suggestion
				response = await base.SendAsync (request, cancellationToken).ConfigureAwait (false);
```

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Web/HttpSourceAuthenticationHandler.cs
```diff
@@ -0,0 +1,180 @@
...
+			while (true) {
+				// Clean up any previous responses
+				if (response != null) {
+					response.Dispose ();
+				}
+
+				// store the auth state before sending the request
+				var beforeLockVersion = credentials.Version;
+
+				response = await base.SendAsync (request, cancellationToken);
```
Therzok:
```suggestion
				response = await base.SendAsync (request, cancellationToken).ConfigureAwait (false);
```

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Web/ProxyAuthenticationHandler.cs
```diff
@@ -0,0 +1,172 @@
...
+		protected override async Task<HttpResponseMessage> SendAsync (
+			HttpRequestMessage request,
+			CancellationToken cancellationToken)
+		{
+			while (true) {
+				// Store the auth start before sending the request
+				var cacheVersion = credentialCache.Version;
+
+				try {
+					var response = await base.SendAsync (request, cancellationToken);
```
Therzok:
```suggestion
					var response = await base.SendAsync (request, cancellationToken).ConfigureAwait (false);
```

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Web/ProxyAuthenticationHandler.cs
```diff
@@ -0,0 +1,172 @@
...
+
+					if (clientHandler.Proxy == null) {
+						return response;
+					}
+
+					if (credentialService == null) {
+						return response;
+					}
+
+					if (!await AcquireCredentialsAsync (request.RequestUri, cacheVersion, cancellationToken)) {
```
Therzok: This seems to be a pattern. Can we have some kind of extension method implementing this?

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Web/ProxyCache.cs
```diff
@@ -30,25 +30,72 @@ class ProxyCache : IProxyCache
...
+			var proxyCredentials = configuredProxy.Credentials ?? CredentialCache.DefaultCredentials;
+			return cache.TryAdd (configuredProxy.Address, proxyCredentials);
+		}
+
+		public void UpdateCredential (Uri proxyAddress, NetworkCredential credentials)
+		{
+			if (credentials == null)
+				throw new ArgumentNullException (nameof (credentials));
+
+			cache.AddOrUpdate (
```
Therzok: Please use this overload which takes a factory state argument:
https://github.com/dotnet/corefx/blob/1c673ffb9c74e796a9068ec310aaa25121310a12/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs#L1091

Then you can pass the credentials as a parameter and not have to capture in the lambda.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Web/ProxyAuthenticationHandler.cs
mrward: Not planning to change this code. This is based on the original NuGet code and I do not want to make it very different from the original. Ideally we would use the same code but I suspect that is unlikely to happen.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Web/ProxyCache.cs
mrward: ConcurrentDictionary.AddOrUpdate Overload does not seem to exist with Mono. I guess this is new for .NET Core.

mrward: Seems to be new in .NET Framework 4.7.2 looking at the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=netframework-4.7.2#System_Collections_Concurrent_ConcurrentDictionary_2_AddOrUpdate__1__0_System_Func__0___0__1__System_Func__0__1___0__1____0_) and switching between 4.7.2 and 4.7.1 -

main/src/addins/MacPlatform/Dialogs/HttpClientOptionsWidget.UI.cs
```diff
@@ -0,0 +1,76 @@
...
+			Name = GettextCatalog.GetString ("Managed")
+		};
+
+		void Build ()
+		{
+			var mainHBox = new HBox ();
+
+			var httpClientHandlerLabel = new Label ();
+			httpClientHandlerLabel.Name = nameof (httpClientHandlerLabel);
+			httpClientHandlerLabel.Text = GettextCatalog.GetString ("HttpClient implementation:");
```
slluis: Why do we need to offer this option? Can't we just use NSUrlSession?

mrward: Mainly because NSUrlSession can be very strict in what it supports and can fail. Mono's implementation is offered as a fallback if NSUrlSession does not work. One example of this was using Fiddler as a proxy with basic auth - NSUrlSession would not work with it whilst Mono's HttpClient handled it.

Another example is that it is very strict on Basic auth headers. Had to add a workaround in a custom http message handler to support VSTS NuGet package sources so NSUrlSession could be used.

mrward:
Updated this with latest master.

slluis (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Web/ProxyCache.cs
Therzok: Okay, so, I've recently seen this used in roslyn. Just do:
```
Version = Guid.NewGuid();
cache[proxyAddress] = credentials;
```

https://github.com/dotnet/roslyn/issues/33179

PR #6632, opened 2018-11-20
[NuGet] Support conditional PackageReferences when multi-targeting

mrward:
Use the GenerateRestoreGraphFile MSBuild target to deterine the
package spec information instead of determining this from the project
model. This allows multi-targeting projects and conditional
PackageReferences to be supported.

Fixes VSTS #583482 - PackageReference does not respect condition
Fixes VSTS #572320 - Support conditional NuGet PackageReferences in
multi-targeting projects
Fixes VSTS #572329 - Show per-framework dependencies when multi-targeting
Fixes VSTS #563083 - Using Paket with new .fsproj format
Fixes VSTS #738993 - Type system not updated after saving edited project file in editor
Fixes VSTS #714265 - NuGet restores the wrong version of Microsoft.AspNetCore.App
Fixes VSTS #711758 - NuGet restore ignore build targets
Fixes VSTS #740214 - NuGet packages are not getting restored for ASP.NET Core projects with error "System.NullReferenceException"


main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/MSBuildPackageSpecCreator.cs
```diff
@@ -0,0 +1,92 @@
...
+		{
+			if (VerboseLogging)
+				return new LoggingProgressMonitor ();
+
+			return new ProgressMonitor ();
+		}
+
+		public static DependencyGraphSpec GetDependencyGraph (string resultsPath)
+		{
+			if (File.Exists (resultsPath) && new FileInfo (resultsPath).Length != 0) {
```
Therzok: We should use the same file info:
```suggestion
			var fileInfo = new FileInfo (resultsPath);
			if (fileInfo.Exists && fileInfo.Length != 0) {
```

main/src/addins/MonoDevelop.PackageManagement/NuGet.Commands/MSBuildUtility.cs
```diff
@@ -0,0 +1,197 @@
...
+					arguments = GetMSBuildArguments (msbuildBinPath, inputTargetPath, solution);
+				}
+
+				using (var monitor = new LoggingProgressMonitor ()) {
+					var process = Runtime.ProcessService.StartProcess (
+						command,
+						arguments,
+						solution.BaseDirectory,
+						monitor.Log,
+						monitor.Log,
```
Therzok: Shouldn't this be ErrorLog?

Therzok:
```suggestion
						monitor.ErrorLog,
```

main/src/addins/MonoDevelop.PackageManagement/NuGet.ProjectManagement/TempFile.cs
```diff
@@ -0,0 +1,71 @@
...
+
+			string tempDirectory = NuGetEnvironment.GetFolderPath (NuGetFolderPath.Temp);
+
+			Directory.CreateDirectory (tempDirectory);
+
+			int count = 0;
+			do {
+				filePath = Path.Combine (tempDirectory, Path.GetRandomFileName () + extension);
+
+				if (!File.Exists (filePath)) {
```
Therzok: We retry 3 times if the file exists?

main/src/addins/MonoDevelop.PackageManagement/NuGet.Commands/MSBuildUtility.cs
mrward: I think I did this after looking at [existing code](https://github.com/mono/monodevelop/blob/0a95558a7da163a86a6887b0dfbe78292b4f4283/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/CustomCommand.cs#L301) which also does not use the ErrorLog.

Therzok: That code should also be changed.

Therzok (approved):

mrward:
Reworked this pull request with some performance improvements. Also fixes a problem with the type system using out of date reference information on project reload.

Therzok (approved):

slluis (changes requested):
Two general comments:
1) There are several references to `IdeApp.Workspace.ActiveConfiguration`. The changes you are doing are on the project system, and the project system should not depend on the IDE status. If the configuration is required for getting package information, it should be provided as an argument.
2) There is logic in MSBuildUtility for running msbuild, which involves locating the right mono, msbuild, building the command line, etc. I'd like this logic to be encapsulated in MonoDevelop.Projects.MSBuild, since MSBuildProjectService uses a similar logic.

mrward:
Moved part of the code out of MSBuildUtility into a new MSBuildProcessService class in the MonoDevelop.Projects.MSBuild project.

Not sure what to do about the IdeApp.Workspace?.ActiveConfiguration. This is used in the NuGet addin, not directly in the project system, and would be based on the current setting in the IDE. I could push this up to higher in the stack if that would be better.

slluis:
If the code above that uses IdeApp.Workspace ends being called directly or indirectly by the project system, then the configuration must be a parameter that has to be provided by the project system. The case I mean is for example, a project model extension invoking that code to find referenced assemblies or something like that. The project system has no concept of active configuration. An invocation to a method is supposed to always have the same effect, no matter what is the configuration selected in the IDE.

If the above code is only invoked by the NuGet service, then maybe the reference may stay. However, you have to be aware (and users of the api must be aware) that the result of the functions may change depending on what the user selects in the IDE, and the service should probably react to active configuration changes.

mrward:
When you restore or install a new NuGet packages the current configuration would be used at that time - it is not captured forever. So in the unusual case where the user has package references that are conditional on the platform/configuration they would need to restore the packages to fix the references. I can test this in VS on Windows - I suspect they do not automatically restore on switching configuration.

mrward:
VS 2017 runs a restore for a .NET Core project when the configuration is switched between Debug/Release even if there are no conditions for the PackageReferences. VS Mac can do the same.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteBuildEngine.cs
```diff
@@ -153,8 +153,7 @@ async Task<int> LoadProject (string projectFile)
 
 		internal async Task UnloadProject (RemoteProjectBuilder remoteBuilder, int projectId)
 		{
-			lock (remoteProjectBuilders)
-				remoteProjectBuilders.Remove (remoteBuilder.File);
+			OnShuttingDownProjectBuilder (remoteBuilder);
```
slluis: I don't quite understand this commit. If a builder is being shut down it doesn't qualify for being reused for running another target. I don't see in which case the builder could still be in use when UnloadProject is called.

mrward: Correct - if UnloadProject is called then the project builder will not be available. The problem is that if there is reference held for the project builder, so it is still in use, then UnloadProject will not be called until the project builder has finished running the target. So a project builder can have its Shutdown method called but UnloadProject will not be called since it is still running. This can result in the project builder being re-used in [GetRemoteProjectBuilder](https://github.com/mono/monodevelop/blob/b7f59d3c01662a66cae069e6dca66ad9e34cb7d4/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteBuildEngine.cs#L117) since it is still in the list even though it has been marked as shutting down.

 I added a unit test which reproduced this problem before fixing the bug.

slluis: I think the real problem is that `GetRemoteProjectBuilder()` doesn't take into account that `RemoteProjectBuilder.AddReference()` will fail if the builder is being shut down. The fix would be for `GetRemoteProjectBuilder()` to check the result of `AddReference` before returning a builder, and if it fails, unlist the builder and create a new one.

mrward: Reworked the code so the AddReference method is checked in GetRemoteProjectBuilder.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteBuildEngine.cs
```diff
@@ -115,9 +115,16 @@ public Task<RemoteProjectBuilder> GetRemoteProjectBuilder (string projectFile, b
...
 			lock (remoteProjectBuilders) {
 				if (remoteProjectBuilders.TryGetValue (projectFile, out var builder)) {
-					if (addReference)
+					if (addReference && builder.IsCompleted) {
+						if (builder.Result.AddReference ())
+							return builder;
+						else // Project builder is shutting down
+							remoteProjectBuilders.Remove (projectFile);
+					} else if (addReference) {
 						builder.ContinueWith (t => t.Result.AddReference (), TaskContinuationOptions.NotOnFaulted);
```
slluis: AddReference should also be checked here, since by the time this continuation is executed, the builder may already have been shut down.

slluis (approved):

PR #6650, opened 2018-11-23
[NuGet] Modify reference in place on updating a NuGet package

mrward:
On updating a NuGet package in a project that uses a packages.config
the old NuGet package is uninstalled and the new one is installed.
This removes the old references and adds new references. If the
references are in an ItemGroup with a condition then the new
reference can be added into a different ItemGroup if there are other
ItemGroups with references. To prevent this from happening the
changes to made to references are cached and not applied to the
project until the NuGet actions have all been run. This allows
a reference remove followed by an reference being added to be
converted into an update of the original reference's HintPath so its
location in the project file is not changed.

Fixes VSTS #697600 - Sometimes updating package changes its Reference
ItemGroup

Keep fully qualifed hint path on updating reference

Updating a NuGet package, where the project had been modified so the
original reference had a fully qualified hint path, would result in
a relative path used for the hint path on updating the reference.
Now if the original hint path was fully qualified then if the hint
path for the reference is changed it is saved using a fully qualified
path.

Fixes VSTS #697601 - Updating package reference adds relative path to
HintPath


Therzok (approved):

slluis (approved):

mrward:

PR #6651, opened 2018-11-23
[Ide] Improve error handling when loading project templates

mrward:
Failing to access the ~/.templateengine folder due to permissions
or not owning the files would result in the New Project dialog not
opening when trying to create a new project. Now a message dialog
is displayed showing the error before the New Project dialog is
opened.

Also moved the initialization of the templating engine out of the
constructor, to prevent the underlying error being hidden by a
TypeInitializationException. Initialization is now attempted each
time the templates are accessed in case the underlying error has
been fixed whilst the IDE is running.

Fixes VSTS #725161 - Failed to open New Project Dialog. Access to the
path "/Users/user/.templateengine/Visual Studio/7.0/..." is denied.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngine.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngine.cs
@@ -52,16 +52,8 @@ namespace MonoDevelop.Ide.Templates
 	{
 		static EngineEnvironmentSettings environmentSettings = new EngineEnvironmentSettings (new MyTemplateEngineHost (), (env) => new SettingsLoader (env));
 		static TemplateCreator templateCreator = new TemplateCreator (environmentSettings);
-
+		static bool initialized;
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/MicrosoftTemplateEngine.cs
```diff
@@ -139,14 +131,29 @@ static void OnItemTemplateExtensionChanged (object sender, ExtensionNodeEventArg
...
 			return projectTemplates;
 		}
 
 		public static IEnumerable<ItemTemplate> GetItemTemplates ()
 		{
+			EnsureInitialized ();
 			return itemTemplates;
 		}
 
+		static void EnsureInitialized ()
```
Therzok: This can race and double init can happen?

mrward: Currently I believe this is only called on the UI thread.

It would be better if this was Task based I guess but the templating provider API is currently public.

PR #6666, opened 2018-11-26
[A11y] Use the default Mac unfocused background colour for treeview rows

iainx:
Use the default Mac unfocused background colour for treeview rows, except in the
solution pad tree view, because the default colour clashes with the custom
background colour. A property has been added to the CellRendererImage so that
it will ignore the selected icon when the treeview is unfocused

Fixes VSTS #648196

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/CellRendererImage.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/CellRendererImage.cs
@@ -128,13 +128,35 @@ namespace MonoDevelop.Components
 			}
 		}
 
+		bool? ignoreSelection;
+
 		protected override void Render (Gdk.Drawable window, Gtk.Widget widget, Gdk.Rectangle background_area, Gdk.Rectangle cell_area, Gdk.Rectangle expose_area, Gtk.CellRendererState flags)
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components/CellRendererImage.cs
```diff
@@ -128,13 +128,31 @@ public CellRendererImage ()
...
+		// On the Mac, the default unfocused selection background is lighter, so the icon should be black
+		// but our `sel` icons are white.
+		//
+		// Except in the solution treeview, because the Mac's default unfocused selection background is too light for the
+		// custom treeview background
+		bool? ignoreSelection;
+#else
+		bool? ignoreSelection = false;
+#endif
+		public bool IgnoreSelectionWhenUnfocused { get => ignoreSelection.GetValueOrDefault (); set => ignoreSelection = value; }
```
slluis: Add at least a blank line between the property and the method.

slluis:
Please use `Platform.IsMac` instead of `#if MAC`, it will make the code easier to read.

slluis:
@iainx can you merge latest master so that we can have a build?

PR #6669, opened 2018-11-27
Task 608410: Accessibility: MAC: Keyboard: MAS2.4.3: The keyboard foc…

daserge:
…us loss is observed after Accepting/Declining the license agreement.

Set Xwt root window to Preferences dialog on showing, reset in the end

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
@@ -770,7 +770,10 @@ namespace MonoDevelop.Ide.Gui
 				? GettextCatalog.GetString ("Options")
 				: GettextCatalog.GetString ("Preferences");
 
+			var initialRootWindow = Xwt.MessageDialog.RootWindow;
 			try {
+				Xwt.MessageDialog.RootWindow = Xwt.Toolkit.CurrentEngine.WrapWindow (ops);
+
 				if (panelId != null)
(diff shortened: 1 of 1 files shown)


daserge:
@sevoku @slluis please take a look once more.
I've tested it with the latest AtkCocoa changes and the PR is still relevant (a11y focus jumps to the main window without the fix).

bratsche:
I just tested with this change and it appears to fix a number of issues in md-addins.

slluis:
I don't understand this fix. What is the root of the problem?

daserge:
The fix in its current state works around the problem that we have nested modal dialogs running on different Xwt toolkits.
Namely Preferences Windows is on Gtk, and [android-sdk-installer/License dialog is XamMac](https://github.com/xamarin/android-sdk-installer/blob/a955c52b9217913b248f0e938ff39a6bc7cfe48c/Xamarin.Installer.AndroidSDK.Manager/Manager/AndroidSDKManager.cs#L758), so the root cause may be in AtkCocoa or Xwt.
Here's a related recent fix:
https://github.com/xamarin/AtkCocoa/commit/cf53a57b2ad6197bc1ed0af67edfffb768a558c6
Maybe @iainx can help?
The problem was [initially fixed via switching to Gtk toolkit](https://github.com/xamarin/android-sdk-installer/pull/170) for nested Updates and License dialogs, but it was rejected (and then partially fixed in AtkCocoa).

bratsche:
@slluis ping, any chance we can get this merged in?

PR #6670, opened 2018-11-27
[Ide] Add support for Welcome dialog providers

rodrmoya:
If there's a dialog provider for the welcome page, prefer that, and only use
Widget version of it as a fallback. And when so, give control to that
window, hiding the main window.

Also fixes behavior of closing the main window on Mac, which should keep
the app menu bar available to the user.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/561910

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
@@ -758,7 +758,7 @@ namespace MonoDevelop.Ide.Gui
 
 		public void ShowGlobalPreferencesDialog (Window parentWindow, string panelId, Action<OptionsDialog> configurationAction = null)
 		{
-			if (parentWindow == null)
+			if (parentWindow == null && IdeApp.Workbench.RootWindow.Visible)
 				parentWindow = IdeApp.Workbench.RootWindow;
 
 			OptionsDialog ops = new OptionsDialog (
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageService.cs
@@ -27,6 +27,7 @@ using System;
 using MonoDevelop.Ide.Gui;
 using Mono.Addins;
 using System.Linq;
+using MonoDevelop.Components;
 
 namespace MonoDevelop.Ide.WelcomePage
 {
@@ -34,29 +35,60 @@ namespace MonoDevelop.Ide.WelcomePage
 	{
 		static bool visible;
 		static WelcomePageFrame welcomePage;
+		static IWelcomeWindowProvider welcomeWindowProvider;
+		static Window welcomeWindow;
 
 		public static event EventHandler WelcomePageShown;
 		public static event EventHandler WelcomePageHidden;
 
 		internal static void Initialize ()
 		{
+			IdeApp.Workbench.RootWindow.Hidden += (sender, e) => {
+				if (!IdeApp.IsExiting && HasWindowImplementation) {
+					ShowWelcomeWindow ();
+				}
+			};
 			IdeApp.Workspace.FirstWorkspaceItemOpened += delegate {
-				HideWelcomePage ();
+				HideWelcomePageOrWindow ();
 			};
 			IdeApp.Workspace.LastWorkspaceItemClosed += delegate {
-				ShowWelcomePage ();
+				if (!IdeApp.IsExiting && !IdeApp.Workspace.WorkspaceItemIsOpening) {
+					ShowWelcomePageOrWindow ();
+				}
 			};
 			IdeApp.Workbench.DocumentOpened += delegate {
-				HideWelcomePage ();
+				HideWelcomePageOrWindow ();
 			};
 			IdeApp.Workbench.DocumentClosed += delegate {
-				if (IdeApp.Workbench.Documents.Count == 0 && !IdeApp.Workspace.IsOpen)
-					ShowWelcomePage ();
+				if (!IdeApp.IsExiting && IdeApp.Workbench.Documents.Count == 0 && !IdeApp.Workspace.IsOpen) {
+					ShowWelcomePageOrWindow ();
+				}
 			};
 		}
 
-		public static bool WelcomePageVisible {
-			get { return visible; }
+		public static bool WelcomePageVisible => visible;
+
+		public static bool WelcomeWindowVisible => welcomeWindow != null && visible;
+
+		public static bool HasWindowImplementation => AddinManager.GetExtensionObjects<IWelcomeWindowProvider> ().Any ();
+
+		public static void ShowWelcomePageOrWindow (bool animate = false)
+		{
+			// Try to get a dialog version of the "welcome screen" first
+			if (!ShowWelcomeWindow ()) {
+				ShowWelcomePage ();
+			}
+		}
+
+		public static void HideWelcomePageOrWindow (bool animate = false)
+		{
+			if (HasWindowImplementation && welcomeWindowProvider != null && welcomeWindow != null) {
+				welcomeWindowProvider.HideWindow (welcomeWindow);
+			} else {
+				HideWelcomePage (animate);
+			}
+
+			visible = false;
 		}
 
 		public static void ShowWelcomePage (bool animate = false)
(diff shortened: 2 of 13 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/FinalProjectConfigurationPage.cs
```diff
@@ -32,7 +32,7 @@
 
 namespace MonoDevelop.Ide.Projects
 {
-	class FinalProjectConfigurationPage
+	public class FinalProjectConfigurationPage
 	{
```
mrward: Just wondering why we have made these New Project dialog classes public in this change?

rodrmoya: NewProjectDialogController is used in the md-addins branch -> https://github.com/xamarin/md-addins/pull/3765

mrward: If there is a problem here we could consider adding a new method to say ProjectOperations which shows the new project dialog instead of exposing the New Project Dialog itself. It already has some NewSolution methods which do something like this already.

mrward: Looking at the code it looks like you could just use `ProjectOperations.NewSolution ()`. It seems to set the same parameters on the NewProjectDialogController.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
```diff
@@ -758,8 +758,10 @@ public void ShowGlobalPreferencesDialog (Gtk.Window parentWindow)
 
 		public void ShowGlobalPreferencesDialog (Window parentWindow, string panelId, Action<OptionsDialog> configurationAction = null)
 		{
+#if !MAC
```
slluis: Why?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/FinalProjectConfigurationPage.cs
slluis: Matt is right, this can't be public.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/INewProjectController.cs
```diff
@@ -33,7 +33,7 @@
 
 namespace MonoDevelop.Ide.Projects
 {
-	interface INewProjectDialogController
+	public interface INewProjectDialogController
```
slluis: This interface can't be public.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/INewProjectDialogBackend.cs
```diff
@@ -26,7 +26,7 @@
 
 namespace MonoDevelop.Ide.Projects
 {
-	interface INewProjectDialogBackend
+	public interface INewProjectDialogBackend
```
slluis: This interface can't be public.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
```diff
@@ -50,7 +50,7 @@ namespace MonoDevelop.Ide.Projects
 	/// <summary>
 	/// To be renamed to NewProjectDialog
 	/// </summary>
-	class NewProjectDialogController : INewProjectDialogController
+	public class NewProjectDialogController : INewProjectDialogController
```
slluis: This class can't be public.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
rodrmoya: Because setting the parent window to the RootWindow on GtkDialog's makes it show the RootWindow when the dialog is focused, which is not what we want when showing the welcome page on a dialog. This is though a temporary fix, as it makes no sense to have to do that for every single Gtk.Dialog in the code, so have been talking with @sevoku about a better, long term solution for this.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageService.cs
```diff
@@ -27,38 +27,87 @@
 using MonoDevelop.Ide.Gui;
 using Mono.Addins;
 using System.Linq;
+using MonoDevelop.Components;
+
+#if MAC
```
slluis: This file should not have any `#if MAC`. All functionality that is specific to the welcome window should be encapsulated in IWelcomeWindowProvider or the Window class. For example:
``` c#
public interface IWelcomeWindowProvider
{
    Window CreateWindow ();
    void ShowWindow (Window w);
    void HideWindow (Window w);
}
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageService.cs
```diff
@@ -27,38 +27,87 @@
...
 		}
 
 		public static bool WelcomePageVisible {
 			get { return visible; }
 		}
 
+		public static bool WelcomeWindowVisible {
+			get {
+#if MAC
+				var w = (NSWindow)welcomeWindow;
```
slluis: You can use the `visible` field for this, no need to query the window.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
```diff
@@ -335,6 +324,18 @@ public static int RunCustomDialog (Dialog dlg, Window parent)
 				else
 					PlaceDialog (dialog, parent);
 			}).Wait ();
+			#else
+			if (dialog.Modal) {
```
slluis: Why not in MAC?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/FinalProjectConfigurationPage.cs
rodrmoya: Right, followed @mrward 's suggestion, to use `ProjectOperations.NewSolution` instead, so have reverted back all those private->public changes

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/INewProjectController.cs
rodrmoya: done in the last push (--force)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
rodrmoya: Same reason as above, this code tries to set a parent always for the dialog, and doing so, as it will find the `RootWindow`, makes the main window being shown, because it's the parent, when the dialog is focused.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
slluis: Please add a comment explaining that.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageService.cs
rodrmoya: Right, much better, changed now!

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageService.cs
rodrmoya: 👍

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
slluis: Maybe you could just check if IdeApp.Workbench.RootWindow is visible, and only use it as default parent window in that case.

rodrmoya: Right, that works too, changed then

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
Therzok: We can run on Mac without md-addins btw. This should be a check for the welcome dialog.

slluis: You could just check for RootWindow.Visible, like above.

Therzok: Shouldn't this have been changed?

rodrmoya: I'd prefer not to, I came to that solution because I couldn't find a good way of mixing GtkWindow's and NSWindow's, so right now, the best solution is to not set a parent for any dialog at all on Mac. That said, I am talking with @sevoku about it and will come up with a better and long term solution in the next PR or so

Therzok: Does stock MonoDevelop still work after this change?

rodrmoya: yes, it should

PR #6671, opened 2018-11-27
[NuGet] Fix null reference exception in compatiblity checker

mrward:
If a project had both a PackageReference and a packages.config file
the package compatiblity checker would run and fail with a null
reference exception. It was treating the project as though it was
using a packages.config file, when it should be treated as a
PackageReference project.

Fixes VSTS #734876 - NRE in package manager when updating target
framework


mrward:

PR #6674, opened 2018-11-27
 Process markup from formatted text to attributedstring in MacToolboxViewItems

netonjm:
Toolbox allows items with markup. In this PR we added support for cover this, this introduces a new problem with accessibility, we need remove the formatted tags to read correctly the content over VoiceOver.

Fixes Bug https://devdiv.visualstudio.com/DevDiv/_workitems/edit/735244

image

### DIFF
--- a/main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport.Toolbox/MacToolboxViewItems.cs
+++ b/main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport.Toolbox/MacToolboxViewItems.cs
@@ -34,6 +34,7 @@ using MonoDevelop.Ide;
 using MonoDevelop.Components;
 using MonoDevelop.DesignerSupport.Toolbox.NativeViews;
 using CoreAnimation;
+using Xwt.Mac;
 
 namespace MonoDevelop.DesignerSupport.Toolbox
 {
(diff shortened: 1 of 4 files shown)


main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport.Toolbox/MacToolboxViewItems.cs
```diff
@@ -196,6 +203,12 @@ public override bool BecomeFirstResponder ()
 			return base.BecomeFirstResponder ();
 		}
 
+		public void SetText (string formattedText)
+		{
+			TitleTextField.AttributedStringValue = NativeViewHelper.GetAttributedStringFromFormattedText (formattedText);
+			TitleTextField.AccessibilityTitle = Regex.Replace (formattedText, @"<[^>]*>", string.Empty);
```
Therzok: Do not do regex to parse xml.

Therzok:
We should not be using regex to parse xml...

main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport.Toolbox/MacToolboxViewItems.cs
Therzok: If we really want ot use a regex, use a precompiled one and cache it. Otherwise, I'd just do parsing of the string by checking the `<` and `>` and appending non tag parts into a stringbuilder.

PR #6685, opened 2018-11-28
Put common Cocoa helpers/classes in a common place

rodrmoya:
These are used both in the new Cocoa Toolbox, as well as in md-addins (PR coming), and, as we add more Cocoa native code, in other places.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/NSStackViewExtensions.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/NSStackViewExtensions.cs
@@ -32,8 +32,8 @@ namespace MonoDevelop.Components.Mac
 {
 	public static class NSStackViewExtensions
 	{
-		const int RenderingPriority = 1000;
-		const int LowPriority = 249;
+		const int RenderingPriority = (int)NSLayoutPriority.Required;
+		const int LowPriority = (int)NSLayoutPriority.DefaultLow - 1;
 
 		public static NSStackView CreateHorizontalStackView (int spacing = 10) => new NSStackView () {
 			Orientation = NSUserInterfaceLayoutOrientation.Horizontal,
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/NSStackViewExtensions.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/NSStackViewExtensions.cs
@@ -1,0 +1,58 @@
+//
+// NSStackViewExtensions.cs
+//
+// Author:
+//       Rodrigo Moya <rodrigo.moya@xamarin.com>
+//
+// Copyright (c) 2018 
+//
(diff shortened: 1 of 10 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.csproj
```diff
@@ -180,6 +180,10 @@
       <Project>{2B8F4F83-C2B3-4E84-A27B-8DEE1BE0E006}</Project>
       <Name>ICSharpCode.NRefactory.Cecil</Name>
     </ProjectReference>
+    <ProjectReference Include="..\..\..\external\xwt\Xwt.XamMac\Xwt.XamMac.csproj">
```
Therzok: Conditional reference here.

Therzok: Also, if Ide will local copy it, remove the local copy from MacPlatform:
https://github.com/mono/monodevelop/blob/92d1b6b8d642604f57c4b8daed567c72f68fa882/main/src/addins/MacPlatform/MacPlatform.csproj#L60

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/NSStackViewExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+
+#if MAC
+using AppKit;
+
+namespace MonoDevelop.Components.Mac
+{
+	public static class NSStackViewExtensions
+	{
+		const int RenderingPriority = 1000;
+		const int LowPriority = 249;
```
netonjm: NSLayoutPriority.DefaultLow

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/NSStackViewExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+using System;
+
+#if MAC
+using AppKit;
+
+namespace MonoDevelop.Components.Mac
+{
+	public static class NSStackViewExtensions
+	{
+		const int RenderingPriority = 1000;
```
netonjm: NSLayoutPriority.Required

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.csproj
rodrmoya: Can't remove it from MacPlatform, seems to need it for building, but added the condition to the ProjectReference

Therzok: Unresolving. Add a `<Private>false</Private>` to MacPlatform.csproj. MacPlatform depends on MonoDevelop.Ide, so it should not be the assembly local copying the code.

rodrmoya: right, done

PR #6689, opened 2018-11-29
Few fixes for Start page integration

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/737147
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/737155


slluis:
It works much better. The only problem is that if the start window is visible and I click on the icon, the window will move to the center of the window. It should just be brought to the front, not moved.

rodrmoya:
Right, the code in md-addins centres the window every time is shown, so just need to do it only once, so I'll send the fix for that in a md-addins PR

rodrmoya:
Fix included in https://github.com/xamarin/md-addins/pull/3888

PR #6690, opened 2018-11-29
[MacPlatform] Hide Start page when restarting

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/737152


slluis:
@rodrmoya This still doesn't work for me. When I click on Restart the start window is hidden, but the preferences window is not closed and VSM does not shut down.

rodrmoya:
Yes, I'm seeing that when running a local build, and seems also on the generated DMG

rodrmoya:
Indeed, that one is better and really fixes the issue, so closing this one

PR #6695, opened 2018-11-29
More `make dist` missing files

directhex:
Yes yes I know it's a bit late for 7.7.0 stable


directhex:
These checks will never pass, since VSTS still doesn't support PRs from forks 😒

PR #6703, opened 2018-11-30
[A11y] Remove focus from the NSAlert dialog when it is closed

iainx:
Fixes VSTS #648644


iainx:
@slluis already have

slluis:
@iain 648664 is not a bug

iainx:
@slluis Ah, typo'd it

PR #6711, opened 2018-12-03
[Editor] Fix brace completion options not working

Therzok:
For some reason, MS.VS.Text.Impl brace completion manager ended up
being used.

Changing the property key ensures our brace completion manager is used.

Fixes VSTS #738348 - [Feedback] VS for Mac is inserting matching braces even though the option is disabled.


Therzok:
@slluis we might want this in master/7.8 too, since we can't guarantee the ordering used here...

KirillOsenkov:
Good catch, no idea how I missed this all this time... Thanks!

Therzok:
It works right in master though, which is weird.

Therzok:
@monojenkins backport master

PR #6714, opened 2018-12-04
[Ide] Fix New Project dialog's language selector being truncated

mrward:
The size of the language button cell renderer did not increase its
size for any text it needed to contain.

Also increased the padding on the right hand side so the button
has some space before the end of the row.

Also changed the button so it expands to the left instead of the
right, for example when switching from C# to VB.NET.

With these changes the behaviour and UI is back to how things
worked in version 7.0.

Fixes VSTS 738565 - C#/F# chooser in New Project should be moved a few
pixels leftward

This change prevents the language button from being truncated, as shown below:

languageselectortruncated

Expansion of the button to the left matches the behaviour in 7.0.

expandtoleftnewprojectdialog


mrward:
Fixed the other UI problems here so the behaviour and look now matches what we had in version 7.0.

PR #6720, opened 2018-12-05
Fixes VSTS Bug 737175: Unchecking the Organize Using > Place System

mkrueger:
directives first setting does not save

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/737175


slluis:

PR #6737, opened 2018-12-07
[Ide] Allow IDE to specify if it wants its window closed by IWelcomeW…

rodrmoya:
…indowProvider

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/737146

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/IWelcomeWindowProvider.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/IWelcomeWindowProvider.cs
@@ -23,17 +23,27 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
-using System;
+
 using MonoDevelop.Components;
 using Mono.Addins;
 
 namespace MonoDevelop.Ide.WelcomePage
 {
+	public class WelcomeWindowShowOptions
+	{
+		public WelcomeWindowShowOptions (bool closeSolution)
+		{
+			CloseSolution = closeSolution;
+		}
+
+		public bool CloseSolution { get; set; }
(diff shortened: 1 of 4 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/IWelcomeWindowProvider.cs
```diff
@@ -33,7 +33,7 @@ namespace MonoDevelop.Ide.WelcomePage
 	public interface IWelcomeWindowProvider
 	{
 		Window CreateWindow ();
-		void ShowWindow (Window window);
+		void ShowWindow (Window window, bool closingSolution);
```
Therzok: API wise, I'd use a `ShowWindowOptions` class or struct so we can add additional parameters to the API without having to break it via the interface.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageService.cs
```diff
@@ -74,10 +74,14 @@ internal static void Initialize ()
 
 		public static bool HasWindowImplementation => AddinManager.GetExtensionObjects<IWelcomeWindowProvider> ().Any ();
 
-		public static void ShowWelcomePageOrWindow (bool animate = false)
+		public static void ShowWelcomePageOrWindow (bool animate = false, WelcomeWindowShowOptions options = null)
```
slluis: The `animate` parameter should be moved to an option in `WelcomeWindowShowOptions`.

slluis: Nevermind, it can be a parameter too, not a big issue.

Therzok: Would prefer animate to be moved too, that way we have proper API implementation here.

slluis: Maybe we don't need that parameter at all anymore and it can be removed.

PR #6744, opened 2018-12-10
Bug 746222: [Feedback] Fold Regions By Default has no effect

mkrueger:
Seems that the folding parser was still used for that. This is the easiest solution to the issue. 

Query that info before the opening can lead to long delays - folding it afterwards leads to screen updates/unwanted scrolling.

Another viable option would be to remove this functionality @slluis  what do you say to that ?
As far as I can see VS.NET does not offer that. 

Maybe we can remove that and make an external addin out of it ? I think this may be a doable option - then ppl who want that can just install it using the addin functionality.

### DIFF
--- a/main/src/addins/CSharpBinding/CSharpBinding.addin.xml
+++ b/main/src/addins/CSharpBinding/CSharpBinding.addin.xml
@@ -194,6 +194,10 @@
 		<Parser class="MonoDevelop.CSharp.Parser.TypeSystemParser" mimeType = "text/x-csharp" />
 	</Extension>
 
+	<Extension path = "/MonoDevelop/TypeSystem/FoldingParser">
+		<Parser class = "MonoDevelop.CSharp.Parser.CSharpFoldingParser" mimeType="text/x-csharp" />
+	</Extension>
+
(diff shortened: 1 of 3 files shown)


slluis:
@Therzok Looks like you removed this, but it seems to be still required for the "Fold regions by default" option.

Therzok:
Seems like my initial analysis from https://github.com/mono/monodevelop/pull/6556 was incorrect.

The folding parser returns a DefaultParsedDocument which _is_ used.

The problem with that piece of code is that it's slow. How does VS Windows handle this?

mkrueger:
VS Windows handles that by not providing this feature.
If you've a faster way of doing it suggest it :). Roslyn is not that fast this time. But does this code has really any impact on file open perception ? I didn't notice any difference TBH.

It's true that it increases the cost of file open - but waiting for roslyn parse information is way slower. Opening the file and correcting the folds when roslyn parse info is ready will result in jumping/scrolling behavior.

Therzok:
I'm okay with re-merging this. This can go away once we import design time targets, implement proper design time builds and caching mechanisms for them.

PR #6750, opened 2018-12-11
[netcore30] Implements netcoreapp30

jtorres:
*NOTE: Work In Progress....*
*Rebased (Dec. 18th)*

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/695014
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/729967
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/729979
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/727333
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/729984
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/752729
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/757629

# Known issues
- when global.json is renamed to ie.e global1.json it's not caught by `FileChanged`
# FIXME
- .NET core 3.0 aka.ms is still not available [here](https://github.com/mono/monodevelop/compare/netcoreapp30?expand=1#diff-45fb0e36ce58ce6f9f6622d7a63f948eR41)
- Mono runtime version is missing  [here](https://github.com/mono/monodevelop/pull/6750/files#diff-433a7fd2cf020622736651e84dc9295cR47)
- .NET Standard 2.1 is still not available (templates, core...)
- Better global.json FileRenamed / Added / Removed handling
- Improve Test based on specific .NET Core SDK version + refactoring
- Better Restoring at Sol level according to [this](https://github.com/mono/monodevelop/pull/6750#discussion_r245921876).

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateStringTagProvider.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateStringTagProvider.cs
@@ -126,12 +126,10 @@ namespace MonoDevelop.DotNetCore.Templating
 		{
 			DotNetCoreVersion dotNetCoreSdk = null;
 
-			if (tag.StartsWith ("DotNetCoreSdk.3.0", StringComparison.OrdinalIgnoreCase)) {
-				dotNetCoreSdk = GetDotNetCoreSdk30Version ();
-			} else if (tag.StartsWith ("DotNetCoreSdk.2.2", StringComparison.OrdinalIgnoreCase)) {
-				dotNetCoreSdk = GetDotNetCoreSdk22Version ();
-			} else if (tag.StartsWith ("DotNetCoreSdk.2.1", StringComparison.OrdinalIgnoreCase)) {
-				dotNetCoreSdk = GetDotNetCoreSdk21Version ();
+			foreach (var sdk in SupportedSDK) {
+				if (tag.StartsWith ($"DotNetCoreSdk.{sdk}", StringComparison.OrdinalIgnoreCase)) {
+					dotNetCoreSdk = GetDotNetCoreSdkVersion (DotNetCoreVersion.Parse (sdk));
+				}
 			}
 
 			if (dotNetCoreSdk == null)
@@ -150,22 +148,15 @@ namespace MonoDevelop.DotNetCore.Templating
 			return string.Empty;
 		}
 
-		DotNetCoreVersion GetDotNetCoreSdk30Version ()
+		DotNetCoreVersion GetDotNetCoreSdkVersion (DotNetCoreVersion version)
 		{
-			return DotNetCoreSdk.Versions
-				.FirstOrDefault (v => v.Major == 3 && v.Minor == 0);
-		}
+			//special case 2.1
+			if (version.Major == 2 && version.Minor == 1)
+				return DotNetCoreSdk.Versions
+				.FirstOrDefault (v => v.Major == version.Major && v.Minor == version.Minor && v.Patch >= 300);
 
-		DotNetCoreVersion GetDotNetCoreSdk22Version ()
-		{
 			return DotNetCoreSdk.Versions
-				.FirstOrDefault (v => v.Major == 2 && v.Minor == 2);
-		}
-
-		DotNetCoreVersion GetDotNetCoreSdk21Version ()
-		{
-			return DotNetCoreSdk.Versions
-				.FirstOrDefault (v => v.Major == 2 && v.Minor == 1 && v.Patch >= 300);
+				.FirstOrDefault (v => v.Major == version.Major && v.Minor == version.Minor);
 		}
 
 		/// <summary>
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateStringTagProvider.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateStringTagProvider.cs
@@ -56,9 +56,8 @@ namespace MonoDevelop.DotNetCore.Templating
 					$"DotNetCoreSdk.{SupportedSDK [i]}.Templates.Web.ProjectTemplates.nupkg",
 					GettextCatalog.GetString (string.Format (".NET Core SDK {0} Web Project Templates NuGet package path", SupportedSDK[i]))
 				);
-
-				//TODO: Ask why 2.1 does not support NUnit
-				if (i >0 ) { //2.2 and before 
+					
+				if (i > 0 ) { //2.2 and before 
 					yield return new StringTagDescription (
 						$"DotNetCoreSdk.{SupportedSDK [i]}.Templates.NUnit3.DotNetNew.Template.nupkg",
 						GettextCatalog.GetString (string.Format (".NET Core SDK {0} NUnit Project Templates NuGet package path", SupportedSDK [i]))
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateStringTagProvider.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateStringTagProvider.cs
@@ -37,44 +37,34 @@ namespace MonoDevelop.DotNetCore.Templating
 	[Extension]
 	class DotNetCoreProjectTemplateStringTagProvider : IStringTagProvider
 	{
+		string [] SupportedSDK = { "2.1", "2.2", "3.0" };
+
 		public IEnumerable<StringTagDescription> GetTags (Type type)
 		{
-			// 2.1 templates
-			yield return new StringTagDescription (
-				"DotNetCoreSdk.2.1.Templates.Common.ProjectTemplates.nupkg",
-				GettextCatalog.GetString (".NET Core SDK 2.1 Common Project Templates NuGet package path")
-			);
+			for (var i = 0; i < SupportedSDK.Length; i++) {
+				yield return new StringTagDescription (
+					$"DotNetCoreSdk.{SupportedSDK[i]}.Templates.Common.ProjectTemplates.nupkg",
+					GettextCatalog.GetString (string.Format (".NET Core SDK {0} Common Project Templates NuGet package path", SupportedSDK [i]))
+				);
 
-			yield return new StringTagDescription (
-				"DotNetCoreSdk.2.1.Templates.Test.ProjectTemplates.nupkg",
-				GettextCatalog.GetString (".NET Core SDK 2.1 Test Project Templates NuGet package path")
-			);
+				yield return new StringTagDescription (
+					$"DotNetCoreSdk.{SupportedSDK [i]}.Templates.Test.ProjectTemplates.nupkg",
+					GettextCatalog.GetString (string.Format (".NET Core SDK {0} Test Project Templates NuGet package path", SupportedSDK[i]))
+				);
 
-			yield return new StringTagDescription (
-				"DotNetCoreSdk.2.1.Templates.Web.ProjectTemplates.nupkg",
-				GettextCatalog.GetString (".NET Core SDK 2.1 Web Project Templates NuGet package path")
-			);
+				yield return new StringTagDescription (
+					$"DotNetCoreSdk.{SupportedSDK [i]}.Templates.Web.ProjectTemplates.nupkg",
+					GettextCatalog.GetString (string.Format (".NET Core SDK {0} Web Project Templates NuGet package path", SupportedSDK[i]))
+				);
 
-			// 2.2 templates
-			yield return new StringTagDescription (
-				"DotNetCoreSdk.2.2.Templates.Common.ProjectTemplates.nupkg",
-				GettextCatalog.GetString (".NET Core SDK 2.2 Common Project Templates NuGet package path")
-			);
-
-			yield return new StringTagDescription (
-				"DotNetCoreSdk.2.2.Templates.Test.ProjectTemplates.nupkg",
-				GettextCatalog.GetString (".NET Core SDK 2.2 Test Project Templates NuGet package path")
-			);
-
-			yield return new StringTagDescription (
-				"DotNetCoreSdk.2.2.Templates.Web.ProjectTemplates.nupkg",
-				GettextCatalog.GetString (".NET Core SDK 2.2 Web Project Templates NuGet package path")
-			);
-
-			yield return new StringTagDescription (
-				"DotNetCoreSdk.2.2.Templates.NUnit3.DotNetNew.Template.nupkg",
-				GettextCatalog.GetString (".NET Core SDK 2.2 NUnit Project Templates NuGet package path")
-			);
+				//TODO: Ask why 2.1 does not support NUnit
+				if (i >0 ) { //2.2 and before 
+					yield return new StringTagDescription (
+						$"DotNetCoreSdk.{SupportedSDK [i]}.Templates.NUnit3.DotNetNew.Template.nupkg",
+						GettextCatalog.GetString (string.Format (".NET Core SDK {0} NUnit Project Templates NuGet package path", SupportedSDK [i]))
+					);
+				}
+			}
 		}
 
 		public object GetTagValue (object instance, string tag)
@@ -131,16 +121,18 @@ namespace MonoDevelop.DotNetCore.Templating
 		}
 
 		/// <summary>
-		/// Only .NET Core SDKs 2.1/2.2 are supported.
+		/// Only .NET Core SDKs 2.1/2.2/3.0 are supported.
 		/// </summary>
 		string GetDotNetCoreSdkTemplatesDirectory (string tag)
 		{
 			DotNetCoreVersion dotNetCoreSdk = null;
 
-			if (tag.StartsWith ("DotNetCoreSdk.2.1", StringComparison.OrdinalIgnoreCase)) {
-				dotNetCoreSdk = GetDotNetCoreSdk21Version ();
+			if (tag.StartsWith ("DotNetCoreSdk.3.0", StringComparison.OrdinalIgnoreCase)) {
+				dotNetCoreSdk = GetDotNetCoreSdk30Version ();
 			} else if (tag.StartsWith ("DotNetCoreSdk.2.2", StringComparison.OrdinalIgnoreCase)) {
 				dotNetCoreSdk = GetDotNetCoreSdk22Version ();
+			} else if (tag.StartsWith ("DotNetCoreSdk.2.1", StringComparison.OrdinalIgnoreCase)) {
+				dotNetCoreSdk = GetDotNetCoreSdk21Version ();
 			}
 
 			if (dotNetCoreSdk == null)
@@ -157,6 +149,12 @@ namespace MonoDevelop.DotNetCore.Templating
 			}
 
 			return string.Empty;
+		}
+
+		DotNetCoreVersion GetDotNetCoreSdk30Version ()
+		{
+			return DotNetCoreSdk.Versions
+				.FirstOrDefault (v => v.Major == 3 && v.Minor == 0);
 		}
 
 		DotNetCoreVersion GetDotNetCoreSdk22Version ()
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateWizard.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateWizard.cs
@@ -129,7 +129,9 @@ namespace MonoDevelop.DotNetCore.Templating
 				if (!SupportsNetCore1x ()) {
 					var highestFramework = DotNetCoreProjectSupportedTargetFrameworks.GetNetCoreAppTargetFrameworks ().FirstOrDefault ();
 					if (highestFramework != null) {
-						if (highestFramework.IsNetCoreApp22 ()) {
+						if (highestFramework.IsNetCoreApp30 ()) {
+							Parameters ["UseNetCore30"] = "true";
+						} else if (highestFramework.IsNetCoreApp22 ()) {
 							Parameters ["UseNetCore22"] = "true";
 						} else if (highestFramework != null && highestFramework.IsNetCoreApp21 ()) {
 							Parameters ["UseNetCore21"] = "true";
@@ -142,8 +144,10 @@ namespace MonoDevelop.DotNetCore.Templating
 				} else {
 					var highestFramework = DotNetCoreProjectSupportedTargetFrameworks.GetNetCoreAppTargetFrameworks ().FirstOrDefault ();
 					if (highestFramework != null) {
-						if (highestFramework.IsNetCoreApp22 ()) {
-							Parameters["UseNetCore22"] = "true";
+						if (highestFramework.IsNetCoreApp30 ()) {
+							Parameters["UseNetCore30"] = "true";
+						} else if (highestFramework.IsNetCoreApp22 ()) {
+							Parameters ["UseNetCore22"] = "true";
 						} else if (highestFramework.IsNetCoreApp21 ()) {
 							Parameters ["UseNetCore21"] = "true";
 						} else if (highestFramework.IsNetCoreApp20 ()) {
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateWizard.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateWizard.cs
@@ -123,7 +123,7 @@ namespace MonoDevelop.DotNetCore.Templating
 			} else {
 				if (!SupportsNetCore1x ()) {
 					var highestFramework = DotNetCoreProjectSupportedTargetFrameworks.GetNetCoreAppTargetFrameworks ().FirstOrDefault ();
-					if (highestFramework != null) {
+					if (highestFramework != null && highestFramework.IsNetCoreAppOrHigher (DotNetCoreVersion.Parse ("2.0"))) {
 						var parameter = highestFramework.GetParameterName ();
 						if (!string.IsNullOrEmpty (parameter))
 							Parameters [parameter] = "true";
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateWizard.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateWizard.cs
@@ -33,6 +33,9 @@ namespace MonoDevelop.DotNetCore.Templating
 {
 	class DotNetCoreProjectTemplateWizard : TemplateWizard
 	{
+		const string defaultParameterNetCore20 = "UseNetCore20";
+		const string defaultParameterNetCore1x = "UseNetCore1x";
+
 		List<TargetFramework> targetFrameworks;
 
 		public override WizardPage GetPage (int pageNumber)
@@ -42,17 +45,11 @@ namespace MonoDevelop.DotNetCore.Templating
 			return page;
 		}
 
-		public override int TotalPages {
-			get { return GetTotalPages (); }
-		}
+		public override int TotalPages => GetTotalPages ();
 
-		public override string Id {
-			get { return "MonoDevelop.DotNetCore.ProjectTemplateWizard"; }
-		}
+		public override string Id => "MonoDevelop.DotNetCore.ProjectTemplateWizard";
 
-		internal IList<TargetFramework> TargetFrameworks {
-			get { return targetFrameworks; }
-		}
+		internal IList<TargetFramework> TargetFrameworks => targetFrameworks;
 
 		/// <summary>
 		/// When only .NET Core 2.0 is installed there is only one option in the drop down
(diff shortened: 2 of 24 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateStringTagProvider.cs
```diff
@@ -131,16 +121,18 @@ string GetTemplateFileNamePrefix (string tag)
 		}
 
 		/// <summary>
-		/// Only .NET Core SDKs 2.1/2.2 are supported.
+		/// Only .NET Core SDKs 2.1/2.2/3.0 are supported.
 		/// </summary>
 		string GetDotNetCoreSdkTemplatesDirectory (string tag)
```
slluis: Can't we make this code more generic? The framework major and minor versions can be parsed from the tag. We can still have a list of supported frameworks. Adding a new supported framework should be as easy as adding a new item to the list.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateWizard.cs
```diff
@@ -129,7 +129,9 @@ void ConfigureDefaultParameters ()
 				if (!SupportsNetCore1x ()) {
 					var highestFramework = DotNetCoreProjectSupportedTargetFrameworks.GetNetCoreAppTargetFrameworks ().FirstOrDefault ();
 					if (highestFramework != null) {
-						if (highestFramework.IsNetCoreApp22 ()) {
```
slluis: The code in this file could also be made more generic. It shouldn't be necessary to add new hardcoded checks for every release. Adding to a list of supported frameworks should be enough.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateWizardPage.cs
```diff
@@ -96,11 +96,13 @@ void UpdateTargetFrameworkParameters ()
 			TargetFramework framework = targetFrameworks [selectedTargetFrameworkIndex];
 			wizard.Parameters ["Framework"] = framework.Id.GetShortFrameworkName ();
 
+			wizard.Parameters ["UseNetCore30"] = framework.IsNetCoreApp30 ().ToString ();
```
slluis: Same as above, it could be more generic.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateStringTagProvider.cs
```diff
@@ -37,44 +37,34 @@ namespace MonoDevelop.DotNetCore.Templating
...
+					GettextCatalog.GetString (string.Format (".NET Core SDK {0} Test Project Templates NuGet package path", SupportedSDK[i]))
+				);
+
+				yield return new StringTagDescription (
+					$"DotNetCoreSdk.{SupportedSDK [i]}.Templates.Web.ProjectTemplates.nupkg",
+					GettextCatalog.GetString (string.Format (".NET Core SDK {0} Web Project Templates NuGet package path", SupportedSDK[i]))
+				);
+
+				//TODO: Ask why 2.1 does not support NUnit
+				if (i >0 ) { //2.2 and before
```
Therzok:
```suggestion
				if (i > 0) { // 2.2 and before 
```

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
```diff
@@ -29,43 +29,81 @@
...
 				return;
+			}
 
-			msbuildSDKsPath = Path.Combine (SdksParentDirectory, "Sdks");
+			DotNetCoreVersion requiredVersion;
+			DotNetCoreVersion.TryParse (specificVersion, out requiredVersion);
+
+			if (requiredVersion == null)
+				//TODO: what kind of Exception?
+				throw new Exception ("TargetFramework version unknown");
```
mrward: Instead of throwing an exception, which will likely break the loading of .NET Core projects we could just indicate a problem somewhere, say on the DotNetCoreSdk class, and just use the latest sdk. Then show this information in the Solution pad with a warning/error icon. Could also add a warning on building - not sure what MSBuild will do here.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
```diff
@@ -29,43 +29,81 @@
...
-			SdksParentDirectory = Path.Combine (sdkRootPath, latestVersion.OriginalString);
-			if (SdksParentDirectory == null)
+			//look at the WHOLE logic, create SDKResovlerFoo (testing), use here
+			DotNetCoreVersion targetVersion = null;
+			var specificVersion = ReadGlobalJson (workingDir);
+
+			//if !global.json, returns latest
+			if (string.IsNullOrEmpty (specificVersion)) {
+				msbuildSDKsPath = GetSdksParentDirectory (GetLatesSdk (dotNetCoreVersion));
+				Exist = true;
```
mrward: Typo - should be 'GetLatestSdk'

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
```diff
@@ -156,5 +185,43 @@ IEnumerable<DotNetCoreVersion> GetInstalledSdkVersions (string sdkRootPath)
...
+		{
+			var globalJsonPath = Path.Combine (workingDir, "global.json");
+			if (!File.Exists (globalJsonPath))
+				return string.Empty;
+
+			using (var r = new StreamReader (globalJsonPath)) {
+				var token = JObject.Parse (r.ReadToEnd ());
+				var version = (string)token.SelectToken ("sdk").SelectToken ("version");
+				return version;
+			}
```
mrward: I guess we should handle a global.json file without an sdk token and ignore it in this case.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
mrward: Not sure this is correct. If my project is .NET Core 2.0 but .NET Core 2.1 SDK is installed then GetLatestSdk will not match anything and return null. I think the old behaviour is correct where it just gets the latest sdk version, and ignores the project's target framework version, - that will match what MSBuild does.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
```diff
@@ -29,43 +29,81 @@
...
+				if (requiredVersion.Major == 2 && requiredVersion.Minor >= 1) {
+					targetVersion = SdkVersions.Where (version => version.Major == requiredVersion.Major
+																	&& version.Minor == requiredVersion.Minor
+																	&& version.Patch > requiredVersion.Patch)
+												.OrderByDescending (version => version.Patch).FirstOrDefault ();
+				} else {
+					targetVersion = SdkVersions.Where (version => version.Major == requiredVersion.Major && version.Minor == requiredVersion.Minor)
+												.OrderByDescending (version => version.Patch).FirstOrDefault ();
+				}
+
```
mrward: Just wondering why we have a check for 2.1 but not for others. Do we not need the same logic for the patch if it does no exactly match?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
```diff
@@ -29,43 +29,81 @@
...
+				if (targetVersion == null)
+					//TODO: what kind of Exception?
+					throw new Exception ($"Couldn't find SDK Version {specificVersion} for {requiredVersion}");
```
mrward: I think this will break the loading of the project. Mentioned this elsewhere, but maybe we can just have a property on the DotNetCoreSdkPaths to indicate this as a problem and then report it elsewhere on the solution pad and when building as a warning.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -335,11 +352,46 @@ protected override void OnItemReady ()
...
+
+			if (Project.ParentSolution == null)
+				return;
+
+			if (Project.ParentSolution.ExtendedProperties.Contains (GlobalJsonPathExtendedPropertyName))
+				return;
+
+			//detect globaljson
+			var globalJsonPath = Path.Combine (Project.ParentSolution.BaseDirectory, "global.json");
+			if (File.Exists (globalJsonPath))
```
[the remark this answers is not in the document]
jtorres: according to global.json documentation: 
> .NET Core SDK looks for a global.json file in the current working directory (which isn't necessarily the same as the project directory) or one of its parent directories.

Now that I read again,, we should check from solution root up to all its parent directories right?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -335,11 +352,46 @@ protected override void OnItemReady ()
...
+		{
+			if (!IdeApp.Workspace.IsOpen)
+				return;
+
+			var progressMonitor = new ProgressMonitor ();
+			foreach (var project in IdeApp.Workspace.GetAllItems<DotNetProject> ()) {
+				if (project.HasFlavor<DotNetCoreProjectExtension> ()) {
+					await project.ReevaluateProject (progressMonitor);
+					RestorePackagesInProjectHandler.Run (project, restoreTransitiveProjectReferences: true, reevaluateBeforeRestore: true);
+				}
```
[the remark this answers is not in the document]
jtorres: hmm not sure,,, so `reevaluateBeforeRestore` should be `false`?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectSupportedTargetFrameworks.cs
```diff
@@ -62,6 +62,12 @@ public IEnumerable<TargetFramework> GetFrameworks ()
 
 		public static IEnumerable<TargetFramework> GetNetStandardTargetFrameworks ()
 		{
+			if (DotNetCoreRuntime.IsNetCore30Installed () || MonoRuntimeInfoExtensions.CurrentRuntimeVersion.SupportsNetStandard21 ())
+				yield return CreateTargetFramework (".NETStandard", "2.1");
+
+			if (DotNetCoreRuntime.IsNetCore2xInstalled () || MonoRuntimeInfoExtensions.CurrentRuntimeVersion.SupportsNetStandard21 ())
+				yield return CreateTargetFramework (".NETStandard", "2.1");
```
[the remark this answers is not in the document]
jtorres: nope, this is wrong.. will fix it

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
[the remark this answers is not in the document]
mrward: If a restore is being run then even with reevaluateBeforeRestore=false, then you should not need to call RevaluateProject. A restore will re-evaluate if the project if it needs to.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
jtorres: This is logic:
> Starting with .NET Core 2.0, the following rules apply when determining which version of the SDK to use:

If no global.json file is found or global.json doesn't specify an SDK version, the latest installed SDK version is used. Latest SDK version can be either release or pre-release - the highest version number wins.
If global.json does specify an SDK version:
If the specified SDK version is found on the machine, that exact version is used.
If the specified SDK version can't be found on the machine, the latest installed SDK patch version of that version is used. Latest installed SDK patch version can be either release or pre-release - the highest version number wins. In .NET Core 2.1 and higher, the patch versions lower than the patch version specified are ignored in the SDK selection.
If the specified SDK version and an appropriate SDK patch version can't be found, an error is thrown.

if that does not make sense to you, let's talk about it in the channel or give me the sdk you are setting and the one you expect, you also can check it out using the new tests for sdk resolver.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
jtorres: according to the documentation: 

> If the specified SDK version and an appropriate SDK patch version can't be found, an error is thrown.

will check on VS on Windows though

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
jtorres: yep, look at the logic I pointed out above regarding to global.json and sdks. More info [here](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json)

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
jtorres: I think I tested it, but I double check it anyway

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
mrward: `If no global.json file is found or global.json doesn't specify an SDK version, the latest installed SDK version is used.` Right now if no global.json file is used, or does not specify a version, the code seems to be using the project's target framework version to match an SDK and does not necessarily use the latest SDK installed. I think we should not be passing a version to GetLatestSdk.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
mrward: Yeah I see that on Windows an exception is thrown - I am just worried this will break the loading of the project. I think we can handle the problem elsewhere - say as a build error or a error icon on the project in the solution pad.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
mrward: The rules have `In .NET Core 2.1 and higher, the patch versions lower than the patch version specified are ignored in the SDK selection.` but the code checks for major == 2 and minor >= 1, not 2.1 and higher.

mrward: Also I do not think we handle this case since the patch version in DotNetCoreVersion is the full number (e.g. 300) not the last two digits: `The patch version is defined by the last two digits (yz) in the last portion of the number (xyz) for SDK versions 2.1.100 and higher. For example, if you specify 2.1.300 as the SDK version, SDK selection finds up to 2.1.399 but 2.1.400 isn't considered a patch version for 2.1.300.` - not sure how this works on Windows.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectSupportedTargetFrameworks.cs
```diff
@@ -62,6 +62,12 @@ public IEnumerable<TargetFramework> GetFrameworks ()
 
 		public static IEnumerable<TargetFramework> GetNetStandardTargetFrameworks ()
 		{
+			if (DotNetCoreRuntime.IsNetCore30Installed () || MonoRuntimeInfoExtensions.CurrentRuntimeVersion.SupportsNetStandard21 ())
+				yield return CreateTargetFramework (".NETStandard", "2.1");
+
+			if (DotNetCoreRuntime.IsNetCore2xInstalled () || MonoRuntimeInfoExtensions.CurrentRuntimeVersion.SupportsNetStandard21 ())
+				yield return CreateTargetFramework (".NETStandard", "2.0");
+
```
mrward: Not sure this if statement is needed. It looks like we will return .NET Standard 2.0 twice from this method since the if below this one is similar.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkInstalledCondition.cs
```diff
@@ -78,42 +74,33 @@ static bool SdkVersionSupported (NodeElement conditionNode, params DotNetCoreVer
...
-			// Special case '2.2', '2.1' and '2.0'.
-			if (requiredSdkversion == "2.2") {
-				return versions.Any (IsNetCoreSdk22);
-			} else if (requiredSdkversion == "2.1") {
+			requiredSdkversion = requiredSdkversion.Replace ("*", string.Empty);
+			DotNetCoreVersion requiredDotNetCoreVersion;
+			DotNetCoreVersion.TryParse (requiredSdkversion, out requiredDotNetCoreVersion);
+
+			if (requiredDotNetCoreVersion == null)
+				return versions.Any (version => version.ToString ().StartsWith (requiredSdkversion, StringComparison.OrdinalIgnoreCase));
```
mrward: This does not look right for .NET Core 2.1 sdks since it has some odd numbering around 2.1.300 and before this. This will match 2.1 when the sdk does not support 2.1. For example .NET Core 2.1.200 SDK only supports .NET Core 2.0. The IsNetCoreSdk21 and IsNetCoreSdk20 methods take into account the patch version when matching.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
```diff
@@ -156,5 +185,36 @@ IEnumerable<DotNetCoreVersion> GetInstalledSdkVersions (string sdkRootPath)
...
+				return version;
+			}
+		}
+
+		internal IEnumerable<DotNetCoreVersion> GetInstalledSdkVersions ()
+		{
+			return Directory.EnumerateDirectories (SdkRootPath)
+				.Select (directory => DotNetCoreVersion.GetDotNetCoreVersionFromDirectory (directory))
+				.Where (version => version != null);
+		}
```
mrward: This code is basically the same as GetInstalledSdkVersions that takes the sdkRootPath. Can this just call the other method?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -53,6 +54,17 @@ public DotNetCoreProjectExtension ()
 			DotNetCoreProjectReloadMonitor.Initialize ();
 		}
 
+		void FileService_FileChanged (object sender, FileEventArgs e)
+		{
+			var globalJson = e.FirstOrDefault (x => x.FileName.ToString ().Contains ("global.json") && !x.FileName.IsDirectory);
+			if (globalJson == null)
```
rodrmoya: Shouldn't this check for the filename == "global.json" rather than "*global.json*" as it seems to do now?

jtorres: filename contains the full path, not the file name

rodrmoya: right so this would match a file in /temp/global.json/peg.cs right? But you only want to check for files called global.json right?

rodrmoya: That is, it should be checking for Path.GetFileName (x.Filename) == “global.json”

rodrmoya: Was just an example path :p What I mean is that if there is a file, inside the solution, called `file_global.json_whatever`, this would catch it, but that's not what you want, which is to get files called `global.json`, right?

jtorres: we also would like to detect that cases, (i.e. file's name has changed from global.json to global.json.old) but it's true that if the rename is something like "global1.json" then this won't detect that change... so I will take a look

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -222,14 +234,12 @@ Task ShowDotNetCoreNotInstalledDialog (bool unsupportedSdkVersion)
...
+					
 				Project.ParentSolution.ExtendedProperties [ShownDotNetCoreSdkInstalledExtendedPropertyName] = "true";
 
 				using (var dialog = new DotNetCoreNotInstalledDialog ()) {
 					dialog.IsUnsupportedVersion = unsupportedSdkVersion;
-					dialog.RequiresDotNetCore22 = Project.TargetFramework.IsNetCoreApp22 ();
-					dialog.RequiresDotNetCore21 = Project.TargetFramework.IsNetCoreApp21 ();
-					dialog.RequiresDotNetCore20 = Project.TargetFramework.IsNetStandard20OrNetCore20 ();
+					dialog.RequiredDotNetCoreVersion = DotNetCoreVersion.Parse (Project.TargetFramework.Id.Version);
 					dialog.Show ();
```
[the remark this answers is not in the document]
jtorres: .net standard 2.1 is still not supported here, on the other hand is true that the message for .net standard may not be clear, the behaviour however is pretty much the same than in master. So I will leave a message and will make message more clear.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -302,11 +313,45 @@ protected override void OnItemReady ()
...
+
+		void ReevaluateAllOpenDotNetCoreProjects ()
+		{
+			if (!IdeApp.Workspace.IsOpen)
+				return;
+				
+			foreach (var project in IdeApp.Workspace.GetAllItems<DotNetProject> ()) {
+				if (project.HasFlavor<DotNetCoreProjectExtension> ()) {
+					RestorePackagesInProjectHandler.Run (project, restoreTransitiveProjectReferences: true, reevaluateBeforeRestore: true);
+				}
```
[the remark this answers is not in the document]
mrward: I think we would need to add a new class to NuGet to make this easy to do from the DotNetCoreProjectExtension. I can add that. We can fix this later on in another pull request.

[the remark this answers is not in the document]
jtorres: cool thanks, then I will leave it as it is right now, will rename the method (as you mentioned) in another PR...

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -53,6 +54,20 @@ public DotNetCoreProjectExtension ()
...
 
+		void FileService_FileChanged (object sender, FileEventArgs e)
+		{
+			var globalJson = e.FirstOrDefault (x => x.FileName.ToString ().Contains ("global.json") && !x.FileName.IsDirectory);
+			if (globalJson == null)
+				return;
+
+			if (Path.GetFileName (globalJson.FileName).IndexOf ("global.json", StringComparison.OrdinalIgnoreCase) == 0)
+				Project.ParentSolution.ExtendedProperties [GlobalJsonPathExtendedPropertyName] = globalJson.FileName.ToString ();
+			else //i.e. global.json has been renamed
```
mrward: Missed this before. Not sure I understand what this is going. FileService.FileChanged occurs when the contents of the file has changed. So I think we can just call DetectSDK. Not sure I understand why the ExtendedProperties is being used here.

I could see maybe ExtendedProperties being used to check if the global.json file changed matches the value stored in ExtendedProperties.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -335,11 +352,46 @@ protected override void OnItemReady ()
...
+			var globalJsonPath = Path.Combine (Project.ParentSolution.BaseDirectory, "global.json");
+			if (File.Exists (globalJsonPath))
+				Project.ParentSolution.ExtendedProperties [GlobalJsonPathExtendedPropertyName] = globalJsonPath;
+			DetectSDK ();
+		}
+
+		void DetectSDK (bool restore = false)
+		{
+			sdkPaths.ResolveSDK (DotNetCoreVersion.Parse (Project.TargetFramework.Id.Version),  Project.ParentSolution.BaseDirectory);
+			DotNetCoreSdk.Update (sdkPaths);
```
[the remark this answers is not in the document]
mrward: Can we get the `Project.ParentSolution.ExtendedProperties [GlobalJsonPathExtendedPropertyName]` property once and then re-use that value instead of getting it twice and also checking the ExtendedProperties contains that property.

Just wondering if we can just do:

```
var globalJsonPathProperty = Project.ParentSolution.ExtendedProperties [GlobalJsonPathExtendedPropertyName] as string;
if (globalJsonPathProperty != null && File.Exists (globalJsonPathProperty)) {
   ....
```

Reading the documentation I think IDictionary will return null if the item does not exist but needs to be checked.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
```diff
@@ -156,5 +194,41 @@ IEnumerable<DotNetCoreVersion> GetInstalledSdkVersions (string sdkRootPath)
...
+			return Path.Combine (SdksParentDirectory, "Sdks");
+		}
+
+		string ReadGlobalJson (string workingDir)
+		{
+			if (string.IsNullOrEmpty (workingDir))
+				return string.Empty;
+
+			var globalJsonPath = new DirectoryInfo (workingDir).GetFiles ("global.json", SearchOption.AllDirectories).FirstOrDefault ();
+			if (globalJsonPath == null)
```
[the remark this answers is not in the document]
mrward: Logic looks good now. Here is how NuGet does this for comparison - https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs#L663-L709 - although it returns all NuGet.Config files it finds instead of just stopping at the first one it finds. NuGet gets the directory, then does a File.Exists. It also supports different filename case - not sure if we need to do that for global.json files on case sensitive file systems.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
mrward: Not sure the code does not look right to me. FileChanged is for file changes not for file renames. There is a FileService.FileRenamed event. For now maybe we should just ignore the file being renamed and just reload on file changes.

Something else here which maybe we should look at later is that if you have say 3 .NET Core projects and global.json changes we will run this code 3 times and run a restore for all projects 3 times. Seems like it should be run once if we are restoring all projects.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -53,6 +54,15 @@ public DotNetCoreProjectExtension ()
 			DotNetCoreProjectReloadMonitor.Initialize ();
 		}
 
+		void FileService_FileChanged (object sender, FileEventArgs e)
+		{
+			var globalJson = e.FirstOrDefault (x => x.FileName.FileName.IndexOf ("global.json", StringComparison.OrdinalIgnoreCase) == 0 && !x.FileName.IsDirectory);
+			if (globalJson == null)
```
mrward: I think the code should check `globalJson == Project.ParentSolution.ExtendedProperties [GlobalJsonPathExtendedPropertyName]` before running DetectSDK.

PR #6753, opened 2018-12-11
Use an exclusive scheduler for FileWatchers

Therzok:
This should fix file watcher test failure.


slluis:

PR #6759, opened 2018-12-12
Fixes VSTS Bug 731943: [Feedback] xUnit Fact DisplayName not shown in

mkrueger:
test explorer if string has a period at the end.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/731943


slluis:
@mkrueger can you please rebase to latest master, so that we can get a green build?

PR #6773, opened 2018-12-14
[release-7.7][NuGet] Fix NuGet sdk resolver

mrward:
Opening a project that required the NuGet sdk resolver would
fail with an error about failing to load a NuGet assembly.

    Could not load file or assembly 'NuGet.Commands, Version=4.8.0.1,
    Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its
    dependencies.

The MSBUILD_NUGET_PATH environment variable is not supported in
a more recent NuGet sdk resolver, after it moved out of MSBuild
source code repository into NuGet's source code repository.
Before the NuGet 4.8 update the NuGet sdk resolver would find
the NuGet assemblies that it required, and were not already
loaded by the IDE, in its working directory. Latest NuGet sdk
resolvers are now included in the msbuild bin directory next to
the NuGet assemblies. Now after updating to NuGet 4.8 the sdk
resolver would fail to load the NuGet assemblies. To fix this
the NuGet addin now loads the NuGet assemblies with an import
so the NuGet sdk resolver can find them.

Fixes VSTS #751992 - NuGet sdk resolver fails to find assembly


mrward:
@monojenkins backport master

PR #6777, opened 2018-12-14
[Debugger] Remove Gtk use from the New Breakpoint dialog

iainx:
Replace the Gtk specific code with Xwt equivalents

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/TextEntryWithCodeCompletion.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/TextEntryWithCodeCompletion.cs
@@ -34,21 +34,18 @@ namespace MonoDevelop.Debugger
 	class TextEntryWithCodeCompletion : TextEntry, ICompletionWidget
 	{
 		CodeCompletionContext ctx;
-		Gtk.Entry gtkEntry;
 
-		Gdk.ModifierType modifier;
+		Xwt.ModifierKeys modifier;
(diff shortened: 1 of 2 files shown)


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/TextEntryWithCodeCompletion.cs
```diff
@@ -241,7 +246,7 @@ public void SetCompletionText (CodeCompletionContext ctx, string partial_word, s
 
 		public Gtk.Style GtkStyle {
```
Therzok: Still Gtk here.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/TextEntryWithCodeCompletion.cs
```diff
@@ -65,32 +62,40 @@ void HandleWindowClosed (object sender, EventArgs e)
 				CompletionContextChanged (this, EventArgs.Empty);
 		}
 
+		char CharFromKey (Xwt.Key xwtKey)
```
Therzok: Could be static.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/TextEntryWithCodeCompletion.cs
iainx: It's a public interface member.

PR #6778, opened 2018-12-15
Disable VB.NET templates for .NET Core

rodrmoya:


rodrmoya:
NOTE: should find a better condition to disable VB.NET support I guess?

mhutch:
I wonder if this should done in a more generic way, e.g. `<Condition id="FeatureSwitch" name="VBNetDotnetCoreTemplates" />` and a `MD_FEATURES_ENABLED="Feature1;Feature2"` env var.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/GlobalFeatures.cs
```diff
@@ -0,0 +1,52 @@
...
+using System;
+using Mono.Addins;
+
+namespace MonoDevelop.Core.FeatureConfiguration
+{
+	public static class GlobalFeatures
+	{
+		public static class Languages
+		{
+			public static FeatureSwitch VisualBasicSupport = FeatureSwitchService.RegisterFeature (
```
mhutch: This is *very* misleadingly named, as it doesn't control VB.NET support at all, it just controls the .NET Core templates

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/GlobalFeatures.cs
```diff
@@ -0,0 +1,52 @@
...
+		}
+	}
+
+	internal class VisualBasicSupportCondition : ConditionType
+	{
+		// FIXME!!
+		public override bool Evaluate (NodeElement conditionNode)
+		{
+			return Environment.GetEnvironmentVariables ().Contains ("MD_ENABLE_VISUALBASIC_SUPPORT")
+				|| BrandingService.ApplicationName.Contains ("MonoDevelop");
```
mhutch: I don't think it's necessary to keep these templates enabled for MD. Also, the semantics of this are weird - I think the intention is "disable for VSMac" but it reads as "disable for any rebranded instance".

Therzok: Maybe we can implement this extension point as returning false in Xamarin.Ide.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/GlobalFeatures.cs
rodrmoya: right, indeed

rodrmoya:
Changed it to be more generic and to explicitly disable it in md-addins -> https://github.com/xamarin/md-addins/pull/3988

main/src/addins/MonoDevelop.DotNetCore/Properties/MonoDevelop.DotNetCore.addin.xml
```diff
@@ -95,18 +95,20 @@
...
-			_overrideName="Console Application"
-			_overrideDescription="Creates a new .NET Core console project."
-			overrideLanguage="VBNet"
-			path="${DotNetCoreSdk.2.1.Templates.Common.ProjectTemplates.nupkg}"
-			icon="md-netcore-console-project"
-			imageId="md-netcore-console-project"
-			wizard="MonoDevelop.DotNetCore.ProjectTemplateWizard"
-			condition="UseNetCore21=true"
-			category="netcore/app/general"/>
+		<Condition id="FeatureSwitchCondition" name="VBNetDotnetCoreTemplates">
```
mhutch: IMO "Condition" is redundant in `FeatureSwitchCondition`, it could just be `FeatureSwitch`

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/GlobalFeatures.cs
```diff
@@ -0,0 +1,53 @@
...
+	internal class FeatureSwitchCondition : ConditionType
+	{
+		public override bool Evaluate (NodeElement conditionNode)
+		{
+			var featureName = conditionNode.GetAttribute ("name");
+			if (String.IsNullOrEmpty (featureName)) {
+				return true;
+			}
+
+			if (Environment.GetEnvironmentVariable ("MD_FEATURES_ENABLED")
```
mhutch: Wonder whether it's worth caching the split string in a field

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/IFeatureSwitchCheck.cs
```diff
@@ -0,0 +1,34 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+
+namespace MonoDevelop.Core.FeatureConfiguration
+{
+	public interface IFeatureSwitchCheck
```
mhutch: Not a huge fan of this name, it seems to be defined from the point of view of the consumer, and doesn't make the purpose super obvious to me. `FeatureSwitchEnabler`, maybe?

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/IFeatureSwitchCheck.cs
```diff
@@ -0,0 +1,34 @@
...
+	{
+		bool IsFeatureEnabled (string featureName);
```
mhutch: I wonder whether this should return a `bool?` so that these can both enable _and_ disable features.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/GlobalFeatures.cs
```diff
@@ -0,0 +1,53 @@
...
+				.Split (';')
+				.Contains (featureName)) {
+				return true;
+			}
+
+			// Fallback to ask extensions, enabling by default
+			return !(AddinManager.GetExtensionObjects<IFeatureSwitchCheck> ("MonoDevelop.Core.FeatureConfiguration.IFeatureSwitchCheck")?
+				.Any (x => !x.IsFeatureEnabled (featureName)) ?? true);
```
mhutch: It might make sense to make the default value (true/false) a per-feature thing. This would allow features to be on by default, and the extension could disable them

rodrmoya:
I'm not too happy with this implementation indeed, as I was trying to unify what we use for disabling/enabling Build Output (https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs#L47) with this, but because of the `NodeElement` argument in `ConditionType`'s `Evaluate`, can't think of a good way of unifying it. Maybe just make everything use the `FeatureSwitch` as BuildOutput does, and, on top of that, write the `ConditionType`-based class?

Or maybe use the `FeatureSwitchService` to register/query each individual feature toggle?

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/GlobalFeatures.cs
rodrmoya: Well, that's what it does, right? It only disables the feature if one of the IFeatureSwitchCheck implementations returns false for that feature.

rodrmoya:
I've started over and simplified the FeatureSwitchService implementation to use the IFeatureSwitchEnabler extension, so that all code is now unified and much simpler

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/GlobalFeatures.cs
rodrmoya: for normal situation (i.e. the IDE) yes, but for tests we set/unset the env vars for testing different situations, so I guess we should be splitting it always

mhutch:
Couple thoughts - since feature switches can now be enabled by default, maybe `IFeatureSwitchController` would be better? Also, having `IFeatureSwitchController.IsFeatureEnabled` return a `bool?` where null means "don't care" would be cleaner than passing in the default value as a parameter.

rodrmoya:
@mhutch done! :)

slluis (approved):

Therzok (approved):

rodrmoya:

rodrmoya:

rodrmoya:
Bug was marked for 7.7, but we don't have the VB.NET templates there, so no need to backport to 7.7

PR #6780, opened 2018-12-17
[A11y] Make the headers on the error pad accessible from the keyboard

iainx:
Normally in Gtk if you focus a table and shift-tab the focus goes to the headers
but in the Errors pad, because the focus chain for the pad is set to be the
logView, the shift-tab focus breaks.

unsetting the focus chain fixes the header navigation issue, but stops the
logView being automatically focused. I have tried about 5 different ways to
make the logView still be the default focus, but none of them worked.
Discussing it with Sevo, we decided that making the logView the default focus
didn't really make sense anyway, so I've removed that.

Fixes VSTS #515118


iainx:
I added it years ago to fix a bug, it's old code, but the bug it fixed wasn't very useful, and @sevoku and I don't think it's the right thing to do anyway

PR #6781, opened 2018-12-17
[C#] Fix NRE in completion querying

Therzok:
The semantic model might not contain a valid symbol at a given
syntax node which can happen in code that's not compilable.

Prevent that by adding a nullcheck

Fixes VSTS #714055 - AddImportCompletionData null reference exception


Therzok:

Therzok:

Therzok:
Not sure how to crop up a test case.

Therzok:
@slluis needs merged

PR #6785, opened 2018-12-18
[MEF] Add metadata about referenced assemblies into the MEF cache

Therzok:
All assemblies now register their referenced assemblies into the composition cache file
(with the exception of assemblies which cannot be found, which will be reported
anyway by the MEF discovery error reporting).

With this, we can now invalidate the cache on assembly dependency changes and
on GAC changes (i.e. mono upgrade)

Fixes VSTS #734010 - MEF cache should invalidate on framework or dependency changes

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
@@ -132,11 +132,11 @@ namespace MonoDevelop.Ide.Composition
 				var fullTimer = System.Diagnostics.Stopwatch.StartNew ();
 				var stepTimer = System.Diagnostics.Stopwatch.StartNew ();
 
-				var assemblies = ReadAssembliesFromAddins (timer);
+				var (mefAssemblies, allAssemblies) = ReadAssembliesFromAddins (timer);
 				timings ["ReadFromAddins"] = stepTimer.ElapsedMilliseconds;
 				stepTimer.Restart ();
 
-				var caching = new Caching (assemblies);
+				var caching = new Caching (mefAssemblies, allAssemblies);
 
 				// Try to use cached MEF data
 
@@ -169,7 +169,7 @@ namespace MonoDevelop.Ide.Composition
 
 		internal static async Task<RuntimeComposition> TryCreateRuntimeCompositionFromCache (Caching caching)
 		{
-			CachedComposition cacheManager = new CachedComposition ();
+			var cacheManager = new CachedComposition ();
 
 			try {
 				using (var cacheStream = caching.OpenCacheStream ()) {
@@ -184,7 +184,7 @@ namespace MonoDevelop.Ide.Composition
 
 		internal static async Task<RuntimeComposition> CreateRuntimeCompositionFromDiscovery (Caching caching, ITimeTracker timer = null)
 		{
-			var parts = await Discovery.CreatePartsAsync (caching.Assemblies);
+			var parts = await Discovery.CreatePartsAsync (caching.MefAssemblies);
 			timer?.Trace ("Composition parts discovered");
 
 			ComposableCatalog catalog = ComposableCatalog.Create (StandardResolver)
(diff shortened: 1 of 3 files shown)


Therzok:
Interestingly enough, with more logging added, we have this now:
```
INFO [2018-12-18 03:21:03Z]: Could not find dependency PresentationCore, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, possibly missing Runtime import
INFO [2018-12-18 03:21:03Z]: Could not find dependency StreamJsonRpc, Version=1.3.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, possibly missing Runtime import
INFO [2018-12-18 03:21:03Z]: Could not find dependency Microsoft.VisualStudio.TestPlatform.ObjectModel, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, possibly missing Runtime import
INFO [2018-12-18 03:21:03Z]: Could not find dependency Microsoft.TestPlatform.CommunicationUtilities, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, possibly missing Runtime import
INFO [2018-12-18 03:21:03Z]: Could not find dependency NuGet.ProjectModel, Version=4.8.0.6, Culture=neutral, PublicKeyToken=31bf3856ad364e35, possibly missing Runtime import
INFO [2018-12-18 03:21:03Z]: Could not find dependency NuGet.ProjectModel, Version=4.8.0.6, Culture=neutral, PublicKeyToken=31bf3856ad364e35, possibly missing Runtime import
INFO [2018-12-18 03:21:03Z]: Could not find dependency NuGet.ProjectModel, Version=4.8.0.6, Culture=neutral, PublicKeyToken=31bf3856ad364e35, possibly missing Runtime import
INFO [2018-12-18 03:21:03Z]: Could not find dependency NuGet.DependencyResolver.Core, Version=4.8.0.6, Culture=neutral, PublicKeyToken=31bf3856ad364e35, possibly missing Runtime import
INFO [2018-12-18 03:21:03Z]: Could not find dependency NuGet.Resolver, Version=4.8.0.6, Culture=neutral, PublicKeyToken=31bf3856ad364e35, possibly missing Runtime import
INFO [2018-12-18 03:21:03Z]: Could not find dependency NuGet.DependencyResolver.Core, Version=4.8.0.6, Culture=neutral, PublicKeyToken=31bf3856ad364e35, possibly missing Runtime import
INFO [2018-12-18 03:21:03Z]: Could not find dependency NuGet.Indexing, Version=4.8.0.6, Culture=neutral, PublicKeyToken=31bf3856ad364e35, possibly missing Runtime import
INFO [2018-12-18 03:21:03Z]: Could not find dependency NuGet.Resolver, Version=4.8.0.6, Culture=neutral, PublicKeyToken=31bf3856ad364e35, possibly missing Runtime import
INFO [2018-12-18 03:21:03Z]: Could not find dependency Microsoft.TestPlatform.VsTestConsole.TranslationLayer, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, possibly missing Runtime import
```

versus the two assemblies for xammac load
```
Could not find `PresentationCore` referenced by assembly `Microsoft.VisualStudio.Language.Intellisense, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a`.
Could not find `StreamJsonRpc` referenced by assembly `Microsoft.VisualStudio.Language, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a`.
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
```diff
@@ -223,18 +226,44 @@ void ReadAssemblies (HashSet<Assembly> assemblies, string extensionPath, ITimeTr
...
+					GatherDependencies (asm);
+					timer.Trace ("End dependencies: " + asm.FullName);
+				}
+
+				void GatherDependencies (Assembly currentAssembly)
+				{
+					foreach (var dependencyName in currentAssembly.GetReferencedAssemblies ()) {
+						try {
+							// Lookup loaded assembly or load GAC one.
+							var dependencyAssembly = Assembly.Load (dependencyName);
```
KirillOsenkov: Not sure it should be using Assembly.Load. I think we try to use our wrappers like Runtime.LoadAssemblyFrom because there are subtle differences between .NET Framework on Windows and Mono.

KirillOsenkov:
Not sure we should be loading the assemblies explicitly (this changes runtime behavior). Some dependency assemblies are not actually required at runtime (the code paths we're exercising are never triggering loading them). Now we're forcefully bringing them in. That is probably explaining the difference in log messages you're seeing - now we're loading more than we'd naturally load during normal execution. Some of this may even break, because for example one of the VS Auth .dlls depends on WindowsBase but it's not actually required for our usecase, so it's never loaded at runtime.

I wish the runtime had an API like "tell me where you would have loaded this assembly from, but don't actually load it". Then we could just get the actual location and grab the MVID from there without loading.

Maybe for now we should optimistically check the MVIDs only for those assemblies that are already loaded into the appdomain at the time MEF cache is running? This would cover most runtime assemblies, which we care about the most.

And, keep the list of all MVIDs we expect, then install an AppDomain.CurrentDomain.AssemblyLoaded handler, and when an assembly loads, check it's MVID against the list we have, and if it doesn't match, show a good error message: "Please restart MonoDevelop because the MEF cache is out-of-date" and invalidate the cache.

I don't know anything about IAssemblyLoader but perhaps you're right.

KirillOsenkov:
Here’s another idea: remember the absolute path on disk where each assembly was loaded from when the cache got created. This way we can check if the MVID changed, and if any of the files is absent, invalidate the entire cache.

Sceptical about ReflectionOnlyLoadFrom because of added perf overhead.

KirillOsenkov:
Here’s how to read the MVID very quickly:
https://github.com/KirillOsenkov/MetadataTools/blob/master/src/PEFile/MvidBenchmark.cs

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
mhutch: Yes but it's a opt in "strict mode". The runtime folks have been asking us to try it.

Even so, we'd want to use the wrappers, as they ensure we don't have issue mixing the Load and LoadFrom contexts. But we'd need to update the wrappers to use the .NET codepath with Mono strict mode.

Therzok:
Will take a look at make it so we only consider runtime assemblies and add an assemblyload hook to validate the MEF cache.

Therzok:
@KirillOsenkov should be ready for another review pass.

PR #6789, opened 2018-12-18
[DesignerSupport] Fix Crasher when disposing the Toolbox

sevoku:
If something fails during toolbox initialization, the event unsubscription might also fail. However it's not necessary to unsubscribe events, we should simply dispose the widgets being used.

@Therzok I'm not sure about all the event handlers, but it might still leak, could you take a quick look if you have time?

Fixes VSTS #753355


slluis:
I think the events still have to be unsubscribed, even if the widget is disposed.

Therzok:
Can we preventively keep event handler removal? Ref-cycles are easy to introduce and hard to get rid of.

Therzok:
Also, disposal is not the same as destruction. You need to `Destroy()` to release all managed refs.

sevoku:
@monojenkins backport release-8.0-p1.1

Therzok:
Way better 👍

netonjm:
+1 here

Descarga Outlook para iOS<https://aka.ms/o0ukef>
________________________________

PR #6794, opened 2018-12-19
Use SQLitePCL.raw 1.1.12 to bring in SQLite 3.26.0

duncanmak:
This is to bring in the Magellan fix that was disclosed recently.

I noticed that there's a mix of target frameworks in the `packages.config`.

```
$ grep net461 packages.config  | wc -l
      96
$ grep net471 packages.config  | wc -l
      13
```

This patch moved the SQLitePCL.raw references to net471, because that's what `MDFrameworkVersion` is set to since https://github.com/mono/monodevelop/commit/975d3bf5edca958b55520cefafc9a9cadc383083.


Therzok:
@duncanmak yes, there's a mix, because IDE nuget support does not properly support our way of referencing nugets with imported targets. We need to upgrade to PackageReference eventually to fix that.

Therzok (approved):
:+1: from me if it goes green

Therzok:
Remaining failure is unrelated to this PR. Can be merged.

KirillOsenkov (approved):

PR #6795, opened 2018-12-19
[Nuget] Add import NuGet.common.dll

sgmunn:
Xamarin.Ide uses this assembly and may need it to be registered. When running some tests via `mdtool` I have seen a failure to load the ide addin because this was not found.

### DIFF
--- a/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement.addin.xml
+++ b/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement.addin.xml
@@ -8,6 +8,7 @@
 		<Import assembly="NuGet.Versioning.dll" />
 		<Import assembly="NuGet.Commands.dll" />
 		<Import assembly="NuGet.LibraryModel.dll" />
+		<Import assembly="NuGet.Common.dll" />
 	</Runtime>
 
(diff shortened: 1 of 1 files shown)


Therzok:
```
➜  MonoDevelop.PackageManagement git:(update-sqlite-nuget) ✗ ls *.dll
Lucene.Net.dll                    NuGet.Credentials.dll             NuGet.Packaging.Core.dll
Microsoft.Web.XmlTransform.dll    NuGet.DependencyResolver.Core.dll NuGet.Packaging.dll
MonoDevelop.PackageManagement.dll NuGet.Frameworks.dll              NuGet.ProjectModel.dll
NuGet.Commands.dll                NuGet.Indexing.dll                NuGet.Protocol.dll
NuGet.Common.dll                  NuGet.LibraryModel.dll            NuGet.Resolver.dll
NuGet.Configuration.dll           NuGet.PackageManagement.dll       NuGet.Versioning.dll
```

Looks like there should be more in there.

mrward:
For the other .dlls it depends on whether they are used elsewhere. Some of them are unlikely to be used by other addins. Ideally they would not be used elsewhere since they typically break the API on every release.

sgmunn:
compat check is failing with same error on master, not this PR

mrward:

PR #6801, opened 2018-12-21
[Ide] Add IVT to templates tests

sgmunn:
Needed to split out forms template tests


sgmunn:
Actually, I don't think we don't need this. Later we might want to refactor template tests to a common lib, but for now I think we can ignore this.

PR #6805, opened 2018-12-29
Add IDE startup preferences

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/729583

### DIFF
--- a/main/src/core/MonoDevelop.Ide/Gui/MonoDevelop.Ide.Gui.OptionPanels.LoadSavePanelWidget.cs
+++ b/main/src/core/MonoDevelop.Ide/Gui/MonoDevelop.Ide.Gui.OptionPanels.LoadSavePanelWidget.cs
@@ -14,8 +14,8 @@ namespace MonoDevelop.Ide.Gui.OptionPanels
 		private global::Gtk.HBox hbox10;
 		private global::Gtk.VBox vbox65;
 		private global::Gtk.CheckButton loadUserDataCheckButton;
-		private global::Gtk.CheckButton openStartWindowCheckButton;
-		private global::Gtk.CheckButton loadPrevProjectCheckButton;
+		private global::Gtk.RadioButton openStartWindowRadioButton;
+		private global::Gtk.RadioButton loadPrevProjectRadioButton;
 		private global::Gtk.VBox vbox19;
 		private global::Gtk.Label saveLabel;
 		private global::Gtk.HBox hbox11;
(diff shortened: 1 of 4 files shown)


Therzok:
In the case where we don't want to show the start window, shoudln't we actually be showing the IDE with an open file dialog?

rodrmoya:
Last commit moves the controls to be grouped radio buttons, so if the openStartWindow option is disabled, it would open the last solution

slluis:
Can you please attach a screenshot for UI review?

rodrmoya:
@slluis this just adds a radio button to the Load/Save options panel:

screenshot 2019-01-07 at 16 53 05

slluis:
I don't understand the change. The options are the same as before: show the start window or load the last solution.

rodrmoya:
hmm, no, I added the "show the start window", 8.0 has this:
image

slluis:
What I mean is that in current master the options are the same: if the "Load previous solution in startup" checkbox is unchecked, it will show the start window. If it is checked, it will load the previous solution. So this PR is just a UI change, which changes a checkbox for two radio buttons, and I don't think the new UI is better. /cc @vancura

rodrmoya:
pinging @hbons who did the design and @asb3993 for the spec side

vancura:
The results of this PR makes the UI more difficult to understand and use than before. We don't need this I am afraid.

asb3993:
The idea behind this was to give users an option on exactly what they want to see on start, similar to how Visual Studio or a browser gives options:

screen shot 2019-01-07 at 11 52 19 am
screen shot 2019-01-07 at 11 33 00 am

The original design (I thought this was attached to the ticket, but can't seem to see it) includes more than two options: 
card-on_startup

The radio buttons only make sense if we have > 2 options, so I'm in agreement with Lluis that as this PR stands, it doesn't add anything to the UI. As per the spec we had the option to "option a specific solution" too, can we add this? Also in the design above, Hylke had added a "Do nothing", which is similar to VS "Empty Environment". @slluis or @rodrmoya do you think there is any value in showing that?

Therzok:
My $0.02 on this:

1. We need a label to separate the ratio buttons from the checkbox above it if we stick with this.
2. `Empty Environment` is the equivalent of showing an empty workspace. This might be a scenario we need to support, given that we have no global search in GTC mode. It would enable the old scenario of - open IDE and drag and drop a solution file or just use one of the built-in tools - i.e. RegexToolkit

rodrmoya:
TBH I don’t see any point in the “open a specific solution” option, which is why I didn’t add it since wanted to discuss it but everyone was on vacation. On the other hand, what @Therzok mentions does indeed make sense to me.

asb3993:
I can see the scenario where a user might tend to use the the same solution 99% of the time and just want to start that up every time they open the IDE. This might not always be the _last_ solution that they had opened though, so it's just a way to allow to use to use the IDE in the way they want.

Agree with @Therzok on

1. adding a label "On Start" as per the design above. Although we should probably make it consistent with the other labels and have them as "On Load"/"On Save"?
2. Adding "Empty Environment" as an empty workspace

hbons:
@asb3993 the "on load/save" options have been removed from the new Preferences design as they're not clear, so no need to change anything there.

hbons:
In any case, if we're not removing those load/save options for now, I agree with @asb3993 that we need a new "Startup" section for this.

rodrmoya:
This now looks like this:

image

Therzok:
@rodrmoya can we move Start before Load?

rodrmoya:
yes, we can: 
image

vancura:
Looks great, thanks @rodrmoya!

PR #6814, opened 2019-01-03
[Ide] Make InfoBar API return a Control

bratsche:
Right now there is no way to remove an InfoBar programmatically once it
has been added. This can be a problem for addins that show an InfoBar.
For example, iOS and Android addins each show an infobar in certain
situations. If Android addin shows the infobar and the developer closes
the solution without hiding the infobar, then loads a different solution
without an Android project the Android infobar will re-appear.

If we return a Control for the addin to reference then the addin can
dispose of it when it's no longer needed.


Therzok:
What is the behaviour of Dispose() with the mac backend? Does it do `RemoveFromSuperview` too?

bratsche:
@Therzok I was talking to @sgmunn and he thought this sounded like the wrong approach. Maybe we can just make it so when the solution is closed and we're returning to the welcome screen then all the existing InfoBars are removed?

PR #6820, opened 2019-01-07
This test did not remove the solution it added to the FileWatcherService

Therzok:
Fixes VSTS #761524 - Test failure: MonoDevelopMetadataReferenceManagerMetadataReferenceCacheTests.ReferenceCacheSnapshotUpdates)


mrward:
Solution derives from WorkspaceItem which calls FileWatcherService.Remove so the tests look OK to me without this change. The solution is disposed in the test tear down. Also all tests that add a new solution dispose it explicitly in the test itself.

Therzok:
:+1: The fix needs to be done elsewhere.

PR #6828, opened 2019-01-07
[PublishToFolder] Sorts out default destination folder

jtorres:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/740318
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/740321

### DIFF
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.Dialogs/PublishToFolderDialog.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.Dialogs/PublishToFolderDialog.cs
@@ -4,13 +4,50 @@ using MonoDevelop.AspNetCore.Commands;
 using MonoDevelop.Core;
 using MonoDevelop.DotNetCore;
 using MonoDevelop.Ide;
+using MonoDevelop.Projects;
 using Xwt;
 
 namespace MonoDevelop.AspNetCore.Dialogs
 {
+	internal class DefaultFolderResolver
+	{
+		internal const string DefaultConfiguration = "Release";
+
+		DotNetProject project;
+		public Uri BinBaseUri => new Uri (Path.Combine (project.BaseDirectory, "bin"));
+
+		public DefaultFolderResolver (DotNetProject project) => this.project = project;
+
+		// The default folder is: "bin/Release/<netcore-version>/publish"
+		// as long as Release exists. If it does not, then we take the active one.
+		public string GetDefaultFolder (UriKind uriKind)
+		{
+			//check if there is a Release configuration
+			bool releaseFound = false;
+			for (int i = 0; i < project.Configurations.Count; i++) {
+				if (project.Configurations [i].Name == DefaultConfiguration) {
+					releaseFound = true;
+					break;
+				}
+			}
+
+			var configuration = DefaultConfiguration;
+			if (!releaseFound) //if there is no Release config, then we take the active one
+				configuration = project.GetActiveConfiguration ();
+
+			var defaultDirectory = Path.Combine (BinBaseUri.ToString (), configuration,
+								project.TargetFramework.Id.GetShortFrameworkName (),
+								"publish");
+
+			if (uriKind == UriKind.Relative)
+				return BinBaseUri.MakeRelativeUri (new Uri (defaultDirectory)).ToString ();
+
+			return defaultDirectory;
+		}
+	}
+
 	class PublishToFolderDialog : Dialog
 	{
-		const string DefaultConfiguration = "Release";
 		VBox mainVBox;
 		Label publishYourAppLabel;
 		VBox browseVBox;
@@ -23,7 +60,7 @@ namespace MonoDevelop.AspNetCore.Dialogs
 		HBox messageBox;
 		Label messageLabel;
 		ImageView messageIcon;
-		Uri BinBaseUri => new Uri (Path.Combine (publishCommandItem.Project.BaseDirectory, "bin"));
+		DefaultFolderResolver defaultDirectoryResolver;
 
 		readonly PublishCommandItem publishCommandItem;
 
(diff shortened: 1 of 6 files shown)


main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.Dialogs/PublishToFolderDialog.cs
```diff
@@ -4,13 +4,50 @@
...
 
 namespace MonoDevelop.AspNetCore.Dialogs
 {
+	internal class DefaultFolderResolver
+	{
+		internal const string DefaultConfiguration = "Release";
+
+		DotNetProject project;
+		public Uri BinBaseUri => new Uri (Path.Combine (project.BaseDirectory, "bin"));
+
```
mrward: Is using the bin directory correct? If I choose a different OutputPath and use dotnet publish it will publish in that directory instead of bin/Release:

```
<Project Sdk="Microsoft.NET.Sdk.Web">

  <PropertyGroup>
    <TargetFramework>netcoreapp2.1</TargetFramework>
    <OutputPath>foobar</OutputPath>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.App" />
  </ItemGroup>

</Project>
```

    ProjectDirectory/foobar/netcoreapp2.1/publish/

Not sure what VS on Windows does here.

I think you can also choose to not include the target framework as part of the folder name.

You can also completely override the publish folder using `<PublishDir>test/somefolder</Publish>`

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.Tests/MonoDevelop.AspNetCore.Tests.csproj
```diff
@@ -1,33 +1,23 @@
...
-    <ConsolePause>false</ConsolePause>
+    <RARResolveFromOutputDirectory>true</RARResolveFromOutputDirectory>
   </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' " />
   <ItemGroup>
     <Reference Include="System" />
+    <Reference Include="System.Xml" />
+    <Reference Include="NuGet.Versioning">
+      <HintPath>..\..\..\..\packages\NuGet.Versioning.4.9.0-rtm.5658\lib\net46\NuGet.Versioning.dll</HintPath>
```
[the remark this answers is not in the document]
jtorres: not sure, I took the same version than MonoDevelop.DotNetCore.Test, look: https://github.com/mono/monodevelop/blob/master/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Tests/MonoDevelop.DotNetCore.Tests.csproj#L19

If so, we should downgrade both, shouldn't we?

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.Dialogs/PublishToFolderDialog.cs
sayedihashimi: @mrward we have the same behavior as VS (Windows), and use the same default path. Users can change the path in the publish profile (.pubxml file under Properties/PublishProfiles).

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.Tests/MonoDevelop.AspNetCore.Tests.csproj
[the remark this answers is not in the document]
mrward: Yeah we should downgrade both - looks like I missed that NuGet 4.9 reference. Not sure how that works since NuGet.Versioning 4.9 should not be restored.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.Dialogs/PublishToFolderDialog.cs
```diff
@@ -4,13 +4,51 @@
...
 using Xwt;
 
 namespace MonoDevelop.AspNetCore.Dialogs
 {
+	internal class DefaultFolderResolver
+	{
+		internal const string DefaultConfiguration = "Release";
+
+		string config = DefaultConfiguration;
+		DotNetProject project;
```
rodrmoya: just a small nitpick: why have `config` and `Configuration`? Can't there just be a `public string Configuration { get; private set; }`?

rodrmoya: Ditto for `DefaultConfiguration`, I guess you can just set `Configuration` to be `"Release"` by default?

jtorres: agreed about the first comment (and fixed), about the second one, I need that as `const` for comparing and testing.

rodrmoya:

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.Tests/MonoDevelop.AspNetCore.Tests.csproj
```diff
@@ -1,33 +1,23 @@
...
-    <ConsolePause>false</ConsolePause>
+    <RARResolveFromOutputDirectory>true</RARResolveFromOutputDirectory>
   </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' " />
   <ItemGroup>
     <Reference Include="System" />
+    <Reference Include="System.Xml" />
+    <Reference Include="NuGet.Versioning">
+      <HintPath>..\..\..\..\packages\NuGet.Versioning.4.8.0\lib\net46\NuGet.Versioning.dll</HintPath>
```
Therzok: Should this be local copied ❓

jtorres: is essentially the same package used in MD.DotNetCore.Test, so I supposed so... @mrward ??

mrward: From what I can remember, it was a while ago, I think the test directly uses a type from that assembly and does not work without it local copied.

mrward: NuGet.Versioning is being Runtime/Imported currently by the NuGet addin. It might be that the test was not using the ide tests base so avoiding the normal addin loading so a local copy was needed.

PR #6840, opened 2019-01-08
[Ide] Draw focus rect for ImageView

bratsche:
Because this is an app-paintable widget it doesn't draw a focus
rectangle as needed. There doesn't seem to be a themable color for
drawing focus rectangles so I'm choosing this style color just
because it seems to look the best.

Fixes VSTS #751980

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/ImageView.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/ImageView.cs
@@ -131,8 +131,8 @@ namespace MonoDevelop.Components
 
 					if (HasFocus) {
 						ctx.SetSourceColor (Styles.BaseSelectionBackgroundColor.ToCairoColor ());
-						ctx.LineWidth = 1.0;
-						ctx.Rectangle (x - Xpad, y - Ypad, image.Width + Xpad, Image.Height + Ypad);
+						ctx.LineWidth = 2.0;
+						ctx.RoundedRectangle (x, y, image.Width, Image.Height, 4.0);
 						ctx.Stroke ();
 					}
 				}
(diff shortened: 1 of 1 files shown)


slluis:
Can you please attach a screenshot of how it looks like?

bratsche:
screen shot 2019-01-09 at 11 56 17 am

bratsche:
screen shot 2019-01-09 at 11 57 05 am

Therzok:
It looks misaligned? Why the `-1` on height/width for the frame height/width?

bratsche:
screen shot 2019-01-09 at 8 59 13 pm

Here's an updated screenshot @Therzok

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/ImageView.cs
```diff
@@ -132,7 +132,7 @@ protected override bool OnExposeEvent (Gdk.EventExpose evnt)
 					if (HasFocus) {
 						ctx.SetSourceColor (Styles.BaseSelectionBackgroundColor.ToCairoColor ());
 						ctx.LineWidth = 1.0;
-						ctx.Rectangle (x, y, image.Width - 1, Image.Height - 1);
+						ctx.Rectangle (x - Xalign, y - Yalign, image.Width + Xalign, Image.Height + Yalign);
```
slluis: Using Xalign/Yalign here makes no sense, the alignment is already applied when calculating the x and y coordinates above.

Therzok:
@bratsche still looks misaligned even in the new screenshot.

vancura:
Can you try a thicker line? `2pt` or even `3pt`? Preferably with rounded corners, like what Quartz does.

bratsche:
screen shot 2019-01-10 at 10 01 21 am

@vancura

vancura:
Wow, that would be so good if it was not clipped from left and right :(

No way, please set it to the old value. Sorry!

bratsche:
Closed this because the correct solution was to just not use ImageView anymore, instead use InformationPopoverWidget.

Issue #6730, opened 2018-12-06
[8.0p1] C# error squiggles never go away


mhutch:
With 8.0 preview 1, I often see behavior where the error squiggles stop updating and never go away even after I fix the problem. The document is still being parsed - intellisense shows new symbols that I type - but the squiggles simply stop being updated.

I can fix this by closing the file and reopening it. I've been it at least half a dozen times in the past hour.

> VS bug [#754930](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/754930)

mkrueger:
I would need a log file ?  Can you reproduce it somehow ?  In which project ?

mhutch:
There's nothing in the log file 😞

xamarin-release-manager:
Fixed in version 8.0.0.1140 (master)

Pull Request #6842 merged by: Mike Krüger
Author: mono
Commit: 057c7c118936fb4e7cfd3bb28fea19404c2d0ee4 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.0.0.1154 (release-8.0)

Pull Request #6842 merged by: Mike Krüger
Author: mono
Commit: 057c7c118936fb4e7cfd3bb28fea19404c2d0ee4 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #6842 merged by: Mike Kr??ger
Author: mono
Commit: 057c7c118936fb4e7cfd3bb28fea19404c2d0ee4 (xamarin/monodevelop)

PR #6842, opened 2019-01-09
Fixes issue #6730 [8.0p1] C# error squiggles never go away A canceled

mkrueger:
operation could leave the editor in a half updated state. This should
prevent that from happening.
+ Fixed a bug where the quick task update was broken.


slluis:
The c# tests are failing.

PR #6845, opened 2019-01-09
[Ide] Fix focus problem when tabbing out of the IDE

Therzok:
The searchbar would indiscriminately trigger FocusOut events every
time the window was focused out. That would cause the focus to move
to the next responder when tabbing out.

Fixes VSTS #754535 - Focusing out/into VisualStudio changes the default focused element on the UI


Therzok:
@slluis ping

Therzok:

PR #6850, opened 2019-01-09
[Windows] Fix documents getting closed on save

Therzok:
File.Replace internally renames the destination file to a temporary file.

When that happens, the FileSystemWatcher then takes that event and makes
the document point to the temporary file.

When done, the File.Replace code will delete the temporary file, causing
the document to close.

Circumvent that by using different IO events, which are possibly slower
but maintain correctness

Fixes VSTS #718767 - File close when save

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core/FileService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core/FileService.cs
@@ -634,16 +634,31 @@ namespace MonoDevelop.Core
 
 		static void WindowsRename (string sourceFile, string destFile)
 		{
-			//Replace fails if the target file doesn't exist, so in that case try a simple move
+			string tmpPath = null;
+
+			// In case the file doesn't exist, we need to use Move
 			if (!File.Exists (destFile)) {
 				try {
 					File.Move (sourceFile, destFile);
 					return;
 				} catch {
 				}
+			} else {
+				tmpPath = Path.GetTempFileName ();
 			}
 
-			File.Replace (sourceFile, destFile, null);
+			// Do not use File.Replace as that generates weird FSW events which make the IDE think
+			// that the document should be monitoring the autosave file.
+			if (tmpPath != null) {
+				File.Copy (destFile, tmpPath, true);
+			}
+
+			File.Copy (sourceFile, destFile, true);
+
+			// If we reached this point, it's safe to assume we succesfully overwrote.
+			if (tmpPath != null) {
+				File.Delete (tmpPath);
+			}
 		}
 
 		static void UnixRename (string sourceFile, string destFile)


KirillOsenkov (approved):

KirillOsenkov:
This is good, however VS Editor also uses File.Replace in their Save code. I'm wondering if we need a different fix, such as wrapping the replace code in 

```
using (FileSystemWatcher.Suspend())
{
    File.Replace(...);
}
```
or something. 

Basically detect the events about the .tmp file and throw them away.

Therzok:
We do have freezing/thawing of events, but that still causes the problematic scenarios.

Therzok:
One other idea, which would also work in the general scenario, would be to make it so when the filewatcher events are unthawed, we process the events in order, and if we have a pattern like:
```
mv a a.tmp
mv b a
rm a.tmp
```

That is just a single `mv b -> a`

Therzok:
@mrward should we try to reduce events in the pattern described above?

Therzok:
Thinking about keeping it for freeze/thaw for now, then maybe we can add some timer code.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/FileService.cs
```diff
@@ -633,17 +633,32 @@ public static void SystemRename (string sourceFile, string destFile)
...
+			// that the document should be monitoring the autosave file.
+			if (tmpPath != null) {
+				File.Copy (destFile, tmpPath, true);
+			}
+
+			File.Copy (sourceFile, destFile, true);
+
+			// If we reached this point, it's safe to assume we succesfully overwrote.
+			if (tmpPath != null) {
+				File.Delete (tmpPath);
```
slluis: This should be in a `finally` clause.

Therzok: I'm reworking the PR to reduce FSW events, as the editor code uses File.Replace regardless and we'd hit this bug still.

slluis (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Core/FileService.cs
```diff
@@ -633,17 +633,32 @@ public static void SystemRename (string sourceFile, string destFile)
...
 				} catch {
 				}
+			} else {
+				tmpPath = Path.GetTempFileName ();
+			}
+
+			// Do not use File.Replace as that generates weird FSW events which make the IDE think
+			// that the document should be monitoring the autosave file.
+			if (tmpPath != null) {
+				File.Copy (destFile, tmpPath, true);
```
netonjm: this could throw an exception in case that the file can not be overwritten?

Therzok: This PR is going to be obsolete anyway, the new editor uses File.Replace.

Therzok:
Closing PR in favor of FileService rework.

Therzok:
https://github.com/mono/monodevelop/pull/7012/commits/306432d40959c552e16ea5bf8bab2265c3a5d184

PR #6854, opened 2019-01-09
[Roslyn] Try fixing flakey GC tests by using a more recent PinHelper …

Therzok:
…from mono


slluis:
I have to admit I have no clue what this is about

Therzok:
@slluis this is an imported test from Roslyn. It fails from time to time, due to mono's GC being conservative, not precise.

The changes in the PinHelper guarantee that the object _is_ collected now for tests on WeakReference.

Therzok:

PR #6855, opened 2019-01-09
[Ide] Cache the active document in the document tracking service

Therzok:
Fixes VSTS #753459 - WrapDocument called too much on solution load


Therzok:
Seems like something needs to be rebuilt against this version.

Therzok:
Opened a PR against xamarin designers, waiting on Unity folks to get an mpack from them too.

PR #6858, opened 2019-01-10
[WIP] Support ItemDefinitionGroup

mrward:
WIP to support [Item definitions](https://docs.microsoft.com/en-us/visualstudio/msbuild/item-definitions).

TODO:

 - [x] Item definition groups elements loaded
 - [x] Item definition item metadata evaluated
 - [x] Handle item defined multiple times
 - [x] Item definition metadata available to project items
    - [x] Classic project
    - [x] Sdk project - Properties window not showing data for Update items
      - Does not work as implemented. Current implementation requires the MSBuildPropertyGroupEvaluated on the MSBuildItem to have access to the root project - which is not always the case.
 - [x] Saving project should not add item definition metadata to project items (currently breaking some tests)
    - [x] Classic project
    - [x] Sdk project - this seems to be OK but there is no code explicitly added to handle this.
 - [ ] Project items need to be able to remove an item definition metadata property (i.e. add an empty element for that property)
 - [ ] Handle new files added to a project that have associated item definition metadata.
    - Currently new files do not necessarily have the item definition metadata so when they are added the Properties window shows the files do not have this data but on reloading the project they will have this data. On saving it seems that the item definition should be checked and empty elements added for properties that are not set. Also for wildcards we have [logic that removes update items and adds an include](https://github.com/mono/monodevelop/blob/3cc0a0901f8c703f26c4e0578c2bdc4cb378f607/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs#L4021-L4028) - which possibly could be handled by adding an empty element for the property instead.
   - [ ] Classic project
   - [ ] Sdk project

Fixes VSTS #731472 - Support ItemDefinitionGroup for generators

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
@@ -1267,7 +1267,6 @@ namespace MonoDevelop.Projects.MSBuild
 				var trueCond = conditionIsTrue && (string.IsNullOrEmpty (item.Condition) || SafeParseAndEvaluate (project, context, item.Condition));
 				if (trueCond) {
 					var it = CreateEvaluatedItem (context, project, project.Project, item, string.Empty);
-					((MSBuildPropertyGroupEvaluated)it.Metadata).ItemName = null;
 					project.EvaluatedItemDefinitions.Add (it);
 				}
 			}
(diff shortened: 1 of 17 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
```diff
@@ -1076,6 +1080,9 @@ void Evaluate (ProjectInfo project, MSBuildEvaluationContext context, MSBuildImp
 						it.IsImported = true;
 						project.EvaluatedItemsIgnoringCondition.Add (it);
 					}
+					foreach (var it in p.EvaluatedItemDefinitions) {
```
Therzok: `project.EvaluatedItemDefinitions.AddRange (p.EvaluatedItemDefinitions)`?

Therzok (approved):

Therzok:
LGTM until now!

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
mrward: Yeah this could be AddRange. I think I was influenced by the other foreach loops but they are only needed to set the IsImported flag.

mrward:
Replaced by branch rebased on master-vnext

PR #6859, opened 2019-01-10
Fixes VSTS Bug 752653: A11y_VS for Mac_Find in Files_Screen Reader:

mkrueger:
Voice Over focus is reading “Find in files” and “Replace in files”
buttons as radio buttons.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/752653

In fact these buttons are radio buttons - but UI wise from user point
of view they aren't.


mkrueger:

iainx:
I don't think the bug is a valid accessibility bug, because these are radio buttons and they are acting like radio buttons, therefore VoiceOver needs to describe them as radio buttons, which is it doing.

However, we think the dialog could use a redesign because it's not a very good dialog, so I'm not closing the bug, just removing it from the accessibility bugs list.

All that said though, this isn't the correct way to fix the issue, VoiceOver would still read these as being Radio Buttons, even if the description said Toggle Button.

PR #6866, opened 2019-01-10
[Ide] Fixed InformationPopoverWidget and EventBoxTooltip to allow Markup.

jstedfast:
This is needed for Android and iOS addins which expect to be able
to use markup in tooltips.


jstedfast:
Related: https://github.com/xamarin/md-addins/pull/4043

jstedfast:
I've updated the PR to have a bool `UseMarkup` property so that existing usage doesn't break

PR #6880, opened 2019-01-14
[Ide] Fix accessibility label for target framework in build options

mrward:
Target framework combo box had no associated accessiblity label.

The options page uses a combo box with an enabled text entry. This
does not seem to be supported with the AtkCocoa library. Tabbing into
the control wants to use the accessibility information for the text
entry where as using Voice Over arrow keys wants to use the
accessibility information for the combo box. Possible solution here
would be to split this into two separate controls - text entry plus
a button which shows a list of items.

Partially Fixes VSTS #753567 - Accessibility: Project Options Build
General: Voice over is not reading labels for the combo boxes and text
fields in the page


mrward:
Closing - replaced by #8493

PR #6887, opened 2019-01-14
Pressing Enter key when Exception Caught tooltip is shown displays di…

nosami:
…alog

Note that pressing escape already works to hide tooltip.

Fixes VSTS #646440


sevoku:
@nosami some tests failed, could you rebase on latest master please?

PR #6888, opened 2019-01-14
[Ide] Fix contract in edit references dialog

mrward:
The secondary text displayed for the selected references on the
right hand side of the dialog was hard to read when a list item
was selected. Now the main IDE secondary selection text style is used
when the list item is selected which has better contrast.

 **Before**

before2
before2dark

 **After**

after2
after2dark

Fixes VSTS #750333 - Accessibility: Content under the ‘System’ below
the selected references is not visible in normal mode and same issue
is observed when we enable invert colors and grey scale


vancura:
Is it possible to move the labels (both of them) `2pt` up?

untitled-1

mrward:
The custom cell renderer currently breaks Voice Over.

mrward:
Moved the text up slightly with the last commit.

movedtext

PR #6890, opened 2019-01-15
Update macdoc reference to fix MonoDoc.app missing files

mrward:
Fixes MonoDoc.app not having all required files in app bundle.
The build of MonoDoc was failing with Xamarin.Mac 5.6 since that
version of Xamarin.Mac requires 10.9 to be used in the info.plist
for MonoDoc.

MMP: error MM0073: Xamarin.Mac 5.6.0 does not support a deployment
target of 10.7 for macOS (the minimum is 10.9). Please select a newer
deployment target in your project's Info.plist.

Fixes VSTS #766219 MonoDoc bundle does not include binary files


mrward:
Build MacDoc step is working now for this on VSTS. Need to check monodoc is working in the app bundle.

slluis:
XM has been downgraded to 5.0 in the 7.8 branch, I guess this is not needed then?

mrward:
VS Mac.app looks good. MonoDoc.app has all its files and you can open it from the Help - API Documentation menu.

PR #6892, opened 2019-01-15
Gtkrc a11y fixes

vancura:
These two commits fix color accessibility issues in these areas:

- Secondary (gray) text color now matches required 4.5:1.
- New macOS selection background which follows changes in macOS Mojave.

Fixes VSTS #612400
Fixes VSTS #754225


slluis:
Can you references the bugs this fixes?

vancura:
@slluis There is no bug. I found the problem while testing another thing.

vancura:
Everything compiles nicely, this seems to be done.

vancura:
@slluis Actually I just found a bug! They already filed the selection background color here: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/612400 — what should I do now so this gets propagated there?

PR #6894, opened 2019-01-15
[A11y] Accessibility fixes for the Help Pad

iainx:
Don't trigger MonoDoc when selection changes
Allow rows to be expanded by the keyboard

Fixes VSTS #753478
Fixes VSTS #753484


iainx:

PR #6902, opened 2019-01-15
[netcore]  Better SDK References handling

jtorres:
- Prevents NRE when explicit references

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/765999
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/768376

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
@@ -93,7 +93,7 @@ namespace MonoDevelop.DotNetCore
 		/// </summary>
 		bool IsSdkProject (DotNetProject project)
 		{
-			return project.MSBuildProject.GetReferencedSDKs ().Any ();
+			return project.MSBuildProject.GetReferencedSDKs ().Length > 0;
 		}
 
 		protected override bool OnGetCanReferenceProject (DotNetProject targetProject, out string reason)
@@ -473,7 +473,6 @@ namespace MonoDevelop.DotNetCore
 
 		protected override void OnBeginLoad ()
 		{
-			dotNetCoreMSBuildProject.Sdk = Project.MSBuildProject.Sdk;
 			base.OnBeginLoad ();
 		}
 
@@ -497,11 +496,8 @@ namespace MonoDevelop.DotNetCore
 
 			var sdks = string.Join (";", Project.MSBuildProject.GetReferencedSDKs ());
 			sdkPaths = DotNetCoreSdk.FindSdkPaths (sdks);
-			if (string.IsNullOrEmpty (dotNetCoreMSBuildProject.Sdk)) {
-				dotNetCoreMSBuildProject.Sdk = sdks;
-			}
+			dotNetCoreMSBuildProject.HasSdk = sdkPaths.Exist;
 		}
-
 		protected override async Task<ImmutableArray<ProjectFile>> OnGetSourceFiles (ProgressMonitor monitor, ConfigurationSelector configuration)
 		{
 			var sourceFiles = await base.OnGetSourceFiles (monitor, configuration);
@@ -615,7 +611,14 @@ namespace MonoDevelop.DotNetCore
 
 		bool IsFSharpSdkProject ()
 		{
-			return HasSdk && dotNetCoreMSBuildProject.Sdk.Contains ("FSharp");
+			if (HasSdk) {
+				var sdks = Project.MSBuildProject.GetReferencedSDKs ();
+				for (var i = 0; i < sdks.Length; i++) {
+					if (sdks [i].Contains ("FSharp"))
+						return true;
+				}
+			}
+			return false;
 		}
 
 		/// <summary>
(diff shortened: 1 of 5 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -472,9 +472,7 @@ protected override void OnBeginLoad ()
 			base.OnBeginLoad ();
 		}
 
-		public bool HasSdk {
-			get { return dotNetCoreMSBuildProject.HasSdk; }
-		}
+		public bool HasSdk => Project.MSBuildProject.GetReferencedSDKs ().Any ();
```
Therzok:
```suggestion
		public bool HasSdk => Project.MSBuildProject.GetReferencedSDKs ().Length > 0;
```

mrward: Out of interest - why Length versus Any? Is this just because the underlying type is an Array behind the scenes or something else?

Therzok: This is .netcore, but results are similar for mono:

https://sharplab.io/#v2:C4LghgzgtgPgAgBgARwIwG4CwAoRLUB0AMgJYB2AjlrgMwoBM+A7EgN45Kcp1oBsSAIwD2QgDZIAqhACmAQTIBPABTlgAbQC6SMACcdASjYcuJuC107i0sgHNgACyQA+JAmomAvsc7fu+fsJikjJE1nb2KmTqWhaG7NgmSACQZtp6BPLK+u5cXtgeQA=

Looking further, we can then look at the source code for [Any](https://source.dot.net/#System.Linq/System/Linq/AnyAll.cs,8788153112b7ffd0,references). What we see here is that it _indiscriminately_ creates the enumerator for the `IEnumerable<T>`, which, for most BCL types, ends up boxed, thus allocates.

(`Count()` is different, due to it checking for `ICollection<T>` [here](https://source.dot.net/#System.Linq/System/Linq/Count.cs,41ef9e39e54d0d0b,references))

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -491,8 +489,12 @@ protected override void OnPrepareForEvaluation (MSBuildProject project)
...
 				return;
-		
-			sdkPaths = DotNetCoreSdk.FindSdkPaths (dotNetCoreMSBuildProject.Sdk);
+
+			var sdks = string.Join (";", Project.MSBuildProject.GetReferencedSDKs ());
+			sdkPaths = DotNetCoreSdk.FindSdkPaths (sdks);
+			if (string.IsNullOrEmpty (dotNetCoreMSBuildProject.Sdk)) {
+				dotNetCoreMSBuildProject.Sdk = sdks;
+			}
 		}
```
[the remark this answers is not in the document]
jtorres: done, please check it out

jtorres:
@mrward I think that we should leave HasSdk as bool since it's been used on tests and is easier thank mock up references, anyway, just pushed it as bool according to the references sdks array length. If you think it should be `string[]` rather just let me know. thanks

mrward:
@jtorres HasSdk as bool is good. No need for it to be a string array. Just needs a flag to tell it to save the .csproj as though it is a sdk style project.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -491,10 +493,12 @@ protected override void OnPrepareForEvaluation (MSBuildProject project)
...
 				return;
-		
-			sdkPaths = DotNetCoreSdk.FindSdkPaths (dotNetCoreMSBuildProject.Sdk);
-		}
 
+			var referencedSdks = project.GetReferencedSDKs ();
+			var sdks = string.Join (";", referencedSdks);
+			sdkPaths = DotNetCoreSdk.FindSdkPaths (sdks);
+			dotNetCoreMSBuildProject.HasSdk = referencedSdks.Length > 0;
+		}
```
mrward: I still think it would be good if we could pass the referencedSdks directly to DotNetCoreSdk.FindSdkPaths instead of using string.Join here. [DotNetCoreSdk.FindSdkPaths](https://github.com/mono/monodevelop/blob/c9b6b49b660ae620d8761f385c484580fb73749d/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs#L152-L157) does a String.Split so we are doing work twice here.

I think we should introduce a new overload for FindSdkPaths. One that takes a string and one that takes a string array. Maybe the one that takes a string is not needed - need to see how that is changed.

    `internal static DotNetCoreSdkPaths FindSdkPaths (string[] sdks)`

Therzok:
Cancelling and rerunning tests, Ide gc transient bug.

PR #6907, opened 2019-01-17
Fixes VSTS Bug 769515: Exception in OnMarkerRemoved

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/769515

The cause was that the line segment markers weren't unparented but I
fixed the crash in IconMargin as well.


mkrueger:

PR #6911, opened 2019-01-17
Fixes VSTS Bug 715151: [Feedback] 64 bit unit testing is not supported

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/715151

Unit testing now values the platfom target setting.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
@@ -744,6 +744,7 @@
     <Compile Include="MonoDevelop.Core.FeatureConfiguration\FeatureSwitchCondition.cs" />
     <Compile Include="MonoDevelop.Projects.MSBuild\MSBuildProcessService.cs" />
     <Compile Include="MonoDevelop.Core.FeatureConfiguration\IFeatureSwitchController.cs" />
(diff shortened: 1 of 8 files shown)


Therzok (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
```diff
@@ -780,6 +780,7 @@
   <ItemGroup>
     <InternalsVisibleTo Include="MonoDevelop.Core.Tests" />
     <InternalsVisibleTo Include="MonoDevelop.Ide.Tests" />
+    <InternalsVisibleTo Include="MonoDevelop.UnitTesting.NUnit" />
```
Therzok: Don't like IVT here. Can't we expose this API somehow?

Therzok (changes requested):
Remove IVT

main/src/core/MonoDevelop.Core/MonoDevelop.Core.csproj
mkrueger: No I don't want to make that API public.

mkrueger:
@Therzok: I don't want to change the public API. We should stop just popping in public helper methods from everywhere just because we need to use them internally.

Therzok (approved):

slluis (approved):
Nice!

PR #6914, opened 2019-01-17
Fixes VSTS Bug 753343: Accessibility: Apply Policies Dialog: Focus is

mkrueger:
not visible after Policy Combo Box


sevoku:
@mkrueger the build failed, can you take a look pls?

PR #6919, opened 2019-01-18
Hacky cache implementation to help speed up iterations on editor work

Therzok:
Hidden behind HackyCache feature toggle.

To enable, for local terminal session
`export MD_FEATURES_ENABLED=HackyCache`

or global
`launchctl setenv MD_FEATURES_ENABLED=HackyCache`

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectItem.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectItem.cs
@@ -49,7 +49,8 @@ namespace MonoDevelop.Projects
 			get {
 				return project;
 			}
-			internal set {
+			// HACK: internal
+			set {
 				project = value;
 				OnProjectSet ();
 			}
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/HackyWorkspaceFilesCache.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/HackyWorkspaceFilesCache.cs
@@ -37,7 +37,7 @@ namespace MonoDevelop.Ide.TypeSystem
 {
 	class HackyWorkspaceFilesCache
 	{
-		readonly bool? enabled = FeatureSwitchService.IsFeatureEnabled ("HackyCache");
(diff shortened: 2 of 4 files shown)


Therzok:
cc @abock @sandyarmstrong @DavidKarlas

DavidKarlas:
What I think should happen here are 2 things:
1) Include also references in cache
2) Always request fresh `sourceFiles = await p.GetSourceFilesAsync (config?.Selector).ConfigureAwait (false);` even when cache is available and then if list between cache and fresh request differs add/remove files to Roslyn project and Roslyn will quickly adapt if just 1 or 2 files are added... This will also avoid need for invalidation...

Therzok:
@slluis can we merge this? It's behind a feature toggle and won't impact anyone else.

Therzok:
> Always request fresh sourceFiles = await p.GetSourceFilesAsync (config?.Selector).ConfigureAwait (false); even when cache is available and then if list between cache and fresh request differs add/remove files to Roslyn project and Roslyn will quickly adapt if just 1 or 2 files are added... This will also avoid need for invalidation...

This is a harder task, as we need to take care what we pass over to roslyn. Sometimes it's better to just reconstruct project from scratch than send over changed values.

slluis:
Why is it named "Hacky cache"? are you proposing this as a hack that we'll have to remove?

Therzok:
@slluis I'll make it a proper cache once we figure out the correct way to load from cache and update data to the workspace.

(i.e., in some cases, simply creating a new document dictionary rather than updating the document list makes roslyn process data faster).

A non-hacky cache means:
a) we always load cache data
b) we process the msbuild information in the background
c) update with the correct mechanism (replace full map vs post update of added/removed items) of the workspace.

I needed this hacky version to unblock the editor team from having 10mins to load the typesystem. This makes it _instant_ after the first time the evaluation is done.

It's currently hidden behind a feature toggle, so it won't impact anyone other than those using this feature.

slluis:
Ok, although if the current version of the cache works and causes no regressions I don't see why it needs a feature flag.

slluis (approved):

Therzok:
@slluis it might cause regressions for sdk style projects with globs, for example, as we invalidate the current cache on csproj change.

When we get the implementation above, we'll be able to enable it for general purpose. Although, there's multiple things to fix here, like improve msbuild builder times

DavidKarlas (approved):
Tested and works as advertised, awesome

PR #6922, opened 2019-01-18
Fixes VSTS Bug 770635: NRE in TextEditorData.GetChunks

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/770635

At least it'll help to find the real cause. Added segment == null
check which is a guess which would expose an issue in the syntax mode
implemention. From where that method is called line is always != null.


main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/TextEditorData.cs
```diff
@@ -488,12 +488,27 @@ public string GetMarkup (int offset, int length, bool removeIndent, bool useColo
...
+				foreach (var segment in highlightedLine.Segments) {
+					if (segment == null) {
+						LoggingService.LogError ("Segment== null insede highlighed line " + highlightedLine + " at " + Environment.StackTrace);
+						continue;
+					}
+					result.Add (segment.WithOffset (segment.Offset + lineOffset));
+				}
+				return TextViewMargin.TrimChunks (result, offset, length);
+			} catch (Exception e) {
+				LoggingService.LogError ("Error while getting chunks.", e);
```
slluis:
```suggestion
				LoggingService.LogInternalError ("Error while getting chunks.", e);
```

main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/TextEditorData.cs
```diff
@@ -488,12 +488,27 @@ public string GetMarkup (int offset, int length, bool removeIndent, bool useColo
...
+			if (document == null)
+				throw new InvalidOperationException ("TextEditorData was disposed.");
+			try {
+				var lineOffset = line.Offset;
+				var syntaxMode = document.SyntaxMode;
+				var highlightedLine = await syntaxMode.GetHighlightedLineAsync (line, CancellationToken.None).ConfigureAwait (false);
+				var result = new List<MonoDevelop.Ide.Editor.Highlighting.ColoredSegment> (highlightedLine.Segments.Count);
+				foreach (var segment in highlightedLine.Segments) {
+					if (segment == null) {
+						LoggingService.LogError ("Segment== null insede highlighed line " + highlightedLine + " at " + Environment.StackTrace);
```
slluis:
```suggestion
						LoggingService.LogInternalError ("Segment== null insede highlighed line " + highlightedLine + " at " + Environment.StackTrace);
```

slluis:
Uh, massive change in TextEditorData.cs

main/src/core/Mono.TextEditor.Shared/Mono.TextEditor/TextEditorData.cs
Therzok: Fixed.

slluis:

PR #6928, opened 2019-01-18
[Ide] Support string tags for paths in ItemTemplates

rodrmoya:


mrward:
Failing tests are in the xml addin and are unrelated to this change.

PR #6931, opened 2019-01-18
[Ide] Various memory retention fixes around the project model

Therzok:
This change allows us to reclaim some memory earlier, rather than relying on the user invalidating the data via a future operation.

Fixes VSTS #773051 - Possible leak of large data via AsyncOperation

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
@@ -1233,23 +1233,22 @@ namespace MonoDevelop.Ide
 
 				OnStartClean (monitor, tt);
 
-				var t = CleanAsync (entry, monitor, tt, false, operationContext)
-					.ContinueWith (ta => {
-						ResetCurrentBuildOperation ();
-						currentBuildOperationOwner = null;
-						return ta.Result;
-					});
+				var t = CleanAsync (entry, monitor, tt, false, operationContext);
 
 				var op = new AsyncOperation<BuildResult> (t, cs);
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
```diff
@@ -1671,10 +1671,13 @@ AsyncOperation<BuildResult> Build (IBuildTarget entry, bool skipPrebuildCheck, C
 					cs = CancellationTokenSource.CreateLinkedTokenSource (cs.Token, cancellationToken.Value);
 				ProgressMonitor monitor = IdeApp.Workbench.ProgressMonitors.GetBuildProgressMonitor ().WithCancellationSource (cs);
 				BeginBuild (monitor, tt, false);
-				var t = BuildSolutionItemAsync (entry, monitor, tt, skipPrebuildCheck, operationContext);
+				var t = BuildSolutionItemAsync (entry, monitor, tt, skipPrebuildCheck, operationContext)
+					.ContinueWith (ta => {
+						ResetCurrentBuildOperation ();
```
slluis: This could potentially be executed before currentBuildOperation is assigned below, which would be incorrect.

slluis (changes requested):

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
Therzok: Fixed!

Therzok:
Fixed the possible race.

slluis (approved):

PR #6944, opened 2019-01-21
Fix some issues in the New Project Dialog

sevoku:
Fixes VSTS #675356
Fixes VSTS #752336

Depends on https://github.com/mono/xwt/pull/912


sevoku:
> With this patch the menu is shown in the mouse position, while it should be shown below the language tag, always in the same position

Right, that requires a small change in Xwt: https://github.com/mono/xwt/pull/912

PR #6945, opened 2019-01-21
Fixes Bug 770919: NRE in OSXEditor.GetLineHeight

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/770919


mkrueger:
Not sure why that happens - but it seems it's possible that the pango font name is not the real system font name. Using pango metrics is ok for most fonts btw.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/OSXEditor.cs
```diff
@@ -66,9 +66,12 @@ public static double GetLineHeight(string fontName)
 		{
 			var editorFont = Xwt.Drawing.Font.FromName(fontName);
 
-			using (var nsFont = NSFont.FromFontName(editorFont.Family, (nfloat)editorFont.Size))
-				using (var lm = new NSLayoutManager())
-			 		return lm.DefaultLineHeightForFont(nsFont);
+			using (var nsFont = NSFont.FromFontName (editorFont.Family, (nfloat)editorFont.Size)) {
```
Therzok: Disposing the NSFont is safe after we include XamMac with this fix: https://github.com/xamarin/xamarin-macios/pull/5433

Otherwise, we should not dispose the NSFont.

Therzok:
@mkrueger commented. We need a newer xammac, and it should be okay without this patch.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/OSXEditor.cs
sandyarmstrong: Yes, this `using` is potentially problematic until a fixed Xamarin.Mac is used. So far it's only been a problem in the `roslyn-upgrade` branch.

When this bug rears its head, it results in a native crash because the managed NSFont instance (which is not null) has been prematurely released on the native side.

mkrueger: The main issue is the Sans 10 string - shouldn't be a valid font name on macOS.

PR #6962, opened 2019-01-23
[Debugger] Show main window when debugging app

rodrmoya:
Now that we have a separate window (Start Window), we need to make
sure we show the main window if it's not visible for operations that
rely on the main window being shown, like is the case for Run->Debug
Application.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/770996

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Extensions.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Extensions.cs
@@ -65,6 +65,10 @@ namespace MonoDevelop.Debugger
 
 		public static AsyncOperation DebugApplication (this ProjectOperations opers, string executableFile, string args, string workingDir, IDictionary<string,string> envVars)
 		{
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Extensions.cs
```diff
@@ -65,6 +65,10 @@ public static AsyncOperation DebugFile (this ProjectOperations opers, string fil
 
 		public static AsyncOperation DebugApplication (this ProjectOperations opers, string executableFile, string args, string workingDir, IDictionary<string,string> envVars)
 		{
+			if (!IdeApp.Workbench.RootWindow.Visible) {
```
Therzok: Shouldn't this happen in attaching to process, etc?

rodrmoya: right, added to the commit now

PR #6966, opened 2019-01-24
Fixes NRE in case of bind a instance non IComponent

netonjm:
In case of crating an Event editor with a instance that desn't inherit from IComponent this thows a NRE breaking the IDE.

This commit, returns null creating this Editor which is covered by the current implementation


sevoku:
Can't we avoid creating the editor at all, if it relies on `Instance is IComponent`? My feeling is that this is the wrong place to check for that, cause it creates a noop editor for no reason.

PR #6972, opened 2019-01-24
 Fixes calculate of the next focused view in SelectorView

netonjm:
Fixes VSTS 648190 - Accessibility: Keyboard: MAS2.4.3: Focus order is inappropriate while navigating in reverse order in the toolbar.

https://devdiv.visualstudio.com/DevDiv/_queries/edit/648190/?triage=true

toolbarfocusfix

To test the issue:
- Open VS4Mac
- Click on the Global Search box
- Press Shift + Tab/Tab to move the focus between the views
- Ensure you can access all buttons in the correct order

### DIFF
--- a/main/src/addins/MacPlatform/MainToolbar/SelectorView.cs
+++ b/main/src/addins/MacPlatform/MainToolbar/SelectorView.cs
@@ -124,22 +124,10 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 
 		internal void OnSizeChanged ()
 		{
-			if (SizeChanged != null) {
-				SizeChanged (this, EventArgs.Empty);
-			}
+			SizeChanged?.Invoke (this, EventArgs.Empty);
 		}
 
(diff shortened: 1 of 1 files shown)


main/src/addins/MacPlatform/MainToolbar/SelectorView.cs
```diff
@@ -584,29 +572,39 @@ public override void KeyDown (NSEvent theEvent)
 
 				// 0x30 is Tab
 				if (theEvent.KeyCode == (ushort)KeyCodes.Tab) {
```
Therzok: How can we have a space inside the tab key?

Therzok: I guess this check should be removed.

netonjm: right ! good catch!

PR #6976, opened 2019-01-25
[VCS] Add instrumentation counters to repository operations

sevoku:
Fixes VSTS 767160

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlService.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlService.cs
@@ -226,7 +226,6 @@ namespace MonoDevelop.VersionControl
 			return repo;
 		}
 
-		readonly static Counter<RepositoryMetadata> Repositories = InstrumentationService.CreateCounter<RepositoryMetadata> ("VersionControl.RepositoryOpened", "Version Control", id:"VersionControl.RepositoryOpened");
 		internal static readonly Dictionary<FilePath,Repository> repositoryCache = new Dictionary<FilePath,Repository> ();
 		public static Repository GetRepositoryReference (string path, string id)
 		{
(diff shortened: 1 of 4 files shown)


sevoku:

mrward:
Is there a VSTS user story that this should link to?

sevoku:
tests are failing, fixing...

sevoku:

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlService.cs
```diff
@@ -847,6 +846,12 @@ public RepositoryMetadata ()
 		{
 		}
 
+		public RepositoryMetadata (VersionControlSystem versionControl)
```
Therzok: We should explicitly disallow the normal constructor, as we could get events that get the metadata bag created via the Counter itself, thus circumventing this.

Therzok: My point:
```
using (var counter = CreateCounter<RepositoryMetadata>())
{
   counter.Metadata.Path = value;
}
```

Will send over a counter with no VCS system set. We should prevent not having one set.

sevoku: That's not easy because of https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/InstrumentationService.cs#L247, but the metadata should not be public and we could throw an `InvalidOperationException` internally.

Therzok: Yup! throw InvalidOperationException in the normal constructor

sevoku: done in 29e31b5

sevoku:
Reminder: don't backport API change from 29e31b5

PR #6978, opened 2019-01-25
[MacPlatform] Ignore hidden items in responder chain

rodrmoya:
PathSelectorView has 3 "sections", one of them being, for some kind of
projects/setups, hidden, but the code was assuming all of them were
always visible when responding to responder chain events. Now make sure
only visible items are processed so that there is never focus loss.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/753499


netonjm:
The problem persists to me:
focus-persiste.

rodrmoya:
Ok, needs to check for `Enabled` also I guess, fixing

rodrmoya:
Fixed now, it takes into account both visible and enabled cells

PR #6988, opened 2019-01-28
Fixes VSTS Bug 675488: [Feedback] Changing tab doesn't allow to

mkrueger:
directly type

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/675488

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/SdiWorkspaceWindow.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/SdiWorkspaceWindow.cs
@@ -262,9 +262,10 @@ namespace MonoDevelop.Ide.Gui
 		{
 			if (subViewToolbar != null)
 				subViewToolbar.Tabs [subViewToolbar.ActiveTab].Activate ();
+			SelectWindow ();
 		}
 		
-		public void SelectWindow()
+		public void SelectWindow ()
 		{
 			var window = tabControl.Toplevel as Gtk.Window;
 			if (window != null) {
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/SdiWorkspaceWindow.cs
```diff
@@ -287,10 +288,14 @@ public void SelectWindow()
 			// The tab change must be done now to ensure that the content is created
 			// before exiting this method.
 			tabControl.CurrentTabIndex = tab.Index;
-
 			// Focus the tab in the next iteration since presenting the window may take some time
 			Application.Invoke ((o, args) => {
 				DockNotebook.ActiveNotebook = tabControl;
+				var editor = ActiveViewContent.GetContent<Ide.Editor.TextEditor> ();
```
Therzok: Should this know of the editor? Shoudln't we do that on ActiveWindowChanged somewhere else?

mkrueger:
Y it still repros.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/SdiWorkspaceWindow.cs
mkrueger: Editor should be in focus if there is an editor. It's a bit more complex for other controls. The proper thing would be that the workspace window stores the widget that had the focus on focus lost and give it back in the OnSelect event.

However I wasn't able to find a good implementation for it. Esp. the search widget is generated after the workspace window initialization and the gtk focus APIs are a bit limited. The workspace window atm controls the focusing on selection that's why it's there. I don't want to clutter that logic up too much.

mkrueger:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/SdiWorkspaceWindow.cs
Therzok: @slluis ^^

slluis: Marius is right, we should not special case the editor here.

mkrueger: Any suggestions ? DeepGrabFocus needs to be bypassed in that case.

mkrueger:

mkrueger:
Moved the logic into the view content. But that means that every view content needs to handle it itself correctly.

Therzok:
This might conflict with @slluis work to rework the document model.

PR #6989, opened 2019-01-28
Add logging on whether hackyowrkspacecache is enabled

Therzok:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/HackyWorkspaceFilesCache.cs
```diff
@@ -49,6 +49,7 @@ public HackyWorkspaceFilesCache (Solution solution)
 			if (!IdeApp.IsInitialized || !enabled || solution == null)
 				return;
 
+			LoggingService.LogDebug ("HackyWorkspaceCache enabled: {0}", enabled.ToString ());
```
sandyarmstrong: Maybe only log if it _is_ enabled? Going to be disabled for 99% of people so it's log noise.

Therzok: It is logged only if enabled. I just forgot to remove the variable value.

PR #6992, opened 2019-01-28
Fixes VSTS Bug 763702: [Feedback] Using NUnit, typing inside

mkrueger:
Assert.That(tes pops up a Create method dialog that gets in the way

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/763702

The old CompletionWindowManager hack doesn't work in this case - so I
introduced a pseudo-parameter (can't change that API or use it in any
sane way). A roslyn refactoring is requesting now the changes and that
happens during completion as well - that's why the insert windows are
popping up. This prevents that from happening.


mkrueger:

mkrueger:
What should an extension method do different ? Doesn't make sense IMO. An additional parameter - what for ? It's just a single call where this parameter would always be true.

Y it's a hack but making a hack more complex for no appearant reason isn't good. IMO an extension method class should never provide any "flag" and an extension method that toggles a static flag in another class is worse than this solution IMO.

Therzok:
We can break API in this release though.

mkrueger:
@Therzok : how to break roslyn API in this release ?

PR #6996, opened 2019-01-29
Fixes VSTS Bug 766267: No keyboard shortcut to run test from the text

mkrueger:
editor

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/766267


mkrueger:

main/src/addins/MonoDevelop.UnitTesting/Services/AbstractUnitTestEditorExtension.cs
```diff
@@ -137,6 +139,45 @@ static void RemoveHandler ()
 
 		List<IUnitTestMarker> currentMarker = new List<IUnitTestMarker>();
 
+		IUnitTestMarker GetMarkerAtCaret ()
```
Therzok: Maybe we can remove some duplicated logic by making it so we get the TestRunner for the caret.
```
TestRunner GetTestRunnerAtCaret (bool debug)
{
			var line = Editor.GetLine (Editor.CaretLine);
			if (line == null)
				return null;
			foreach (var marker in Editor.GetLineMarkers (line)) {
				if (marker is IUnitTestMarker result)
					return new TestRunner (result.UnitTest.UnitTestIdentifier, DocumentContext.Project, debug);
			}
			return null;
}
```

main/src/addins/MonoDevelop.UnitTesting/Services/AbstractUnitTestEditorExtension.cs
```diff
@@ -137,6 +139,45 @@ static void RemoveHandler ()
...
+				return null;
+			foreach (var marker in Editor.GetLineMarkers (line)) {
+				if (marker is IUnitTestMarker result)
+					return result;
+			}
+			return null;
+		}
+
+		[CommandHandler (TextEditorCommands.RunTestAtCaret)]
+		protected void OnRunTestAtCaret ()
```
Therzok: With the change above, this would be:
```
	GetTestRunnerAtCaret (debug: false)?.Run (this, EventArgs.Empty);
```

main/src/addins/MonoDevelop.UnitTesting/Services/AbstractUnitTestEditorExtension.cs
```diff
@@ -277,82 +318,82 @@ public override void PopupContextMenu (UnitTestLocation unitTest, int x, int y)
...
-				{
-					var test = UnitTestService.SearchTestByDocumentId (testCase);
-					if (test != null) {
-						RunTest (test); 
-						timeoutHandler = 0;
-					} else {
-						return true;
-					}
-					return false;
+			bool TimeoutHandler ()
```
Therzok: Doesn't seem used at all?

mkrueger: Not part of this PR.

PR #6998, opened 2019-01-29
Fixes VSTS Bug 780196: [Feedback] ディレクティブの入力時に他のregionが開く-Another

mkrueger:
region opens when entering a directive

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/780196


slluis:
There are several unit test failures. Can you add a unit test for this bug?

PR #6999, opened 2019-01-29
[DotNetCore] Create a DotNetCoreSdkPaths instance per project

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/779492

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
@@ -36,11 +36,10 @@ namespace MonoDevelop.DotNetCore
 	public static class DotNetCoreSdk
 	{
 		static readonly Version DotNetCoreVersion2_1 = new Version (2, 1, 0);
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
```diff
@@ -36,14 +36,6 @@ namespace MonoDevelop.DotNetCore
...
 	{
 		static readonly Version DotNetCoreVersion2_1 = new Version (2, 1, 0);
-		static readonly DotNetCoreSdkPaths sdkPaths;
-
-		static DotNetCoreSdk ()
-		{
-			sdkPaths = new DotNetCoreSdkPaths ();
-			sdkPaths.ResolveSDK ();
-			Update (sdkPaths);
-		}
```
mrward: We need this. This needs to be put back. The original code did not call Update.

rodrmoya: Ok, changed it to use a copy for initialisation and then another every time FindSdkPaths is called

PR #7003, opened 2019-01-30
[Ide] Add New Folder dialog for adding new folders

mrward:
Due to various problems with the Solution window when adding a new
folder directly into the tree and then editing the label, instead
a New Folder dialog is used when creating a new folder for a project.

Fixes VSTS #721437 Can't focus on a folder and name it during creation.

Needs a UI review. Is using a separate message alert for errors OK or should an icon + text be in the dialog itself?

newfolderalertmessage

Also has some voice over problems since the native XWT toolkit is being used - there is a bug open for one problem about not announcing the label for the text box - also seeing odd behaviour with the alert message being in the wrong position and the text is not selected, unlike over alert dialogs.

newfoldervoiceover

### DIFF
--- a/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
+++ b/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
@@ -256,7 +256,7 @@
 			_description = "Adds and existing folder and its contents"
 			_label = "_Add Existing Folder..." />
 	<Command id = "MonoDevelop.Ide.Commands.ProjectCommands.NewFolder"
-			_label = "New _Folder"
+			_label = "New _Folder…"
 			_description = "Create a new folder"
 			icon  = "md-new-folder" />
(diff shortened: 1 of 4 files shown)


mrward:
OK - going to look at getting rid of the alert message dialog and instead display the warning/error in the New Folder dialog.

vancura:
I need to figure out if we could approve something like this without the UXBoard, or if everything that changes UI in any way needs to be generally approved by the board. I'll be here as soon as possible.

mrward:
OK updated this. The New Folder dialog now shows a warning/error icon next to the folder name instead of showing a separate error/warning dialog. There are still several accessibility problems when using the native XWT toolkit - which will be addressed in XWT.

newfolderdialoginlinewarning
newfolderdialog
newfolderdialogwarninginuse
newfolderwarning
newfolderdialogerror

vancura:
In the queue for UXBoard review today. We might need more changes :(

PR #7006, opened 2019-01-30
[Ide] Don't set full screen mode on main window until it is shown

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/738914


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DefaultWorkbench.cs
```diff
@@ -146,13 +146,34 @@ private void DoNotCompile ()
...
 			}
 			set {
-				DesktopService.SetIsFullscreen (this, value);
+				// If this window is not visible, don't set full screen mode
+				// until it is, as that would conflict with other windows we
+				// might be opening before (Start Window, for instance)
+				if (Visible) {
+					DesktopService.SetIsFullscreen (this, value);
+				} else if (value) {
+					Shown += OnEnableFullScreenOnShown;
```
Therzok: We need to register this event once and only once.

rodrmoya: Much better now indeed

PR #7012, opened 2019-01-31
Rework FileService allowing us to avoid File.Replace bug that closes documents

Therzok:
This PR aims to integrate the proposal found [here](https://paper.dropbox.com/doc/FileService-API-breakages-and-optimizations--AWqJFOGJmAqRDNdUa8A6OC7BAg-Lmx4sBpGbT2V0vd2IGD1S) to allow Files not to be closed when the new vs editor backend is used.

Fixes VSTS #718767 - File close when save

TODO List:
* [x] Problem 1 - Simplify EventData
* [x] Problem 2 - Fix handler value bug
* [x] Problem 3 - Merging events is now fixed
* [x] Problem 4 - Event reduction.

### DIFF
--- a/main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring/RefactoringService.cs
+++ b/main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring/RefactoringService.cs
@@ -108,7 +108,7 @@ namespace MonoDevelop.Refactoring
 			}
 			public void FileRename (object sender, FileCopyEventArgs e)
 			{
-				foreach (FileCopyEventInfo args in e) {
+				foreach (FileEventInfo args in e) {
 					foreach (Change change in changes) {
 						var replaceChange = change as TextReplaceChange;
 						if (replaceChange == null)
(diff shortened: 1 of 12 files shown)


Therzok:
Writing the event queue trimming code, I've wanted to make a few notes on github so I don't forget
a) FileService is frozen
b) These events are queued
```
-- a exists
remove a
create b
create a
```

This would effectively become: `create b` and `change a`. I imagine that there is no dependency of `b` on `a` since, `a` was removed before `b` was created.

```
create a
change a
change a
create b
change a
```

This would effectively become `create a`, `change a`, `create b`, `change a`, as we don't want to break the dependency

Therzok:
Actually, I'm not sure anymore we need to trigger multiple change events for `a`. One should suffice.

slluis (approved):

Therzok:
Let's see, it should pass tests now.

Therzok:
Also, testing on windows to see if it repros.

Therzok:
This works on mac, but not on Windows. Not sure exactly what's going wrong there, but not sure we can fix it for P3.

mrward (approved):
Looks good to me.

slluis (approved):

PR #7019, opened 2019-01-31
[FileWatcherService] Switch to a better implementation for watchers for Mac/Win

Therzok:
PathTree was supposed to fix an issue with the OSX file watcher, and that was having one thread corresponding to one FileSystemWatcher.

Switch to a ref-counted system where we have multiple, more granular watchers, preventing us from having a too wide cone of directories to watch.

Keep the old watcher logic on Linux, where the 1:1 FSW - Thread mapping remains.


mrward (commented):
Do we need some logic here to handle overlapping directories on the Mac?

Seems like anybody could say 'watch these directories' and there could be overlaps?

Solution 1: /root/solution1/
Solution 2: /root/solution2
                   /root/solution1/projectA

ProjectA is in both solution 2 and solution 1.

Presumably we would could more than one file watcher, for a particular directory, due to capturing all events for child directories, so we would be getting multiple events due to the multiple watchers.

Do we not need something similar to what is done in the workspace item's GetRootDirectories - but presumably more complicated since it would need to handle a directory being set at a higher level?

https://github.com/mono/monodevelop/blob/a5a368f77a9046edf4e6b914977799c9cb973aa4/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceItem.cs#L215

So we remove directories that already exist due to having a parent path being watched. Also handling someone adding a root directory above all existing directories.

Therzok:
Good point! I'll validate and see what we can do.

mrward:
Looks like there was something like this before PathTree was added:

https://github.com/mono/monodevelop/blob/5cc345dbdc497d33d1cc85329b5aa176156ab9a4/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs#L182

Therzok:
Okay, so I'm refactoring the code a bit, it seems like more of the code is actually share-able.

Will make it so we actually do take into account parent directories and only trigger actual watcher updates when we need to.

PR #7020, opened 2019-01-31
[Ide] Fix NRE in ExportSolutionDialog initialization

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/681358

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/ExportSolutionDialog.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/ExportSolutionDialog.cs
@@ -1,4 +1,4 @@
-// ExportProjectDialog.cs
+﻿// ExportProjectDialog.cs
 //
 // Author:
 //   Lluis Sanchez Gual <lluis@novell.com>
@@ -43,7 +43,7 @@ namespace MonoDevelop.Ide.Projects
 		{
 			this.Build();
 			
-			labelNewFormat.Text = selectedFormat.ProductDescription;
+			labelNewFormat.Text = selectedFormat?.ProductDescription ?? item.FileFormat.ProductDescription;
 			
 			formats = MSBuildFileFormat.GetSupportedFormats (item).ToArray ();
 			foreach (var format in formats) {
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
@@ -1,4 +1,4 @@
-//
+﻿//
 // ProjectOperations.cs
 //
 // Author:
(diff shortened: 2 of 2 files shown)


rodrmoya:
Ah sorry, it is indeed available in the popup menu, in Tools->Convert....

mrward:
This has worked in the past. Does not work in VS Mac 7.7 nor 7.6. But seems to work in VS Mac 7.5.

Seems like a change here may have broken it - https://github.com/mono/monodevelop/commit/e1b22249418126b333241fdbfd0310794016f5e9 - which seems to be in 7.5 although the version I have seems to be OK.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/ProjectCommands.cs
```diff
@@ -464,8 +464,11 @@ internal class ExportSolutionHandler : CommandHandler
 		protected override void Update (CommandInfo info)
 		{
 			// FIXME: Once we fix Workspaces to offer Visual Studio formats (instead of the deprecated MonoDevelop 1.0 format), we can allow exporting of Workspaces as well.
-			if (!(IdeApp.ProjectOperations.CurrentSelectedItem is Solution) && !(IdeApp.ProjectOperations.CurrentSelectedItem is SolutionItem))
```
slluis: Why exclude sdk-style projects?

rodrmoya: The command makes no sense, AFAICS, for SDK-style projects, @mrward right?

slluis: This is about exporting a solution to a new location. I don't see why it wouldn't make sense for a sdk-style project.

mrward: Yeah makes sense. I have not seen the dialog in a while - I thought it just did a conversion in place. There is a text box where you can specify a different location if you want to. So it can export to another location - so you would want the sdk project copied in that case.

I guess we can just revert the change to the ProjectCommands.Update method.

rodrmoya: Ah, thought it was for converting the current solution's format, since the menu says "Convert solution format". I guess we should then change the menu item's text? and probably revert 
https://github.com/mono/monodevelop/pull/7020/commits/381dd67383b1bfe1f4e4a39a2b459e273d400b57 ?

rodrmoya: As discussed, since this is for exporting, leaving only the 1st commit which fixes the crash.

PR #7031, opened 2019-02-02
[Mac] Add "New Instance" command to menu item when right-clicking on …

DavidKarlas:
…instance of IDE in Mac Dock
A picture is worth a thousand words:
image

Enough `open -n`

### DIFF
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -266,6 +266,7 @@ namespace MonoDevelop.MacIntegration
 						e.Reply = NSApplicationTerminateReply.Now;
 					}
 				};
+				appDelegate.ShowDockMenu += AppDelegate_ShowDockMenu;
 			}
 
 			// Listen to the AtkCocoa notification for the presence of VoiceOver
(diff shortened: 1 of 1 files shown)


main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -298,6 +299,27 @@ public override Xwt.Toolkit LoadNativeToolkit ()
...
+				return;
+			var menu = new NSMenu ();
+			var newInstanceMenuItem = new NSMenuItem ();
+			newInstanceMenuItem.Title = GettextCatalog.GetString ("New Instance");
+			newInstanceMenuItem.Activated += NewInstanceMenuItem_Activated;
+			menu.AddItem (newInstanceMenuItem);
+			e.DockMenu = menu;
+		}
+
+		void NewInstanceMenuItem_Activated (object sender, EventArgs e)
```
Therzok: I guess this can be static.

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -298,6 +299,27 @@ public override Xwt.Toolkit LoadNativeToolkit ()
...
+		{
+			var bundlePath = NSBundle.MainBundle.BundlePath;
+			NSWorkspace.SharedWorkspace.LaunchApplication (NSUrl.FromFilename (bundlePath), NSWorkspaceLaunchOptions.NewInstance, new NSDictionary (), out NSError error);
+			if (error != null)
+				LoggingService.LogError ($"Failed to start new instance: { error.LocalizedDescription}");
```
Therzok:
```suggestion
				LoggingService.LogError ($"Failed to start new instance: {error.LocalizedDescription}");
```

main/src/addins/MacPlatform/MacPlatform.cs
Therzok:
```suggestion
		static void NewInstanceMenuItem_Activated (object sender, EventArgs e)
```

mrward:
This is nice. Just wondering about the label 'New Instance'.

- VS Code uses 'New Window'
- Terminal uses 'New Window'
- Finder uses 'New Finder Window'

DavidKarlas:
@mrward but macOS API says NewInstance, joke aside
All those apps don’t actually spawn another app which has new icon in Dock.
Would likw others input, I don’t care much about name tbh :)

slluis:
"New Instance" is ok, maybe some day we'll add support for "New Window".

vancura:
What about "New Window"? "New Instance" sounds too robotic. @hbons thoughts?

vancura:
Yeah, @Therzok is right, I haven't seen the rest of the comments, sorry (too swamped).

Keep New Instance then. On the end we're making an editor for people who know what an instance is, this is not iPhoto :)

hbons:
Agree with @slluis. If it launches with another dock icon: "New Instance" is ok. Otherwise "New Window".

jmatthiesen:
FYI - I just added a reference in the description to a VSTS item for this, for ease of tracking with release notes.

PR #7035, opened 2019-02-03
Dispose MonoDevelopAnalyers otherwise they are kept in memory by File…

DavidKarlas:
…WatcherService

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectSystemHandler.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectSystemHandler.cs
@@ -42,7 +42,7 @@ namespace MonoDevelop.Ide.TypeSystem
 {
 	public partial class MonoDevelopWorkspace
 	{
-		internal class ProjectSystemHandler
+		internal class ProjectSystemHandler : IDisposable
 		{
 			readonly MonoDevelopWorkspace workspace;
 			readonly ProjectDataMap projectMap;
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectSystemHandler.cs
```diff
@@ -174,6 +175,7 @@ internal async Task<ProjectInfo> LoadProject (MonoDevelop.Projects.Project p, Ca
 					references.Select (x => x.CurrentSnapshot),
 					analyzerReferences: analyzerFiles.SelectAsArray (x => {
 						var analyzer = new MonoDevelopAnalyzer (x, hostDiagnosticUpdateSource.Value, projectId, workspace, loader, LanguageNames.CSharp);
+						analyzersToDispose.Add (analyzer);
```
Therzok: Why don't we just capture the immutable array created here and dispose of that?

Therzok: And by that, I mean have a `List<ImmutableArray<MonoDevelopAnalyzer>>` which will in turn end up being more efficient.

DavidKarlas: I'm not sure what you mean, can you make commit that does the change and push it to branch to update PR?

Therzok: Actually, it might be better to have this on the large object heap so it's not scanned often.

PR #7043, opened 2019-02-04
[NuGet] Prevent custom tool being run during NuGet restore

mrward:
On creating a new Xamarin.Forms project sometimes an error reported
by the custom tool service would be added to the Errors list.

MSB4057: The target "UpdateDesignTimeXaml" does not exist in the project.

The best way to reproduce this seems to be to remove the Xamarin.Forms
NuGet packages from ~/.nuget/packages/xamarin.forms and then restart
the IDE and create a Xamarin.Forms project with iOS, Android and
.NET Standard.

During the NuGet restore the project is re-evaluated. This causes
the .xaml and .xaml.cs files to be removed and then added back to
the project. This is because the files now have different MSBuild
metadata after the Xamarin.Forms MSBuild targets are re-evaluated.
When the files are added back this then causes the custom tool service
to run the UpdateDesignTimeXaml msbuild target. Whilst this seems to
work for App.xaml it fails for the next file it tries MainPage.xaml.
To workaround this the NuGet addin will freeze the custom tool service
so it queues possible updates whilst the restore is being done. After
the restore is completed the custom tool service is unfrozen and the
project file events are then processed. This seems to prevent the
error about the missing UpdateDesignTimeXaml target.

Also fixed the custom tool service not checking the correct file
when iterating over a set of project files in one of the Update
methods. The file was taken from the enumerator at the start of
the method and then whilst enumerator.MoveNext () was called the
file being checked to see if a generator should be run was not the
current file as specified by the enumerator.

Fixes VSTS #672679 - UpdateDesignTimeXaml does not exist error on
creating new Xamarin.Forms project

### DIFF
--- a/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/PackageManagementEventsMonitor.cs
+++ b/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/PackageManagementEventsMonitor.cs
@@ -26,9 +26,9 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Threading.Tasks;
 using MonoDevelop.Core;
+using MonoDevelop.Ide.CustomTools;
 using NuGet.PackageManagement;
 using NuGet.ProjectManagement;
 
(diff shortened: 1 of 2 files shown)


mrward:
How do we want it to work?

I do not think freezing it for one project during the re-evaluation will fix the problem. The error seems to happen after the re-evaluation has completed not during re-evaluation but I can check if this the case. I think we will need to wait until the NuGet restore has completed for the entire solution. We could have the NuGet addin freeze all project events whilst a restore is happening - I just wonder if that would break things - but this does not seem like what you are suggesting since NuGet currently is an IDE feature not a project model feature.

mrward:
Alternative simpler fix here - https://github.com/mono/monodevelop/pull/7151

PR #7057, opened 2019-02-05
 [Roslyn] Rework persistent storage location service 

Therzok:
Brings it in line with VSWorkspace implementation.

We only support one main solution for persistence, the others
are ignored. So a secondary solution in the workspace will not use persistence.

Also, we do not handle the case where we do:
* Open Solution A
* Open Solution B
* Close Solution A

Instead of using persistence for B, we do not. We will use it for the next solution
that is opened.

Fixes VSTS #786109 - SQLitePersistentStorage not disposed on solution close and other issues


Therzok:
@slluis we might want this in 7.8?

slluis:
@Therzok will this have an important effect on performance?

Therzok:
There's two implications of this:

a) The first time we open a solution, we create a persistent storage for it and indefinitely leak it
b) If we reload the solution, we will not be able to use the persistent storage.

This means that we have a definite leak if we switch solutions and if we switch from and back to a solution, it won't use persistence.

That means caches are not used, computations are done, etc. I'd call it severe impact for this scenario, which is most likely not an edge case (switching between two solutions).

PR #7062, opened 2019-02-06
[A11y] Fix up memory pressure caused by lots of arrays being created

Therzok:
By using native NSMutableArrays we avoid copying a lot of arrays between
native and managed, thus avoid a lot of GC pressure that can happen while
modifying a11y properties.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs
@@ -37,6 +37,7 @@ using Foundation;
 using Gdk;
 using ObjCRuntime;
 using MetalPerformanceShaders;
+using MonoDevelop.Components.Mac;
 
 namespace MonoDevelop.Components.AtkCocoaHelper
 {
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs
```diff
@@ -331,16 +343,12 @@ public static void SetTabs (this Atk.Object o, params Atk.Object [] tabs)
 				return;
 			}
 
-			NSObject [] realTabs = new NSObject [tabs.Length];
-			int i = 0;
-			foreach (var tab in tabs) {
-				realTabs [i] = (NSObject)GetNSAccessibilityElement (tab);
-				i++;
+			using (var realTabs = new NSMutableArray ((nuint)tabs.Length)) {
```
Therzok: Fix me

Therzok:

iainx:
My only thought would be, would it be easier to maintain if all the bits like `Messaging.void_objc_msgSend_IntPtr (nsa.Handle, selSetAccessibilityServesAsTitleForUIElements_Handle, titleElements != null ? titleElements.Handle : IntPtr.Zero);` were inside helper methods?

Therzok:
@iainx done

Therzok:
@slluis @sevoku PTAL

PR #7065, opened 2019-02-06
[C#] Fix NRE caused by #7059

Therzok:


DavidKarlas (changes requested):
Why 2 ifs? There is existing one few lines below...

Therzok:
It looks like it can handle non-DotNetExecutionCommand, so it's better to handle ProcessExecutionCommand separately.

DavidKarlas (approved):

Therzok:
I.e. old code did not return if DotNetExecutionCommand wasn't passed, so it's being safe and handling the scenario where we don't get a DotNetExecutionCommand 'right'.

mrward (approved):

PR #7066, opened 2019-02-06
Copy Property Editing Resource Bundle to Output Dir

CartBlanche:
Depends on https://github.com/xamarin/Xamarin.PropertyEditing/pull/516

I presume we may need another step to ensure the `PropertyEditingResource.bundle` file is included with the installer and is along side `Xamarin.PropertyEditing.Mac.dll`


CartBlanche:
@Therzok AFAIK `MonoDevelop.DesignerSupport` references `Xamarin.PropertyEditing.Mac.dll`. So are you saying that `MonoDevelop.DesignerSupport` should be responsible for copying both `Xamarin.PropertyEditing.Mac.dll` and the `PropertyEditingResource` sub-dirs and it's files into `PropertyEditingResource.bundle` inside the `MonoDevelop.Startup/build/bin` directory?

Therzok:
Yes, it looks like `MonoDevelop.DesignerSupport` should handle the copying.

Why is copying to `build/bin` needed? Won't having it in the DesignerSupport's addin directory be enough?

PR #7083, opened 2019-02-07
[Core] Fix error running project with empty working directory

mrward:
With a project that has a run configuration to run an external
program but did not set the working directory it was not possible
to run the external program. The working directory, which is an
empty string, was being passed to Directory.Exists which returns
false and an error about a missing directory was shown in the
status bar. Now the Directory.Exists check is only made for a
non-empty string.

Fixes VSTS #787952 - Empty working directory prevents project from
running


Therzok (approved):
Thanks for handling this!

PR #7089, opened 2019-02-08
Update TestLogSynchronizationContextThroughWrites to make it possible to troubleshoot it when it fails

kg:
This test is failing on CI and it's basically impossible to troubleshoot it without reproducing locally.

(Also, mock-based unit tests should not be written this way, but rewriting it is a larger undertaking)


slluis:
Already merged.

PR #7095, opened 2019-02-08
Harden the null checking around unregistering object val in debugger

nosami:
MERP / DR Watson shows this as one of the most frequent crashes in VSMac

```
    Fault:       Crash
    thread 0:
        IsManaged:   False
        NativeFrames:0
        Managed Frames:
            MonoDevelop.Debugger.dll!ObjectValueTreeView.UnregisterValue+0
            MonoDevelop.Debugger.dll!ObjectValueTreeView.RefreshRow+13
            MonoDevelop.Debugger.dll!ObjectValueTreeView.OnButtonPressEvent+89
            40f46e55-eb67-454a-b904-dc9e5644a3a1!6003ba7+16 (UNKNOWN MVID)
```

Ensure that object value can never be null when calling UnregisterValue. It was
checked from some call sites previously, but not all.

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValueTreeView.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValueTreeView.cs
@@ -980,6 +980,8 @@ namespace MonoDevelop.Debugger
 
 		void RefreshRow (TreeIter iter, ObjectValue val)
 		{
+			if (val == null)
+				return;
 			UnregisterValue (val);
 			
 			RemoveChildren (iter);
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValueTreeView.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValueTreeView.cs
@@ -1013,8 +1013,7 @@ namespace MonoDevelop.Debugger
 
 			while (store.IterChildren (out citer, iter)) {
 				var val = (ObjectValue) store.GetValue (citer, ObjectColumn);
-				if (val != null)
-					UnregisterValue (val);
+				UnregisterValue (val);
 				RemoveChildren (citer);
 				store.Remove (ref citer);
 			}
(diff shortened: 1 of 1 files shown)


nosami:
cc: @leculver

mrward:
If val can be null when RefreshRow is called - then should we also handle a null val in RefreshRow? It seems to be used after the UnregisterValue call later on in the RefreshRow method.

It seems like maybe this should be still be called - https://github.com/mono/monodevelop/blob/16e55901a9635a31f7b101db379e2842b0b941c6/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValueTreeView.cs#L983-L989

Also looking at where the RefreshRow method is called it seems that we already have the val. Not sure why we are getting it again in RefreshRow - could it be passed instead?

nosami:
@mrward yes, good catch. I'll pass the value in too.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValueTreeView.cs
```diff
@@ -1904,7 +1906,7 @@ protected override bool OnButtonPressEvent (Gdk.EventButton evnt)
 					var val = (ObjectValue)store.GetValue (it, ObjectColumn);
 					if (DebuggingService.ShowValueVisualizer (val)) {
 						UpdateParentValue (it);
-						RefreshRow (it);
```
DavidKarlas: I'm not sure if this is same code as before, because if `if (DebuggingService.ShowValueVisualizer (val)) {` make changes I'm not 100% if `RefreshRow (it, val);` will display old or new value.

mrward: Needs testing. Presumably the visualizer would only be updating the ObjectValue it gets passed not creating a new one?

nosami: I just tested this to be doubly sure. RefreshRow displays the updated value.

debugger-update

PR #7097, opened 2019-02-08
Defensive programming in Dock Container

nosami:
MERP / Dr Watson show this as one of the most frequent crashes in VSMac

```
    Fault:       Crash
    thread 0:
        IsManaged:   False
        NativeFrames:0
        Managed Frames:
            MonoDevelop.Ide.dll!DockContainer.UpdatePlaceholder+20
            MonoDevelop.Ide.dll!DockFrame.UpdatePlaceholder+0
            MonoDevelop.Ide.dll!DockItemTitleTab.OnMotionNotifyEvent+d0
            40f46e55-eb67-454a-b904-dc9e5644a3a1!6003c23+16 (UNKNOWN MVID)
```

which points to `placeholderWindow.AllowDocking = allowDocking;` as the line
that is failing.

It isn't obvious at all how placeholderWindow could ever be null at this point.
(I don't see access to the placeholderWindow field from threads other than the
UI), so I have just made the method more robust.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockContainer.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockContainer.cs
@@ -1,5 +1,4 @@
 //
-// DockContainer.cs
 //
 // Author:
 //   Lluis Sanchez Gual
@@ -36,6 +35,7 @@ using Gtk;
 using Gdk;
 using System.Linq;
 using MonoDevelop.Components.AtkCocoaHelper;
+using MonoDevelop.Core;
 using MonoDevelop.Ide.Gui;
 
 namespace MonoDevelop.Components.Docking
@@ -403,7 +403,11 @@ namespace MonoDevelop.Components.Docking
 		
 		internal bool UpdatePlaceholder (DockItem item, Gdk.Size size, bool allowDocking)
 		{
-			MonoDevelop.Core.Runtime.AssertMainThread ();
+			if (!Runtime.IsMainThread) {
+				var msg = "UpdatePlaceholder called from background thread.";
+				LoggingService.LogInternalError ($"{msg}\n{Environment.StackTrace}", new InvalidOperationException (msg));
+			}
+
 			var placeholderWindow = this.placeholderWindow;
 			var padTitleWindow = this.padTitleWindow;
 
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockContainer.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockContainer.cs
@@ -403,6 +403,7 @@ namespace MonoDevelop.Components.Docking
 		
 		internal bool UpdatePlaceholder (DockItem item, Gdk.Size size, bool allowDocking)
 		{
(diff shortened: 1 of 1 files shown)


nosami:
cc @leculver

re:https://github.com/mono/monodevelop/blob/cb6407da5fec9b68997cf41896140d547fadb987/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockContainer.cs#L406-L412

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockContainer.cs
```diff
@@ -52,8 +52,8 @@ class DockContainer: Container
 
 		bool needsRelayout = true;
 
-		PlaceholderWindow placeholderWindow;
-		PadTitleWindow padTitleWindow;
+		volatile PlaceholderWindow placeholderWindow;
+		volatile PadTitleWindow padTitleWindow;
```
Therzok: Why volatile here? These should only be accessed/changed on the UI thread.

nosami: Because, it seems that _somehow_ that they are accessed on a background thread. Why, I can't see.

But that seems to be the only explanation why placeholderWindow could ever be null here on line 412 (and I can't think of any other way for this line to fail)

https://github.com/mono/monodevelop/blob/cb6407da5fec9b68997cf41896140d547fadb987/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockContainer.cs#L406-L412

Therzok: @nosami we should add some mainthread assertions then. It's not safe to use Gtk# widgets on a background thread.

leculver: Agreed on the assertions.

Note the `volatile` markers _shouldn't_ be doing anything here, but one of the challenges with this failure is that we know _something_ is causing VS for Mac to crash on `placeholderWindow.AllowDocking = allowDocking;`.  From everything we looked at, this should be impossible...and yet it's still happening.  This change essentially lets us narrow down the failure by eliminating things that should be impossible...but theoretically might be the cause (such as someone calling HidePlaceholder on a non-UI thread).

This change is doing exactly that:  Making what we _think_ is impossible _actually_ be impossible...and if we still get failures on this code we will use that information to try again to figure out what the next thing could be here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockContainer.cs
```diff
@@ -403,7 +403,11 @@ internal void ShowPlaceholder (DockItem draggedItem)
 		
 		internal bool UpdatePlaceholder (DockItem item, Gdk.Size size, bool allowDocking)
 		{
-			if (placeholderWindow == null)
+			MonoDevelop.Core.Runtime.AssertMainThread ();
```
Therzok: Should we rather just log when it happens from a background thread rather than asserting? We're replacing a NRE with a main thread misuse exception, so we're not fixing the crasher.

Maybe LogInternalError if we're not on the UI thread with a stacktrace.

PR #7111, opened 2019-02-11
Fixes VSTS Bug 769480: Code fix preview window is too small

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/769480

Resize was missing.


mkrueger:
@monojenkins

PR #7115, opened 2019-02-11
Fixes VSTS Bug 769480: Code fix preview window is too small

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/769480

Resize was missing.


mkrueger:

main/src/addins/MonoDevelop.Refactoring/MonoDevelop.Refactoring/RefactoringPreviewTooltipWindow.cs
```diff
@@ -135,6 +135,7 @@ void RequestPopup (Xwt.Rectangle rect)
 						if (temp.LineKind == LineKind.Normal && temp.TextMarkup == null) {
 							var newText = editor.GetMarkup (temp.Offset, temp.Length, new MarkupOptions (MarkupFormat.Pango, false));
 							diff.LineResults [i] = new LineResult (newText, temp.Offset, temp.Length, temp.LineKind, temp.XNeedsMeasure);
+							QueueResize ();
```
Therzok: Why not resize at the end though?

mkrueger: Doesn't matter much hof often resize is queued - it's only resizing once.

Therzok: This seems like the kind of code that would end up called a lot from the UI thread. Regardless, looks okay to me.

PR #7123, opened 2019-02-11
Update TestLogSynchronizationContextThroughWrites to work against both old and new BCL TextWriter implementations

kg:
This should make the test able to handle the difference in call counts that is introduced by updating to a newer version of mono. I wasn't able to identify what specific file changed (since we have 8000 different TextWriters in-tree) but the change in call counts matches what I see in the corefx source.

Hard to test locally so MD CI should tell me whether this works against older BCL, and then the other CI I hit the bug on will tell me whether I've fixed the issue...


kg:
```
==== SUBMODULE VERSION MISMATCH ====
Submodule xwt is problematic
```
What does this mean? How do I fix it? Is it just a flake due to the state of master when I ran my tests?

Issue #7104, opened 2019-02-11
[VersionControl] Error fetching from Git server


jsuarezruiz:
Fetching sometimes gives an error or without any error never ends.

```
| at MonoDevelop.VersionControl.Git.GitConfigurationDialog.OnButtonFetchClicked (System.Object sender, System.EventArgs e) [0x000af] 
in /Users/jsuarezruiz/Projects/MonoDevelop/monodevelop/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitConfigurationDialog.cs:375<br>  
at System.Runtime.CompilerServices.AsyncMethodBuilderCore+<>c.<ThrowAsync>b__6_0 (System.Object state) [0x00000] 
in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-08/external/bockbuild/builds/mono-x64/mcs/class/referencesource/mscorlib/system/runtime/compilerservices/AsyncMethodBuilder.cs:1023<br>  
at MonoDevelop.Ide.DispatchService+GtkSynchronizationContext+TimeoutProxy.HandlerInternal (System.IntPtr data) [0x00016] 
in /Users/jsuarezruiz/Projects/MonoDevelop/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/DispatchService.cs:77<br><br>----------<br><br>  
ERROR [2019-02-11 09:24:31Z]: An unhandled exception has occurred. 
Terminating MonoDevelop? False<br>System.NullReferenceException: Object reference not set to an instance of an object<br>  
at MonoDevelop.Components.TreeViewState.Save () [0x0000c] 
in /Users/jsuarezruiz/Projects/MonoDevelop/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/TreeViewState.cs:59 <br>  
at MonoDevelop.VersionControl.Git.GitConfigurationDialog.FillRemotes () [0x0000e] 
in /Users/jsuarezruiz/Projects/MonoDevelop/monodevelop/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitConfigurationDialog.cs:147 <br>  
at MonoDevelop.VersionControl.Git.GitConfigurationDialog.OnButtonFetchClicked (System.Object sender, System.EventArgs e) [0x00101] 
[...]

PR #7128, opened 2019-02-12
[VersionControl] Fixed error fetching from Git server

jsuarezruiz:
We get an error trying to fetch the path to a remote branch. If a branch is selected, we will look for the parent element to obtain the appropriate path.

Fixes gh #7104


jsuarezruiz:
@sevoku I squashed the commits and, by the way, I changed the comment to the good one.

PR #7147, opened 2019-02-14
Fixes VSTS Bug 790861: Text not redrawn when auto-formatting

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/790861

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
@@ -203,20 +203,11 @@ namespace Microsoft.VisualStudio.Platform
 			var handler = _highlightingStateChanged;
 			if (handler != null) {
 				foreach (Mono.TextEditor.MdTextViewLineCollection.MdTextViewLine line in textView.TextViewLines) {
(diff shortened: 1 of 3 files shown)


mkrueger:
@monojenkins backport master

mkrueger:
Fixes the issue as well is a perf. improvement. There is basically no reason to check it that way, setting a flag there is easier.

main/src/addins/MonoDevelop.SourceEditor2/VSEditor/TagBasedSyntaxHighlighting.cs
```diff
@@ -203,20 +203,11 @@ private void OnClassificationChanged (object sender, ClassificationChangedEventA
...
-						handler (this, new LineEventArgs (line.line));
-						continue;
-					}
-					for (int i = 0; i < oldSegments.Count; i++) {
-						if (newSegments [i].ColorStyleKey != oldSegments [i].ColorStyleKey) {
-							handler (this, new LineEventArgs (line.line));
-							break;
-						}
-					}
+					if (!line.HasDrawn) {
```
Therzok: What resets this though?

mkrueger: Basically that fires a redraw when it's newly created - default value is false.

Issue #7119, opened 2019-02-11
[VersionControl] Error committing changes of removed files (Git)


jsuarezruiz:
After deleting files, and try to commit the changes.
```
ERROR [2019-02-11 08:45:15Z]: Version control operation failed: 
LibGit2Sharp.EmptyCommitException: No changes; nothing to commit.
  at LibGit2Sharp.Repository.Commit (System.String message, LibGit2Sharp.Signature author, LibGit2Sharp.Signature committer, LibGit2Sharp.CommitOptions options) [0x000dd] in /Users/jsuarezruiz/Projects/MonoDevelop/monodevelop/main/external/libgit2sharp/LibGit2Sharp/Repository.cs:1084 
  at LibGit2Sharp.RepositoryExtensions.Commit (LibGit2Sharp.IRepository repository, System.String message, LibGit2Sharp.Signature author, LibGit2Sharp.Signature committer) [0x00001] in /Users/jsuarezruiz/Projects/MonoDevelop/monodevelop/main/external/libgit2sharp/LibGit2Sharp/RepositoryExtensions.cs:293 
  at MonoDevelop.VersionControl.Git.GitRepository.OnCommit (MonoDevelop.VersionControl.ChangeSet changeSet, MonoDevelop.Core.ProgressMonitor monitor) [0x000dc] in /Users/jsuarezruiz/Projects/MonoDevelop/monodevelop/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs:867 
  at MonoDevelop.VersionControl.Repository.Commit (MonoDevelop.VersionControl.ChangeSet changeSet, MonoDevelop.Core.ProgressMonitor monitor) [0x0005d] in /Users/jsuarezruiz/Projects/MonoDevelop/monodevelop/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs:593 
  at MonoDevelop.VersionControl.CommitCommand+CommitWorker.Run () [0x00081] in /Users/jsuarezruiz/Projects/MonoDevelop/monodevelop/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/CommitCommand.cs:70 
[...]

PR #7149, opened 2019-02-14
[VersionControl] Fix problem notifying deleted files (Git)

jsuarezruiz:
After deleting a file, and trying to make Commit the change with the deleted file does not appear.
It can cause problems when trying to commit without detecting the change.

Before:
delete-fails
Now:
delete-ok

Fixes gh #7119


main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlService.cs
```diff
@@ -513,29 +506,44 @@ static void OnFileAdded (object s, ProjectFileEventArgs e)
...
-			Repository repo = GetRepository (args.Project);
-			if (repo.IsVersioned (path) && repo.CanRemove (path)) {
-				repo.Remove (path);
-				NotifyFileStatusChanged (repo, path, args.ProjectFile.Subtype == Subtype.Directory);
+			foreach (var file in args) {
+				string path = file.ProjectFile.FilePath;
+				Repository repo = GetRepository (file.Project);
+				if (repo != null) {
+					using (ProgressMonitor monitor = GetStatusMonitor ()) {
+						repo.Update (path, false, monitor);
```
Therzok: This will cause a pull for every renamed file.

jsuarezruiz: Right, I just changed this part.

slluis:
Shouldn't this be handled through VersionControlFileSystemExtension ?

jsuarezruiz:
You're right, Lluis. `VersionControlFileSystemExtension` already manages all these events (add, delete files, modify folders, etc). In addition, it invokes `FileStatusChanged`, used by the `StatusView` to refresh the VCS changes. I have added more invocations  of FileStatusChanged in other `FileService` operations such as deleting files, etc.

slluis:
NotifyFilesChanged is meant to be called when a file is modified, not when it is deleted. Can't the status view just listen to file removed events?

main/src/core/MonoDevelop.Core/MonoDevelop.Core/FileService.cs
```diff
@@ -340,6 +340,8 @@ public static void NotifyFileRemoved (string fileName)
 		public static void NotifyFilesRemoved (IEnumerable<FilePath> files)
 		{
 			try {
+				foreach (var fsFiles in files.GroupBy (f => GetFileSystemForPath (f, false)))
```
Therzok: Why?

Therzok: We shoudl listen to FileService.FileRemoved in the StatusView.

slluis: Not really, it should listen to status change events from VersionControlService.

Therzok (dismissed):

slluis (changes requested):
I don't think the fix needs to be done in StatusView. StatusView already handles the case of a file that changes its status to Deleted. If you use the Remove command from the version control menu, the StatusView will be correctly updated. 

Physically deleting a file from the solution pad is a just a special case of the above. The file is both removed from version control and deleted from disk. The difference is that in the first case, the Remove command [calls](https://github.com/mono/monodevelop/blob/master/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/AddRemoveMoveCommand.cs#L135) `VersionControlService.NotifyFileStatusChanged()` to inform that the file has been removed from version control, and StatusView, which is subscribed to that event, updates the list correctly. In the second case, nobody is calling `NotifyFileStatusChanged()`.

So the question is, who should be calling `NotifyFileStatusChanged()` in the second case? We are currently using different approaches for notifying status changes. On one hand, the version control commands (such the Remove command I mentioned) make explicit calls to `NotifyFileStatusChanged()` when operations are completed. On the other hand, repository implementations call that method in different cases (for example GitRepository does it when stashing or switching branches). I think that notifying status changes should be the responsibility of repository implementations, since only repositories know when status really changes, and ensures that notifications will be properly sent no matter who initiated an operation.

And that's is really the source of the bug. The Remove command calls `DeleteFiles()` on the repository, and then calls `NotifyFileStatusChanged()`. When a file is deleted from the solution pad, `DeleteFiles()` ends being called through `VersionControlFileSystemExtension`, but nobody calls `NotifyFileStatusChanged()`. So the fix should be to just make the repository call `NotifyFileStatusChanged()` in `DeleteFiles()`, and that should cover both cases.

When we redesign the version control infrastructure we should make it more explicit that the responsibility of notifying status changes is on the repository.

jsuarezruiz:
Thanks for the detailed feedback Lluis. I have undone the previous changes by adding your suggestion, calling `NotifyFileStatusChanged()` in `DeleteFiles()` and iworks (ignoring the cache problems cases solved in another PR).

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlService.cs
```diff
@@ -513,29 +515,30 @@ static void OnFileAdded (object s, ProjectFileEventArgs e)
...
-			string path = args.ProjectFile.FilePath;
-			Repository repo = GetRepository (args.Project);
-			if (repo.IsVersioned (path) && repo.CanRemove (path)) {
-				repo.Remove (path);
-				NotifyFileStatusChanged (repo, path, args.ProjectFile.Subtype == Subtype.Directory);
+			foreach (var file in args) {
+				Repository repo = GetRepository (args.CommonProject);
+				if (repo != null) {
+					string path = file.ProjectFile.FilePath;
+					NotifyFileStatusChanged (new FileUpdateEventArgs (repo, path, file.ProjectFile.Subtype == Subtype.Directory));
```
slluis: This shouldn't be necessary, since it will be invoked by Repository.DeleteFiles

main/src/core/MonoDevelop.Core/MonoDevelop.Core/FileService.cs
```diff
@@ -340,6 +340,8 @@ public static void NotifyFileRemoved (string fileName)
...
+					fsFiles.Key.NotifyFilesDeleted (fsFiles);
```
slluis: This shouldn't be necessary.

jsuarezruiz:
To simplify the review and the historical, we will unify this PR with https://github.com/mono/monodevelop/pull/7314 in a new one.

jsuarezruiz:
New related PR: https://github.com/mono/monodevelop/pull/7345

PR #7157, opened 2019-02-14
[Tests] Fix Build Output feature toggle test

rodrmoya:
See https://github.com/xamarin/md-addins/pull/4201


mrward:
The BuildOutputEnabledInV8 test is failing. Does an md-addin build need to be re-run?

rodrmoya:
Yes, md-addins branch was built before this was pushed, so re-building both

PR #7158, opened 2019-02-14
New document/view architecture

slluis:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/595176
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/691254

**REVIEW WARNING: This PR is built on top of new-service-model (PR #7172), a branch that was initially part of this PR but that I moved to a different PR to make it easier to review.**

**Also, it is easier to review this PR commit by commit. Most of the meat is in the two first commits. The other commits are changes to migrate to the new model and to track api changes.**

This PR implements a new architecture for displaying views in the shell. The main changes of this architecture compared to the old one are:

* Introduction of a Model/View/Controller pattern. By having the data in a Model it will be possible to easily support displaying and editing a file in more than one view, even in different documents.
* Extensibility at document and controller level, so that add-ins can plug into file operations and provide custom commands.
* Proper support of split and tabbed views. Extensions can provide visual designers and previews, and the shell decides the best way to present them.

## Document models

A model is the data being shown in a view. DocumentModel is the base class for models that can be edited in a view. FileModel is a base class for file-based models. TextFileModel is the base class for models stored in text files. TextBufferFileModel is a model that stores data in a TextBuffer. DocumentModelRegistry is a service that can be used to share and get shared models.

## Document controllers

The new DocumentController and related classes replace ViewContent and BaseView content. A document controller is a class that implements the logic for loading, displaying and interacting with the contents of a document. The FileDocumentController class is a controller base class specialized in showing the content of files.

DocumentController supports extensions using the DocumentControllerClass. Extensions can hook into controller operations and provide custom UI.

## Document views

Document controllers can provide views using the DocumentView class and derived classes. Unlike the old ViewContent class, which created a Control, the DocumentController class creates a DocumentView. A DocumentView is a component of the shell that can show the content provided by a document controller. The DocumentViewContent subclass can show a Control, and the DocumentViewContainer can show a set of views using tabs or splits.

SdiWorkspaceWindow has changed to support the new DocumentView model. However, all the interaction between DocumentView and the shell implementation is now done through interfaces so that porting to a native toolkit will be easier in the future, and also makes the code easier to unit test.

There are several GtkShellDocumentView* classes that implement the functionality of DocumentView and related classes.

All old ViewContent implementations have been migrated to use DocumentModel and DocumentView.

## Shell documents

Document handling logic has been moved from Workbench to a new DocumentManager class.

The Document class is not a subclass of DocumentContext anymore. It has a DocumentContext property instead. All logic in charge of parsing documents has been moved to the new RoslynDocumentExtension class, a document controller extension that is created only when the document is parseable.

## Unit test infrastructure

The unit test infrastructure has been improved to make it easier to test functionality that depens on the document model. There are new mock classes for IShell and related interfaces. Unit tests now can use the [RequireService] attribute to force the initialization of a service before starting a test.

## TODO

* [ ] More unit tests
   * [ ] Test project reload for views
* [x] Proper UI for view collections that support tabs + split mode
* [x] Translation catalog is not yet initialized when start window is shown
* [ ] Colors of start window are wrong
* [x] Application menu has wrong label


main/src/core/MonoDevelop.Core/MonoDevelop.Core.Text/TextFileUtility.cs
```diff
@@ -285,17 +305,31 @@ static Stream CreateStream (FilePath tmpPath)
...
 		}
 
+		public static void WriteText (Stream stream, string text, Encoding encoding, bool writeBom)
+		{
+			if (writeBom) {
+				var bom = encoding.GetPreamble ();
+				if (bom != null && bom.Length > 0)
+					stream.Write (bom, 0, bom.Length);
+			}
+			var sw = new StreamWriter (stream, encoding);
```
Therzok: Might be better to wrap this in a using and set the `leaveOpen: true`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentModel.cs
```diff
@@ -0,0 +1,496 @@
...
+					LoggingService.LogError ("RaiseChanged failed", ex);
+				}
+			}
+		}
+
+		internal class DocumentModelData
+		{
+			ImmutableList<DocumentModel> linkedModels = ImmutableList<DocumentModel>.Empty;
+			Dictionary <Type, ModelRepresentation> representations = new Dictionary<Type, ModelRepresentation> ();
+			SemaphoreSlim dataLock = new SemaphoreSlim (1);
```
Therzok: Unless we set the maximum count too, releasing dataLock twice in a row will allow 2 concurrent accesses.

```suggestion
			SemaphoreSlim dataLock = new SemaphoreSlim (1, 1);
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentModel.cs
```diff
@@ -0,0 +1,496 @@
...
+			int changeVersion;
+			object changeLock = new object ();
+
+			public bool IsShared => linkedModels.Count > 1;
+
+			public bool IsLinked => Id != null;
+
+			public object Id { get; set; }
+
+			public async Task LinkNew (DocumentModel model)
```
Therzok: Should these take CancellationTokens?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentModel.cs
```diff
@@ -0,0 +1,496 @@
...
+					throw new InvalidOperationException ();
+
+				if (!representations.TryGetValue (type, out var representation)) {
+					representations [type] = representation = (ModelRepresentation)Activator.CreateInstance (type);
+					representation.DocumentModelData = this;
+				}
+				return representation;
+			}
+
+			public async Task<ModelRepresentation> GetRepresentation (Type type)
```
Therzok: Should these take CancellationTokens?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentModel.cs
```diff
@@ -0,0 +1,496 @@
...
+				} finally {
+					if (existingRep != null)
+						existingRep.WaitHandle.Release ();
+					dataLock.Release ();
+				}
+
+				return representation;
+			}
+
+			public async Task<bool> Unlink (DocumentModel model)
```
Therzok: Should these take CancellationTokens?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentModel.cs
```diff
@@ -0,0 +1,496 @@
...
+				}
+
+				if (targetRep.CurrentVersion == currentChangeVersion || sourceModel == null || sourceModel == targetRep)
+					return; // Not changed
+
+				try {
+					// Any operation that requires the lock of several representation must take
+					// the model lock first, to avoid deadlocks
+					if (!dataLocked) {
+						await dataLock.WaitAsync ();
```
Therzok: should we do these in parallel?

```
await Task.WhenAll (dataLock.WaitAsync(), targetRep.WaitHandle.WaitAsync(), sourceModel.WaitHandle.WaitAsync())`
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentModel.cs
```diff
@@ -0,0 +1,496 @@
...
+						targetRep.WaitHandle.Release ();
+						sourceModel.WaitHandle.Release ();
+					}
+				}
+				RaiseChangedEvent (targetRep.GetType ());
+			}
+
+			void RaiseChangedEvent (Type repType)
+			{
+				foreach (var m in linkedModels.Where (m => m.RepresentationType == repType))
```
Therzok: Can we just roll in the lambda here so we don't capture the local?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentModel.cs
```diff
@@ -0,0 +1,496 @@
...
+
+		internal class DocumentModelData
+		{
+			ImmutableList<DocumentModel> linkedModels = ImmutableList<DocumentModel>.Empty;
+			Dictionary <Type, ModelRepresentation> representations = new Dictionary<Type, ModelRepresentation> ();
+			SemaphoreSlim dataLock = new SemaphoreSlim (1);
+
+			ModelRepresentation lastChangedRepresentation;
+			int changeVersion;
+			object changeLock = new object ();
```
Therzok: readonly here to ensure lock integrity.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentModelRegistry.cs
```diff
@@ -0,0 +1,131 @@
...
+
+namespace MonoDevelop.Ide.Gui.Documents
+{
+	/// <summary>
+	/// Registry that keeps track of currently active and shared document models
+	/// </summary>
+	[DefaultServiceImplementation]
+	public class DocumentModelRegistry: Service
+	{
+		SemaphoreSlim dataLock = new SemaphoreSlim (1);
```
Therzok: same reasoning as the other semaphore.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/FileModel.cs
```diff
@@ -0,0 +1,168 @@
...
+
+			protected override Task OnLoad ()
+			{
+				content = File.ReadAllBytes (FilePath);
+				return Task.CompletedTask;
+			}
+
+			protected override void OnCreateNew ()
+			{
+				content = new byte [0];
```
Therzok: `Array.Empty<byte> ()`

main/tests/Ide.Tests/MonoDevelop.Ide.FindInFiles/SearchResultWidgetTests.cs
```diff
@@ -29,18 +29,19 @@
...
 
 namespace MonoDevelop.Ide.FindInFiles
 {
 	[TestFixture]
+	[RequireService (typeof(TextEditorService))]
 	class SearchResultWidgetTests : IdeTestBase
 	{
 		[SetUp]
-		public void SetUp ()
+		public async Task SetUp ()
```
Therzok: Empty method?

main/src/addins/MonoDevelop.RegexToolkit/MonoDevelop.RegexToolkit/ValidateActionCodeRefactoringProvider.cs
```diff
@@ -89,9 +89,9 @@ public sealed override async Task ComputeRefactoringsAsync(CodeRefactoringContex
 			
 			context.RegisterRefactoring (CodeAction.Create(
 				GettextCatalog.GetString("Validate regular expression"),
-				t2 => {
-					ShowRegexToolkitHandler.RunRegexWindow ().Regex = regex.Value.ToString ();
-					return Task.FromResult(document);
+				async t2 => {
+					(await ShowRegexToolkitHandler.RunRegexWindow ()).Regex = regex.Value.ToString ();
```
Therzok: Should we explicitly move to the UI thread here?

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.csproj
```diff
@@ -516,6 +515,7 @@
     <Compile Include="Gui\MonoDevelop.VersionControl.UrlBasedRepositoryEditor.cs" />
     <Compile Include="Gui\MonoDevelop.VersionControl.Views.DiffWidget.cs" />
     <Compile Include="Gui\MonoDevelop.VersionControl.Views.LogWidget.cs" />
+    <Compile Include="MonoDevelop.VersionControl\VersionControlDocumentController.cs" />
```
Therzok: This file is not in the PR

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentModel.cs
slluis: I don't think so. We need to take the locks always in the same order, to avoid deadlocks.

main/tests/Ide.Tests/MonoDevelop.Ide.Editor/DocumentReloadTests.cs
```diff
@@ -86,13 +86,10 @@ public void ShowFileChangedWarning (bool multiple)
...
-			public override Task Load (FileOpenInformation fileOpenInformation)
+			public async Task Load ()
 			{
-				var fileName = fileOpenInformation.FileName;
-				string text = text = TextFileUtility.ReadAllText (fileName, out Encoding encoding);
-				Document.Editor.Text = text;
-				ContentName = fileName;
-				return Task.FromResult (true);
+				var content = await TextFileUtility.ReadAllTextAsync (FilePath);
+				Document.Editor.Text = (await TextFileUtility.ReadAllTextAsync (FilePath)).Text;
```
Therzok:
```csharp
Document.Editor.Text = content.Text;
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
```diff
@@ -0,0 +1,917 @@
...
+				.Select (ex => ex.SourceExtensionNode.Id).ToList ();
+
+			var newExtensions = new List<DocumentControllerExtension> ();
+
+			ExtensionNode lastAddedNode = null;
+
+			// Ensure conditions are re-evaluated.
+			extensionContext = CreateExtensionContext ();
+
+			using (extensionChain.BatchModify ()) {
```
Therzok: :heart:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
```diff
@@ -0,0 +1,917 @@
...
+		}
+
+		/// <summary>
+		/// Called to check if this controller supports being assigned a model of the provided type
+		/// </summary>
+		/// <returns><c>true</c>, if can assign model was oned, <c>false</c> otherwise.</returns>
+		/// <param name="type">Type.</param>
+		protected virtual bool OnCanAssignModel (Type type)
+		{
+			return Model?.GetType () == type;
```
Therzok: Should we use IsAssignableFrom here for people who extend controllers?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
slluis: :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
slluis: I guess so.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
```diff
@@ -28,357 +28,268 @@
...
+			window.Document = this;
+
+			view = await controller.GetDocumentView ();
+			view.IsRoot = true;
+			window.SetRootView (view.CreateShellView (window));
+
+			LastTimeActive = DateTime.Now;
+			this.window.CloseRequested += Window_CloseRequested;
+
+			var workspace = Runtime.PeekService<RootWorkspace> ();
```
Therzok: Should we not do this when the service is initialized?

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.csproj
slluis: I accidentally added it in the service model PR: https://github.com/mono/monodevelop/pull/7172/commits/3f08768e1c48470d3c77365bf1225e2f48ad8fbc

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
slluis: Yes, I'll fix.

Therzok: I'm thinking we should also be returning a registration token from `RunWhenServiceInitialized` which we can dispose if we don't have the service initialized by the time the document is disposed.

Therzok: i.e.: https://github.com/mono/monodevelop/pull/7172/files#r257746754

The reasoning for this is that for example, the service callback list will keep the document alive due to the Func capturing this. So we need a way to clear that so the doc actually goes away.

slluis: Heh, I implemented the registration token before seeing your comment. Yes, the reason I used PeekService here was to avoid holding the document reference forever. In practice this will never happen, but using WhenInitialized and canceling it when disposing the document is the right thing to do.

slluis: Fixed by https://github.com/mono/monodevelop/pull/7172/commits/36b90fc92f1649013378b160ec602aab0111250f.

slluis: Well, not yet fixed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
slluis: I'm going to change this to return false by default. If model replacement is supported, the document controller should explicitly say so by overriding OnCanAssignModel.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentModel.cs
slluis: No, it this operation can't be cancelled.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentModel.cs
slluis: I don't think it is worth making this cancellable. It should be fast.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentModel.cs
slluis: Not worth it.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Text/TextFileUtility.cs
slluis: Fixed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentModel.cs
slluis: Fixed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentModel.cs
slluis: Fixed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/FileModel.cs
slluis: Fixed.

main/tests/Ide.Tests/MonoDevelop.Ide.FindInFiles/SearchResultWidgetTests.cs
slluis: Fixed.

main/tests/Ide.Tests/MonoDevelop.Ide.Editor/DocumentReloadTests.cs
slluis: Fixed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
slluis: Fixed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
slluis: Fixed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
```diff
@@ -0,0 +1,917 @@
...
+					IsReadOnlyChanged?.Invoke (this, EventArgs.Empty);
+				}
+			}
+		}
+
+		/// <summary>
+		/// Owner project or solution item
+		/// </summary>
+		/// <value>The owner.</value>
+		public WorkspaceObject Owner {
```
mhutch: What about documents with multiple owners e.g. documents that are in multiple projects?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
```diff
@@ -0,0 +1,917 @@
...
+					DocumentTitleChanged?.Invoke (this, EventArgs.Empty);
+				}
+			}
+		}
+
+		/// <summary>
+		/// Icon of the document tab
+		/// </summary>
+		/// <value>The stock icon identifier.</value>
+		protected string DocumentIconId {
```
mhutch: Shouldn't this be an `IconId `?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
```diff
@@ -0,0 +1,917 @@
...
+					accessibilityDescription = value;
+					AccessibilityDescriptionChanged?.Invoke (this, EventArgs.Empty);
+				}
+			}
+		}
+
+		/// <summary>
+		/// Icon of the document tab
+		/// </summary>
+		/// <value>The stock icon identifier.</value>
```
mhutch: Bad copy/paste

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
```diff
@@ -0,0 +1,917 @@
...
+			get {
+				CheckInitialized ();
+				if (extensionChain != null)
+					return extensionChain.GetAllExtensions ().OfType<DocumentControllerExtension> ().Select (c => c.ProjectReloadCapability).Append (OnGetProjectReloadCapability ()).Min ();
+				else
+					return OnGetProjectReloadCapability ();
+			}
+		}
+
+		public bool ShowNotification { get; internal set; }
```
mhutch: What does this do?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
```diff
@@ -0,0 +1,917 @@
...
+				model = value;
+				if (model != null)
+					model.Changed += Model_Changed;
+				OnModelChanged (oldModel, model);
+				RefreshExtensions ().Ignore ();
+				ModelChanged?.Invoke (this, EventArgs.Empty);
+			}
+		}
+
+		void Model_Changed (object sender, EventArgs e)
```
mhutch: Isn't this redundant?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
```diff
@@ -0,0 +1,917 @@
...
+
+			if (extensions.Count - defaultExts.Count > 0)
+				OnContentChanged ();
+		}
+
+		protected virtual void OnExtensionChainCreated ()
+		{
+		}
+
+		ExtensionContext CreateExtensionContext ()
```
mhutch: Shouldn't this context have some conditions - mimetype, file extension, project type, etc?

mhutch: Ah, I see, some filtering attributes are implemented on ExportDocumentControllerBaseAttribute. Would be nice to have project type and project capabilities (and `AppliesToAttribute`).

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentViewContainer.cs
```diff
@@ -0,0 +1,275 @@
...
+		/// <summary>
+		/// Raised when the active view of this container changes
+		/// </summary>
+		public EventHandler ActiveViewChanged;
+
+		/// <summary>
+		/// Gets or sets the display modes supported by this container
+		/// </summary>
+		/// <value>The supported modes.</value>
+		public DocumentViewContainerMode SupportedModes {
```
mhutch: This entire concept of display modes is very unclear to me. Do you have docs explaining how this works?

mhutch: Specifically, I'd like to know how display modes factor into the following cases:

* Android designer with tabs for source, split, and design
* Xaml previewer with tabs for source and split

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
slluis: This represents the currently active owner, which is what really matters. If the document belongs to more than one project, there will be a selector for the active owner.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
slluis: I want to make it clear that this is the icon that will be shown in the document tab, which may be different from an icon for the view tab (in case of a document with multiple views).

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
slluis: That's a TODO, not yet implemented.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
slluis: Ah yes, that's some left over code.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
slluis: Good point! I'll add those.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentViewContainer.cs
slluis: In the case of the designer and the previewer, it would set both tabs and split as supported modes. How those modes will be presented to the user is still TBD. One simple to implement option is to show the tabs, with an extra "Split" tab that shows source and design in a split, which is what the initial PR from David was doing. Another option is to have a button in the bottom bar for switching between tab and split modes. We'll have to discuss this with the UX board.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
mhutch: I mean the type, instead of `string`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentViewContainer.cs
mhutch: We had a fairly lengthy UX discussion before settling on the approach in David's PR, I'm not sure I want to reopen that can of worms :)

Is there any way to express that only "source" and "source/preview split" views are valid, and that the "preview" view cannot be displayed outside of the split? Or to express what the split view should be labeled?

slluis: No, there is no way to express that "preview" can't be displayed outside of the split. I don't see why add this restriction. Also, there is no way to indicate what the split label should be.

mhutch: Well, in the case of the XAML previewer for example, the preview by itself is not particularly useful. So makes sense to remove that tab, and call the split mode "Preview".

slluis: I still think that even in the XAML previewer case makes sense. But ok, let me think about it. It could be supported with an additional mode.

mhutch: I agree that it _could_ be useful, but that doesn't mean it's useful enough to justify adding additional complexity to the UI in the form of another button/option. As far as I know no-one's _asked_ for this feature, after all :)

main/tests/Ide.Tests/MonoDevelop.Ide.Gui.Documents/DocumentControllerTests.cs
```diff
@@ -0,0 +1,147 @@
...
+using System;
+using System.Threading.Tasks;
+using UnitTests;
+using MonoDevelop.Components;
+using System.Threading;
+
+namespace MonoDevelop.Ide.Gui.Documents
+{
+	[TestFixture]
+	public class DocumentControllerTests: IdeTestBase
```
Therzok: Nice

Therzok (dismissed):

slluis:
I'll be back!

PR #7164, opened 2019-02-15
[CSharpBinding] Fixed failing unit test.

mkrueger:


mkrueger:
This solution prevents to create a cached OptionsSet object and queries the "real" options every time. Should even use less memory for translating roslyn options from ours.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.OptionProvider/CSharpDocumentOptionsProvider.cs
```diff
@@ -87,12 +76,20 @@ static string GetPath(Document document)
 
 		internal class DocumentOptions : IDocumentOptions
 		{
-			readonly OptionSet optionSet;
+			readonly static IEnumerable<string> types = Ide.IdeServices.DesktopService.GetMimeTypeInheritanceChain (CSharpFormatter.MimeType);
```
Therzok: Can this change at runtime?

mkrueger: No

slluis:
I cherrypicked the fix into new-doc-model.

PR #7166, opened 2019-02-16
[SourceControl] Adds back button Blame window when show prior authors

netonjm:
In the Authors tab, there is a history of all the users that modified the file in the right panel.
When we press Show "authors before this change", our context has changed with new information, but there is no way to go back to the author's page.

This PR adds the toolbar with a Back button.

Fixes VSTS #790407 - Blame view: no way back in history (once you select to show blame before the current, you can’t come back)

https://devdiv.visualstudio.com/DevDiv/_queries/edit/790407/?triage=true

back

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/SelectRepositoryDialog.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/SelectRepositoryDialog.cs
@@ -69,6 +69,7 @@ namespace MonoDevelop.VersionControl.Dialogs
 				defaultPath = VersionControlDefaultPath;
 				entryFolder.Text = defaultPath;
 				buttonOk.Label = GettextCatalog.GetString ("_Checkout");
+				UpdateCheckoutButton ();
 			} else {
 				labelTargetDir.Visible = false;
 				boxFolder.Visible = false;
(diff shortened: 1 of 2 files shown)


main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/BlameWidget.cs
```diff
@@ -164,9 +188,22 @@ public BlameWidget (VersionControlDocumentInfo info)
 				QueueDraw ();
 			};
 			editor.DoPopupMenu = ShowPopup;
+
+			revertButton = new DocumentToolButton ("vc-revert-command", GettextCatalog.GetString ("Back to Authors"));
+			revertButton.GetNativeWidget<Gtk.Widget> ();
+			revertButton.Clicked += RevertRevisionClicked;
```
Therzok: Not unsubscribed?

sevoku:
This doesn't look like the right approach, showing/hiding the toolbar just for a button to reset the view is a bit confusing.
I think it'd be better if we had a command to navigate back, similar to `ShowBlameBefore`. It can be called `BlameCommands.ShowPreviousBlame` and every time `OnShowBlameBefore ()` is called, we would pack the current revision on a stack and pop+show it on `OnShowPreviousBlame ()`. The command would be disabled if the stack is empty and the stack would be cleared on `Reset ()`.

We could also implement global navigation here, but this is out of scope for now, something to consider for the big VCS redesign.

netonjm:
Continues in: 

https://github.com/mono/monodevelop/pull/7173

sorry about this new PR, seems after the rebase something was broken.

PR #7159, opened 2019-02-14
 Initial impl for reusable workspaces

Therzok:
This changes the implementation to use a Primary and Secondary workspaces definitions
to help us with leaking everything.

Currently, every MonoDevelopWorkspace is leaked via roslyn's DiagnosticService, as they
hold onto the workspace for event subscription.

Fixes VSTS #788777 - Change the MonoDevelopWorkspace model to have a primary and secondary workspaces

PR #7167, opened 2019-02-16
[Roslyn] Fixes multiple leaks in the workspace integration

Therzok:
This mitigates the leak that happen at the workspace level by making the leak really small (just the workspace instance with no data in it).

This should help with our current situation where we leak ~everything.

There's still a few leaks here and there, but I'll open a different PR for that, as I have automated the leak retention process using the stress test tool.

Fixes VSTS #788777 - Change the MonoDevelopWorkspace model to have a primary and secondary workspaces

Should supersede https://github.com/mono/monodevelop/pull/7159

### DIFF
--- a/main/tests/StressTest/MonoDevelop.StressTest/StressTestApp.cs
+++ b/main/tests/StressTest/MonoDevelop.StressTest/StressTestApp.cs
@@ -56,7 +56,7 @@ namespace MonoDevelop.StressTest
 		public int Iterations { get; set; } = 1;
 		ProfilerProcessor profilerProcessor;
 
-		const int cleanupIteration = int.MaxValue;
+		const int cleanupIteration = int.MinValue;
 
 		public void Start ()
 		{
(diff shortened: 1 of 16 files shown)


main/tests/StressTest/MonoDevelop.StressTest/StressTestApp.cs
```diff
@@ -89,7 +91,12 @@ public void Start ()
 			for (int i = 0; i < Iterations; ++i) {
 				scenario.Run ();
 				ReportMemoryUsage (i);
+				// This is to prevent leaking of AppQuery instances.
+				TestService.Session.DisconnectQueries ();
```
DavidKarlas: Maybe this should be called before `ReportMemoryUsage (i);`?

Therzok: Yes, we could do that.

PR #7172, opened 2019-02-18
New service model

slluis:
**REVIEW WARNING: This PR is not 100% complete and will not build since it is part of PR #7158, which changes the document/view architecture. I just created this PR to make the service model changes easier to review.**

**Also, it is easier to review this PR commit by commit.**

This PR implements a new service model for MonoDevelop. Until now we have a mix of approaches for implementing services. Some are implemented as instance classes, some as static classes, and there is no standard way of initializing them.

The new service model has three main goals:

* Homogenize how services are implemented, instantiated, initialized and accessed.
* Support asynchronous and on-demand initialization of services, to reduce and parallelize the of work that needs to be done at IDE startup.
* Improve support for unit testing by allowing discrete service initialization and using mock test services.

The model is very simple:

* The new ServiceProvider class provides methods for registering and getting references to services. 
* Services implement the interface `IService`, either directly or through the abstract class `Service`. The interface defines the async method `Task Initialize(ServiceProvider)` which is invoked the first time a service is requested.

In this new model all services are instances (no more static classes, with the exception of some special cases), and it is possible to get any service using the main service provider available in `Runtime.ServiceProvider`. The `Runtime` class also provides some shortcut methods for getting services. For example:

```
var fontService = await Runtime.GetService<FontService> ();
...
```

The ´GetService()` method can be used to get a service. Services are created on demand, so if the service has not been requested before, it will be created and initialized at this time.

Getting a service is an asynchronous operation (since it may involve initialization). This may not always be convenient, since services also need to be used from methods that are not asynchronous. To make use of services easier MonoDevelop.Ide defines the class IdeServices, which has references to all services that are initialized at startup. Referencing services from IdeServices is safe once the Ide is initialized. In some scenarios (such as in unit tests or when implementing a command line command), IdeServices still can be used, but will only have references to those services that have been initialized.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/487937.


main/src/addins/MacPlatform/Dialogs/MacOpenFileDialogHandler.cs
```diff
@@ -220,7 +220,7 @@ static bool FillViewers (List<FileViewer> currentViewers, NSPopUpButton button,
 				i++;
 			}
 			
-			foreach (var vw in DisplayBindingService.GetFileViewers (filename, null)) {
+			foreach (var vw in IdeServices.DisplayBindingService.GetFileViewers (filename, null).Result) {
```
Therzok: Can we make this async somehow?

slluis: I guess it can be done, but it requires changes in the dialog that are better done in another PR.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/BasicServiceProvider.cs
```diff
@@ -0,0 +1,214 @@
...
+		Dictionary<Type, object> servicesByType = new Dictionary<Type, object> ();
+		Dictionary<Type, List<object>> initializationCallbacks = new Dictionary<Type, List<object>> ();
+		bool disposing;
+
+		public override async Task<T> GetService<T> ()
+		{
+			CheckValid ();
+			Task<object> currentInitTask = null;
+
+			lock (services) {
```
Therzok: Does this mean we can't initialize services in parallel?

Therzok: Ah, nevermind, we can.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/BasicServiceProvider.cs
```diff
@@ -0,0 +1,214 @@
...
+					// Look in all registered services
+					service = services.OfType<T> ().FirstOrDefault ();
+					if (service == null) {
+						// Create a new service instance
+						LoggingService.LogInfo ("Creating service: " + typeof (T));
+						service = (T)Activator.CreateInstance (GetImplementationType (typeof (T)), true);
+						services.Add (service);
+						servicesByType [typeof (T)] = service;
+						if (service is IService serviceInstance) {
+							var timer = Stopwatch.StartNew ();
```
Therzok: Can we not use InstrumentationService here?

main/src/core/MonoDevelop.Core/MonoDevelop.Core/BasicServiceProvider.cs
```diff
@@ -0,0 +1,214 @@
...
+							var completionTask = new TaskCompletionSource<object> ();
+							initializationTasks [service] = completionTask;
+							serviceInstance.Initialize (this).ContinueWith (t => {
+								LoggingService.LogInfo ($"Service {typeof (T)} initialized in {timer.ElapsedMilliseconds} ms.");
```
Therzok: Timer can be skewed based on threadpool availability?

main/src/core/MonoDevelop.Core/MonoDevelop.Core/BasicServiceProvider.cs
```diff
@@ -0,0 +1,214 @@
...
+using System.Diagnostics;
+
+namespace MonoDevelop.Core
+{
+	/// <summary>
+	/// A basic implementation of a service provider
+	/// </summary>
+	public class BasicServiceProvider : ServiceProvider
+	{
+		List<object> services = new List<object> ();
```
Therzok:
```suggestion
		readonly List<object> services = new List<object> ();
```

To ensure lock integrity.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/BasicServiceProvider.cs
```diff
@@ -0,0 +1,214 @@
...
+			List<object> list;
+			lock (services) {
+				if (disposing)
+					return;
+				list = services.ToList ();
+				services.Clear ();
+				disposing = true;
+			}
+
+			await Task.WhenAll (list.OfType<IService> ().Select (s => s.Dispose ()));
```
Therzok:
```suggestion
			await Task.WhenAll (list.OfType<IService> ().Select (s => s.Dispose ()).ToArray ());
```

Therzok: Actually, this is Dispose. Not that necessary.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/BasicServiceProvider.cs
```diff
@@ -0,0 +1,214 @@
...
+			}
+		}
+
+		/// <summary>
+		/// Executes an action when a service is initialized
+		/// </summary>
+		/// <param name="action">Action to run</param>
+		/// <typeparam name="T">Service type</typeparam>
+		/// <remarks>This method does not cause the initialization of the service.</remarks>
+		public override void WhenServiceInitialized<T> (Action<T> action)
```
Therzok: As stated on the other PR, I think we need to pass an IDisposable registration which would remove the callback when disposed.

That way we can't run a document's registration after the document is disposed.

slluis: Fixed by https://github.com/mono/monodevelop/pull/7172/commits/36b90fc92f1649013378b160ec602aab0111250f.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/BasicServiceProvider.cs
slluis: Could be, but threadpool availability is just one more factor that can slow down service initialization, I don't think we need to special case it.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
```diff
@@ -203,22 +196,30 @@ static IdeApp ()
...
-			projectOperations = new ProjectOperations ();
+			await Runtime.GetService<ProjectOperations> ();
 			helpOperations = new HelpOperations ();
-			commandService = new CommandManager ();
-			ideServices = new IdeServices ();
+
+			await Runtime.GetService<TextEditorService> ();
+			await Runtime.GetService<NavigationHistoryService> ();
+			await Runtime.GetService<DisplayBindingService> ();
+
```
rodrmoya: Can these services be initialised in parallel? If so, maybe this could be changed to ```await Task.WhenAll (Runtime.GetService<TextEditorService> (), await Runtime.GetService<NavigationHistoryService> (), ...)```?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
```diff
@@ -203,22 +196,30 @@ static IdeApp ()
...
+		public static async Task Initialize (ProgressMonitor monitor)
 		{
 			// Already done in IdeSetup, but called again since unit tests don't use IdeSetup.
 			DispatchService.Initialize ();
 
+			commandService = await Runtime.GetService<CommandManager> ();
+			await Runtime.GetService<DesktopService> ();
+			await Runtime.GetService<FontService> ();
+			await Runtime.GetService<TaskService> ();
+
```
rodrmoya: ditto here, if they don't have dependencies between each other?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
```diff
@@ -334,6 +344,14 @@ static bool FMOpenTimerExpired (Action timeCompletion)
 			return false;
 		}
 
+		public static void BringToFront ()
+		{
+			Initialized += (sender, e) => {
+				if (!Ide.WelcomePage.WelcomePageService.HasWindowImplementation)
+					Workbench.Present ();
```
rodrmoya: Not sure what this is supposed to do, but if you want to check if the Welcome Window is visible, there is a `WelcomeWindowVisible` property in `WelcomePageService`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
Therzok: They're UI thread bound, so I think not. We're using the UI thread scheduler for this.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
Therzok: If we don't have a welcome window, show the workbench.

rodrmoya: Right 😂but when is this called?

Therzok: Oh, I meant this might be to circumvent the usage of the welcome page.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
slluis: I'll add a WhenAll, although there is little chance of parallelization since they run on the UI thread. In any case, a WhenAll won't hurt.

slluis: Fixed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
slluis: Done.

Therzok:
@slluis one idea comes to mind with this PR.

Can we remove startup handlers and add on service init handlers?

So addins can plug in handling when a service is initialized, not on IDE startup.

slluis:
> @slluis one idea comes to mind with this PR.
> 
> Can we remove startup handlers and add on service init handlers?
> 
> So addins can plug in handling when a service is initialized, not on IDE startup.

Maybe, but in practice it wouldn't make much of a difference, since all services end being initialized at startup.

Therzok:
Was already merged to master-vnext.

PR #7173, opened 2019-02-18
[SourceControl] Add a new command to go back  to previous blame

netonjm:
In the Authors tab, there is a history of all the users that modified the file in the right panel.

When we press Show "authors before this change", our context has changed with new information, but there is no way to go back to the author's page.

This PR adds a new command to go back  to previous blame

Fixes VSTS #790407 - Blame view: no way back in history (once you select to show blame before the current, you can’t come back)

https://devdiv.visualstudio.com/DevDiv/_queries/edit/790407/?triage=true

blame

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/BlameWidget.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/BlameWidget.cs
@@ -560,7 +560,7 @@ namespace MonoDevelop.VersionControl.Views
 			[CommandUpdateHandler (BlameCommands.ShowPreviousBlame)]
 			protected void OnUpdateShowPreviousBlame (CommandInfo cinfo)
 			{
-				cinfo.Enabled = history.Count == 0;
+				cinfo.Enabled = history.Count > 0;
 			}
 
 			protected override bool OnButtonReleaseEvent (EventButton evnt)
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/BlameWidget.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/BlameWidget.cs
@@ -532,12 +532,11 @@ namespace MonoDevelop.VersionControl.Views
 			protected void OnShowBlameBefore ()
 			{
 				var current = menuAnnotation?.Revision;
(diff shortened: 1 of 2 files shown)


main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/BlameWidget.cs
```diff
@@ -526,7 +532,9 @@ protected void OnShowLog ()
 			protected void OnShowBlameBefore ()
 			{
 				var current = menuAnnotation?.Revision;
-				Revision rev = current?.GetPrevious () ?? widget.info.History.FirstOrDefault ();
+				history.Push (current);
```
sevoku: this should happen after the null check later, because if `current?.GetPrevious () == null` we'd push the revision onto the stack without actually navigating anywhere.

netonjm: Fixed!

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/BlameWidget.cs
```diff
@@ -55,6 +56,9 @@ class BlameWidget : Bin
 			get {
 				return revision;
 			}
+			set {
```
Therzok:
```suggestion
			private set {
```

netonjm:
Fixed @sevoku 👍

main/src/addins/VersionControl/MonoDevelop.VersionControl/VersionControl.addin.xml
```diff
@@ -193,7 +193,9 @@
 		<Command id = "MonoDevelop.VersionControl.Views.BlameCommands.ShowLog"
 			_label = "S_how Log"/>
 		<Command id = "MonoDevelop.VersionControl.Views.BlameCommands.ShowBlameBefore"
-			_label = "Show _Authors prior to this change"/>
+			_label = "Show _Authors prior to this change"/>			
+		<Command id = "MonoDevelop.VersionControl.Views.BlameCommands.ShowPreviousBlame"
+			_label = "Show _Previous Blame"/>
```
sevoku: Let's call this `Back to Previous view`
```suggestion
			_label = "Back to _Previous view"/>
```

PR #7174, opened 2019-02-18
[Ide] For Multiplatform projects, adds multi solution config

jtorres:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/681741

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/TemplateWizard.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/TemplateWizard.cs
@@ -28,6 +28,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using MonoDevelop.Core;
 using MonoDevelop.Ide.Projects;
 using MonoDevelop.Projects;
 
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/TemplateWizard.cs
```diff
@@ -109,6 +110,35 @@ static IEnumerable<TemplateParameter> GetValidParameters (string parameters)
 
 		public virtual void ItemsCreated (IEnumerable<IWorkspaceFileObject> items)
 		{
+			var solution = (Solution)items.FirstOrDefault ();
```
Therzok:
```csharp
if (!(items.FirstOrDefault() is Solution solution)
    return; 
```

Therzok: Actually, what if the first item is not a Solution?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/TemplateWizard.cs
```diff
@@ -109,6 +110,35 @@ static IEnumerable<TemplateParameter> GetValidParameters (string parameters)
...
+		}
+
+		/// <summary>
+		/// Adds MultiStartupConfiguration when there are 
+		/// more than one project and one of them is a Backend project
+		/// </summary>
+		/// <param name="solution">Solution.</param>
+		void MultiProjectStartUp (Solution solution)
+		{
+			if (Parameters ["CreateBackEndProject"] != "True" || Parameters ["IncludeBackEndProject"] != "True")
```
mrward: MonoDevelop.Core.StringParsing has an extension method you can use here to do something like:

`!Parameters.GetBoolValue ("CreateBackEndProject") || !Parameters.GetBoolValue ("IncludeBackEndProject")`

Instead of comparing a string.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/TemplateWizard.cs
```diff
@@ -109,6 +110,35 @@ static IEnumerable<TemplateParameter> GetValidParameters (string parameters)
 
 		public virtual void ItemsCreated (IEnumerable<IWorkspaceFileObject> items)
 		{
+			var solution = (Solution)items.FirstOrDefault ();
+			if (solution == null)
+				return;
+
+			MultiProjectStartUp (solution);
+		}
```
rodrmoya: Just a small style thing, methods should probably contain a verb, so `CreateMultiProjectStartup` sounds better, if not, at first sight, this looks like a type.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Templates/TemplateWizard.cs
jtorres: Do we have templates like this?

mrward: If a solution is created then items will contain a single solution.

If a new project is added to a solution then items will contain a project. Could even contain more than one, for example when adding a new Forms project to an existing solution.

Based on the current New Project dialog implementation this code looks OK.

Therzok: Yes, but, the item will not be a `Solution`, but a `SolutionItem`, so that'll cause a InvalidCastException.

mrward: Good point 👍

PR #7177, opened 2019-02-18
[SourceControl] Hiddes merge tab in case of no conflicting file

netonjm:
Hiddes merge tab in case of no conflicting file

Fixes VSTS #790409 - Merge view: should be hidden if there is no merge pending for the current file
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/790409

conflict

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/MergeView.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/MergeView.cs
@@ -25,6 +25,7 @@
 // THE SOFTWARE.
 using MonoDevelop.Components;
 using MonoDevelop.Core;
+using System.Linq;
 
 namespace MonoDevelop.VersionControl.Views
 {
@@ -34,7 +35,8 @@ namespace MonoDevelop.VersionControl.Views
 	
 	class MergeView : BaseView, IMergeView
 	{
-		VersionControlDocumentInfo info;
+		readonly VersionControlDocumentInfo info;
+		readonly FileEventInfo fileEventInfo;
 		MergeWidget widget;
 		readonly MergeWidgetContainer widgetContainer;
 		readonly Gtk.Label NoMergeConflictsLabel;
@@ -44,6 +46,7 @@ namespace MonoDevelop.VersionControl.Views
 		public MergeView (VersionControlDocumentInfo info) : base (GettextCatalog.GetString ("Merge"), GettextCatalog.GetString ("Shows the merge view for the current file"))
 		{
 			this.info = info;
+			fileEventInfo = new FileEventInfo (info.Item.Path.FullPath, info.Item.IsDirectory);
 			widgetContainer = new MergeWidgetContainer ();
 			NoMergeConflictsLabel = new Gtk.Label () { Text = GettextCatalog.GetString ("No merge conflicts detected.") };
 			FileService.FileChanged += FileService_FileChanged;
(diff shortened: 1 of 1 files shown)


Therzok:
Maybe we should just show 'no merge conflicts detected', rather than hiding the tab.

If we get a merge conflict while the document is open, the subview will not be attached, so a file needs to be closed and reopened to handle that.

netonjm:
@Therzok , tell me what do you think about this approach 👍

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/MergeView.cs
```diff
@@ -36,34 +36,111 @@ class MergeView : BaseView, IMergeView
...
+			if (isConflicted) {
 				if (widget == null) {
 					widget = new MergeWidget ();
 					widget.Load (info);
 				}
-				
-				return widget;
+
+				if (widgetContainer.Content != widget) {
+					FileService.FileChanged -= FileService_FileChanged;
```
sevoku: This looks a bit fragile to me. To avoid double subscription or NREs, we could just subscribe in the constructor (unsubscribe on dispose) and add a check in FileService_FileChanged. Also while this view is open and the `NoMergeConflictsLabel` shown, if a file status changes from/to conflicting, will the view be reloaded?

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/MergeView.cs
```diff
@@ -36,34 +36,103 @@ class MergeView : BaseView, IMergeView
...
+				}
 			}
 		}
 
-		public MergeView (VersionControlDocumentInfo info) : base (GettextCatalog.GetString ("Merge"), GettextCatalog.GetString ("Shows the merge view for the current file"))
+		void FileService_FileChanged (object sender, FileEventArgs e)
 		{
-			this.info = info;
+			//content is null when is deselected
+			if (widgetContainer.Content == null) {
```
sevoku: we should also check that our currently shown file has changed, and not reaload on any file change

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/MergeView.cs
```diff
@@ -36,34 +36,103 @@ class MergeView : BaseView, IMergeView
...
+			RefreshMergeEditor ();
 		}
 
 		protected override void OnSelected ()
 		{
-			widget.UpdateLocalText ();
-			widget.info.Start ();
+			info.Start ();
+			RefreshContent ();
+			RefreshMergeEditor ();
```
sevoku: Won't this discard all changes when the user switches between views?

sevoku: I think this is actually out of scope for now. The behavior didn't change (we did the same before). The whole thing needs to be rewritten, let's postpone that.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/MergeView.cs
netonjm: Fixed

sevoku:

PR #7179, opened 2019-02-19
Add more leak retention fixes

Therzok:
This brings down all ProjectFile and Document leaks to 0 on the stress test app.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
@@ -97,25 +97,37 @@ namespace MonoDevelop.Ide.Gui
 				workbench.WorkbenchTabsChanged += WorkbenchTabsChanged;
 				IdeApp.Workspace.StoringUserPreferences += OnStoringWorkspaceUserPreferences;
 				IdeApp.Workspace.LoadingUserPreferences += OnLoadingWorkspaceUserPreferences;
-				
-				IdeApp.FocusOut += delegate(object o, EventArgs args) {
-					if (!fileEventsFrozen) {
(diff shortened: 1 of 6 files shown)


Therzok:
Nice, this uncovered use-after-dispose issues.

Therzok:
Something is happening on CI making the IDE not have focus, thus failing the test. Rebasing.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectSystemHandler.cs
```diff
@@ -266,6 +266,9 @@ async Task<SolutionInfo> CreateSolutionInfoInternal (MonoDevelop.Projects.Soluti
 							return await CreateSolutionInfoInternal (solution, token).ConfigureAwait (false);
 						}
 
+						if (token.IsCancellationRequested)
+							return (solution, null);
```
alanmcgovern: why not `token.ThrowIfCancellationRequested ()`?

Therzok: I don't want to bring down the entire IDE :p

Therzok: This exception will not bubble to GetWorkspaceAsync, but this is the internal workspace loading mechanism and we don't want it to throw.

So return null here and handle that to pass on to the actual tcs which will set it's been canceled.

Therzok:
Closing for now. Will add pieces of this PR in other PRs.

PR #7181, opened 2019-02-19
[Core] Add a counter for installed user addins

mrward:
Reports the addin id and version for all installed and enabled user
addins.

Fixes VSTS #712478 - Add Telemetry to identify extensions installed

Currently this does not handle addins being disabled/enabled or installed/uninstalled.

The AddinManager.AddinUnloaded event can be used for disabling and uninstalling.

However AddinManager.AddinLoaded fires on startup or when an addin is lazily loaded. So it does not necessarily indicate an addin is being installed or enabled. It seems like we would need the AddinEngine's ActivateAddin to fire an event. That seems to be called by the AddinDatabase on installing a new addin.

Another possibility would be to cache the list of user installed addins and every time the AddinLoaded event is fired the AddinManager's Registry is checked to see if there is a difference. Although this would mean using memory just for a counter that might not be used during an IDE session.

Or not as good, we could send an event on shutdown listing the installed addins at the end of the IDE session.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
@@ -123,7 +123,6 @@ namespace MonoDevelop.Core
 					AddinManager.Registry.Update (null);
 				setupService = new AddinSetupService (AddinManager.Registry);
 				Counters.RuntimeInitialization.Trace ("Initialized Addin Manager");
-				ReportUserAddins ();
 				
 				PropertyService.Initialize ();
 
@@ -154,25 +153,6 @@ namespace MonoDevelop.Core
 			} finally {
 				Counters.RuntimeInitialization.EndTiming ();
 			}
-		}
-
-		static void ReportUserAddins ()
-		{
-			Counters.UserAddins.Inc (1, null, GetUserAddinsMetadata ());
-		}
-
-		static IDictionary<string, object> GetUserAddinsMetadata ()
-		{
-			var metadata = new Dictionary<string, object> ();
-
-			foreach (Addin addin in AddinManager.Registry.GetModules (AddinSearchFlags.IncludeAddins | AddinSearchFlags.LatestVersionsOnly)) {
-				if (addin.IsUserAddin && addin.Enabled) {
-					string id = Addin.GetFullId (addin.Namespace, addin.LocalId, null);
-					metadata [id] = addin.Version;
-				}
-			}
-
-			return metadata;
 		}
 
 		static void RegisterAddinRepositories ()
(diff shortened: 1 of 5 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Core/Runtime.cs
```diff
@@ -154,7 +155,26 @@ public static void Initialize (bool updateAddinRegistry)
...
+			Counters.UserAddins.Inc (1, null, GetUserAddinsMetadata ());
+		}
+
+		static IDictionary<string, object> GetUserAddinsMetadata ()
+		{
+			var metadata = new Dictionary<string, object> ();
+
+			foreach (Addin addin in AddinManager.Registry.GetModules (AddinSearchFlags.IncludeAddins | AddinSearchFlags.LatestVersionsOnly)) {
+				if (addin.IsUserAddin && addin.Enabled) {
+					string id = Addin.GetFullId (addin.Namespace, addin.LocalId, null);
```
Therzok: I think this counts as PII. Should we hash them and have a list of well known IDs?

mrward: Doing the PII hashing in md-addins.

mrward:
Another idea here might be to check the user installed adds before and after the addin manager dialog is displayed (AddinManagerWindow) and send out a counter if they have changed. Seems to be multiple places where the addin manager dialog can be triggered - main menu, check for updates (AddinsUpdateHandler), opening an .mpack file.

mrward:
Changed this around a bit. Moved the counter to MonoDevelop.Ide. Also handle addins being installed/enabled/uninstalled/disabled from the Extension Manager.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Updater/AddinsUpdateHandler.cs
```diff
@@ -123,6 +128,68 @@ public static void HideAlert ()
...
+				oldUserAddins = GetUserAddinsList ();
+			}
+
+			List<string> GetUserAddinsList ()
+			{
+				var addins = new List<string> ();
+				foreach (Addin addin in GetEnabledUserAddins ()) {
+					addins.Add (addin.Id);
+				}
+				addins.Sort ();
```
Therzok: Do we need to sort it?

Therzok (approved):

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Updater/AddinsUpdateHandler.cs
mrward: Sorting it for the comparison. It does not look like they are sorted when returned from the AddinManager. Was assuming we could not rely on the order returned from the AddinManager when looking for changes.

mrward:
Implementing a different approach.

PR #7184, opened 2019-02-20
Freeze/Thaw events on using file generators

Therzok:
This change should allow us to control file rename events better, avoiding
a file being moved to its temporary file

Best when combined with #7012

Possibly
Fixes VSTS #767154 - [Feedback] Compile errors in non-active tabs can corrupt the active tab


Therzok:
Moved the freeze/thaw into CustomToolService.

Issue #7176, opened 2019-02-18
[VersionControl] Remove all the remotes source


jsuarezruiz:
If in a project using Git as VCS all remotes source are deleted, nothing happens.
The project is still under VCS, but without any remote.

In these cases, we can not allow to delete all the remotes (there always remains at least one); or the project is no longer under VCS.

Therzok:
You can still operate on a repository with no remote. You're comitting and doing everything locally.

Maybe the network commands (Fetch, Push) command should show that there's no remote configured, thus nowhere to push.

sevoku:
Yes, I agree with @Therzok , and we should show the Publish command again (in this case it won't initialize the repo, but only push it to a new remote)

jsuarezruiz:
Perfect. So, we must disable some network commands like push, fetch, update solution, etc. and show again the publish command.

jsuarezruiz:
Sent PR: https://github.com/mono/monodevelop/pull/7189

PR #7189, opened 2019-02-20
[Version Control] If the Repository has no remotes, disable remote commands and show publish command

jsuarezruiz:
If the Repository has no remotes, we already disabled options like push, etc.
With this changes we add the option to publish visible without any remotes.

Fixes gh #7176

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/PublishCommand.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/PublishCommand.cs
@@ -10,10 +10,16 @@ namespace MonoDevelop.VersionControl
 {
 	internal class PublishCommand 
 	{
-		public static bool Publish (WorkspaceObject entry, FilePath localPath, bool test)
+		public static bool Publish (VersionControlItemList items, WorkspaceObject entry, FilePath localPath, bool test)
 		{
-			if (test)
+			if (test) {
+				var canUpdate = !items.Any (it => it.VersionInfo.CanUpdate);
+
+				if (canUpdate)
+					return true;
+
 				return VersionControlService.CheckVersionControlInstalled () && VersionControlService.GetRepository (entry) == null;
+			}
 
 			List<FilePath> files = new List<FilePath> ();
 
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Commands.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Commands.cs
@@ -127,7 +127,16 @@ namespace MonoDevelop.VersionControl
 		{
 			return true;
 		}
-	}	
+	}
+
+	class PublishCommandHandler : SolutionVersionControlCommandHandler
+	{
+		protected override bool RunCommand (VersionControlItemList items, bool test)
+		{
+			VersionControlItem it = items [0];
+			return PublishCommand.Publish (items, it.WorkspaceObject, it.Path, test);
+		}
+	}
 
 	class UpdateCommandHandler: SolutionVersionControlCommandHandler
 	{
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlNodeExtension.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlNodeExtension.cs
@@ -465,7 +465,7 @@ namespace MonoDevelop.VersionControl
 				case Commands.Publish:
 					VersionControlItem it = items [0];
 					if (items.Count == 1 && it.IsDirectory && it.WorkspaceObject != null)
-						res = PublishCommand.Publish (it.WorkspaceObject, it.Path, test);
+					res = PublishCommand.Publish (items, it.WorkspaceObject, it.Path, test);
 					break;
 				case Commands.Annotate:
 					res = await BlameCommand.Show (items, test);
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/VersionControl.addin.xml
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/VersionControl.addin.xml
@@ -152,6 +152,7 @@
 		<Command id = "MonoDevelop.VersionControl.Commands.Publish"
 			_label = "_Publish in Version Control..."
 			icon = "vc-push"
+			defaultHandler="MonoDevelop.VersionControl.PublishCommandHandler"
 			description = "Publish actual project into repository."/>
 		<Command id = "MonoDevelop.VersionControl.Commands.Checkout"
 			defaultHandler = "MonoDevelop.VersionControl.CheckoutCommand"


main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/PublishCommand.cs
```diff
@@ -10,10 +10,16 @@ namespace MonoDevelop.VersionControl
 {
 	internal class PublishCommand 
 	{
-		public static bool Publish (WorkspaceObject entry, FilePath localPath, bool test)
+		public static bool Publish (VersionControlItemList items, WorkspaceObject entry, FilePath localPath, bool test)
 		{
-			if (test)
+			if (test) {
+				var canUpdate = !items.Any (it => it.VersionInfo.CanUpdate);
```
Therzok: I think a better name would be `canPublish`?

jsuarezruiz: I think it is better, changed.

slluis:
@sevoku can you review?

jsuarezruiz:
Thanks for the feedback @sevoku!. I have applied new changes. First of all, in the dialog to publish (disabled tabs, hidden combo to select VCS, etc.). 
screenshot 2019-03-05 at 12 35 37
On the other hand, in the `OnPublish` method from the `GitRepository` we avoid initialize the repository if it already exists and check if there are any pending local change to commit. Of course, we add the remote if it does not exist, etc.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/SelectRepositoryDialog.cs
```diff
@@ -75,6 +79,26 @@ public SelectRepositoryDialog (SelectRepositoryMode mode)
...
+				notebook.RemovePage (1);    // Remove "Registered Repositories" Tab.
+				hbox1.Visible = false;      // Hide VCS selector.
+
+				// Select "Git"
+				Gtk.TreeIter iter;
+				repCombo.Model.GetIterFirst (out iter);
+				do {
+					GLib.Value thisRow = new GLib.Value ();
+					repCombo.Model.GetValue (iter, 0, ref thisRow);
+					if ((thisRow.Val as string).Equals ("Git")) {
```
sevoku: This is not right, VCS should not know anything about Git or do Git specific logic. Looking at the dialog code, we could check `(thisRow.Val as string).Equals (repo.VersionControlSystem.Name)`.

But this looks a bit hacky, the dialog will still create a new repo instance etc.  A better approach would be to define a new `ConfigureRepository` dialog which would be used if repo is not initialized (==null). It would shown only the configuration widget from `vcs.CreateRepositoryEditor (repo)` and the optional commit entries (+Cancel/Ok buttons).

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
```diff
@@ -552,26 +552,34 @@ static void CollectFiles (List<FilePath> directories, FilePath dir, bool recursi
 				.Select (f => new FilePath (f)));
 		}
 
-		protected override Repository OnPublish (string serverPath, FilePath localPath, FilePath[] files, string message, ProgressMonitor monitor)
+		protected override Repository OnPublish (string serverPath, FilePath localPath, FilePath [] files, string message, ProgressMonitor monitor)
 		{
-			// Initialize the repository
-			RootRepository = new LibGit2Sharp.Repository (LibGit2Sharp.Repository.Init (localPath));
 			RootPath = localPath;
```
sevoku: Check `RootPath != localPath` and `RootRepository != null` first. If this repository object is already initialized, we should not replace the RootRepository object.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/SelectRepositoryDialog.cs
jsuarezruiz: We are using that line to ensure that we use Git as VCS since this case can only be given using Git. That line will not be necessary because in another Issue we are alphabetically ordering VCS. However, creating a new dialog seems to me the best option. Do we make the new dialog?

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
jsuarezruiz: Added validations.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/SelectRepositoryDialog.cs
jsuarezruiz: Ok. Added a new Dialog (`ConfigureRepositoryDialog`).

Screenshot 2019-03-18 at 16 02 55

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
```diff
@@ -826,42 +826,47 @@ static void CollectFiles (List<FilePath> directories, FilePath dir, bool recursi
...
 		protected override Repository OnPublish (string serverPath, FilePath localPath, FilePath[] files, string message, ProgressMonitor monitor)
 		{
+			if(!RootPath.Equals(localPath))
+				RootPath = localPath;
+
 			// Initialize the repository
-			RootPath = localPath;
-			RootRepository = new LibGit2Sharp.Repository (LibGit2Sharp.Repository.Init (localPath));
+			if (RootRepository == null) {
+				if (!RootPath.Combine (".git").IsDirectory)
```
netonjm: Add ".git" extension  like a constant, not sure but I think there are other constant of this extension somewhere

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
```diff
@@ -826,42 +826,47 @@ static void CollectFiles (List<FilePath> directories, FilePath dir, bool recursi
...
-						CredentialsProvider = (url, userFromUrl, types) => GitCredentials.TryGet (url, userFromUrl, types, credType)
-					});
-				} catch(VersionControlException vcex) {
+			// Push
+			RetryUntilSuccess (monitor, credType => RootRepository.Network.Push (RootRepository.Head, new PushOptions {
+				OnPushStatusError = delegate (PushStatusError e) {
 					RootRepository.Dispose ();
 					RootRepository = null;
 					if (RootPath.Combine (".git").IsDirectory)
 						Directory.Delete (RootPath.Combine (".git"), true);
```
netonjm: same here

jsuarezruiz: Yes, we had a constant in `GitVersionControl`. Changed!

jsuarezruiz:
have made rebase and squash of this branch. I have also removed changes that only related with code format (spaces, etc).

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
```diff
@@ -905,26 +905,38 @@ static void CollectFiles (List<FilePath> directories, FilePath dir, bool recursi
...
+				RootPath = localPath;
+
 			// Initialize the repository
-			RootPath = localPath;
-			RootRepository = new LibGit2Sharp.Repository (LibGit2Sharp.Repository.Init (localPath));
+			if (RootRepository == null) {
+				if (!RootPath.Combine (GitVersionControl.GitExtension).IsDirectory)
+					RootRepository = new LibGit2Sharp.Repository (LibGit2Sharp.Repository.Init (RootPath));
+
+				RootRepository = new LibGit2Sharp.Repository (RootPath);
```
mkrueger: Wouldn't that set the RootRepository a 2nd time ? Is an else missing here ?

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/PublishCommand.cs
```diff
@@ -1,42 +1,78 @@
...
 
-			if (files.Count == 0)
-				return false;
-	
+			if (!isInitialized) {
+				// Build the list of files to be checked in			
+				if (localPath == entry.BaseDirectory) {
+					GetFiles (files, entry);
+				} else if (entry is Project) {
+					foreach (ProjectFile file in ((Project)entry).Files.GetFilesInPath (localPath))
```
mkrueger: Style request: 
can you use more {} - it's ok for a 1 lined if but in a foreach we're always using blocks. Just a minor style issue.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/PublishCommand.cs
```diff
@@ -1,42 +1,78 @@
...
+			}
+
+			bool hasChanges = false;
+			if (repository != null) {
+				var directoryVersionInfo = repository.GetDirectoryVersionInfo (localPath, true, true);
+				if (directoryVersionInfo != null)
+					hasChanges = directoryVersionInfo.Any (vi => vi.HasLocalChanges);
+
+				if (hasChanges) {
+					foreach (var vi in directoryVersionInfo.Where (vi => vi.HasLocalChanges)) {
```
mkrueger: Minor perf issue - if you go through a collection with a foreach don't use Where only slows things down and doesn't improve readability.

In general all LINQ/Lambdas should be considered as slow. Doesn't make much difference in a command but in low level code it does.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/PublishCommand.cs
jsuarezruiz: Sure, thanks for the feedback. Changed.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/PublishCommand.cs
jsuarezruiz: I have applied changes to stop using LINQ in this case.

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitVersionControl.cs
```diff
@@ -35,7 +35,7 @@ abstract class GitVersionControl : VersionControlSystem
 		string version = null;
 		bool failedToInitialize;
 
-		const string GitExtension = ".git";
+		public const string GitExtension = ".git";
```
sevoku: Minor: shouldn't this be `internal` rather than `public`?

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
jsuarezruiz: Right!

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitVersionControl.cs
jsuarezruiz: Yes, can be `internal` and in this case is better. Changed.

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
```diff
@@ -907,26 +907,38 @@ static void CollectFiles (List<FilePath> directories, FilePath dir, bool recursi
...
-				cs.AddFile (fp);
-			}
 
-			// Create the initial commit
-			cs.GlobalComment = message;
-			Commit (cs, monitor);
+			if (files.Length > 0) {
+				// Add the project files
+				foreach (FilePath fp in files) {
+					LibGit2Sharp.Commands.Stage (RootRepository, RootRepository.ToGitPath (fp));
```
Therzok: Not a big issue, but we should probably pass all the strings in one go, so libgit2 does less processing.

There's [Stage(IRepository repo, IEnumerable<string> paths)](https://github.com/libgit2/libgit2sharp/blob/master/LibGit2Sharp/Commands/Stage.cs#L51)

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/ConfigureRepositoryDialog.cs
```diff
@@ -0,0 +1,220 @@
...
+using MonoDevelop.Core;
+using Xwt;
+
+namespace MonoDevelop.VersionControl.Dialogs
+{
+	public class ConfigureRepositoryDialog : Dialog
+	{
+		Repository repo;
+		IRepositoryEditor currentEditor;
+		bool hasChanges;
```
Therzok: should this be readonly? Asking because we unsubscribe the event handler based on this value, and if we can change, we can leave a dangling handler.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/ConfigureRepositoryDialog.cs
```diff
@@ -0,0 +1,220 @@
...
+
+		VBox repoContainer;
```
Therzok: Urgh, because we added all these fields, we will have to null them out too. Is there a way to refactor the gui creation code to avoid having so many fields?

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/ConfigureRepositoryDialog.cs
```diff
@@ -0,0 +1,220 @@
...
+		{
+			UpdateCheckoutButton ();
+		}
+
+		void OnCancel (object sender, EventArgs e)
+		{
+			Respond (Command.Cancel);
+		}
+
+		void OnModify (object sender, EventArgs e)
```
Therzok: OnModify does not actually show the intent of the handler:

```suggestion
		void OnOkClicked (object sender, EventArgs e)
```

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/PublishCommand.cs
```diff
@@ -50,7 +88,34 @@ public static bool Publish (WorkspaceObject entry, FilePath localPath, bool test
...
 				dlg.Dispose ();
 			}
-			return true;
+		}
+
+		static void ValidateRepository (Repository repository)
+		{
+			if (repository != null && repository is UrlBasedRepository urlBasedRepository) {
+				if (string.IsNullOrEmpty (urlBasedRepository.Url)) {
+					var tempRepository = repository.VersionControlSystem.CreateRepositoryInstance ();
```
Therzok: Could be simplified if we just wrap it in a using.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/ConfigureRepositoryDialog.cs
jsuarezruiz: Yeah, right, is better. Changed.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/ConfigureRepositoryDialog.cs
jsuarezruiz: Changed!

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/ConfigureRepositoryDialog.cs
```diff
@@ -0,0 +1,225 @@
...
+namespace MonoDevelop.VersionControl.Dialogs
+{
+	public class ConfigureRepositoryDialog : Dialog
+	{
+		readonly Repository repo;
+		readonly bool hasChanges;
+
+		IRepositoryEditor currentEditor;
+
+		VBox repoContainer;
```
Therzok: All these fields need to be nulled out or not to be fields.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/PublishCommand.cs
jsuarezruiz: Changed!

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/ConfigureRepositoryDialog.cs
jsuarezruiz: I have changed the creation of the visual elements a bit to avoid unnecessary fields.

Issue #7106, opened 2019-02-11
[VersionControl] Add validations creating branches in Git


jsuarezruiz:
There aren't validations creating branches, we can obtain errors depending on name length, etc.

```
ERROR [2019-02-08 10:51:42Z]: An unhandled exception has occurred. Terminating MonoDevelop? False
LibGit2Sharp.InvalidSpecificationException: invalid path for filesystem '/Users/jsuarezruiz/Projects/DefaultCollection/TipCalcCIGit/_git/TipCalcCIGit/.git/refs/heads/dsfsdfdsdsfert245906u430rieetñyu+çilñop+ç09+98op7uydsfsdfdsdsfert245906u430rieetñyu+çilñop+ç09+98op7uydsfsdfdsdsfert245906u430rieetñyu+çilñop+ç09+98op7uydsfsdfdsdsfert245906u430rieetñyu+çilñop+ç09+98op7uydsfsdfdsdsfert245906u430rieetñyu+çilñop+ç09+98op7uydsfsdfdsdsfert245906u430rieetñyu+çilñop+ç09+98op7uydsfsdfdsdsfert245906u430rieetñyu+çilñop+ç09+98op7uydsfsdfdsdsfert245906u430rieetñyu+çilñop+ç09+98op7uydsfsdfdsdsfert245906u430rieetñyu+çilñop+ç09+98op7uydsfsdfdsdsfert245906u430rieetñyu+çilñop+ç09+98op7uydsfsdfdsdsfert245906u430rieetñyu+çilñop+ç09+98op7uy': File name too long
  at LibGit2Sharp.Core.Ensure.HandleError (System.Int32 result) [0x00064] in /Users/jsuarezruiz/Projects/MonoDevelop/monodevelop/main/external/libgit2sharp/LibGit2Sharp/Core/Ensure.cs:155
  at LibGit2Sharp.Core.Ensure.ZeroResult (System.Int32 result) [0x0000c] in /Users/jsuarezruiz/Projects/MonoDevelop/monodevelop/main/external/libgit2sharp/LibGit2Sharp/Core/Ensure.cs:173
[...]

PR #7191, opened 2019-02-20
[VersionControl] Error creating branches 

jsuarezruiz:
The first thing I considered is to add validations in the name of branches. Libgit2 has a method for it: https://libgit2.org/libgit2/#HEAD/group/reference/git_reference_is_valid_name
But we are using it: http://source.monodevelop.com/#MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/EditBranchDialog.cs,126

With this changes, show an error message to notify the user that there has been a problem creating the branch.

Fixes gh #7106

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/EditBranchDialog.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/EditBranchDialog.cs
@@ -37,6 +37,8 @@ namespace MonoDevelop.VersionControl.Git
 {
 	partial class EditBranchDialog : Gtk.Dialog
 	{
+		const int GitRefnameMax = 1024;
+
 		readonly ListStore comboStore;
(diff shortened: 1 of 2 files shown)


slluis:
Is there any information we can extract from the exception that we can use to give more information to the user?

Therzok:
```
int | 1 if the reference name is acceptable; 0 if it isn't
-- | --
```

This is all that libgit2 supplies, no error message.

Therzok:
git_reference__normalize_name does set an error that has some message, might be better than is_valid_name

jsuarezruiz:
@sevoku I added the branch name max length.
screenshot 2019-02-21 at 10 36 30

Libgit2sharp return some information about the problems, like this one:
screenshot 2019-02-21 at 10 54 50

BTW, with large names, the size of the dialog grows to adapt. Playing with the limits, we can have a dialog that does not fit on the screen.
screenshot 2019-02-21 at 10 37 00

Added in this PR the dialog width size:
screenshot 2019-02-21 at 10 44 17

Issue #7194, opened 2019-02-21
Silent fatal crash opening a solution in MD (no error displayed and force closed)


netonjm:
Seems to be related to libgit2 from VersionControl

Git Log:
https://gist.github.com/netonjm/5076aa1512a184242005f92a2bfe57b6

Therzok:
Eh, what:
```
ERROR [2019-02-21 12:03:45Z]: Failed to write MEF cache
System.ArgumentException: An item with the same key has already been added. Key: Microsoft.GeneratedCode, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null
  at System.Collections.Generic.Dictionary`2[TKey,TValue].TryInsert (TKey key, TValue value, System.Collections.Generic.InsertionBehavior behavior) [0x002b5] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-10/external/bockbuild/builds/mono-x64/external/corefx/src/Common/src/CoreLib/System/Collections/Generic/Dictionary.cs:677 
  at System.Collections.Generic.Dictionary`2[TKey,TValue].Add (TKey key, TValue value) [0x00000] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-10/external/bockbuild/builds/mono-x64/external/corefx/src/Common/src/CoreLib/System/Collections/Generic/Dictionary.cs:240 
  at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement] (System.Collections.Generic.IEnumerable`1[T] source, System.Func`2[T,TResult] keySelector, System.Func`2[T,TResult] elementSelector, System.Collections.Generic.IEqualityComparer`1[T] comparer) [0x0008c] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-10/external/bockbuild/builds/mono-x64/external/corefx/src/System.Linq/src/System/Linq/ToCollection.cs:142 
  at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement] (System.Collections.Generic.IEnumerable`1[T] source, System.Func`2[T,TResult] keySelector, System.Func`2[T,TResult] elementSelector) [0x00000] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-10/external/bockbuild/builds/mono-x64/external/corefx/src/System.Linq/src/System/Linq/ToCollection.cs:100 
[...]

PR #7199, opened 2019-02-21
[MEF] Fix possible issue where generated code assemblies cause a mef …

Therzok:
…cache write failure

Seen in a log from #7194

Fixes VSTS 798008 - Possible duplicate assembly added in MEF assembly list


Therzok:
I should add a test for this

PR #7200, opened 2019-02-21
[Ide] Just let the content GrabFocus when document tab is activated

sevoku:
and don't SelectWindow, which would result in an infinite loop
activating the tab and selecting the window over and over again.

Fixes VSTS #797754


sevoku:

PR #7206, opened 2019-02-21
[VersionControl] Improve Repository Disposing

sevoku:
Fixes VSTS #797939

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs
@@ -89,10 +89,11 @@ namespace MonoDevelop.VersionControl
 				Dispose ();
 		}
 
-		internal bool Disposed { get; private set; }
-		public virtual void Dispose ()
+		public bool IsDisposed { get; protected set; }
+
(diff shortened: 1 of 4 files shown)


main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs
```diff
@@ -89,10 +89,11 @@ internal void Unref ()
...
 		}
 
-		internal bool Disposed { get; private set; }
-		public virtual void Dispose ()
+		public bool IsDisposed { get; protected set; }
+
+		protected virtual void Dispose (bool disposing)
 		{
-			Disposed = true;
+			IsDisposed = true;
```
Therzok: Maybe set IsDisposed in `Dispose()` so it's set when disposing?

sevoku: Hhm, I think having it here gives the subclass more control over the state, even without touching it directly. If the client depends on IsDisposed for some logic, it's ensured that it will enter the disposed state with `base.Dispose (bool)`.

Therzok: Sure!

PR #7210, opened 2019-02-21
[Roslyn] Fix NRE of solution when cancelling load

Therzok:
If a Workspace is disposed before load is finished, we would have a NRE on trying to add to the ConcurrentDictionary. Prevent that from happening.

Fixes VSTS 797986 - Unloading a solution while it's being loaded by the roslyn workspace can lead to NRE


Therzok:
@slluis green

PR #7217, opened 2019-02-22
[NuGet] Fix possible NullReferenceException

mrward:
Handle a project being disposed when checking to see if it supports
NuGet. A disposed project will have a null MSBuildProject property.

Fixes VSTS #797960 - Null reference in DotNetCoreNuGetProject


main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/DotNetCoreNuGetProject.cs
```diff
@@ -108,7 +108,11 @@ public DotNetCoreNuGetProject (DotNetProject project, ConfigurationSelector conf
 
 		public static bool CanCreate (DotNetProject project)
 		{
-			return project.MSBuildProject.GetReferencedSDKs ().Any ();
+			var msbuildProject = project.MSBuildProject;
```
Therzok: Ideally, everything would use the BindTask's cancellationToken property of the DotNetProject, but this works 👍

PR #7223, opened 2019-02-22
[Mac] Remove Network option from Preferences

mrward:
Use an environment variable instead of a options page in Preferences
to allow Mono's HttpClientHandler to be used instead of Xamarin.Mac's
NSUrlSessionHandler. Since this is an advanced setting providing
a way in the UI to change this is not required.

Fixes VSTS #796914 - Remove Network option from Preferences


main/src/addins/MacPlatform/MacPlatformSettings.cs
```diff
@@ -23,14 +23,17 @@
...
-using MonoDevelop.Core;
+using System;
 
 namespace MonoDevelop.MacIntegration
 {
 	static class MacPlatformSettings
 	{
-		public static readonly ConfigurationProperty<bool> UseNSUrlSessionHandler
-			= ConfigurationProperty.Create ("MonoDevelop.MacIntegration.UseNSUrlSessionHandler", true);
+		static MacPlatformSettings ()
```
Therzok: Do we need a static constructor? Can't we just inline the value?

mrward: Updated.

mrward:
Tried re-running the tests several times but the MacPlatform Tests seem to be timing out. Seem unrelated to this change. Also the tests are sometimes timing out on master.

slluis:

PR #7230, opened 2019-02-22
Fixes unexpected overlaps in pads with native contents in a DockGroup

netonjm:
When we attach some Pad which contains a native view to a DockGroup, we are not handling how this native view acts when the content is shown or hidden.

To try to normalize this I created a NativePadContent which encapsulates all this logic to reuse in other cases, right now implemented in the Toolbox.

clipissue

Fixes VSTS #790903 - Toolbox and Document Outline overlap each other
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/790903

### DIFF
--- a/main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport/MacToolboxPad.cs
+++ b/main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport/MacToolboxPad.cs
@@ -1,0 +1,146 @@
+﻿//
+// ToolboxPad.cs: The pad that hold the MD toolbox.
+//
+// Authors:
+//   Michael Hutchinson <m.j.hutchinson@gmail.com>
+//
+// Copyright (C) 2006 Michael Hutchinson
+//
+//
+// This source code is licenced under The MIT License:
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+#if MAC
+using System;
+using MonoDevelop.Ide.Gui;
+using MonoDevelop.Components;
+using Gtk;
+
+namespace MonoDevelop.DesignerSupport
+{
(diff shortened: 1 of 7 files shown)


main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport/ToolboxPad.cs
```diff
@@ -31,106 +31,145 @@
...
-using Xwt;
-#if MAC
-using MonoDevelop.Components.Mac;
-#endif
+using Gtk;
 
 namespace MonoDevelop.DesignerSupport
 {
-	public class ToolboxPad : PadContent
+	public class ToolboxPad :
```
sevoku: This looks quite hacky, shouldn't we have two separate classes for Gtk and Native?

netonjm: Fixed

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/NativePadContent.cs
```diff
@@ -0,0 +1,106 @@
...
+		{
+			if (widget != null) {
+			
+				widget.Destroy ();
+				widget.Dispose ();
+				widget = null;
+			}
+
+			if (window != null) {
+				window.PadContentShown -= OnPadContentShown;
```
sevoku: Set and check an IsDisposed flag or set `window = null` to avoid crasher in the unlikely case of double dispose.

main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport/MacToolboxPad.cs
```diff
@@ -0,0 +1,151 @@
...
+
+namespace MonoDevelop.DesignerSupport
+{
+	public class MacToolboxPad : NativePadContent
+	{
+		Widget widget;
+
+		protected override void Initialize (IPadWindow window)
+		{
+			base.Initialize (window);
```
Therzok: Unneeded.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/NativePadContent.cs
netonjm: good catch! do you recommend both or only with a window = null we are safe?

Therzok: Just set window to null.

main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport/MacToolboxPad.cs
netonjm: Fixed

main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport/ToolboxPad.cs
```diff
@@ -28,109 +28,31 @@
...
-using Xwt;
-#if MAC
-using MonoDevelop.Components.Mac;
-#endif
+using Gtk;
 
 namespace MonoDevelop.DesignerSupport
 {
-	public class ToolboxPad : PadContent
+	class ToolboxPad : PadContent
```
sevoku: Oh, this breaks the GtkSharp designer extension

slluis:
I don't think this approach is right. Why not just hide/show the widget in DockItem or DockContainerItem? I don't see the need for NativePadContent.

sevoku:
The idea was to abstract such things out to `NativePadContent` which actually makes much sense (to unify more embedding code in the future), but we didn't think about how much this breaks the API for designers. We should rethink Pad API in general for the next major release. With having different pad implementations we should replace `Workbench.GetPad<T>(className)` (which is too strict and static) with `Workbench.GetPad (string padId)` and hide the content implementations from the public APIs completely. This would allow us to use different content implementations for different situations much easier.

slluis:
We may improve the API in the future, but for now it is just a matter of the shell doing the right thing with native widgets. If native widgets needs to be explicitly hidden when deselected, let's handle it in the docking system.

netonjm:
This PR closes in favour of https://github.com/mono/monodevelop/pull/7247

PR #7232, opened 2019-02-24
[Ide] Don't show NPD in its "constructor"

rodrmoya:
Doing so means the dialog is visible before actually running it via
MessageService, which tries to retrieve the current focused window,
which would be the NPD dialog itself, completely breaking the
child window positioning code.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/796030


Therzok:
Does the old welcome page still work right?

rodrmoya:
> Does the old welcome page still work right?

yes, it should, nothing has changed there, if you run a monodevelop-only build (that is, no IWelcomeWindowProvider implementations found) it should display the old welcome page

rodrmoya:

rodrmoya:
Test failure seems unrelated

PR #7240, opened 2019-02-26
[Debugger] Disable debug commands when IDE window is not visible

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/801643


rodrmoya:

PR #7247, opened 2019-02-26
 Fixes hide/show native Toolbox when window shows/hides content 

netonjm:
We were not hiding the native view of the Toolbox when Window need show/hide contents

Fixes VSTS #790903 - Toolbox and Document Outline overlap each other
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/790903

clipissue

Instructions to test the Fix:
1) Open any project
2) Focus any other Pad window from the same container like the Toolbox
3) Press directly the auto-hide button from the toolbox window to de-attach to the container
4) Press for a second time the auto-hide button to attach the toolbox window to the container

Expected: the selected item is your first selected item and the toolbox is not overlapping


slluis:

PR #7255, opened 2019-02-27
[Core][Tests] Add basic ProgressMonitor tests

sevoku:
This PR adds more tests for progress monitors. Some tests are based on assumptions that are not necessarily right and need clarification.

### DIFF
--- a/main/tests/MonoDevelop.Core.Tests/MonoDevelop.Core/ProgressMonitorTests.cs
+++ b/main/tests/MonoDevelop.Core.Tests/MonoDevelop.Core/ProgressMonitorTests.cs
@@ -32,6 +32,7 @@ using System.Threading;
 using System.Text;
 using System.Collections.Concurrent;
 using UnitTests;
+using System.Linq;
 
 namespace MonoDevelop.Core
 {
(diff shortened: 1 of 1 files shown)


main/tests/MonoDevelop.Core.Tests/MonoDevelop.Core/ProgressMonitorTests.cs
```diff
@@ -154,6 +155,247 @@ public void TestLogSynchronizationContextThroughWrites ()
...
+				// 2 steps
+				monitor.Step (2);
+				stepped += 2;
+				Assert.That (childTask.Progress, Is.EqualTo (childStepProgress * stepped).Within (taskProgressTestPrecision));
+
+				expectedMainProgress += childStepProgress * stepped;
+				Assert.That (mainTask.Progress, Is.EqualTo (expectedMainProgress).Within (taskProgressTestPrecision));
+				Assert.That (monitor.Progress, Is.EqualTo (expectedMainProgress).Within (taskProgressTestPrecision));
+
+				monitor.EndTask ();
```
slluis: There should be a Step() call after EndTask()

sevoku: That's a main part of the problem I described in #7256, in order to reflect child progress in the monitor, this step should not be required and is not obvious. If this is required, we need to find a way to synchronize that.

Therzok:
Can we add the non-failing tests to master and keep the failing one in this PR?

PR #7258, opened 2019-02-27
[NetCore] Sets starting working directory according to VS on Windows

jtorres:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/788012


mrward:
Is this changing the working directory for StartActions.Program if it is not set? As long as this is consistent with what we do for non .NET Core console projects then it is OK.

Have we checked how ASP.NET Core projects work on Windows here? I did a quick test and Directory.GetCurrentDirectory is the project root directory for ASP.NET Core projects in VS 2017. So I think this change will break that behaviour.

mrward:
Or maybe this works already for ASP.NET Core projects since they have a different AspNetCoreExecutionCommand? I was assuming the working directory logic was shared.

slluis:

PR #7262, opened 2019-02-27
[Core] Handle missing language on saving shared assets project

mrward:
If a shared assets project was missing the import for its projItems
then this could result in a null reference exception when the project
was saved.

    <Import Project="Shared.projitems" Label="Shared" />

This was because the language is not set if the above import is
missing.

Handle a shared assets project type guid used for C# project

If the solution file is edited by hand and a shared assets project
entry was converted to a C# project file but the shared assets
project type guid is kept then the project would be loaded as a
shared assets project. To avoid this a check is made to see if the
project file extension is correct on reading the project file. If
not then an exception is thrown indicating there is a problem with
the project type guid. This avoids the project being treated as a
shared assets project incorrectly.

Fixes VSTS #802484 Unable to save Xamarin.Android-Tests.sln after adding
an existing shared project


slluis:

PR #7264, opened 2019-02-27
[Ide] Fix centering of dialogs over native NSWindow's

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/802604

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
@@ -423,16 +423,16 @@ namespace MonoDevelop.Ide
 				if (nsParent == null || !nsParent.IsVisible) {
 					nsChild.Center ();
 				} else {
-					int x = (int)(Math.Max (0, nsParent.Frame.Left + (nsParent.Frame.Width - nsChild.Frame.Width) / 2));
-					int y = (int)(Math.Max (0, nsParent.Frame.Top + (nsParent.Frame.Height - nsChild.Frame.Height) / 2));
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
```diff
@@ -444,9 +444,10 @@ static void CenterWindow (Window childControl, Window parentControl)
 					gtkChild.Move (x, y);
 #if MAC
 				} else if (nsParent != null) {
-					x = (int)(Math.Max (0, nsParent.Frame.Left + (nsParent.Frame.Width - w) / 2));
-					y = (int)(Math.Max (0, nsParent.Frame.Top + (nsParent.Frame.Height - h) / 2));
-					gtkChild.Move (x, y);
+					nsChild = GtkMacInterop.GetNSWindow (gtkChild);
+					x = (int) Math.Max (0, nsParent.Frame.Left + (nsParent.Frame.Width - w) / 2);
+					y = (int) Math.Max (0, nsParent.Frame.Top + (nsParent.Frame.Height - h) / 2);
+					nsChild.SetFrame (new CoreGraphics.CGRect (x, y, w, h), true);
```
sevoku: `w` and `h` won't take the window title into account here, because `gtkChild.GetSize (out var w, out var h)` returns the window size without decoration, resulting in a too small height:

grafik

I'd also use `SetFrameOrigin` to not mess around with the actual dialog size.

this is how MacPlatform does it: https://github.com/mono/monodevelop/blob/master/main/src/addins/MacPlatform/MacPlatform.cs#L1136

rodrmoya: done

rodrmoya:

PR #7265, opened 2019-02-28
Fixes Open current TextQuestionDialog using centralized RunCustomDialog

netonjm:
This PR fixes NRE trying to show the TextQuestionDialog because expects a Gtk window but it's getting an NSWindow.  

Just fixed GetParentForModalWindow to also check NSApplication.SharedApplication.ModalWindow and RunCustomDialog now calls CenterWindow.

Probably with this change is also **fixed focus the parent view in natives modal views**, because the parent was not well calculated.

Fixes Bug #802168 - Android app archive and publishing is broken - fails at MessageService.GetPassword.
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/802168

curioso

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
@@ -348,9 +348,14 @@ namespace MonoDevelop.Ide
 			Runtime.RunInMainThread (() => {
 				// If there is a native NSWindow model window running, we need
 				// to show the new dialog over that window.
-				if (NSApplication.SharedApplication.ModalWindow != null)
-					dialog.Shown += HandleShown;
-				else
+				if (NSApplication.SharedApplication.ModalWindow != null) {
+					EventHandler shownHandler = null;
+					shownHandler = (s,e) => {
+						ShowCustomModalDialog (dialog, parent);
+						dialog.Shown -= shownHandler;
+					};
+					dialog.Shown += shownHandler;
+				} else
 					PlaceDialog (dialog, parent);
 			}).Wait ();
 			#endif
@@ -373,11 +378,11 @@ namespace MonoDevelop.Ide
 		}
 
 		#if MAC
-		static void HandleShown (object sender, EventArgs e)
+		static void ShowCustomModalDialog (Gtk.Window dialog, Window parent)
 		{
-			var dialog = (Gtk.Window)sender;
-			var nsdialog = GtkMacInterop.GetNSWindow (dialog);
+			CenterWindow (dialog, parent);
 
+			var nsdialog = GtkMacInterop.GetNSWindow (dialog);
 			// Make the GTK window modal WRT the current modal NSWindow
 			var s = NSApplication.SharedApplication.BeginModalSession (nsdialog);
 
@@ -387,7 +392,6 @@ namespace MonoDevelop.Ide
 				dialog.Unrealized -= unrealizer;
 			};
 			dialog.Unrealized += unrealizer;
-			dialog.Shown -= HandleShown;
 		}
 		#endif
 		
@@ -596,7 +600,6 @@ namespace MonoDevelop.Ide
 					Caption = caption,
 					Value = initialValue,
 					IsPassword = isPassword,
-					TransientFor = parent ?? DesktopService.GetParentForModalWindow ()
 				};
 				if (dialog.Run ())
 					return dialog.Value;
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -970,7 +970,7 @@ namespace MonoDevelop.MacIntegration
 
 		public override Window GetParentForModalWindow ()
 		{
-			return NSApplication.SharedApplication.KeyWindow ?? NSApplication.SharedApplication.MainWindow;
+			return NSApplication.SharedApplication.ModalWindow ?? NSApplication.SharedApplication.KeyWindow ?? NSApplication.SharedApplication.MainWindow;
 		}
 
 		public override Window GetFocusedTopLevelWindow ()


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
```diff
@@ -594,12 +594,32 @@ public string GetTextResponse (Window parent, string question, string caption, s
 					Question = question,
 					Caption = caption,
 					Value = initialValue,
-					IsPassword = isPassword,
-					TransientFor = parent ?? DesktopService.GetParentForModalWindow ()
+					IsPassword = isPassword
+				};
+
+#if MAC
+				if (parent == null) {
```
sevoku: Shouldn't all this happen in [`TextQuestionDialog.RunDefault ()`](https://github.com/mono/monodevelop/blob/ad0f20e38619915304a7bde84013c72ee3c210b3/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Extensions/TextQuestionDialog.cs#L105) and not here?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
```diff
@@ -594,12 +594,32 @@ public string GetTextResponse (Window parent, string question, string caption, s
...
+					parent = NSApplication.SharedApplication.ModalWindow;
+				}
+#endif
+				if (parent == null) {
+					parent = DesktopService.GetParentForModalWindow ();
+				}
+
+				var customDialog = dialog.Create ();
+				customDialog.Shown += (sender, e) => {
+					CenterWindow (customDialog, parent);
```
sevoku: `RunCustomDialog` should do that, if it doesn't then it's a bug there. It shouldn't be necessary to call it here again. See also #7264

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Extensions/TextQuestionDialog.cs
```diff
@@ -62,40 +62,53 @@ public class TextQuestionDialog : PlatformDialog<TextQuestionDialogData>
 			get { return data.IsPassword; }
 			set { data.IsPassword = value; }
 		}
-		
+
+		public Gtk.Dialog Create ()
```
Therzok: Should this be public API?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
netonjm: Yes! agree!

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
netonjm: 👍

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Extensions/TextQuestionDialog.cs
```diff
@@ -62,40 +62,53 @@ public class TextQuestionDialog : PlatformDialog<TextQuestionDialogData>
...
-
-				md.DefaultResponse = ResponseType.Cancel;
-				
-				md.Child.ShowAll ();
-				
+				md = Create ();
 				var response = (ResponseType) MonoDevelop.Ide.MessageService.RunCustomDialog (md, TransientFor);
 				if (response == ResponseType.Ok) {
-					Value = responseEntry.Text;
 					return true;
```
rodrmoya: why was this moved to the `Response` event handler?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
rodrmoya: Yes please, just run everything via `RunCustomDialog`, it already has code to show, parent and centre the window correctly. In fact, this code should just create the dialog and call its `RunDefault`. No need to create the dialog here, right?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Extensions/TextQuestionDialog.cs
sevoku: no, it shouldn't

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
netonjm: Fixed

netonjm:
Hey guys I changed completely the approach could you take a look? @rodrmoya  @sevoku  @Therzok

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
```diff
@@ -348,9 +348,14 @@ public static int RunCustomDialog (Dialog dlg, Window parent)
...
 				// If there is a native NSWindow model window running, we need
 				// to show the new dialog over that window.
-				if (NSApplication.SharedApplication.ModalWindow != null)
-					dialog.Shown += HandleShown;
-				else
+				if (NSApplication.SharedApplication.ModalWindow != null) {
+					EventHandler shownHandler = null;
+					shownHandler = (s,e) => {
+						ShowCustomModalDialog (dialog, parent);
+						dialog.Shown -= shownHandler;
```
Therzok: What value of shownHandler is captured here? It might not unsubscribe the event?

netonjm: that's is a common pattern, no?

shownHandler is suscribed on shown:

https://github.com/mono/monodevelop/pull/7265/files/0e847640b8f5b6e33711ad6db6f31c678f7ebc3b#diff-e33cd0734d92a14cc69170d8404ffca3R353

and unsubscribed when finished:

https://github.com/mono/monodevelop/blob/0e847640b8f5b6e33711ad6db6f31c678f7ebc3b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs#L357 

it's not correct?

Therzok: Just tested a small app, it's working fine!

slluis:

PR #7275, opened 2019-02-28
Add TimeToCode timer with intellisense data 

Therzok:
Fixes VSTS #767147 - Add telemetry to measure time to intellisense available

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
@@ -347,9 +347,27 @@ namespace MonoDevelop.Ide
 			Counters.TimeToCode.Inc ("SolutionLoaded", ttcMetadata);
 
 			timeToCodeSolutionTimer = null;
+			timeToCodeWorkspaceTimer = Stopwatch.StartNew ();
+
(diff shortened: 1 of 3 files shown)


Therzok:
@rodrmoya @iainx I've tried this locally, and telemetry monitor isn't picking up TTC events.

That is because the open timer expired:
http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs,6b6434186b540a0e,references

What do you suggest doing here?

rodrmoya:
hmm, not sure, but I guess the 1st thing that comes to mind is to start the timeToCode timer only if we're opening a file?

iainx:
TimeToCode is only supposed to be for opening Solutions/Files from Finder. Not opening files from inside the IDE itself, that's why there's the timer, because when we start the IDE, we don't know if it was started by the user from Launchpad, or if it was started by Finder, so we wait 2.5seconds to see if we get the "Load a file" Apple Event.

If we want to have Intellisense timings available for all solutions it should be added to the `Ide.Shell.SolutionOpened` event maybe?

Therzok:
@iainx I think Lluis actually wants 2 timers for this, one like timetocode, one just to measure the actual monodevelopworkspace load

Therzok:
But regardless, TTC is completely busted.

Therzok:

PR #7278, opened 2019-03-01
[VersionControl] Fixes error in PasswordService; refuses to (re)store Git Credentials

jsuarezruiz:
Actually, the keychain implementation (macOS) has several issues:
- Missing protocols (especially "https" for Git and Azure DevOps)
- The URL Path part is not stored correctly

Due to these issues, we can not update an existing keychain item sometimes, and generate new ones on each query, the service always returns the first entry for a service. 

We have changed the implementation of KeyChain to use the security APIs of Xamarin.Mac directly.

Fixes #795844


sevoku:

@Therzok @slluis the tests pass now!

PR #7281, opened 2019-03-01
Fixes remove minimize and maximize from SelectRepositoryDialog + allways on top

netonjm:
When we were opening the SelectRepositoryDialog (Checkout menu), this window was appearing always on top of all other system windows, this was because we were not set correctly the Parent in this GTK.Dialog when the parent window was native because this parent was null.

Included the case in our RunCustomDialog from MessageService, to handle this in all dialogs with the same problem.

Also Removed Minimize and Maximize button from the window toolbar 

dialog

Fixes VSTS #801733 - [x] When cloning a git repo starting from GTC, SCC dialogs appear on top of other apps.
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/801733

Fixes VSTS #801737 - - [x] Select Repository doesn't re-appear after being hidden and reactivated (staring with GTC)
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/801737


sevoku:

PR #7287, opened 2019-03-01
Add more missing files for `make dist` to be valid

directhex:
It looks like my 7.7 changes from #6695 were never merged, so I rewrote them again from scratch when I ran into the same issues in 7.8 :man_shrugging:


Therzok:
Do we still want to merge this? If so, a monojenkins backport to master should make it actually run in VSTS

PR #7290, opened 2019-03-04
[NuGet] Make custom AvailableItemNames available after install

mrward:
Installing a NuGet package into a project that used PackageReferences
would not re-evaluate the project. This resulted in any custom
AvailableItemNames not being available to be used as a build action
in the Solution pad. Also needed to clear the cached build actions
on re-reading the project to ensure the latest items are available
after an re-evaluation.

Fixes VSTS #672198 - Users are unable to set the Build Action
"GoogleServicesJson" when adding a package that references Google
Play Services Basement and deploying an application


Therzok (approved):

mrward:

PR #7294, opened 2019-03-04
[AspNet] TypeScript files default BuildAction

jtorres:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/776769

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
@@ -790,6 +790,9 @@ namespace MonoDevelop.DotNetCore
 			if (globItems.Count == 1)
 				return globItems [0].Name;
 
+			if (Path.GetExtension (fileName).IndexOf (".ts", StringComparison.OrdinalIgnoreCase) == 0)
+				return BuildAction.TypeScriptCompile;
+
 			return base.OnGetDefaultBuildAction (fileName);
 		}
 	}
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
@@ -790,9 +790,6 @@ namespace MonoDevelop.DotNetCore
 			if (globItems.Count == 1)
 				return globItems [0].Name;
 
-			if (Path.GetExtension (fileName).IndexOf (".ts", StringComparison.OrdinalIgnoreCase) == 0)
-				return BuildAction.TypeScriptCompile;
-
 			return base.OnGetDefaultBuildAction (fileName);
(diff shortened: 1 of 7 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -790,6 +790,9 @@ protected override string OnGetDefaultBuildAction (string fileName)
 			if (globItems.Count == 1)
 				return globItems [0].Name;
 
+			if (Path.GetExtension (fileName).IndexOf (".ts", StringComparison.OrdinalIgnoreCase) == 0)
```
slluis: Special-casing typescript here doesn't look correct.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/BuildAction.cs
```diff
@@ -47,6 +47,7 @@ public static class BuildAction
 		public const string Resource = "Resource";
 		public const string SplashScreen = "SplashScreen";
 		public const string EntityDeploy = "EntityDeploy";
+		public const string TypeScriptCompile = "TypeScriptCompile";
```
slluis: This is specific to typescript support, the constant should not be defined here.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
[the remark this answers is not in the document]
jtorres: like this `	if (DesktopService.GetMimeTypeInheritanceChainForFile (fileName).Contains ("text/x-typescript"))` ?

jtorres: hmm why no?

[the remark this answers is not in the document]
Therzok: Yup

slluis: DotNetCoreProjectExtension is part of the core project system. Typescript support is provided by an extension. If we special case typescript here, it means that
1) The project system does not provide the ability to customize the behavior for this specific case. If that's is the case, then we need to fix the project system because if not, developers of other extensions won't be able to implement a similar behavior for other file types, and we can't be patching the project system every time an add-in developer needs some new behavior on their file types.
2) OR the project system provides the extensibility mechanism, in which case that should be used instead of special casing here.

jtorres: makes perfect sense 👍

slluis (approved):

rodrmoya (approved):

mrward (approved):

rodrmoya:

PR #7302, opened 2019-03-05
[NuGet] Raise install/uninstall events for PackageReference projects

mrward:
The NuGet addin API has a PackageReferenceAdded event and a
PackageReferenceRemoved. These were being raised if a project had
a packages.config file but not if the project used PackageReferences.

Fixes VSTS #807927 - Get the events working for an Android project
that uses PackageReferences


mrward:

PR #7303, opened 2019-03-05
[Ide] Fix progress dialog in front of confirmation dialog

mrward:
Adding an existing folder, that does not exist inside the project and
has more than 10 files, was opening a progress monitor dialog in front
of the confirmation dialog asking how to copy the files into the
project. A change to the MessageDialogProgressMonitor affected its
parent causing it to appear above the confirmation dialog. To fix this
the MessageDialogProgressMonitor now has a constructor that takes the
parent window.

Fixes VSTS #807518 - New ASP.NET Core Empty > .NET Core 2.x project,
add Existing Folder "bootstrap-4.3.1-dist", "Adding files..." dialog
in the front.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
@@ -1979,7 +1979,7 @@ namespace MonoDevelop.Ide
 			ProgressMonitor monitor = null;
 			
 			if (files.Length > 10) {
-				monitor = new MessageDialogProgressMonitor (true);
+				monitor = new MessageDialogProgressMonitor (true, true, true, true, MessageService.RootWindow);
 				monitor.BeginTask (GettextCatalog.GetString("Adding files..."), files.Length);
 			}
 			
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.ProgressMonitoring/MessageDialogProgressMonitor.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.ProgressMonitoring/MessageDialogProgressMonitor.cs
@@ -57,10 +57,16 @@ namespace MonoDevelop.Ide.ProgressMonitoring
 		{
 		}
 		
-		public MessageDialogProgressMonitor (bool showProgress, bool allowCancel, bool showDetails, bool hideWhenDone): base (Runtime.MainSynchronizationContext)
+		public MessageDialogProgressMonitor (bool showProgress, bool allowCancel, bool showDetails, bool hideWhenDone)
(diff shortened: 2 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
```diff
@@ -1979,7 +1979,7 @@ public IList<ProjectFile> AddFilesToProject (Project project, FilePath[] files,
 			ProgressMonitor monitor = null;
 			
 			if (files.Length > 10) {
-				monitor = new MessageDialogProgressMonitor (true);
+				monitor = new MessageDialogProgressMonitor (true, true, true, true, MessageService.RootWindow);
 				monitor.BeginTask (GettextCatalog.GetString("Adding files..."), files.Length);
```
rodrmoya: if this window can be opened without the main window opened, please use `DesktopService.GetFocusedTopLevelWindow` here also, if not, then it's ok to use `RootWindow`, although I guess, unless it really needs the RootWindow (and the latter is opened already), I'd say better to use the DesktopService call.

netonjm: @rodrmoya it's possible to create an analyzer to catch this? cc @Therzok

mrward: This window is only opened with the main window opened. You need to have the Solution pad visible in order to trigger this.

Not sure what change is being suggested here. The code here should be using `DesktopService.GetFocusedTopLevelWindow` to pass into MessageDialogProgressMonitor? The code was using that originally but inside the MessageDialogProgressMonitor and that does not seem to work. It seemed to end up with the wrong parent and was in front of the confirmation dialog. If there is a better way to fix this I am happy to change it. As a quick test I did fix was to revert the change in MessageDialogProgressMonitor which also fixed but I guess we need that code :)

rodrmoya: if this dialog needs the main window, then it's ok to use `RootWindow`. If it can be opened without the RootWindow visible (like File->NewSolution, Tools->Assembly browser, etc), then it needs to not rely on `RootWindow`, as that will make `RootWindow` visible when it shouldn't.

But again, if this can only be opened when the RootWindow is visible, then it's ok, no need for changing anything

Therzok: @netonjm Analyzer wouldn't work in this case. If we want to use GetFocusedTopLevelWindow everywhere, we should just obsolete RootWindow or validate all its usage. I'm not sure how this could be determined at static analysis time.

slluis:

PR #7306, opened 2019-03-06
Remove MetadataReferences from ProjectReferences

Therzok:
These are not needed, as they're handled via in-memory compilation. This should improve performance and prevent recompiling of the assembly on editing a root project


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.MetadataReferenceHandler.cs
```diff
@@ -132,34 +132,6 @@ async Task<bool> AddMetadataAssemblyReferences (AddMetadataReferencesData data)
 				}
 			}
 
-			bool AddMetadataProjectReferences (AddMetadataReferencesData data)
```
Therzok: This is already done here: https://github.com/mono/monodevelop/pull/7306/files#diff-b3d7afe03541f257c303184c89439b15R151

Therzok:

PR #7314, opened 2019-03-06
[VersionControl] Fix problem detecting sln and csproj changes

jsuarezruiz:
In the StatusView, obtaining the status of files using `GetVersionInfo`, we obtain an incorrect value (without local changes). It seems to be a cache problem. For this reason, sometimes trying to add the change to the changeset we do not do it and we do not update the UI correctly.
Added validation where we verify the status of the file ignoring the cache (in necessary case).

Fixes VSTS #675483
Fixes VSTS #672672

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/StatusView.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/StatusView.cs
@@ -1001,7 +1001,7 @@ namespace MonoDevelop.VersionControl.Views
 			VersionInfo newInfo;
 			try {
 				// Reuse remote status from old version info
-				newInfo = vc.GetVersionInfo (args.FilePath);
+				newInfo = vc.GetVersionInfo (args.FilePath, VersionInfoQueryFlags.IgnoreCache);
 				if (found && newInfo != null) {
 					VersionInfo oldInfo = statuses [oldStatusIndex];
 					if (oldInfo != null) {
@@ -1034,9 +1034,11 @@ namespace MonoDevelop.VersionControl.Views
 			}
 			else {
 				if (FileVisible (newInfo)) {
-					statuses.Add (newInfo);
-					changeSet.AddFile (newInfo);
-					AppendFileInfo (newInfo, wasExpanded);
+					if (!statuses.Any(s => s.LocalPath == newInfo.LocalPath)) {
+						statuses.Add (newInfo);
+						changeSet.AddFile (newInfo);
+						AppendFileInfo (newInfo, wasExpanded);
+					}
 				}
 			}
 			return true;
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/StatusView.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/StatusView.cs
@@ -11,7 +11,6 @@ using MonoDevelop.Components.Commands;
 using MonoDevelop.Projects;
 using MonoDevelop.Ide;
 using System.Text;
-using System.Threading.Tasks;
 
 namespace MonoDevelop.VersionControl.Views
 {
@@ -303,7 +302,7 @@ namespace MonoDevelop.VersionControl.Views
 				Image = new Xwt.ImageView (ImageService.GetIcon ("vc-status-refresh", IconSize.Menu)).ToGtkWidget ()
 			};
 			btnRefresh.Image.Show ();
-
+			
 			btnRefresh.Clicked += new EventHandler (OnRefresh);
 			toolbar.Add (btnRefresh);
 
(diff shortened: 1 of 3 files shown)


main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/StatusView.cs
```diff
@@ -1031,17 +1029,28 @@ bool OnFileStatusChanged (FileUpdateEventInfo args)
...
+				} else {
+					InalidateChangeSet (newInfo.LocalPath);
 				}
 			}
 			return true;
 		}
 
+		void InalidateChangeSet(FilePath path)
+		{
+			if (!changeSet.ContainsFile (path)) {
```
sevoku: what happens if the changeset already contains the file?

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/StatusView.cs
```diff
@@ -999,6 +996,7 @@ bool OnFileStatusChanged (FileUpdateEventInfo args)
 			}
 
 			VersionInfo newInfo;
+
 			try {
 				// Reuse remote status from old version info
 				newInfo = vc.GetVersionInfo (args.FilePath);
```
sevoku: Shouldn't this bypass the cache too?

Therzok: Why would we bypass the cache? We shold have proper invalidation caused by FileService

sevoku: The main problem is that we update the cache in the background. On file changes we only mark an entry as invalid here: https://github.com/mono/monodevelop/blob/master/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlFileSystemExtension.cs#L125 and then update them asynchronously here: https://github.com/mono/monodevelop/blob/master/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs#L255.

So the right solution would be to fire `VersionControlService.NotifyFileStatusChanged` after the cache update (in `Repository.RunQueries()`) and not immediately on file change.

sevoku:
@jsuarezruiz Lluis is right, see my comment here: https://github.com/mono/monodevelop/pull/7314#discussion_r263286534

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/StatusView.cs
slluis: @sevoku exactly. The repository knows when version status changes, so the repository should be responsible for notifying the change.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionInfoCache.cs
```diff
@@ -23,181 +23,146 @@
...
-				fileStatus [versionInfo.LocalPath] = versionInfo;
-			} finally {
-				fileLock.ExitWriteLock ();
-			}
+			if (fileStatus.TryGetValue (versionInfo.LocalPath, out var vi) && vi.Equals (versionInfo)) {
+				vi.RequiresRefresh = false;
+				return;
+			}
+
+			fileStatus.AddOrUpdate (versionInfo.LocalPath, versionInfo, (f, v) => versionInfo);
```
Therzok:
```suggestion
			fileStatus[versionInfo.LocalPath] = versionInfo;
```

The setter internally handles concurrent scenarios.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionInfoCache.cs
```diff
@@ -23,181 +23,146 @@
...
+			foreach (var versionInfo in versionInfos) {
+				if (!versionInfo.IsInitialized)
+					versionInfo.Init (repo);
+
+				if (fileStatus.TryGetValue (versionInfo.LocalPath, out var vi) && vi.Equals (versionInfo)) {
+					vi.RequiresRefresh = false;
+					continue;
+				}
+
+				fileStatus.AddOrUpdate (versionInfo.LocalPath, versionInfo, (f, v) => versionInfo);
```
Therzok:
```suggestion
			fileStatus[versionInfo.LocalPath] = versionInfo;
```

The setter internally handles concurrent scenarios.

jsuarezruiz:
To simplify the review and the historical, we will unify this PR with https://github.com/mono/monodevelop/pull/7149 in a new one.

jsuarezruiz:
New related PR: https://github.com/mono/monodevelop/pull/7345

PR #7318, opened 2019-03-07
[StressTest] Add a way to detect leaks on iterations / scenarios

Therzok:
This PR provides the groundwork for detecting leaks of various objects via the StressTest application.

What it does, is that it searches for an attribute [NoLeak("myclassname")] on a test scenario, enabling the tool to report object counts after an iteration or at the end for a given type.


Therzok:
Things left to do:
* Write results to a result file which would then be uploaded to artifacts
* Make some makeshift retention path detection, so we can report this in the results file
* Enable this in CI.

Therzok:
Waiting for build to finish, then triggering leaktest release.

Therzok:
We should be diffing object counts between runs, not using the global count.

Therzok:
Cleaning my VS cache and preferences make the Solution held by a thread stack. Not sure how to fix it. It is a root to itself as thread stack. Maybe we should ignore these scenarios.

Therzok:
I can revert the cancellation fixes if we don't want them, but it's good to sanitize their usage.

Therzok:
But all in all, it seems to function.

main/tests/StressTest/MonoDevelop.StressTest/Profiler/Heapshot.cs
```diff
@@ -0,0 +1,97 @@
...
+
+				if (count > 0)
+					result[name] = (count, classId);
+			}
+
+			return result;
+		}
+
+
+		//public static IEnumerable<(LogHeapRootSource, long[])> SearchRoots(this Heapshot heapshot, long objAddr)
```
Therzok: TODO: Remove this.

main/tests/StressTest/MonoDevelop.StressTest/Profiler/Heapshot.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+	public class NativeHeapshot
+	{
+		public readonly Dictionary<long, HeapRootRegisterEvent> Roots = new Dictionary<long, HeapRootRegisterEvent> ();
+		public readonly Dictionary<long, int> ObjectsPerClassCounter = new Dictionary<long, int> ();
+		public Dictionary<long, ClassLoadEvent> ClassInfos;
+		public readonly Dictionary<long, long> ObjectToType = new Dictionary<long, long> ();
+		public readonly Dictionary<long, List<long>> TypeToObjectList = new Dictionary<long, List<long>> ();
+
+		public AdjacencyGraph<long, Edge<long>> Graph = new AdjacencyGraph<long, Edge<long>> ();
```
Therzok: Maybe we can use an `STaggedEdge` for the graph, thus we can set all the data we want in the tag.

main/tests/StressTest/MonoDevelop.StressTest/Profiler/Heapshot.cs
```diff
@@ -8,55 +8,138 @@ namespace MonoDevelop.StressTest
...
+		public NativeHeapshot(HashSet<string> trackedTypeNames)
+		{
+			this.trackedTypeNames = new HashSet<string> (trackedTypeNames);
+		}
+
+		public void AddObject (TypeInfo typeInfo, HeapObjectEvent heapObjectEvent)
+		{
+			if (heapObjectEvent.ObjectSize == 0) {
+				// This means it's just reporting references
+				// TODO: Validate if we need to handle it.
```
Therzok: This might need to be handled to prevent false positives from going through.

Therzok:
@slluis should be ready for merge.

PR #7320, opened 2019-03-07
[IDE] Track whether a file load was initiated by the File Manager better

iainx:
Instead of using a timer that isn't very accurate, track whether or not a
file load was initiated by the FileManager by passing a boolean to the events

Fixes VSTS #806672 https://devdiv.visualstudio.com/DevDiv/_workitems/edit/806672

### DIFF
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -57,6 +57,7 @@ using Xwt.Mac;
 using MonoDevelop.Components.Mac;
 using System.Reflection;
 using MacPlatform;
+using MonoDevelop.Projects;
 
 namespace MonoDevelop.MacIntegration
 {
@@ -582,6 +583,14 @@ namespace MonoDevelop.MacIntegration
 //			Gtk.Rc.ParseString (gtkrc);
 		}
 
+		static TimeToCodeMetadata.DocumentType GetDocumentTypeFromFilename (string filename)
+		{
+			if (Projects.Services.ProjectService.IsWorkspaceItemFile (filename) || Projects.Services.ProjectService.IsSolutionItemFile (filename)) {
+				return TimeToCodeMetadata.DocumentType.Solution;
+			}
+			return TimeToCodeMetadata.DocumentType.File;
+		}
+
 		void GlobalSetup ()
 		{
 			//FIXME: should we remove these when finalizing?
@@ -628,21 +637,30 @@ namespace MonoDevelop.MacIntegration
 				};
 
 				ApplicationEvents.OpenDocuments += delegate (object sender, ApplicationDocumentEventArgs e) {
-					//OpenFiles may pump the mainloop, but can't do that from an AppleEvent, so use a brief timeout
-					GLib.Timeout.Add (0, delegate {
+					//OpenFiles may pump the mainloop, but can't do that from an AppleEvent
+					GLib.Idle.Add (delegate {
 						Ide.WelcomePage.WelcomePageService.HideWelcomePageOrWindow ();
-						IdeApp.ReportTimeToCode = true;
+						var trackTTC = IdeApp.StartTimeToCodeLoadTimer ();
 						IdeApp.OpenFiles (e.Documents.Select (
-							doc => new FileOpenInformation (doc.Key, null, doc.Value, 1, OpenDocumentOptions.DefaultInternal))
-						);
+							doc => new FileOpenInformation (doc.Key, null, doc.Value, 1, OpenDocumentOptions.DefaultInternal)),
+							null
+						).ContinueWith ((result) => {
+							if (!trackTTC) {
+								return;
+							}
+
+							var firstFile = e.Documents.First ().Key;
+
+							IdeApp.TrackTimeToCode (GetDocumentTypeFromFilename (firstFile));
+						});
 						return false;
 					});
 					e.Handled = true;
 				};
 
 				ApplicationEvents.OpenUrls += delegate (object sender, ApplicationUrlEventArgs e) {
-					GLib.Timeout.Add (0, delegate {
-						IdeApp.ReportTimeToCode = true;
+					GLib.Idle.Add (delegate {
+						var trackTTC = IdeApp.StartTimeToCodeLoadTimer ();
 						// Open files via the monodevelop:// URI scheme, compatible with the
 						// common TextMate scheme: http://blog.macromates.com/2007/the-textmate-url-scheme/
 						IdeApp.OpenFiles (e.Urls.Select (url => {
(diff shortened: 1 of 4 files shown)


iainx:
@monojenkins backport release-8.0

mrward (commented):
Should we make all use of InitiatedByFileManager internal? Do we want/need this to be public?

iainx:
We don't, no.

iainx:
But there were a few places that we couldn't make it internal due to internal not being visible IIRC

iainx:
@slluis Updated to take your idea into consideration

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -628,21 +637,25 @@ void GlobalSetup ()
 				};
 
 				ApplicationEvents.OpenDocuments += delegate (object sender, ApplicationDocumentEventArgs e) {
-					//OpenFiles may pump the mainloop, but can't do that from an AppleEvent, so use a brief timeout
-					GLib.Timeout.Add (0, delegate {
+					//OpenFiles may pump the mainloop, but can't do that from an AppleEvent
+					MonoDevelop.Core.Runtime.RunInMainThread ( () => {
```
Therzok: Runtime.RunInMainThread is sync if we're already on the main thread.

Therzok: So this changes behaviour.

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -628,21 +637,25 @@ void GlobalSetup ()
...
 						Ide.WelcomePage.WelcomePageService.HideWelcomePageOrWindow ();
-						IdeApp.ReportTimeToCode = true;
+						IdeApp.StartTimeToCodeTimer ();
 						IdeApp.OpenFiles (e.Documents.Select (
-							doc => new FileOpenInformation (doc.Key, null, doc.Value, 1, OpenDocumentOptions.DefaultInternal))
-						);
-						return false;
+							doc => new FileOpenInformation (doc.Key, null, doc.Value, 1, OpenDocumentOptions.DefaultInternal)),
+							null
+						).ContinueWith ((result) => {
```
Therzok: I keep forgetting the basic TPL APIs, does this continue on the same scheduler (main thread) or thread pool?

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -628,21 +637,25 @@ void GlobalSetup ()
...
+							IdeApp.SendTimeToCodeTelemetry (GetDocumentTypeFromFilename (firstFile));
+						});
 					});
 					e.Handled = true;
 				};
 
 				ApplicationEvents.OpenUrls += delegate (object sender, ApplicationUrlEventArgs e) {
-					GLib.Timeout.Add (0, delegate {
-						IdeApp.ReportTimeToCode = true;
+					Core.Runtime.RunInMainThread (() => {
```
Therzok: Same here, change of semantics.

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -628,21 +637,27 @@ void GlobalSetup ()
 				};
 
 				ApplicationEvents.OpenDocuments += delegate (object sender, ApplicationDocumentEventArgs e) {
-					//OpenFiles may pump the mainloop, but can't do that from an AppleEvent, so use a brief timeout
-					GLib.Timeout.Add (0, delegate {
+					//OpenFiles may pump the mainloop, but can't do that from an AppleEvent
+					GLib.Idle.Add (delegate {
 						Ide.WelcomePage.WelcomePageService.HideWelcomePageOrWindow ();
-						IdeApp.ReportTimeToCode = true;
+						IdeApp.StartTimeToCodeTimer ();
```
slluis: This is not right, it should be measuring time since startup, not since the open message was received.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
```diff
@@ -318,20 +323,43 @@ static void KeyBindingFailed (object sender, KeyBindingFailedEventArgs e)
...
+			timeToCodeTimer.Stop ();
+			GLib.Source.Remove (ttcTimeout);
+			ttcTimeout = 0;
+
+			if (ttcMetadata == null) {
+				timeToCodeTimer = null;
+				throw new Exception ("SendTimeToCode called before initialisation completed");
+			}
+
+			LoggingService.LogInfo ("Processing TTC");
```
slluis: All of those LogInfo should be LogDebug

slluis (changes requested):

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
```diff
@@ -101,8 +102,25 @@ public static class IdeApp
...
+
+		internal static void OnStartupCompleted (StartupMetadata startupMetadata)
 		{
+			ttcMetadata = new TimeToCodeMetadata {
+				StartupTime = startupMetadata.CorrectedStartupTime
+			};
+			ttcMetadata.AddProperties (startupMetadata);
+
+			LoggingService.LogInfo ("TTC starting");
+			ttcTimeout = GLib.Timeout.Add (ttcDuration, () => {
```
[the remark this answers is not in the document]
slluis: BTW, I think ttcDuration of 6s is too much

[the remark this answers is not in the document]
iainx: 6s was what was needed for my machine when I tested it. I had 3 seconds initially but it was timing out.

main/src/addins/MacPlatform/MacPlatform.cs
iainx: It's in theory corrected at the end.

But if we using a timer since startup, and don't receive an open event, then the timer will never be stopped, unless we use yet another timer to track whether we got the open event

slluis: Well, a Stopwatch is not a real timer, it just stores a timestamp, it doesn't matter if we don't stop it.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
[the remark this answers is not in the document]
slluis: 6 seconds between Start Window showing up and getting the open document event? that's a lot. For me there isn't even time to render the Start Window.

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -582,6 +583,17 @@ void PatchGtkTheme ()
 //			Gtk.Rc.ParseString (gtkrc);
 		}
 
+		static TimeToCodeMetadata.DocumentType GetDocumentTypeFromFilename (string filename)
+		{
+			if (Projects.Services.ProjectService.IsWorkspaceItemFile (filename)) {
+				return TimeToCodeMetadata.DocumentType.Solution;
+			} else if (Projects.Services.ProjectService.IsSolutionItemFile (filename)) {
+				return TimeToCodeMetadata.DocumentType.File;
```
slluis: No no, IsWorkspaceItemFile() will return true for any WorkspaceItem subclass (such as a Solution). IsSolutionItemFile() will return true for any SolutionItem subclass (such as a Project). In both cases it should return a TimeToCodeMetadata.DocumentType.Solution.

slluis (approved):

slluis:

PR #7323, opened 2019-03-08
[SourceEditor] Takes line length as column when cursor is set at the …

jtorres:
…end of the caret.

- When the cursor is at the end of the line, the caret Column is higher than line length so that after adding a new breakpoint, this is set at the line after. With this commit, we prevent it.
- Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/792482


slluis:
@DavidKarlas make sure the new editor doesn't have this bug.

slluis:

slluis:
@DavidKarlas we should check if the .net core debugger uses 0 or 1-based columns, but no matter what, this fix is still necessary. The editor should not create breakpoints in a column that is not valid.

rodrmoya:
As for the .net core debugger using 0 or 1, @nosami any idea?

PR #7327, opened 2019-03-08
Ide: Add command to close all tabs to the right

sandyarmstrong:
Identical behavior to VScode.

image

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
@@ -107,11 +107,56 @@ namespace MonoDevelop.Ide.Commands
 			var active = IdeApp.Workbench.ActiveDocument;
 			return active == null ? null : active.Window.ViewContent;
 		}
+
+		protected override void Update (CommandInfo info)
+		{
+			var documents = IdeApp.Workbench.Documents;
+			var activeDoc = IdeApp.Workbench.ActiveDocument;
+
+			if (activeDoc == null) {
+				info.Enabled = false;
+				return;
+			}
+
+			var activeNotebook = ((SdiWorkspaceWindow)activeDoc.Window).TabControl;
+
+			// Disable if only document in tab strip
+			foreach (var doc in documents) { 
+				if (doc != activeDoc && ((SdiWorkspaceWindow)doc.Window).TabControl == activeNotebook) {
+					info.Enabled = true;
+					return;
+				}
+			}
+
+			info.Enabled = false;
+		}
 	}
 
 	class CloseAllToTheRightHandler : CloseAllButThisHandler
 	{
 		protected override bool StartAfterException => true;
+
+		protected override void Update (CommandInfo info)
+		{
+			var documents = IdeApp.Workbench.Documents;
+			var activeDoc = IdeApp.Workbench.ActiveDocument;
+
+			if (activeDoc == null) {
+				info.Enabled = false;
+				return;
+			}
+
+			var activeNotebook = ((SdiWorkspaceWindow)activeDoc.Window).TabControl;
+
+			// Disable if right-most document in tab strip
+			for (int i = documents.Count - 1; i >= 0; i--) {
+				var doc = documents [i];
+				if (((SdiWorkspaceWindow)doc.Window).TabControl == activeNotebook) {
+					info.Enabled = doc != activeDoc;
+					return;
+				}
+			}
+		}
 	}
 
 	class ToggleMaximizeHandler : CommandHandler
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
@@ -27,14 +27,14 @@
 //
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 
 using Gtk;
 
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
```diff
@@ -64,26 +67,39 @@ protected override void Run ()
...
-				.ToArray ();
+			var docs = new List<Document> ();
+			var dirtyDialogShown = false;
+			var startRemoving = except == null || !StartAfterException;
+
+			foreach (var doc in IdeApp.Workbench.Documents) {
+				if (((SdiWorkspaceWindow)doc.Window).TabControl == activeNotebook) {
+					if (except != null && doc.Window.ViewContent == except) {
+						startRemoving = true;
+						continue;
```
Therzok: Won't this ignore the active document in CloseAll?

sandyarmstrong: I tested Close All, Close Except, Close Tab, Close To The Right, all with tab reordering. I think we're good. Thanks for checking!

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
```diff
@@ -92,15 +108,20 @@ protected override ViewContent GetDocumentException ()
 			return active == null ? null : active.Window.ViewContent;
 		}
 	}
-	
+
+	class CloseAllToTheRightHandler : CloseAllButThisHandler
+	{
+		protected override bool StartAfterException => true;
```
sevoku: What about enabling it only if there are any documents to the right?
We could add the same to `CloseAllButThisHandler` but without the `.SkipWhile()` part.

```suggestion
		protected override bool StartAfterException => true;
		
		
		protected override void Update (CommandInfo info)
		{
			var active = IdeApp.Workbench.ActiveDocument;
			if (active == null)
				return;

			var activeNotebook = ((SdiWorkspaceWindow)active.Window).TabControl;
			info.Enabled = IdeApp.Workbench.Documents.SkipWhile (doc => doc != active).Any (doc => doc != active && ((SdiWorkspaceWindow)doc.Window).TabControl == activeNotebook);
		}
```

sandyarmstrong: @sevoku I have mixed feelings about this, which is why I left it out. How often is `Update` called for menu items? I know it's a lot for always-visible toolbar commands.

VScode doesn't bother disabling for the right-most tab, for what it's worth.

Happy to add this if we're not worried about `Update` doing work.

netonjm: I think the update is only called when the popup is opened. You only need to check is if is the last element of the array.

sandyarmstrong: Last element of the array with matching `TabControl`, yeah. I'll update on Monday, thanks!

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
```diff
@@ -64,43 +67,106 @@ protected override void Run ()
...
+				info.Enabled = false;
+				return;
+			}
+
+			var activeNotebook = ((SdiWorkspaceWindow)activeDoc.Window).TabControl;
+
+			// Disable if right-most document in tab strip
+			for (int i = documents.Count - 1; i >= 0; i--) {
+				var doc = documents [i];
+				if (((SdiWorkspaceWindow)doc.Window).TabControl == activeNotebook) {
```
netonjm: we can remove this from the iteration, we get always the same value

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileTabCommands.cs
```diff
@@ -64,43 +67,106 @@ protected override void Run ()
...
+
+			if (activeDoc == null) {
+				info.Enabled = false;
+				return;
+			}
+
+			var activeNotebook = ((SdiWorkspaceWindow)activeDoc.Window).TabControl;
+
+			// Disable if right-most document in tab strip
+			for (int i = documents.Count - 1; i >= 0; i--) {
```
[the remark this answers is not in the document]
sandyarmstrong: The last member of the `documents` list could be in a different tab strip.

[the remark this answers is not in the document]
netonjm: Sorry I was assuming we were checking only checking the documents from the tabstrip of the *selected tab*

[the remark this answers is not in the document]
sevoku: Hhm, I'm actually not sure, are the docs in `documents` sorted by their occurrence in the tab view?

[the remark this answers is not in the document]
sandyarmstrong: Yes.

sandyarmstrong:
Can't merge because Mac tests are timing out still.

sandyarmstrong:
This already merged.

PR #7332, opened 2019-03-11
[Core] Fix MSBuildEngineManager disposed error

mrward:
Dispose the MSBuildEngineManager after the project active tasks have
completed. This should avoid the MSBuildEngineManager being used
by the project after it has been disposed.  Also prevent new tasks
being bound in BindTask once the dispose has started. If BindTask
is called when a dispose has started for a parent object a
cancelled task is returned from BindTask.

```
  System.ObjectDisposedException: Cannot access a disposed object.
  Object name: 'MSBuildEngineManager'.
    at MonoDevelop.Projects.MSBuild.MSBuildEngineManager.GetEngine (System.Boolean supportsMSBuild) [0x0001e] in monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildEngineManager.cs:55
    at MonoDevelop.Projects.MSBuild.MSBuildProject.LoadNativeInstance (System.Boolean evaluateItems) [0x0005b] in monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs:498
    at MonoDevelop.Projects.MSBuild.MSBuildProjectInstance.Evaluate () [0x00019] in monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProjectInstance.cs:92
    at MonoDevelop.Projects.ProjectRunConfiguration.Initialize (MonoDevelop.Projects.Project project) [0x00038] in monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectRunConfiguration.cs:52
    at MonoDevelop.Projects.AssemblyRunConfiguration.Initialize (MonoDevelop.Projects.Project project) [0x00001] in monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/AssemblyRunConfiguration.cs:49
    at MonoDevelop.DotNetCore.DotNetCoreRunConfiguration.Initialize (MonoDevelop.Projects.Project project) [0x00019] in monodevelop/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreRunConfiguration.cs:87
    at MonoDevelop.Projects.Project.CreateRunConfigurationInternal (System.String name) [0x00009] in monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs:343
    at MonoDevelop.Projects.Project.WriteRunConfigurations (MonoDevelop.Core.ProgressMonitor monitor, MonoDevelop.Projects.MSBuild.MSBuildProject msproject, MonoDevelop.Projects.MSBuild.IMSBuildPropertySet globalGroup) [0x00043] in monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs:3444
    at MonoDevelop.Projects.Project.OnWriteProject (MonoDevelop.Core.ProgressMonitor monitor, MonoDevelop.Projects.MSBuild.MSBuildProject msproject) [0x00045] in monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs:3299
    at MonoDevelop.Projects.DotNetProject.OnWriteProject (MonoDevelop.Core.ProgressMonitor monitor, MonoDevelop.Projects.MSBuild.MSBuildProject msproject) [0x00001] in monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs:2037
```

Fixes VSTS #797958 - MSBuildEngineManager disposed error

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
@@ -99,8 +99,11 @@ namespace MonoDevelop.Projects
 		public Task<T> BindTask<T> (Func<CancellationToken, Task<T>> f)
 		{
 			var t = f (disposeCancellation.Token);
-			lock (activeTasks)
+			lock (activeTasks) {
+				if (disposeCancellation.IsCancellationRequested)
+					return Task.FromCanceled<T> (disposeCancellation.Token);
 				activeTasks.Add (t);
+			}
 			t.ContinueWith (tr => {
 				lock (activeTasks)
 					activeTasks.Remove (t);
@@ -118,13 +121,32 @@ namespace MonoDevelop.Projects
 		public Task BindTask (Func<CancellationToken, Task> f)
 		{
 			var t = f (disposeCancellation.Token);
-			lock (activeTasks)
+			lock (activeTasks) {
+				if (disposeCancellation.IsCancellationRequested)
+					return Task.FromCanceled (disposeCancellation.Token);
 				activeTasks.Add (t);
+			}
 			t.ContinueWith (tr => {
 				lock (activeTasks)
 					activeTasks.Remove (t);
 			});
 			return t;
+		}
+
+		/// <summary>
+		/// Gathers all the tasks for this WorkspaceObject and all children that need to finish before
+		/// this object can be disposed.
+		/// </summary>
+		internal void GetAllActiveTasksForDispose (List<Task> tasks)
+		{
+			lock (activeTasks) {
+				disposeCancellation.Cancel ();
+				tasks.AddRange (activeTasks);
+			}
+
+			foreach (var child in GetChildren ()) {
+				child.GetAllActiveTasksForDispose (tasks);
+			}
 		}
 
 		/// <summary>
@@ -210,14 +232,10 @@ namespace MonoDevelop.Projects
 
 			Disposed = true;
 
-			disposeCancellation.Cancel ();
+			var allTasks = new List<Task> ();
+			GetAllActiveTasksForDispose (allTasks);
 
-			Task[] allTasks;
-
-			lock (activeTasks)
-				allTasks = activeTasks.ToArray ();
-
-			if (allTasks.Length > 0)
+			if (allTasks.Count > 0)
 				Task.WhenAll (allTasks).ContinueWith (t => OnDispose (), TaskScheduler.FromCurrentSynchronizationContext ());
 			else
 				OnDispose ();
--- a/main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/SolutionTests.cs
+++ b/main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/SolutionTests.cs
@@ -1010,6 +1010,51 @@ namespace MonoDevelop.Projects
 				Runtime.Preferences.SkipBuildingUnmodifiedProjects.Set (oldPrefSkipUnmodified);
 			}
 		}
+
+		[Test]
+		public async Task SolutionDisposed_ActiveProjectTasks_MSBuildEngineManagerNotDisposedUntilProjectTasksCompleted ()
+		{
+			var en = new CustomSolutionItemNode<TestProjectExtension> ();
+			WorkspaceObject.RegisterCustomExtension (en);
+			try {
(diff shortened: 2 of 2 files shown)


Therzok:
My take on this is that APIs don't properly pass the cancellationtoken that is in BindTask, thus they're not cancelled on dispose.

Although, if the implementation does not take cancellation into account, it'll still NRE.

So on that front, looks good to me!

mrward:
Not using the BindTask cancellation token is another problem which needs looking at. I tried making that change, based on your comment a while ago, whilst looking at this bug for the GetReferenceAssemblies method - however it did not fix the MSBuildEngineManager being used after it is disposed. There seem to be some cases where it could be used whilst the token was disposed.

We also have some null references in the RunMSBuildTarget method if the project is disposed. Random thoughts here:
- Ensure BindTask cancellation token is associated with the progress monitor here.
   - Had several other IDE log errors occur here when the BindTask's token was passed through with GetReferenceAssemblies. Aborting the thread, and some others.
- Check the MSBuildProject is not null before using it. Then if the progress monitor has been cancelled handle that case - maybe by throwing a task canceled operation - or perhaps return no items or a special error TargetEvaluationResult (might be better).

Therzok (approved):

slluis (changes requested):
We already delay disposal when there are tasks running: https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs#L220. The problem is that this code only takes into account tasks initiated in the object being disposed, but not in child objects. I think a better solution would be to recursively get all the children (there is a GetChildren() method in WorkspaceObject) and then get all the tasks for all the children. Also, if an object hierarchy is being disposed, it should not allow new tasks to be bound to any object in the hierarchy.

mrward:
Updated after review.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
```diff
@@ -99,8 +100,11 @@ protected void AssertMainThread ()
 		public Task<T> BindTask<T> (Func<CancellationToken, Task<T>> f)
 		{
 			var t = f (disposeCancellation.Token);
-			lock (activeTasks)
+			lock (activeTasks) {
+				if (disposing)
+					return Task.FromCanceled<T> (new CancellationToken (true));
```
Therzok: Shouldn't we use disposeCancellation here?

Therzok (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
mrward: The disposeCancellation may not be cancelled at this point so I did not want to use it. Calling Solution.Dispose will mark all child items as `disposing` but they may not yet have had their Dispose method called if there are still active tasks. SolutionFolder disposes its child items in its OnDispose method which could be called sometime later on.

I did think about using the disposeCancellation if it was cancelled but just returning a task that is cancelled seemed to be less code.

However this is making me think about the disposing. Seems like SolutionFolder will now wait for all its projects to finish their tasks before Project.Dispose is called. Which may avoid the null MSBuildProject but means tasks are not going to be cancelled as early as they were before.

slluis (approved):
👍

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
mrward: Mmm, maybe when we get the active tasks for the child items we should also call `disposeCancellation.Cancel()` for each child item. The child item's Dispose method could be called later on when the tasks have completed. Then we could use the disposeCancellation in the BindTask instead of a separate disposing flag.

slluis: Makes sense

mrward: Updated the code to use the disposeCancellation.

Therzok (approved):
Looks great!

slluis:

PR #7340, opened 2019-03-14
[Version Control] Fixes error adding new files to the Solution Items under Version Control

jsuarezruiz:
We have some changes here. The first one was adding the file in the solution. If there is no "Solution Items" folder, we create it.
On the other hand, trying to add the file to the repository (Add command) a validation was made taking into account the ActiveDocument that we will only have if the document is open (it is not always the case, or can even deal with a file that is not text). Validation changed.

Fixes VSTS #715157

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Commands.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Commands.cs
@@ -5,6 +5,7 @@ using MonoDevelop.Ide;
 using MonoDevelop.Ide.Gui;
 using MonoDevelop.Core;
 using MonoDevelop.VersionControl.Views;
+using System.Linq;
 
 namespace MonoDevelop.VersionControl
 {
@@ -77,7 +78,7 @@ namespace MonoDevelop.VersionControl
 		}
 	}
 	
-	class FileVersionControlCommandHandler: CommandHandler
+	class FileVersionControlCommandHandler : CommandHandler
 	{
 		protected static VersionControlItemList GetItems ()
 		{
@@ -88,22 +89,37 @@ namespace MonoDevelop.VersionControl
 				list.Add (it);
 			return list;
 		}
-		
+
 		protected static VersionControlItem GetItem ()
 		{
+			Repository repo = null;
+			FilePath filePath = null;
+
+			if (IdeApp.ProjectOperations.CurrentSelectedItem as IFileItem != null) {
+				filePath = (IdeApp.ProjectOperations.CurrentSelectedItem as IFileItem).FileName;
+			}
+
+			var currentSelectedWorkspaceItem = IdeApp.ProjectOperations.CurrentSelectedWorkspaceItem;
+			if (currentSelectedWorkspaceItem != null) {
+				repo = VersionControlService.GetRepository (currentSelectedWorkspaceItem);
+			}
+
+			Project project = IdeApp.ProjectOperations.CurrentSelectedProject;
 			Document doc = IdeApp.Workbench.ActiveDocument;
-			if (doc == null || !doc.IsFile)
-				return null;
-			
-			Project project = doc.Project ?? IdeApp.ProjectOperations.CurrentSelectedProject;
-			if (project == null)
-				return null;
-			
-			Repository repo = VersionControlService.GetRepository (project);
+
+			if (doc != null) {
+				filePath = doc.FileName;
+				project = doc.Project;
+
+				if (repo == null) {
+					repo = VersionControlService.GetRepository (project);
+				}
+			}
+
 			if (repo == null || repo.VersionControlSystem == null || !repo.VersionControlSystem.IsInstalled)
 				return null;
-			
-			return new VersionControlItem (repo, project, doc.FileName, false, null);
+
+			return new VersionControlItem (repo, project, filePath, false, null);
 		}
 		
 		protected sealed override void Run ()
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs
@@ -2,20 +2,20 @@
 using System;
 using System.IO;
 using System.Text;
-using System.Collections;
 using System.Collections.Generic;
 using MonoDevelop.Core.Serialization;
 using MonoDevelop.Core;
 using System.Linq;
 using System.Threading;
-using MonoDevelop.Core.Instrumentation;
 using MonoDevelop.Ide;
+using MonoDevelop.Projects;
 
 namespace MonoDevelop.VersionControl
 {
 	[DataItem (FallbackType=typeof(UnknownRepository))]
 	public abstract class Repository: IDisposable
 	{
+		FilePath filePath;
 		string name;
 		VersionControlSystem vcs;
 		
@@ -26,8 +26,11 @@ namespace MonoDevelop.VersionControl
 
 		public FilePath RootPath
 		{
-			get;
-			protected set;
+			get { return filePath; }
+			protected set {
+				filePath = value;
+				InitFileWatcher (filePath);
+			}
 		}
 		
 		public event EventHandler NameChanged;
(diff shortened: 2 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/SystemFileNodeBuilder.cs
```diff
@@ -176,7 +176,19 @@ public void IncludeFileToProject ()
...
+							if (solutionFolder != null) {
+								if (solutionFolder.IsRoot) {
+									// Don't allow adding files to the root folder. VS doesn't allow it
+									// If there is no existing folder, create one
+									solutionFolder = solutionFolder.ParentSolution.DefaultSolutionFolder;
+								}
+
+								solutionFolder.Files.Add (file.Path);
+							}
+
```
slluis: It seems to me that all the above changes can be replaced by just `sol.DefaultSolutionFolder.Files.Add(file.Path)` (sol.RootFolder will never be null, and sol.RootFolder.IsRoot is always true)

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs
```diff
@@ -105,6 +108,8 @@ protected virtual void Dispose (bool disposing)
 
 			infoCache?.Dispose ();
 			infoCache = null;
+
+			ShutdownFileWatcher ();
```
sevoku: This should probably happen at the beginning, what if the FSW spawns new events, while/after query shutdown above

slluis:
The approach of the fix is not quite right. The problem is in the solution pad, where for some reason commands are not enabled. Solution pad commands are not handled by `FileVersionControlCommandHandler`. This class handles commands that apply to the active document, but not for the solution pad. Solution pad commands are handled by a [tree node command handler](https://github.com/mono/monodevelop/blob/master/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlNodeExtension.cs#L232), together with a corresponding [NodeBuilderExtension](https://github.com/mono/monodevelop/blob/master/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlNodeExtension.cs#L18). Maybe for some reason the node builder extension is not being applied to solution items, so handlers are not enabled.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/SystemFileNodeBuilder.cs
jsuarezruiz: You're right. Your proposal is simpler and does the same. Thanks for the feedback!

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs
```diff
@@ -1052,6 +1057,20 @@ public virtual bool GetFileIsText (FilePath path)
...
+		void InitFileWatcher (FilePath directoryPath)
+		{
+			FileWatcherService.WatchDirectories (this, new HashSet<FilePath> {
+				directoryPath
+			}).Ignore ();
+		}
+
+		void ShutdownFileWatcher()
+		{
+			FileWatcherService.WatchDirectories (this, new HashSet<FilePath> {
```
Therzok:
```suggestion
			FileWatcherService.WatchDirectories (this, null);
```

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Commands.cs
```diff
@@ -88,22 +89,37 @@ protected static VersionControlItemList GetItems ()
...
 			return list;
 		}
-		
+
 		protected static VersionControlItem GetItem ()
 		{
+			Repository repo = null;
+			FilePath filePath = null;
+
+			if (IdeApp.ProjectOperations.CurrentSelectedItem as IFileItem != null) {
```
Therzok:
```suggestion
			if (IdeApp.ProjectOperations.CurrentSelectedItem is IFileItem fileItem) {
				filePath = fileITem.FileName;
```

jsuarezruiz:
@Lluis, you are right. The problem seems to be the node type validation from `VersionControlNodeExtension` in the method `CanBuildNode` (`SolutionFolderItem`). Will send changes soon!

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs
```diff
@@ -26,8 +26,11 @@ public abstract class Repository: IDisposable
 
 		public FilePath RootPath
 		{
-			get;
-			protected set;
+			get { return filePath; }
+			protected set {
+				filePath = value;
+				InitFileWatcher (filePath);
```
slluis: Why is this necessary for this bug?

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlNodeExtension.cs
```diff
@@ -21,11 +21,12 @@ class VersionControlNodeExtension : NodeBuilderExtension
...
-			return typeof(ProjectFile).IsAssignableFrom (dataType)
-				|| typeof(SystemFile).IsAssignableFrom (dataType)
-				|| typeof(ProjectFolder).IsAssignableFrom (dataType)
-				|| typeof(WorkspaceObject).IsAssignableFrom (dataType);
+			return typeof (ProjectFile).IsAssignableFrom (dataType)
+				|| typeof (SystemFile).IsAssignableFrom (dataType)
+				|| typeof (ProjectFolder).IsAssignableFrom (dataType)
+				|| typeof (WorkspaceObject).IsAssignableFrom (dataType)
+				|| typeof (SolutionFolder).IsAssignableFrom (dataType)
+				|| typeof (SolutionFolderFileNode).IsAssignableFrom (dataType);
```
slluis: The new node types are not handled in BuildNode()

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs
jsuarezruiz: Good question. This FileWatcher is NOT necessary to resolve this issue (Version Control commands are disabled in root element, Solution Items).
So, why was it added?
If we have the following structure:
-Root
-- README.md
-- src/solution.sln

When opening the solution, we will have FileWatcher for everything that is in the src folder. But if we edit the README.md file externally, the file will not appear as modified in the StatusView.

Thinking about it, it's a different problem that maybe fits better in another Issue and PR. I'm going to remove this part to focus on the fix of this particular Issue.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlNodeExtension.cs
jsuarezruiz: We have added the types `SolutionFolder` and `SolutionFolderFileNode` in `CanBuildNode`. Lluis, you're right, we should manage this new types in the `BuildNode` method. However, both types are managed by `WorkspaceObject` (get VersionInfo and Add a folder overlay). Seems correct but, maybe we must apply a different behavior in this new types?

slluis: I guess it is ok for SolutionFolderFileNode, but SolutionFolderFileNode is not a WorkspaceObject.

jsuarezruiz: Yes, was ok for `SolutionFolder` and not for `SolutionFolderFileNode`. Added changes to add the overlay to SolutionFolderFileNode type items.

sevoku:

Therzok:
@sevoku are we still interested in getting this PR merged?

Therzok:
@sevoku are we still interested in getting this PR merged?

PR #7342, opened 2019-03-14
[DotNetCore] Show package updates in Solution window

mrward:
For package references directly specified in the project then if an
updated package is available this information will now be shown
on the Dependencies folder, the NuGet folder and the top level
package in the Solution window.

Fixes VSTS #710259 - .Net Standard project does not show available NuGet
package updates
Fixes VSTS #935714 - Highlight the available update more prominently in the solution pad

Before:

Before

After:

UpdateNuGet

Update menu now shows the version if one is available for both SDK style projects and classic projects.

UpdateMenu1

UpdateMenu2

UpdateClassic

Made things more consistent with how NuGet packages are displayed for non sdk style projects:

- The Properties window will now show the package id and version when a NuGet package is selected in the solution window.

PackagePropertiesWindow

Non-SDK style project changes:

 - Show the version in the Solution window
   - Previously they would not show the version and to see the version you would have to right click the package and open the context menu.
 - Right click context menu no longer shows the package version: `Version 1.2.3`
   - No longer needed since the version is displayed in the Solution window.

PackageVersionLabel

ClassicUpdates

PackageInstallingLabel


jmatthiesen:
This looks great to me. @mrward did you still have some questions for design about the presentation of the updates? I know we'd discussed using a different icon to show when updates are available, and then continue to show the version currently installed next to that. For now, I think this is good.

mrward:
I have created a ticket on the UX design board for a review for this.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.NodeBuilders/DependenciesNode.cs
```diff
@@ -25,19 +25,30 @@
...
 using MonoDevelop.Projects;
+using NuGet.Packaging;
 
 namespace MonoDevelop.DotNetCore.NodeBuilders
 {
 	class DependenciesNode
 	{
 		public static readonly string NodeName = "Dependencies";
 
+		IUpdatedNuGetPackagesInWorkspace updatedPackagesInWorkspace;
```
Therzok:
```suggestion
		readonly IUpdatedNuGetPackagesInWorkspace updatedPackagesInWorkspace;
```

mrward:
I think we are good now - removing the do not merge label. No more UI changes planned here. UX board approved.

PR #7343, opened 2019-03-15
[Version Control] Revision shown in "Unified-Diff" incorrect

jsuarezruiz:
Entering to the DiffView we avoid reloading the local changes with the current version. We keep the option selected in the local version selector.

Fixes VSTS #801631

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/DiffView.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/DiffView.cs
@@ -101,7 +101,6 @@ namespace MonoDevelop.VersionControl.Views
 		protected override void OnSelected ()
 		{
 			info.Start ();
-			ComparisonWidget.UpdateLocalText ();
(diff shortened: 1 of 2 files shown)


main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/DiffView.cs
```diff
@@ -101,7 +101,6 @@ public int GetLineInCenter (Mono.TextEditor.MonoTextEditor editor)
 		protected override void OnSelected ()
 		{
 			info.Start ();
-			ComparisonWidget.UpdateLocalText ();
```
sevoku: we should check here which revision is selected and skip this step only if it's not `Local`, otherwise the view won't reflect local (unsaved) changes to the file.

jsuarezruiz: Changed!

sevoku:
Closed via ddcad1d0aafea52d1a404e2513d87ccbc643d914

PR #7345, opened 2019-03-15
[Version Control] Fix problem notifying VCS changes

jsuarezruiz:
Updated `VersionInfoCache` to avoid cache problems.
Call `NotifyFileStatusChanged` deleting files and directories in the `Repository` class.

Fixes GH #7119 
Fixes VSTS #675483
Fixes VSTS #672672

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs
@@ -29,7 +29,9 @@ namespace MonoDevelop.VersionControl
 			get;
 			protected set;
 		}
-		
+
+		internal FilePath RepositoryPath { get; set; }
+
 		public event EventHandler NameChanged;
 		
 		protected Repository ()
@@ -102,6 +104,8 @@ namespace MonoDevelop.VersionControl
 					recursiveDirectoryQueryQueue.Clear ();
 				}
 			}
+
+			VersionControlService.repositoryCache.TryRemove (RepositoryPath.CanonicalPath, out _);
 
 			infoCache?.Dispose ();
 			infoCache = null;
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs
@@ -258,7 +258,7 @@ namespace MonoDevelop.VersionControl
 				else {
 					// If there is no cached status, query it asynchronously
 					vi = new VersionInfo (p, "", Directory.Exists (p), VersionStatus.Versioned, null, VersionStatus.Versioned, null);
-					infoCache.SetStatus (vi, false);
+					infoCache.SetStatus (vi, true);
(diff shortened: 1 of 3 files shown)


main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionInfoCache.cs
```diff
@@ -23,20 +23,19 @@
...
+
 namespace MonoDevelop.VersionControl
 {
 	class VersionInfoCache : IDisposable
 	{
-		ReaderWriterLockSlim fileLock = new ReaderWriterLockSlim();
-		Dictionary<FilePath,VersionInfo> fileStatus = new Dictionary<FilePath, VersionInfo> ();
-		ReaderWriterLockSlim directoryLock = new ReaderWriterLockSlim ();
-		Dictionary<FilePath,DirectoryStatus> directoryStatus = new Dictionary<FilePath, DirectoryStatus> ();
+		static readonly ConcurrentDictionary<FilePath, VersionInfo> fileStatus = new ConcurrentDictionary<FilePath, VersionInfo> ();
```
Therzok: Why static?

jsuarezruiz: Good question. Initially, in the PR it was static. But it was a mistake. The problem comes from here: https://github.com/mono/monodevelop/pull/7345/files#diff-f6191d34b79d44987b335f29c6f89616R108
We came to create a different instance of the repository (with a different cache) which caused us to update values from one cache and recover from another. The static comes from there, but was incorrect because leaving the static values could trigger problems working with submodules. Fixed!

slluis:
I've been testing the PR and the bugs seem to be fixed. However, I'm getting a crash. Here are the steps to reproduce:
* Create a blank Forms project, enabling the option to add git support
* Once the project is created and restored, right click on the solution, click on "Review and Commit".
* Make sure all files are selected in the status view, then click on commit.

I consistently get a native crash: https://gist.github.com/slluis/23b66522b4506d3de09de8c5b26de3e9

sevoku:
The native crash doesn't seem to be related, I've seen similar crashes on some older mono version in different tree views, might be atk-cocoa related. Also can't repro this with a release-8.0 rebase. I'll try to provision both master and release-8.0 with exact dependencies and try to repro again.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/Repository.cs
```diff
@@ -258,7 +262,7 @@ public IEnumerable<VersionInfo> GetVersionInfo (IEnumerable<FilePath> paths, Ver
 				else {
 					// If there is no cached status, query it asynchronously
 					vi = new VersionInfo (p, "", Directory.Exists (p), VersionStatus.Versioned, null, VersionStatus.Versioned, null);
-					infoCache.SetStatus (vi, false);
+					infoCache.SetStatus (vi, true);
```
sevoku:
```suggestion
					infoCache.SetStatus (vi, false);
```
This change shouldn't be required, the queue processor will pick this up and notify (that was broken, but now fixed with the other changes in this PR)

PR #7351, opened 2019-03-19
[Shell] Language selection show each language in its own locale

jsuarezruiz:
With this change, each language name under language selection shows in its native translation.

Now, we have situations like this.
korean-problem

Languages are impossible to select for a non-Korean speaker. Changed to:
Screenshot 2019-03-19 at 13 47 52

_NOTE: Native translation from every language po file used in MonoDevelop._

Fixes VSTS #679609


jsuarezruiz:
With this change, we will have the same behaviour as Visual Studio for Windows.
Screenshot 2019-03-19 at 12 51 14

jsuarezruiz:
Good question @sevoku. @slluis What do you think?

slluis:
Yeah, we can sort.

jsuarezruiz:
Applied the sort. cc @sevoku

netonjm:
If we are translating the titles, Catalan should be **Català**? cc @slluis  💃

PR #7354, opened 2019-03-19
[Version Control] Avoid creating .git if it already exists

jsuarezruiz:
When creating a new project into a folder which already has already been setup as a git repo, avoid create the .git folder.

Fixes VSTS #660470

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/ProjectTemplateHandler.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/ProjectTemplateHandler.cs
@@ -29,6 +29,9 @@ using LibGit2Sharp;
 using MonoDevelop.Ide.Projects;
 using MonoDevelop.Ide.Templates;
 using MonoDevelop.Core;
+using MonoDevelop.Projects;
+using MonoDevelop.Ide;
+using System.Linq;
 
 namespace MonoDevelop.VersionControl.Git
 {
@@ -37,12 +40,32 @@ namespace MonoDevelop.VersionControl.Git
 		public void Run (NewProjectConfiguration config)
 		{
 			if (config.UseGit) {
-				if (config.CreateGitIgnoreFile) {
-					CreateGitIgnoreFile (config.SolutionLocation);
-				}
+				bool isUsingGit = IsUsingGit (config.SolutionLocation);
+				if (!isUsingGit) {
+					if (config.CreateGitIgnoreFile) {
+						CreateGitIgnoreFile (config.SolutionLocation);
+					}
 
-				CreateGitRepository (config.SolutionLocation);
+					CreateGitRepository (config.SolutionLocation);
+				}
 			}
+		}
+
+		bool IsUsingGit (string path)
+		{
+			bool isGitRepository = false;
+			DirectoryInfo repoDirectory = new DirectoryInfo (path);
+
+			while (repoDirectory != null) {
+				FilePath filePath = new FilePath (repoDirectory.FullName);
+
+				if (filePath.Combine (".git").IsDirectory) {
+					isGitRepository = true;
+					break;
+				}
+				repoDirectory = repoDirectory.Parent;
+			}
+			return isGitRepository;
 		}
 
 		void CreateGitIgnoreFile (FilePath solutionPath)


sevoku:
I like this idea, lets's add a `IVersionControlProjectTemplateHandler.CanCreateRepository (FilePath atPath)`. `FinalProjectConfigurationPage` could have an additional `ParentFolderChanged` event and the `NewProjectController` would subscribe to that and set `IsUseGitEnabled` based on the result of CanCreateRepository.

jsuarezruiz:
Matt, I liked your idea. Thanks guys for the feedback!. I have made changes to disable the Git option if a .git folder already exists (in the selected path).
However, I have some doubts. Should we show a warning icon with a popup and information in this case? Thus, in case of doubt the user knows why it is disabled. Another one, should we check for a .gitignore file and if exists disable the option to create the gitignore file?

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/ProjectTemplateHandler.cs
```diff
@@ -1,70 +1,85 @@
...
+namespace MonoDevelop.VersionControl.Git
+{
+	public class ProjectTemplateHandler : IVersionControlProjectTemplateHandler
+	{
+		public bool CanCreateRepository (FilePath filePath)
+		{
+			bool isGitRepository = false;
+			DirectoryInfo repoDirectory = new DirectoryInfo (filePath);
+
+			while (repoDirectory != null) {
```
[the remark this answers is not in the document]
Therzok: This also handles the case of a broken `.git` directory, so 👍 from me.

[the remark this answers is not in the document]
jsuarezruiz: Nice suggestion. Changed!

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkProjectConfigurationWidget.cs
```diff
@@ -201,7 +201,9 @@ void BrowseButtonClicked ()
 			FilePath startingFolder = GetStartingFolder ();
 			FilePath selectedFolder = BrowseForFolder (startingFolder);
 			if (selectedFolder != null) {
-				locationTextBox.Text = selectedFolder;
+				locationTextBox.Text = selectedFolder;
+				useGitCheckBox.Sensitive = projectConfiguration.IsUseGitEnabled;
+				useGitCheckBox.Active &= projectConfiguration.IsUseGitEnabled;
 			}
```
mrward: It has been a while since I looked at this code. If this changes the useGitCheckBox.Active state should we also be changing projectConfiguration.UseGit?

I am just wondering since the new project dialog always calls the CreateVersionControlItems method and the only time we do not call Run on the version control project template handler is if we are adding a new project to an existing solution.

https://github.com/mono/monodevelop/blob/c52a375198809d70b552565ff434aceb5711e4c3/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs#L935

The version control project template handler just checks the config.UseGit property. So is it possible to have git enabled, but then select a location where git already exists, and then create the solution and the version control system will then try to create git again?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
```diff
@@ -283,7 +283,13 @@ void CreateFinalConfigurationPage ()
 		{
 			finalConfigurationPage = new FinalProjectConfigurationPage (projectConfiguration);
 			finalConfigurationPage.ParentFolder = ParentFolder;
-			finalConfigurationPage.IsUseGitEnabled = IsNewSolution && (versionControlHandler != null);
+			finalConfigurationPage.IsUseGitEnabled = IsNewSolution && versionControlHandler != null;
+			finalConfigurationPage.ParentFolderChanged += (sender, e) => {
+				finalConfigurationPage.UseGit = versionControlHandler != null;
+				finalConfigurationPage.IsUseGitEnabled = IsNewSolution && versionControlHandler != null &&
+					versionControlHandler.CanCreateRepository (projectConfiguration.Location);
+				finalConfigurationPage.CreateGitIgnoreFile = versionControlHandler.CanCreateRepository (projectConfiguration.Location);
```
sevoku: Shouldn't thins check if .gitignore exists? Also versionControlHandler might be null.

jsuarezruiz: Added null validation.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/FinalProjectConfigurationPage.cs
```diff
@@ -179,7 +182,7 @@ public string GetValidSolutionName ()
 		}
 
 		public bool IsGitIgnoreEnabled {
-			get { return config.UseGit && IsUseGitEnabled && gitIgnoreEnabled; }
+			get { return config.UseGit && gitIgnoreEnabled && EnableCreateGitIgnore(); }
 		}
```
mrward: Can we add tests for this new logic?

There are some tests for the IsGitIgnoreEnabled flag in the NewProjectDialogTests.

Also should the IsUseGitEnabled check be removed here? That is used

https://github.com/mono/monodevelop/blob/f9839cdd26fd7710315b7ad7fcb7bc7b585223c9/main/tests/Ide.Tests/MonoDevelop.Ide.Projects/NewProjectDialogTests.cs#L273

jsuarezruiz:
The behavior with .gitignore with this changes:
- .gitginore exists in project folder (not parent, given the selected target folder already exists) > disable and uncheck checkbox.
- .gitignore exists in a parent folder > enable and uncheck checkbox.
- .gitignore does not exist (neither target nor parent folder) > enable and check the checkbox.

I have reviewed the tests too:
Screenshot 2019-06-10 at 12 42 57

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/FinalProjectConfigurationPage.cs
```diff
@@ -231,15 +239,19 @@ void CheckIsValid ()
...
 
 		void OnIsValidChanged ()
-		{
-			if (IsValidChanged != null) {
-				IsValidChanged (this, new EventArgs ());
-			}
-		}
-
+		{
+			IsValidChanged?.Invoke (this, new EventArgs ());
```
mkrueger: Use EventArgs.Empty here - EventArgs is a class this avoids the creation of a new object. (I see it was in old code but we should clean up such things)

jsuarezruiz:
Hey @mkrueger, thanks for the feedback but, what do you mean by the `EventArgs`?

sevoku:
@jsuarezruiz I think @mkrueger meant to use `EventArgs.Empty` instead of creating a new instance

jsuarezruiz:
Ah, ok!. Changed. Thanks guys.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/FinalProjectConfigurationPage.cs
```diff
@@ -259,6 +271,44 @@ public void UpdateFromParameters ()
...
+			if (Location == solutionPath) {
+				FilePath gitIgnoreFilePath = solutionPath.Combine (".gitignore");
+				return File.Exists (gitIgnoreFilePath);
+			}
+			return false;
+		}
+
+		public bool GitIgnoreFileExistsInParentFolder ()
+		{
+			IEnumerable<DirectoryInfo> di = GetAllParentDirectories (new DirectoryInfo(config.SolutionLocation));
```
Therzok: We could possibly simplify the code a lot by doing something similar to [here](https://github.com/mono/monodevelop/blob/7f62365a6a17301abc7c218452a1fe9a2e45d58f/main/src/addins/VersionControl/MonoDevelop.VersionControl.Subversion.Unix/MonoDevelop.VersionControl.Subversion.Unix/SvnClient.cs) ?

Therzok: Link got broken, here it is: https://github.com/mono/monodevelop/blob/7f62365a6a17301abc7c218452a1fe9a2e45d58f/main/src/addins/VersionControl/MonoDevelop.VersionControl.Subversion.Unix/MonoDevelop.VersionControl.Subversion.Unix/SvnClient.cs#L203-L211

sevoku: agree, it should be much simpler: https://github.com/mono/monodevelop/blob/7f62365a6a17301abc7c218452a1fe9a2e45d58f/main/src/addins/VersionControl/MonoDevelop.VersionControl.Subversion.Unix/MonoDevelop.VersionControl.Subversion.Unix/SvnClient.cs#L203-L211

jsuarezruiz: Right. I have applied changes to simplify that part.

PR #7356, opened 2019-03-19
[Ide] Prevent InvalidOperationException when calling GetCodeAnalysisProjectAsync with a token

garuma:
The way cancellation was previously registered in the method would easily lead to InvalidOperationException being thrown since the registration on the cancellation token was kept indefinitely. As such disposing the token source after the taskSource has long been completed would throw.

With this patch we keep and dispose the cancellation token registration as soon as possible and switch every TaskCompletionSource access to the `Try` pattern to avoid spurious failures.


slluis:

alanmcgovern:
Oops, i see jeremie already requested a backport :p

PR #7359, opened 2019-03-20
[Ide] Unload deleted projects and solutions.

mrward:
When a project file is deleted, or its parent directory is removed,
the project will be unloaded. If a solution file is deleted then a
message dialog will be displayed indicating that the solution will
be closed.

Deleting a solution's parent directory is not detected by the file
watcher.

Child items for a workspace are not handled. A workspace can contain
child workspaces or solutions. Only the top level workspace item can
be closed and workspaces do not currently support having unloaded
child items.

Fixes VSTS #807849 - Nothing happens if the open project gets removed


mrward:
Rebased against master-vnext. Closing this.

PR #7370, opened 2019-03-27
[aspnetcore] Extends dev certification check for aspnet.core 2.1 and higher

jtorres:
- Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/827640


rodrmoya:
hmm, was going to merge this, but to avoid further conflicts, @jtorres can you rebase on top of master-vnext?

PR #7381, opened 2019-04-03
[msbuild] Update to work with ToolsVersion=`Current`

radical:
- the recent msbuild update changed the toolsVersion from `15.0` to
`Current`.
- we add a second toolset `Current` to the app.config for the remote
builder.
- VSMac still tries to pick `15.0`
	- with older mono, msbuild will use the 15.0 toolset
	- with newer mono, vsmac asks for 15.0 but msbuild picks
	`current` instead.

- VSMac needs to pick `Current` for newer mono, which will be fixed in a
follow up PR.

Fixes VS bug #831219 - 6.x Mono MSBuild support is broken

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteBuildEngineManager.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteBuildEngineManager.cs
@@ -561,7 +561,7 @@ namespace MonoDevelop.Projects.MSBuild
 				ConfigFileUtilities.SetOrAppendSubelementAttributeValue (runtimeElement, "AppContextSwitchOverrides", "value", "Switch.System.IO.UseLegacyPathHandling=false");
 			}
 
-			foreach (var toolset in doc.Root.Elements ("msbuildToolsets").FirstOrDefault ()?.Elements ("toolset")) {
+			foreach (var toolset in doc.Root.Elements ("msbuildToolsets").FirstOrDefault ()?.Elements ("toolset") ?? Enumerable.Empty<XElement> ()) {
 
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteBuildEngineManager.cs
```diff
@@ -561,8 +561,7 @@ static void UpdateMSBuildExeConfigFile (TargetRuntime runtime, string sourceConf
 				ConfigFileUtilities.SetOrAppendSubelementAttributeValue (runtimeElement, "AppContextSwitchOverrides", "value", "Switch.System.IO.UseLegacyPathHandling=false");
 			}
 
-			var toolset = doc.Root.Elements ("msbuildToolsets").FirstOrDefault ()?.Elements ("toolset")?.FirstOrDefault ();
-			if (toolset != null) {
+			foreach (var toolset in doc.Root.Elements ("msbuildToolsets").FirstOrDefault ()?.Elements ("toolset")) {
```
mrward: I think this will throw a null reference exception if any of those elements are missing. I guess we should put this outside the foreach and check to see if it is null. Or use something like:

`foreach (var toolset in doc.Root.Elements ("msbuildToolsets").FirstOrDefault ()?.Elements ("toolset") ?? Enumerable.Empty<XElement> ()) {

mrward:
@jtorres has tested a local build of this with Mono 6.3 and it looks good. I created a local branch on mono/monodevelop and the tests all pass (ignoring the leak test problem which is unrelated).

Presumably we need this on master-vnext though instead of master?

Therzok (approved):

mrward:
This has been merged to master-vnext already.

PR #7396, opened 2019-04-07
Add support for VS2019 in winbuild.bat script

wjk:
Does what the title says. Not much else to say. 😄

### DIFF
--- a/main/winbuild.bat
+++ b/main/winbuild.bat
@@ -3,6 +3,8 @@ setlocal enableextensions enabledelayedexpansion
 
 if exist "%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe" (
 	FOR /F "delims=" %%E in ('"%ProgramFiles(x86)%\Microsoft Visual Studio\installer\vswhere.exe" -latest -property installationPath') DO (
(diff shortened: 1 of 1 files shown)


wjk:
Unfortunately, even after this change was applied I am so far unable to get MonoDevelop to build on Windows with VS2019. It's just one thing after another. The main issue I've been fighting is that the MSBuild assemblies shipped with VS2019 target net472, while MonoDevelop targets net461; I had to retarget the entire solution to get it to resolve MSBuild correctly. Unfortunately, this would break compatibility all sorts of ways if merged into upstream. I'm not entirely certain where to go from here, given that Windows support is a very low priority for MonoDevelop in this day and age.

Therzok:
Hey @wjk, currently, there's development happening on a private branch, and everything's been bumped to .net472 and this change has already been done there.

The changes will be merged back to the public repo soon, but there's no exact date that I know of right now.

KirillOsenkov:
This change is OK in and of itself, but I'd just wait for our fixes that would also update to net472 etc.

Therzok:
Support for VS2019 has been added via https://github.com/mono/monodevelop/commit/9b0bf92e1618d3e6663cd0db12ec2afcc74a19bb#diff-b67e00fd42ad8aeaab6794c311b3642a

Therzok:
Thanks for the PR!

PR #7432, opened 2019-04-30
Removed using Compile items directly from the project [WIP]

dellis1972:
If your project uses Conditions on `Compile` items, or runs a target before CoreCompile to dynamically add/remove items you will get IntelliSense errors. This is because monodevelop is including ALL `Compile` files in the csproj even if they are not really used. 

I think we really only need the results of `GetItemsFromCoreCompileDependenciesAsync` which will only include the bits the compiler will actually get. This seems to be how Visual Studio works. The attached sample project does not have any IntelliSense errors in it in VS 2019, however you will see errors in the IDE about the `EmptyClass.Config` property being declared already. The project will build perfectly fine however. 

[CompileRemoveTest.zip](https://github.com/mono/monodevelop/files/3131444/CompileRemoveTest.zip)

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -531,15 +531,11 @@ namespace MonoDevelop.Projects
 		/// </summary>
 		protected virtual async Task<ImmutableArray<ProjectFile>> OnGetSourceFiles (ProgressMonitor monitor, ConfigurationSelector configuration)
 		{
-			// pre-load the results with the current list of files in the project
-			var evaluatedItems = await GetEvaluatedSourceFiles (configuration);
-
 			// add in any compile items that we discover from running the CoreCompile dependencies
 			var coreCompileResult = await compileEvaluator.GetItemsFromCoreCompileDependenciesAsync (this, monitor, configuration);
 			var evaluatedCompileItems = coreCompileResult.SourceFiles;
 
-			var results = new HashSet<ProjectFile> (evaluatedItems, ProjectFileFilePathComparer.Instance);
-			results.UnionWith (evaluatedCompileItems);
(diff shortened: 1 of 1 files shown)


mrward:
We could do with tests for this.

There seems to be some whitespace changes here.

Also be interesting to know why GetEvaluatedSourceFiles is not working. It should be re-evaluating the project based on the current configuration so not sure why it is not ignoring items based on the condition. It seems like this should work:

    <Compile Include="EmptyClass.Debug.cs" Condition=" '$(Configuration)' == 'Debug' " />

mrward:
What intellisense errors are you seeing? I see one about a partial class which only has one class which seems correct. If I remove the partial from all the EmptyClass.cs files it seems OK even though there are two classes with the same name, since they are conditionally applied.

dellis1972:
@mrward thing is its not just `Condition` on the items. If an MSbuild task before `CoreCompile` uses

```
<Target Name="_Foo" BeforeTargets="CoreCompile">
<ItemGroup>
<Compile Remove="Foo.cs" />
</ItemGroup>
</Target>
```

That too will not be detected and `Foo.cs` will still be included in the items the Intellisense  uses.
I'll see if I can fix up the whitespace and add a unit test of some sort. Not very familiar with the monodevelop codebase so will need to have a dig around.

dellis1972:
@mrward ok weird.. I don't get the error now closed and reopened VSForMac. I'll try a repo with the `Target` problem.

mrward:
Yeah I can see that a condition might be different. The attached project does not use a target and seems to work.

With `<Target Name="_Foo" BeforeTargets="CoreCompile">` that is not going to be run by the project model when it evaluates the MSBuild items based on the current configuration. So I could see things not working there with the Compile Remove.

dellis1972:
OK, For some reason this is working 🤔 I'll close for now till I can repo again.

PR #7439, opened 2019-05-03
Bump minimum Mono version

rodrmoya:
This is needed because of the MSBuild/NuGet bumps done in Mono to support
the latest 2.1, 2.2 and 3.0 SDKs, which are MSBuild16/NuGet5-based. With
old Mono versions, even basic builds will fail with all sort of packaging,
missing targets, etc issues.


rodrmoya:

PR #7447, opened 2019-05-06
Fix several toolbox issues

slluis:
Fix several toolbox issues:

* SourceEditorView.cs, CodeTemplateToolboxProvider.cs: In the old editor, use a private
  toolbox provider for snippets. The global provider tries to get the
  editor from the consumer, and that's not possible anymore.

* ToolboxService.cs: Support views that provide several
  IToolboxDynamicProvider instances.

* Document.cs, DocumentController.cs: Added GetContent() method to
  DocumentController to get the content for the active view hierarchy.
  Removed some duplicated code in Document.

* DocumentView.cs, DocumentViewContainer.cs: When a view gets the focus, make sure it
  is set as active. This is important when in Split mode, since now
  clicking in the different views of the split will set it as active.

Also fixed some focus issues with tab views. When switching between tabs, ensure the activated tab gets the focus. Also fixed issue in Shown event, that was not properly invoked for
view containers. Added unit tests.


slluis:

slluis:
Reverted some changes in Document and DocumentController that were causing a test failure.

slluis:

PR #7449, opened 2019-05-06
Fix several builder creation issues

slluis:
To get a project builder RemoteBuildEngineManager first gets a build
engine and then it gets the project builder from it. The process of
getting the project builder was done inside the engine lock and that
was causing a lot of thread contention that lead to a collapse of
builder creation. To avoid this problem builders are now created
outside of the engine lock. With this change there is now the
possibility of the engine being shutdown before the builder can be
obtained from it. If that happens a recursive call is made to get a
new builder from a new engine.

This patch also improves the handling of engine shutdown. The project
builder creation logic now ensures that builders have a reference
count of 1 when they are created and registerd in the builders list.
This was done before in three steps: create -> register builder -> add
reference. In some cases this lead to the engine to be disposed before
the reference was added.

Fixes VSTS #841060 - Hang when opening a file and lots of background threads


slluis:

mrward (approved):

slluis:

slluis:

Therzok:
Is there anything I can help on with this PR?

slluis:
@Therzok find what is causing the test failures :)

Therzok:
Rebased this PR in order to see more recent test results.

Therzok:
Looking at the time it takes to finish the test runs, as there seems to be a slight regression.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/RemoteBuildEngine.cs
```diff
@@ -126,15 +126,17 @@ public Task<RemoteProjectBuilder> GetRemoteProjectBuilder (string projectFile)
...
 		/// Gets or creates a project builder for the provided project
 		/// </summary>
 		/// <returns>The remote project builder.</returns>
 		/// <param name="projectFile">Project to build.</param>
-		async Task<RemoteProjectBuilder> InternalGetRemoteProjectBuilder (string projectFile, bool create)
+		Task<RemoteProjectBuilder> InternalGetRemoteProjectBuilder (string projectFile, bool create)
 		{
 			if (IsShuttingDown)
-				return null;
+				return nullRemoteBuilderTask;
```
Therzok: What I meant is do not allocate task.

Therzok:
Debugging this lead to the conclusion that the build engine is disposed by the time connecting is attempted. Need to figure out why.

slluis:
@monojenkins rebase

PR #7457, opened 2019-05-07
Fixes VSTS Bug 849520: TextViewMargin.CreateLinePartLayout creates a

mkrueger:
large amount of (handled) exceptions

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/849520

Not sure if the stack trace can happen because it's from an former
version and I couldn't find any in my log. But I added some bounds
checking in case the highlighting engine gives a chunk that's no
longer inside the text bounds.


slluis:

PR #7461, opened 2019-05-07
Fix NRE when opening file

slluis:
Fixes VSTS #849798 - NullReferenceException in SetTitle because
window.Document is null


slluis:

PR #7464, opened 2019-05-07
Avoid unnecessary editor extension chain reinitializaiton

slluis:
Fixes VSTS #888272 - InitializeExtensionChain() called twice on editor start


slluis:

PR #7465, opened 2019-05-07
Fix NRE at startup

slluis:
It is handled, so it is not a crash.


slluis:

PR #7469, opened 2019-05-07
[Ide] Refresh build action when adding existing files to project

mrward:
If an existing file is copied or moved into a project then the
default build action from the project may be incorrect if the file
uses file globs. DefaultMSBuildEngine's FindGlobItemsIncludingFile
will not find a match if the file does not exist at the target
location. To fix this, if no build action was explicitly set when
adding an existing file to the project, the build action is
refreshed after the file has been copied or moved to the target
location.

This allows an existing .xaml file to be added to a Xamarin.Forms
.NET Standard project and for it to be added as an EmbeddedResource
instead of a None item.

Fixes VSTS 813456 - XAML files added are not actually added


mrward:

PR #7481, opened 2019-05-08
[release-8.1] Fix welcome window initialization

slluis:
With the latest changes to the initialization process (new service model) the welcome window initialization has been removed from the MD initialization process (the welcome page service implementation had to take care of showing the window). This resulted in some unexpected welcome window behaviour if it was shown too early before the command manager and the global menu were initialized.

This PR reduces command dependencies on the workbench and allows to initialize the global menu and the welcome window early enough to avoid noticable startup delays (before the workbench initialization starts), but not too early so the command system has enough information to build the global menu. At the same time MD gets back the control over when and how to show the welcome window.

Fixes VSTS #850578
Fixes VSTS #857394

Depends on https://github.com/xamarin/md-addins/pull/4552
Backport of https://github.com/mono/monodevelop/pull/7480

### DIFF
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -401,10 +401,16 @@ namespace MonoDevelop.MacIntegration
 			return map;
 		}
 
+		string currentCommandMenuPath, currentAppMenuPath;
+
 		public override bool SetGlobalMenu (CommandManager commandManager, string commandMenuAddinPath, string appMenuAddinPath)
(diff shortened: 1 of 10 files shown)


rodrmoya (approved):

slluis:
Green in md-addins.

sandyarmstrong:
Filed a new GTC hang after this merged: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/889788. On the whole, GTC init weirdness is much improved, though.

PR #7484, opened 2019-05-08
Bump fsharp

nosami:


nosami:
The reason the build started to fail yesterday was because Fantomas had been updated without bumping its version number. This has since been rectified and Fantomas is now including a build number on it's nightly feed.

PR #7495, opened 2019-05-08
Fix NRE when switching documents

slluis:
Fixes VSTS #889280 - Cannot move between xaml files
Fixes VSTS #889281 - IDE crashes when clicking on a document tab to swap
documents


slluis:

PR #7494, opened 2019-05-08
Fix scope of GetContent() for several views

slluis:
Reviewed calls to GetContent and changed several calls to limit the scope
to the current active view.

For ToolboxService, changed the scope to look for provider in all document.

Fixes VSTS #866686 - Xamarin.Forms previewer does not show toolbox content


slluis:

PR #7510, opened 2019-05-09
Fixes VSTS Bug 889278: Exceptions are thrown whenever a change is

mkrueger:
undone/redone in a xaml file

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/889278


mkrueger:
Another approach would be to remove the assertion. I'm not really sure how that works but I wouldn't invest much time in the old editor infrastructure. It's likely that the undo/redo system is not compatible to the vs.net infrastructure.

mkrueger:

PR #7512, opened 2019-05-09
Catch exception from having a bad LangVersion setting in .csproj

bratsche:
If LangVersion is an invalid value and throws an exception, the general
page fails to load. So now we catch this exception and display the
error on the page so the user can correct it.

Fixes VSTS #802073


mrward:
Not sure about the UI change. Maybe we could use an error icon instead of adding the text. Or instead we could just add the unknown version to the combo box and show something like `5555 (Unknown version)`. /cc @vancura 

LanguageOptions

Another problem here is that if you open the project options go to the Build - General page and click OK the language version in the project is changed to default. Not sure whether we should be changing the value here - nothing is shown as selected in the drop down but we are changing it when the OK button is clicked. Seems like we should try to keep the old value if the user does not change it.

vancura:
I'd do both: please add a `gtk-warning` in front of the whole line, and move the description inside the dropdown. Also, the text "Unknown LangVersion string is '5555'" is too robotic. Let's do what Matt says, "5555 (Unknown Version)" (just with capital V).

The line would look like:

`[gtk-warning] C# Language Version [5555 (Unknown Version)]`

(The second half of the line is the dropdown with its content)

Is this clear or do you need an image? :)

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CompilerOptionsPanelWidget.cs
```diff
@@ -108,7 +116,9 @@ public CompilerOptionsPanelWidget (DotNetProject project)
 						langVerStore.AppendValues (text, version);
 					}
 				} catch (Exception ex) {
-					label2.Markup = GettextCatalog.GetString ("C# Language Version (<b>{0}</b>):", ex.Message);
+					var badVersion = ExtractBadVersion (ex.Message);
+					label2.Markup = GettextCatalog.GetString ("C# Language Version [{0} (Unknown Version)]:", badVersion);
+					langVerStore.AppendValues (ExtractBadVersion (ex.Message), LanguageVersion.Preview);
```
mrward: Can we just read the langVersion property from the project instead of using a regex to get the information from the exception? Either using the MSBuildProject or the ProjectProperties. One of those should give you the raw string without having to parse the exception due to the enum conversion not working.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CompilerOptionsPanelWidget.cs
```diff
@@ -101,21 +109,32 @@ public CompilerOptionsPanelWidget (DotNetProject project)
...
-					langVerStore.AppendValues (text, version);
+				try {
+					if (unsupportedLanguageVersions.Contains (version) && compilerParameters.LangVersion != version) {
+						// Mono's MSBuild does not currently support C# 8.
+					} else {
+						langVerStore.AppendValues (text, version);
+					}
+				} catch (Exception ex) {
+					var badVersion = ExtractBadVersion (ex.Message);
+					label2.Markup = GettextCatalog.GetString ("C# Language Version [{0} (Unknown Version)]:", badVersion);
```
mrward: I was assuming the `BadLanguageVersion (Unknown version)` was in the combo box itself. The label would not be changed.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CompilerOptionsPanelWidget.cs
mrward: My idea of using the MSBuildProject or ProjectProperties is not right since the langversion can be configuration specific.

It looks like the [ProjectConfiguration.Properties](https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectConfiguration.cs#L169) could be used to get the raw lang version.

If not then I think another approach would be to have a custom exception in the [CSharpCompilerParameters.LangVersion](https://github.com/mono/monodevelop/blob/master/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CSharpCompilerParameters.cs#L225) which would have the LanguageVersion that is incorrect as a property on the exception so we could avoid having to use a regex.

bratsche:
@mrward I think I have everything fixed in this. If you have time would you mind taking a look again?

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CompilerOptionsPanelWidget.cs
```diff
@@ -98,26 +100,47 @@ public CompilerOptionsPanelWidget (DotNetProject project)
...
 			var langVerStore = new ListStore (typeof (string), typeof(LanguageVersion));
-			foreach (var (text, version) in CSharpLanguageVersionHelper.GetKnownLanguageVersions ()) {
-				if (unsupportedLanguageVersions.Contains (version) && compilerParameters.LangVersion != version) {
-					// Mono's MSBuild does not currently support C# 8.
-				} else {
-					langVerStore.AppendValues (text, version);
+			var langVersions = CSharpLanguageVersionHelper.GetKnownLanguageVersions ();
+			var badSet = new HashSet<string> ();
+
+			foreach (var (text, version) in langVersions) {
```
mrward: Not sure why we need a loop here - we would only have a single value. The compilerParameters.LangVersion has a single value so we should only have one bad value if it is invalid. Could we not just capture the value before the loop and wrap it in a try catch?

```
LangVersion langVersion;
string badVersion = null;

try {
    langVersion = compilerParameters.LangVersion;
} catch () {
    badVersion = configuration.Properties.GetProperty ("LangVersion").Value;
}

// Do the loop if no badVersion found.
```

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CompilerOptionsPanelWidget.cs
```diff
@@ -98,26 +100,47 @@ public CompilerOptionsPanelWidget (DotNetProject project)
...
+				TreeIter iter;
+				if (langVerStore.GetIterFirst (out iter)) {
+					do {
+						try {
+							var val = (LanguageVersion)(int)langVerStore.GetValue (iter, 1);
+							if (val == compilerParameters.LangVersion) {
+								langVerCombo.SetActiveIter (iter);
+								break;
+							}
+						} catch (Exception ex) {
```
mrward: Not sure we need a try/catch here. This should not fail since we already handle the bad compilerParameters.LangVersion.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CompilerOptionsPanelWidget.cs
```diff
@@ -98,26 +100,47 @@ public CompilerOptionsPanelWidget (DotNetProject project)
...
-					if (val == compilerParameters.LangVersion) {
-						langVerCombo.SetActiveIter (iter);
-						break;
-					}
-				} while (langVerStore.IterNext (ref iter));
+			if (badSet.Any ()) {
+				langVersionWarningIcon.Visible = true;
+				foreach (var s in badSet) {
+					var badIter = langVerStore.AppendValues (s, LanguageVersion.Default);
+					langVerCombo.SetActiveIter (badIter);
```
mrward: Can we get a screenshot? I am not sure we need all this text. I think just `{0} (Unknown Version)` in the combo box would be enough since the label already has the `C# Language Version` part.

If I OK the dialog it looks like we would change the project config's lang version to Default. Is that we want to happen here? I think we should just keep the value and not change the user's project. Also allow them to select a different value and then go back to the unknown version and still use it. It might require some changes to how we store the data in the TreeStore, or maybe we could just keep the bad value and a way to identify it was selected and special case that on saving.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CompilerOptionsPanelWidget.cs
bratsche: I think we need that loop because we are also populating the ListStore with all the other language versions. But yeah, we can assume there is only one bad version in the list I guess and get rid of the HashSet and the loop later that iterates through it.

Therzok: LGTM

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CompilerOptionsPanelWidget.cs
```diff
@@ -98,26 +100,41 @@ public CompilerOptionsPanelWidget (DotNetProject project)
...
+				try {
+					if (unsupportedLanguageVersions.Contains (version) && compilerParameters.LangVersion != version) {
+						// Mono's MSBuild does not currently support C# 8.
+					} else {
+						langVerStore.AppendValues (text, version);
+					}
+				} catch (Exception ex) {
+					badVersion = configuration.Properties.GetProperty ("LangVersion").Value;
 				}
 			}
```
mrward: Still seems to be a potential problem here. If the `if` causes an exception to be thrown when the compilerParameters.LangVersion is checked, before all lang versions are appended, then the combo box would be missing items. I think we should get the lang version from the compilerParameters.LangVersion outside this loop inside a try/catch, store that in a nullable variable, handle that as a bad version. Then in the loop only check the lang version if it is not null in that if statement.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CompilerOptionsPanelWidget.cs
```diff
@@ -98,26 +100,41 @@ public CompilerOptionsPanelWidget (DotNetProject project)
...
-				do {
-					var val = (LanguageVersion)(int)langVerStore.GetValue (iter, 1);
-					if (val == compilerParameters.LangVersion) {
-						langVerCombo.SetActiveIter (iter);
-						break;
-					}
-				} while (langVerStore.IterNext (ref iter));
+			if (badVersion != null) {
+				var badIter = langVerStore.AppendValues (GettextCatalog.GetString ("{0} (Unknown Version)", badVersion), LanguageVersion.Default);
+				langVerCombo.SetActiveIter (badIter);
```
mrward: This still looks like we are re-setting the value back to Default if the user clicks OK in preferences. Is that what we want to do here? If so I think we should add a comment indicating that is what will happen here in the code.

bratsche: Thanks, I think I've got this fixed now. I added another column in the list store to determine if the value is a "bad" value, and if so then when we store things I don't set the LangVersion. So if you hit OK then it doesn't affect the language version.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CompilerOptionsPanelWidget.cs
bratsche: Thanks, makes sense now. I've changed this in the latest commit to be (I think) what you're describing.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CompilerOptionsPanelWidget.cs
```diff
@@ -98,26 +100,44 @@ public CompilerOptionsPanelWidget (DotNetProject project)
...
+				badVersion = configuration.Properties.GetProperty ("LangVersion").Value;
+			}
 
-			var langVerStore = new ListStore (typeof (string), typeof(LanguageVersion));
-			foreach (var (text, version) in CSharpLanguageVersionHelper.GetKnownLanguageVersions ()) {
-				if (unsupportedLanguageVersions.Contains (version) && compilerParameters.LangVersion != version) {
+			foreach (var (text, version) in langVersions) {
+				if (unsupportedLanguageVersions.Contains (version)) {
 					// Mono's MSBuild does not currently support C# 8.
 				} else {
```
mrward: I think we are nearly there.

The original code did:

`if (unsupportedLanguageVersions.Contains (version) && compilerParameters.LangVersion != version) {`

Now we are only doing:

`if (unsupportedLanguageVersions.Contains (version)) {`

The idea behind the original code was that even if VS Mac does not support 8.0 yet, if the project used that value we would include it in the list. Now it looks like we would not show 8.0 if the project used it.

bratsche: So we should just remove that check and append the values anyway?

mrward: No. The idea was that you should not be able to select an unsupported lang version if Mono does not support it for a new project. However if you already have that configured, then we allow it.

bratsche: Okay, I think that's fixed now. Thanks!

vancura:
Wow, isn't that the internal Gtk warning icon? I haven't seen that in ages :)

vancura:
We have so many issues in these elements, but I am not sure we have time/manpower to change this now, especially as we'll eventually have the new Settings:

image

Different horizontal alignment:

image

Top left: the real `md-warning` icon, `16pt`.
Bottom right: the icon rendered with our code, `24pt`. No idea why. It's upscaled and blurry.

image

See this space:

image

It's not in our latest example (this issue): 

image

Is there anything we can do about these issues?

bratsche:
Screen Shot 2019-06-05 at 3 51 21 PM

@vancura This looks like it improves things in the area of the dialog I'm working on. Does this look okay to you?

vancura:
This is amazing, @bratsche! Thanks a lot!

PR #7513, opened 2019-05-09
[GTC] Fixes current window child hierarchy (parents) using RunCustomDialog

netonjm:
This fixes the way of we show dialogs in our Messaging.Service, seems BeginModalSession is not parenting correctly the child windows and provokes unexpected results like odd flashes when changing desktops (CMD + LEFT/RIGHT ARROW).

Also now like a child window is attached to it's parent and it moves accordingly with it

Fixes VSTS #823695 - About info flashes briefly if left open while switching desktops
Fixes VSTS #844450 - Run/RunWith/CustomParameters window flashes briefly if left open while switching desktops
Fixes VSTS #844452 - Project/Set Startup project flashes briefly if left open while switching desktops
Fixes VSTS #899539

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
@@ -392,29 +392,23 @@ namespace MonoDevelop.Ide
 			var nsdialog = GtkMacInterop.GetNSWindow (dialog);
 			// Make the GTK window modal WRT the current modal NSWindow
 
+			NSWindow nativeParent = null;
 			try {
-				var nativeParent = (NSWindow)parent;
+				nativeParent = parent;
 				nativeParent.AddChildWindow (nsdialog, NSWindowOrderingMode.Above);
-				var s = NSApplication.SharedApplication.BeginModalSession (nsdialog);
-
-				EventHandler unrealizer = null;
-				unrealizer = delegate {
-					if (nativeParent != null) {
-						nativeParent.RemoveChildWindow (nsdialog);
-					}
-					NSApplication.SharedApplication.EndModalSession (s);
-					dialog.Unrealized -= unrealizer;
-				};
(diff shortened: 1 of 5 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
```diff
@@ -384,20 +384,38 @@ public static int RunCustomDialog (Dialog dlg, Window parent)
...
 
-			EventHandler unrealizer = null;
-			unrealizer = delegate {
-				NSApplication.SharedApplication.EndModalSession (s);
-				dialog.Unrealized -= unrealizer;
-			};
-			dialog.Unrealized += unrealizer;
+			try {
+				var nativeParent = (NSWindow)parent;
+				nativeParent.AddChildWindow (nsdialog, NSWindowOrderingMode.Above);
```
Therzok: Hmm, why would this throw? Also, don't we need to handle removing the unrealizer?

netonjm: I think it was because the direct cast from Window

Therzok: We should avoid throwing an exception if we can. And yes, I misread the order of actions, unrealizer is called on all the cases.

netonjm:
Please do not merge! I'm hitting and issue trying to 
nativeParent.AddChildWindow (nsdialog, NSWindowOrderingMode.Above); 
when nativeParent is a gtk view

netonjm:
rebased! ands seems to work now

sevoku:
I'm not sure whether/how it will work with nested modals... I'll verify...

sevoku:

sevoku:
Soo, I've simplified the code a bit. Tested different scenarios incl. publishing wizard (native) with gtk children. Seems to be working.

sevoku:

PR #7514, opened 2019-05-09
Fix focus notification issue with the old editor

slluis:
The ITextView implementation relied on the ActiveDocumentChanged
event to be raised after the editor is fully initialized. The handler of that
event would call QueueAggregateFocusCheck() to update the focus
status of the text view.

Since views are now loaded asynchronously, ActiveDocumentChanged is now raised
before the editor is loaded, so QueueAggregateFocusCheck is not called.
To fix this issue, QueueAggregateFocusCheck is now explicitly called once
the editor is initialized.

Fixes VSTS #807522 - [WebToolsE2E][VSforMac]Type keyword in cshtml file, there
is no intellisense.

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/MonoTextEditor.ITextView.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/Mono.TextEditor/Gui/MonoTextEditor.ITextView.cs
@@ -405,8 +405,7 @@ namespace Mono.TextEditor
 		/// </remarks>
 		private void SubscribeToEvents ()
 		{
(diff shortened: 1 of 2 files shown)


slluis:

slluis:

KirillOsenkov:
@slluis thanks! I've confirmed that this fixes both remaining WebTools issues (completion in Razor C# works again and smart indent is correct)

slluis:
@KirillOsenkov which other bug does this fix?

KirillOsenkov:
I didn't file the Smart Indent one because I suspected it would be the same root cause.

PR #7516, opened 2019-05-09
[Mac] Run Preferred32bit applications in x64

Therzok:
macOS has deprecated 32bit applications, thus Preferred32bit should
run applications in 64bit, as it has no impact on the assembly, like
Required32Bit - we should check how to handle that when 32bit is completely
deprecated.

Fixes VSTS #890092 - NUnit tests are executed in a 32-bit process
Fixes https://dev.azure.com/devdiv/DevDiv/_workitems/edit/749922


Therzok:

mrward (approved):

PR #7520, opened 2019-05-09
[GTC] Recent projects don't show up as disabled right away if you remove the actual project in Finder

jsuarezruiz:
We capture any change in the recent files. In case of modifying the name of the file, delete it, etc. We notify the change.

Fixes VSTS #738481

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentFileStorage.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentFileStorage.cs
@@ -33,14 +33,13 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
 using System.Text;
-using System.Threading;
 using System.Xml;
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentFileStorage.cs
```diff
@@ -103,7 +103,12 @@ void EnableWatching ()
 			watcher.Changed += FileChanged;
 			watcher.Deleted += FileChanged;
 			watcher.Renamed += HandleWatcherRenamed;
-			watcher.EnableRaisingEvents = true;
+			watcher.EnableRaisingEvents = true; 
+            if (recentFilesWatcher == null) {
```
Therzok:
```suggestion
			if (recentFilesWatcher == null) {
```

jsuarezruiz: Changed!

PR #7524, opened 2019-05-09
[GTC] Support reordering pinned projects

jsuarezruiz:
Added a new method into the RecentItems to allow reorder.

Fixes VSTS #842643

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentOpen.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentOpen.cs
@@ -127,7 +127,16 @@ namespace MonoDevelop.Ide.Desktop
 				LoggingService.LogError ("Failed to add item to recent files list.", e);
 			}
 		}
-		
+
+		public override void ReorderProject (int from, int to)
+		{
+			try {
+				ReorderFavoriteFile (from, to);
+			} catch (Exception e) {
+				LoggingService.LogInternalError ("Failed to reorder favorite item from recent files list.", e);
+			}
+		}
+
 		public override void NotifyFileRemoved (string fileName)
 		{
 			try {
@@ -199,6 +208,7 @@ namespace MonoDevelop.Ide.Desktop
 		public abstract void ClearFiles ();
 		public abstract void AddFile (string fileName, string displayName);
 		public abstract void AddProject (string fileName, string displayName);
+		public abstract void ReorderProject (int from, int to);
 		public abstract void NotifyFileRemoved (string filename);
 		public abstract void NotifyFileRenamed (string oldName, string newName);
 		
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentOpen.cs
```diff
@@ -227,6 +237,19 @@ internal void SetFavoriteFile (FilePath file, bool favorite)
 			PropertyService.SaveProperties ();
 		}
 
+		internal void ReorderFavoriteFile (int from, int to)
```
netonjm: Why we need this method? Could not we move the code to ReorderProject?

jsuarezruiz: Yes, we could do it without problems because it is only used from one place.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentOpen.cs
```diff
@@ -199,6 +208,7 @@ public IList<RecentFile> GetFiles ()
 		public abstract void ClearFiles ();
 		public abstract void AddFile (string fileName, string displayName);
 		public abstract void AddProject (string fileName, string displayName);
+		public abstract void ReorderProject (int from, int to);
```
sevoku: Why abstract? Favorites are not overridable, it's purely internal to `RecentFiles`.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentOpen.cs
sevoku: This should be just here in `RecentFiles`, no need to make it abstract or virtual.

sevoku: Wouldn't `ReorderFavoriteFile (FilePath file, FilePath beforeFile)` be more convenient?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentOpen.cs
```diff
@@ -227,6 +237,19 @@ internal void SetFavoriteFile (FilePath file, bool favorite)
 			PropertyService.SaveProperties ();
 		}
 
+		internal void ReorderFavoriteFile (int from, int to)
+		{
+			if (!favoriteFiles.Any ())
+				return;
+
+			string item = favoriteFiles [from];
```
sevoku: this will throw if the index is not valid, for convenience I'd rather use file names directly and just throw if `favoriteFiles.IndexOf (file) < 0`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentOpen.cs
```diff
@@ -224,6 +214,28 @@ public void AddFile (string fileName, WorkspaceObject project)
...
+		{
+			try {
+				if (!favoriteFiles.Any ())
+					return;
+
+				if (favoriteFiles.IndexOf (from) < 0 || favoriteFiles.IndexOf (to) < 0)
+					throw new ArgumentException ("The projects are not on the list.");
+
+				int fromIndex = favoriteFiles.IndexOf (from);
+				int toIndex = favoriteFiles.IndexOf (to);
```
sevoku: Please do the `<0` check here to avoid double lookup and add the param name to the `ArgumentException`

netonjm:
Should we include some tests?

jsuarezruiz:
Yes. Will include some unit tests!

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/RecentOpen.cs
jsuarezruiz: Changed.

netonjm:
tests!

PR #7526, opened 2019-05-09
[DotNetCore] Improve SDK not supported message

rodrmoya:
Clearly specify what versions the user has to install, to avoid confusion.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/889382

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
@@ -180,14 +180,14 @@ namespace MonoDevelop.DotNetCore
 			SdkRootPath = path;
 		}
 
-		internal static string GetNotSupportedVersionMessage (string currentPath, string version = "")
+		internal static string GetNotSupportedVersionMessage (string version = "")
 		{
 			string GetMessage (DotNetCoreVersion currentVersion)
 			{
 				return GettextCatalog.GetString ("NET Core {0}.{1} SDK version {2} is not compatible with this version of Visual Studio for Mac. Install the latest update to the .NET Core {0}.{1} SDK by visiting {3}.", currentVersion.Major, currentVersion.Minor, currentVersion.ToString (), DotNetCoreDownloadUrl.GetDotNetCoreDownloadUrl (currentVersion));
 			}
 
-			var installedVersion = DotNetCoreSdk.Versions.OrderByDescending (x => x).FirstOrDefault ();
+			var installedVersion = Versions.OrderByDescending (x => x).FirstOrDefault ();
 			if (installedVersion != null) {
 				if (installedVersion < DotNetCoreVersion.MinimumSupportedSdkVersion) {
 					return GetMessage (installedVersion);
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
@@ -179,5 +179,26 @@ namespace MonoDevelop.DotNetCore
 		{
 			SdkRootPath = path;
 		}
+
+		internal static string GetNotSupportedVersionMessage (string currentPath, string version = "")
+		{
+			string GetMessage (DotNetCoreVersion currentVersion)
+			{
+				return GettextCatalog.GetString ("NET Core {0}.{1} SDK version {2} is not compatible with this version of Visual Studio for Mac. Install the latest update to the .NET Core {0}.{1} SDK by visiting {3}.", currentVersion.Major, currentVersion.Minor, currentVersion.ToString (), DotNetCoreDownloadUrl.GetDotNetCoreDownloadUrl (currentVersion));
+			}
+
+			var installedVersion = DotNetCoreSdk.Versions.OrderByDescending (x => x).FirstOrDefault ();
+			if (installedVersion != null) {
+				if (installedVersion < DotNetCoreVersion.MinimumSupportedSdkVersion) {
+					return GetMessage (installedVersion);
+				} else if (installedVersion.Major == 2 && installedVersion.Minor == 2 && installedVersion < DotNetCoreVersion.MinimumSupportedSdkVersion22) {
+					return GetMessage (installedVersion);
+				} else if (installedVersion.Major == 3 && installedVersion < DotNetCoreVersion.MinimumSupportedSdkVersion30) {
+					return GetMessage (installedVersion);
+				}
+			}
+
+			return GettextCatalog.GetString (".NET Core {0} SDK is required to build this application, and is not installed. Install the latest update to the .NET Core 2.2 SDK by visiting {1}.", version, DotNetCoreDownloadUrl.GetDotNetCoreDownloadUrl (version));
+		}
 	}
 }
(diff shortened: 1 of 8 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreVersion.cs
```diff
@@ -227,5 +228,21 @@ public static bool IsSdkSupported (DotNetCoreVersion version)
...
 			return false;
 		}
+
+		public static string GetNotSupportedVersionMessage (string currentPath, string version = "")
+		{
+			var installedVersion = DotNetCoreSdk.Versions.OrderByDescending (x => x).FirstOrDefault ();
+			if (installedVersion != null) {
+				if (installedVersion < MinimumSupportedSdkVersion) {
+					return GettextCatalog.GetString ("The version of the .NET Core SDK currently installed ({0}) is not supported, please download the minimum supported version ({1})", installedVersion.ToString (), MinimumSupportedSdkVersion.ToString ());
+				} else if (installedVersion.Major == 2 && installedVersion.Minor == 2 && installedVersion < MinimumSupportedSdkVersion22) {
```
mrward: Not sure about the wording here `the minimum supported version`. Do we want them to download the minimum version? That seems to be what the message is suggesting. Our updater/installer might be offering a different version. //cc @vancura 

Also these three messages are the same - would it make sense to convert them into a single method that takes the installed and minimum version?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreVersion.cs
```diff
@@ -227,5 +228,21 @@ public static bool IsSdkSupported (DotNetCoreVersion version)
...
+					return GettextCatalog.GetString ("The version of the .NET Core SDK currently installed ({0}) is not supported, please download the minimum supported version ({1})", installedVersion.ToString (), MinimumSupportedSdkVersion22.ToString ());
+				} else if (installedVersion.Major == 3 && installedVersion < MinimumSupportedSdkVersion30) {
+					return GettextCatalog.GetString ("The version of the .NET Core SDK currently installed ({0}) is not supported, please download the minimum supported version ({1})", installedVersion.ToString (), MinimumSupportedSdkVersion30.ToString ());
+				}
+			}
+
+			return GettextCatalog.GetString (".NET Core {0} SDK is not installed. This is required to build and run .NET Core {0} projects.", version);
+		}
 	}
```
mrward: We will also need to fix the test that assumes .NET Core 2.2 is installed.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreVersion.cs
rodrmoya: Right, changed now both things

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreVersion.cs
rodrmoya: Right again, fixing

rodrmoya:
Maybe `GetDownloadUrl` could be in `DotNetSdk`?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreVersion.cs
```diff
@@ -227,5 +228,49 @@ public static bool IsSdkSupported (DotNetCoreVersion version)
...
+		{
+			//special case for 2.0, 3.0, ..
+			if (version.Minor == 0)
+				return string.Format (DotNetCoreDownloadUrl, version.Major);
+
+			return string.Format (DotNetCoreDownloadUrl, $"{version.Major}{version.Minor}");
+		}
+
+		public static string GetNotSupportedVersionMessage (string currentPath, string version = "")
+		{
```
mrward: currentPath does not seem to be used.

mrward:
Error messages about unsupported versions and download urls seem separate to the DotNetCoreSdk class to me. Moving GetNotSupportedVersionMessage to DotNetCoreSdk might be OK. Download urls still seem separate to me. The DotNetCoreSdk class is used for finding the installed sdks.

rodrmoya:
See https://github.com/xamarin/md-addins/pull/4595 for the md-addins part

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
```diff
@@ -179,5 +179,26 @@ internal static void SetSdkRootPath (string path)
...
+				if (installedVersion < DotNetCoreVersion.MinimumSupportedSdkVersion) {
+					return GetMessage (installedVersion);
+				} else if (installedVersion.Major == 2 && installedVersion.Minor == 2 && installedVersion < DotNetCoreVersion.MinimumSupportedSdkVersion22) {
+					return GetMessage (installedVersion);
+				} else if (installedVersion.Major == 3 && installedVersion < DotNetCoreVersion.MinimumSupportedSdkVersion30) {
+					return GetMessage (installedVersion);
+				}
+			}
+
+			return GettextCatalog.GetString (".NET Core {0} SDK is required to build this application, and is not installed. Install the latest update to the .NET Core 2.2 SDK by visiting {1}.", version, DotNetCoreDownloadUrl.GetDotNetCoreDownloadUrl (version));
```
jtorres: hmm shouldn't we indicate the expected SDK, not the ` 2.2 SDK` one?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
```diff
@@ -179,5 +179,26 @@ internal static void SetSdkRootPath (string path)
 		{
 			SdkRootPath = path;
 		}
+
+		internal static string GetNotSupportedVersionMessage (string version = "")
+		{
+			string GetMessage (DotNetCoreVersion currentVersion)
+			{
+				return GettextCatalog.GetString ("NET Core {0}.{1} SDK version {2} is not compatible with this version of Visual Studio for Mac. Install the latest update to the .NET Core {0}.{1} SDK by visiting {3}.", currentVersion.Major, currentVersion.Minor, currentVersion.ToString (), DotNetCoreDownloadUrl.GetDotNetCoreDownloadUrl (currentVersion));
```
jtorres: here the same `Install the latest update to the .NET Core {0}.{1}` so that `{0}{1}` should be the expected version

Therzok:
LGTM

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
rodrmoya: Right, fixed now

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
rodrmoya: Unni suggested to not mention a specific version, just point users to the "latest 2.x", so this is on purpose, as we don't have to tell people to install 2.2.1543 when maybe there's a much newer 2.2.3600

rodrmoya:

PR #7528, opened 2019-05-09
[DotNetCore] Split out SDK style project support

mrward:
Move SDK style project specific code out of the .NET Core addin into
MonoDevelop.Core. NuGet specific code has been moved into the NuGet
addin. The DotNetCoreProjectExtension was supporting all SDK style
projects now it supports .NET Core and .NET Standard projects that
use the .NET Core SDK.

Fixes VSTS #830544 - Split out SDK style project support from the
.NET Core addin


main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/PackageManagementSdkProjectExtension.cs
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		public override void Dispose ()
+		{
+			Project.Modified -= OnProjectModified;
+			base.Dispose ();
+		}
+
+		/// <summary>
+		/// Cannot use SolutionItemExtension.OnModified. It does not seem to be called.
```
slluis: How so? It should be called.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/TargetFrameworkExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+using MonoDevelop.Core.Assemblies;
+
+namespace MonoDevelop.Projects
+{
+	public static class TargetFrameworkExtensions
```
slluis: Maybe those extension methods could not be added to TargetFrameworkMoniker as properties.

slluis (approved):
:+1:

main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/PackageManagementSdkProjectExtension.cs
mrward: It looks like SolutionItemExtension.OnModified was fixed at some point after this  - https://github.com/mono/monodevelop/commit/c35ef1ac44c5da7a326862a96e4b093e1930d3ac

mrward: Although switching to use SolutionItemExtension.OnModified seems to cause code to be run earlier than before, which breaks the logic, so I may leave it as is.

slluis: You can try calling base.OnModified() first, and then the handler code.

mrward: Yeah that is what I tried. The problem using SolutionItemExtension.OnModified is that it gets called on initial load so the initial setting of the FileName will cause this to be called.

SolutionItemExtension.OnModified could be used - the project extension would need to have a flag to ignore calls to OnModified until ItemReady has been called.

mrward:
Update based on the review feedback. Moved TargetFrameworkExtensions into the TargetFrameworkMoniker class as a new property. Using SolutionItemExtension.OnModified method instead of using the Project.Modified event.

slluis:
It looks great!

jtorres (approved):
neat! 👍

PR #7529, opened 2019-05-09
Fix document focus issues

slluis:
Fixes VSTS #832026 - Double click in Search Results for XAML file does not
set focus to the edit window

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentView.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentView.cs
@@ -239,7 +239,7 @@ namespace MonoDevelop.Ide.Gui.Documents
 
 		static WeakReference<DocumentView> lastFocusGrabRequest = new WeakReference<DocumentView> (null);
 
-		public void GrabFocus ()
+		public async void GrabFocus ()
 		{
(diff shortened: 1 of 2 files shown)


slluis:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentView.cs
```diff
@@ -239,7 +239,7 @@ internal virtual void OnActivated ()
 
 		static WeakReference<DocumentView> lastFocusGrabRequest = new WeakReference<DocumentView> (null);
 
-		public void GrabFocus ()
+		public async void GrabFocus ()
```
Therzok: If anyone does GrabFocus, it will not focus the view by the time GrabFocus has finished.

So we might have a race like:
```
    documentView.GrabFocus();
    bool hasFocus = documentView.HasFocus; // will return false
```

slluis: I don't think that's a problem.

Therzok: It could be for integration tests though, since we don't actually focus until later on in the method.

PR #7408, opened 2019-04-12
Workaround for .NET Core debugger

usr-sse2:
Fixes issue #6793 until `vsdbg` supports filtering by exception type. Now it offers only "all" and "user-unhandled" exception filters.

PR #7531, opened 2019-05-09
[Debugger] Refactored unit tests to better allow testing of the DotNe…

jstedfast:
…tCore debugger backend

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
+++ b/main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
@@ -1,45 +1,280 @@
 using System;
 using System.Linq;
+using System.Text;
+using System.Globalization;
+
 using Microsoft.VisualStudio.Shared.VSCodeDebugProtocol.Messages;
+
 using Mono.Debugging.Backend;
 using Mono.Debugging.Client;
+using Mono.Debugging.Evaluation;
 
 namespace MonoDevelop.Debugger.VsCodeDebugProtocol
 {
 	class VSCodeObjectSource : IObjectValueSource
 	{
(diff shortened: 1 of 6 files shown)


jstedfast:
This depends on https://github.com/xamarin/md-addins/pull/4611 getting merged.

main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
```diff
@@ -1,38 +1,271 @@
...
-			this.variablesReference = variablesReference;
 			this.parentVariablesReference = parentVariablesReference;
-			this.val = val;
+			this.variablesReference = variablesReference;
+			this.evalName = evalName;
+			this.frameId = frameId;
+
+			if (type == null) {
+				if (IsCSError (118, "is a namespace but is used like a variable", val, out string ns)) {
+					this.display = this.name = this.val = ns;
```
mrward: Not sure where this error is from - are these messages always in English? Just wondering what would happen if the machine's default locate is not using English.

jstedfast: Yea, I thought about that too... you're probably right that I should ignore the message string and depend only on the CS error code.

jstedfast: Altho... https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0118 - CS0118 is not limited to a namespace being used as a variable.

Same for CS0119: https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0119

mrward: Another contributor was doing something similar with the .NET Core debugger - they tested their changes with Russian - https://github.com/mono/monodevelop/pull/7408#issuecomment-488247168 - maybe it would just work :)

PR #7532, opened 2019-05-09
[Dock] Don't disable hover when hiding a dock item

iainx:
Hiding and showing a dock item causes any hovers within the next 1.5seconds to be ignored.
It's possible to move into the dockbar to try to show a new pad within that 1.5 seconds, causing the pad to not appear.

Fix this by only disable the hover when showing the pad, not hiding.

Fixes VSTS #851588


slluis:

PR #7538, opened 2019-05-10
Fixes VSTS Bug 825007: [Feedback] Log files for Visual Studio 2019

mkrueger:
Preview crash

https://dev.azure.com/devdiv/DevDiv/_workitems/edit/825007

I doubt that this is responsible for the crash but it ensures, that
newMarkers is always != null at that point of the IdleHandler.


slluis:

PR #7539, opened 2019-05-10
Implement VSTS User Story 859993: Redesign and Improve our Search Engine

mkrueger:
Improves performance of find in files.

Fixes VSTS #959361


mkrueger:
It's already a lot faster but it's not yet finished.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/PatternSearcher.cs
```diff
@@ -0,0 +1,142 @@
...
+			while ((index = Find (text, index, end)) > -1) {
+				result.Add (index);
+				index++;
+			}
+			return result.ToArray ();
+		}
+
+		public int Find (string text, int startIndex, int endIndex)
+		{
+			return Find (text.AsSpan (), startIndex, endIndex);
```
Therzok:
```suggestion
			return Find (text.AsSpan (startIndex, endIndex - startIndex));
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/PatternSearcher.cs
```diff
@@ -0,0 +1,142 @@
...
+		}
+
+		public int Find (ReadOnlySpan<char> text, int startIndex, int endIndex)
```
Therzok: We don't need this overload if we use ReadOnlySpan, as ReadOnlySpan is a slice.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/PatternSearcher.cs
```diff
@@ -0,0 +1,142 @@
...
+		readonly int [] skipTable;
+
+		readonly bool isCaseSensitive;
+		readonly bool wholeWordsOnly;
+
+		public PatternSearcher (string pattern, bool isCaseSensitive, bool wholeWordsOnly)
+		{
+			this.isCaseSensitive = isCaseSensitive;
+			this.wholeWordsOnly = wholeWordsOnly;
+			patternArray = this.isCaseSensitive ? pattern.ToCharArray () : pattern.ToLowerInvariant ().ToCharArray ();
```
Therzok: This could be a ReadOnlySpan too.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
```diff
@@ -0,0 +1,247 @@
...
+			var scope = Scope.Create (model);
+			if (scope == null)
+				return;
+			currentScope = scope;
+			fileProviderTask = Task.Run (async delegate {
+				return await scope.GetFilesAsync (model);
+			});
+		}
+
+		internal static void SearchReplace (FindInFilesModel model, Scope scope, Task<IReadOnlyList<FileProvider>> getFilesTask, System.Action UpdateStopButton, System.Action<SearchResultPad> UpdateResultPad)
```
Therzok:
```suggestion
		internal static void SearchReplace (FindInFilesModel model, Scope scope, Task<ImmutableArray<FileProvider>> getFilesTask, System.Action UpdateStopButton, System.Action<SearchResultPad> UpdateResultPad)
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesModel.cs
```diff
@@ -0,0 +1,147 @@
...
+		public string FileMask {
+			get {
+				return file_mask;
+			}
+			set {
+				file_mask = value;
+				if (file_mask == null) {
+					split_file_masks = null;
+				} else {
+					split_file_masks = file_mask.Split (separators, StringSplitOptions.RemoveEmptyEntries);
```
Therzok: Style, replace whole if/else blocks with:
```csharp
split_file_masks = file_mask?.Split (separators, StringSplitOptions.RemoveEmptyEntries);
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
```diff
@@ -238,20 +233,21 @@ IEnumerable<SearchResult> RegexSearch (FindInFilesModel model, ProgressMonitor m
...
+			var findResults = model.PatternSearcher.FindAll (content);
+			var result = new SearchResult [findResults.Length];
+			for (var i = 0; i < findResults.Length; i++) {
+				result[i] = new SearchResult (provider, findResults [i], model.FindPattern.Length);
 			}
+			return result;
 		}
 
-		public void Replace (FileProvider provider, IEnumerable<SearchResult> searchResult, string replacePattern)
+		public void Replace (FileProvider provider, SearchResult[] searchResult, string replacePattern)
```
Therzok: Maybe use ImmutableArray here?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/PatternSearcher.cs
mkrueger: Span is just length + ptr - doesn't make a difference.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
mkrueger: 100% not.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
mkrueger: To make things even slower ? No.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/PatternSearcher.cs
mkrueger: We don't need the overload with the string when the framework will change - .NET core doesn't need it.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesModel.cs
mkrueger: Y makes sense - not my code btw.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/PatternSearcher.cs
mkrueger: No because it's not a reference type.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/PatternSearcher.cs
Therzok: Cleans up code. You don't need to track indices with a span and you're already using span.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/PatternSearcher.cs
Therzok: We don't have .NET Core now. And having ROSpan and start/end is weird as API.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
```diff
@@ -209,37 +180,34 @@ public IEnumerable<SearchResult> FindAll (Scope scope, ProgressMonitor monitor,
...
 
-		// Took: 17743
-
-		IEnumerable<SearchResult> FindAll (ProgressMonitor monitor, FileProvider provider, TextReader content, string pattern, string replacePattern, FilterOptions filter)
+		SearchResult[] FindAll (FindInFilesModel model, ProgressMonitor monitor, FileProvider provider)
 		{
-			if (string.IsNullOrEmpty (pattern))
-				return Enumerable.Empty<SearchResult> ();
+			if (string.IsNullOrEmpty (model.FindPattern))
+				return new SearchResult[0];
```
Therzok:
```suggestion
				return Array.Empty<SearchResult> ();
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/PatternSearcher.cs
mkrueger: ok weird API is a valid point :).

mkrueger:
I added an overload with just ReadOnlySpan.
 
The one with the indices however is useful and not going away usually I don't want to Find a pattern in the whole text and not just a subsequence.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
```diff
@@ -0,0 +1,247 @@
...
+			StartScopeTask (this, EventArgs.Empty);
+		}
+
+		void StartScopeTask (object sender, EventArgs e)
+		{
+			var scope = Scope.Create (model);
+			if (scope == null)
+				return;
+			currentScope = scope;
+			fileProviderTask = Task.Run (async delegate {
```
Therzok: No need for async await here, just return the task directly.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
```diff
@@ -0,0 +1,245 @@
...
+
+			if (model.ReplacePattern != null && !find.ValidatePattern (model, model.ReplacePattern)) {
+				MessageService.ShowError (GettextCatalog.GetString ("Replace pattern is invalid"));
+				return;
+			}
+			var cancelSource = new CancellationTokenSource ();
+			searchTokenSource = cancelSource;
+			var token = cancelSource.Token;
+
+			currentTask = Task.Run (async delegate {
```
Therzok: This could easily be:
```
currentTask = getFilesTask.ContinueWith (t => {
   ... all the rest of the code.
}, TaskScheduler.Default);
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
```diff
@@ -0,0 +1,245 @@
...
+				var files = await getFilesTask;
+				using (var searchMonitor = IdeApp.Workbench.ProgressMonitors.GetSearchProgressMonitor (true)) {
+
+					if (UpdateResultPad != null) {
+						Application.Invoke ((o, args) => {
```
Therzok: Rather than using Application.Invoke, prefer using `Runtime.RunInMainThread()`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
```diff
@@ -0,0 +1,245 @@
...
+							UpdateResultPad (searchMonitor.ResultPad);
+						});
+					}
+
+					if (UpdateStopButton != null) {
+						Application.Invoke ((o, args) => {
```
Therzok: `Runtime.RunInMainThread()`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
```diff
@@ -0,0 +1,245 @@
...
+							UpdateStopButton ();
+						});
+					}
+					var timer = DateTime.Now;
```
Therzok: Use a stopwatch for timings.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
```diff
@@ -0,0 +1,245 @@
...
+							foreach (var result in find.FindAll (model, files, searchMonitor, token)) {
+								if (token.IsCancellationRequested)
+									return;
+								results.Add (result);
+							}
+							searchMonitor.ReportResults (results);
+						} finally {
+							searchMonitor.EndTask ();
+						}
+					} catch (Exception ex) {
```
Therzok: Do we not want to log exceptions?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
```diff
@@ -0,0 +1,245 @@
...
+						string fileString = string.Format (GettextCatalog.GetPluralString ("in {0} file.", "in {0} files.", find.SearchedFilesCount), find.SearchedFilesCount);
+						message = GettextCatalog.GetString ("Search completed.") + Environment.NewLine + matches + " " + fileString;
+						searchMonitor.ReportSuccess (message);
+					}
+					if (message != null)
+						searchMonitor.ReportStatus (message);
+					searchMonitor.Log.WriteLine (GettextCatalog.GetString ("Search time: {0} seconds."), (DateTime.Now - timer).TotalSeconds);
+				}
+				if (UpdateStopButton != null) {
+					Application.Invoke ((o, args) => {
```
Therzok: Runtime.RunInMainThread

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
```diff
@@ -0,0 +1,245 @@
...
+					} catch (Exception ex) {
+						errorMessage = ex.Message;
+						LoggingService.LogError ("Error while search", ex);
+					}
+					string message = null;
+					if (errorMessage != null) {
+						message = GettextCatalog.GetString ("The search could not be finished: {0}", errorMessage);
+						searchMonitor.ReportError (message, null);
+					} else if (!searchMonitor.CancellationToken.IsCancellationRequested) {
+						string matches = string.Format (GettextCatalog.GetPluralString ("{0} match found", "{0} matches found", find.FoundMatchesCount), find.FoundMatchesCount);
```
Therzok: This won't provide enough context for translators. Maybe we should do the plural checks manually and have 2 different strings.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
```diff
@@ -0,0 +1,245 @@
...
+						string fileString = string.Format (GettextCatalog.GetPluralString ("in {0} file.", "in {0} files.", find.SearchedFilesCount), find.SearchedFilesCount);
+						message = GettextCatalog.GetString ("Search completed.") + Environment.NewLine + matches + " " + fileString;
+						searchMonitor.ReportSuccess (message);
+					}
+					if (message != null)
+						searchMonitor.ReportStatus (message);
+					searchMonitor.Log.WriteLine (GettextCatalog.GetString ("Search time: {0} seconds."), (DateTime.Now - timer).TotalSeconds);
```
Therzok:
```suggestion
					searchMonitor.Log.WriteLine (GettextCatalog.GetString ("Search time: {0} seconds.", (DateTime.Now - timer).TotalSeconds));
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
```diff
@@ -0,0 +1,245 @@
...
+					});
+				}
+			});
+		}
+
+		#region Static methods
+
+		static FindInFilesController currentFindInFilesController;
+		static FindInFilesSession find;
+		static CancellationTokenSource searchTokenSource = new CancellationTokenSource ();
```
Therzok: Why are these static? Shouldn't this be per-controller?

Therzok: If this is service code, maybe this should be in a separate class, from the controller.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
```diff
@@ -727,197 +712,35 @@ protected override void OnDestroyed ()
...
-				RegexSearch = checkbuttonRegexSearch.Active,
-				WholeWordsOnly = checkbuttonWholeWordsOnly.Active
-			};
-		}
-
-		static FindReplace find;
 		void HandleReplaceClicked (object sender, EventArgs e)
 		{
-			SearchReplace (comboboxentryFind.Entry.Text, comboboxentryReplace.Entry.Text ?? "", GetScope (), GetFilterOptions (), () => UpdateStopButton (), UpdateResultPad);
+			RequestFindAndReplace?.Invoke (this, EventArgs.Empty);
```
Therzok: Should we signal whether this is a find or a replace button click?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
```diff
@@ -1,363 +1,259 @@
...
+		{
+			IsRunning = false;
+		}
+
+		public bool ValidatePattern (FindInFilesModel filter, string pattern)
+		{
+			if (filter.RegexSearch) {
+				try {
+					_ = new Regex (pattern);
+					return true;
```
Therzok:
```suggestion
```

We're already returning true at the end of the method.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
```diff
@@ -1,363 +1,259 @@
...
+			}
+		}
+
+		public IEnumerable<SearchResult> FindAll (FindInFilesModel model, IReadOnlyList<FileProvider> fileList, ProgressMonitor monitor, CancellationToken token)
+		{
+			if (model.RegexSearch) {
+				RegexOptions regexOptions = RegexOptions.Compiled | RegexOptions.Multiline;
+				if (!model.CaseSensitive)
+					regexOptions |= RegexOptions.IgnoreCase;
+				regex = new Regex (model.FindPattern, regexOptions);
```
Therzok: We're not reusing the regex. So Compiled is actually going to hurt.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
```diff
@@ -1,363 +1,259 @@
...
+						MaxDegreeOfParallelism = Environment.ProcessorCount,
+						CancellationToken = token
+					};
+					Parallel.ForEach (contents, options, content => {
+						if (token.IsCancellationRequested)
+							return;
+						try {
+							Interlocked.Increment (ref searchedFilesCount);
+							if (model.InReplaceMode) {
+								content.Text = content.Reader.ReadToEnd ();
```
Therzok: Why do we need to get the whole file content?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
```diff
@@ -1,363 +1,259 @@
...
+			if (string.IsNullOrEmpty (model.FindPattern))
+				return Array.Empty<SearchResult> ();
+
+			if (model.RegexSearch)
+				return RegexSearch (model, monitor, provider);
+
+			return Search (model, provider);
+		}
+
+		SearchResult[] RegexSearch (FindInFilesModel model, ProgressMonitor monitor, FileProvider provider)
```
Therzok: Return an ImmutableArray and use an ImmutableArray Builder? Mostly as sanity, so we don't do defensive copies of the array.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
```diff
@@ -1,363 +1,259 @@
...
+		{
+			string content = IdeApp.Workbench.GetDocumentText (provider.FileName);
+			var results = new List<SearchResult> ();
+			if (!model.InReplaceMode) {
+				foreach (Match match in regex.Matches (content)) {
```
Therzok: Can we use:

var match = Regex.Match (content);
while (match.IsSuccess) {
    ... etc
    match = match.NextMatch ();
}
```

It's possible that regex.Matches does not eagerly do the matches, so in case of cancellation we do extra work.

File: main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs [hunk 24]
```diff
@@ -1,363 +1,259 @@
...
+					if (provider.SelectionStartPosition > -1 && match.Index < provider.SelectionStartPosition)
+						continue;
+					if (provider.SelectionEndPosition > -1 && match.Index + match.Length > provider.SelectionEndPosition)
+						continue;
+					if (!model.WholeWordsOnly || FindInFilesModel.IsWholeWordAt(content, match.Index, match.Length))
+						results.Add(new SearchResult(provider, match.Index, match.Length));
+				}
+			} else {
+				var matches = new List<Match> ();
+				foreach (Match match in regex.Matches(content))
```
Therzok: Same here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
```diff
@@ -1,363 +1,259 @@
...
+				provider.EndReplace ();
+			}
+			return results.ToArray ();
+		}
+
+		SearchResult[] Search (FindInFilesModel model, FileProvider provider)
+		{
+			string content = IdeApp.Workbench != null ? IdeApp.Workbench.GetDocumentText (provider.FileName) : File.ReadAllText (provider.FileName);
+			var findResults = model.PatternSearcher.FindAll (content);
+			var result = new SearchResult [findResults.Length];
```
Therzok: Why do we do a copy here?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/PatternSearcher.cs
```diff
@@ -0,0 +1,145 @@
...
+
+namespace MonoDevelop.Ide.FindInFiles
+{
+	class PatternSearcher
+	{
+		readonly char [] patternArray;
+		readonly int patternLength;
+
+		const int ShiftTableLength = 128;
+		readonly int [] shiftTable;
```
Therzok: Prefer ImmutableArray here to avoid modifying shiftTable.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/PatternSearcher.cs
```diff
@@ -0,0 +1,145 @@
...
+			}
+			return result;
+		}
+
+		public int [] FindAll (string text)
+		{
+			return FindAll (text.AsSpan ());
+		}
+
+		public int [] FindAll (ReadOnlySpan<char> text)
```
Therzok: Maybe return ImmutableArray here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/PatternSearcher.cs
```diff
@@ -0,0 +1,145 @@
...
+		{
+			var result = new List<int> ();
+			int index = 0;
+			int end = text.Length;
+			while ((index = Find (text, index, end)) > -1) {
```
Therzok: Can we add a test for this? I think this will return too many results for a scenario where we want to search for "ab" in "aa ab"

Therzok: In example, I think we should be setting index to the index where we found a match + 1.

Therzok: Ah, nevermind, misunderstood the implementation.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.AllOpenFilesScope.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Text;
+using System.Threading;
+
+namespace MonoDevelop.Ide.FindInFiles
+{
+
+	abstract partial class Scope
+	{
+		class AllOpenFilesScope : Scope
```
Therzok:
```suggestion
		sealed class AllOpenFilesScope : Scope
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.AllOpenFilesScope.cs
```diff
@@ -0,0 +1,68 @@
...
+							continue;
+						results.Add (new FileProvider (document.FileName, document.Owner as Project));
+					}
+				}
+				return Task.FromResult ((IReadOnlyList<FileProvider>)results);
+			}
+
+			public override string GetDescription (FindInFilesModel model)
+			{
+				if (!model.InReplaceMode)
```
Therzok: Invert this condition:
```csharp
return model.InReplaceMode
    ? GettextCatalog.GetString ("Replacing '{0}' in all open documents", model.FindPattern)
    : GettextCatalog.GetString ("Looking for '{0}' in all open documents", model.FindPattern);
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.DirectoryScope.cs
```diff
@@ -0,0 +1,144 @@
...
+		class DirectoryScope : Scope
+		{
+			readonly string path;
+			readonly bool recurse;
+
+			public override PathMode PathMode {
+				get { return PathMode.Absolute; }
+			}
+
+			[Obsolete ("Unused - will be removed")]
```
Therzok: Also add EditorBrowsable(Advanced)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.DirectoryScope.cs
```diff
@@ -0,0 +1,144 @@
...
+						yield break;
+					try {
+						var readPermission = new FileIOPermission (FileIOPermissionAccess.Read, curPath);
+						readPermission.Demand ();
+					} catch (Exception e) {
+						LoggingService.LogError ("Can't access path " + curPath, e);
+						yield break;
+					}
+
+					foreach (string fileName in Directory.EnumerateFiles (curPath, "*")) {
```
Therzok:
```suggestion
					foreach (string fileName in Directory.EnumerateFiles (curPath)) {
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.DocumentScope.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Text;
+using System.Threading;
+
+namespace MonoDevelop.Ide.FindInFiles
+{
+
+	abstract partial class Scope
+	{
+		class DocumentScope : Scope
```
Therzok: sealed

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.DirectoryScope.cs
```diff
@@ -0,0 +1,144 @@
...
+
+			public override Task<IReadOnlyList<FileProvider>> GetFilesAsync (FindInFilesModel filterOptions, CancellationToken cancellationToken)
+			{
+				EnsureFileNamesLoaded (filterOptions);
+				return Task.FromResult<IReadOnlyList<FileProvider>> (fileNames);
+			}
+
+			public override string GetDescription (FindInFilesModel model)
+			{
+				if (!model.InReplaceMode)
```
Therzok: Same, invert condition.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.DocumentScope.cs
```diff
@@ -0,0 +1,63 @@
...
+				var doc = IdeApp.Workbench.ActiveDocument;
+				var textBuffer = doc.GetContent<ITextBuffer> ();
+				if (textBuffer != null)
+					return Task.FromResult<IReadOnlyList<FileProvider>> (new [] { new OpenFileProvider (textBuffer, doc.Owner as Project, doc.FileName) });
+				return EmptyFileProviderTask;
+			}
+
+			public override string GetDescription (FindInFilesModel model)
+			{
+				if (!model.InReplaceMode)
```
Therzok: Invert condition here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.DocumentScope.cs
```diff
@@ -0,0 +1,63 @@
...
+			public override PathMode PathMode {
+				get { return PathMode.Hidden; }
+			}
+
+			public override Task<IReadOnlyList<FileProvider>> GetFilesAsync (FindInFilesModel filterOptions, CancellationToken cancellationToken = default)
+			{
+				var doc = IdeApp.Workbench.ActiveDocument;
+				var textBuffer = doc.GetContent<ITextBuffer> ();
+				if (textBuffer != null)
+					return Task.FromResult<IReadOnlyList<FileProvider>> (new [] { new OpenFileProvider (textBuffer, doc.Owner as Project, doc.FileName) });
```
Therzok:
```csharp
					return Task.FromResult<IReadOnlyList<FileProvider>> (ImmutableArray.Create (new OpenFileProvider (textBuffer, doc.Owner as Project, doc.FileName)));
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.SelectionScope.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Text.Editor;
+using System.Threading;
+
+namespace MonoDevelop.Ide.FindInFiles
+{
+
+	abstract partial class Scope
+	{
+		class SelectionScope : Scope
```
Therzok:
```suggestion
		sealed class SelectionScope : Scope
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.WholeProjectScope.cs
```diff
@@ -0,0 +1,82 @@
...
+using MonoDevelop.Core;
+using System.Threading.Tasks;
+using System.Threading;
+
+namespace MonoDevelop.Ide.FindInFiles
+{
+
+	abstract partial class Scope
+	{
+		class WholeProjectScope : Scope
```
Therzok:
```suggestion
		sealed class WholeProjectScope : Scope
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.WholeProjectScope.cs
```diff
@@ -0,0 +1,82 @@
...
+							continue;
+						results.Add (new FileProvider (file.Name, project));
+					}
+				}
+				return results;
+			}
+
+			public override string GetDescription (FindInFilesModel model)
+			{
+				if (!model.InReplaceMode)
```
Therzok: invert if

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.WholeProjectScope.cs
```diff
@@ -0,0 +1,82 @@
...
+					foreach (ProjectFile file in await project.GetSourceFilesAsync (conf)) {
+						if ((file.Flags & ProjectItemFlags.Hidden) == ProjectItemFlags.Hidden)
+							continue;
+						if (!filterOptions.IncludeCodeBehind && file.Subtype == Subtype.Designer)
+							continue;
+						if (!filterOptions.IsFileNameMatching (file.Name))
+							continue;
+						if (!IdeServices.DesktopService.GetFileIsText (file.Name))
+							continue;
+						if (!alreadyVisited.Add (file.FilePath.FullPath))
```
Therzok: Why do we need to do full path here?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.WholeSolutionScope.cs
```diff
@@ -0,0 +1,121 @@
...
+using MonoDevelop.Core;
+using System.Threading.Tasks;
+using System.Threading;
+
+namespace MonoDevelop.Ide.FindInFiles
+{
+
+	abstract partial class Scope
+	{
+		class WholeSolutionScope : Scope
```
Therzok:
```suggestion
		sealed class WholeSolutionScope : Scope
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.WholeSolutionScope.cs
```diff
@@ -0,0 +1,121 @@
...
+
+				var options = new ParallelOptions ();
+				options.MaxDegreeOfParallelism = 4;
+
+				Parallel.ForEach (IdeApp.Workspace.GetAllSolutionItems ().OfType<SolutionFolder> (),
+								  options,
+								  () => new List<FileProvider> (),
+								  (folder, loop, providers) => {
+									  foreach (var file in folder.Files.Where (f => filterOptions.IsFileNameMatching (f.FileName) && File.Exists (f.FullPath))) {
+										  if (!IdeServices.DesktopService.GetFileIsText (file.FullPath))
```
Therzok: please call FullPath only once.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.WholeSolutionScope.cs
```diff
@@ -0,0 +1,121 @@
...
+				}
+
+				var alreadyVisited = new HashSet<string> ();
+				var results = new List<FileProvider> ();
+
+				var options = new ParallelOptions ();
+				options.MaxDegreeOfParallelism = 4;
+
+				Parallel.ForEach (IdeApp.Workspace.GetAllSolutionItems ().OfType<SolutionFolder> (),
+								  options,
```
Therzok: way too much indent.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.WholeSolutionScope.cs
```diff
@@ -0,0 +1,121 @@
...
+										  results.AddRange (providers);
+									  }
+								  });
+
+				return Task.FromResult ((IReadOnlyList<FileProvider>)results);
+			}
+
+			public override string GetDescription (FindInFilesModel model)
+			{
+				if (!model.InReplaceMode)
```
Therzok: Invert if.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
```diff
@@ -250,6 +251,34 @@ public Document GetDocument (FilePath filePath)
...
 
+		/// <summary>
+		/// Gets the document text as string. Either from an open document or a file on disk.
+		/// </summary>
+		internal string GetDocumentText (string fileName, bool readBinaryFiles = false)
+		{
+			if (fileName == null)
+				throw new ArgumentNullException (nameof (fileName));
+
+			foreach (var doc in documentManager.Documents) {
```
Therzok: DocumentManager has a method to get a Document by FilePath.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
mkrueger: This PR is not about the regex part.

Regexes can be addressed in another PR. This PR already took too long to merge for just the "search" part. Concentrate on that.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/PatternSearcher.cs
mkrueger: Returning complexer objects does not improve memory or speed in any way.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
mkrueger: Yes it does - but as said this PR is not about the replace part.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
mkrueger: This PR doesn't touch replace.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
mkrueger: This PR doesn't touch replace.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
mkrueger: This PR is not about the regex part - that should be done seperately.

Mixing replace & regexes in here would only delay the merge even more.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
mkrueger: ImmutableArray is not faster or more memory efficient than arrays. Array is the best thing to use.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
mkrueger: same here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/PatternSearcher.cs
mkrueger: Table is not modified - if someone does it's his fault.
ImmutableArray is only less performant - this PR is about increasing, not decreasing search performance.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/PatternSearcher.cs
mkrueger: There are tests for the pattern searcher in this PR btw.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
mkrueger: A copy of what ? 

Pattern searcher gives back an array of indices. SearchResults are more complex the indices need to be converted. However the PatternSearcher could create SearchResult instead.
The PatternSearcher is atm independent of the search engine and could be reused.

But I'll think about changing that. Maybe I'll do some speed tests if it's worth it.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.AllOpenFilesScope.cs
mkrueger: changed - in the other scopes as well.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
Therzok: It's not about performance. It's about contract.

Since this PR remodels the API, we could definitely go for correctness, and have an ImmutableArray to ensure the immutability of the data.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.AllOpenFilesScope.cs
mkrueger: fixed - in other scopes as well.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.DirectoryScope.cs
mkrueger: fixed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.DocumentScope.cs
mkrueger: done. Here it is no performance problem - but doesn't really help to improve the code much.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.WholeSolutionScope.cs
mkrueger: fixed - was a very old indentation engine - didn't correct it in that PR.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.WholeProjectScope.cs
mkrueger: file.Name is enough - fixed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
mkrueger: fixed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.WholeSolutionScope.cs
```diff
@@ -0,0 +1,121 @@
...
+									  }
+									  return providers;
+								  },
+								  (providers) => {
+									  lock (results) {
+										  results.AddRange (providers);
+									  }
+								  });
+
+				Parallel.ForEach (IdeApp.Workspace.GetAllProjects (),
```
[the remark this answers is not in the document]
mkrueger: This doesn't work anymore at all - SolutionFolders are not in that list anymore.
This code is broken - I'll replace it with a single parallel.ForEach.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
```diff
@@ -184,18 +185,18 @@ public IEnumerable<SearchResult> FindAll (FindInFilesModel model, IReadOnlyList<
...
 				return RegexSearch (model, monitor, provider);
 
 			return Search (model, provider);
 		}
 
-		SearchResult[] RegexSearch (FindInFilesModel model, ProgressMonitor monitor, FileProvider provider)
+		ImmutableArray<SearchResult> RegexSearch (FindInFilesModel model, ProgressMonitor monitor, FileProvider provider)
 		{
 			string content = IdeApp.Workbench.GetDocumentText (provider.FileName);
 			var results = new List<SearchResult> ();
```
Therzok: Use ImmutableArray.Builder here - it's optimized for this scenario.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
```diff
@@ -233,21 +234,21 @@ SearchResult[] RegexSearch (FindInFilesModel model, ProgressMonitor monitor, Fil
...
-			return results.ToArray ();
+			return results.ToImmutableArray ();
 		}
 
-		SearchResult[] Search (FindInFilesModel model, FileProvider provider)
+		ImmutableArray<SearchResult> Search (FindInFilesModel model, FileProvider provider)
 		{
 			string content = IdeApp.Workbench != null ? IdeApp.Workbench.GetDocumentText (provider.FileName) : File.ReadAllText (provider.FileName);
 			var findResults = model.PatternSearcher.FindAll (content);
 			var result = new SearchResult [findResults.Length];
```
Therzok: ImmutableArray.Builder here too.

Therzok: And since we know the length of the array, we can then use `MoveToImmutable()` to avoid array copies.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.DirectoryScope.cs
```diff
@@ -54,14 +55,15 @@ sealed class DirectoryScope : Scope
...
+			ImmutableArray<FileProvider> fileNames;
 
 			private void EnsureFileNamesLoaded (FindInFilesModel filterOptions)
 			{
-				if (fileNames != null)
+				if (fileNamesLoaded)
 					return;
-				fileNames = GetFileNames (filterOptions).Select (file => new FileProvider (file)).ToArray ();
+				fileNamesLoaded = true;
+				fileNames = ImmutableArray.Create (GetFileNames (filterOptions).Select (file => new FileProvider (file)).ToArray ());
```
Therzok: No need to `ToArray()`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.SelectionScope.cs
```diff
@@ -43,13 +44,13 @@ sealed class SelectionScope : Scope
...
 
-			public override Task<IReadOnlyList<FileProvider>> GetFilesAsync (FindInFilesModel filterOptions, CancellationToken cancellationToken = default)
+			public override Task<ImmutableArray<FileProvider>> GetFilesAsync (FindInFilesModel filterOptions, CancellationToken cancellationToken = default)
 			{
 				var doc = IdeApp.Workbench.ActiveDocument;
 				var textView = doc.GetContent<ITextView> (true);
 				if (textView != null) {
 					var selection = textView.Selection.SelectedSpans.FirstOrDefault ();
-					return Task.FromResult<IReadOnlyList<FileProvider>> (new [] { new OpenFileProvider (textView.TextBuffer, doc.Owner as Project, doc.FileName, selection.Start, selection.End) });
+					return Task.FromResult(ImmutableArray<FileProvider>.Empty.Add (new OpenFileProvider (textView.TextBuffer, doc.Owner as Project, doc.FileName, selection.Start, selection.End)));
```
Therzok: `ImmutableArray.Create ((FileProvider)new OpenFileProvider (textView.TextBuffer, doc.Owner as Project, doc.FileName, selection.Start, selection.End));`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesSession.cs
mkrueger: fixed - wanted to do that differently.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.SelectionScope.cs
mkrueger: fixed.

slluis:
It would be nice to have performance regression tests for this. @Therzok could you please advice about how to implement those?

Therzok:
We want a benchmark test, not an integration perf test, right?

If so, then the method to do so is:
* Add a reference to MonoDevelop.PerformanceTesting
* Attribute the test you want benchmark with `[Benchmark]`(https://github.com/mono/monodevelop/blob/560a020231b345761fadfa63f99ddd8a64d6c1ad/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.PerfTests/SteppingTests.cs#L50)
* At the end of the test, do `Benchmark.SetTime ()`
* Add a new step in CI that would do: `mdtool run-md-tests --performance MonoDevelop.Ide.Tests.dll --xml=TestResult_Perf_MonoDevelop.Ide.Tests.dll.xml`

That should get you setup with a new perf test

Therzok:
@manish if you can provide guidance for new perftest format, it would be better.

mkrueger:
I'll continue later on that issue.

mkrueger:
Reopened, thought that I can optimize it even more without a lookup table, tried instead to hard wire the pattern.

But that didn't make it faster to my surprise.

Therzok:
There's a few remaining comments, if you check the diff page. Could you please take a look, @mkrueger ?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.WholeSolutionScope.cs
```diff
@@ -0,0 +1,110 @@
...
+					Parallel.ForEach (solution.GetAllItems<SolutionFolderItem> (),
+						options,
+						() => new List<FileProvider> (),
+						(item, loop, providers) => {
+							if (item is SolutionFolder folder) {
+								foreach (var file in folder.Files.Where (f => filterOptions.IsFileNameMatching (f.FileName) && File.Exists (f.FileName))) {
+									if (!IdeServices.DesktopService.GetFileIsText (file.FileName))
+										continue;
+									lock (alreadyVisited) {
+										if (alreadyVisited.Contains (file.FileName))
```
Therzok:
```csharp
if (!alreadyVisited.Add (file.FileName))
	continue;
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.WholeSolutionScope.cs
```diff
@@ -0,0 +1,110 @@
...
+											continue;
+										alreadyVisited.Add (file.FileName);
+									}
+									providers.Add (new FileProvider (file.FileName));
+								}
+							}
+
+							if (item is Project project) {
+								var conf = project.DefaultConfiguration?.Selector;
+								foreach (var file in project.GetSourceFilesAsync (conf).Result) {
```
Therzok:
```suggestion
								foreach (var file in await project.GetSourceFilesAsync (conf)) {
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.SelectionScope.cs
```diff
@@ -0,0 +1,67 @@
...
+				get { return PathMode.Hidden; }
+			}
+
+			public override Task<ImmutableArray<FileProvider>> GetFilesAsync (FindInFilesModel filterOptions, CancellationToken cancellationToken = default)
+			{
+				var doc = IdeApp.Workbench.ActiveDocument;
+				var textView = doc.GetContent<ITextView> (true);
+				if (textView != null) {
+					var selection = textView.Selection.SelectedSpans.FirstOrDefault ();
+					return Task.FromResult(ImmutableArray.Create<FileProvider> (new OpenFileProvider (textView.TextBuffer, doc.Owner as Project, doc.FileName, selection.Start, selection.End)));
```
Therzok: This can NRE on `selection`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.SelectionScope.cs
```diff
@@ -0,0 +1,67 @@
...
+			public override PathMode PathMode {
+				get { return PathMode.Hidden; }
+			}
+
+			public override Task<ImmutableArray<FileProvider>> GetFilesAsync (FindInFilesModel filterOptions, CancellationToken cancellationToken = default)
+			{
+				var doc = IdeApp.Workbench.ActiveDocument;
+				var textView = doc.GetContent<ITextView> (true);
+				if (textView != null) {
+					var selection = textView.Selection.SelectedSpans.FirstOrDefault ();
```
Therzok: Should we just search in the whole document if nothing is selected?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
```diff
@@ -201,15 +179,19 @@ public static string FormatPatternToSelectionOption (string pattern, bool regex)
 
 			InitFromProperties ();
 			
-			if (showReplace)
+			if (model.InReplaceMode)
 				toggleReplaceInFiles.Toggle ();
 			else
 				toggleFindInFiles.Toggle ();
+			comboboxentryFind.Entry.Text = model.FindPattern;
+			comboboxentryFind.Entry.Changed += delegate {
```
Therzok: Possible ref cycle

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
```diff
@@ -221,17 +203,20 @@ public static string FormatPatternToSelectionOption (string pattern, bool regex)
...
 			comboboxentryFind.Entry.SelectRegion (0, comboboxentryFind.ActiveText.Length);
 			comboboxentryFind.GrabFocus ();
+
 			DeleteEvent += delegate { Destroy (); };
 			UpdateStopButton ();
 			UpdateSensitivity ();
 			if (!buttonSearch.Sensitive) {
 				comboboxScope.Active = (int)SearchScope.Directories;
 			}
+			searchentryFileMask.Entry.Changed += delegate {
```
Therzok: possible RefCycle:
```
dialog -> entry (via UI hierarchy)
entry -> dialog (via this capture)
```

Need to remove event handler on destroy

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
```diff
@@ -386,17 +373,19 @@ void ShowReplaceUI ()
 			AddActionWidget (buttonReplace, 0);
 			buttonReplace.GrabDefault ();
 			
-			replaceMode = true;
-			
 			Requisition req = SizeRequest ();
 			Resize (req.Width, req.Height);
+			comboboxentryReplace.Entry.Changed += delegate {
```
Therzok: Possible ref cycle as above.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
```diff
@@ -432,6 +419,10 @@ void ShowDirectoryPathUI ()
 			hboxPath = new HBox ();
 			comboboxentryPath = new ComboBoxEntry ();
 			comboboxentryPath.Destroyed += ComboboxentryPathDestroyed;
+			comboboxentryPath.Entry.Changed += delegate {
```
Therzok: Ref cycle

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
```diff
@@ -0,0 +1,246 @@
...
+
+		void Dispose ()
+		{
+			currentFindDialog?.Destroy ();
+			currentFindDialog = null;
+		}
+
+		FindInFilesModel model;
+		Scope currentScope;
+		Task<ImmutableArray<FileProvider>> fileProviderTask;
```
Therzok: Why do we hold onto this task?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.SelectionScope.cs
mkrueger: Not part of this PR.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.WholeSolutionScope.cs
mkrueger: me neither. But currently that .Result doesn't lock up. I'll plan to rework find in files that should've been only the 1st step. But took months to merge…

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
mkrueger: Both refs get collected they should be GCed once dialog is collected.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
mkrueger: Same here

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
mkrueger: same here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs
mkrueger: same here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.WholeSolutionScope.cs
mkrueger: These days I do that - must've been older code :).

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.SelectionScope.cs
mkrueger: Got introduced in the new editor changes. /fixed

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
mkrueger: old code.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
mkrueger: old code.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
mkrueger: old code.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
mkrueger: old code.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesController.cs
mkrueger: old code.

btw. why doesn't that work for translators ?  In general the plural string is helpful because there are languages that have more then 1 and "many" ,like 1,2,3, many etc. However we don't support that yet.

main/tests/Ide.Tests/MonoDevelop.Ide.FindInFiles/FindInFilesModelTests.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Threading.Tasks;
+using System.Linq;
+
+namespace MonoDevelop.Ide.FindInFiles
+{
+	[TestFixture]
+	public class FindInFilesModelTests : IdeTestBase
+	{
+		[Test]
+		public void TestIsFileNameMatching ()
```
Therzok: Please add tests for all the public APIs of FindInFilesModel. Just 2 tests are not enough to ensure it's functional.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesModel.cs
```diff
@@ -0,0 +1,143 @@
...
+
+		public string FileMask {
+			get {
+				return file_mask;
+			}
+			set {
+				file_mask = value;
+				split_file_masks = file_mask?.Split (separators, StringSplitOptions.RemoveEmptyEntries);
+
+				evaluator = FileNameEvaluator.CreateFileNameEvaluator (split_file_masks);
```
Therzok: What I mean by testing all APIs, setting `FileMask` to `null` will NRE in this call

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesModel.cs
```diff
@@ -0,0 +1,143 @@
...
+
+		string findPattern;
+		public string FindPattern { get => findPattern; set { patternSearcher = null; findPattern = value; } }
+
+		public string ReplacePattern { get; set; }
+
+		PatternSearcher patternSearcher;
+		public PatternSearcher PatternSearcher {
+			get {
+				return patternSearcher ?? (patternSearcher = new PatternSearcher (FindPattern, CaseSensitive, WholeWordsOnly));
```
Therzok: If FindPattern is null, this will throw in the constructor.

mkrueger: FindPattern needs to be != null.

main/tests/Ide.Tests/MonoDevelop.Ide.FindInFiles/FindInFilesModelTests.cs
mkrueger: It's 2 tests more than what we had. It's old code - not new one btw.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesModel.cs
Therzok: Nothing enforces that.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesModel.cs
mkrueger: == null case can't happen from what I see at that point - but y the old code could handle that /fixed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesModel.cs
mkrueger: The exception does. Other than that I fixed that in that case. doesn't make sense to be null but it'll work.

mhutch:
this deadlocked on cmd-shift-f

https://gist.github.com/mhutch/e9389894a04be0ae889f844b50e2ba16

mkrueger:
> this deadlocked on cmd-shift-f
> 
> https://gist.github.com/mhutch/e9389894a04be0ae889f844b50e2ba16

From that it's not related - seems to stuck up in Microsoft.CodeAnalysis.SQLite - that's a different issue. However the find in files queries the project system that may trigger roslyn for creating code behind files. 

But that's the case in the current implementation as well that's not touched and unrelated to these changes. I'll rebase that to master. See if it happens again.

A use case for that would've been helpful.

mkrueger:
moved to another repository.

PR #7548, opened 2019-05-10
Fix focus issue with navigation

slluis:
Don't change the first responder of the document being focused if it is
already a child of the main view.

Fixes VSTS #847834 - Go To Declaration causes editor to lose focus


slluis:

PR #7553, opened 2019-05-11
[Ide] GLib logging now uses loginternalerror so we can track glib exc…

Therzok:
…eptions

Fixes VSTS #867387 - [Telemetry] Objective-C based crashes should log exception information to telemetry

cc @leculver

### DIFF
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -272,16 +272,10 @@ namespace MonoDevelop.MacIntegration
 		[DllImport (FoundationLib)]
 		static extern NSUncaughtExceptionHandler NSGetUncaughtExceptionHandler ();
 
-		internal static void RegisterUncaughtExceptionHandler ()
+		static void RegisterUncaughtExceptionHandler ()
 		{
 			oldHandler = NSGetUncaughtExceptionHandler ();
 			NSSetUncaughtExceptionHandler (uncaughtHandler);
-		}
-
-		internal static void UnregisterUncaughtExceptionHandler ()
-		{
-			NSSetUncaughtExceptionHandler (oldHandler);
-			oldHandler = null;
 		}
 
 		public override Xwt.Toolkit LoadNativeToolkit ()
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -235,6 +235,55 @@ namespace MonoDevelop.MacIntegration
 			}
 		}
 
+		const string FoundationLib = "/System/Library/Frameworks/Foundation.framework/Versions/Current/Foundation";
+
+		delegate void NSUncaughtExceptionHandler (IntPtr exception);
+
+		static readonly NSUncaughtExceptionHandler uncaughtHandler = HandleUncaughtException;
+		static NSUncaughtExceptionHandler oldHandler;
+
+		static void HandleUncaughtException(IntPtr exceptionPtr)
+		{
+			// It's non-null guaranteed by objc.
+			Debug.Assert (exceptionPtr != IntPtr.Zero);
+
+			var nsException = ObjCRuntime.Runtime.GetNSObject<NSException> (exceptionPtr);
+			var exception = new MarshalledObjCException (nsException, new StackTrace(true).ToString ());
+			LoggingService.LogInternalError (exception.Message, exception);
+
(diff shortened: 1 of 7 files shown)


main/tests/Ide.Tests/MonoDevelop.Ide.Gui/GLibLoggingTests.cs
```diff
@@ -0,0 +1,62 @@
...
+
+				LoggingService.RegisterCrashReporter (crashReporter);
+
+				GLib.Log.Write ("Gtk", GLib.LogLevelFlags.Warning, "{0}", "should not be captured");
+				Assert.IsNull (crashReporter.LastException);
+
+				GLib.Log.Write ("Gtk", GLib.LogLevelFlags.Critical, "{0}", "critical should be captured");
+				Assert.That (crashReporter.LastException.Message, Contains.Substring ("critical should be captured"));
+
+				// Error will cause the application to exit, so we can't test for that, but it follows the same code as Critical.
```
Therzok: There is no way to set error as non-fatal:
https://github.com/GNOME/glib/blob/4326d5e27db954b6883b2cc77b1cf6408cdf76e2/glib/gmessages.c#L772-L773

slluis:
Is here any way to know if any of those exceptions are going to bring down the IDE?

Therzok:
Yes. We can check whether the logging flags contain Fatal.

Therzok:
I'll fix that up

Therzok:
@slluis pushed a fix with fatal/non-fatal behaviour.

Therzok:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/GLibLogging.cs
```diff
@@ -268,13 +270,32 @@ static void LoggerMethod (IntPtr logDomainPtr, LogLevelFlags logLevel, IntPtr me
 			case LogLevelFlags.Error:
 			case LogLevelFlags.Critical:
 			default:
-				LoggingService.LogError (msg);
+				var exception = new CriticalGtkException (message, stackTraceString);
+				if (logLevel.HasFlag (LogLevelFlags.FlagFatal))
+					LoggingService.LogFatalError ("Fatal GLib error", exception);
```
leculver: Why pass a static string to either of these methods?  I would have expected:

```suggestion
					LoggingService.LogFatalError (msg, exception);
				else
					LoggingService.LogInternalError (msg, exception);
```

This is more in line with the one for obj-c in the previous file.  I worry that setting the string here will end up putting the wrong message in telemetry, but I'm not 100% sure how all of that plumbs through.

Therzok: The static message is for our log files only:
http://source.monodevelop.com/#MonoDevelop.Core/MonoDevelop.Core/LoggingService.cs,572

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -235,6 +235,49 @@ static void CheckGtkVersion (uint major, uint minor, uint micro)
...
+		static NSUncaughtExceptionHandler oldHandler;
+
+		static void HandleUncaughtException(IntPtr exceptionPtr)
+		{
+			// It's non-null guaranteed by objc.
+			Debug.Assert (exceptionPtr != IntPtr.Zero);
+
+			var nsException = ObjCRuntime.Runtime.GetNSObject<NSException> (exceptionPtr);
+			var exception = new MarshalledObjCException (nsException, new StackTrace(true).ToString ());
+			LoggingService.LogInternalError (exception.Message, exception);
```
leculver: One remaining question.  This creates a nested exception, right?  (I don't know what ObjCException does behind the scenes offhand.)  Does `exception` here contain the same message from `nsException`?  If so this looks good to me.  If not there might be a slight issue here because VS telemetry seems to ignore the nested parts of exceptions...from what I can tell.

Therzok: NSException does not subclass Exception. I should've used the same pattern as in the glib logger to log a static message. Here is [ObjCException](https://github.com/xamarin/xamarin-macios/blob/master/src/Foundation/ObjCException.cs#L29-L65)

leculver: Ahh looking at the code, this looks fine too.

Therzok:
Adding @rolfbjarne for review. Does the ObjC exception handling look okay to you?

PR #7557, opened 2019-05-13
[Ide] Handle NuGet restore cancelled whilst type system frozen

mrward:
If a NuGet restore is cancelled then the type system may stop
processing if it is frozen waiting for the restore to finish since a
TaskCanceledException would be thrown.

To avoid this the TypeSystemService.FreezeTask is used inside a
try/catch so any exceptions can be ignored. Whether the NuGet restore
is cancelled or fails should not stop the type system from getting
project information.

Fixes VSTS #891329 - Handle cancelling NuGet restore whilst type
system is waiting for restore

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService.cs
@@ -774,6 +774,17 @@ namespace MonoDevelop.Ide.TypeSystem
 
 		public static Func<Task> FreezeLoad = () => Task.CompletedTask;
 
+		internal static async Task SafeFreezeLoad ()
+		{
+			try {
+				await FreezeLoad ();
+			} catch (Exception) {
+				// Ignore exceptions, such as the task being cancelled. We want to freeze the load
+				// whilst the NuGet restore is being run and continue after it has finished, cancelled,
+				// or thrown an exception.
+			}
+		}
+
 		public Task ProcessPendingLoadOperations ()
 		{
 			lock (workspaceLoadLock) {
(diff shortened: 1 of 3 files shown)


mrward:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService.cs
```diff
@@ -774,6 +774,17 @@ public MonoDevelop.Projects.Project GetMonoProject (Microsoft.CodeAnalysis.Docum
 
 		public static Func<Task> FreezeLoad = () => Task.CompletedTask;
 
+		internal static async Task SafeFreezeLoad ()
+		{
+			try {
+				await FreezeLoad ();
+			} catch (Exception) {
```
Therzok: We might want to LogInternalError on Exception and swallow just OperationCanceledException

mrward: Not sure it is worth it. Any non-cancel exception here would be caught by the NuGet addin.

Therzok: Then we should filter OperationCanceledException only. General catch exception is something we want to eliminate.

mrward: If this freeze NuGet restore task throws an exception (not sure that it will) then it would stop the type system from continuing if there is no general catch exception. The NuGet addin will log exceptions itself but the type system would still receive the task's non-cancelled exception. The type system does not care about what happens when the freeze task finishes (either successfully or is cancelled or an error occurs) - it is just used to prevent it from running when the freeze task is active. So I am not sure we should remove the general catch exception here.

Therzok: Fine by me then.

PR #7559, opened 2019-05-13
[Ide] Generate valid project name from app name

mrward:
The New Project dialog has extra rules for the project name and
solution name on the final page of the dialog, such as not allowing
spaces. The generation of a valid project and solution name from the
app name specified in a project template wizard used different logic
so the names shown on the final page may still be considered invalid
on the final page. For example AppName='Hello, Mac' in the Cocoa App
project templates generates a project and solution name of 'Hello,Mac'
which the New Project dialog considers invalid.

To prevent this the same logic is now used when generating a valid
project and solution name from the project template wizard's app name.

Fixes VSTS #825164 - VS4M partially renames the project but it's still
invalid

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectConfiguration.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectConfiguration.cs
@@ -72,12 +72,12 @@ namespace MonoDevelop.Ide.Projects
 
 		public string GetValidProjectName ()
 		{
-			return GetValidDir (ProjectName);
+			return GenerateValidProjectName (ProjectName);
 		}
 
 		public string GetValidSolutionName ()
 		{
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectConfiguration.cs
```diff
@@ -125,6 +125,8 @@ static string GetValidDir (string name)
 					continue;
```
Therzok: We should probably cache `GetInvalidPathChars()`, it's creating a copy on every call.

mrward: Yeah I did think the same thing when I looked at the code. The GetValidDir code was copied from the original New Project dialog and I did not change it.

PR #7562, opened 2019-05-13
[8.1][Ide] GLib logging now uses loginternalerror so we can track glib exceptions

Therzok:
Fixes VSTS #867387 - [Telemetry] Objective-C based crashes should log exception information to telemetry

cc @leculver


Therzok:
Hmmm, the PR is still messed up. I'll fix.

@slluis Lee said we might want this in for 8.1 P2.

leculver:
@Therzok If possible, yes.  p2 would be the best route so we can try to fix some of these problems before 8.1 RTM.  If it's not feasible in that timeframe or would otherwise risk getting preview 2 out the door, then we can live without it.  Whichever is best for the product.

Therzok:
@slluis can you merge?

Therzok:
Cherry-picked the other commits from https://github.com/mono/monodevelop/pull/7598

slluis (approved):

PR #7576, opened 2019-05-15
MSBuild uses better GC configuration.

Therzok:


Therzok:
Results are inconclusive. Not sure if there's any big gain - but test applications might be small enough to not be affected by this change.

Therzok:
Maybe this is why?
https://github.com/mono/mono/blob/c5b88ec4f323f2bdb7c7d0a595ece28dae66579c/scripts/xbuild.in#L2

Therzok:
```
➜  main git:(sample-profiler) cat /Library/Frameworks/Mono.framework/Versions/Current/Commands/msbuild
#!/bin/sh
MONO_GC_PARAMS="nursery-size=64m,$MONO_GC_PARAMS" exec /Library/Frameworks/Mono.framework/Versions/5.18.1/bin/mono $MONO_OPTIONS /Library/Frameworks/Mono.framework/Versions/5.18.1/lib/mono/msbuild/15.0/bin/MSBuild.dll "$@"
```

YEP. Only the msbuild builder should be affected.

Therzok:
Gains of a few seconds across test runs, it should be good to go.

Therzok:
DotNetCore tests are behaving differently, but not sure what's causing that.

Therzok:
Overall, actual test times improved - downloading the drop is taking longer in this branch than master.

slluis (approved):

mrward (approved):

PR #7581, opened 2019-05-16
Fixes VSTS Bug 893674: [Feedback] Goto line jumps to wrong line or

mkrueger:
crashes

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/893674

Bug got introduced by the new editor changes. Pattern.Column == 0 was
the problem so the SnapshotPoint got negative and that explains the
line -1 problem as well.


mkrueger:

slluis:
Does it apply to 8.0 too?

mkrueger:
Should be but it's a small bug after all, because it's a hidden feature.

PR #7583, opened 2019-05-16
[Core] Fix empty ReferenceSourceTarget added for new project reference

mrward:
Adding a reference to another project would result in an empty
ReferenceSourceTarget being added to the project file.

    <ProjectReference Include="..\lib\lib.csproj">
      <Project>{7A11FE57-9070-408C-A100-F1C720FFED6C}</Project>
      <Name>lib</Name>
      <ReferenceSourceTarget></ReferenceSourceTarget>
    </ProjectReference>

Support for ItemDefinitionGroups was added recently and one exists for
ProjectReference which defines the ReferenceSourceTarget to be
'ProjectReference'. Since the ProjectReference item added to the project
in the IDE did not have the ReferenceSourceTarget set an empty value
was being used on saving the project. Now the ReferenceSourceTarget
is set for new project references to be ProjectReference which will
result in the ReferenceSourceTarget not being added to the project file.

Fixes VSTS #893190 - Empty ReferenceSourceTarget added to project


Therzok (approved):

Therzok:
❤️

slluis (approved):

PR #7590, opened 2019-05-17
[NuGet] Fix multiple NuGet restores run on target framework changed

mrward:
[NuGet] Restore only projects affected when target framework changed

When the target framework changed in a .NET Core project all .NET
Core projects in the soluton were restored. Now only the project and
any of its transitive project references are restored.

Fixes VSTS #895149 - Target Framework change in one project restores
all .NET Core projects

[NuGet] Fix adding non-sdk style project triggering NuGet restore

Adding a new non-sdk style project to md-addins would cause a NuGet
restore to run multiple times. The problem was that the Xamarin.Mac
SDK style projects change their target framework on re-evaluation from
.NET to Xamarin.Mac. This was causing the .NET Core project extension
to run a restore for all sdk style projects in the solution, multiple
times. To avoid this the target framework changes during re-evaluation
are ignored. We are only interested in changes due to the user making
changes to the target framework.

Fixes VSTS #893871 - Adding a non-sdk style project to a solution can
trigger a NuGet restore

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
@@ -263,8 +263,6 @@ namespace MonoDevelop.DotNetCore
 			if (!IdeApp.IsInitialized)
 				return;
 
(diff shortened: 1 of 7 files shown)


mrward:
Failing tests look unrelated - DDRIT and the Xamarin.Forms tests which have been failing for other pull requests.

PR #7598, opened 2019-05-18
[Ide] Add proper handling of gtk/objc exceptions

Therzok:
We need to throw the exceptions, otherwise nothing processes stacktrace information properly, which is only set by the runtime


Therzok:
cc @leculver we can't avoid this.

PR #7599, opened 2019-05-18
[UITests] Try fix race, not doing a wait for the idle might end up wi…

Therzok:
…th methods not being called


alanmcgovern:
https://github.com/mono/monodevelop/blob/7f828fcd55a7096104fa5c3fbcfde0d3cda53005/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest/AutoTestSession.cs#L378

Isn't the default to pass `wait: true`, so this change is a no-op?

PR #7602, opened 2019-05-20
Fixes VSTS Bug 860976: [Watson] System.NullReferenceException in

mkrueger:
MonoDevelop.Ide.dll!MonoDevelop.Ide.Gui.Document.GetContent
(MonoDevelop.Ide.dll)

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/860976

### DIFF
--- a/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.Wrappers/TooltipProviderWrapper.cs
+++ b/main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.Wrappers/TooltipProviderWrapper.cs
@@ -56,11 +56,11 @@ namespace MonoDevelop.SourceEditor.Wrappers
 		static MonoDevelop.Ide.Editor.TextEditor WrapEditor (MonoTextEditor editor)
 		{
 			foreach (var doc in IdeApp.Workbench.Documents) {
-				var textEditor = doc.Editor;
-				if (textEditor == null)
-					continue;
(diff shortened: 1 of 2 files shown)


mkrueger:
GetContent can throw a null reference exception that shouldn't be but with the info from the bug I couldn't find the cause of the issue. 

But that should work-around it.

slluis:
@mkrueger looking at the stack trace of the bug, it's clear it belongs to 8.0. I don't think it can happen in 8.1 or master.

main/src/addins/MonoDevelop.SourceEditor2/MonoDevelop.SourceEditor.Wrappers/TooltipProviderWrapper.cs
```diff
@@ -56,11 +56,11 @@ public TooltipProviderWrapper (MonoDevelop.Ide.Editor.TooltipProvider provider)
...
 			foreach (var doc in IdeApp.Workbench.Documents) {
-				var textEditor = doc.Editor;
-				if (textEditor == null)
-					continue;
-				if (textEditor.FileName == editor.FileName)
-					return textEditor;
+				if (doc.FileName == editor.FileName) {
+					var textEditor = doc.Editor;
+					if (textEditor != null)
+    					return textEditor;
```
Therzok: Whitespace is mixed here.

mkrueger:
@slluis ok then I close this PR.

PR #7603, opened 2019-05-20
Fix issue when closing floating document notebook

slluis:
When closing the floating window, only the first document of the notebook
was closed before returning from the close event handler.
The fix ensures that all documents are closed.

Fixes VSTS 895763 - Document close issue with floating documents


slluis:

PR #7605, opened 2019-05-20
Fixes VSTS Bug 866898: Project file not reloaded in editor after

mkrueger:
changes made to project
Fixes VSTS Bug 897623: Unable to save changes after undoing changes for reloaded file

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/866898

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditorViewContent.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditorViewContent.cs
@@ -1,4 +1,4 @@
-//
+﻿//
 // TextEditorViewContent.cs
 //
 // Author:
@@ -61,6 +61,14 @@ namespace MonoDevelop.Ide.Editor
 
 		public TextEditorViewContent ()
 		{
+		}
+
+		protected override Task OnLoad (bool reloading)
+		{
+			return base.OnLoad (reloading).ContinueWith (delegate {
+				if (textEditor != null)
+					textEditor.IsDirty = false;
+			});
 		}
 
 		protected override async Task OnInitialize (ModelDescriptor modelDescriptor, Properties status)
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditorViewContent.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditorViewContent.cs
@@ -63,12 +63,13 @@ namespace MonoDevelop.Ide.Editor
 		{
 		}
 
-		protected override Task OnLoad (bool reloading)
+		protected override async Task OnLoad (bool reloading)
 		{
-			return base.OnLoad (reloading).ContinueWith (delegate {
-				if (textEditor != null)
-					textEditor.IsDirty = false;
-			});
+			await base.OnLoad (reloading);
+			if (isDisposed)
+				return;
+ 			textEditor.SetNotDirtyState ();
+			textEditor.IsDirty = false;
 		}
 
 		protected override async Task OnInitialize (ModelDescriptor modelDescriptor, Properties status)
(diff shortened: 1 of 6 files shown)


mrward:
With a .csproj open in the IDE if you add a reference the .csproj file is reloaded but it is marked as unsaved/dirty. So further project changes trigger a message saying the file is changed outside the IDE.

Not sure what the expected behaviour is here but this is different to VS Mac 8.0, and older versions, and also how the new text editor in VS Mac 8.1/8.2 works for C# files. I was expecting the file to be reloaded and not marked as dirty/unsaved.

mkrueger:
It's caused by the new document infrastructure changes. It's now using the vs.net infrastructure maybe I can find a work around but reloading now creates an edit operation.

mrward:
OK reloading is working now.

One problem that remains is that the dirty flag is no longer set if you edit the .csproj file in the text editor after it has been reloaded. Which means you cannot save changes.

Repro steps:

1. Open/Create a C# library project.
2. Open .csproj into text editor.
3. Make changes to .csproj
4. file is marked as dirty/unsaved and can be saved. Save the .csproj file.
5. Edit references and add a reference to System.Xml.
6. .csproj file is reloaded in text editor - file is not marked as dirty/unsaved.
7. Make some changes to the .csproj in the text editor.

Expected result:

File is marked as unsaved/dirty and can be saved.

Actual result:

No unsaved/dirty icon shown in text editor tab label. csproj file cannot be saved.

mkrueger:
fixed something doesn't work in the document layer. Have implemented a work around for it.

sevoku:
I've verified the scenario mentioned by @mrward and it seems to be working now too.

sevoku:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
```diff
@@ -377,6 +377,8 @@ public Task Reload ()
 		async Task ReloadTask ()
 		{
 			await controller.Reload ();
+			if (Editor != null)
+				Editor.IsDirty = false; // Fixes the old editor dirty state.
```
slluis: Please don't add more editor-specific code to Document.cs

mkrueger: I tried to fix it otherwise the API you're using is not implemented in the old editor infrastructure.

I did not break it - atm I don't see another way than that to fix this.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditorViewContent.cs
```diff
@@ -63,6 +63,16 @@ public TextEditorViewContent ()
 		{
 		}
 
+		protected override Task OnLoad (bool reloading)
+		{
+			return base.OnLoad (reloading).ContinueWith (delegate {
+				if (textEditor != null) {
+					textEditor.SetNotDirtyState ();
+					textEditor.IsDirty = false;
+				}
```
Therzok: Need to specify TaskScheduler if we want UI thread updates.

Therzok: Also, cancellation token that's cancelled on dispose.

mkrueger: if it 's diposed then textEditor == null.

mkrueger:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditorViewContent.cs
```diff
@@ -63,6 +63,15 @@ public TextEditorViewContent ()
 		{
 		}
 
+		protected override async Task OnLoad (bool reloading)
+		{
+			await base.OnLoad (reloading);
+			if (isDisposed)
+				return;
+ 			textEditor.SetNotDirtyState ();
```
sevoku: @mkrueger [Xamarin.Forms tests fail with an NRE](https://devdiv.visualstudio.com/DevDiv/_release?releaseId=339171&_a=release-contribution-tab-ms.vss-test-web.test-result-in-release-management) in this line

slluis:

PR #7606, opened 2019-05-20
Fix exception in DocumentController.GetContent()

slluis:
The text editor was doing an incorrect type check in the GetContent implementation.
The document changed event is now raised in a more safe way, so that an exception
in the handler won't crash the IDE, and will ensure that all handlers are
invoked no matter if there is an exception.

Fixes VSTS #891457 - FATAL ERROR: An unhandled exception has occurred.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
@@ -550,18 +550,25 @@ namespace MonoDevelop.Ide.Gui.Documents
 
 			CheckInitialized ();
 			foreach (var c in OnGetContents (type))
-				yield return c;
+				yield return AssertValid (c, type);
 
 			if (extensionChain != null) {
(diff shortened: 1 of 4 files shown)


slluis:

mrward (approved):

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
```diff
@@ -550,20 +550,27 @@ public IEnumerable<object> GetContents (Type type)
...
 				}
 			}
 			if (linkedController != null) {
 				foreach (var c in linkedController.GetContents (type))
-					yield return c;
+					yield return AssertValid (c, type);
 			}
 		}
 
+		object AssertValid (object ob, Type type)
```
Therzok: Should we be doing this on release builds?

slluis: Yes, it can help tracking down issues reported by users.

PR #7608, opened 2019-05-20
[TextEditor] Don't manually dispose the TextDocument instance.

garuma:
Instead the task of disposing it should be left to the document model registry. In case where said instance is shared, the manual dispose would corrupt the version still stored inside the registry causing all sorts of failures when reopening documents.


slluis:
@garuma good catch! is there any specific bug already reported that you think might be caused by this?

garuma:
There may be one that I think one my team member encountered but I'm checking with him if he actually ended up logging it.

garuma:
At quick glance the test failures look to be fluke. Should they be restarted or can the PR be merged as-is?

slluis:

PR #7546, opened 2019-05-10
TextViewContent: Clean up CanHandleCommand

sandyarmstrong:
Only look for a find presenter when querying about find
commands.

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/890051

PR #7610, opened 2019-05-20
Fix Find commands availability.

KirillOsenkov:
The previous fix for https://devdiv.visualstudio.com/DevDiv/_workitems/edit/890051 was subtly incorrect. It prevented the control flow for find commands from falling through. Turns out when Find is not focused the control flow needs to fall through to the default.

A different way to fix it is to check TextView for null to avoid the NullReferenceException from 890051.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/893794 - Unable to search in text editor with inline search

### DIFF
--- a/main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewContent.Commands.cs
+++ b/main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewContent.Commands.cs
@@ -190,9 +190,12 @@ namespace MonoDevelop.TextEditor
 
 		bool CanHandleCommand (object commandId)
 		{
-			if (commandsSupportedWhenFindPresenterIsFocused.Contains (commandId)) {
-				var findPresenter = Imports.FindPresenterFactory?.TryGetFindPresenter (TextView);
-				return findPresenter != null && findPresenter.IsFocused;
+			// check TextView for null because of https://devdiv.visualstudio.com/DevDiv/_workitems/edit/890051
+			if (TextView is ITextView textView) {
+				var findPresenter = Imports.FindPresenterFactory?.TryGetFindPresenter (textView);
+				if (findPresenter != null && findPresenter.IsFocused) {
+					return commandsSupportedWhenFindPresenterIsFocused.Contains (commandId);
+				}
 			}
 
 			return true;


KirillOsenkov:
I want to test this fix thoroughly before merging.

Therzok:
We need this in master too.

KirillOsenkov:
Yes, will backport if this works.

sandyarmstrong:
Looking at this now, I think that in https://github.com/mono/monodevelop/pull/7546 I might have just forgotten an `else return false;`? I'll test today.

DavidKarlas:
FYI: Fix works for me

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewContent.Commands.cs
```diff
@@ -190,9 +190,12 @@ void FormatBufferUpdate (EditorOperationsInterface op, CommandInfo info)
 
 		bool CanHandleCommand (object commandId)
 		{
-			if (commandsSupportedWhenFindPresenterIsFocused.Contains (commandId)) {
-				var findPresenter = Imports.FindPresenterFactory?.TryGetFindPresenter (TextView);
-				return findPresenter != null && findPresenter.IsFocused;
+			// check TextView for null because of https://devdiv.visualstudio.com/DevDiv/_workitems/edit/890051
+			if (TextView is ITextView textView) {
```
sandyarmstrong: This should be an actual null check. We already know `TextView` is `ITextView`.

sandyarmstrong: (and yes, that means we need to update the generic constraint to make sure `TView` is a reference type)

abock: `is` _is_ an actually null check for reference types. It's even more of an null check that `== null` because you cannot override `is`.

sandyarmstrong: Sure, but are you advocating that we replace all `== null` checks with `is` checks?

sandyarmstrong: I think it's pretty clear what reads better here, is all I'm saying. But yes, obviously this works. :) And thank you to Kirill for fixing my mistake!

abock: I'm not advocating that (here). You said that it should an "actual null check" and what @KirillOsenkov has here is a better/best null check.

sandyarmstrong: LOL. Ok, would "a literal null check" be better phrasing for you?

KirillOsenkov: The reason I use the is pattern here is because I wanted to grab a copy of TextView and work with it, avoiding reading the TextView twice. If it can randomly change to null then I want to avoid the race condition.

abock:
@monojenkins backport master

abock:
@KirillOsenkov please remove `do-not-merge` if you are happy with the PR.

KirillOsenkov:
@monojenkins backport master

PR #7619, opened 2019-05-22
TextEditor: Enable multi-caret commands during Find

sandyarmstrong:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/896797


abock:

PR #7624, opened 2019-05-22
[NuGet] Fix being unable to add or update NuGet package sources

mrward:
After updating to NuGet 5 it was not possible to add, remove or update
package sources in Preferences if a package source had a username and
password associated with it. The problem was that the package source
url was being used when saving the username and password information
in the packageSourceCredentials section in the NuGet.Config file. This
resulted in an invalid xml element name so the save would fail. In
NuGet 4.8 the PackageSourceCredential.Source was not used when saving
the NuGet.Config file. With NuGet 4.9 and later it is being used.
However it is not the source url but the source name.

Fixes VSTS #857001 - Unable to update NuGet package sources


mrward:

PR #7630, opened 2019-05-22
Fix 762833: Implement selection drag and drop within/across text views

DavidKarlas:


abock:

PR #7642, opened 2019-05-23
[PropertyGrid] Fixes NRE on a second call to EndEditing and currentEditor is null

netonjm:
This issue was detected by Watson, and probably could be related call EndEditing 2 times and we don't check for null the currentEditor.

I'm not completely sure about the fix, because at the end we refresh 2 times the grid, maybe we could don't allow call EndEditing 2 times?

Fixes VSTS #861004 - [Watson] System.NullReferenceException in MonoDevelop.Components.PropertyGrid.PropertyGridTable::EndEditing (System.NullReferenceException_MonoDevelop.Ide.dll)


slluis:

Issue #7561, opened 2019-05-13
Find in files - open files - returns no results on re-opening solution


mrward:
Build Information
Release ID: 801002462
Git revision: 29162fa2ff21bf2018bd5d3a7f60866fb62d547b
Build date: 2019-05-10 21:38:59+00
Build branch: release-8.1
Xamarin extensions: 735d3d5d1daaee9cf9b045febed581680960e8b2

Opening a solution that has several files that get re-opened, then trying to search for text using the Find in Files, with 'Look in' set to 'All open files', does not return find any results for files that have not been made visible in the text editor.

To repro:

1. Create a new C# library project.
2. Add a new empty class called Class2.
3. Add a new empth class called Class3.
4. Close solution and re-open it.
5. MyClass, Class2, and Class3 should be open. Class3 is visible.
6. Open the Find in files dialog, select Look in - All open files, and search for Class2.

Expected result:

Search finds Class2 in Class2.cs file.

Actual result:

No search results found.
After selecting the Class2 tab then find in files will work.

> VS bug [#892073](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/892073)

xamarin-release-manager:
Fixed in version 8.1.0.2581 (release-8.1)

Pull Request #7656 merged by: Vsevolod Kukol
Author: mono
Commit: 70ea11980e9c1fccfe18948dbc175eca77a6b5a4 (mono/monodevelop)

PR #7647, opened 2019-05-23
Fixes issue #7561 Find in files - open files - returns no results on

mkrueger:
re-opening solution This is better than forcing the load, if many
files are open forcing to load all at once without need is worse than
the possibility to load a file twice.


slluis:

PR #7650, opened 2019-05-23
[A11y] Check the functions have been set correctly before calling

iainx:
If we try to call a null func<> then we'll get an NRE, so check they're set and warn if not.

Fixes VSTS #780731

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs
@@ -638,17 +638,17 @@ namespace MonoDevelop.Components.AtkCocoaHelper
 																  Func<AtkCocoa.Range> visibleCharacterRange)
 		{
 			return new AccessibilityElementProxy (new RealAccessibilityElementNavigableStaticTextProxy () {
-				Contents = contents,
-				NumberOfCharacters = numberOfCharacters,
-				InsertionPointLineNumber = insertionPointLineNumber,
-				GetFrameForRange = frameForRange,
-				GetLineForIndex = lineForIndex,
-				GetRangeForLine = rangeForLine,
-				GetStringForRange = stringForRange,
-				GetRangeForIndex = rangeForIndex,
-				GetStyleRangeForIndex = styleRangeForIndex,
-				GetRangeForPosition = rangeForPosition,
-				GetVisibleCharacterRange = visibleCharacterRange
+				Contents = contents ?? throw new ArgumentNullException (nameof (contents)),
+				NumberOfCharacters = numberOfCharacters ?? throw new ArgumentNullException (nameof (numberOfCharacters)),
+				InsertionPointLineNumber = insertionPointLineNumber ?? throw new ArgumentNullException (nameof (insertionPointLineNumber)),
+				GetFrameForRange = frameForRange ?? throw new ArgumentNullException (nameof (frameForRange)),
+				GetLineForIndex = lineForIndex ?? throw new ArgumentNullException (nameof (lineForIndex)),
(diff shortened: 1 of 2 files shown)


iainx:

Therzok:
Is there any way we can determine whether `InsertionPointLineNumber` has to be set? Maybe we can make it mandatory via the constructor (aka passing them as parameters to `TextElementProxy()`)

iainx:
They are mandatory, but to break ref cycles need to be nulled out when the parent object is disposed. So I've made them mandatory with the constructor, but left in the try catch in case the object is still hanging around after the parent object is disposed, so we can get a stack trace for a possible bug.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs
```diff
@@ -624,9 +625,31 @@ public static AccessibilityElementProxy ButtonElementProxy ()
...
+																  Func<int, int> lineForIndex,
+																  Func<int, AtkCocoa.Range> rangeForLine,
+																  Func<AtkCocoa.Range, string> stringForRange,
+																  Func<int, AtkCocoa.Range> rangeForIndex,
+																  Func<int, AtkCocoa.Range> styleRangeForIndex,
+																  Func<Point, AtkCocoa.Range> rangeForPosition,
+																  Func<AtkCocoa.Range> visibleCharacterRange)
+		{
+			return new AccessibilityElementProxy (new RealAccessibilityElementNavigableStaticTextProxy () {
+				Contents = contents,
```
Therzok: on all the lines below:

```suggestion
				Contents = contents ?? throw new ArgumentNullException (nameof(contents)),
```

Therzok:
Perfect!

slluis:
> They are mandatory, but to break ref cycles need to be nulled out when the parent object is disposed. So I've made them mandatory with the constructor, but left in the try catch in case the object is still hanging around after the parent object is disposed, so we can get a stack trace for a possible bug.

So do we know what is causing VSTS #780731? This PR will avoid the crash, but there is still something trying to invoke a null function.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs
```diff
@@ -624,9 +625,31 @@ public static AccessibilityElementProxy ButtonElementProxy ()
...
+																  Func<int, int> lineForIndex,
+																  Func<int, AtkCocoa.Range> rangeForLine,
+																  Func<AtkCocoa.Range, string> stringForRange,
+																  Func<int, AtkCocoa.Range> rangeForIndex,
+																  Func<int, AtkCocoa.Range> styleRangeForIndex,
+																  Func<Point, AtkCocoa.Range> rangeForPosition,
+																  Func<AtkCocoa.Range> visibleCharacterRange)
+		{
+			return new AccessibilityElementProxy (new RealAccessibilityElementNavigableStaticTextProxy () {
+				Contents = contents ?? throw new ArgumentNullException (nameof (contents)),
```
Therzok: Even here, the `Func<>`s should be externally readonly.

See here: http://source.monodevelop.com/#MonoDevelop.SourceEditor/Mono.TextEditor/Gui/TextViewMargin.cs,141

We shouldn't allow (at least public) mutable state. Disposing (i.e. on a background thread) might cause racing values.

iainx:
> So do we know what is causing VSTS #780731? This PR will avoid the crash, but there is still something trying to invoke a null function.

I suspect something is keeping a reference to the object after its parent has been finalised and the delegates have been nulled, where that is I'm not sure yet. This patch will at least let us track when it occurs and prevents the NRE.

This only affects the old Gtk editor, so it will become less of an issue as the new editor is improved

slluis:

PR #7657, opened 2019-05-24
[Mac] Cleanup MONO_REGISTRY_PATH on startup

rodrmoya:
If started from a Xamarin.Mac app, this env var is set pointing to a (maybe)
non-existing folder (~/Library/Application Support/Whatever), which causes
code that tries to access the registry fail if that folder doesn't exist. So,
since we want to use Mono's registry, make sure the env var is cleaned up
on startup.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/896438

### DIFF
--- a/main/build/MacOSX/monostub.mm
+++ b/main/build/MacOSX/monostub.mm
@@ -269,6 +269,14 @@ main (int argc, char **argv)
 				req_mono_version = version_obj;
 		}
 
+		// Xamarin.Mac sets MONO_REGISTRY_PATH to ~/Library/Application Support/...,
+		// so in the case of the Xamarin Installer, this gets set to a non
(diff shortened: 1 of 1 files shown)


rodrmoya:

main/build/MacOSX/monostub.mm
```diff
@@ -269,6 +269,7 @@
 				req_mono_version = version_obj;
 		}
 
+		unsetenv ("MONO_REGISTRY_PATH");
```
slluis: Could you add a comment explaining why was this added?

rodrmoya: Done!

rodrmoya:
Backported manually at https://github.com/mono/monodevelop/pull/7660

PR #7661, opened 2019-05-24
[Core] Fix error saving project with ToolsVersion set to Current

mrward:
Saving an SDK style project that set ToolsVersion="Current" would
fail as an attempt was made to parse the version using System.Version.

    System.ArgumentException: Version string portion was too short or too long.
    Parameter name: input
      at System.Version.ParseVersion

Fixes VSTS #897947 - SDK Projects with ToolsVersion="Current"

### DIFF
--- a/main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/ProjectLoadSaveTests.cs
+++ b/main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/ProjectLoadSaveTests.cs
@@ -1393,6 +1393,19 @@ namespace MonoDevelop.Projects
 				Assert.That (p.LoadError, Contains.Substring ("circular dependency"));
 			}
 		}
+
+		[Test]
+		public async Task Save_SdkProject_ToolsVersionCurrent ()
+		{
+			string projectFile = Util.GetSampleProject ("tools-version-current", "netstandard.csproj");
+			using (var p = await Services.ProjectService.ReadSolutionItem (Util.GetMonitor (), projectFile) as DotNetProject) {
+				p.DefaultNamespace = "Test";
+				await p.SaveAsync (Util.GetMonitor ());
+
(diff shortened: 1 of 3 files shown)


mrward:

main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/ProjectLoadSaveTests.cs
```diff
@@ -1393,6 +1393,19 @@ public async Task LoadProject_ImportHasCircularDependency ()
...
 			}
 		}
+
+		[Test]
+		public async Task Save_SdkProject_ToolsVersionCurrent ()
+		{
+			string projectFile = Util.GetSampleProject ("tools-version-current", "netstandard.csproj");
+			using (var p = await Services.ProjectService.ReadSolutionItem (Util.GetMonitor (), projectFile) as DotNetProject) {
+				p.DefaultNamespace = "Test";
+				await p.SaveAsync (Util.GetMonitor ());
```
Therzok: Should we also check the return value of save async?

Therzok (approved):
LGTM!

main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/ProjectLoadSaveTests.cs
mrward: Project.SaveAsync does not return anything to check. Without the code fix SaveAsync was throwing an exception. I thought about doing an Assert.DoesNotThrow to indicate where the problem was.

Therzok: Well, no need for that, throwing code would show us where the problem is.

It seems like SaveAsync always returns true anyway...

slluis (approved):

PR #7667, opened 2019-05-24
Fix tab stacking issue in designers

slluis:
When a tabbed view has attached views, there is a tab stacking effect
that is very confusing.

To avoid this problem, when a tabbed view has a child that also has
tabs, the tabs are now shown in the parent's tab area, next to the
parent's tabs.

Before:

Captura de Pantalla 2019-05-24 a les 19 30 20

After:

Captura de Pantalla 2019-05-24 a les 19 30 30

Fixes VSTS# 893310 - New split view results in stacked tabs with the
same name

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Shell/GtkShellDocumentViewContainer.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Shell/GtkShellDocumentViewContainer.cs
@@ -65,7 +65,6 @@ namespace MonoDevelop.Ide.Gui.Shell
 			rootTabsBox.PackEnd (bottomBarBox, false, false, 0);
 
 			tabstrip = new Tabstrip ();
-			tabstrip.Show ();
 			bottomBarBox.PackStart (tabstrip, true, true, 0);
 
 			rootTabsBox.Show ();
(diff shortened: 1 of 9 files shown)


slluis:

jsuarezruiz:
I've tried the branch. Works fine. I have tried in different resolutions, themes, etc.

Screenshot 2019-05-27 at 10 49 43

Screenshot 2019-05-27 at 10 47 25

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Shell/GtkShellDocumentViewContainer.cs
```diff
@@ -156,6 +163,87 @@ public void SetCurrentMode (DocumentViewContainerMode mode, GtkShellDocumentView
...
+				foreach (var child in attachedChildTabstrips)
+					DetachTabstrip (child);
+				attachedChildTabstrips.Clear ();
+				return;
+			}
+
+			// Get a list of children with visible tabstrips. Notice that if any of the child tabstrips is hidden, then this
+			// method will be called and the tabstrip will be removed
+
+			var childrenWithTabs = GetAllViews ().OfType<GtkShellDocumentViewContainer> ().Where (c => c.tabstrip.Visible).ToList ();
```
netonjm: we could improve this call somehow?

slluis: What do you suggest?

netonjm: not sure, but maybe include a filtering in the GetAllView calls?

netonjm: Oh! forgot it, I just figured we are talking about a very low number of views 👍

PR #7671, opened 2019-05-24
Fix issues when opening a document

slluis:
When opening a document, the view may take long time to initialize (for example when loading a designer). This used to delay the whole document initialization sequence. During that initialization the corresponding tab in the shell was already visible, and that caused some bugs since there was
code expecting a fully initialized document.

Two things have been done to solve this problem. First, the document initialization doesn't wait anymore for the view to be initialized, so that when the tab is show, the document initialization is complete. Second, checks have been added to the Document and SdiWorkspaceWindow classes
to properly handle documents that don't yet have a view.

Also removed obsolete WorkbenchWindow api.

Fixes VSTS #889288 - Opening a document does not focus the new document
Fixes VSTS #889300 - NavigationHistoryService triggers an exception which crashes the IDE


slluis:

slluis:

PR #7675, opened 2019-05-25
Preserve C# documents added by Razor when reloading a project.

KirillOsenkov:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/889145 - [WebToolsE2E][VSforMac]First time to add MVC View Page file(.cshtml), the intellisense and colorization doesn't work at once and shows un-save state. It needs to reopen or insert.

When a project is reloaded the C# buffers we added manually are lost, and so we need to keep track of them and readd them on each reload. This explains why we lost classification and other Roslyn features for C# chunks of .cshtml files when a project was reloaded while a .cshtml file was open.


KirillOsenkov:
Tested and this fixes 889145.

KirillOsenkov:
@monojenkins backport master

KirillOsenkov:
@slluis this WebTools fix would be nice to get in

slluis:
I'm approving, but if we still need this after switching web editors to the new editor, we'll need a better solution.

PR #7682, opened 2019-05-27
Add support for automatic file nesting

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/781656
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/857165

### DIFF
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreNestingRulesProvider.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreNestingRulesProvider.cs
@@ -30,7 +30,7 @@ using MonoDevelop.Projects.FileNesting;
 namespace MonoDevelop.AspNetCore
 {
 	[Extension]
-	class AspNetCoreNestingRulesProvider : NestingRulesProvider
+	sealed class AspNetCoreNestingRulesProvider : NestingRulesProvider
 	{
 		public AspNetCoreNestingRulesProvider ()
 		{
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreNestingRulesProvider.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreNestingRulesProvider.cs
@@ -34,7 +34,7 @@ namespace MonoDevelop.AspNetCore
 	{
 		public AspNetCoreNestingRulesProvider ()
 		{
-			SourceFile = AddinManager.CurrentAddin.GetFilePath ("Resources/AspNetCore.filenesting.json");
+			SourceFile = AddinManager.CurrentAddin.GetFilePath (Path.Combine ("Resources", "AspNetCore.filenesting.json"));
 		}
 	}
 }
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreNestingRulesProvider.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreNestingRulesProvider.cs
@@ -24,6 +24,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+using System.IO;
 using Mono.Addins;
 using MonoDevelop.Projects.FileNesting;
 
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreNestingRulesProvider.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreNestingRulesProvider.cs
@@ -1,10 +1,10 @@
 //
-// FileNestingService.cs
+// AspNetCoreNestingRulesProvider.cs
 //
 // Author:
 //       Rodrigo Moya <rodrigo.moya@xamarin.com>
 //
-// Copyright (c) 2019 Microsoft, Inc. (http://microsoft.com)
+// Copyright (c) 2019 Microsoft Corp. (http://microsoft.com)
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -24,36 +24,17 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-using System;
-using System.Collections.Generic;
+using Mono.Addins;
+using MonoDevelop.Projects.FileNesting;
 
-namespace MonoDevelop.Projects.FileNesting
+namespace MonoDevelop.AspNetCore
 {
-	internal static class FileNestingService
+	[Extension]
+	class AspNetCoreNestingRulesProvider : NestingRulesProvider
 	{
-		static List<NestingRulesProvider> rulesProviders;
-
-		static FileNestingService ()
+		public AspNetCoreNestingRulesProvider ()
 		{
-		}
-
-		public static string GetParentFile (string inputFile)
-		{
-			if (rulesProviders == null) {
-				rulesProviders = new List<NestingRulesProvider> ();
-				foreach (var pr in Mono.Addins.AddinManager.GetExtensionObjects<NestingRulesProvider> ()) {
-					rulesProviders.Add (pr);
-				}
-			}
-
-			foreach (var rp in rulesProviders) {
-				var parentFile = rp.GetParentFile (inputFile);
-				if (!String.IsNullOrEmpty (parentFile)) {
-					return parentFile;
-				}
-			}
-
-			return null;
+			SourceFile = AddinManager.CurrentAddin.GetFilePath ("Resources/AspNetCore.filenesting.json");
 		}
 	}
 }
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
@@ -57,15 +57,12 @@ namespace MonoDevelop.Projects.FileNesting
 		public static string GetParentFile (string inputFile)
 		{
 			if (rulesProviders == null) {
-				rulesProviders = ImmutableList<NestingRulesProvider>.Empty;
-				foreach (var pr in AddinManager.GetExtensionObjects<NestingRulesProvider> ()) {
-					rulesProviders = rulesProviders.Add (pr);
-				}
+				rulesProviders = ImmutableList<NestingRulesProvider>.Empty.AddRange (AddinManager.GetExtensionObjects<NestingRulesProvider> ());
 			}
 
 			foreach (var rp in rulesProviders) {
 				var parentFile = rp.GetParentFile (inputFile);
-				if (!String.IsNullOrEmpty (parentFile)) {
+				if (!string.IsNullOrEmpty (parentFile)) {
 					return parentFile;
 				}
 			}
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
@@ -54,7 +54,7 @@ namespace MonoDevelop.Projects.FileNesting
 			}
 		}
 
-		public static string GetParentFile (string inputFile)
+		public static FilePath GetParentFile (FilePath inputFile)
 		{
 			if (rulesProviders == null) {
 				rulesProviders = ImmutableList<NestingRulesProvider>.Empty.AddRange (AddinManager.GetExtensionObjects<NestingRulesProvider> ());
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
@@ -26,7 +26,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.IO;
 
 namespace MonoDevelop.Projects.FileNesting
 {
@@ -37,7 +36,8 @@ namespace MonoDevelop.Projects.FileNesting
 		static FileNestingService ()
 		{
 			rulesProviders = new List<NestingRulesProvider> ();
-			rulesProviders.Add (new NestingRulesProvider ());
+			// FIXME: this is temporary, remove when we load rules from json files
+			rulesProviders.Add (new NestingRulesProvider ("/Users/rodrigo/Downloads/Settings1.filenesting.json"));
 		}
 
 		public static string GetParentFile (string inputFile)
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
@@ -25,24 +25,41 @@
 // THE SOFTWARE.
 
 using System;
-using System.Collections.Generic;
+using System.Collections.Immutable;
+using Mono.Addins;
+using MonoDevelop.Core;
 
 namespace MonoDevelop.Projects.FileNesting
 {
 	internal static class FileNestingService
 	{
-		static List<NestingRulesProvider> rulesProviders;
+		static ImmutableList<NestingRulesProvider> rulesProviders;
 
 		static FileNestingService ()
 		{
+			AddinManager.AddExtensionNodeHandler (typeof (NestingRulesProvider), HandleRulesProviderExtension);
+		}
+
+		static void HandleRulesProviderExtension (object sender, ExtensionNodeEventArgs args)
+		{
+			if (rulesProviders != null) {
+				var pr = args.ExtensionObject as NestingRulesProvider;
+				if (pr != null) {
+					if (args.Change == ExtensionChange.Add) {
+						rulesProviders = rulesProviders.Add (pr);
+					} else {
+						rulesProviders = rulesProviders.Remove (pr);
+					}
+				}
+			}
 		}
 
 		public static string GetParentFile (string inputFile)
 		{
 			if (rulesProviders == null) {
-				rulesProviders = new List<NestingRulesProvider> ();
-				foreach (var pr in Mono.Addins.AddinManager.GetExtensionObjects<NestingRulesProvider> ()) {
-					rulesProviders.Add (pr);
+				rulesProviders = ImmutableList<NestingRulesProvider>.Empty;
+				foreach (var pr in AddinManager.GetExtensionObjects<NestingRulesProvider> ()) {
+					rulesProviders = rulesProviders.Add (pr);
 				}
 			}
 
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
@@ -52,10 +52,10 @@ namespace MonoDevelop.Projects.FileNesting
 			}
 		}
 
-		public static FilePath GetParentFile (FilePath inputFile)
+		public static FilePath GetParentFile (Project project, FilePath inputFile)
 		{
 			foreach (var rp in rulesProviders) {
-				var parentFile = rp.GetParentFile (inputFile);
+				var parentFile = rp.GetParentFile (project, inputFile);
 				if (!string.IsNullOrEmpty (parentFile)) {
 					return parentFile;
 				}
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
@@ -1,0 +1,45 @@
+//
+// FileNestingService.cs
+//
+// Author:
+//       Rodrigo Moya <rodrigo.moya@xamarin.com>
+//
+// Copyright (c) 2019 Microsoft, Inc. (http://microsoft.com)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+using System;
+using System.IO;
+
+namespace MonoDevelop.Projects.FileNesting
+{
+	public static class FileNestingService
+	{
+		public static string GetDependentFile (string file)
+		{
+			var dependentFile = file.Substring (0, file.Length - 3);
+			if (file.EndsWith (".cshtml.cs", StringComparison.OrdinalIgnoreCase) && File.Exists (dependentFile)) {
+				Console.WriteLine ($"**** Nesting {dependentFile} under {file}");
+				return dependentFile;
+			}
+
+			return null;
+		}
+	}
+}
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
@@ -33,7 +33,7 @@ namespace MonoDevelop.Projects.FileNesting
 {
 	internal static class FileNestingService
 	{
-		static ImmutableList<NestingRulesProvider> rulesProviders;
+		static ImmutableList<NestingRulesProvider> rulesProviders = ImmutableList<NestingRulesProvider>.Empty;
 
 		static FileNestingService ()
 		{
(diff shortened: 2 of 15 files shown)


Therzok:
This should be easily testable. Needs unit tests.

rodrmoya:
> This should be easily testable. Needs unit tests.

Yeah, work in progress, and will add lots of tests

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -0,0 +1,55 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+using System;
+using System.Collections.Generic;
+
+namespace MonoDevelop.Projects.FileNesting
+{
+	public static class FileNestingService
+	{
```
mrward: Does this need to be public? MonoDevelop.Core.Tests should have access to MonoDevelop.Core's internals.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
```diff
@@ -0,0 +1,146 @@
...
+							return parentFile;
+					}
+				}
+				break;
+
+			case NestingRuleKind.FileSuffixToExtension:
+				if (inputFile.EndsWith (AppliesTo, StringComparison.OrdinalIgnoreCase)) {
+					foreach (var pt in patterns) {
+						parentFile = inputFile.Replace (AppliesTo, pt);
+						if (CheckParentForFile (inputFile, parentFile)) {
```
mrward: This looks like it will replace all occurrences of AppliesTo in the file path? I think we want to only replace the last part.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
```diff
@@ -0,0 +1,146 @@
...
+			if (File.Exists (parentFile) && inputFile != parentFile) {
+				LoggingService.LogInfo ($"Applied rule for nesting {inputFile} under {parentFile}");
+				return true;
+			}
+
+			return false;
+		}
+
+		public string GetParentFile (string inputFile)
+		{
```
mrward: Just wondering if there is a reason to not use FilePath here for the inputFile and the return value? I do not have a preference either way, but when I see `Path.` methods being used I usually think about using the FilePath class.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: I was planning to inject the .json from the AspNetCore extension, so that each extension can provide the nesting rules they want, so that's why I made it public. But maybe there's a better way to inject it? Or maybe we should just have the asp.net core rules available to all projects?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
rodrmoya: Right

rodrmoya: Fixed now

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: Moved it back to not public, as I implemented it via ExtensionPoint/Extension in https://github.com/mono/monodevelop/pull/7682/commits/627693a8d52f0768f716f1132173015d701f0b50

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreNestingRulesProvider.cs
```diff
@@ -0,0 +1,40 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+using Mono.Addins;
+using MonoDevelop.Projects.FileNesting;
+
+namespace MonoDevelop.AspNetCore
+{
+	[Extension]
+	class AspNetCoreNestingRulesProvider : NestingRulesProvider
```
Therzok:
```suggestion
	sealed class AspNetCoreNestingRulesProvider : NestingRulesProvider
```

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreNestingRulesProvider.cs
```diff
@@ -0,0 +1,40 @@
...
+	{
+		public AspNetCoreNestingRulesProvider ()
+		{
+			SourceFile = AddinManager.CurrentAddin.GetFilePath ("Resources/AspNetCore.filenesting.json");
```
Therzok:
```suggestion
			SourceFile = AddinManager.CurrentAddin.GetFilePath (Path.Combine ("Resources", "AspNetCore.filenesting.json");
```

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -0,0 +1,59 @@
...
+		static List<NestingRulesProvider> rulesProviders;
+
+		static FileNestingService ()
+		{
+		}
+
+		public static string GetParentFile (string inputFile)
+		{
+			if (rulesProviders == null) {
+				rulesProviders = new List<NestingRulesProvider> ();
```
Therzok: We're caching the list, but the providers can change at runtime. Listen for extension changes to get providers.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -0,0 +1,59 @@
...
+				foreach (var pr in Mono.Addins.AddinManager.GetExtensionObjects<NestingRulesProvider> ()) {
+					rulesProviders.Add (pr);
+				}
+			}
+
+			foreach (var rp in rulesProviders) {
+				var parentFile = rp.GetParentFile (inputFile);
+				if (!String.IsNullOrEmpty (parentFile)) {
```
Therzok:
```suggestion
				if (!string.IsNullOrEmpty (parentFile)) {
```

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
```diff
@@ -0,0 +1,147 @@
...
+	internal class NestingRule
+	{
+		public const string AllFilesWildcard = ".*";
+
+		List<string> patterns;
+
+		public NestingRule (NestingRuleKind kind, string appliesTo, IEnumerable<string> patterns)
+		{
+			Kind = kind;
+			AppliesTo = appliesTo ?? ".*";
```
Therzok:
```suggestion
			AppliesTo = appliesTo ?? AllFilesWildcard;
```

Therzok: Also, what happens when appliesTo is `string.Empty`?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
```diff
@@ -0,0 +1,147 @@
...
+			this.patterns = patterns?.ToList ();
```
Therzok: Patterns is used later on, and it can be null here. Possible NREs.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
```diff
@@ -0,0 +1,150 @@
...
+
+		List<NestingRule> nestingRules;
+
+		public NestingRulesProvider ()
+		{
+		}
+
+		public NestingRulesProvider (string fromFile)
+		{
+			SourceFile = fromFile;
```
Therzok: Check argument for null?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: Yes, although right now we just want to support the ASP.NET Core rules (or, if we find it useful, the "standard" plugins), so want to keep it as simple as possible, but because of the failing test in CI (https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-environment-logs&releaseId=357250&environmentId=1668000) I might have to some small refactoring to the `FileNestingService`, so might include that

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: I'm going to change this to use FilePath's, so not applying this commit, as this will change

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
Therzok: Disabling/enabling the aspnetcore addin can break this code. It's a matter of reliability, not simplicity here.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
rodrmoya: right, fixed now

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
rodrmoya: Done, will push shortly

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: Done

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreNestingRulesProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+// THE SOFTWARE.
+
+using System.IO;
+using Mono.Addins;
+using MonoDevelop.Projects.FileNesting;
+
+namespace MonoDevelop.AspNetCore
+{
+	[Extension]
+	sealed class AspNetCoreNestingRulesProvider : NestingRulesProvider
```
mrward: Should we support an id here to allow ordering of file nesting providers? Not sure how to do this using the attribute - usually do this in the .addin.xml.

If I write a file nesting provider and want it to be used instead of the AspNetCoreNestingRulesProvider we can do that if each one has an id, using insertbefore/insertafter.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
```diff
@@ -0,0 +1,150 @@
...
+				LoggingService.LogError ($"Unable to parse {provider.SourceFile}: {ex}");
+			}
+
+			return false;
+		}
+
+		public string GetParentFile (string inputFile)
+		{
+			if (nestingRules == null) {
+				if (!File.Exists (SourceFile)) {
```
mrward: If the source file does not exist or fails to load should we set a flag the first time this happens instead of checking File.Exists and trying to load every time GetParentFile is called?

main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/MSBuildGlobTests.cs
```diff
@@ -1379,6 +1379,7 @@ public async Task RenameXamlAndXamlCSharpFileAtSameTime (bool addXamlFilesBefore
 
 				// Project xml should be unchanged.
 				var projectXml = File.ReadAllText (p.FileName);
+				Console.WriteLine ($"Comparing expected {expectedProjectXml} to {projectXml}");
 				Assert.AreEqual (expectedProjectXml, projectXml);
```
mrward: Debug code?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
```diff
@@ -0,0 +1,150 @@
...
+		{
+			SourceFile = fromFile;
+		}
+
+		public string SourceFile { get; protected set; }
+
+		void AddRule (NestingRuleKind kind, string appliesTo, IEnumerable<string> patterns)
+		{
+			if (nestingRules == null) {
+				nestingRules = new List<NestingRule> ();
```
iantoalms: Why not just use a field initializer and avoid the null check

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
```diff
@@ -0,0 +1,150 @@
...
+				if (prop.Value.Type == JTokenType.Array) {
+					provider.AddRule (kind, prop.Name, (prop.Value as JArray).Select (x => x.Value<string> ()));
+				}
+			}
+		}
+
+		static bool LoadFromFile (NestingRulesProvider provider)
+		{
+			try {
+				using (var reader = new StreamReader (provider.SourceFile)) {
```
iantoalms: The new syntax for this is great: https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#using-declarations

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
```diff
@@ -0,0 +1,150 @@
...
+					var json = JObject.Parse (reader.ReadToEnd ());
+					if (json != null) {
+						var parentNode = json ["dependentFileProviders"] [TokenNameAdd] as JObject;
+						foreach (var rp in parentNode.Properties ()) {
```
iantoalms: nit: rp? Would be nice to have a more descriptive name

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -0,0 +1,76 @@
...
+					} else {
+						rulesProviders = rulesProviders.Remove (pr);
+					}
+				}
+			}
+		}
+
+		public static string GetParentFile (string inputFile)
+		{
+			if (rulesProviders == null) {
```
iantoalms: Just field initialize rulesProviders and avoid the null check here and above?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
```diff
@@ -0,0 +1,150 @@
...
+			} catch (Exception ex) {
+				LoggingService.LogError ($"Unable to parse {provider.SourceFile}: {ex}");
+			}
+
+			return false;
+		}
+
+		public string GetParentFile (string inputFile)
+		{
+			if (nestingRules == null) {
```
iantoalms: This can be considerably simplified. If you get rid of the null check by having a field initializer:

		if (File.Exists (SourceFile) && LoadFromFile (this)) {
			foreach (var rule in nestingRules) {
				string parentFile = rule.GetParentFile (inputFile);
				if (!String.IsNullOrEmpty (parentFile)) {
					// Stop at the 1st rule found
					return parentFile;
				}
			}
		}
		return null;

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreNestingRulesProvider.cs
rodrmoya: Good question. Right now, there's only 1 `NestingRulesProvider`, but indeed we might want to add more. I've been thinking about this, and maybe a solution would be to associate those providers with specific project types (or all), so the current provider would just apply to ASP.NET Core projects, another one could be added for Xamarin.ios/Mac/Android, etc.

About the `insertAfter/insertBefore`, not sure, as in theory the order should be a configurable setting for the user (once we support custom file nesting UI), but might be an option.

main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/MSBuildGlobTests.cs
rodrmoya: Removed

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
rodrmoya: The null value is used to know if the rules need to be loaded from file, so if I field initialise it, the check here (https://github.com/mono/monodevelop/pull/7682/files/8efa0b1234e601b97eb4e3a9a21e86d964d862f3#diff-9f0f0b39ec9b79ec2d5db396c3fde582R128) would just have to check for ' Count` on the collection, and in some cases it might do an extra read from the file every time it is called.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreNestingRulesProvider.cs
mrward: OK - if these can be ordered by the user, which seems to be the case in VS Win, then insertbefore/after is not useful here. I guess the ordering will need to be defined somewhere else and applied in the FileNestingService.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: Same reason as above, really want to make sure file is read only once

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
rodrmoya: 👍

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
rodrmoya: hmm, but what if the file gets created after? Not that it can happen right now, as there's only 1 provider that gets loaded on startup, but it might happen when we have more. You mean then just marking as not existing and use FSW to detect if/when it gets created?

mrward: Mostly thinking about avoiding File.Exists being called for each file in the Solution pad. Was not thinking about reloading of the file due to changes by the user.

rodrmoya: ah, but this `SourceFile` is the .json file with the rules, not every file in the solution. But, that said, I think I do a File.Exists for every file returned by this, so will check.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
rodrmoya: Changed

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
rodrmoya: But we don't want to load the file every single time, just once, hence the null initial value/check.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -0,0 +1,73 @@
...
+using System;
+using System.Collections.Immutable;
+using Mono.Addins;
+using MonoDevelop.Core;
+
+namespace MonoDevelop.Projects.FileNesting
+{
+	internal static class FileNestingService
+	{
+		static ImmutableList<NestingRulesProvider> rulesProviders;
```
Therzok: Why are we using an ImmutableList here?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
rodrmoya: hmm, something's not setup correctly for C#8 then, as changing the code to do that shows a warning, about the statement being empty and `reader` not being defined:

image

@Therzok ? this is vsmac 8.2 (build 832)

rodrmoya: C# language version is set to 8 for that project

Therzok: it should be:
```csharp
using var reader = new StreamReader (provider.SourceFile);
```

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: It's a very small collection (only 1 item max now) and solves concurrent access, which is needed because the ExtensionNodeHandler can be ran on a separate thread than the one used for resolving, although I guess it could use a Concurrent* collection instead, although for such a small collection, not sure it's worth.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
Therzok: Putting the extra set of parens makes it behave as if it were:

```csharp
using (var reader = new StreamReader (provider.SourceFile)) {}
```

rodrmoya: ugh, indeed 👍

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
Therzok: Understood, now I see the _why_.

iantoalms (approved):

jtorres (approved):
apart of the moving files issue, looks good

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
```diff
@@ -0,0 +1,151 @@
...
+				LoggingService.LogError ($"Unable to parse {provider.SourceFile}: {ex}");
+			}
+
+			return false;
+		}
+
+		public FilePath GetParentFile (FilePath inputFile)
+		{
+			if (nestingRules == null) {
+				if (!File.Exists (SourceFile)) {
```
mrward: Still doing a File.Exists here. Which will be called for every file in the project if the rules file does not exist, or if the load fails.

rodrmoya: Right, fixed now

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -0,0 +1,73 @@
...
+				var parentFile = rp.GetParentFile (inputFile);
+				if (!string.IsNullOrEmpty (parentFile)) {
+					return parentFile;
+				}
+			}
+
+			return null;
+		}
+	}
+}
```
mrward: This looks like an unusual pattern. The existing code either calls AddinManager.GetExtensionObjects<T> so the tree is re-created every time, or it uses AddinManager.AddExtensionNodeHandler. No code uses both.

If you call AddinManager.AddExtensionNodeHandler then this will be called [multiple times initially with all addins](https://github.com/mono/mono-addins/blob/a543fa0011098232041c99bc5acb6416cb4acbe0/Mono.Addins/Mono.Addins/ExtensionNode.cs#L170).

So I think we should one of these approaches. Since this is going to be called a lot I think it makes sense to catch the rule providers and handle changes using just AddinManager.AddExtensionNodeHandler.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
```diff
@@ -0,0 +1,148 @@
...
+							return parentFile;
+					}
+				}
+				break;
+
+			case NestingRuleKind.FileSuffixToExtension:
+				string inputFileString = inputFile;
+				if (inputFileString.EndsWith (AppliesTo, StringComparison.OrdinalIgnoreCase)) {
+					int suffixPosition = inputFileString.LastIndexOf (AppliesTo, StringComparison.OrdinalIgnoreCase);
+					foreach (var pt in patterns) {
```
mrward: If inputFileFile ends with AppliesTo it seems that we could find the index by substracting the length of AppliesTo from inputFileString instead of using LastIndexOf here.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		public NestingRuleKind Kind { get; private set; }
+
+		public string AppliesTo { get; private set; }
+
+		bool CheckParentForFile (FilePath inputFile, FilePath parentFile)
+		{
+			if (File.Exists (parentFile) && inputFile != parentFile) {
+				LoggingService.LogInfo ($"Applied rule for nesting {inputFile} under {parentFile}");
```
mrward: I am just wondering if there is a better way of doing this which avoids using File.Exists. Should we instead be using the Project.Files information? That would avoid using File.Exists. I am assuming that CheckParentForFile will be called after the Project.Files collection has been populated.

Do we want file nesting to apply to files that are not explicitly in the project? Not sure it has any affect in practice - unless if we enable show all files would files be nested then?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
```diff
@@ -0,0 +1,156 @@
...
+				throw new ArgumentNullException (nameof (fromFile));
+			SourceFile = fromFile;
+		}
+
+		public string SourceFile { get; protected set; }
+
+		void AddRule (NestingRuleKind kind, string appliesTo, IEnumerable<string> patterns)
+		{
+			if (nestingRules == null) {
+				nestingRules = new List<NestingRule> ();
```
mrward: Minor: I guess this check is not needed now. nestingRules is initialized in GetParentFile now.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
rodrmoya: Yes, I think the best would be to pass the `Project` and check the `Files` collection indeed.

rodrmoya: Fixed in https://github.com/mono/monodevelop/pull/7682/commits/cf69e5a72e781f80565958a5d886d8fdaef1c19a

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
rodrmoya: Fixed in https://github.com/mono/monodevelop/pull/7682/commits/2f22de536e952a2dff9f67b055fe288b29776b36

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
rodrmoya: https://github.com/mono/monodevelop/pull/7682/commits/f0b97debe27c97833fe9522ec7f9205a8d365bfb

mrward (approved):
Code looks good to me.

Therzok:
We should've added a performance test for this before merging?

PR #7683, opened 2019-05-27
Launch settings fix

jtorres:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/862782
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/918958
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/933588
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/933582
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/934345
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/934367
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/934322
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/918985
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/755485
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/958219

Specs https://microsoft-my.sharepoint.com/:w:/p/mhutch/EfsOmr-aDnJMkI1Dv1QQIrcBd-kon2e8qB9QicNVkq3E-A?e=mYbAcu

Mote: rebased

User Cases:

given a ASP.NET Core project 

- when a new project is created then it should contain a default launchSettings.json profile called like project name and it should be a Default run config
- when a new Run Config is added then it should update launchSettings.json and add a new profile
- when a Run Config is removed then it should update launchSettings.json and remove that profile
- when a run config is modified the it should update launchSettings.json and the profile updated
- when a profile is added in the launchSettings.json then it should add a new Run Config
- when a profile is removed in the launchSettings.json then it should remove that Run Config
- when a profile is updated in the launchSettings.json then it should update the Run Config
- when a project with more than one profiles in the launchSettings.json is loaded then it should create a new Run Config for each profile which type == "Project"

### DIFF
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
@@ -43,18 +43,17 @@ namespace MonoDevelop.AspNetCore
 
 		bool updating;
 		Dictionary<string, AspNetCoreRunConfiguration> aspNetCoreRunConfs = new Dictionary<string, AspNetCoreRunConfiguration> ();
-		readonly object profiles = new object ();
 		LaunchProfileProvider launchProfileProvider;
 
 		protected override ProjectRunConfiguration OnCreateRunConfiguration (string name)
 		{
 			InitLaunchSettingsProvider ();
 
-			if (aspNetCoreRunConfs.ContainsKey (name))
-				return aspNetCoreRunConfs [name];
+			if (aspNetCoreRunConfs.TryGetValue (name, out var aspNetCoreRunConfiguration))
+				return aspNetCoreRunConfiguration;
 
 			var profile = new LaunchProfileData ();
-			if (!launchProfileProvider.Profiles.ContainsKey (name)) {
+			if (!launchProfileProvider.Profiles.TryGetValue (name, out var _)) {
 				if (name == "Default") {
 					profile = launchProfileProvider.AddNewProfile (Project.DefaultNamespace);
 					launchProfileProvider.Profiles [Project.DefaultNamespace] = profile;
@@ -234,64 +233,64 @@ namespace MonoDevelop.AspNetCore
 			if (launchSettings == null)
 				return;
 
-			lock (profiles) {
-				updating = true;
+			updating = true;
 
-				launchProfileProvider.LoadLaunchSettings ();
+			launchProfileProvider.LoadLaunchSettings ();
 
-				foreach (var profile in launchProfileProvider.Profiles) {
-					if (profile.Value.CommandName != "Project")
-						continue;
+			foreach (var profile in launchProfileProvider.Profiles) {
+				if (profile.Value.CommandName != "Project")
+					continue;
 
-					var key = string.Empty;
+				var key = string.Empty;
 
-					if (profile.Key == this.Project.DefaultNamespace) {
-						key = "Default";
-					} else {
-						key = profile.Key;
-					}
-
-					var runConfig = this.Project.RunConfigurations.FirstOrDefault (x => x.Name == key);
-					if (runConfig == null) {
-						var projectRunConfiguration = new AspNetCoreRunConfiguration (key, profile.Value) {
-							StartAction = "Project"
-						};
-						this.Project.RunConfigurations.Add (projectRunConfiguration);
-					} else {
-                        if (runConfig is AspNetCoreRunConfiguration aspNetCoreRunConfiguration) {
-							var index = Project.RunConfigurations.IndexOf (runConfig);
-							aspNetCoreRunConfiguration.CurrentProfile = profile.Value;
-							this.Project.RunConfigurations [index] = runConfig;
-						}
-					}
+				if (profile.Key == this.Project.DefaultNamespace) {
+					key = "Default";
+				} else {
+					key = profile.Key;
 				}
 
-				var itemsRemoved = new RunConfigurationCollection ();
-
-				foreach (var config in Project.RunConfigurations) {
-					var key = config.Name;
-
-					if (config.Name == "Default") {
-						key = this.Project.DefaultNamespace;
+				var runConfig = this.Project.RunConfigurations.FirstOrDefault (x => x.Name == key);
+				if (runConfig == null) {
+					var projectRunConfiguration = new AspNetCoreRunConfiguration (key, profile.Value) {
+						StartAction = "Project"
+					};
+					this.Project.RunConfigurations.Add (projectRunConfiguration);
+				} else {
+					if (runConfig is AspNetCoreRunConfiguration aspNetCoreRunConfiguration) {
+						var index = Project.RunConfigurations.IndexOf (runConfig);
+						aspNetCoreRunConfiguration.CurrentProfile = profile.Value;
+						this.Project.RunConfigurations [index] = runConfig;
 					}
-
-					if (launchProfileProvider.Profiles.ContainsKey (key))
-						continue;
-
-					itemsRemoved.Add (config);
 				}
-
-				Project.RunConfigurations.RemoveRange (itemsRemoved);
-
-				updating = false;
 			}
+
+			var itemsRemoved = new RunConfigurationCollection ();
+
+			foreach (var config in Project.RunConfigurations) {
+				var key = config.Name;
+
+				if (config.Name == "Default") {
+					key = this.Project.DefaultNamespace;
+				}
+
+				if (launchProfileProvider.Profiles.TryGetValue (key, out var _))
+					continue;
+
+				itemsRemoved.Add (config);
+			}
+
+			Project.RunConfigurations.RemoveRange (itemsRemoved);
+
+			updating = false;
 		}
 
-		protected override void OnRemoveRunConfiguration (string name)
+		protected override void OnRemoveRunConfiguration (IEnumerable<SolutionItemRunConfiguration> items)
 		{
 			if (updating || launchProfileProvider == null)
 				return;
-			launchProfileProvider.Profiles.TryRemove (name, out var _);
+			foreach (var item in items) {
+				launchProfileProvider.Profiles.TryRemove (item.Name, out var _);
+			}
 			launchProfileProvider.SaveLaunchSettings ();
 		}
 	}
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
@@ -53,16 +53,14 @@ namespace MonoDevelop.AspNetCore
 				return aspNetCoreRunConfiguration;
 
 			var profile = new LaunchProfileData ();
-			if (!launchProfileProvider.Profiles.TryGetValue (name, out var _)) {
-				if (name == "Default") {
-					profile = launchProfileProvider.AddNewProfile (Project.DefaultNamespace);
-					launchProfileProvider.Profiles [Project.DefaultNamespace] = profile;
-				} else {
-					profile = launchProfileProvider.AddNewProfile (name);
-					launchProfileProvider.Profiles [name] = profile;
-				}
+
+			var key = name != "Default" ? name : this.Project.DefaultNamespace;
+
+			if (!launchProfileProvider.Profiles.TryGetValue (key, out var _)) {
+				profile = launchProfileProvider.AddNewProfile (key);
+				launchProfileProvider.Profiles [key] = profile;
 			} else {
-				profile = launchProfileProvider.Profiles [name];
+				profile = launchProfileProvider.Profiles [key];
 			}
 
 			var aspnetconf = new AspNetCoreRunConfiguration (name, profile);
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
@@ -35,19 +35,55 @@ using MonoDevelop.Ide;
 using MonoDevelop.Projects;
 
 namespace MonoDevelop.AspNetCore
-{
+{   
 	[ExportProjectModelExtension]
 	class AspNetCoreProjectExtension : DotNetCoreProjectExtension
 	{
-		AspNetCoreRunConfiguration aspNetCoreRunConf;
 		public const string TypeScriptCompile = "TypeScriptCompile";
 
+		bool updating;
+		readonly Dictionary<string, AspNetCoreRunConfiguration> aspNetCoreRunConfs = new Dictionary<string, AspNetCoreRunConfiguration> ();
+		readonly object profiles = new object ();
+        LaunchProfileProvider launchProfileProvider;
+         
 		protected override ProjectRunConfiguration OnCreateRunConfiguration (string name)
 		{
-			if (aspNetCoreRunConf == null) {
-				aspNetCoreRunConf = new AspNetCoreRunConfiguration (name, this.Project);
+			InitLaunchSettingsProvider ();
+
+			if (aspNetCoreRunConfs.ContainsKey (name))
+				return aspNetCoreRunConfs [name];
+
+			var profile = new LaunchProfileData ();
+			if (!launchProfileProvider.Profiles.ContainsKey (name)) {
+				if (name == "Default") {
+					profile = launchProfileProvider.AddNewProfile (Project.DefaultNamespace);
+					launchProfileProvider.Profiles [Project.DefaultNamespace] = profile;
+				} else {
+					profile = launchProfileProvider.AddNewProfile (name);
+					launchProfileProvider.Profiles [name] = profile;
+				}
+			} else {
+				profile = launchProfileProvider.Profiles [name];
 			}
-			return aspNetCoreRunConf;
+
+			var aspnetconf = new AspNetCoreRunConfiguration (name, profile);
+            aspnetconf.SaveRequested += Aspnetconf_Save;
+			aspNetCoreRunConfs.Add (name, aspnetconf);
+			return aspnetconf;
+		}
+
+		void Aspnetconf_Save (object sender, EventArgs e)
+		{
+            if (sender is AspNetCoreRunConfiguration config && config.IsDirty)
+			    launchProfileProvider.SaveLaunchSettings ();
+		}
+
+        void InitLaunchSettingsProvider ()
+		{
+			if (launchProfileProvider == null) {
+				launchProfileProvider = new LaunchProfileProvider (this.Project.BaseDirectory, this.Project.DefaultNamespace);
+				launchProfileProvider.LoadLaunchSettings ();
+			}
 		}
 
 		protected override bool SupportsObject (WorkspaceObject item)
@@ -76,8 +112,7 @@ namespace MonoDevelop.AspNetCore
 		private ExecutionCommand CreateAspNetCoreExecutionCommand (ConfigurationSelector configSel, DotNetProjectConfiguration configuration, ProjectRunConfiguration runConfiguration)
 		{
 			FilePath outputFileName;
-			var aspnetCoreRunConfiguration = runConfiguration as AspNetCoreRunConfiguration;
-			if (aspnetCoreRunConfiguration == null)
+			if (!(runConfiguration is AspNetCoreRunConfiguration aspnetCoreRunConfiguration))
 				return null;
 			if (aspnetCoreRunConfiguration.StartAction == AssemblyRunConfiguration.StartActions.Program)
 				outputFileName = aspnetCoreRunConfiguration.StartProgram;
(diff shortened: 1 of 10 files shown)


main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -116,27 +152,109 @@ protected override string OnGetDefaultBuildAction (string fileName)
...
+					var runConfig = this.Project.RunConfigurations.FirstOrDefault (x => x.Name == key);
+					if (runConfig == null) {
+						var projectRunConfiguration = new AspNetCoreRunConfiguration (key, profile.Value) {
+							StartAction = "Project"
+						};
+						this.Project.RunConfigurations.Add (projectRunConfiguration);
+					} else {
+						var index = Project.RunConfigurations.IndexOf (runConfig);
+						this.Project.RunConfigurations [index] = runConfig;
+					}
```
rodrmoya: hmm, not sure what this does, it seems to just set to the item that was retrieved in line 219 again, by retrieving its index and setting it by that index??

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/ILaunchProfile.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace MonoDevelop.AspNetCore
 {
-	interface ILaunchProfile
+	public interface ILaunchProfile
 	{
```
rodrmoya: Does this need to be public?

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/LaunchProfileData.cs
```diff
@@ -10,9 +10,8 @@
 
 namespace MonoDevelop.AspNetCore
 {
-
 	[JsonObject (MemberSerialization.OptIn)]
-	internal class LaunchProfileData
+	public class LaunchProfileData
 	{
```
rodrmoya: Same here, does it need to be public?

jtorres: yes, `LaunchProfileData` is used by AspnNetCoreConfiguration which is public

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/ILaunchProfile.cs
jtorres: same, since LaunchProfileData is public, it should be public

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
jtorres: right somehow I missed a step before `this.Project.RunConfigurations [index] = runConfig` that copies `profile` values to `runconfig`... will fix

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -116,27 +152,109 @@ protected override string OnGetDefaultBuildAction (string fileName)
...
+					this.Project.RunConfigurations.Add (projectRunConfiguration);
+				}
+			}
 		}
 
 		public override void Dispose ()
 		{
 			base.Dispose ();
 			FileService.FileChanged -= FileService_FileChanged;
+			foreach (var conf in aspNetCoreRunConfs) {
```
Therzok: Maybe also clear/null out the dictionary here.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreRunConfiguration.cs
```diff
@@ -216,5 +177,7 @@ internal bool UsingHttps ()
 
 			return false;
 		}
+
+		public void OnSaveRequested () => SaveRequested?.Invoke (this, EventArgs.Empty);
```
Therzok: Does this have to be public?

Therzok:
There's a lot of mixed whitespace here.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/LaunchProfileProvider.cs
```diff
@@ -24,37 +24,51 @@
...
+		internal string LaunchSettingsJsonPath => Path.Combine (baseDirectory, "Properties", "launchSettings.json");
 		const string DefaultGlobalSettings = @"{
     						""windowsAuthentication"": false,
     						""anonymousAuthentication"": true
   							}";
 
-		public LaunchProfileProvider (Project project)
+		public LaunchProfileData DefaultProfile {
+			get {
+				if (!Profiles.ContainsKey (defaultNamespace)) {
```
mrward: TryGet?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -3060,7 +3060,8 @@ internal void OnRunConfigurationsAdded (IEnumerable<SolutionItemRunConfiguration
 
 		internal void OnRunConfigurationRemoved (IEnumerable<SolutionItemRunConfiguration> items)
 		{
-
+			foreach (var s in items)
+				ProjectExtension.OnRemoveRunConfiguration (s.Name);
 		}
```
mrward: I think we should follow the same pattern as for NotifyItemsRemoved.

Project.NotifyItemsRemoved -> ProjectExtension.OnItemsRemoved -> Project.OnItemsRemoved

Also I think it should pass the full set of items instead of the name of each one individually.

We could also implement the same pattern when a run config is added whilst we are added the pattern for removing.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -116,27 +152,109 @@ protected override string OnGetDefaultBuildAction (string fileName)
...
 		void FileService_FileChanged (object sender, FileEventArgs e)
 		{
-			var launchSettingsPath = aspNetCoreRunConf?.launchProfileProvider?.launchSettingsJsonPath;
+			var launchSettingsPath = launchProfileProvider?.LaunchSettingsJsonPath;
 			var launchSettings = e.FirstOrDefault (x => x.FileName == launchSettingsPath && !x.FileName.IsDirectory);
 			if (launchSettings == null)
 				return;
+			
+			lock (profiles) {
+				updating = true;
```
mrward: FileService_FileChanged I think runs on the UI thread so we do not need a lock here. A lock would make sense if we were running this on another background thread.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -37,15 +38,51 @@ namespace MonoDevelop.AspNetCore
...
 		{
-			if (aspNetCoreRunConf == null) {
-				aspNetCoreRunConf = new AspNetCoreRunConfiguration (name, this.Project);
+			InitLaunchSettingsProvider ();
+
+			if (aspNetCoreRunConfs.ContainsKey (name))
+				return aspNetCoreRunConfs [name];
+
+			var profile = new LaunchProfileData ();
+			if (!launchProfileProvider.Profiles.ContainsKey (name)) {
```
Therzok: TryGetValue here?

rodrmoya:
I wonder if we could easily have a high level test, that created a project, then a couple run configs and built on both configs and checking some condition?

rodrmoya (approved):

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -121,22 +154,111 @@ protected override void OnItemReady ()
...
+						StartAction = "Project"
+					};
+					this.Project.RunConfigurations.Add (projectRunConfiguration);
+				} else {
+					if (runConfig is AspNetCoreRunConfiguration aspNetCoreRunConfiguration) {
+						var index = Project.RunConfigurations.IndexOf (runConfig);
+						aspNetCoreRunConfiguration.UpdateProfile (profile.Value);
+						this.Project.RunConfigurations [index] = runConfig;
+					}
+				}
```
mrward: This code is almost, but not quite, identical to similar code in OnItemReady - can this code be shared instead of duplicating it here?

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/ILaunchProfile.cs
mrward: Asking the same question as Rodrigo here - does this and the LaunchProfileData need to be public. Whilst AspNetCoreConfiguration is public the LaunchProfileData is not public on AspNetCoreConfiguration. It seems like we could make LaunchProfileData and ILaunchProfile non-public.

mrward (commented):

Just two minor questions about public classes and some possible code duplication.

mrward:
Adding or removing a profile from launchSettings.json - the main toolbar does not update with this new run configuration. If you add/remove a run config from project options the main toolbar updates.

Adding a new user specific run config adds an entry into the launchSettings.json - not sure what the expected behaviour here. Not sure I understand the mappings here and how things should be synchronised. The spec mentions ignoring user specific run configurations - but not sure if that is just on load or all the time.

jtorres:
I need to run a smoke test after lasts commits

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/ILaunchProfile.cs
jtorres: ok, set LaunchProfileData to internal

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -121,22 +154,113 @@ protected override void OnItemReady ()
...
 		}
 
 		public override void Dispose ()
 		{
 			base.Dispose ();
 			FileService.FileChanged -= FileService_FileChanged;
+			foreach (var conf in aspNetCoreRunConfs) {
+				conf.Value.SaveRequested -= Aspnetconf_Save;
+			}
+			aspNetCoreRunConfs.Clear ();
```
Therzok: No need to clear if we set it to null.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -121,22 +154,113 @@ protected override void OnItemReady ()
...
+
+			updating = false;
+		}
+
+		protected override void OnRemoveRunConfiguration (IEnumerable<SolutionItemRunConfiguration> items)
+		{
+			if (updating || launchProfileProvider == null)
+				return;
+			foreach (var item in items) {
+				launchProfileProvider.Profiles.TryRemove (item.Name, out var _);
```
Therzok:
```suggestion
				launchProfileProvider.Profiles.TryRemove (item.Name, out _);
```

Therzok (commented):
LGTM!

slluis (changes requested):
We need to document test scenarios and use cases.

mrward:
- Could not test everything since reloading the solution with new run configs throws the null reference exception that Sayed reported.

- Seems to be some problems with shared versus user run configs. From the spec the user specific run configs should be ignored but when you create a user specific run config the launchsettings.json file is updated.

- A shared run configuration does not seem to be created for the Default launchSettings.json profile. From the spec I was assuming that would be the case - maybe that is not the case for Default.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -37,15 +38,47 @@ namespace MonoDevelop.AspNetCore
...
+			InitLaunchSettingsProvider ();
+
+			if (aspNetCoreRunConfs.TryGetValue (name, out var aspNetCoreRunConfiguration))
+				return aspNetCoreRunConfiguration;
+
+			var profile = new LaunchProfileData ();
+
+			var key = name != "Default" ? name : this.Project.DefaultNamespace;
+
+			if (!launchProfileProvider.Profiles.TryGetValue (key, out var _)) {
```
Therzok:
```suggestion
			if (!launchProfileProvider.Profiles.TryGetValue (key, out profile)) {
```

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -37,15 +38,47 @@ namespace MonoDevelop.AspNetCore
...
+				profile = launchProfileProvider.AddNewProfile (key);
+				launchProfileProvider.Profiles [key] = profile;
+			} else {
+				profile = launchProfileProvider.Profiles [key];
```
Therzok: Since we avoided double lookup, we can remove this block.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/LaunchProfileProvider.cs
```diff
@@ -36,27 +38,42 @@ namespace MonoDevelop.AspNetCore
...
+					var defaultProfile = CreateDefaultProfile ();
```
Therzok:
```csharp
return Profiles [defaultNamespace] = CreateDefaultProfile ();
```

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/LaunchProfileProvider.cs
```diff
@@ -36,27 +38,42 @@ namespace MonoDevelop.AspNetCore
...
 	internal class LaunchProfileProvider
 	{
-		DotNetProject project;
+		readonly string baseDirectory;
+		readonly string defaultNamespace;
 		readonly object fileLocker = new object ();
 		public IDictionary<string, JToken> GlobalSettings { get; private set; }
 		internal JObject ProfilesObject { get; private set; }
-		internal string launchSettingsJsonPath => Path.Combine (project.BaseDirectory, "Properties", "launchSettings.json");
+		public ConcurrentDictionary<string, LaunchProfileData> Profiles { get; set; }
```
Therzok: Why can we publicly set the Profiles dictionary? Should this be readonly?
```suggestion
		public ConcurrentDictionary<string, LaunchProfileData> Profiles { get; }
```

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/LaunchProfileProvider.cs
```diff
@@ -69,25 +86,25 @@ public void LoadLaunchSettings ()
...
 		}
 
 		JObject TryParse ()
 		{
 			try {
-				return JObject.Parse (File.ReadAllText (launchSettingsJsonPath));
-			}
-			catch {
+				return JObject.Parse (File.ReadAllText (LaunchSettingsJsonPath));
+			} catch {
```
Therzok:
```csharp
try {
        using var fs = File.OpenRead (LaunchSettingsJsonPath);
	using var sr = new StreamReader (fs);
	using var reader = new JsonTextReader (sr);
	return JObject.Load (reader);
} catch (JsonReaderException e) {
	LoggingService.LogInternalError ($"Failed to parse Json {LaunchSettingsJsonPath}", e);
}
```

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/LaunchProfileProvider.cs
```diff
@@ -120,10 +137,22 @@ void ReportError (string message, Exception ex)
 			MessageService.ShowError (message);
 		}
 
-		public LaunchProfileData CreateDefaultProfile ()
+		public LaunchProfileData AddNewProfile (string name)
```
Therzok: Should we do validation if the profile already exists?

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/LaunchProfileProvider.cs
```diff
@@ -120,10 +137,22 @@ void ReportError (string message, Exception ex)
...
+		{
+			if (Profiles == null)
+				Profiles = new ConcurrentDictionary<string, LaunchProfileData> ();
+
+			var newProfile = CreateProfile (name);
```
Therzok: Replace with:
```csharp
return Profiles [name] = CreateProfile (name); 
```

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/LaunchProfileProvider.cs
```diff
@@ -139,9 +168,62 @@ public LaunchProfileData CreateDefaultProfile ()
 		void CreateAndAddDefaultLaunchSettings ()
 		{
 			GlobalSettings.Add ("iisSettings", JToken.Parse (DefaultGlobalSettings));
-			var profilesData = new Dictionary<string, LaunchProfileData> ();
-			profilesData.Add (project.DefaultNamespace, CreateDefaultProfile ());
-			SaveLaunchSettings (profilesData.ToSerializableForm ());
+			var profiles = new Dictionary<string, LaunchProfileData> {
+				{ defaultNamespace, CreateDefaultProfile () }
+			};
+			Profiles = new ConcurrentDictionary<string, LaunchProfileData> (profiles);
```
Therzok: Remove the redundant intermediate dictionary, and set the value directly:
```csharp
			Profiles = new ConcurrentDictionary<string, LaunchProfileData>();
			Profiles.TryAdd (defaultNamespace, CreateDefaultProfile ());
```

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/LaunchProfileProvider.cs
```diff
@@ -139,9 +168,62 @@ public LaunchProfileData CreateDefaultProfile ()
...
+
+				var runConfig = project.RunConfigurations.FirstOrDefault (x => x.Name == key);
+				if (runConfig == null) {
+					var projectRunConfiguration = new AspNetCoreRunConfiguration (key, profile.Value) {
+						StartAction = "Project",
+						StoreInUserFile = false
+					};
+					project.RunConfigurations.Add (projectRunConfiguration);
+				} else if (runConfig is AspNetCoreRunConfiguration aspNetCoreRunConfiguration) {
+					var index = project.RunConfigurations.IndexOf (runConfig);
```
Therzok: Validate the return value is `>= 0` or Debug.Assert it if we expect it to always be found:
```csharp
if (index >= 0) ... // do stuff

// or

Debug.Assert (index >= 0, "Expected value to be found");
```

Therzok: I expect it's the latter. Maybe it would be worth refactoring so we get the index of the first run configuration matching the `Name`, and do the `>= 0` check there. I'm wary of race conditions.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/LaunchProfileProvider.cs
```diff
@@ -139,9 +168,62 @@ public LaunchProfileData CreateDefaultProfile ()
...
+			var itemsRemoved = new RunConfigurationCollection ();
+
+			foreach (var config in project.RunConfigurations) {
+				var key = config.Name;
+
+				if (config.Name == "Default") {
+					key = project.DefaultNamespace;
+				}
+
+				if (Profiles.TryGetValue (key, out var _))
```
Therzok:
```suggestion
				if (Profiles.TryGetValue (key, out _))
```

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -37,15 +38,47 @@ namespace MonoDevelop.AspNetCore
...
+			}
+
+			var aspnetconf = new AspNetCoreRunConfiguration (name, profile);
+			aspnetconf.SaveRequested += Aspnetconf_Save;
+			aspNetCoreRunConfs.Add (name, aspnetconf);
```
Therzok: Threading concerns? If the value was added already by another thread, we can race here.

Therzok: If we expect this to be called from the main thread only, add a `Runtime.AssertMainThread()` at the top of the method.

Therzok:
First review pass, mostly LGTM!

rodrmoya:
@monojenkins backport pr-8.3-launch-settings-fix

PR #7684, opened 2019-05-27
Fix document reuse logic

slluis:
Fixed the logic that determines if a document can be reused or not, it didn't work properly when opening a document using a custom model descriptor.

Also adjusted the logic for reusing file documents and added unit tests.

Fixes bug #891266 - Assembly browser back/forward navigation opens new Assembly Browser window


slluis:

PR #7685, opened 2019-05-27
[Watson] Fixes NRE when FindDockGroupItem doesn't find any item

netonjm:
This issue was detected by Watson and I can't reproduce it, but my feeling: this is it happening with dock panels with AllowDocking = false (like the Solution Explorer).

This call FindDockGroupItem will end in NRE (id not found) if we try to dock this panel at the same time we the system is Splitting a groupview or an item is Removed for some reason (the collection is cleaned on this cases). 

Fixes VSTS #861013 - [Watson] System.NullReferenceException in MonoDevelop.Components.Docking.DockContainer::DockInPlaceholder (MonoDevelop.Ide.dll)


netonjm:
@slluis added a try/catch block 👍

slluis:

PR #7695, opened 2019-05-28
[IDE] Overall energy consumption and UI hang improvements

Therzok:
There are various small things added up.

Most notably, the assembly browser is now more responsive and faster to load, and should no longer leak.

Other small improvements

### DIFF
--- a/main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserViewContent.cs
+++ b/main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserViewContent.cs
@@ -67,11 +67,13 @@ namespace MonoDevelop.AssemblyBrowser
 			FillWidget ();
 		}
 
-		protected override async Task OnInitialize (ModelDescriptor modelDescriptor, Properties status)
+		protected override Task OnInitialize (ModelDescriptor modelDescriptor, Properties status)
 		{
 			if (modelDescriptor is FileDescriptor fileDescriptor) {
 				Load (fileDescriptor.FilePath);
 			}
+
+			return Task.CompletedTask;
 		}
 
 		protected override bool OnTryReuseDocument (ModelDescriptor modelDescriptor)
@@ -92,9 +94,9 @@ namespace MonoDevelop.AssemblyBrowser
 		{
 			var loader = widget.AddReferenceByFileName (filePath);
 			if (loader != null) {
-				loader.LoadingTask.ContinueWith (delegate {
-					widget.SelectAssembly (loader);
-				});
+				loader.LoadingTask
+					.ContinueWith (t => widget.SelectAssembly (t.Result), Runtime.MainTaskScheduler)
+					.Ignore ();
 			}
 		}
 
@@ -105,12 +107,12 @@ namespace MonoDevelop.AssemblyBrowser
 
 		protected override void OnDispose ()
 		{
-			if (currentWs != null) 
+			if (currentWs != null) {
 				currentWs.WorkspaceLoaded -= Handle_WorkspaceLoaded;
+			}
 
 			widget = null;
-			if (Disposed != null)
-				Disposed (this, EventArgs.Empty);
+			Disposed?.Invoke (this, EventArgs.Empty);
 			base.OnDispose ();
 		}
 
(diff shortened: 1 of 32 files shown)


main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserViewContent.cs
```diff
@@ -169,40 +176,29 @@ public void FindDerivedClasses ()
...
-						LoggingService.LogError ("Error while adding project " + project.Name + " to the tree.", e);
-					}
-				}
-				await Task.WhenAll (allTasks).ContinueWith (delegate {
-					Runtime.RunInMainThread (delegate {
+				var token = cts.Token;
+
+				var workspace = IdeApp.TypeSystemService.GetWorkspaceAsync (IdeApp.ProjectOperations.CurrentSelectedSolution)
+					.ContinueWith (t => {
+						if (token.IsCancellationRequested)
```
mrward: Was wondering if this needed to handle the task exception but it looks like GetWorkspaceAsync would not cause an exception.

mrward (approved):
👍

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserViewContent.cs
Therzok: I had another small patch around this, but it's just a fake hang improvement.

sevoku (approved):

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserWidget.cs
```diff
@@ -914,24 +914,26 @@ void HPaneExpose (object sender, Gtk.ExposeEventArgs args)
...
-					try {
-						if (currentAssembly != null) {
-							OpenFromAssembly (url, currentAssembly);
-						} else {
-							OpenFromAssemblyNames (url);
-						}
-					} catch (Exception e) {
-						LoggingService.LogError ("Error while opening the assembly browser with id:" + url, e);
+				if (d.IsFaulted) {
+					LoggingService.LogError ("Failed to load assemblies", d.Exception);
```
mkrueger: Shouldn't be a return here ?

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/TreeNodes/Cecil/TypeDefinitionNodeBuilder.cs
```diff
@@ -98,33 +98,32 @@ static string GetSource (ITypeDefinition type)
...
 		{
 			var type = (ITypeDefinition)dataObject;
-			var list = new System.Collections.ArrayList ();
 			if (type.DirectBaseTypes.Any ())
-				list.Add (new BaseTypeFolder (type));
+				builder.AddChild (new BaseTypeFolder (type));
 			bool publicOnly = Widget.PublicApiOnly;
 
-			foreach (var field in type.Fields.OrderBy (m => m.Name, StringComparer.InvariantCulture)) {
+			foreach (var field in type.Fields) {
```
mkrueger: Shouldn't the names be ordered ?

mkrueger (approved):

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserWidget.cs
Therzok: definitions will be empty if it's faulted, so I guess I can do a return there too.

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/TreeNodes/Cecil/TypeDefinitionNodeBuilder.cs
Therzok: Yes, but the sorting is done at treenode level. We don't need to do it here.

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserWidget.cs
Therzok: I'm changing a few things here, to better handle faulting.

Therzok: I've thought this through.

There's a few scenarios which are handled by explicitly not returning on faulted.

a) IsFaulted is set when even just one task is faulted, but after all have run. What if we can actually find the type we're looking for? 
b) If the current assembly loader task fails, it'll be caught by the try/catch a bit further below.

Added a comment.

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/TreeNodes/Cecil/TypeDefinitionNodeBuilder.cs
Therzok: See the [comparer](https://github.com/mono/monodevelop/blob/5bbe3317491f6f588f5fb7938ed1c1ef8f1169f2/main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/TreeNodes/AssemblyBrowserTypeNodeBuilder.cs#L60)

Therzok:
@slluis PTAL

Therzok:
Rebased and fixed conflicts.

slluis (approved):

PR #7696, opened 2019-05-28
[Core] Improve UpdateChannel's GetHashCode()

bratsche:
Equals() and GetHashCode() are out of sync and not checking the same
fields.


main/src/core/MonoDevelop.Core/MonoDevelop.Core.Setup/UpdateChannel.cs
```diff
@@ -116,7 +116,7 @@ public bool Equals (UpdateChannel a)
 
 		public override int GetHashCode ()
 		{
-			return Idx;
+			return Id.GetHashCode() + Idx;
 		}
```
mrward: I think we should just be using Id for GetHashCode and Equals. Idx looks to be just used for ordering when sorting.

mrward: It seems unlikely we support the same Id multiple times but with a different Idx. The UpdateChannel Ids are supposed to be unique as far as I am aware.

bratsche: Thanks @mrward, updated.

mrward (approved):

Therzok (approved):

PR #7700, opened 2019-05-28
Fix some NullReferences in Json files

KirillOsenkov:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/861121 - "System.NullReferenceException" when closing a solution while a json file is open & the solution doesn't close.

This is the MonoDevelop side of the fix for 861121. For non-C# files they could still be a part of WebEditorRoslynWorkspace, and not MonoDevelopWorkspace, and we need to gracefully deal with that.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageService.cs
@@ -126,8 +126,15 @@ namespace MonoDevelop.Ide.WelcomePage
 				}
 				WelcomePageShown?.Invoke (welcomePage, EventArgs.Empty);
 				welcomePage.UpdateProjectBar ();
-				((DefaultWorkbench)IdeApp.Workbench.RootWindow).BottomBar.Visible = false;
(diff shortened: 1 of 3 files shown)


slluis:
@monojenkins backport master

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageService.cs
```diff
@@ -126,8 +126,15 @@ public static void ShowWelcomePage (bool animate = false)
...
+				
+				var rootWindow = (DefaultWorkbench)IdeApp.Workbench.RootWindow;
+				if (rootWindow.BottomBar is MonoDevelopStatusBar statusBar) {
+					statusBar.Visible = false;
+				}
+
+				if (rootWindow.DockFrame is Components.Docking.DockFrame dockFrame) {
+					dockFrame.AddOverlayWidget (welcomePage, animate);
+				}
 				welcomePage.GrabFocus ();
```
slluis: What are the above changes for? they look unrelated to the bug

KirillOsenkov: I basically fixed all the exceptions I saw when working with a .json file. With this PR there are no more exceptions along the codepaths I tested.

slluis:
@monojenkins backport master

PR #7701, opened 2019-05-29
Packaging: More thread-safety in NuGetPackageServicesProxy

sandyarmstrong:
Remove more unnecessary or problematic uses of Runtime.RunInMainThread,
again due to Roslyn blocking the main thread, and then calling services
like this from async tasks it also blocks on.

Interesting note: the async call in GetInstalledVersions is actually
entirely synchronous, but we cannot change all the methods involved
because they are implementing interfaces from NuGet. So running it on
the threadpool is totally unnecessary, but protects us in case things
change in the future.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/900631


sandyarmstrong:
Will backport to `release-8.1` after this passes review.

sandyarmstrong:

PR #7707, opened 2019-05-29
Fix 895583: Breakpoints not visible after renaming file

DavidKarlas:
This is two part fix, part is in DebuggingService which is updating actual breakpoint to new file
And other part is in new editor breakpoint integration which now doesn't use `readonly string file` but instead `textDocument.FilePath` which is mutable and updates when file is renamed and also listens for additional event.


slluis:
Also, the debugger changes is missing unit tests.

DavidKarlas:
@slluis https://github.com/mono/debugger-libs/pull/226 addresses requested changes. I think this can be merged regardless.

PR #7708, opened 2019-05-29
[Ide] Added a disposed token to the document controller API.

mkrueger:
The disposed token can be used by continue with calls that shouldn't
continue on disposed instances.

### DIFF
--- a/main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewContent.cs
+++ b/main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewContent.cs
@@ -1,4 +1,4 @@
-//
+﻿//
 // Copyright (c) Microsoft Corp. (https://www.microsoft.com)
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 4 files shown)


mkrueger:
Set the DisposedToken in constructor, because the token source gets set to null in Dispose.

Therzok:
I'm okay with having disposed surfaced as protected.

sevoku:
+1 for it to be protected, and maybe ad a `public bool IsDisposed`?

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewContent.cs
```diff
@@ -841,4 +838,3 @@ public void WriteTextTo (TextWriter writer, int offset, int length)
 				=> throw new NotImplementedException ();
 		}
 	}
```
Therzok:
```suggestion
	}
}
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
```diff
@@ -384,12 +384,26 @@ public abstract class DocumentController : IDisposable, ICommandDelegatorRouter
 			}
 		}
 
+		CancellationTokenSource disposedTokenSource = new CancellationTokenSource ();
+
+		/// <summary>
+		/// Returns a dispose token that's canceled when the document controller disposes.
+		/// </summary>
+		protected CancellationToken DisposedToken { get; private set; }
```
Therzok:
```suggestion
		protected CancellationToken DisposedToken { get; }
```

We never change it, so let's not set it anywhere else except the constructor.

Therzok:
@mkrueger can you fix the build?

PR #7727, opened 2019-05-30
Updated references to KirillOsenkov/vs-editor-api, xamarin/md-addins

xamarin-release-manager:
Updated references to KirillOsenkov/vs-editor-api, xamarin/md-addins


abock:
Pushing this through so the md-addins bump will build since the repos must be in sync.

PR #7734, opened 2019-05-30
[Mac] Use the main DispatchQueue when processing memory pressure events

Therzok:
Processing this on a background thread will cause an AppKit NSEventThread to spin up on a background thread, which would never finish running
This thread would also be running at 100% CPU, causing a specific thread to loop and burn until the process ends

Fixes VSTS #894581 - VS Mac processes still running after quitting the app?
Fixes VSTS #802626 - [Feedback] VS usually consumes an entire CPU core when idle, killing battery life (and making my lap hot)

### DIFF
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -1360,7 +1360,7 @@ namespace MonoDevelop.MacIntegration
 
 			public MacMemoryMonitor ()
 			{
-				DispatchSource = new DispatchSource.MemoryPressure (notificationFlags, DispatchQueue.DefaultGlobalQueue);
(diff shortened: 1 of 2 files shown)


Therzok:

slluis:

leculver:
@slluis @Therzok I suspect (but am not certain) that this causes our Unhealthy Session rate to go up.  The background process that monitors VS for Mac likely would get confused by the fact that VS for Mac is still running, and would appear hanging.  If we decide to ship another 8.0.*, I would suggest including this change as well.  (I wouldn't ship a new 8.0 just for this change, but opportunistically fixing this problem would be a good idea, IMO.)

migueldeicaza:
The debugging that lead to this patch must have been epic!

PR #7694, opened 2019-05-28
[NuGet] Update to NuGet 5.1

mrward:
Update to NuGet 5.1.0.6013

Fixes VSTS #899305 - Bump to NuGet 5.1

PR #7738, opened 2019-05-30
Revert "[NuGet] Update to NuGet 5.1"

Therzok:
Reverts mono/monodevelop#7694

GitHub page didn't show the label for `do-not-merge`


mrward:
I think we should be OK with leaving NuGet 5.1 merged in master since it does not break anything. NuGet 5.1 is planned for VS Mac 8.2, along with Mono 6.0 being updated to use NuGet 5.1.

PR #7742, opened 2019-05-30
[Mac] ObjCMarshalledException is still not reported right

Therzok:
The app crashes before the handler is reached, fix this by reporting it as fatal - usually the app really just does crash.


Therzok:

Therzok:
Telemetry monitor is still not picking it up.

Therzok:
Fixed!

slluis:

PR #7744, opened 2019-05-30
Navigation: Fix activation with new editor

sandyarmstrong:
Intended to fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/901424.

This isn't actually working at all. Thought I'd take a few minutes and try to fix quickly but clearly I'm still missing something. This _does_ ultimately call `GrabFocus` on the managed gtk host (and it's on the main thread when that call happens) so I'm not sure what's wrong.

Also, I could have sworn that this fixed https://devdiv.visualstudio.com/DevDiv/_workitems/edit/841029, but now I can't repro the fix.

So basically this does nothing but I can't spend more time on it right now.


sandyarmstrong:
Fixed properly in https://github.com/mono/monodevelop/pull/7791

PR #7748, opened 2019-05-31
[IDE] only popup the menu on single clicks.

iainx:
The way Gtk treats double/triple clicks is breaking something in Xwt and crashing when triple clicking on the language
button

Fixes VSTS #849556

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
@@ -179,7 +179,9 @@ namespace MonoDevelop.Ide.Projects
 				return;
 			}
 
-			if (languageCellRenderer.IsLanguageButtonPressed (args.Event)) {
+			// Only display the popup menu on a single press, ignore anything else
+			// Fixes a crash when triple clicking. VSTS #849556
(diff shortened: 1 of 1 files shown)


iainx:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -177,9 +177,11 @@ void TemplatesTreeViewButtonPressed (object o, ButtonPressEventArgs args)
...
 				return;
-			}
-
-			if (languageCellRenderer.IsLanguageButtonPressed (args.Event)) {
+			}
+
+			// Only display the popup menu on a single press, ignore anything else
+			// Fixes a crash when triple clicking. VSTS #849556
+			if (args.Event.Type == Gdk.EventType.ButtonPress && languageCellRenderer.IsLanguageButtonPressed (args.Event)) {
 				HandlePopup (template, args.Event.Time);
```
mrward: Minor: Could have the EventType check in the IsLanguageButtonPressed method.

iainx: Well, my thinking was that if it's not a ButtonPress event then we don't even care if it's in the language button, and save a method call. I know, it's hardly a performance bottleneck :)

PR #7756, opened 2019-05-31
Fix Application Output hang when using ASP.Net Core Angular template

iantoalms:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/893213

Added method in the LogView to sanitize the logging text output from the node.js service for the Angular/Bootstrap functionality in the ASP.Net Core Angular template.

The logging text contained ANSI/VT100 color escape sequences and some control characters, which were causing the pango text engine to slow down significantly in order to create custom glyphs with the control codes.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs
@@ -59,6 +59,7 @@ namespace MonoDevelop.Ide.Gui.Components
 		QueuedTextWrite lastTextWrite;
 		GLib.TimeoutHandler outputDispatcher;
 		bool outputDispatcherRunning = false;
+		readonly Regex consoleTextSanitizerRegex = new Regex ("(\\e[[0-9]*m)|(\b)");
 
 		const int MAX_BUFFER_LENGTH = 4000 * 1024;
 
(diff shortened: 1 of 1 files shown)


rodrmoya:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs
```diff
@@ -693,6 +696,23 @@ bool ShouldAutoScroll ()
...
+		///
+		/// Specifically we remove color escape sequences (\e[ + color code + m) and the backspace (\b) control
+		/// character.
+		/// 
+		/// </summary>
+		/// <param name="text"></param>
+		/// <returns>original text with control characters and escape sequences removed</returns>
+		string SanitizeConsoleText(string text)
+		{
+			return Regex.Replace (text, "(\\e[[0-9]*m)|(\b)", string.Empty);
```
Therzok: Can we compile this regex into a static instance?

Therzok: I assume this is going to be used a lot, so having a compiled regex would result in faster processing.

iantoalms: Good idea. Done.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs
```diff
@@ -58,6 +59,7 @@ public class LogView : Gtk.VBox
 		QueuedTextWrite lastTextWrite;
 		GLib.TimeoutHandler outputDispatcher;
 		bool outputDispatcherRunning = false;
+		readonly Regex consoleTextSanitizerRegex = new Regex ("(\\e[[0-9]*m)|(\b)");
```
Therzok:
```suggestion
		readonly Regex consoleTextSanitizerRegex = new Regex ("(\\e[[0-9]*m)|(\b)", RegexOptions.Compiled);
```

rodrmoya:

PR #7762, opened 2019-05-31
Fixes resize dock containers in our dock system moving  SplitterWidget to native

netonjm:
This PR fixes the resize behaviour in dock containers when mouse is over native embed areas

To solve this, we had to deal with some problems:
- Create a MacSplitterWidget (because we need this view above the GTK layer to capture some mouse over events),
- Doesn't break GTK
- Capture and not let GTK swallow the mouse events 
- Introduces SplitterGtkNSViewHostWidget based in a new GtkNSViewHostWidget, this layer encapsules  the logic to handle drag logic between toolkits

In order not to break GTK, I tried first to encapsule both backends and expose it over a wrapper class  SplitterWidgetWrapper with same interface.

In case of the cocoa backend, I created a native view (MacSplitterWidget) which GTK embeds into a SplitterGtkNSViewHostWidget extending from GtkNSViewHostWidget. This host widget has a DragCaptureEventMonitor with some logic to capture cocoa events and not let GTK swallow the mouse events.

scroll

Fixes VSTS #790689 - Pads can only be resized if mousing left to right over edge

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockContainer.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockContainer.cs
@@ -27,8 +27,6 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
-
-
 using System;
 using System.Collections.Generic;
 using Gtk;
@@ -37,10 +35,11 @@ using System.Linq;
 using MonoDevelop.Components.AtkCocoaHelper;
 using MonoDevelop.Core;
 using MonoDevelop.Ide.Gui;
+using GLib;
 
 namespace MonoDevelop.Components.Docking
 {
-	class DockContainer: Container
+	class DockContainer : Container
 	{
 		DockLayout layout;
 		DockFrame frame;
@@ -48,7 +47,7 @@ namespace MonoDevelop.Components.Docking
 		List<TabStrip> notebooks = new List<TabStrip> ();
 		List<DockItem> items = new List<DockItem> ();
 
-		List<SplitterWidget> splitters = new List<SplitterWidget> ();
+		List<SplitterWidgetWrapper> splitters = new List<SplitterWidgetWrapper> ();
 
 		bool needsRelayout = true;
 
@@ -190,7 +189,7 @@ namespace MonoDevelop.Components.Docking
 			}
 			foreach (var s in splitters)
 				if (s.Parent != null)
-					callback (s);
+					callback (s.Widget);
 		}
 		
 		protected override bool OnExposeEvent (Gdk.EventExpose evnt)
@@ -288,7 +287,7 @@ namespace MonoDevelop.Components.Docking
 			for (int n=0; n < splitters.Count; n++) {
 				var s = splitters [n];
 				if (s.Parent != null)
-					Remove (s);
+					Remove (s.Widget);
 			}
 
 			// Hide the splitters that are not required
@@ -307,12 +306,19 @@ namespace MonoDevelop.Components.Docking
 					var s = splitters [n];
 					if (!s.Visible)
 						s.Show ();
-					Add (s);
+					Add (s.Widget);
 				} else {
-					var s = new SplitterWidget ();
+
+					SplitterWidgetWrapper s = null;
+#if MAC
+					var widget = new SplitterMacHostWidget ();
+#else
+					var widget = new SplitterWidget ();
+#endif
+					s = new SplitterWidgetWrapper (widget);
 					splitters.Add (s);
 					s.Show ();
-					Add (s);
+					Add (s.Widget);
 				}
 			}
 		}
@@ -494,7 +500,7 @@ namespace MonoDevelop.Components.Docking
 			}
 		}
 		
-		internal class SplitterWidget: EventBox
+		internal class SplitterWidget: EventBox, ISplitterWidget
 		{
 			static Gdk.Cursor hresizeCursor = new Gdk.Cursor (CursorType.SbHDoubleArrow);
 			static Gdk.Cursor vresizeCursor = new Gdk.Cursor (CursorType.SbVDoubleArrow);
@@ -505,7 +511,9 @@ namespace MonoDevelop.Components.Docking
 
 			DockGroup dockGroup;
 			int dockIndex;
-	
+
+			public Widget Widget => this;
+
 			public SplitterWidget ()
 			{
 				Accessible.SetRole (AtkCocoa.Roles.AXSplitter);
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
@@ -62,22 +62,25 @@ namespace MonoDevelop.Components.Docking
 		public override void MouseExited (NSEvent theEvent)
 		{
 			if (!dragging) {
-				NSCursor.ArrowCursor.Set ();
+				SetDefaultCursor ();
 			}
 			hover = false;
 			base.MouseExited (theEvent);
 		}
 
+		NSCursor currentCursor;
+
 		void SetResizeCursor ()
 		{
-			if (dockGroup == null) {
+			if (dockGroup == null || currentCursor != null) {
 				return;
 			}
 			if (dockGroup.Type == DockGroupType.Horizontal) {
-				NSCursor.ResizeLeftRightCursor.Set ();
+				currentCursor = NSCursor.ResizeLeftRightCursor;
 			} else {
-				NSCursor.ResizeUpDownCursor.Set ();
+				currentCursor = NSCursor.ResizeUpDownCursor;
 			}
+			currentCursor.Push ();
 		}
 
 		public override void MouseMoved (NSEvent theEvent)
@@ -120,12 +123,20 @@ namespace MonoDevelop.Components.Docking
 			AddRemoveFilter (true);
 		}
 
+		void SetDefaultCursor ()
+		{
+			if (currentCursor != null) {
+				currentCursor.Pop ();
+				currentCursor = null;
+			}
+		}
+
 		public override void MouseUp (NSEvent theEvent)
 		{
 			if (hover) {
 				SetResizeCursor ();
 			} else {
-				NSCursor.ArrowCursor.Set ();
+				SetDefaultCursor ();
 			}
 			RaiseEndDrag ();
 			base.MouseUp (theEvent);
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
@@ -48,8 +48,7 @@ namespace MonoDevelop.Components.Docking
 		{
 			var options = NSTrackingAreaOptions.ActiveInKeyWindow |
 				NSTrackingAreaOptions.InVisibleRect |
-				NSTrackingAreaOptions.MouseEnteredAndExited |
-				NSTrackingAreaOptions.EnabledDuringMouseDrag;
+				NSTrackingAreaOptions.MouseEnteredAndExited;
 			var trackingArea = new NSTrackingArea (default, options, this, null);
 			AddTrackingArea (trackingArea);
 		}
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
@@ -3,7 +3,6 @@
 // Author:
 //   Jose Medrano
 //
-
 //
 // Copyright (C) 2918 Microsoft
 //
@@ -31,7 +30,6 @@
 using System;
 using AppKit;
 using Gtk;
-using MonoDevelop.Components.Mac;
 
 namespace MonoDevelop.Components.Docking
 {
@@ -39,7 +37,6 @@ namespace MonoDevelop.Components.Docking
 	{
 		DockGroup dockGroup;
 		int dockIndex;
-		Widget parent;
 
 		int dragPos, dragSize;
 		bool hover, dragging;
@@ -164,8 +161,8 @@ namespace MonoDevelop.Components.Docking
 			else
 				Gdk.Window.RemoveFilterForAll (Filter);
 		}
-		static Gdk.FilterReturn Filter (IntPtr xevent, Gdk.Event evnt) => Gdk.FilterReturn.Remove;
 
+		static Gdk.FilterReturn Filter (IntPtr xevent, Gdk.Event evnt) => Gdk.FilterReturn.Remove;
 	}
 
 }
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
@@ -1,0 +1,174 @@
+﻿//
+//
+// Author:
+//   Jose Medrano
+//
+
+//
+// Copyright (C) 2918 Microsoft
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+#if MAC
+
+using System;
+using AppKit;
+using Gtk;
+using MonoDevelop.Components.Mac;
+
+namespace MonoDevelop.Components.Docking
+{
+	internal class MacSplitterWidget : NSView
+	{
+		DockGroup dockGroup;
+		int dockIndex;
+		Widget parent;
+
+		int dragPos, dragSize;
+		bool hover, dragging;
+
+		public MacSplitterWidget ()
+		{
+			var options = NSTrackingAreaOptions.ActiveInKeyWindow |
+				NSTrackingAreaOptions.InVisibleRect |
+				NSTrackingAreaOptions.MouseEnteredAndExited |
+				NSTrackingAreaOptions.EnabledDuringMouseDrag;
+			var trackingArea = new NSTrackingArea (default, options, this, null);
+			AddTrackingArea (trackingArea);
+		}
+
+		public override void MouseEntered (NSEvent theEvent)
+		{
+			if (!dragging) {
+				SetResizeCursor ();
+			}
+			hover = true;
+			base.MouseEntered (theEvent);
+		}
+
+		public override void MouseExited (NSEvent theEvent)
+		{
+			if (!dragging) {
+				NSCursor.ArrowCursor.Set ();
+			}
+			hover = false;
+			base.MouseExited (theEvent);
+		}
+
+		void SetResizeCursor ()
+		{
+			if (dockGroup == null) {
+				return;
+			}
+			if (dockGroup.Type == DockGroupType.Horizontal) {
+				NSCursor.ResizeLeftRightCursor.Set ();
+			} else {
+				NSCursor.ResizeUpDownCursor.Set ();
+			}
+		}
+
+		public override void MouseMoved (NSEvent theEvent)
+		{
+			if (!dragging) {
+				SetResizeCursor ();
+			}
+			base.MouseMoved (theEvent);
+		}
+
+		void RaiseEndDrag ()
+		{
+			if (dragging) {
+				dragging = false;
+				AddRemoveFilter (false);
+			}
+		}
+
+		public void Init (DockGroup grp, int index)
+		{
+			dockGroup = grp;
+			dockIndex = index;
+		}
+
+		public override void MouseDown (NSEvent theEvent)
+		{
+			dragging = true;
+
+			var point = NSEvent.CurrentMouseLocation;
+			var obj = dockGroup.VisibleObjects [dockIndex];
+
+			if (dockGroup.Type == DockGroupType.Horizontal) {
+				dragPos = (int)point.X;
+				dragSize = obj.Allocation.Width;
+			} else {
+				dragPos = (int)point.Y;
+				dragSize = obj.Allocation.Height;
+			}
+
+			AddRemoveFilter (true);
+		}
+
+		public override void MouseUp (NSEvent theEvent)
+		{
+			if (hover) {
+				SetResizeCursor ();
+			} else {
+				NSCursor.ArrowCursor.Set ();
+			}
+			RaiseEndDrag ();
+			base.MouseUp (theEvent);
+		}
+
+		public override void MouseDragged (NSEvent theEvent)
+		{
+			SetResizeCursor ();
+
+			var point = NSEvent.CurrentMouseLocation;
+			int newpos;
+			if (dockGroup.Type == DockGroupType.Horizontal) {
+				newpos = (int)point.X;
+			} else {
+				newpos = (int)point.Y;
+			}
+
+			if (newpos != dragPos) {
+				int nsize;
+				if (dockGroup.Type == DockGroupType.Horizontal) {
+					nsize = dragSize + (newpos - dragPos);
+				} else {
+					nsize = dragSize - (newpos - dragPos);
+				}
+				dockGroup.ResizeItem (dockIndex, nsize);
+			}
+		}
+
+		void AddRemoveFilter (bool enable)
+		{
+			if (enable)
+				Gdk.Window.AddFilterForAll (Filter);
+			else
+				Gdk.Window.RemoveFilterForAll (Filter);
+		}
+		static Gdk.FilterReturn Filter (IntPtr xevent, Gdk.Event evnt) => Gdk.FilterReturn.Remove;
+
+	}
+
+}
+
+#endif
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
@@ -43,7 +43,7 @@ namespace MonoDevelop.Components.Docking
 
 		public MacSplitterWidget ()
 		{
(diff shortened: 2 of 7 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockContainer.cs
```diff
@@ -307,12 +306,21 @@ void LayoutWidgets ()
...
 				} else {
-					var s = new SplitterWidget ();
+
+					SplitterWidgetWrapper s = null;
+#if MAC
+					var splitter = new MacSplitterWidget ();
+					var widget = Mac.GtkMacInterop.NSViewToGtkWidget<MacSplitterWidget, SplitterContainerWidget> (splitter);
+
+#else
+                     var widget = new SplitterWidget ();
```
Therzok: Indent is off.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
```diff
@@ -0,0 +1,181 @@
...
+		{
+			dockGroup = grp;
+			dockIndex = index;
+		}
+
+		public override void MouseDown (NSEvent theEvent)
+		{
+			dragging = true;
+
+			var point = NSEvent.CurrentMouseLocation;
```
Therzok: Shouldn't we use theEvent here?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
```diff
@@ -0,0 +1,181 @@
...
+				NSCursor.ResizeLeftRightCursor.Set ();
+			} else {
+				NSCursor.ArrowCursor.Set ();
+			}
+			RaiseEndDrag ();
+		}
+
+		public void MouseDrag (NSEvent theEvent)
+		{
+			var point = NSEvent.CurrentMouseLocation;
```
Therzok: theEvent here?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
```diff
@@ -0,0 +1,181 @@
...
+		}
+
+		NSTrackingArea trackingArea;
+		public override void UpdateTrackingAreas ()
+		{
+			if (trackingArea != null) {
+				RemoveTrackingArea (trackingArea);
+				trackingArea.Dispose ();
+			}
+			var options = NSTrackingAreaOptions.MouseMoved | NSTrackingAreaOptions.ActiveInKeyWindow | NSTrackingAreaOptions.MouseEnteredAndExited;
```
Therzok: Can we not use something like NSTrackingInVisibleRect?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
```diff
@@ -0,0 +1,181 @@
...
+			if (!dragging) {
+				SetDefaultCursor ();
+			}
+			hover = true;
+			base.MouseEntered (theEvent);
+		}
+
+		public override void MouseExited (NSEvent theEvent)
+		{
+			if (!dragging) {
```
Therzok: This can be a cursor tracking NSTrackingArea:
https://developer.apple.com/documentation/appkit/nstrackingareaoptions/nstrackingcursorupdate?language=objc

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/NativeContainerWidget.cs
```diff
@@ -0,0 +1,80 @@
...
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+#if MAC
+
+using System;
+
+namespace MonoDevelop.Components
+{
+	public class NativeContainerWidget : Gtk.Widget
```
Therzok: Can't we just use `Xwt.WrapWidget`?

http://source.monodevelop.com/#Xwt/Xwt/Toolkit.cs,a4d19a71b683bd89,references

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
netonjm: No because for some reason nsevent is not completely accurate in mouse position (not sure why), probably I could remove the parameter

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
netonjm: No because for some reason nsevent is not completely accurate in mouse position (not sure why), probably we could remove the parameter, and with this we ensure mouse position is correct.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/SplitterContainerWidget.cs
```diff
@@ -0,0 +1,105 @@
...
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+
+using System;
+using AppKit;
+using Gdk;
+
```
rodrmoya: shouldn't this have `#if MAC`?

netonjm: yep, good catch!

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
Therzok: Gtk and MacOS coordinates are different. I think Gtk does not account for the toolbar height (when the toolbar is in the titlebar) and a few other things in full screen.

Fine by me!

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
netonjm: Why? In theory, there will not be views above this, so the size will always be Frame of the view

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
netonjm: The problem is we want to change the icon in all cases only when is not dragging

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/NativeContainerWidget.cs
netonjm: The idea behind this class is instead embed a native view into a dump Gtk.Widget class, give a little more functionality, for example encapsulate code that always repeats every time we embed the view  (CanFocus = true,Sensitive = true)  or get the native content. In addition if we normalize this in other places (like Pads) it allows us to distinguish easily in the gtk world when this is a native embed view.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/SplitterContainerWidget.cs
netonjm: Fixed!

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockContainer.cs
```diff
@@ -573,6 +581,11 @@ protected override bool OnMotionNotifyEvent (Gdk.EventMotion e)
 				}
 				return base.OnMotionNotifyEvent (e);
 			}
+
+			public void SetSize (Rectangle rect)
+			{
+				OnSizeAllocated (rect);
```
slluis: This should be SizeAllocate(rect)

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/NativeContainerWidget.cs
slluis: We could use a subclass of [GtkNSViewHost](https://github.com/mono/monodevelop/blob/f9839cdd26fd7710315b7ad7fcb7bc7b585223c9/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/GtkNSViewHost.cs#L34) for this.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/SplitterContainerWidget.cs
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		public void Init (DockGroup grp, int index)
+		{
+			splitter?.Init (grp, index);
+		}
+
+		public void SetSize (Rectangle rect)
+		{
+			OnSizeAllocated (rect);
```
slluis: should be SizeAllocate

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/NativeContainerWidget.cs
netonjm: Completely forgot this class, yeah makes sense!

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/SplitterContainerWidget.cs
netonjm: Fixed

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/DockContainer.cs
netonjm: Fixed

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/NativeContainerWidget.cs
netonjm: Fixed!

netonjm:
rebased and fixed issues @slluis @Therzok @sevoku

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/SplitterGtkNSViewHostWidget.cs
```diff
@@ -0,0 +1,90 @@
...
+
+#if MAC
+
+using System;
+using AppKit;
+using Gdk;
+
+namespace MonoDevelop.Components.Docking
+{
+	class SplitterGtkNSViewHostWidget : Gtk.GtkNSViewHost, ISplitterWidget
```
Therzok:
```suggestion
	sealed class SplitterGtkNSViewHostWidget : Gtk.GtkNSViewHost, ISplitterWidget
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/SplitterWidgetWrapper.cs
```diff
@@ -0,0 +1,82 @@
...
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+
+using Gdk;
+using MonoDevelop.Components.Docking;
+
+interface ISplitterWidget
```
Therzok: Move this inside the namespace?

netonjm: Yes! Good catch!

netonjm: Fixed !

slluis:

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
```diff
@@ -0,0 +1,178 @@
...
+			} else {
+				dragPos = (int)point.Y;
+				dragSize = obj.Allocation.Height;
+			}
+
+			DragStarted?.Invoke (this, EventArgs.Empty);
+		}
+
+		NSTrackingArea trackingArea;
+		public override void UpdateTrackingAreas ()
```
abock: Why do you need to override `UpdateTrackingAreas` and remove/add a new tracking area at all? You should be able to create a single tracking area in the constructor and (as @Therzok was I think alluding to) and pass it ` NSTrackingAreaOptions.InVisibleRect` - Cocoa will update the tracking area for you when the view's visible frame changes.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
```diff
@@ -0,0 +1,178 @@
...
+		{
+			if (trackingArea != null) {
+				RemoveTrackingArea (trackingArea);
+				trackingArea.Dispose ();
+			}
+			var options = NSTrackingAreaOptions.MouseMoved | NSTrackingAreaOptions.ActiveInKeyWindow | NSTrackingAreaOptions.MouseEnteredAndExited;
```
abock: Tracking areas support mouse dragging out of the box. You are effectively recreating drag operations via mouse down/move/up. Your tracking area options should be:

```csharp
NSTrackingAreaOptions.ActiveInKeyWindow |
NSTrackingAreaOptions.InVisibleRect |
NSTrackingAreaOptions.MouseEnteredAndExited |
NSTrackingAreaOptions.EnabledDuringMouseDrag
```

And then override `MouseDragged` to update the splitter. The local event monitor can invoke this method as well.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/GtkNSViewHost.cs
```diff
@@ -31,7 +31,7 @@
 
 namespace Gtk
 {
-	public sealed class GtkNSViewHost : Widget
+	public class GtkNSViewHost : Widget
```
abock: This is sealed for a reason... I do not want different host implementations for native embedding. By unsealing, you are exposing (and thus giving license to) very intricate GTK methods, many of which are already overridden for overriding. I think you only do this because you wanted a disposal hook. If that is the case, just add an event/callback to the host.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
netonjm: Fixed!

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
netonjm: Fixed!

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/GtkNSViewHost.cs
netonjm: Fixed

netonjm:
I have already discovered what my problem was, I was not canceling correctly the Gtk window events and and that's why I was losing my mouse events! YOU WERE RIGHT GUYS @abock  @abock ! Thanks to this now I can kill all the other layering and simplify the things! THANK YOU!

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
```diff
@@ -0,0 +1,174 @@
...
+		}
+
+		void AddRemoveFilter (bool enable)
+		{
+			if (enable)
+				Gdk.Window.AddFilterForAll (Filter);
+			else
+				Gdk.Window.RemoveFilterForAll (Filter);
+		}
+		static Gdk.FilterReturn Filter (IntPtr xevent, Gdk.Event evnt) => Gdk.FilterReturn.Remove;
```
Therzok: Line before this and remove empty lines below.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
```diff
@@ -0,0 +1,174 @@
...
+		public void Init (DockGroup grp, int index)
+		{
+			dockGroup = grp;
+			dockIndex = index;
+		}
+
+		public override void MouseDown (NSEvent theEvent)
+		{
+			dragging = true;
+
```
abock: I would think you don't need to do anything other than `AddRemoveFilter` in `MouseDown` since you'll do all the real work in `MouseDragged` (where you can also set `dragging = true`).

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
```diff
@@ -0,0 +1,174 @@
...
+				return;
+			}
+			if (dockGroup.Type == DockGroupType.Horizontal) {
+				NSCursor.ResizeLeftRightCursor.Set ();
+			} else {
+				NSCursor.ResizeUpDownCursor.Set ();
+			}
+		}
+
+		public override void MouseMoved (NSEvent theEvent)
```
abock: You should be able to fully remove `MouseMoved` now that you have `MouseDragged`. You may also not even need to track this `dragging` state?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
```diff
@@ -0,0 +1,174 @@
...
+				SetResizeCursor ();
+			}
+			hover = true;
+			base.MouseEntered (theEvent);
+		}
+
+		public override void MouseExited (NSEvent theEvent)
+		{
+			if (!dragging) {
+				NSCursor.ArrowCursor.Set ();
```
abock: You should push and pop cursors, not set them, if you're going to manage them manually like this. Assuming `Arrow` is not correct.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
```diff
@@ -0,0 +1,174 @@
...
+		{
+			var options = NSTrackingAreaOptions.ActiveInKeyWindow |
+				NSTrackingAreaOptions.InVisibleRect |
+				NSTrackingAreaOptions.MouseEnteredAndExited |
+				NSTrackingAreaOptions.EnabledDuringMouseDrag;
+			var trackingArea = new NSTrackingArea (default, options, this, null);
+			AddTrackingArea (trackingArea);
+		}
+
+		public override void MouseEntered (NSEvent theEvent)
```
abock: Have you tried just setting the cursor via overriding `CursorUpdate` (and make sure to add the flag to the tracking area). You may not need to handle this at all in `MouseEntered` and `MouseExited`.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/SplitterGtkNSViewHostWidget.cs
```diff
@@ -0,0 +1,80 @@
...
+//
+
+#if MAC
+
+using Gdk;
+using Gtk;
+
+namespace MonoDevelop.Components.Docking
+{
+	sealed class SplitterMacHostWidget : ISplitterWidget
```
abock: In the previous version, you were doing quite a bit of tear down in `Dispose` of the subclass of the host. Do you no longer need to worry about that, considering the tracking area changes now in place?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
netonjm: the dockGroup to resize an item needs the relative distance to grow, how we can calculate it without set a starting point?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/SplitterGtkNSViewHostWidget.cs
Therzok: This was the [Dispose](https://github.com/mono/monodevelop/compare/9c6cca658242447f3c97add638becb08a3f5b27c#diff-90f7eb2cc9a4b0c64d75c45646360b24R92).

Seems like they're not needed anymore.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/SplitterWidgetWrapper.cs
```diff
@@ -0,0 +1,88 @@
...
+		void SizeAllocate (Gdk.Rectangle rect);
+		bool Visible { get; set; }
+		void Hide ();
+		void Show ();
+		Gtk.Widget Widget { get; }
+	}
+
+	class SplitterWidgetWrapper : ISplitterWidget
+	{
+		public Gtk.Widget Parent {
```
Therzok: Minor, style nit:
```
public Gtk.Widget Parent => nativeWidget.Parent;
public Gtk.Widget Widget => nativeWidget.Widget;
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
```diff
@@ -0,0 +1,173 @@
...
+		DockGroup dockGroup;
+		int dockIndex;
+		Widget parent;
+
+		int dragPos, dragSize;
+		bool hover, dragging;
+
+		public MacSplitterWidget ()
+		{
+			var options = NSTrackingAreaOptions.ActiveInKeyWindow |
```
Therzok: Nit:
```suggestion
			const NSTrackingAreaOptions options = NSTrackingAreaOptions.ActiveInKeyWindow |
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/SplitterGtkNSViewHostWidget.cs
netonjm: Yeah, this was old code. I added this class becase the docking system uses ISplitterWidget which sometimes needs to get the current GtkWidget which in case of Cocoa backend is the GtkNSViewHost. In the old implementation this was "easy" because this interface was implemented in this layer, but because it's sealed I created an intermedia layer to do this

and if we sealed the GtkNSViewHost this could be a wra

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/SplitterWidgetWrapper.cs
netonjm: Fixed

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
netonjm: Without use CursorUpdate override and NSTrackingAreaOptions.CursorUpdate, it's necessary to handle cursor changes in mouse in/out

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
```diff
@@ -0,0 +1,170 @@
...
+			base.MouseExited (theEvent);
+		}
+
+		void SetResizeCursor ()
+		{
+			if (dockGroup == null) {
+				return;
+			}
+			if (dockGroup.Type == DockGroupType.Horizontal) {
+				NSCursor.ResizeLeftRightCursor.Set ();
```
Therzok: These should be Push/Pop

slluis:

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
```diff
@@ -0,0 +1,170 @@
...
+
+			AddRemoveFilter (true);
+		}
+
+		public override void MouseUp (NSEvent theEvent)
+		{
+			if (hover) {
+				SetResizeCursor ();
+			} else {
+				NSCursor.ArrowCursor.Set ();
```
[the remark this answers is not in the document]
netonjm: Fixed

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
netonjm: Fixed

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
netonjm: Fixed

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
netonjm: I've tried it, but I have not managed to make it work. can we leave the current implementation and include this comment in a future PR ?

Therzok:
I'm okay with tracking future improvements in a separate issue and merging this as-is!

netonjm:
@abock I think I fixed all the big issues. Related to the cursor logic, I'm not familiar with CursorUpdate mechanism from cocoa, I did it in the old way which probably adds some additional logic I think we are ok. Maybe we could include the CursorUpdate change as @Therzok pointed like an improvement in a future PR?

sevoku:

abock:
This is still not fully functional. For example, I cannot resize split editors.

sevoku:
Good catch @abock, for me a splitview broke all GTK events now

abock:
@netonjm I have a feeling that the new native splitter subview might have a lower z-order than the native editor view.

We might need to extend the host to add the native view using `addSubview:positioned:relativeTo:` so you can specify `NSWindowAbove` to ensure the splitter view is positioned in front of anything else native.

netonjm:
mmm... maybe but all the native views are receiving events, if was a zorder issue, this should not work, no? But yes we can try! Maybe we can also check this https://github.com/mono/monodevelop/pull/7762/files#diff-9da72707c4714334ac8e978f2b06ba24R171

slluis:
I forgot about document splits. Those are not handled by the dock system, so no wonder they still don't work. That split is a VPanedThin.

sevoku:
@slluis it's actually HPanedThin, but it's a different issue. As discussed on Slack, I've created https://github.com/mono/monodevelop/pull/7838 to disable the thin splitter for the new editor. But we need to find a better solution.

@abock @netonjm it seems to be really the ordering, with 4222db22b08da434b71dc9eba75073a0901763b2 I can't reproduce the issue with lost GDK events anymore. Previously non of the exit events fired under some weird conditions.

abock:
@sevoku that's great to hear. It makes sense too - the splitters are probably realized before the pane content views are, so they would have been below one level.

netonjm:
With the new z-order changes and validity checks, it seems to work as expected. We could do a final individual verification to make sure we do not forget any details without covering and GTK is not broken in any case not contemplated.

sevoku:

sevoku:

PR #7764, opened 2019-05-31
[Mac] Don't force resize the window when positioning it

iainx:
Stop force resizing the window, because in some situations it can cause the NSWindow and the GtkWindow to be different
sizes.

Fixes VSTS #900933

### DIFF
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -1281,7 +1281,7 @@ namespace MonoDevelop.MacIntegration
 			y += GetTitleBarHeight (w);
 			var dr = FromDesktopRect (new Gdk.Rectangle (x, y, width, height));
 			var r = w.FrameRectFor (dr);
-			w.SetFrame (r, true);
+
 			base.PlaceWindow (window, x, y, width, height);
 		}
 


iainx:

iainx:
the `w.SetFrame()` call is the cause, I’m thinking there’s a race between that call and the Gtk window listening for the NSWindow frame being changed

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -1272,38 +1272,6 @@ internal override void AddChildWindow (Gtk.Window parent, Gtk.Window child)
 			w.AddChildWindow (overlay, NSWindowOrderingMode.Above);
 		}
 
-		internal override void PlaceWindow (Gtk.Window window, int x, int y, int width, int height)
```
Therzok: Why did we do this in the past? Was this for the searchbar popup?

slluis: Agree, I'd like to have a justification of why this code is not necessary anymore.

iainx: Apart from the call at start up, it's called from two other places (in the monodevelop codebase): `DockFrameTopLevel.UpdateWindowPos` and `DockItem.SetFloatMode`. None of the docking behaviours seem to be affected by the change, but I'll give it a full test on Monday

sevoku: I think we still need the desktop bounds calculation, otherwise the windows will open with "y-titlebar" height (higher by that amount) on each restart, because we store the GDK location which does not include the decorations. We should just not call w.SetFrame, but adjust the location by the titlebar height.

iainx:
[Mac] Don't resize the window frame when placing the window
Calling setFrame on the NSWindow resizes the window, but not the Gtk contents causing offset issues in 8.1

There appears to be a race condition caused by making the startup sequence more async than in previous versions. This is triggered by starting MonoDevelop from Finder by double clicking on a solution file. The call to Ide.OpenFilesAsync occurs
earlier in the startup sequence now, and the call to Workbench.Present causes the error. Workbench.Show is called later in
the method by OpenWorkbenchItem, and if we remove the first call to Workbench.Present, then the workbench window opens
correctly, but slightly later. I am still tracking this down, but removing this setFrame call seems to be the least
dangerous fix for this bug until we fully understand what is going on.

Fixes VSTS #900933

iainx:
@lluis @sevoku I've made those changes, which should fix it, but I'd prefer a fuller understanding of the issue so we can fix it properly, and to understand if the same issue can be caused elsewhere

slluis:

PR #7772, opened 2019-06-03
Fixes VSTS Bug 904321: [Feedback] Xamarin Android Breakpoint not work

mkrueger:
in debug

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/904321

Exception from feedback ticket :
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/896686

Produces many out of range exceptions in text segment marker. This
checks now for empty lines - can happen during an edit session when a
line is cleared. Exceptions in that case prevent further drawing in
that line. This is now changed and the drawing code is hardened
against exceptions in line markers.


mkrueger:

mkrueger:

Issue #4917, opened 2018-05-27
Hang in CommandManager due to async update of keybinding


Therzok:
We have the following:
https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandHandler.cs#L77-L100
https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs#L505-L507

We have a possible issue with the code, given an `UpdateAsync` which looks like:

```
protected override async Task UpdateAsync (CommandArrayInfo info, CancellationToken cancelToken)
{
    // Do some UI work
    UpdateMyLabels();

    // case 1 at this point, we return the task
    await Task.Run (() => SomeBgWork());

    // it can either be configureawait(false) and runtime.runinmainthread or configureawait(true)

    // Therefore, at this point, we need to run on the UI thread, but that is blocked by CommandManager.ProcessKeyEvent which holds the UI thread
}
```

when bg work is started, we return a task, there is a window to race, that is pretty big, depending on how much bg code takes, because we almost instantly go to a synchronous wait on keypress.

~CoreCLR (implicitly, mono 2018-02+) have a smart mechanism through which if a given task has no timeout/cancellation, it'll inline the task to run on the given thread: http://source.dot.net/#System.Private.CoreLib/src/System/Threading/Tasks/Task.cs,2910~

~Alas, since the cancellation token _is_ bound to a cancellation token source, that cannot save us, since we always pass in a cancellation token bound to a source:
https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandHandler.cs#L79
https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandHandler.cs#L91
https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs#L2624
http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs,2678~

EDIT: it seems it won't if it's already in running state, which is the case here.

I guess we should either wait a finite time ~or somehow figure out a way to inline the task on the UI thread~. Otherwise, we'll hang. I could reproduce this with go to declaration, but not sure how, as the race window there is _incredibly small_
[...]

PR #7773, opened 2019-06-03
[Shell] Fix fatal error changing keybindings prefs to ReSharper 

jsuarezruiz:
The crash:
```
FATAL ERROR [2019-04-30 14:04:35Z]: Visual Studio failed to start. Some of the assemblies required to run Visual Studio (for example gtk-sharp)may not be properly installed in the GAC.
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
  at System.Collections.Generic.List`1[T].get_Item (System.Int32 index) [0x0000e] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-08/external/bockbuild/builds/mono-x64/external/corefx/src/Common/src/CoreLib/System/Collections/Generic/List.cs:163 
  at MonoDevelop.Components.Commands.CommandManager.ProcessKeyEventCore (Gdk.EventKey ev) [0x0016e] in /Users/vsts/agent/2.150.0/work/1/s/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs:544 
  at MonoDevelop.Components.Commands.CommandManager.ProcessKeyEvent (Gdk.EventKey ev) [0x00000] in /Users/vsts/agent/2.150.0/work/1/s/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs:452 
  at MonoDevelop.Components.Commands.CommandManager.OnNSEventKeyPress (AppKit.NSEvent ev) [0x0006b] in /Users/vsts/agent/2.150.0/work/1/s/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs:352 
  at ObjCRuntime.Trampolines+SDLocalEventHandler.Invoke (System.IntPtr block, System.IntPtr theEvent) [0x00014] in /Library/Frameworks/Xamarin.Mac.framework/Versions/5.6.0.25/src/Xamarin.Mac/ObjCRuntime/Trampolines.g.cs:1032 
  at (wrapper native-to-managed) ObjCRuntime.Trampolines+SDLocalEventHandler.Invoke(intptr,intptr)
  at (wrapper managed-to-native) Gtk.Application.gtk_main()
  at Gtk.Application.Run () [0x00001] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-08/external/bockbuild/builds/gtk-sharp-None/gtk/Application.cs:145 
  at MonoDevelop.Ide.IdeStartup.Run (MonoDevelop.Ide.MonoDevelopOptions options) [0x00429] in /Users/vsts/agent/2.150.0/work/1/s/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs:216 
```

Fixes VSTS #860275

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -564,8 +564,7 @@ namespace MonoDevelop.Components.Commands
 				if (cinfo.Enabled && cinfo.Visible) {
 					if (!dispatched)
 						dispatched = DispatchCommand (command.Id, null, null, CommandSource.Keybinding, ev.Time, cinfo);
-					if (command != null)
-						conflict.Add (command);
+					conflict.Add (command);
 				} else
 					bypass = true; // allow Gtk to handle the event if the command is disabled
 			}
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -552,7 +552,8 @@ namespace MonoDevelop.Components.Commands
 			var dispatched = false;
 
 			for (int i = 0; i < commands.Count; i++) {
-				CommandInfo cinfo = GetCommandInfo (commands[i].Id, new CommandTargetRoute ());
+				Command command = commands [i];
+				CommandInfo cinfo = GetCommandInfo (command.Id, new CommandTargetRoute ());
 				if (cinfo.IsUpdatingAsynchronously)
 					cinfo.UpdateTask.Wait (); // Not nice, but we need a synchronous result here
 				if (cinfo.Bypass) {
@@ -562,8 +563,9 @@ namespace MonoDevelop.Components.Commands
 
 				if (cinfo.Enabled && cinfo.Visible) {
 					if (!dispatched)
-						dispatched = DispatchCommand (commands [i].Id, null, null, CommandSource.Keybinding, ev.Time, cinfo);
-					conflict.Add (commands [i]);
+						dispatched = DispatchCommand (command.Id, null, null, CommandSource.Keybinding, ev.Time, cinfo);
+					if (command != null)
+						conflict.Add (command);
 				} else
 					bypass = true; // allow Gtk to handle the event if the command is disabled
 			}
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -552,8 +552,7 @@ namespace MonoDevelop.Components.Commands
 			var dispatched = false;
 
 			for (int i = 0; i < commands.Count; i++) {
-				Command command = commands [i];
-				CommandInfo cinfo = GetCommandInfo (command.Id, new CommandTargetRoute ());
+				CommandInfo cinfo = GetCommandInfo (commands [i].Id, new CommandTargetRoute ());
 				if (cinfo.IsUpdatingAsynchronously)
 					cinfo.UpdateTask.Wait (); // Not nice, but we need a synchronous result here
 				if (cinfo.Bypass) {
@@ -563,8 +562,10 @@ namespace MonoDevelop.Components.Commands
 
 				if (cinfo.Enabled && cinfo.Visible) {
 					if (!dispatched)
-						dispatched = DispatchCommand (command.Id, null, null, CommandSource.Keybinding, ev.Time, cinfo);
-					conflict.Add (command);
+						dispatched = DispatchCommand (commands [i].Id, null, null, CommandSource.Keybinding, ev.Time, cinfo);
+					if (!bindings.BindingExists (binding))
+						break;
+					conflict.Add (commands [i]);
 				} else
 					bypass = true; // allow Gtk to handle the event if the command is disabled
 			}


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -562,8 +563,9 @@ bool ProcessKeyEventCore (Gdk.EventKey ev)
 
 				if (cinfo.Enabled && cinfo.Visible) {
 					if (!dispatched)
-						dispatched = DispatchCommand (commands [i].Id, null, null, CommandSource.Keybinding, ev.Time, cinfo);
-					conflict.Add (commands [i]);
+						dispatched = DispatchCommand (command.Id, null, null, CommandSource.Keybinding, ev.Time, cinfo);
+					if (command != null)
+						conflict.Add (command);
 				} else
```
mrward: Not sure this fixes the problem. The exception callstack based on the commit in the bug report seems to be when [accessing an item using an index out of the range of the list](https://github.com/xamarin/monodevelop/blob/b00242a6169f195f7434f61f37263e351b129cb2/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs#L544) and is not due to a null command in the list.

The problem looks to me as though the list returned from the [KeyBindingManager](https://github.com/xamarin/monodevelop/blob/b00242a6169f195f7434f61f37263e351b129cb2/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/KeyBindingManager.cs#L756) is being changed during this method call. If this method is run on the UI thread does that mean the KeyBindingManager is being modified on another thread? Or is it somehow being changed whilst the loop is occurring. Be useful to understand how that could happen.

A possible fix here would be to close the list, but it would be more useful to see if we can understand why the list is being modified.

mrward: I guess it makes sense that changing the current key bindings preferences updates the commands in the KeyBindingManager. So does that happen during this loop?

jsuarezruiz: Thanks for the feedback Matt!. The change is not really a null protection, rather, create a clone of the command.
The problem is that the command collection is empty. When trying to access an element by index, we have the error. I have reviewed where the collection of commands changes and when changing the pref keybinding, after this line: https://github.com/xamarin/monodevelop/blob/2957d9b328df505c5bfac78ac0d634f046bca0e7/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs#L1637 we have the collection empty.
I will review everything better. I also want to see if I can determine from which version it occurs.

sevoku: Since it looks like the list gets modified "somehow", the safer hotfix would be to create a copy with `commands.ToArray()` and loop over that copy, rather than to create a local var for each command inside the loop. Otherwise it might also break if something calls a commands.Remove while this is running.

Therzok: @sevoku I would verify if `commands` is modified on a background thread, other than foreground thread only. If it's modified on foreground only, ToArray() will work, otherwise, we have bigger problems.

Therzok: Resending comment:
@sevoku I would verify if commands is modified on a background thread, other than foreground thread only. If it's modified on foreground only, ToArray() will work, otherwise, we have bigger problems

It returns a reference to the command list in KeyBindingService which is modified [here](http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Components.Commands/KeyBindingManager.cs,1c6511dfa7cfbc99,references) and [here](http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Components.Commands/KeyBindingManager.cs,c68ad38bdc4ae67f,references). So technically, we have possible races here too.

Therzok: Ah, I just noticed Matt already commented the exact same thing. github didn't scroll properly.

sevoku: @Therzok so what you mean is #4917, right? In that case, I'd propose to add the `ToArray()` and merge this, to harden it a bit (won't hurt) and then focus on #4917 separately.

Therzok: ToArray() works if you're modifying a collection during enumeration on the same thread. This is a classic race condition scenario.

Therzok: That's a different bug.

jsuarezruiz: Thanks to everyone for the feedback. After reviewing everything thoroughly, I understand what is happening:
1. We have the Visual Studio key bindings scheme (190 commands by default).
2. Press CMD+,; this execute the method https://github.com/mono/monodevelop/blob/ed437173bd7358e64ca1a547e4744b1b74fdf47b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs#L460 and open the preferences dialog.
3. At this point, we use a new thread https://github.com/mono/monodevelop/blob/ed437173bd7358e64ca1a547e4744b1b74fdf47b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs#L565, then we change the bindings scheme to Rider (135 commands). Note that the CMD+, command will not exists in Rider scheme.
4. The execution continues after complete DispatchCommand with https://github.com/mono/monodevelop/blob/ed437173bd7358e64ca1a547e4744b1b74fdf47b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs#L566 BUT the commands collection changed before in https://github.com/mono/monodevelop/blob/305bddc1e9dcd90caf25f8bed1a042efc317528b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/KeyBindingManager.cs#L656
5. Crash!. Why?. Because we are trying to access to a command of the collection that has changed (in fact, the command no longer exists with the Rider schema).

We can use a copy of the collection in ProcessKeyEventCore (ToArray) or, verify if the evaluated command still exist and if not, break the loop. I just applied changes to do the second option.

sevoku:

PR #7774, opened 2019-06-03
Fix 901093: Completing a snippet adds a newline

DavidKarlas:
Fix 901091: Caret jumps to previous line after completing a prop snippet


abock:

PR #7779, opened 2019-06-04
Fix mime type assignment issues in FileModel

slluis:
The MimeType property was not properly set in some cases, such as when
creating a new file, or when saving a file with a new name and extension.

FileModel now ensures that the MimeType property is always up to date,
and TextBufferFileModel now updates the content type of the buffer
when creating a new document or when saving a document.

Fixes VSTS #825629 - No error message shows up in TypeScript files
Fixes VSTS #889145 - First time to add MVC View Page file(.cshtml), the intellisense and colorization doesn't work at once and shows un-save state. It needs to reopen or insert.
Fixes VSTS #894970 - Double quotes are not automatically matched after tapping the quote key

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentModel.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentModel.cs
@@ -103,13 +103,19 @@ namespace MonoDevelop.Ide.Gui.Documents
 
 		public void CreateNew ()
 		{
+			InitializeNew ();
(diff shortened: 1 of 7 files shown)


slluis:

KirillOsenkov:
Unit-test failure is a native crash:
`2019-06-04 13:05:54.559 vstool[1360:33709] error: * Assertion at mono-threads.c:1886, condition `previous_token == NULL || previous_token == INTERRUPT_STATE' not met`

I think this happened when aborting the file system watcher thread? Yup, https://github.com/mono/mono/issues/10832

PR #7788, opened 2019-06-05
Fixes VSTS Bug 901463: Copy and paste is broken or copies wrong text

mkrueger:
in "Changes" window

Fixes VSTS Bug 901464: Copy/paste is broken in "authors" tab (wrong text or nothing happens)

Fixes Bug 901463: Copy and paste is broken or copies wrong text in "Changes"
window

DiffView was still using the old SharpDevelop clipboard command system
for cut/copy/paste. Moved to monodevelop command handlers.

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/BlameView.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/BlameView.cs
@@ -1,4 +1,4 @@
-//
+﻿//
 // BlameView.cs
 //
 // Author:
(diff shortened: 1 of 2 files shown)


mkrueger:

mkrueger:

mkrueger:
Added https://devdiv.visualstudio.com/DevDiv/_workitems/edit/901464

It's the same issue.

slluis:
That is weird, IClipboardHandler should still work.

slluis:
I found a problem in how content interfaces like IClipboardHandler are discovered for attached views like the control version views. However, fixing the root of the problem requires some intrusive changes we can do for 8.1, so let's merge this fix.

PR #7791, opened 2019-06-05
Fix 901424: After click in the document navigation editor doesn't get focus

DavidKarlas:
Just wanted to note, same problem is in old editor, and this fix fixes both.


slluis:

PR #7792, opened 2019-06-05
Fixes VSTS Bug 905613: [Feedback] Cannot Copy Results of Find in Files

mkrueger:
…

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/905613

AppendFormat was wrong there - CopyData is already formatted.
AppendFormat can easily throw an exception depending on the text
copied. This patch fixes it.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/SearchResultWidget.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/SearchResultWidget.cs
@@ -630,14 +630,18 @@ namespace MonoDevelop.Ide.FindInFiles
 			var sb = new StringBuilder ();
 			foreach (TreePath p in treeviewSearchResults.Selection.GetSelectedRows (out model)) {
 				TreeIter iter;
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/SearchResultWidget.cs
```diff
@@ -630,14 +630,18 @@ public void CopySelection ()
...
-				if (!model.GetIter (out iter, p))
+				if (!model.GetIter (out iter, p)) {
 					continue;
+				}
 				var result = store.GetValue (iter, SearchResultColumn) as SearchResult;
-				if (result == null)
+				if (result == null) {
 					continue;
-
-				sb.AppendFormat (result.GetCopyData (this));
```
Therzok: wow. we were lucky.

mkrueger: Yes, but if you copy more data chances are that this fails

PR #7801, opened 2019-06-06
MonoDevelopWorkspace: Prevent hangs during OpenDocument

sandyarmstrong:
LightBulb actions like "Generate class in new file" result in the
async `DocumentManager.OpenDocument` flow hanging as soon as context
switches back to the UI thread. In this case, we do not want to be
pumping the UI thread. In other cases, like navigation, we do.

As a low-impact fix, put a timeout on the UI thread pumping.

In the future, we'll want to improve the async flow of navigation and
suggestions in Roslyn.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/892087

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
@@ -565,10 +565,19 @@ namespace MonoDevelop.Ide.TypeSystem
 			if (mdProject == null)
 				return;
 
-			var task = OpenDocumentWithTextViewAsync (doc, mdProject, activate);
-			// Can't wait for the task to finish synchronously since doing so would deadlock the UI thread.
-			while (!task.IsCompleted) {
-				DispatchService.RunPendingEvents (30);
+			using (var cts = new CancellationTokenSource ()) {
+				// Timeout on the wait because:
+				// * Document may not show a text view at all
+				// * In the case of lightbulb actions like "Generate class in new file",
+				//   document open may be blocked as long as we keep pumping
+				var task = Task.WhenAny (
+					Task.Delay (1000, cts.Token),
+					OpenDocumentWithTextViewAsync (doc, mdProject, activate));
+
+				// Can't wait for the task to finish synchronously since doing so would deadlock the UI thread.
+				while (!task.IsCompleted) {
+					DispatchService.RunPendingEvents (30);
+				}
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -563,13 +563,20 @@ public override void OpenDocument (DocumentId documentId, bool activate = true)
...
-				return;
-
-			var task = OpenDocumentWithTextViewAsync (doc, mdProject, activate);
+				return;
+
+			// Timeout on the wait because:
+			// * Document may not show a text view at all
+			// * In the case of lightbulb actions like "Generate class in new file",
+			//   document open may be blocked as long as we keep pumping
+			var task = Task.WhenAny (
```
Therzok: Not a blocker, but usually the pattern when using Task.Delay is:
```csharp
using (var cts = new CancellationTokenSource ()) {
    var delay = Task.Delay (1000, cts.Token);
    // the rest of the code
}
```

To avoid having too many delay timer tasks alive. (in example, the delay task will remain alive for 1 second even if the other task completed).

sandyarmstrong: Good call, fixed!.

sandyarmstrong:

slluis:
Why is pumping the UI thread a problem in this scenario?

Therzok:
@slluis From what I remember from the call I had with @sandyarmstrong last night, the UI would just have a progress monitor stuck loading infinitely. This is via the code generation code fix.

To repro, try typing:
```csharp
private NonExisting value;
```

Then generate the `NonExisting` class via a code fix in a new file.

sandyarmstrong:
@slluis I don't fully understand the mechanics of it, but basically roslyn code actions execute on the main thread, and this particular one ends up calling `Workspace.OpenDocument` (which is synchronous). But our async `DocumentManager.OpenDocument` flow gets blocked as soon as it context switches back to the main thread. Removing the UI thread pumping fixes that.

I did a lot of testing with UI thread pumping totally removed. The only regression was that some navigation features that opened a file for the first time would occasionally not navigate to the actual line on the first try. So I decided the best compromise for now was to keep the pump in, but ensure it does not run indefinitely.

sandyarmstrong:
Close in favor of https://github.com/mono/monodevelop/pull/7823

PR #7806, opened 2019-06-06
Fix 801652: New Editor: broken history of references navigation

DavidKarlas:
This fix restores functionality to what it was before  new document model change that changed what is stored in `textView.Properties` bag which made navigation to not work anymore.


abock:
I would rather take the other PR which fully fixes the old editor.

DavidKarlas:
This is obsolete now...

PR #7808, opened 2019-06-06
Fix old editor settings/issues

mkrueger:
Fixes some issues I saw using the old editor.

Fixes VSTS #915554 - Old editor adds BOM to file on saving

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditorViewContent.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditorViewContent.cs
@@ -88,6 +88,7 @@ namespace MonoDevelop.Ide.Editor
 
 				await Init (editor, impl);
 				HasUnsavedChanges = impl.IsDirty;
+				await UpdateStyleParent (Owner, editor.MimeType, default);
 
 				// Editor extensions can provide additional content
 				NotifyContentChanged ();
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditorViewContent.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditorViewContent.cs
@@ -76,6 +76,7 @@ namespace MonoDevelop.Ide.Editor
 		{
 			fileDescriptor = modelDescriptor as FileDescriptor;
 			await base.OnInitialize (modelDescriptor, status);
+			Encoding = fileDescriptor.Encoding;
 		}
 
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditorViewContent.cs
```diff
@@ -76,6 +76,7 @@ protected override async Task OnInitialize (ModelDescriptor modelDescriptor, Pro
 		{
 			fileDescriptor = modelDescriptor as FileDescriptor;
 			await base.OnInitialize (modelDescriptor, status);
+			Encoding = fileDescriptor.Encoding;
```
Therzok: Why is this needed? Was this supposed to be `textEditor.Encoding`? If so, we need to override `OnEncodingChanged` and set it from there.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditorViewContent.cs
```diff
@@ -87,6 +88,7 @@ protected override async Task<Control> OnGetViewControlAsync (CancellationToken
 
 				await Init (editor, impl);
 				HasUnsavedChanges = impl.IsDirty;
+				await UpdateStyleParent (Owner, editor.MimeType, default);
```
Therzok:
```suggestion
				await UpdateStyleParent (Owner, editor.MimeType, editorOptionsUpdateCancellationSource.Token);
```

And also initialize the CancellationTokenSource regardless of update being called.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditorViewContent.cs
mkrueger: It's needed because it needs to be null if not set.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/TextEditorViewContent.cs
mkrueger: Needs to run on GetView.

Therzok: Well, if we get an update while the view is constructed, we will have conflicting data being set.

mkrueger: hm, in any case it's the wrong token OnGetViewControlAsync seems to get a token - wrong sorting of parameters btw.

PR #7809, opened 2019-06-06
[Mac] Retry unmounting because of an issue in the Mojave bots

rodrmoya:
With the move to Xcode 10.2, we need to build on Mojave bots, and there's
a bug (https://github.com/al45tair/dmgbuild/issues/10) that makes unmounting
time out on the bots, which is fixed by just retrying :/


rodrmoya:

rodrmoya:
@monojenkins backport release-8.0

nosami:
Should we do a force detach if it fails 5 times? Like here? https://bitbucket.org/al45tair/dmgbuild/pull-requests/11/retry-for-hdiutil-couldnt-unmount-disk2/diff#Ldmgbuild/core.pyT528

PR #7810, opened 2019-06-06
[Workspace] Swap the order of cache clearing in the manager.

Therzok:
By doing so, we ensure that any in-flight operation to update an assembly from disk must happen before we clear the caches.


slluis:

PR #7813, opened 2019-06-06
[Version Control] Popping a git stash that has conflicts gives invalid warning and no further logs

jsuarezruiz:
Applied changes to show more information about the stash operation status (conflicts, stash not found, uncommitted changes, etc.).

Fixes VSTS #901094

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitService.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitService.cs
@@ -25,6 +25,7 @@
 // THE SOFTWARE.
 
 using System;
+using System.Text;
 using System.Linq;
 using MonoDevelop.Core;
 using MonoDevelop.Ide;
@@ -153,8 +154,8 @@ namespace MonoDevelop.VersionControl.Git
 			var t = Task.Run (delegate {
 				try {
 					var res = repo.ApplyStash (monitor, s);
-					ReportStashResult (res);
-					return true;
+					ReportStashResult (repo, res, null);
+					return res == StashApplyStatus.Applied;
 				} catch (Exception ex) {
 					string msg = GettextCatalog.GetString ("Stash operation failed.");
 					monitor.ReportError (msg, ex);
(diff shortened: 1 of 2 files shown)


main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitService.cs
```diff
@@ -166,20 +166,67 @@ public static Task<bool> ApplyStash (GitRepository repo, int s)
...
+			string msg;
+			StashResultType stashResultType;
+
+			switch (status) {
+			case StashApplyStatus.Conflicts:
+				msg = GettextCatalog.GetString ("A conflicting change has been detected in the index.");
+
+				// Include conflicts in the msg
+				using (var RootRepository = new LibGit2Sharp.Repository (repo.RootPath)) {
+					if (!RootRepository.Index.IsFullyMerged) {
```
jsuarezruiz: IMPORTANT, this piece of code is not working as expected. For example, we do Pop of a Stash.
Throws away the (topmost, by default) stash after applying. In case of conflicts, when the changes have been applied, they should be in the index and the following code could obtain information about it.
However, the stash is NOT being applied in case of conflicts and when trying to obtain them, we have nothing.
Any ideas?

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitService.cs
```diff
@@ -166,20 +166,67 @@ public static Task<bool> ApplyStash (GitRepository repo, int s)
...
-				});
+			string msg;
+			StashResultType stashResultType;
+
+			switch (status) {
+			case StashApplyStatus.Conflicts:
+				msg = GettextCatalog.GetString ("A conflicting change has been detected in the index.");
+
+				// Include conflicts in the msg
+				using (var RootRepository = new LibGit2Sharp.Repository (repo.RootPath)) {
```
sevoku: can't we just use `repo` directly?

jsuarezruiz: We need to access the Index to check for conflicts. For that we use a Repository of Libgit2Sharp. `repo` is GitRepository type. To be able to use it, we would need to add a method or property that exposes the conflicts available in the Internal `LibGit2Sharp.Repository` Index.

Therzok: repo.RootRepository should be acessible here (which would give you a libgit2sharp one).

jsuarezruiz: You're right. Changed!.

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitService.cs
sevoku: By default LibGit2Sharp should do an automatic dirty merge, so it seems to be a bug. Can we identify somehow whether the stah was applied or not and notify the user with an Error instead of a Warning, in case it didn't work (probably just check if the stash is still there)?

jsuarezruiz: Ok!. We verify the stashes count before and after doing the Pop to check if the Stash has been applied or not.

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitService.cs
```diff
@@ -166,20 +166,69 @@ public static Task<bool> ApplyStash (GitRepository repo, int s)
...
+			switch (status) {
+			case StashApplyStatus.Conflicts:
+				bool stashApplied = false;
+				msg = GettextCatalog.GetString ("A conflicting change has been detected in the index. ");
+				// Include conflicts in the msg
+				if (stashCount != null && repo is GitRepository gitRepo) {
+					int actualStashCount = gitRepo.GetStashes ().Count ();
+					stashApplied = actualStashCount != stashCount;
+					if (stashApplied) {
+						msg += GettextCatalog.GetString ("The following conflicts have been found:") + Environment.NewLine;
```
mkrueger: Consider using a string builder you're appending an unknown number of strings here.

jsuarezruiz: Right. Applied changes to use a `StringBuilder`.

sevoku:

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitService.cs
```diff
@@ -166,20 +167,70 @@ public static Task<bool> ApplyStash (GitRepository repo, int s)
...
+			Warning
+		}
+
+		static void ShowStashResult (string msg, StashResultType stashResultType)
+		{
+			Runtime.RunInMainThread (delegate {
+				switch (stashResultType)
+				{
+					case StashResultType.Error:
+						IdeApp.Workbench.StatusBar.ShowError (msg);
```
sevoku: I didn't realize that this would be only the status bar, which won't show the full message, especially if there are conflicting files. Shouldn't we maybe pass the monitor from the caller and report there? Or show an alert if there is an error?

jsuarezruiz: Man, right. Added an alert in case of warning or error.

Therzok:
@sevoku ping

sevoku:

Sorry, I totally forgot this one.

PR #7816, opened 2019-06-06
[CompositionManager] Allow the IDE to recover from a state where the …

Therzok:
…composition is broken

Prompt the user with a dialog telling them that the current IDE state is broken and requires a restart, clearing the MEF caches.

Fixes VSTS #892931 - MEF Cache goes haywire

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
@@ -254,7 +254,7 @@ namespace MonoDevelop.Ide.Composition
 				if (e is IOException)
 					return;
 
-				if (!IdeApp.Initialized) {
+				if (!IdeApp.IsInitialized) {
 					Console.WriteLine (e);
 					return;
 				}
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
@@ -241,17 +241,34 @@ namespace MonoDevelop.Ide.Composition
 
 		sealed class IdeRuntimeCompositionExceptionHandler : RuntimeCompositionExceptionHandler
 		{
+			static class Strings
+			{
+				public static string Quit = GettextCatalog.GetString ("Quit");
+				public static string Restart = GettextCatalog.GetString ("Restart");
+			}
+
 			public override void HandleException (string message, Exception e)
 			{
 				base.HandleException (message, e);
 
-				if (e is InvalidRuntimeCompositionException) {
+				if (!(e is IOException)) {
 					// TODO: Fix text.
-					var text = GettextCatalog.GetString ("Some extensions failed to initialize, requiring a restart of {0}", BrandingService.ApplicationName);
-					var restartLabel = GettextCatalog.GetString ("Restart");
+					var text = GettextCatalog.GetString ("There was a problem loading one or more extensions and {0} needs to be restarted.", BrandingService.ApplicationName);
+					var quitButton = new AlertButton (Strings.Quit);
+					var restartButton = new AlertButton (Strings.Restart);
 
-					MessageService.GenericAlert (IdeServices.DesktopService.GetFocusedTopLevelWindow (), Gui.Stock.Error, text, secondaryText: null, new AlertButton (restartLabel));
-					IdeApp.Restart (false).Ignore ();
+					var result = MessageService.GenericAlert (
+						IdeServices.DesktopService.GetFocusedTopLevelWindow (),
+						Gui.Stock.Error,
+						text,
+						secondaryText: null,
+						quitButton,
+						restartButton
+					);
+					if (result == restartButton)
+						IdeApp.Restart (false).Ignore ();
+					else
+						IdeApp.Exit ().Ignore ();
 				}
 			}
 		}
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
@@ -251,26 +251,31 @@ namespace MonoDevelop.Ide.Composition
 			{
 				base.HandleException (message, e);
 
-				if (!(e is IOException)) {
-					// TODO: Fix text.
-					var text = GettextCatalog.GetString ("There was a problem loading one or more extensions and {0} needs to be restarted.", BrandingService.ApplicationName);
-					var quitButton = new AlertButton (Strings.Quit);
-					var restartButton = new AlertButton (Strings.Restart);
+				if (e is IOException)
+					return;
 
-					var result = MessageService.GenericAlert (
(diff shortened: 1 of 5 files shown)


Therzok:
cc @vancura @hbons need help on the text.

The UI currently looks like this, using stock functionality:
Screenshot 2019-06-06 at 13 56 50
Screenshot 2019-06-06 at 13 56 34

Therzok:
Context: The IDE gets into an unrecoverable state shortly after startup, requiring a restart to flush some caches written to disk.

Therzok:
It's already reported via the stock [ExceptionHandler](https://github.com/mono/monodevelop/blob/6a3ff493645a5441dcc174797b29930f4819917f/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs#L53-L59)

jmatthiesen:
@Therzok Can we get into an "infinite loop" situation where we make the fixes here and restart, but then the error comes up again? Is there a point at which we identify some problematic extensions and can disable them (probably future feature work)?

Therzok:
Nope, that should never happen. A restart and cache clear should definitely fix it.

Therzok:
I've improved on the message and UX, now there's a Quit/Restart button and improved wording. Thanks @jmatthiesen

Therzok:
Can't repro the hangs locally, this is weird.

Therzok:
That's weird. Why is this the case on CI?
```
2019-06-18T03:51:54.1523640Z Error: did not find one of the files in sources//Library/Frameworks/Mono.framework/External/monodoc/MonoTouch-man
2019-06-18T03:51:59.2318420Z MonoDevelop.Ide.Composition.CompositionManager+InvalidRuntimeCompositionException: Input assemblies contained Microsoft.VisualStudio.Text.UI.Wpf, Version=16.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a but composition extension point didn't
2019-06-18T03:51:59.2319070Z at MonoDevelop.Ide.Composition.CompositionManager+Caching.WriteMefCacheControl (Microsoft.VisualStudio.Composition.ComposableCatalog catalog, MonoDevelop.Core.Instrumentation.ITimeTracker timer) [0x00101] in /Users/vsts/agent/2.153.2/work/1/s/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs:245
2019-06-18T03:51:59.2319520Z at MonoDevelop.Ide.Composition.CompositionManager+Caching.WriteMefCache (Microsoft.VisualStudio.Composition.RuntimeComposition runtimeComposition, Microsoft.VisualStudio.Composition.ComposableCatalog catalog, Microsoft.VisualStudio.Composition.CachedComposition cacheManager) [0x00025] in /Users/vsts/agent/2.153.2/work/1/s/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs:213
2019-06-18T03:51:59.2319870Z at MonoDevelop.Ide.Composition.CompositionManager+Caching+<>c__DisplayClass19_0.<Write>b__1 () [0x0004d] in /Users/vsts/agent/2.153.2/work/1/s/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.Caching.cs:194
```

Therzok:
@KirillOsenkov @DavidKarlas any idea?

Therzok:
Few more assemblies missing.

Therzok:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
```diff
@@ -238,5 +238,45 @@ void ReadAssemblies (HashSet<Assembly> assemblies, string extensionPath)
 				}
 			}
 		}
+
+		sealed class IdeRuntimeCompositionExceptionHandler : RuntimeCompositionExceptionHandler
+		{
+			static class Strings
+			{
+				public static string Quit = GettextCatalog.GetString ("Quit");
+				public static string Restart = GettextCatalog.GetString ("Restart");
```
sandyarmstrong: Maybe a dumb question, but why put these in a static class like this? Isn't it just extra work, considering this scenario is rare, and even when it gets hit, the strings only will be loaded once before app is restarted?

Therzok: To be honest, I put them here for one reason. Maintainability. I discussed with @sevoku in the past, that Strings should be similar to Styles classes, so we can do theme-ing in one central place.

Such behaviour could, for example, allow us to toggle the language at runtime, and also caching the string results.

Therzok: Here's an example: https://github.com/mono/monodevelop/blob/0d2f08879297b3e6e1bcad2db30dff67f9fc792b/main/src/addins/MacPlatform/MainToolbar/Styles.cs#L31

There's a few things that have been historically problematic, in my opinion:

* GettextCatalog.GetString marshals a string to native, native gives us the translated result which we marshal. Allocations like this can appear in hot code.
* We require a restart to change the IDE language. Such organization enablesu s to switch it at runtime.
* We're putting strings together in the same place. It's easy to then look at what strings we have in the UI and decide whether language is correct or not, as a non-developer.

On the other hand, yeah, we might want ResX or etc. Strongly typed resource names will generate the same thing, so we can hopefully reduce the noise!

abock: I absolutely disagree with the reasoning for using a `Strings` class like this... when you abstract your language from its context, your translations tend to lose fidelity.

Using performance as a reason to hoist gettext-like calls into this kind of grouping does not make sense. Anything presented to the user, including UI should not be in a hot path, at all. Period.

In the case of `Styles`, caching is far more acceptable: you have large resources that may be reused, and are far less contextual than natural language.

Therzok: I am completely in agreement with your points.

I can inline the calls, easily. I thought this pattern would be good to employ, since it's basically what ~strongly typed resx generates.

I'm not using performance as a reason, I was highlighting other possible benefits.

Yes, the drawback is that all these strings that in memory until the end if time.

I'll follow up with a commit reverting this.

sevoku: I think all opinions are valid and make more or less sense depending on the context.

Fact is that we need to think about string normalization like we did for colors. The introduction of `Styles` classes was first of all for unification and maintainability of resources, perf for static resources was a nice side effect. We see similar maintainability issues for strings (different letter casing, different wording, etc. to describe the same in different places).

Performance on the other hand depends a lot on the context, often we can't avoid these calls to be in a hot path (like Marius wrote: lists is the best example, Gtk and native). So there is really no place for "absolutism" in this discussion.

In this particular case we talk about 2 strings for an alert which will trigger in very rare cases, hence I don't see a reason to hold them in memory, however they won't be loaded before they're needed for the first time, if I'm not wrong. So I don't see a reason to spend time on reverting this change either.

PR #7817, opened 2019-06-06
Fix crash when switching configurations

slluis:
Fixes VSTS #910800


slluis:

Issue #5260, opened 2018-07-03
Implement Debugger checksum mismatch workaround for line endings


KirillOsenkov:
Tomas Matousek recommends the following workaround used by the Windows debugger:

if the file was found, but the checksum doesn't match what's mentioned in the .pdb, look at the line endings of the file. Try convert the line endings to opposite (\n -> \r\n, \r\n -> \n) and try calculating that checksum. If that matches, open the file with changed line endings.

This is because the .pdbs can get generated on Windows where Git checks out a file with CRLF, whereas on Mac the file is checked out as-is with LF.

> VS bug [#642329](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/642329)

xamarin-release-manager:
Fixed in version 8.3.0.282 (master)

Pull Request #7822 merged by: Jeffrey Stedfast
Author: mono
Commit: 56d5c462ffc9b1c0067cbf27195ba243e7063160 (mono/monodevelop)

xamarin-release-manager:
> [sync] [VS-21] Comment by **xamarinc**



Fixed in version 8.3.0.282 (master)



Pull Request #7822 merged by: Jeffrey Stedfast  
 Author: mono  
 Commit: 56d5c462ffc9b1c0067cbf27195ba243e7063160 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.2.0.707 (release-8.2)

Pull Request #7822 merged by: Jeffrey Stedfast
Author: mono
Commit: 56d5c462ffc9b1c0067cbf27195ba243e7063160 (mono/monodevelop)

xamarin-release-manager:
> [sync] [VS-24] Comment by **xamarinc**



Fixed in version 8.2.0.707 (release-8.2)



Pull Request #7822 merged by: Jeffrey Stedfast  
 Author: mono  
 Commit: 56d5c462ffc9b1c0067cbf27195ba243e7063160 (mono/monodevelop)

xamarin-release-manager:
Fixed in version 8.6.0.802 (master)

Pull Request #7822 merged by: Jeffrey Stedfast
Author: mono
Commit: 56d5c462ffc9b1c0067cbf27195ba243e7063160 (xamarin/monodevelop)

xamarin-release-manager:
> [sync] [VS-31] Comment by **xamarinc**



Fixed in version 8.6.0.802 (master)



[...]

PR #7822, opened 2019-06-06
[Debugger] Improved source checksum logic for UNIX and DOS line endings

jstedfast:
Since it is possible that the assemblies being debugged could have
been compiled on a different system (potentially with different line
endings), compute hashes for the source code in 3 formats:

1. The source code file with line endings as-is
2. The source code file with DOS line endings
3. The source code file with UNIX line endings

Note: sometimes source code may have mixed line endings, so it is
important to calculate the source code as-is and not just
canonicalized UNIX and DOS line endings.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/642329
Fixes issue #5260

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/SourceCodeLookup.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/SourceCodeLookup.cs
@@ -23,15 +23,17 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
+
 using System;
-using MonoDevelop.Core;
-using System.Collections.Generic;
 using System.IO;
-using System.Security.Cryptography;
 using System.Linq;
-using MonoDevelop.Ide;
-using MonoDevelop.Projects;
+using System.Collections.Generic;
+
 using Mono.Debugging.Client;
+
+using MonoDevelop.Ide;
+using MonoDevelop.Core;
+using MonoDevelop.Projects;
 
 namespace MonoDevelop.Debugger
 {
@@ -120,50 +122,9 @@ namespace MonoDevelop.Debugger
 			return FilePath.Null;
 		}
 
-		public static bool CheckFileHash (FilePath file, byte[] hash)
+		public static bool CheckFileHash (FilePath path, byte[] checksum)
 		{
-			if (hash == null)
-				return false;
-			if (File.Exists (file)) {
-				using (var fs = File.OpenRead (file)) {
-					// Roslyn SHA1 checksum always starts with 20
-					if (hash.Length > 0 && hash [0] == 20)
-						using (var sha1 = SHA1.Create ()) {
-							if (sha1.ComputeHash (fs).Take (15).SequenceEqual (hash.Skip (1))) {
-								return true;
-							}
-						}
-					if (hash.Length > 0 && hash [0] == 32)
-						using (var sha1 = SHA256.Create ()) {
-							if (sha1.ComputeHash (fs).Take (15).SequenceEqual (hash.Skip (1))) {
-								return true;
-							}
-						}
-					if (hash.Length == 20) {
-						using (var sha1 = SHA1.Create ()) {
-							fs.Position = 0;
-							if (sha1.ComputeHash (fs).SequenceEqual (hash)) {
-								return true;
-							}
-						}
-					}
-					if (hash.Length == 32) {
-						using (var sha256 = SHA256.Create ()) {
-							fs.Position = 0;
-							if (sha256.ComputeHash (fs).SequenceEqual (hash)) {
-								return true;
-							}
-						}
-					}
-					fs.Position = 0;
-					using (var md5 = MD5.Create ()) {
-						if (md5.ComputeHash (fs).SequenceEqual (hash)) {
-							return true;
-						}
-					}
-				}
-			}
-			return false;
+			return SourceLocation.CheckFileHash (path, checksum);
 		}
 
 		/// <summary>
@@ -223,4 +184,3 @@ namespace MonoDevelop.Debugger
 		}
 	}
 }
-


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/SourceCodeLookup.cs
```diff
@@ -120,49 +120,101 @@ public static FilePath TryDebugSourceFolders (FilePath originalFile, byte[] hash
...
-						}
-					}
-					if (hash.Length == 32) {
-						using (var sha256 = SHA256.Create ()) {
-							fs.Position = 0;
-							if (sha256.ComputeHash (fs).SequenceEqual (hash)) {
-								return true;
-							}
-						}
+			var unixBuffer = new byte[4096];
```
Therzok:
```suggestion
			var pool = ArrayPool<byte>Shared;
			var unixBuffer = pool.Rent (4096);
```

And at the end of the method:

```csharp
			pool.Return (unixBuffer);
```

And same for the others.

Therzok:
Otherwise, LGTM!

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/SourceCodeLookup.cs
```diff
@@ -120,49 +122,107 @@ public static FilePath TryDebugSourceFolders (FilePath originalFile, byte[] hash
...
+			try {
+				while ((count = stream.Read (buffer, 0, buffer.Length)) > 0) {
+					int unixIndex = 0, dosIndex = 0;
+
+					for (int i = 0; i < count; i++) {
+						var c = buffer[i];
+
+						if (c == (byte) '\r') {
+							if (pc == (byte) '\r')
+								unixBuffer[unixIndex++] = pc;
```
KirillOsenkov: I think we should wrap the single-line statements under `if` with a block, otherwise it's really dangerous if whitespace gets messed up someone will make mistakes later.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/SourceCodeLookup.cs
```diff
@@ -120,49 +122,107 @@ public static FilePath TryDebugSourceFolders (FilePath originalFile, byte[] hash
...
-					if (hash.Length == 32) {
-						using (var sha256 = SHA256.Create ()) {
-							fs.Position = 0;
-							if (sha256.ComputeHash (fs).SequenceEqual (hash)) {
-								return true;
-							}
+			var unixBuffer = ArrayPool<byte>.Shared.Rent (4096 + 1);
+			var dosBuffer = ArrayPool<byte>.Shared.Rent (8192 + 1);
+			var buffer = ArrayPool<byte>.Shared.Rent (4096);
+			byte pc = 0;
```
KirillOsenkov: The meaning of `pc` is not clear from reading the code, should we spell it out?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/SourceCodeLookup.cs
```diff
@@ -120,49 +122,107 @@ public static FilePath TryDebugSourceFolders (FilePath originalFile, byte[] hash
...
 						}
+
+						pc = c;
 					}
-					fs.Position = 0;
-					using (var md5 = MD5.Create ()) {
-						if (md5.ComputeHash (fs).SequenceEqual (hash)) {
-							return true;
+
+					hash.TransformBlock (buffer, 0, count, null, 0);
```
KirillOsenkov: Should we use named arguments here, to clarify what those literals are? 0 and null? Otherwise reading this code is difficult.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/SourceCodeLookup.cs
```diff
@@ -120,49 +122,107 @@ public static FilePath TryDebugSourceFolders (FilePath originalFile, byte[] hash
...
+				using (var dos = HashAlgorithm.Create (algorithm)) {
+					using (var unix = HashAlgorithm.Create (algorithm)) {
+						stream.Position = 0;
+
+						ComputeHashes (stream, hash, dos, unix);
+
+						if (checksum[0] == size && checksum.Length < size) {
+							int take = checksum.Length - 1;
+
+							return hash.Hash.Take (take).SequenceEqual (checksum.Skip (1)) ||
```
KirillOsenkov: Should we cache some subexpressions here in a separate local variable? Such as `checksum.Skip(1)`? To minimize extra allocations.

DavidKarlas:
@jstedfast check debugger-libs which also checks for checksums when placing breakpoints afair

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/SourceCodeLookup.cs
jstedfast: "Previous char" :-)

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/SourceCodeLookup.cs
jstedfast: done

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/SourceCodeLookup.cs
KirillOsenkov: spell it out then ;)

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/SourceCodeLookup.cs
jstedfast: Those are `outputBuffer` and `outputOffset`. I'll add named args to make this more obvious.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/SourceCodeLookup.cs
jstedfast: I was considering writing a custom comparison method since it's like 5 lines of code instead of using Linq. I'll just do that and then there will be no allocations at all :-)

jstedfast:
Added unit tests and improved patch based on suggestions.

jstedfast:
Pushed the logic down into debugger-libs

DavidKarlas:
Just wanted to comment that I tried this on Roslyn code base and it works as advertised!

PR #7823, opened 2019-06-06
Fix issue in generate new class refactoring

slluis:
Fixes VSTS #892087

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
@@ -565,10 +565,21 @@ namespace MonoDevelop.Ide.TypeSystem
 			if (mdProject == null)
 				return;
 
-			var task = OpenDocumentWithTextViewAsync (doc, mdProject, activate);
-			// Can't wait for the task to finish synchronously since doing so would deadlock the UI thread.
-			while (!task.IsCompleted) {
-				DispatchService.RunPendingEvents (30);
+			// This method can be called by Roslyn or the editor in a context which is not the GTK UI context
+			// that MonoDevelop uses. In that case, before starting async an operation that may queue
+			// task continuations into the current context, we switch to the GTK context, so that
+			// whatever is queued will be dispatched when we run RunPendingEvents.
+
+			var oldContext = SynchronizationContext.Current;
+			try {
+				SynchronizationContext.SetSynchronizationContext (Runtime.MainSynchronizationContext);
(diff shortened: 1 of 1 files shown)


slluis:

sandyarmstrong:
Working well for me!

KirillOsenkov:
@slluis could you please explain (ideally in the comment) why we need to set the Sync Context? What was the old context on that thread and why switching to the main sync context works fine?

KirillOsenkov:
@slluis never mind. I'm blind.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -565,10 +565,21 @@ public override void OpenDocument (DocumentId documentId, bool activate = true)
...
+			var oldContext = SynchronizationContext.Current;
+			try {
+				SynchronizationContext.SetSynchronizationContext (Runtime.MainSynchronizationContext);
+				var task = OpenDocumentWithTextViewAsync (doc, mdProject, activate);
+				// Can't wait for the task to finish synchronously since doing so would deadlock the UI thread.
+				while (!task.IsCompleted) {
+					DispatchService.RunPendingEvents (30);
+				}
+			} finally {
+				SynchronizationContext.SetSynchronizationContext (oldContext);
```
KirillOsenkov: Perhaps in the future we could add a method `void Runtime.PumpingWait(Func<Task> asyncMethod)` because this sounds generally useful.

Therzok: I would advise against this pattern being sprinkled everywhere.

https://blogs.msdn.microsoft.com/jfoscoding/2005/08/06/keeping-your-ui-responsive-and-the-dangers-of-application-doevents/

Therzok: If we have specific tasks which need to run in a pumped loop, we could use JTF, fine grained where needed, otherwise it would be messy and surface some impossible to debug issues.

avodovnik: I just ran into a similar issue, when Roslyn called us to Rename a document. I tried using JTF but that didn't work - I'm not entirely sure at this moment. However, this solution did. While I agree that it might not be something to just sprinkle everywhere, it appears that we **have** to do something to fix the mismatch between the various async patterns that we're using. Until then, the `Runtime.PumpingWait` method seems like a _reasonable_ idea.

PR #7836, opened 2019-06-07
[Ide] Handle exception disposing pad on closing the workbench.

mrward:
An error disposing a pad on closing the workbench results in
a fatal exception which may crash the IDE. Handle any error on
disposing the pad with a try/catch.

Fixes VSTS #911276 - [Watson] Crash in PackageConsolePad


mrward:

PR #7838, opened 2019-06-07
[Ide] Use regular Gtk.HPaned for split view with the new editor

sevoku:
This is a workaround until we find the right solution for the broken GDK events with embedded native views.

Fixes VSTS #910879


sevoku:

PR #7841, opened 2019-06-07
Harden the MacSplitterWidget GDK event filter

sevoku:
Make sure to remove the global GDK event filter when
the widget is removed from a superview or disposed, and
add a santiy check to the filter func, ensuring to restore
regular GDK operation.

Additionally recover GDK events after longer inactivity
to recover from the rare case when some other native
widget starts stealing events, so the drag mode never
ends, blocking GDK forever.

See #7762

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
@@ -52,6 +52,7 @@ namespace MonoDevelop.Components.Docking
 
 		public override void MouseEntered (NSEvent theEvent)
 		{
+			lastEventTimestamp = theEvent.Timestamp;
 			if (!dragging) {
 				SetResizeCursor ();
 			}
@@ -61,6 +62,7 @@ namespace MonoDevelop.Components.Docking
 
 		public override void MouseExited (NSEvent theEvent)
 		{
+			lastEventTimestamp = theEvent.Timestamp;
 			if (!dragging) {
 				SetDefaultCursor ();
 			}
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Docking/MacSplitterWidget.cs
```diff
@@ -165,15 +171,55 @@ public override void MouseDragged (NSEvent theEvent)
...
 		}
 
-		static Gdk.FilterReturn Filter (IntPtr xevent, Gdk.Event evnt) => Gdk.FilterReturn.Remove;
+		Gdk.FilterReturn Filter (IntPtr xevent, Gdk.Event evnt)
+		{
+			if (enableGdkEventFiler && !isDisposed && Window != null) {
+				// always recover GDK events after 2 seconds of inactivity, this makes sure
+				// that we don't lose GDK events forever if some other native view steals events
+				// which are required for the drag exit condition.
+				if (Foundation.NSProcessInfo.ProcessInfo.SystemUptime - lastEventTimestamp < 2.0)
```
abock: What happens in the case that, for example, I drag some text from the native editor but do not release the drag within two seconds? Same with say, selection dragging that performs a long document scroll that lasts more than two seconds?

I like the idea of a failsafe here, but we need to be sure that it's not inadvertently cancelling or interfering other operations that may be in flight.

sevoku: This shouldn't affect the editor or scrolling at all, since the filter func is for the splitter only. The only downside of the fail-safe timeout is that if you start dragging (resizing a pad) and hold for 2 seconds without moving the mouse, then it'll recover GDK events. But that's ok too, since the drag is already started (which was the main problem here, it was not possible to drag).

PR #7842, opened 2019-06-07
Bump Roslyn to 3.2.0-beta4-19312-10

sandyarmstrong:


sandyarmstrong:
This depends on https://github.com/xamarin/vs-editor-core/pull/333 (I have a matching md-addins branch to bump to that for this build).

NuGet.config
```diff
@@ -3,6 +3,7 @@
 	<packageSources>
 		<add key="Nuget Official" value="https://api.nuget.org/v3/index.json" />
 		<add key="Roslyn Nightlies" value="https://dotnet.myget.org/F/roslyn/api/v3/index.json" />
+		<add key="roslyn-analyzers" value="https://dotnet.myget.org/F/roslyn-analyzers/api/v3/index.json" />
```
sandyarmstrong: Had to add this because `Microsoft.CodeAnalysis.Features` now has a dependency on `Microsoft.CodeAnalysis.FlowAnalysis.Utilities@2.9.3-beta1.19301.2`.

rodrmoya:
In md-addins#4702 we are bumping to this mono (https://github.com/mono/mono/commits/2019-02) where Roslyn is 3.1.x, so I wonder if we should be in sync?

Therzok:
We've been shipping newer roslyn than mono historically. Our project options version for the language is done based on mono roslyn, not IDE roslyn. We just support newer IDE features and fixes this way.

sandyarmstrong:
@rodrmoya it would be amazing if the Roslyn we used at build time was in sync with the Roslyn we used in the IDE, but I don't think that's possible in the short term. We have been drastically out-of-sync for a very long time, so really this smaller gap is a huge improvement. :)

But yes, as @Therzok says, we need to be able to quickly update Roslyn in the IDE in order to take fixes and various IDE/editor features. In the long-term, if there are improvements in upstream Roslyn's processes, the ideal would be for Mono to be able to update with the same rapidity.

Therzok (approved):

NuGet.config
Therzok: https://github.com/dotnet/roslyn/pull/36283/files#diff-95cdc8ff82808b3e443f24a26fd05afbR128

abock (approved):

PR #7847, opened 2019-06-08
[IDE] Restore the memento after the toolbar has been attached

iainx:
The memento needs to be restored after the toolbar has been attached to the window to take it into consideration in the size calculations. This caused problems when combined with Workbench.Present: If the workbench window was first shown using Present, subsequent calls to EnsureLayout would never update the window size to take the toolbar height into consideration, which caused the mouse pointer offset bugs.

Also unify the confusion between Workbench.Show and Workbench.Present, deprecating the former.

Fixes VSTS #911290

### DIFF
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -1281,7 +1281,7 @@ namespace MonoDevelop.MacIntegration
 			y += GetTitleBarHeight (w);
 			var dr = FromDesktopRect (new Gdk.Rectangle (x, y, width, height));
 			var r = w.FrameRectFor (dr);
-
+			w.SetFrame (r, true);
 			base.PlaceWindow (window, x, y, width, height);
 		}
 
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebugCommands.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebugCommands.cs
@@ -553,7 +553,7 @@ namespace MonoDevelop.Debugger
 		protected override void Run ()
 		{
 			if (!IdeApp.Workbench.Visible) {
-				IdeApp.Workbench.Show ();
+				IdeApp.Workbench.Present ();
 			}
 			var breakpointsPad = IdeApp.Workbench.Pads.FirstOrDefault (p => p.Id == "MonoDevelop.Debugger.BreakpointPad");
 			if (breakpointsPad != null) {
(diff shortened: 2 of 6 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/DesktopService.cs
```diff
@@ -398,7 +398,9 @@ internal MainToolbarController CreateMainToolbar (Gtk.Window window)
 
 		internal void AttachMainToolbar (Gtk.VBox parent, MainToolbarController toolbar)
 		{
+			Console.WriteLine ("Attach main toolbar in desktop");
 			PlatformService.AttachMainToolbar (parent, toolbar.ToolbarView);
+			Console.WriteLine ("Attached");
```
slluis: Debug code

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Shell/DefaultWorkbench.cs
```diff
@@ -585,8 +585,9 @@ void ApplicationNameChanged (object sender, EventArgs e)
 						requestedBounds.Width = MinimumWidth;
 					if (requestedBounds.Height < MinimumHeight)
 						requestedBounds.Height = MinimumHeight;
+
```
sevoku: Unrelated / formatting only changes in this file?

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -1276,14 +1277,16 @@ internal override void AddChildWindow (Gtk.Window parent, Gtk.Window child)
 
 		internal override void PlaceWindow (Gtk.Window window, int x, int y, int width, int height)
 		{
-			if (window.GdkWindow == null)
-				return; // Not yet realized
+			if (window.GdkWindow == null) {
```
sevoku: Just formatting changes here?

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -1276,14 +1277,16 @@ internal override void AddChildWindow (Gtk.Window parent, Gtk.Window child)
...
+				return;
+			}
 
 			NSWindow w = GtkQuartz.GetWindow (window);
 			y += GetTitleBarHeight (w);
 			var dr = FromDesktopRect (new Gdk.Rectangle (x, y, width, height));
 			var r = w.FrameRectFor (dr);
 			w.SetFrame (r, true);
```
sevoku: So, the change in 788db1611148a030de9df8b2ae09eb453aa69e94 is not required after all? Please rebase to resolve conflict and revert that commit, if that's not required.

sevoku: Yep, after testing that doesn't seem to be required, works perfectly fine!

iainx:
@slluis @sevoku Formatting and debugging changes removed and reverted the previous "fix"

PR #7848, opened 2019-06-10
Support multi-target framework intellisense

mrward:
Type system service now supports multi-target frameworks.
Path bar supports multi-target framework selection.

MultiTargetFrameworkDropdown

MultiTargetIntellisense

Framework execution targets shown in main toolbar.

ExecutionTarget

SelectFrameworkExecutionTarget

Fixes VSTS #572311 - Multi-framework support in Roslyn workspace
Fixes VSTS #572342 - Support picking target framework for intellisense
Fixes VSTS #937105 Show (Multiple TargetFrameworks) in project options
Fixes VSTS #572309 - Execution target support for multi-framework projects
Fixes VSTS #935387 - System.NullReferenceException exception in MonoDevelop.Projects.DotNetProject.<OnGetReferencedAssemblies>d__XXX.MoveNext() - Transitive references null reference exception fixed since this code no longer existx.
Fixes VSTS #572307 - Ensure navigation/refactoring operations on multi-frameworks projects are coalesced
Fixes VSTS #843603 Files with build action 'None' are still used by Intellisense

 - [x] Cache framework specific configuration
     - Ensure the cache is refreshed when project options page changes config settings.
 - [x] OnExecute project extension should be framework aware
        - Do not want ASP.NET Core project extension checking for certificate status when .NET Framework is the execution target.
  - [x] Text editor path bar does not update after changing from a single framework to multiple.
    - [x] See type system errors after saving the edited .csproj - [Gist](https://gist.github.com/mrward/75870b499d8c111597f5b41cf1ad4b53) 
    - [x] Changing frameworks in .csproj does not always update the path bar. Sometimes see no owner projects, sometimes no workspace changed event is fired.
        - [x] After the project file reload the OpenDocumentReferences stored in the type system have an associated Project but its ParentSolution is null. This results in TryOpenDocumentInWorkspace doing nothing which presumably then breaks the editor path bar.


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ItemFrameworkConfigurationSelector.cs
```diff
@@ -0,0 +1,55 @@
...
+		public override ItemConfiguration GetConfiguration (IConfigurationTarget target)
+		{
+			var original = configuration.GetConfiguration (target);
+
+			if (original is DotNetProjectConfiguration itemConfiguration) {
+				ItemConfiguration.ParseConfigurationId (original.Id, out string name, out string platform);
+				var copy = new DotNetProjectConfiguration (name, platform, framework);
+				copy.CopyFrom (itemConfiguration);
+
+				return copy;
```
slluis: That is weird. A selector is supposed to return one of the configurations in the provided target, not create a new one. Why is this required?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ItemConfiguration.cs
```diff
@@ -84,17 +91,28 @@ public static void ParseConfigurationId (string id, out string name, out string
...
+					return name;
 			}
 		}
 		
 		public string Platform {
 			get { return platform ?? string.Empty; }
 		}
 
+		public string Framework {
+			get { return framework; }
```
slluis: Isn't the concept of framework is specific to .net? in that case this should be part of DotNetProjectConfiguration.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ItemFrameworkConfigurationSelector.cs
mrward: Honestly this is a bit of a hack which I do not like.

The framework information is needed when running an MSBuild target so I was trying to get this to the DotNetProject when it runs MSBuild to get references etc without adding an extra framework parameter to each method. Currently the active selector in the workbench has no knowledge of the framework so the type system service is cheating here and creates another selector and sets the framework it wants.

main/src/addins/CSharpBinding/MonoDevelop.CSharp/CSharpPathedDocumentExtension.cs
```diff
@@ -614,5 +618,32 @@ void CancelUpdatePath ()
 			src.Cancel ();
 			src = new CancellationTokenSource ();
 		}
+
+		class OwnerProjectInfo
```
Therzok:
```suggestion
		sealed class OwnerProjectInfo
```

Therzok: Maybe, since we implemented Equals and GetHashCode, we should also:
```suggestion
		class OwnerProjectInfo : IEquatable<T>
```

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.NodeBuilders/DependenciesNodeBuilder.cs
```diff
@@ -62,22 +63,37 @@ public override bool HasChildNodes (ITreeBuilder builder, object dataObject)
...
+		public override void OnNodeRemoved (object dataObject)
+		{
+			var dependenciesNode = (DependenciesNode)dataObject;
+			dependenciesNode.PackageDependencyCache.PackageDependenciesChanged -= OnPackageDependenciesChanged;
+		}
+
+		void OnPackageDependenciesChanged (object sender, EventArgs e)
+		{
+			var cache = (PackageDependencyNodeCache)sender;
+			var project = cache.Project;
```
Therzok: Nit: cache.Project not used elsewhere, seems like a verbose local.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
```diff
@@ -1193,50 +1230,15 @@ protected override async Task OnClearCachedData ()
...
-			SetProperty (metadata, "ReferenceSourceTarget", reference.ReferenceSourceTarget);
-
-			return new AssemblyReference (path, metadata);
-		}
-
-		void SetProperty (MSBuildPropertyGroupEvaluated metadata, string name, string value)
-		{
-			var property = new MSBuildPropertyEvaluated (MSBuildProject, name, value, value);
-			metadata.SetProperty (name, property);
+			await base.OnClearCachedData ();
```
Therzok: Not sure this method needs to be async anymore.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -1571,16 +1577,23 @@ static string[] GetTargetFrameworks (MSBuildProject project)
 			return null;
 		}
 
+		internal protected IEnumerable<string> GetTargetFrameworks ()
+		{
+			var frameworks = GetTargetFrameworks (MSBuildProject);
+			return frameworks ?? Enumerable.Empty<string> ();
```
Therzok:
```suggestion
			return frameworks ?? Array.Empty<string> ();
```

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SdkProjectExtension.cs
```diff
@@ -266,86 +267,108 @@ internal protected override async Task OnReevaluateProject (ProgressMonitor moni
...
-			ConfigurationSelector configuration,
-			System.Threading.CancellationToken token)
+		internal protected override string OnGetDefaultBuildAction (string fileName)
 		{
-			var references = new List<AssemblyReference> ();
-
-			var traversedProjects = new HashSet<string> (StringComparer.OrdinalIgnoreCase);
-			traversedProjects.Add (Project.ItemId);
+			string include = MSBuildProjectService.ToMSBuildPath (Project.ItemDirectory, fileName);
+			var globItems = Project.MSBuildProject.FindGlobItemsIncludingFile (include).ToList ();
```
Therzok: Maybe we can do:
```suggestion
			var globItems = Project.MSBuildProject.FindGlobItemsIncludingFile (include).SingleOrDefault ();
```

To avoid matching all the items.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SdkProjectExtension.cs
```diff
@@ -266,86 +267,108 @@ internal protected override async Task OnReevaluateProject (ProgressMonitor moni
...
+			SolutionItemRunConfiguration runConfig)
 		{
-			foreach (var reference in await base.OnGetReferences (configuration, token).ConfigureAwait (false)) {
-				if (!reference.IsProjectReference) {
-					if (includeNonProjectReferences) {
-						references.Add (reference);
-					}
-					continue;
-				}
+			var targets = new List<ExecutionTarget> ();
```
Therzok:
```suggestion
			var targets = new List<ExecutionTarget> (Project.TargetFrameworkMonikers.Length);
```

Therzok:
LGTM!

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SdkProjectExtension.cs
mrward: Not sure I can change this. There may not be any matches - but could use FirstOrDefault instead. If there is more than one match we do not know what to return so fallback to asking the project for its default build action. Although not sure I have seen any examples of multiple matches here.

Therzok: Yeah, was just trying to avoid a new collection allocation, as it seems redundant.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Gui/DotNetCoreRuntimeOptionsPanelWidget.cs
```diff
@@ -49,43 +49,49 @@ public DotNetCoreRuntimeOptionsPanelWidget (DotNetProject project)
...
-				if (cmp != 0)
-					return cmp;
-				return string.CompareOrdinal (y.Id.Version, x.Id.Version);
-			});
+				//sort by id ascending, version descending
+				frameworks.Sort ((x, y) => {
+					var cmp = string.CompareOrdinal (x.Id.Identifier, y.Id.Identifier);
+					if (cmp != 0)
+						return cmp;
+					return string.CompareOrdinal (y.Id.Version, x.Id.Version);
```
Therzok: I know this was already broken in previous versions and not a big cause of worry for now, but, for example, given the same identifier, a framework list with versions `8.0`, `9.0`, `10.0` would be sorted incorrectly, as:
```
9.0
8.0
10.0
```

Filed https://devdiv.visualstudio.com/DevDiv/_workitems/edit/950504

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/TargetFrameworkExecutionTarget.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace MonoDevelop.Core.Execution
+{
+	internal class TargetFrameworkExecutionTarget : ExecutionTarget
+	{
+		public TargetFrameworkExecutionTarget (TargetFrameworkMoniker framework)
+		{
+			Id = framework.ShortName;
+			Name = framework.ShortName;
+			Framework = framework;
+			Image = "md-framework-dependency";
```
Therzok: Should we have a constant for this in Stock?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
```diff
@@ -2087,6 +2088,131 @@ protected override async Task OnReevaluateProject (ProgressMonitor monitor)
...
+				return frameworkSpecificConfigurations.TryGetValue ((name, platform, framework), out configuration);
+		}
+
+		void CacheFrameworkSpecificConfiguration (string name, string platform, string framework, DotNetProjectConfiguration configuration)
+		{
+			lock (frameworkSpecificConfigurationsLock)
+				frameworkSpecificConfigurations [(name, platform, framework)] = configuration;
+		}
+
+		object frameworkSpecificConfigurationsLock = new object ();
```
Therzok:
```suggestion
		readonly object frameworkSpecificConfigurationsLock = new object ();
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.OptionPanels/RuntimeOptionsPanel.cs
```diff
@@ -72,46 +72,52 @@ public RuntimeOptionsPanelWidget (DotNetProject project, IEnumerable<ItemConfigu
...
+					cmp = string.CompareOrdinal (y.Id.Version, x.Id.Version);
+					if (cmp != 0)
+						return cmp;
+					return string.CompareOrdinal (x.Id.Profile, y.Id.Profile);
+				});
+
+				for (int i = 0; i < frameworks.Count; i++) {
+					var fx = frameworks[i];
+					if (project.TargetFramework.Id == fx.Id) {
+						string name = notInstalled? GettextCatalog.GetString ("{0} (Not installed)", fx.Name) : fx.Name;
```
Therzok:
```suggestion
						string name = notInstalled ? GettextCatalog.GetString ("{0} (Not installed)", fx.Name) : fx.Name;
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.MetadataReferenceHandler.cs
```diff
@@ -65,57 +65,70 @@ ImmutableArray<MonoDevelopMetadataReference> CreateDefaultMetadataReferences ()
...
-				if (!(proj is MonoDevelop.Projects.DotNetProject netProject)) {
-					return CreateDefaultMetadataReferences ();
-				}
+				if (!string.IsNullOrEmpty (framework))
+					config = new MonoDevelop.Projects.DotNetProjectFrameworkConfigurationSelector (config, framework);
 
-				var data = new AddMetadataReferencesData {
-					Result = new List<MonoDevelopMetadataReference> (),
+				var data = new AddReferencesData {
+					References = new List<MonoDevelopMetadataReference> (),
```
Therzok: Note to file a bug to use `ImmutableArray.Builder`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectDataMap.cs
```diff
@@ -48,56 +48,89 @@ public ProjectDataMap (MonoDevelopWorkspace workspace)
 				Workspace = workspace;
 			}
 
-			internal ProjectId GetId (MonoDevelop.Projects.Project p)
+			internal ProjectId GetId (MonoDevelop.Projects.Project p, string framework = null)
 			{
 				lock (gate) {
-					return projectIdMap.TryGetValue (p, out ProjectId result) ? result : null;
+					if (projectIdMap.TryGetValue (p, out var frameworkMappings)) {
+						var map = frameworkMappings.FirstOrDefault (f => f.Framework == framework);
```
Therzok: Don't want to prematurely optimize, but `GetId` is called consistently a lot throughout the IDE usage.

Can we get rid of the lambda capture and LINQ usage?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectDataMap.cs
```diff
@@ -164,6 +214,18 @@ internal ProjectId[] GetProjectIds ()
 					return projectDataMap.Keys.ToArray ();
 				}
 			}
+
+			class FrameworkMap
```
Therzok: Nit:
```suggestion
			sealed class FrameworkMap
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectSystemHandler.cs
```diff
@@ -175,29 +199,45 @@ VersionStamp GetVersionStamp (MonoDevelop.Projects.Project project)
...
+				if (dotNetProject == null)
+					return null;
+
 				var workspace = Runtime.PeekService<RootWorkspace> ();
-				return workspace != null ? p.GetConfiguration (workspace.ActiveConfiguration) as MonoDevelop.Projects.DotNetProjectConfiguration : p.DefaultConfiguration as MonoDevelop.Projects.DotNetProjectConfiguration;
+				var config = workspace != null ? p.GetConfiguration (workspace.ActiveConfiguration) as DotNetProjectConfiguration : p.DefaultConfiguration as DotNetProjectConfiguration;
+				if (config == null || string.IsNullOrEmpty (framework))
+					return config;
+
+				return await dotNetProject.GetConfigurationAsync (config.Name, config.Platform, framework);
```
Therzok:
```suggestion
				return await dotNetProject.GetConfigurationAsync (config.Name, config.Platform, framework).ConfigureAwait (false);
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectSystemHandler.cs
```diff
@@ -97,11 +97,27 @@ class SolutionData
...
+				MonoDevelop.Projects.Project p,
+				CancellationToken token,
+				MonoDevelop.Projects.Project oldProject,
+				ProjectCacheInfo cacheInfo,
+				string framework)
+			{
+				var projectId = projectMap.GetOrCreateId (p, oldProject, framework);
 
-				var config = GetDotNetProjectConfiguration (p);
+				var config = await GetDotNetProjectConfiguration (p, framework);
```
Therzok:
```suggestion
				var config = await GetDotNetProjectConfiguration (p, framework).ConfigureAwait (false);
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -1497,24 +1498,26 @@ void OnProjectModified (object sender, MonoDevelop.Projects.SolutionItemModified
...
-							}
-							try {
-								lock (projectModifyLock) {
-									ProjectInfo newProjectContents = t.Result;
-									newProjectContents = AddVirtualDocuments (newProjectContents);
-									OnProjectReloaded (newProjectContents);
-									Runtime.RunInMainThread (() => IdeServices.TypeSystemService.UpdateRegisteredOpenDocuments ()).Ignore();
+						foreach (string framework in ProjectHandler.GetFrameworks (project)) {
+							var projectInfo = ProjectHandler.LoadProject (project, cts.Token, null, null, framework).ContinueWith (t => {
+								if (t.IsCanceled)
```
Therzok: Modify the ContinueWith call to harden and simplify the code a bit:

Specify `TaskScheduler` (`Task.Default` from the looks of it) for deterministic scheduling of the continuation.

Specify `TaskContinuationOptions.NotOnCanceled` and remove this IsCanceled check.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService_WorkspaceHandling.cs
```diff
@@ -511,14 +511,31 @@ async void OnSolutionItemAdded (object sender, MonoDevelop.Projects.SolutionItem
...
 					if (oldProject != null) {
-						projectInfo = ws.AddVirtualDocuments (projectInfo);
-						ws.OnProjectReloaded (projectInfo);
+						oldProjectIds = ws.GetProjectIds (oldProject).ToHashSet ();
 					}
-					else {
-						ws.OnProjectAdded (projectInfo);
+
+					foreach (string framework in ws.GetFrameworks (project)) {
+						var projectInfo = await ws.LoadProject (project, CancellationToken.None, oldProject, framework);
```
Therzok: Can we not do this work on parallel?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/WorkspaceFilesCache.cs
```diff
@@ -78,34 +78,57 @@ void LoadCache (Solution sol)
...
-					TryDeleteFile (cacheFilePath);
-					continue;
 				}
+
+				if (cacheList != null)
+					cachedItems [projectFilePath] = cacheList;
+			}
+		}
+
+		internal IEnumerable<string> GetFrameworks (Project p)
```
Therzok: Wonder if there is a way to deduplicate this into an extension method without polluting the internal API.

Therzok:
Done with review pass 1. Looks really good!

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Gui/DotNetCoreRuntimeOptionsPanelWidget.cs
Therzok: Marking as resolved, as I don't think it's reasonable to expect a fix in this PR.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.MetadataReferenceHandler.cs
Therzok: Nothing to gain here, did some benchmarking.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
mrward: What should TaskScheduler be set to?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService_WorkspaceHandling.cs
mrward: Just wondering if there is a benefit to doing it in parallel. Each framework would be locked when accessing project information, such as getting reference assemblies for that framework.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
Therzok: TaskScheduler.Default. The code seems to want the continuation to run on the threadpool.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService_WorkspaceHandling.cs
Therzok: Hmmm, the only thing I'm thinking about here is that if we have 10 projects targeting 2 frameworks, we now do twice the work serially.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/TargetFrameworkExecutionTarget.cs
mrward: This icon is currently in the NuGet addin. I guess it needs to be moved to Core since it is being used there.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/WorkspaceFilesCache.cs
mrward: For now I have changed the WorkspaceFilesCache to use the method on the MonoDevelopWorkspace.ProjectSystemHandler.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/TargetFrameworkExecutionTarget.cs
mrward: Making this a stock icon does not help here. Stock Icons are in MonoDevelop.Ide not MonoDevelop.Core so I cannot use it from here.

Therzok: Gotcha! Thanks!

Therzok:
LGTM iteration 2

mrward:
Test failure is happening on master so not specific to this branch.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -2919,21 +2932,21 @@ ProjectConfiguration LoadConfiguration (ProgressMonitor monitor, ConfigData cgrp
 			return config;
 		}
 
-		MSBuildProjectInstance CreateProjectInstaceForConfiguration (string conf, string platform, bool onlyEvaluateProperties = true)
+		protected internal MSBuildProjectInstance CreateProjectInstanceForConfiguration (string conf, string platform, string framework = null, bool onlyEvaluateProperties = true)
```
slluis: Does it need to be protected? It could just be internal.

mrward: Changed it to be internal not protected.

Therzok (approved):

Therzok:
Took it for a spin, works properly from what I've tested! Nice work!

sgmunn:
UI test for project system is from Nuget changes and is fixed in master.
.Net core tests are from mono bug. 

Neither failings are related as far as we know.

sgmunn (approved):
works good

slluis (approved):

PR #7849, opened 2019-06-10
Fixes new editor debugger tooltip not showing errors

DavidKarlas:
Fix 842970: [Feedback] Editing tooltips are shown instead of debugging tooltips
Fix 855907: [Feedback] Debugging Variable Values Regression, Visual Studio for Mac Preview 3 New Editor Preview
VSWindows doesn't show intelisense tooltips, but old editor did and it makes sense to show intelisense tooltips rather then nothing... But last time I was bringing this feature to new editor I didn't copy old editor logic instead took shortcuts that resulted in this 2 bugs, now old editor logic is copied and all should be good going forward.
Also brought another related feature I liked in old editor, which is evaluating selected text when hovering selected text, useful for evaluating parts of `if` statements and to evaluate method calls.

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.VSTextView/QuickInfo/DebuggerQuickInfoSource.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.VSTextView/QuickInfo/DebuggerQuickInfoSource.cs
@@ -87,56 +87,70 @@ namespace MonoDevelop.Debugger.VSTextView.QuickInfo
 			var triggerPoint = session.GetTriggerPoint (textBuffer);
 			var point = triggerPoint.GetPoint (snapshot);
 
+			if (!view.Selection.IsEmpty) {
+				foreach (var span in view.Selection.SelectedSpans) {
+					if(span.Contains(point)) {
+						await EvaluateAndShowTooltipAsync (session, view, point, new DataTipInfo (snapshot.CreateTrackingSpan (span, SpanTrackingMode.EdgeInclusive), snapshot.GetText (span)), cancellationToken);
+						return null;
+					}
+				}
+			}
+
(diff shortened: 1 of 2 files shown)


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.VSTextView/QuickInfo/DebuggerQuickInfoSource.cs
```diff
@@ -87,56 +87,70 @@ public async Task<QuickInfoItem> GetQuickInfoItemAsync (IAsyncQuickInfoSession s
...
-				await session.DismissAsync ();
-				await provider.joinableTaskContext.Factory.SwitchToMainThreadAsync ();
-				this.lastView = view;
-				val.Name = debugInfo.Text;
-				window = new DebugValueWindow ((Gtk.Window)gtkParent.Toplevel, textDocument?.FilePath, textBuffer.CurrentSnapshot.GetLineNumberFromPosition (debugInfo.Span.GetStartPoint (textBuffer.CurrentSnapshot)), DebuggingService.CurrentFrame, val, null);
-				Ide.IdeApp.CommandService.RegisterTopWindow (window);
-				var bounds = view.TextViewLines.GetCharacterBounds (point);
-				view.LayoutChanged += LayoutChanged;
+				await EvaluateAndShowTooltipAsync (session, view, point, debugInfo, cancellationToken);
+				return null;
```
Therzok: This always returns null?

DavidKarlas: Yes, because we actually Dismiss quickinfo tooltip and display DebugObjectTreeWindow(gtk) on ower own, Windows debugger does same thing

PR #7850, opened 2019-06-10
Fixes VSTS 893547: Extract Interface Dialog buttons don't close dialog

avodovnik:
It appears a [change](https://github.com/mono/monodevelop/commit/5f30ab59a877f97724d68350d0b846822b2c4e23#diff-43ce8a1c5ce56b4b8c9e5bde48d8efe9) removed an explicit call to `Dispose` the dialog. That meant that while the service received and acted upon the return value (i.e. it generated the extracted interface, etc.) the actual dialog was never closed/disposed of. 

I've wrapped it in a `using` statement, which seems to have done the trick.

VSTS link: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/893547


mrward:
Tested this locally and looks good to me. There was an exception typing in the editor due to a missing telemetry logger method but that has been fixed elsewhere, and is flagged in the binary compat failure here.

Therzok:
Yeah, this PR probably needs a rebase.

avodovnik:
> Yeah, this PR probably needs a rebase.

Not sure this will do the trick, but I've rebased it nonetheless :-). Let's see...

abock:

PR #7854, opened 2019-06-10
[Watson][Toolbox] Ensures ItemSize is always positive and log if exception occurs

netonjm:
We are hitting this issue from Catalina users, not tested yet in a Catalina machine.
probably happens because we are calling the Refresh after resize the toolbox pad, and to calculate the item size it a list mode it uses the Frame.width.  
I moved Refresh call after the ShowAll call from the gtk widget, and also added a condition to ensure never our item size is going to be negative, and also added a try/catch with some logging information

Fixes VSTS #905849 - NSInternalInconsistencyException negative or zero item sizes are not supported in the flow layout in MacToolboxWidget

### DIFF
--- a/main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport.Toolbox/MacToolboxWidget.cs
+++ b/main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport.Toolbox/MacToolboxWidget.cs
@@ -270,36 +270,32 @@ namespace MonoDevelop.DesignerSupport.Toolbox
 
 		public void RedrawItems (bool invalidates, bool reloads)
 		{
-			try {
-				NSIndexPath selected = null;
-				if (SelectionIndexPaths.Count > 0) {
-					selected = (NSIndexPath)SelectionIndexPaths.ElementAt (0);
-				}
-				if (IsListMode) {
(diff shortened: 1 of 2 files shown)


main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport.Toolbox/MacToolboxWidget.cs
```diff
@@ -270,36 +270,39 @@ public override void MouseDown (NSEvent theEvent)
...
+				if (reloads) {
+					ReloadData ();
+				}
 
-			if (selected != null) {
-				SelectionIndexPaths = new NSSet (selected);
+				if (selected != null) {
+					SelectionIndexPaths = new NSSet (selected);
+				}
+			} catch (Exception ex) {
```
Therzok: Just to note, in case we have a native exception, we're already logging it and _it will always crash the IDE_.

sevoku: right, @netonjm catching here is not necessary, let's remove that.

netonjm: Fixed!

slluis:

netonjm:
Tested in Catalina and now it seems to work ok!

PR #7856, opened 2019-06-11
Fixes VSTS Bug 911693: New document model dispose logic was regressed

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/911693

Revert "[Ide] Added a disposed token to the document controller API."

This reverts commit 800c10b8d37ebc6aeb9367390bc3015244614701.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
@@ -426,14 +426,19 @@ namespace MonoDevelop.Ide.Gui.Documents
 
 		public void Dispose ()
 		{
-			if (!disposed) {
+			if (disposed)
+				return;
+			try {
+				linkedController?.Dispose ();
+				OnDispose ();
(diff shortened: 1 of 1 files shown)


mkrueger:
I would prefer having the broken document controllers fixed - discussion is on VSTS.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
```diff
@@ -427,9 +413,6 @@ public async Task Initialize (ModelDescriptor modelDescriptor, Properties status
 		public void Dispose ()
 		{
 			if (!disposed) {
-				disposedTokenSource.Cancel ();
-				disposedTokenSource.Dispose ();
-				disposedTokenSource = null;
 				linkedController?.Dispose ();
 				disposed = true;
```
sevoku: Instead of revering the change, can't we simply set `disposed = true;` at the very end of `Dispose()`, so it's still false for `OnDispose()` implementations? And maybe the same for the token...

slluis:
@monojenkins backport release-8.2-preview1

Issue #7825, opened 2019-06-06
Unable to search in open files if assembly browser open


mrward:
Repro:

1). Open a C# project.
2) MyClass.cs and the assembly browser open.
3) Search - Find in files
4) Enter search text and make sure `Look in` is set to `All open files`.
5) Click Find.

Expected result:

Open files searched.

Actual result:

Error message in status bar: "The search could not be finished. The value cannot be null. Parameter path."

OK in VS Mac 8.0.9.

Build Information
Release ID: 801002709
Git revision: ca9e5df825980a539b51d48beaf40166e6dedb3c
Build date: 2019-06-05 19:39:47+00
Build branch: release-8.1
Xamarin extensions: 5d3e608c91fa6e3539c6040d2aba1b24802a30bd

```
Error while search
System.ArgumentNullException: Value cannot be null.
Parameter name: path
  at System.IO.Path.InsecureGetFullPath (System.String path) [0x00212] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-08/external/bockbuild/builds/mono-x64/mcs/class/corlib/System.IO/Path.cs:425 
  at System.IO.Path.GetFullPath (System.String path) [0x00000] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-08/external/bockbuild/builds/mono-x64/mcs/class/corlib/System.IO/Path.cs:283 
  at MonoDevelop.Ide.FindInFiles.FindReplace.FindAll (MonoDevelop.Ide.FindInFiles.Scope scope, MonoDevelop.Core.ProgressMonitor monitor, System.String pattern, System.String replacePattern, MonoDevelop.Ide.FindInFiles.FilterOptions filter, System.Threading.CancellationToken token) [0x00157] in /Users/vsts/agent/2.152.1/work/1/s/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindReplace.cs:124 
  at MonoDevelop.Ide.FindInFiles.FindInFilesDialog+<>c__DisplayClass62_0.<SearchReplace>b__0 () [0x0009f] in /Users/vsts/agent/2.152.1/work/1/s/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/FindInFilesDialog.cs:890 
```

> VS bug [#910920](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/910920)

xamarin-release-manager:
[...]

PR #7858, opened 2019-06-11
Fixes issue #7825 Unable to search in open files if assembly browser

mkrueger:
open


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/Scope.cs
```diff
@@ -276,6 +276,8 @@ public override IEnumerable<FileProvider> GetFiles (ProgressMonitor monitor, Fil
 				if (textBuffer != null) {
 					yield return new OpenFileProvider (textBuffer, document.Owner as Project, document.FileName);
 				} else {
+					if (document.IsViewOnly || document.FileName.IsNullOrEmpty)
```
slluis: You could just check document.IsFile

mkrueger: Does that work for untitled files then ? I don't think I can safe the file name check here.

sevoku:

PR #7859, opened 2019-06-11
Fix 841029: Can't list classes in navigation bar

DavidKarlas:


mrward:
The bug here is that if you have multiple classes defined in a single file you cannot jump to the other class using the path bar if the cursor is currently in one class?

That is the only thing I can see here.

Also seems that in VS Mac 8.1 the old editor crashes when you try to use the path bar to test the same thing.

DavidKarlas:
> The bug here is that if you have multiple classes defined in a single file you cannot jump to the other class using the path bar if the cursor is currently in one class?
Correct

> Also seems that in VS Mac 8.1 the old editor crashes when you try to use the path bar to test the same thing.
No idea, seems to work fine on master.

PR #7860, opened 2019-06-11
[release-8.2][Core] Fix GTK# assembly reference resolution on Mono 6.0

mrward:
Mono 6.0 returns files and directories unsorted which was resulting
in .NET 4.5 being used as the target framework for resolved GTK#
assemblies instead of the oldest target framework 4.0. This caused
some ProjectReference validation tests to fail since they were using
.NET 4.0 projects. To fix this the target framework information is
sorted so older frameworks are checked before newer target frameworks,
so the oldest supported framework is used.


Therzok (approved):

Therzok (approved):

rodrmoya (approved):

mrward:
Only the leak tests are failing/crashing, which is also happening on the release-8.2 branch.

slluis:
@monojenkins backport master

mrward:
@monojenkins backport master

PR #7861, opened 2019-06-11
[IDE] Listen for NSNotification to know if the app was started from Finder

iainx:
On Mac  when the app is started from Finder, the requested files are passed to the app asynchronously, so we need to listen for the event.

Fixes VSTS #737145


netonjm:
Could we include a NotInitialized state like a default enum or some flag?

sevoku:
@iainx what's the status of this?

iainx:
Just finishing it up, but I don't think anyone's going to be happy with it

PR #7865, opened 2019-06-11
Fix NRE in DropDownBoxListWindow

slluis:
Fixes VSTS #916199 - Crashes when selecting item on members DropDown list box


slluis:

PR #7867, opened 2019-06-11
Force High Sierra SDK

slluis:
Just a test


Therzok:
SDK_PATH seems empty.

Therzok:
SDK_PATH seems empty. There is no 10.13 SDK in Xcode 10.1/10.2

PR #7870, opened 2019-06-11
[CSharpBinding] Allow choosing C# 8

sandyarmstrong:
Remove `CSharp8` and `LatestMajor` from unsupported list.

Fixes https://devdiv.visualstudio.com/DevDiv/_workItems/edit/916849


sandyarmstrong:

abock:
Opened the explicit backport because `master` seems to be too much in flux and we want a CI build that works to verify.

mrward:
I guess we have no plans to support this conditionally based on the Mono version?

Therzok:
_is comtemplating bumping minimum mono version to 6.x_

slluis:
@monojenkins backport release-8.2-preview1

sandyarmstrong:
@mrward I think that's a lovely idea, but probably requires more discussion. One concern is adding support for a Mono version that we will never dogfood with.

PR #7872, opened 2019-06-11
[TextEditor] Allow extensions to opt file types into new editor

mhutch:
This adds a new extension point that extensions can use to opt individual file types into the new editor, optionally controlled by feature flags.

This means that extensions can light up new file types in the new editor without changing the new editor itself.

### DIFF
--- a/main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/Properties/MonoDevelop.TextEditor.addin.xml
+++ b/main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/Properties/MonoDevelop.TextEditor.addin.xml
@@ -13,6 +13,11 @@
 		<Description>Context menu for the navigation bar.</Description>
 		<ExtensionNode name="Map" type="MonoDevelop.TextEditor.CommandMappingExtensionNode" />
 	</ExtensionPoint>
+
+	<ExtensionPoint path="/MonoDevelop/TextEditor/SupportedFileTypes" name="Opts in file types to be opened in the new editor">
+		<ExtensionNode name="SupportedFileType" type="MonoDevelop.TextEditor.SupportedFileTypeExtensionNode" />
+	</ExtensionPoint>
+
 <!--
 	<ExtensionPoint path = "/MonoDevelop/TextEditor/ContextMenu/NavigationBar" name = "Navigation bar context menu">
 		<Description>Context menu for the navigation bar.</Description>
@@ -42,7 +47,9 @@
 		<ExtensionNode name="Class" />
 	</ExtensionPoint>
 
-	--><!-- Extensions -->
+	-->
+
+	<!-- Extensions -->
 
   	<Extension path="/MonoDevelop/Ide/Composition">
 		<!--
@@ -50,6 +57,12 @@
 		<Assembly file="../../../bin/Microsoft.VisualStudio.Platform.VSEditor.dll"/>
 		-->
 		<Assembly file="MonoDevelop.TextEditor.dll"/>
+	</Extension>
+
+	<Extension path="/MonoDevelop/TextEditor/SupportedFileTypes">
+		<SupportedFileType id="csharp" extensions=".cs,.csx" />
+		<SupportedFileType id="xaml" extensions=".xaml" FeatureSwitch="DesignersNewEditor" />
+		<SupportedFileType id="androidxml" extensions=".xml,.axml" BuildAction="AndroidResource" FeatureSwitch="DesignersNewEditor" />
 	</Extension>
 
 	<Extension path="/MonoDevelop/Ide/Commands">
--- a/main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/SupportedFileTypeExtensionNode.cs
+++ b/main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/SupportedFileTypeExtensionNode.cs
@@ -1,0 +1,43 @@
+//
+// Copyright (c) Microsoft Corp. (https://www.microsoft.com)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+using Mono.Addins;
+
+namespace MonoDevelop.TextEditor
+{
+	sealed class SupportedFileTypeExtensionNode : ExtensionNode
+	{
+		[NodeAttribute ("extensions", "Comma separated list of file extensions. The file must match one of these or one of the mime types.")]
+		public string [] Extensions { get; private set; }
+
+		[NodeAttribute ("mimeTypes", "Comma separated list of mime types. The file must match one of these or one of the extensions.")]
+		public string [] MimeTypes { get; private set; }
+
+		[NodeAttribute ("buildAction", Description = "If specified, the file must have this build action")]
+		public string BuildAction { get; private set; }
+
+		[NodeAttribute ("featureFlag", Description = "ID of a feature flag that can be used to enable/disable editing of this file type in the new editor")]
+		public string FeatureFlag { get; private set; }
+
+		[NodeAttribute ("featureFlagDefault", Description = "Default value of the feature flag")]
+		public bool FeatureFlagDefault { get; private set; }
+	}
+}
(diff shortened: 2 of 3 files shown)


main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/SupportedFileTypeExtensionNode.cs
```diff
@@ -0,0 +1,43 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+using Mono.Addins;
+
+namespace MonoDevelop.TextEditor
+{
+	class SupportedFileTypeExtensionNode : ExtensionNode
```
Therzok:
```suggestion
	sealed class SupportedFileTypeExtensionNode : ExtensionNode
```

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewDisplayBinding.cs
```diff
@@ -45,56 +45,47 @@ protected override IEnumerable<DocumentControllerDescription> GetSupportedContro
...
-			if (!(fileName.HasExtension (".xml") || fileName.HasExtension (".axml")))
-				return false;
-
-			const string AndroidResourceBuildAction = "AndroidResource";
-			var buildAction = (ownerProject as Project)?.GetProjectFile (fileName)?.BuildAction;
-			return string.Equals (buildAction, AndroidResourceBuildAction, StringComparison.Ordinal);
+			bool BuildActionAndFeatureFlagMatch (SupportedFileTypeExtensionNode node)
+			{
+				if (!string.IsNullOrEmpty (node.BuildAction)) {
+					var buildAction = (modelDescriptor.Owner as Project)?.GetProjectFile (modelDescriptor.FilePath)?.BuildAction;
```
Therzok: This won't work for files from imported targets.

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewDisplayBinding.cs
```diff
@@ -45,56 +45,47 @@ protected override IEnumerable<DocumentControllerDescription> GetSupportedContro
...
+					if (!string.Equals (buildAction, node.BuildAction, StringComparison.OrdinalIgnoreCase)) {
+						return false;
+					}
+				}
+				if (!string.IsNullOrEmpty (node.FeatureFlag)) {
+					if (!(FeatureSwitchService.IsFeatureEnabled (node.FeatureFlag) ?? node.FeatureFlagDefault)) {
```
Therzok: Maybe this should be the first check, as its the least expensive?

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewDisplayBinding.cs
```diff
@@ -45,56 +45,47 @@ protected override IEnumerable<DocumentControllerDescription> GetSupportedContro
...
-			//".cshtml",
-			//".css",
-			//".html",
-			//".js",
-			//".json",
-			//".ts"
-		};
+			bool ExtensionMatch (SupportedFileTypeExtensionNode node) =>
+				node.Extensions != null
+				&& node.Extensions.Any (modelDescriptor.FilePath.HasExtension);
```
Therzok: Prefer explicit lambda arguments to method group. Method Groups have poor code generation on the compiler side.

```suggestion
				&& node.Extensions.Any (ext => modelDescriptor.FilePath.HasExtension (ext));
```

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/Properties/MonoDevelop.TextEditor.addin.xml
```diff
@@ -52,6 +59,12 @@
 		<Assembly file="MonoDevelop.TextEditor.dll"/>
 	</Extension>
 
+	<Extension path="/MonoDevelop/TextEditor/SupportedFileTypes">
+		<SupportedFileType id="csharp" extensions=".cs,.csx" />
+		<SupportedFileType id="xaml" extensions=".xaml" FeatureSwitch="DesignersNewEditor" />
```
garuma: Should we define this on our side instead?

mhutch: That would make sense :)

It doesn't matter if there are duplicate registrations so I'd suggest we merge this PR, you add the change on your side, then we remove this line.

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewDisplayBinding.cs
mhutch: Ugh, I keep forgetting about that one. Can we fix the compiler already?

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewDisplayBinding.cs
mhutch: Those won't show up in the solution tree though, will they? Also I'm just replicating what was already here...

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/Properties/MonoDevelop.TextEditor.addin.xml
garuma: Perfect

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewDisplayBinding.cs
mhutch: Good point, though probably negligible

abock:
FYI, the plan by 8.3 is for the new editor to blacklist types it cannot open (e.g. F#) instead of whitelist types it can.

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/SupportedFileTypeExtensionNode.cs
mhutch: Applied, thanks!

I wonder whether we could come up with a good analyzer for this e.g. classes with no virtuals, private/internal classes with no subclasses.

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/Properties/MonoDevelop.TextEditor.addin.xml
```diff
@@ -52,6 +59,12 @@
...
+		<SupportedFileType id="androidxaml" extensions=".xml,.axml" BuildAction="AndroidResource" FeatureSwitch="DesignersNewEditor" />
```
decriptor: Is this a typo? `androidxaml` we already confuse people enough with the silly `.axml`.

mhutch:
@abock should we merge this in the meantime? I imagine it would still be useful as a mechanism to allow the blocklisted extensions (F#, generic XML, ASPX, MSBuild, .addin.xml, etc) to override the block smoothly when transitioning over.

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/Properties/MonoDevelop.TextEditor.addin.xml
mhutch: yeah, stray x, will remove

DavidKarlas:
@mhutch yes, merge this and I will rebase textmate on this

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/Properties/MonoDevelop.TextEditor.addin.xml
mhutch: That will also allow you to flip the feature switch to on-by-default without touching this file :)

mhutch:
Failures look completely unrelated. Ok to merge?

abock:
@slluis care to review?

slluis:
It looks good to me.

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewDisplayBinding.cs
Therzok: https://github.com/dotnet/roslyn/pull/6642

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewDisplayBinding.cs
```diff
@@ -45,56 +45,48 @@ protected override IEnumerable<DocumentControllerDescription> GetSupportedContro
 				yield break;
 			}
 
-			if (modelDescriptor.FilePath == null || !(IsSupportedFileExtension (modelDescriptor.FilePath) || IsSupportedDesignerFileName (modelDescriptor.FilePath, modelDescriptor.Owner))) {
-				yield break;
-			}
-
-			bool supported = false;
+			var nodes = Mono.Addins.AddinManager.GetExtensionNodes<SupportedFileTypeExtensionNode> ("/MonoDevelop/TextEditor/SupportedFileTypes");
```
Therzok: Maybe we should cache this and listen to extension node change events? No use querying this _every time_.

mhutch: I don't think that's necessary, it's not particularly expensive. The main value of the event is to cache derived data.

abock: @mhutch thoughts? Looks like this does rebuild a list on each call.

Therzok: Batch opening docs will cause wasteful processing over here. Imo, switch to extension node events here.

abock: https://vsmac.dev/918078

abock:
Test failure is Xamarin.Forms timing out. Something about Java.

PR #7874, opened 2019-06-12
Fixes VSTS Bug 901467: Suddenly and silently native crash in the IDE

mkrueger:
because LibGit2Sharp (Ide closed)

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/901467

Running that as blocking operation could lead to locks/very long
delays running repository.Blame in locked context should fix a
possible multi threaded blame issue. However I couldn't reproduce the
issue.
Improved blame code a bit.


sevoku:

PR #7882, opened 2019-06-12
Fixes VSTS 915954: "Rename File to {NewClassName}.cs" code action does not work

avodovnik:
Adding support for the `Rename File to...` code action.

Ran into a similar issue to https://github.com/mono/monodevelop/pull/7823 and took the same workaround approach. Would make sense to discuss this, to see if there is a better (working) solution.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
@@ -1319,23 +1319,7 @@ namespace MonoDevelop.Ide.TypeSystem
 			}
 
 			var guiDoc = documentManager.Documents.FirstOrDefault (d => d.IsFile && d.Name == document.Name);
-
-			// This method can be called by Roslyn or the editor in a context which is not the GTK UI context
-			// that MonoDevelop uses. In that case, before starting async an operation that may queue
-			// task continuations into the current context, we switch to the GTK context, so that
-			// whatever is queued will be dispatched when we run RunPendingEvents.
-			// See: https://github.com/mono/monodevelop/pull/7823
-			var oldContext = SynchronizationContext.Current;
-			try {
-				SynchronizationContext.SetSynchronizationContext (Runtime.MainSynchronizationContext);
-				Task task = guiDoc.IsDirty ? guiDoc.Save () : null;
-				// Can't wait for the task to finish synchronously since doing so would deadlock the UI thread.
-				while (task != null && !task.IsCompleted) {
-					DispatchService.RunPendingEvents (30);
-				}
-			} finally {
-				SynchronizationContext.SetSynchronizationContext (oldContext);
-			}
+			DispatchService.PumpingWait (() => guiDoc.IsDirty ? guiDoc.Save () : null);
 
 			// TODO: the Visual Studio Windows implementation of this also adds an undo unit to the
 			// global undo manager which we don't currently support.
@@ -1345,9 +1329,45 @@ namespace MonoDevelop.Ide.TypeSystem
 				Path.GetExtension (info.Name),
 				nx => !mdProject.Files.Any (p => string.Equals(p.Name, nx, FilePath.PathComparison)));
 
+			var mdFile = mdProject.GetProjectFile (document.Name);
+
 			FileService.RenameFile (guiDoc.FilePath, newName);
+
+			var childrenToRename = GetDependentFilesToRename (mdFile, newName);
+			if (childrenToRename != null) {
+				// we also need to rename children!
+				foreach (var child in childrenToRename) {
+					FileService.RenameFile (child.File.FilePath, child.NewName);
+				}
+			}
+
+			DispatchService.PumpingWait (() => mdProject.SaveAsync (new ProgressMonitor ()));
+
+			base.ApplyDocumentInfoChanged (id, info);
 		}
 
+		static List<(MonoDevelop.Projects.ProjectFile File, string NewName)> GetDependentFilesToRename (
+			MonoDevelop.Projects.ProjectFile file,
+			string newName)
+		{
+			if (!file.HasChildren)
+				return null;
+
+			List<(MonoDevelop.Projects.ProjectFile File, string NewName)> files = null;
+
+			string oldName = file.FilePath.FileName;
+			foreach (MonoDevelop.Projects.ProjectFile child in file.DependentChildren) {
+				string oldChildName = child.FilePath.FileName;
+				if (oldChildName.StartsWith (oldName, StringComparison.CurrentCultureIgnoreCase)) {
+					string childNewName = newName + oldChildName.Substring (oldName.Length);
+
+					if (files == null)
+						files = new List<(MonoDevelop.Projects.ProjectFile projectFile, string name)> ();
+					files.Add ((child, childNewName));
+				}
+			}
+			return files;
+		}
 
 		void FailIfDocumentInfoChangesNotSupported (Document document, DocumentInfo updatedInfo)
 		{
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
@@ -1328,7 +1328,7 @@ namespace MonoDevelop.Ide.TypeSystem
 				.FirstOrDefault (d => d.IsFile
 					&& document.FilePath.Equals (d.FilePath, FilePath.PathComparison));
 
-			DispatchService.PumpingWait (() => guiDoc.IsDirty ? guiDoc.Save () : null);
+			DispatchService.PumpingWait (() => guiDoc.IsDirty ? guiDoc.Save () : Task.CompletedTask);
 
 			// TODO: the Visual Studio Windows implementation of this also adds an undo unit to the
 			// global undo manager which we don't currently support.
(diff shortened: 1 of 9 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -1295,8 +1296,90 @@ protected override void ApplyProjectReferenceRemoved (ProjectId projectId, Proje
...
+				SynchronizationContext.SetSynchronizationContext (oldContext);
+			}
+
+			// TODO: the Visual Studio Windows implementation of this also adds an undo unit to the
+			// global undo manager which we don't currently support.
+
+			var newName = NameGenerator.GenerateUniqueName (
+				Path.GetFileNameWithoutExtension (info.Name),
+				Path.GetExtension (info.Name),
+				nx => mdProject.Files.FirstOrDefault (p => StringComparer.OrdinalIgnoreCase.Compare (p.Name, nx) == 0) == null);
```
Therzok: Please use FilePath.PathComparer here. It uses the right comparison type in a cross platform manner.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -1295,8 +1296,90 @@ protected override void ApplyProjectReferenceRemoved (ProjectId projectId, Proje
...
+
+			FileService.RenameFile (guiDoc.FilePath, newName);
```
Therzok: Can newName be null? We should guard against that

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
Therzok: Also, shouldn't an equals check be way more efficient than comparing? The reason I'm asking is that O(n) string comparisons are expensive on big projects. Maybe we should use string.Equals (a, b, FilePath.PathComparer).

Therzok: We're not doing any sorting, so equals should be enough (and faster!)

avodovnik: Good shout, thanks! For the first pass, I was doing a naive implementation, following the [Visual Studio workspace implementation](http://source.roslyn.io/#Microsoft.VisualStudio.LanguageServices/Implementation/ProjectSystem/Extensions/ProjectItemsExtensions.cs,7f0b7161c3882525). :-) And you're right, we're only looking at equality, so it should be perfectly fine.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
avodovnik: I don't see how, *but* that doesn't make it impossible. I've put in a check higher up now.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -1295,8 +1296,110 @@ protected override void ApplyProjectReferenceRemoved (ProjectId projectId, Proje
...
+				mdProject = GetMonoProject (project);
+				if (mdProject == null) {
+					LoggingService.LogWarning ("Couldn't find project for newly file file {0} (Project {1}).",
+						info.Name, info.Id.ProjectId);
+
+					return;
+				}
+			}
+
+			var guiDoc = documentManager.Documents.FirstOrDefault (d => d.IsFile && d.Name == document.Name);
```
slluis: You should be comparing document.FilePath

slluis:
Looks good. We'll need unit tests.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
avodovnik: To be honest, I don't think this is right as it is. I don't think it's trying to compare File Paths... even the "original" Windows implementation seems to be comparing names http://source.roslyn.io/#Microsoft.VisualStudio.LanguageServices/Implementation/ProjectSystem/VisualStudioWorkspaceImpl.cs,1168 

What's also weird is that the updated `DocumentInfo`'s name is actually just that, a name, without the path. I think I might need to rethink what is correct here.

slluis: In any case, if you want to compare the file name, you should use document.FilePath.FileName.

avodovnik:
This was a complicated one (at least for me), so I wanted to make sure @Therzok's commit message is also noted here in the conversation:

> After debugging with @avodovnik, we found out that 2 registrations were done for the same document, one still pointing to the old path.
> 
> Subscribing to the file document controller events on initialization fixes this, as the handler that was checking renames is subscribed there.
> 
> Avoid manually saving a project, as there is a method that does that in TryApplyChanges.
> 
> The Freeze/Thaw events are required because the roslyn file associations need to happen before the text buffer registration, otherwise the text buffer is associated with the miscellaneous files workspace  as our workspace does not contain the new document file path yet.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -1295,8 +1298,124 @@ protected override void ApplyProjectReferenceRemoved (ProjectId projectId, Proje
...
+					return;
+				}
+			}
+
+			var guiDoc = documentManager
+				.Documents
+				.FirstOrDefault (d => d.IsFile
+					&& document.FilePath.Equals (d.FilePath, FilePath.PathComparison));
+
+			DispatchService.PumpingWait (() => guiDoc.IsDirty ? guiDoc.Save () : null);
```
Therzok:
```suggestion
			DispatchService.PumpingWait (() => guiDoc.IsDirty ? guiDoc.Save () : Task.CompletedTask);
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -1295,8 +1298,124 @@ protected override void ApplyProjectReferenceRemoved (ProjectId projectId, Proje
...
+			// global undo manager which we don't currently support.
+
+			var newName = NameGenerator.GenerateUniqueName (
+				Path.GetFileNameWithoutExtension (info.Name),
+				Path.GetExtension (info.Name),
+				nx => !mdProject.Files.Any (p => string.Equals (p.Name, nx, FilePath.PathComparison)));
+
+			var mdFile = mdProject.GetProjectFile (document.FilePath);
+			if (mdFile == null) {
+				// TODO: log this
```
Therzok: Maybe just Debug.Assert?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -1295,8 +1298,124 @@ protected override void ApplyProjectReferenceRemoved (ProjectId projectId, Proje
...
+				return;
+			}
+
+			DispatchService.PumpingWait (() => {
+				return Runtime.RunInMainThread (async () => {
```
Therzok:
```suggestion
				return Runtime.RunInMainThread (() => {
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -1295,8 +1298,124 @@ protected override void ApplyProjectReferenceRemoved (ProjectId projectId, Proje
...
+
+					if (files == null)
+						files = new List<(MonoDevelop.Projects.ProjectFile projectFile, string name)> ();
+					files.Add ((child, childNewName));
+				}
+			}
+			return files;
+		}
+
+		void FailIfDocumentInfoChangesNotSupported (Document document, DocumentInfo updatedInfo)
```
Therzok:
```suggestion
		static void FailIfDocumentInfoChangesNotSupported (Document document, DocumentInfo updatedInfo)
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService_WorkspaceHandling.cs
```diff
@@ -310,6 +310,7 @@ public IDisposable RegisterOpenDocument (WorkspaceObject owner, FilePath filePat
 
 		void TryRegisterOpenDocument (OpenDocumentReference reference)
 		{
+ 			Console.WriteLine ($"Try Register Open Document: {reference.FilePath}");
```
Therzok: few more writelines here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
avodovnik: Inside the `if` statement? Is that what we're doing for this? Or is there a logging pattern?

Therzok: You could use logging service or Debug.Assert, when it should never happen.

avodovnik: It's still *possible* that it happens... :-) Let's use Logging Service, then.

Therzok:
Still needs tests

slluis:
@avodovnik this bug is targeting 8.2, so it should have been backported to release-8.2.

slluis:

PR #7887, opened 2019-06-12
Bump proppy to the latest master, to sync with designer

BretJohnson:


netonjm:
would we also need to bump designer in md-addins?

BretJohnson:
Yeah, we'll want to bump the designer as well.
@garuma - what's the process for that? Alan / you bump it manually every few days, to get the latest in master builds?

garuma:
I will be doing an insertion in master tomorrow so we can get this included at that time as well.

slluis:
What about 8.2? do we need to track the d16-2 branch?

garuma:
@slluis this is only master for now, we are still evaluating if it's proppy is ready at all for us and if not we will be doing this in 8.3

PR #7888, opened 2019-06-12
[dotnetcore/aspnetcore] Hands AppendTargetFrameworkToOutputPath

jtorres:
When AppendTargetFrameworkToOutputPath  is true (default) we have to hide target framework id in the OptionsDialog, otherwise will be appended and it duplicated

+ Code formatting

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/810355

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProjectConfiguration.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProjectConfiguration.cs
@@ -1,4 +1,4 @@
-//
+﻿//
 // DotNetProjectConfiguration.cs
 //
 // Author:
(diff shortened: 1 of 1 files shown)


mrward (commented):
Should we add tests?

Will this break the DotCoreProjectExtension? It looks like it uses the OutputDirectory there to find the path to the assembly - https://github.com/mono/monodevelop/blob/master/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs#L160

Will CompiledOutputName work now? That seems to append the OutputDirectory to the assembly name to find the path.

I am just wondering if we need to introduce a new property which is the output directory that is used in the project options dialog instead of changing the OutputDirectory on the project configuration. Or have the logic in the project options dialog so it checks the AppendTargetFrameworkToOutputPath and has the logic implemented in the DotNetProjectConfiguration. Not sure what is better here.

Currently we are reading the MSBuild `OutputPath` and use that as the OutputDirectory but not in all cases now. My worry is that some code will be expecting the OutputDirectory to be the real OutputPath.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProjectConfiguration.cs
```diff
@@ -27,98 +27,82 @@
...
-			PublicSign = pset.GetValue<bool> (nameof(PublicSign));
-			assemblyKeyFile = pset.GetPathValue ("AssemblyOriginatorKeyFile", FilePath.Empty);
-			if (string.IsNullOrEmpty (assemblyKeyFile))
-				assemblyKeyFile = pset.GetPathValue ("AssemblyKeyFile", FilePath.Empty);
+			OutputAssembly = pset.GetValue ("AssemblyName");
+			appendTargetFrameworkToOutputPath = pset.GetValue<bool> ("AppendTargetFrameworkToOutputPath");
+			if (appendTargetFrameworkToOutputPath)
+				// if appendTargetFrameworkToOutputPath is true (default value) then "OutputPath" property
+				// will append TargetFramework.Id to default output directory (i.e. bin/($Configuration)/netcoreapp30)
+				// and OptionsDialog - output directory hides TargtetFramework.ID, otherwise TargetFramework will be appended again
```
Therzok:
```suggestion
				// and OptionsDialog - output directory hides TargetFramework.ID, otherwise TargetFramework will be appended again
```

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProjectConfiguration.cs
```diff
@@ -27,98 +27,82 @@
...
-		public bool DelaySign {
-			get { return delaySign; }
-			set { delaySign = value; }
-		}
-
+		public bool SignAssembly { get; set; } = false;
+		public bool DelaySign { get; set; } = false;
 		public bool PublicSign { get; set; }
 
 		internal string OldAssemblyKeyFile {
```
Therzok: Can be removed completely.

PR #7893, opened 2019-06-13
[Core] Report SDK resolution errors

mrward:
If an SDK cannot be resolved using the MSBuild SDK resolvers then
an exception is now thrown so the error is reported in the IDE.
Previously the error would be logged in the IDE log and the error
displayed in the IDE would usually be a confusing 'Invalid
configuration mapping'. This also makes the behaviour more consistent
with Visual Studio on Windows.

Before:

No load error. Solution pad shows:

CurrentBehaviour

After:

Load error dialog on opening solution.

SdkNotFoundLoadError

Solution pad error tooltip.

SolutionPadSdkNotFoundError

Multiple project load errors.

MultipleSdkNotFoundErrors

Note that the error text here is generated by the SDK resolvers not by VS Mac.

VS Win:

Does not show error information on loading the solution.

VSWinProjectLoad

On reloading the project an error dialog is displayed.

ReloadError

Fixes VSTS #588773 - Add support for NuGet SDK resolver


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/SdkResolution.cs
```diff
@@ -79,17 +80,26 @@ public static SdkResolution GetResolver (TargetRuntime runtime)
 				throw;
 			}
 
+			StringBuilder errorMessage = null;
```
Therzok: Please use [StringBuilderCache](http://source.monodevelop.com/#MonoDevelop.Core/MonoDevelop.Core/StringBuilderCache.cs,43)

mrward: Changed to use the StringBuilderCache.

rodrmoya (approved):

Therzok (approved):

sgmunn:
@mrward a suggestion here, if the dialog is still shown (master might have disabled showing that anyway) is to derive the exception from UserException which will show a short header and the details of the error in a nicer formatted way.

mrward:
This dialog will be shown since this is an error not a warning about an unsupported project. The change in master was not to show the dialog for unsupported project types, which have a special exception.

I can take a look at using the UserException. We would need to add a short message to the exception, something better than VS Win's: 'The project cannot be opened by the project system...', about not being able to resolve the SDK.

Also it looks like the tooltip in the Solution pad would then not have the full details of the exception, just the short message we added, since UserException.Message does not include the details part of the error.

The full detailed errors is only shown in the dialog which is only shown on loading the solution. If you edit the .csproj then you only get the error in the solution pad. So you will only get the full set of errors on initial load. Not sure if that is a good or bad thing. I guess it depends if the short error message has enough information on its own.

mrward:
Changed to use a UserException. Screenshots updated.

sgmunn (approved):

mrward:

mrward:

PR #7895, opened 2019-06-13
Updated reference to mono/debugger-libs@e7061f6

xamarin-release-manager:
Updated reference to mono/debugger-libs@e7061f6


slluis:
Closing since debugger-libs has already been bumped.

PR #7901, opened 2019-06-14
[Perf Tests] Add type system perf test

mrward:
- Added a perf tests for the type system service loading.
 - Ensure UI Tests clear the tmp directory before running a test to prevent old files being used.

Optimisations:

 - Improve ProjectFile.get_ProjectVirtualPath efficiency (it's always computed, slow)
   - Not done. This value is used once by the type system service. Does not seem to be any benefit to caching it.
 - Improve ProjectFile.set_Project on ProjectFile when it's done in its initializer
 - Move ProjectSystemHandler.updatingProjectDataLock might be overzealous? Make it smaller scope, just projectDataMap manipulation?
    - Already implemented previously

TODO:

 - [x] Update benchmark baseline timings

Fixes VSTS #700297 - Gathering typesystem information can be optimized

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectFile.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectFile.cs
@@ -331,7 +331,7 @@ namespace MonoDevelop.Projects
 					var oldLink = link;
 					link = value;
 
-					OnVirtualPathChanged (oldLink, link);
+					VirtualPathChanged?.Invoke (this, new ProjectFileVirtualPathChangedEventArgs (this, oldLink, link));
 					OnChanged ("Link");
 				}
 			}
@@ -479,7 +479,7 @@ namespace MonoDevelop.Projects
 			base.OnProjectSet ();
 			if (Project != null) {
 				base.Include = Include;
-				OnVirtualPathChanged (FilePath.Null, ProjectVirtualPath);
+				VirtualPathChanged?.Invoke (this, new ProjectFileVirtualPathChangedEventArgs (this, FilePath.Null, ProjectVirtualPath));
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


Therzok (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectFile.cs
```diff
@@ -479,7 +479,8 @@ protected override void OnProjectSet ()
 			base.OnProjectSet ();
 			if (Project != null) {
 				base.Include = Include;
-				OnVirtualPathChanged (FilePath.Null, ProjectVirtualPath);
```
Therzok: @mrward would probably be simpler to remove `OnVirtualPathChanged` and just do:

```
VirtualPathChanged?.Invoke (this, new ProjectFileVirtualPathChanged (this, FilePath.Null, ProjectVirtualPath);
```

This won't compute/create anything unless the handler is not null.

mrward: Sure I can change that. Would also need to change the Link property which also uses OnVirtualPathChanged.

mrward: Or are we OK with OnVirtualPathChanged being used elsewhere but not here?

Therzok: I think that method is redundant, it just wraps the event handler and doesn't simplify the code by much:
http://source.monodevelop.com/#MonoDevelop.Core/MonoDevelop.Projects/ProjectFile.cs,ce13c2316f4dd85e

Therzok: There's also OnPathChanged which suffers from the same problem:
http://source.monodevelop.com/#MonoDevelop.Core/MonoDevelop.Projects/ProjectFile.cs,194

Therzok (approved):

Therzok:
cc @slluis

mrward:
Going to need to rebase this again.

Therzok:
We should just merge this for now.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectFile.cs
Therzok: Filed https://github.com/mono/monodevelop/issues/7997

slluis (approved):

mrward:

PR #7904, opened 2019-06-14
Fix tests that rely on object collection

BrzVlad:
If we want to be sure that an object will be collected, all its references must be done within PerformNoPinAction. This also means that the object must be allocated in such a region. Also we should immediately collect after the action, so the object is dead for good, and not bring back additional references to it (for example by calling weakReference.IsAlive which is going to access the Target and might keep it alive on stack, previously producing failures of these tests).

Unify line endings because it screwed up my editor.

Fixes https://github.com/mono/mono/issues/14882


Therzok:
@monojenkins backport master

Therzok:
^ this is to get it building, as we don't support forks

PR #7906, opened 2019-06-14
Fixed Snippets Drag&Drop and added ability for Android Designer to do D&D

DavidKarlas:
Dropping things on `ITextView` is handled by different `IDropHandler`s. `IDropHandler` is chosen based on `[DropFormat ("XXX")]` and priority... So now when ToolboxPad starts dragging session we always insert `MonoDevelopToolBox` so our `MonoDevelopToolBoxDropHandler` is picked to handle dropping. Which then executes `ConsumeItem` for currently selected item on `Document` that is parent of `textView` that was dropped on.
I also added `IDragDataToolboxNode` interface which allows others e.g. Android Designer to define their own `DropFormat` and `byte[]` data when dragging starts that they want to pass to their own `IDropHandler`.

### DIFF
--- a/main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport.Toolbox/IDragDataToolboxNode.cs
+++ b/main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport.Toolbox/IDragDataToolboxNode.cs
@@ -1,0 +1,49 @@
+//
+// IDragDataToolboxNode.cs
+//
+// Author:
+//       David Karlaš <david.karlas@microsoft.com>
+//
+// Copyright (c) 2019 Microsoft Corporation
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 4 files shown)


DavidKarlas:
I can't see way how to use GTK API for handling dropping things on CocoaTextView which is NSView.
Could you explain how this could be done?

main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport.Toolbox/ToolboxDropHandler.cs
```diff
@@ -0,0 +1,86 @@
...
+using MonoDevelop.DesignerSupport;
+using MonoDevelop.DesignerSupport.Toolbox;
+using MonoDevelop.Ide.Gui.Documents;
+
+namespace MonoDevelop.DesignerSupport.Toolbox
+{
+	[Export (typeof (IDropHandlerProvider))]
+	[DropFormat (DropFormat)]
+	[Name (nameof(MonoDevelopToolBoxDropHandlerProvider))]
+	class MonoDevelopToolBoxDropHandlerProvider : IDropHandlerProvider
```
Therzok:
```suggestion
	sealed class MonoDevelopToolBoxDropHandlerProvider : IDropHandlerProvider
```

main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport.Toolbox/ToolboxDropHandler.cs
```diff
@@ -0,0 +1,86 @@
...
+	{
+		public const string DropFormat = "MonoDevelopToolBox";
+		public IDropHandler GetAssociatedDropHandler (ICocoaTextView cocoaTextView)
+		{
+			return new MonoDevelopToolBoxDropHandler (cocoaTextView);
+		}
+	}
+
+	class MonoDevelopToolBoxDropHandler : IDropHandler
```
Therzok:
```suggestion
	sealed class MonoDevelopToolBoxDropHandler : IDropHandler
```

mhutch:
Can you document IDragDataToolboxNode? Do we have a concrete use case for it?

mhutch:
Also, I feel like the editor should be aware of the toolbox, rather than the inverse. And will this compile on !mac?

DavidKarlas:
> Can you document IDragDataToolboxNode? Do we have a concrete use case for it?

Added documentation. Here is concrete use in Android Designer https://github.com/xamarin/designer/commit/a77a6305cb8

> Also, I feel like the editor should be aware of the toolbox, rather than the inverse. And will this compile on !mac?

I moved `ToolboxDropHandler.cs` to TextEditor.Cocoa integration. Sorry for breaking !mac, tnx for noticing that in time. I really wish we had CI for this :(

netonjm:
> I can't see way how to use GTK API for handling dropping things on CocoaTextView which is NSView.
> Could you explain how this could be done?

Ahh, gotcha I see your point, yeah... I remember I had same problem trying to start drag and drop from the native toolbox, I needed to hack the Gtk drag mechanism using same trick like Xwt and worked https://github.com/mono/xwt/blob/master/Xwt.Gtk/Xwt.GtkBackend/WidgetBackend.cs#L1146

But yes, remember it was too hacky, and if we could move to a native way without break GTK it would be great 👍

DavidKarlas:
This is not breaking GTK. Why do you think it breaks it?

abock:

PR #7908, opened 2019-06-14
Map "Show import items" option to equivalent Roslyn feature

sandyarmstrong:
Roslyn implemented this feature as a per-language setting, not a global
intellisense setting like MonoDevelop has. In the future, we should
reorganize this setting to reflect that (and match Windows).

Ensure that the Roslyn feature is disabled in old editor completion. The
items show up correctly in the list, but insertion fails to actually add
the `using` statement. The old editor can continue using its existing
implementation of this feature.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/756548


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices.Options/RoslynPreferences.cs
```diff
@@ -114,6 +115,12 @@ internal PerLanguagePreferences (string language, RoslynPreferences preferences)
 					language + ".SeparateImportDirectiveGroups"
 				);
 
+				ShowItemsFromUnimportedNamespaces = preferences.Wrap<bool?> (
+					new OptionKey (CompletionOptions.ShowItemsFromUnimportedNamespaces, language),
+					IdeApp.Preferences.AddImportedItemsToCompletionList.Value,
```
Therzok: This won't listen to changes from the main property? We should also add a changed handler to set PerLanguage ShowItemsFromUnimportedNamespaces value to the new global AddImportedItemsToCompletionList value.

PR #7915, opened 2019-06-17
[Core] Fix file nesting after installing NuGet package in a F# project

mrward:
Creating a blank F# Xamarin.Forms project and then installing a NuGet
package into the .NET Standard project would result in the file
nesting being broken in the Solution pad. The MainPage.xaml.fs file
would be nested underneath the App.xaml file.

The problem was that on saving the project the MSBuildItem that was
associated with the Update glob was being used as the SourceItem for the
.xaml.fs files. This resulted in the Update glob item that defined
the DependentUpon metadata value of '%(FileName)' being modified on
saving with the evaluated name of the .xaml file. Then when the
project was re-evaluated the file nesting was then broken since the
DependentUpon had App.xaml set for all files.

To fix this the MSBuildItem will look for a SourceItem that has the
same parent as itself. For the .fsproj this means the last MSBuild
item in the project for that file will be used instead of the MSBuild
item from any .targets imported afterwards, such as the
Xamarin.Forms.DefaultItems.targets file.

Fixes VSTS #636641 - Adding a nuget to an F# Xamarin Forms app breaks
the file nesting

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildItem.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildItem.cs
@@ -342,8 +342,10 @@ namespace MonoDevelop.Projects.MSBuild
 			get {
 				if (sourceItem is MSBuildItem)
 					return (MSBuildItem) sourceItem;
-				else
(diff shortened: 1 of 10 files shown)


Therzok (approved):

rodrmoya (approved):

mrward:

PR #7921, opened 2019-06-18
Bump to new MD analyzer nuget

Therzok:
Exception handlers considered empty will trigger a build and edit-time warning.

Fixes VSTS #774649 - SDL - Validate all new code is free of 'swallow everything' exception handlers

### DIFF
--- a/main/src/addins/MacPlatform/MacInterop/Carbon.cs
+++ b/main/src/addins/MacPlatform/MacInterop/Carbon.cs
@@ -221,7 +221,7 @@ namespace MonoDevelop.MacInterop
 			try {
 				if (GetCurrentProcess (out ProcessSerialNumber psn) == 0)
 					CPSSetProcessName (ref psn, name);
-			} catch (TypeLoadException e) {} //EntryPointNotFoundException?
+			} catch (TypeLoadException) {} //EntryPointNotFoundException?
 		}
 
 		struct ProcessSerialNumber {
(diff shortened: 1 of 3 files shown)


main/src/addins/MacPlatform/MacInterop/Carbon.cs
```diff
@@ -215,15 +215,15 @@ internal static string UnConvertCharCode (int i)
...
-		
+
 		public static void SetProcessName (string name)
 		{
 			try {
 				if (GetCurrentProcess (out ProcessSerialNumber psn) == 0)
 					CPSSetProcessName (ref psn, name);
-			} catch {} //EntryPointNotFoundException?
+			} catch (TypeLoadException) {} //EntryPointNotFoundException?
 		}
```
mrward: Previously we were catching all exceptions now we are handling only a TypeLoadException. Is that the only exception that is thrown here? Do we not want to catch all exceptions here? If this fails with another exception do we want it to stop/break the IDE?

Therzok: Well, given the only 2 methods being called here are PInvokes, those are kinda the only exceptions that can be thrown.

Therzok: There are 2 other possible pinvokes exceptions (ExecutionEngine and AccessViolation), but both tear down the runtime.

mrward: Was just wondering 👍

PR #7923, opened 2019-06-18
[IDE] Only set TransientFor when showing the window, not when centering

iainx:
It seems that, on Mojave and for ... unknown reasons, setting TransientFor more than once on a window makes the transient window's subwindows (like popup menus, tooltips) appear behind it if the root window is fullscreen.

Fixes VSTS #911342
Fixes VSTS #735232
Fixes VSTS #893128
Fixes VSTS #888180
Fixes VSTS #890152
Fixes VSTS #850411
Fixes VSTS #918636
Fixes VSTS #848640
Fixes VSTS #856017
Fixes VSTS #891805
Fixes VSTS #893128


iainx:
This isn't the "correct" fix, there seems to be something going on with Gdk. But it's a simple fix that solves a lot of issues, and IMO is still a correct fix to make, as TransientFor shouldn't be set in a method about centring the dialog.

slluis:

slluis:

PR #7925, opened 2019-06-18
Add a new "Open in Terminal" command

jsuarezruiz:
Add a new "Open in Terminal" command (like Reveal in Finder)

Screenshot 2019-06-18 at 10 51 21
Screenshot 2019-06-18 at 10 51 34

Feature VSTS #786721

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectPadContextMenu.addin.xml
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectPadContextMenu.addin.xml
@@ -106,9 +106,6 @@
 			<CommandItem id = "MonoDevelop.Ide.Commands.ProjectCommands.ExportPolicy" />
 			<SeparatorItem id = "PolicyToolsEnd" />
 		</Condition>
-		<Condition id="ItemType" value="IFolderItem">
-			<CommandItem id = "MonoDevelop.Ide.Commands.FileCommands.OpenInTerminal" />
-		</Condition>
 		<Condition id="ItemType" value="Solution|Project|UnknownSolutionItem">
 			<CommandItem id = "MonoDevelop.Ide.Commands.ProjectCommands.EditSolutionItem" />
 		</Condition>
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectPadContextMenu.addin.xml
```diff
@@ -118,6 +118,7 @@
 	</Condition>
 	<Condition id="ItemType" value="IFolderItem|IFileItem">
 		<CommandItem id = "MonoDevelop.Ide.Commands.FileCommands.OpenContainingFolder" />
+		<CommandItem id = "MonoDevelop.Ide.Commands.FileCommands.OpenInTerminal" />
```
slluis: There is already an OpenInTerminal command in the Tools submenu for the folder.

jsuarezruiz: Right. My original idea was to expose in a more visible way the option to open the terminal from the ProjectPad. I also added the option in the document tabs (removed).
After talking with @sevoku, the idea is the following:
- Move the "Open in Terminal" option from the Tools submenu to something more direct.
- Show the option not only with folders, also with files (will use the folder that contains the file).

What do you think?

sevoku: @jsuarezruiz after discussing this with @slluis offline, I agree that we don't really need that for files, this is a folder operation. The main problem is discoverability, I didn't know that we have that at all. However it's due to the context how we use the product, same can be valid for other users with other features, so we need to make a cut at some point. The main problem that I see here is a bit different, for regular folders it's the only option in the tools menu, and submenus don't make sense with just one option, however this is not easy to solve with our command structure and so out of scope here.

jsuarezruiz: Ok. I am going to close the PR.

PR #7950, opened 2019-06-19
Fixes VSTS 896684: Override keyword does not show protocol completions

avodovnik:
Setting IsExclusive to true and ensuring our `ProtocolCompletionMemberProvider` gets called even after `override` keyword is already written. 

This change will ensure our completion data will get surfaced *alongisde* completion suggestions from the other exclusive provider, which is the actual override provider. There was a recent Roslyn change, starting from 16.2.P4 where exclusivity simply means grouping - showing all exclusive ones, not just whoever gets there first. Additionally, we are now ensuring our provider also gets called when there is an insertion/deletion going on (i.e. after the user writes override).

Note: This should only start actually working when we get the Roslyn change that includes the merge from [this PR](https://github.com/dotnet/roslyn/pull/36088), specifically commit: `3a364d2a2600c4a61b2ac7e68c1ac380dc07164a`. I've added the `do-not-merge` label onto it, because it's impossible to properly _test_ before that change. While I've checked, and it _should_ not have any adverse effects, it's probably worth being prudent. 

VSTS: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/896684

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/CompletionProvider/ProtocolMemberCompletionProvider.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/CompletionProvider/ProtocolMemberCompletionProvider.cs
@@ -97,6 +97,22 @@ namespace MonoDevelop.CSharp.Completion.Provider
 					return;
 
 				var startToken = token.GetPreviousTokenIfTouchingWord (position);
+
+				// We're marking the context as exclusive, to resolve the
+				// regression in the new editor that appeared since introducing
+				// Roslyn's provider mechanism. The actual override completion
+				// provider took precedence and therefore this one was ignored.
+				// We submitted a PR for roslyn, that allows multiple providers,
+				// marked as exclusive, to render suggestions.
+				// See: https://github.com/dotnet/roslyn/pull/36088
+				//
+				// Important note, though, we should only set exclusivity when
+				// we are actually competing with an override provider - that
+				// is, if the token before this one is the override keyword. If
+				// it's not, we're not competing, and therefore shouldn't set
+				// this to true, as we'll be the ones stealing the spotlight.
+				context.IsExclusive = startToken.IsKind (SyntaxKind.OverrideKeyword);
+
 				TryDetermineReturnType (startToken, model, cancellationToken, out ITypeSymbol returnType, out SyntaxToken tokenBeforeReturnType);
 				if (returnType == null) {
 					var enclosingType = model.GetEnclosingSymbol (position, cancellationToken) as INamedTypeSymbol;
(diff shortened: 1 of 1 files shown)


main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/CompletionProvider/ProtocolMemberCompletionProvider.cs
```diff
@@ -468,5 +483,23 @@ static bool TryDetermineModifiers (ref SyntaxToken startToken, SourceText text,
 					.WithIsSealed (isSealed);*/
 			return overrideToken.IsKind (SyntaxKind.OverrideKeyword) && IsOnStartLine (text, startLine, overrideToken.Parent.SpanStart);
 		}
+
+		public override bool ShouldTriggerCompletion (SourceText text, int position, CompletionTrigger trigger, Microsoft.CodeAnalysis.Options.OptionSet options)
```
sandyarmstrong: Have you verified that behavior is still correct in old editor with this additional change?

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/CompletionProvider/ProtocolMemberCompletionProvider.cs
```diff
@@ -97,6 +97,21 @@ public override async Task ProvideCompletionsAsync (Microsoft.CodeAnalysis.Compl
 					return;
 
 				var startToken = token.GetPreviousTokenIfTouchingWord (position);
+
+				// We're marking the context as exclusive, to resolve the regression
+				// that appeared since introducing Roslyn's provider mechanism. The
```
sandyarmstrong: Maybe specifically mention here that this is a regression only in the new editor, which interacts with Roslyn completion differently than the old editor.

sandyarmstrong:
Relevant Roslyn update here: https://github.com/mono/monodevelop/pull/8019

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/CompletionProvider/ProtocolMemberCompletionProvider.cs
```diff
@@ -97,6 +97,21 @@ public override async Task ProvideCompletionsAsync (Microsoft.CodeAnalysis.Compl
...
+				// actual override completion provider took precdence and therefore
```
KirillOsenkov: `precdence` -> `precedence`

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/CompletionProvider/ProtocolMemberCompletionProvider.cs
```diff
@@ -468,5 +483,23 @@ static bool TryDetermineModifiers (ref SyntaxToken startToken, SourceText text,
 					.WithIsSealed (isSealed);*/
 			return overrideToken.IsKind (SyntaxKind.OverrideKeyword) && IsOnStartLine (text, startLine, overrideToken.Parent.SpanStart);
 		}
+
+		public override bool ShouldTriggerCompletion (SourceText text, int position, CompletionTrigger trigger, Microsoft.CodeAnalysis.Options.OptionSet options)
+		{
+			// This method is overriden because Roslyn's Completion Service calls us, in cases
```
KirillOsenkov: `overridden`

sandyarmstrong:
Works as expected with Roslyn bump in https://github.com/mono/monodevelop/pull/8019.

main/src/addins/CSharpBinding/MonoDevelop.CSharp.Completion/CompletionProvider/ProtocolMemberCompletionProvider.cs
sandyarmstrong: I have tested in the old editor and we seem good here.

sandyarmstrong:
Fixed all comments, rebased, removed `do-not-merge` label. This is good to go now!

abock:

PR #7954, opened 2019-06-19
Fixes exception trying to call terminal.app from Catalina

netonjm:
This problem arises when trying to Debug / Run a normal console application.

The exception:
image

This was caused by the fact that the Terminal.app has changed in Catalina and it's now in:
/System/Applications/Utilities/Terminal.app

With the fix:

lola

Fixes VSTS #918848 - [Catalina] Cannot debug a normal console application

### DIFF
--- a/main/src/addins/MacPlatform/MacExternalConsoleProcess.cs
+++ b/main/src/addins/MacPlatform/MacExternalConsoleProcess.cs
@@ -80,12 +80,18 @@ using exec -a to name the process has problems because the terminal can get into
 where running an explicit exec causes it to quit after the exec runs, so we can't use the 
 bash pause on exit trick
 */ 
-
 		//use full path to fix command failure when users have another app called "Terminal"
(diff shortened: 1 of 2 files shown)


Therzok (approved):

iainx:
While the fix looks good, would be better to have an array of possible paths for the terminal, the old path and the new path, and the first time the path is requested, check if the path is valid, and if we can't find the terminal in either place, pop up a dialog for the user "Unable to find Terminal.app". It might future proof us if the Terminal.app moves again, and it could be the start to allow users to set a  custom terminal

iainx (approved):

sevoku (approved):

sevoku (approved):
Agree with Iain, but we actually need a much more generic solution with customizable terminal support (incl. 3rd party). We have planned that already for future work.

sevoku:

PR #7965, opened 2019-06-20
Add some null checks that would prevent showing an info bar if an item array wasn't passed.

Therzok:


Therzok:
cc @drasticactions

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices/MonoDevelopInfoBarService.cs
```diff
@@ -72,11 +73,9 @@ void ShowInfoBar (bool activeView, string message, params InfoBarUI [] items)
...
 				}
 			}, _listener.BeginAsyncOperation (nameof (ShowInfoBar)));
-		}
 
-		static InfoBarItem [] ToUIItems (InfoBarUI [] items)
-		{
-			return items.Select (x => new InfoBarItem (x.Title, ToUIKind (x.Kind), x.Action, x.CloseAfterAction)).ToArray ();
+			static InfoBarItem [] ToUIItems (InfoBarUI [] items)
+				=> items?.Select (x => new InfoBarItem (x.Title, ToUIKind (x.Kind), x.Action, x.CloseAfterAction)).ToArray ();
 		}
```
mrward: I thought this would throw a null reference exception if items was null but it doesn't seem. What is this magic? :)

mrward: I guess the `items?` is bypassing the entire right hand side of that statement if items is null. Now it makes sense to me.

Therzok: Yes. It's a short circuit statement:
https://gist.github.com/Therzok/0828ff8367f6ce55dda60f8f54a83111

See compiler generated code

Therzok:

sevoku:

PR #7969, opened 2019-06-20
Fixes VSTS Bug 901126: Updating from git w/ remote & local changes

mkrueger:
stashes local changes but never pops the stash

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/901126

CommonPreMergeRebase changes the gitupdateoptions without that change
the local changes are not unstashed in the postmergerebase operation.


sevoku:

PR #7984, opened 2019-06-21
[CSharpBinding] Fix path bar sometimes not showing owner projects

mrward:
Opening a C# file from a shared project would sometimes not show
the owner projects in the path bar. Also affects multi-target
projects. The WorkspaceRegistration.WorkspaceChanged event would
sometimes not be generated on opening the text editor so the
owner projects were not added. To prevent the code is run on
creating the CSharpPathedDocumentExtension so that the owner
projects have been added in case the WorkspaceChanged event does
not fire.

Fixes VSTS #934792 - Editor path bar sometimes does not list
project information


Therzok (approved):

DavidKarlas (approved):

mrward:

PR #7992, opened 2019-06-24
Make DesignerSupport work better with the new document model

garuma:
This PR improves toolbox and property pad support for designers that host a source editor (or other IDE-provider controllers).

The main idea is to ensure that the top-level controller in such scenario is always queried first when anything related to those two components is needed (for instance to determine what's the right implementation of `IToolboxConsumer` or `ICustomPropertyPadProvider` to use).

It does this by:
1. Ensuring `DocumentViewContainer.GetActiveControllerHierarchy` returns controllers top-down so that  a`GetContent` query (for the active view) let the parent controller appropriately control delegation to its sub-controllers
2. Adding support for recognizing `DocumentView` when walking the command chain to determine what property pad implementation to show and make use of the `GetContent` machinery instead of relying purely on interface implementation.

@slluis while I was peering in the code I noticed that `DocumentViewContainer.GetAllControllers` calls `base.GetActiveControllerHierarchy` which seems strange. Normal or real issue?

Implements https://devdiv.visualstudio.com/DevDiv/_workitems/edit/891571


main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport/PropertyPadVisitor.cs
```diff
@@ -93,18 +94,33 @@ public bool Visit (object ob)
...
-			else if (ob is IPropertyPadProvider) {
+			if (ob is DocumentView dv) {
+				foreach (var dc in dv.GetTopLevelDocumentView ().GetActiveControllerHierarchy ()) {
+					var ppProvider = dc.GetContent<IPropertyPadProvider> ();
+					if (ppProvider != null) {
+						DesignerSupport.Service.SetPadContent (ppProvider, activeWidget);
+						found = true;
+						return true;
+					}
+					var cppProvider = dc.GetContent<ICustomPropertyPadProvider> ();
```
decriptor: C++ provider? ;)

Therzok: customPropertyPadProvider :p but yeah.

garuma:
@slluis ping

slluis:
I'm concerned by the change of GetActiveControllerHierarchy(), since it completely changes the semantics of that methods and other methods that depend on it. I need to review the code and see what can be the consequences.

garuma:
As far as I could see the only place where it would change things is when `DocumentViewContainer` is used and that's only us + regex + GTK designer. All other cases have only one controller and thus that method only ever return one of them so no ordering issue.

garuma:
@Therzok if you can provide guidance as to why the leak tests are failing, that giant retain graph is fairly unparsable to me

Therzok:
There are cases which we don't weed out. There's a task holding everything alive on the stack. Rerunning it should fix it.

PR #7993, opened 2019-06-24
Fixes VSTS Bug 935392: System.NullReferenceException exception in

mkrueger:
MonoDevelop.SourceEditor.SourceEditorView.MonoDevelop.Ide.Editor.ITextEditorImpl.GrabFocus()

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/935392

Without a repro it's hard to tell where that happens. I added dispose
checks & a null check on different levels of the call. It's caused by
the document level - GrabFocus beeing called on a disposed editor.


sevoku:

PR #7995, opened 2019-06-24
Implemented collection pooling

mkrueger:
We allocate small collections in some hot spots.

Using object pooling would prevent allocating/deallocating a lot of collections on the heap and lower the GC work.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core/ObjectPoolUtil.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core/ObjectPoolUtil.cs
@@ -1,0 +1,99 @@
+//
+// CollectionCache.cs
+//
+// Author:
+//       Mike Krüger <mikkrg@microsoft.com>
+//
+// Copyright (c) 2019 Microsoft Corporation. All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.ObjectPool;
+
+namespace MonoDevelop.Core
+{
+	public static class ObjectPoolUtil
+	{
+		const int maximumCollectionSize = 1024;
+
+		public static ObjectPool<List<T>> CreateListPool<T> () => ObjectPool.Create (new PooledListPolicy<T> ());
+
+		public static ObjectPool<HashSet<T>> CreateHashSetPool<T> () => ObjectPool.Create (new PooledHashSetolicy<T> ());
+
+		sealed class PooledHashSetolicy<T> : PooledObjectPolicy<HashSet<T>>
+		{
+			public override HashSet<T> Create () => new HashSet<T> ();
+
+			public override bool Return (HashSet<T> obj)
+			{
+				if (obj.Count > maximumCollectionSize)
+					return false;
+				obj.Clear ();
+				return true;
+			}
+		}
+
+		public static ObjectPool<Dictionary<T, U>> CreateDictionaryPool<T, U> () => ObjectPool.Create (new PooledDictionaryPolicy<T, U> ());
+
+		sealed class PooledDictionaryPolicy<T, U> : PooledObjectPolicy<Dictionary<T, U>>
+		{
+			public override Dictionary<T, U> Create () => new Dictionary<T, U> ();
+
+			public override bool Return (Dictionary<T, U> obj)
+			{
+				if (obj.Count > maximumCollectionSize)
+					return false;
+				obj.Clear ();
+				return true;
+			}
+		}
+
+		public static ObjectPool<Stack<T>> CreateStackPool<T> () => ObjectPool.Create (new PooledStackPolicy<T> ());
+
+		sealed class PooledStackPolicy<T> : PooledObjectPolicy<Stack<T>>
+		{
+			public override Stack<T> Create () => new Stack<T> ();
+
+			public override bool Return (Stack<T> obj)
+			{
+				if (obj.Count > maximumCollectionSize)
+					return false;
+				obj.Clear ();
+				return true;
+			}
+		}
+
+		public static ObjectPool<Queue<T>> CreateQueuePool<T> () => ObjectPool.Create (new PooledQueuePolicy<T> ());
+
+		sealed class PooledQueuePolicy<T> : PooledObjectPolicy<Queue<T>>
+		{
+			public override Queue<T> Create () => new Queue<T> ();
+
+			public override bool Return (Queue<T> obj)
+			{
+				if (obj.Count > maximumCollectionSize)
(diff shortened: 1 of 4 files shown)


Therzok:
We have object pooling via Microsoft.Extensions.ObjectPool

main/src/core/MonoDevelop.Core/MonoDevelop.Core/CollectionCache.cs
```diff
@@ -0,0 +1,152 @@
...
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.ObjectPool;
+
+namespace MonoDevelop.Core
+{
+	public static class CollectionCache
+	{
+		public static List<T> AllocateList<T> () => Pool<T, List<T>>.Allocate ();
```
Therzok: I'm not sure we want a global collection cache. This would keep some objects in memory regardless of the component being used or not.

We could provide a global collection policy that would clear and check for capacity. But I wouldn't make static pools for a generic `List<T>`.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/CollectionCache.cs
```diff
@@ -0,0 +1,152 @@
...
+			internal static U Allocate () => pool.Get ();
+
+			internal static void Deallocate (U collection)
+			{
+				if (collection == null)
+					throw new ArgumentNullException (nameof (collection));
+				pool.Return (collection);
+			}
+
+			class PooledObjectPolicy : PooledObjectPolicy<U>
```
Therzok: In example, I'd make just this public API and that's it. Every code path should have a local pool they're responsible for, either per service or per instance. A global pool for a type used only on solution load would just 'leak' objects for a long time.

Therzok (changes requested):
We need to have a better placement for collection pooling, usually scoped to the pool user, not as a service.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/CollectionCache.cs
mkrueger: Wouldn't leak - it would be constant which is a big difference. 

Other than that - yes it's not useful for every case but there are often useful.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/CollectionCache.cs
mkrueger: Hard to tell that's why I want to discuss it. I'm sure you see in the profiler as well that we've high memory traffic of arrays beeing created. 

Re-using collections is something we can try to lower the memory traffic. I agree that it's not useful for every case but we need to find the hot spots for memory traffic and reduce them.

ATM I don't have a profiler for that :/. In the old editor there would've been several places where collection pooling would be useful.

Therzok: Yes, but what I mean is make the services in the assemblies that need to hold it alive, not in Core. Having too many users of a pool means that we can potentially have more than 1000 pooled objects kept alive simply because too many requested them, and returning them will add them onto the pool.

In example, say a pool has a default of 16 members, if we end up having too many concurrent users, we can possibly request 1000. That means 1000 allocations that will remain in memory _forever_. It's not a constant.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/CollectionCache.cs
Therzok: Using ObjectPool in the text editor would be better than having this in Core.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/CollectionCache.cs
mkrueger: It's a constant it'll always have an upper limit.

Therzok: What I meant is we can possibly have 16 collections with 1000 members each. So that's 16000 objects that will indefinitely remain in memory. The internal arrays for List<T> count. I'm still not sold on having APIs for open-ended collections `List<T>`. Local pools for _specific types_ is what we need.

Since we have addin unloading, for example, if an addin for a given T is unloaded, this will crash.

rodrmoya: Maybe retained collections could have an expiration time? Not sure though if that would make the code better or worse, as, I guess, this means storing creation date per collection, iterate over all collections regularly, etc

Therzok: Expiration time does not solve the crash I mentioend above.****

Therzok: Again, how much more code is just doing:
```
new ObjectPool<List<T>> (MonoDevelop.Core.CollectionPoolPolicy ();
```

Therzok:
I'm okay with the objectpool usage, just not exposing this kind of API.

Therzok:
@mkrueger how would having something like, sound to you? That way we keep the pool _outside_ MD.Core, and we just provide utility helpers.

```
static class CollectionPoolHelpers
{
    public ObjectPool<List<T>> CreateListPool<T> ();
    public ObjectPool<HashSet<T>> CreateHashSetPool<T> ();
}
```

mkrueger:
CollectionPoolHelpers sounds ok that would work. I'll implement that - but using more pooling is something we should do at many parts.

Therzok (approved):
Much better!

sevoku (approved):
LGTM! Can we fix the build and merge this?

main/src/core/MonoDevelop.Core/MonoDevelop.Core/ObjectPoolUtil.cs
```diff
@@ -0,0 +1,86 @@
...
+				if (obj.Count > maximumCollectionSize)
+					return false;
+				obj.Clear ();
+				return true;
+			}
+		}
+
+		public static ObjectPool<Stack<T>> CreateStackPool<T> () => ObjectPool.Create (new PooledStackPolicy<T> ());
+
+		class PooledStackPolicy<T> : PooledObjectPolicy<Stack<T>>
```
Therzok:
```suggestion
		sealed class PooledStackPolicy<T> : PooledObjectPolicy<Stack<T>>
```

main/src/core/MonoDevelop.Core/MonoDevelop.Core/ObjectPoolUtil.cs
```diff
@@ -0,0 +1,86 @@
...
+				if (obj.Count > maximumCollectionSize)
+					return false;
+				obj.Clear ();
+				return true;
+			}
+		}
+
+		public static ObjectPool<Queue<T>> CreateQueuePool<T> () => ObjectPool.Create (new PooledQueuePolicy<T> ());
+
+		class PooledQueuePolicy<T> : PooledObjectPolicy<Queue<T>>
```
Therzok:
```suggestion
		sealed class PooledQueuePolicy<T> : PooledObjectPolicy<Queue<T>>
```

main/src/core/MonoDevelop.Core/MonoDevelop.Core/ObjectPoolUtil.cs
```diff
@@ -0,0 +1,86 @@
...
+
+		public static ObjectPool<List<T>> CreateListPool<T> () => ObjectPool.Create (new PooledListPolicy<T> ());
+
+		public static ObjectPool<HashSet<T>> CreateHashSetPool<T> () => ObjectPool.Create (new PooledCollectionPolicy<T, HashSet<T>> ());
+
+		public static ObjectPool<Dictionary<T, U>> CreateDictionaryPool<T, U> () => ObjectPool.Create (new PooledCollectionPolicy<KeyValuePair<T, U>, Dictionary<T, U>> ());
+
+		class PooledCollectionPolicy<T, U> : PooledObjectPolicy<U> where U : class, ICollection<T>, new()
+		{
+			public override U Create () => new U ();
```
Therzok: This might bite us. It uses reflection (Activator.CreateInstance). We can probably have a pooled HashSet policy.

mkrueger: wtf - .NET what have you done ?!?!?!

The question is: Can the JIT optimize that ?

Therzok (approved):

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor/SegmentTree.cs
```diff
@@ -30,6 +30,7 @@
 using System.Text;
 using System.Web.UI.WebControls;
 using System.Diagnostics;
+using MonoDevelop.Core;
```
Therzok:
```suggestion
using MonoDevelop.Core;
using Microsoft.Extensions.ObjectPool;
```

Therzok:
Added a commit to fix the build

Therzok:
@sevoku can you take alook at merge this?

Therzok (approved):

Therzok:
@monojenkins rebase

PR #7996, opened 2019-06-24
[Ide] Fix missing xaml file breaking intellisense

mrward:
If a project referenced a .xaml file that did not exist then the
type system would throw a file not exception and not load any
type information for the solution. This then broke code completion,
go to declaration, etc. To avoid this error a check is made to
ensure the file exists before trying to generate projections for the
.xaml file.

Fixes VSTS #918867 - Missing xaml file breaks intellisense


slluis:

slluis:

PR #8001, opened 2019-06-24
[Ide] Fix state corruption with new-document-model and old editor

garuma:
The controller for the old editor didn't properly synchronize the value of HasUnsavedChanges with the internal textEditor dirty flag.

This would cause normal saves of the model to perpetuate a state where the editor controller's HasUnsavedChanges flag would never get reset to true because the internal dirty state would have been kept back.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/936466

> Fixes VSTS #936466


slluis:
@mkrueger can you review?

mkrueger:
This only works before load -  right ? It's not really clear to me what's fixed here. But if it works in your use-case it's ok.

mkrueger:
I would propose https://github.com/mono/monodevelop/pull/8027

sevoku:

PR #8003, opened 2019-06-25
Fixes VSTS Bug 936228: Exact match and regex match mimetypes no longer

mkrueger:
match

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/936228

The file name evaluator didn't take the full path - it only worked
with file name. Which is an unusual API. I changed it that it works
with full path as well so a future mis-use of the API is prevented.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Extensions/FileNameEvalutor.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Extensions/FileNameEvalutor.cs
@@ -54,6 +54,16 @@ namespace MonoDevelop.Ide.Extensions
 
 		public abstract bool SupportsFile (string fileName);
 
+		static string SafeGetFileName (string fileName)
+		{
+			try {
+				return System.IO.Path.GetFileName (fileName);
+			} catch (Exception e) {
+				LoggingService.LogInternalError (e);
+				return fileName;
+			}
+		}
+
 		class RegexFileNameEvaluator : FileNameEvalutor
 		{
 			Regex regex;
@@ -81,7 +91,7 @@ namespace MonoDevelop.Ide.Extensions
 
 			public override bool SupportsFile (string fileName)
 			{
-				return regex.IsMatch (fileName);
+				return regex.IsMatch (SafeGetFileName (fileName));
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Extensions/FileNameEvalutor.cs
```diff
@@ -54,6 +55,15 @@ static IEnumerable<string> SplitPatterns (IEnumerable<string> patterns, char sep
 
 		public abstract bool SupportsFile (string fileName);
 
+		static string SafeGetFileName (string fileName)
+		{
+			try {
+				return System.IO.Path.GetFileName (fileName);
+			} catch {
```
Therzok: The IDE should trigger a warning here. We need to catch a more specific exception or at least log it.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Extensions/FileNameEvalutor.cs
```diff
@@ -28,6 +28,7 @@
 using System.Linq;
 using System.Text.RegularExpressions;
 using MonoDevelop.Core;
+using Cairo;
```
Therzok: where is cairo used?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Extensions/FileNameEvalutor.cs
mkrueger: Added

mkrueger:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Extensions/FileNameEvalutor.cs
Therzok: This was marked as resolved, but still don't see Cairo being used.

slluis:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Extensions/FileNameEvalutor.cs
mkrueger: push failed for some reason.

PR #8025, opened 2019-06-26
[Roslyn] Pass app.config file for projects which specify it

Therzok:
Fixes VSTS #802643 - CSharpCompilationOptions is missing binding redirect information

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CSharpCompilerParameters.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CSharpCompilerParameters.cs
@@ -29,6 +29,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Globalization;
+using System.IO;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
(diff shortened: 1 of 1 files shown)


main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CSharpCompilerParameters.cs
```diff
@@ -159,11 +160,55 @@ public override CompilationOptions CreateCompilationOptions ()
...
+
+				if (items != null) {
+					// Probe for item in None and Content
+					var appConfig = "app.config".AsSpan ();
+					foreach (var item in items) {
+						if (item.Name != "Content" && item.Name != "None")
+							continue;
+
+						var fileNameStart = item.Include.LastIndexOf (Path.DirectorySeparatorChar) + 1;
+						var fileName = item.Include.AsSpan (fileNameStart);
```
Therzok: Should return full path.

Therzok:
Needs tests

Therzok:
I couldn't come up with a test case that would yield different results.

Therzok:
After further analysis, closing the bug and the PR. Seems like it only uses the app config to support some silverlight scenario with system libraries. Not worth the extra work done to find the app config file on load just for a scenario that mono and .netcore don't support.

The item probed for is `supportPortability`

PR #8026, opened 2019-06-26
[IDE] Listen for NSNotification to know if the app was started from F…

iainx:
…inder

On Mac  when the app is started from Finder, the requested files are passed to the app asynchronously, so we need to listen for the event.

Fixes VSTS #737145

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageService.cs
@@ -81,7 +81,14 @@ namespace MonoDevelop.Ide.WelcomePage
 			if (!hideWelcomePage && HasWindowImplementation) {
 				await Runtime.GetService<DesktopService> ();
 				var commandManager = await Runtime.GetService<CommandManager> ();
-				await ShowWelcomeWindow (new WelcomeWindowShowOptions (false));
+
+				var reason = await IdeApp.LaunchCompletionSource.Task;
+
+				if (IdeApp.LaunchReason == IdeApp.LaunchType.Normal) {
+					await ShowWelcomeWindow (new WelcomeWindowShowOptions (false));
+				} else if (IdeApp.LaunchReason == IdeApp.LaunchType.Unknown) {
+					LoggingService.LogInternalError ("LaunchCompletion is still Unknown", new Exception ());
+				}
 			}
 		}
 
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
@@ -192,9 +192,8 @@ namespace MonoDevelop.Ide
 			get => launchType;
 			internal set {
 				launchType = value;
-				if (!LaunchCompletionSource.Task.IsCompleted) {
-					LaunchCompletionSource.SetResult (value);
-				} else {
+
+				if (!LaunchCompletionSource.TrySetResult (value)) {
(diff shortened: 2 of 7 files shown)


iainx:
Some notes:

* We can't put this logic inside MacPlatform, as Cocoa is initialised at the same time as Gtk, so by the time MacPlatform is initialised, we've already missed the NSNotification
* This delays the GTC window until the NSNotification has been fired, to work around the race condition between which happens first. The delay adds milliseconds to the GTC window appearing, so isn't a big issue.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
```diff
@@ -180,6 +180,24 @@ internal static void OnStartupCompleted ()
...
+			Normal,
+			LaunchedFromFileManager
+		}
+
+		static LaunchType launchType = LaunchType.Unknown;
+		public static LaunchType LaunchReason {
+			get => launchType;
+			set {
+				launchType = value;
+				LaunchCompletionSource.SetResult (value);
```
sevoku: This will throw an InvalidOperationException if called more than once. Make the setter internal and add a check for the completion?

Therzok: There's TrySetResult.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageService.cs
```diff
@@ -81,7 +81,14 @@ internal static async Task Initialize (bool hideWelcomePage)
 			if (!hideWelcomePage && HasWindowImplementation) {
 				await Runtime.GetService<DesktopService> ();
 				var commandManager = await Runtime.GetService<CommandManager> ();
-				await ShowWelcomeWindow (new WelcomeWindowShowOptions (false));
+
+				var reason = await IdeApp.LaunchCompletionSource.Task;
```
sevoku: If we don't get the `DidFinishLaunchingNotification` notification (either when running on other platforms or for some weird reason on Mac) and `startupInfo.HasFiles == false`, we won't show the WelcomeWindow at all (since it waits for the LaunchReason to be set). I think we should have some fall-back scenario, where we know that the notification must have been there and set `LaunchReason = LaunchType.Normal`.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
```diff
@@ -117,6 +117,11 @@ int Run (MonoDevelopOptions options)
 			IdeTheme.InitializeGtk (BrandingService.ApplicationName, ref args);
 
 			startupInfo = new StartupInfo (options, args);
+			if (startupInfo.HasFiles) {
+				// If files are present, consider started from the commandline as being the same as file manager.
+				// On macOS, we need to wait until the DidFinishLaunching notification to find out we were launched from the Finder
+				IdeApp.LaunchReason = IdeApp.LaunchType.LaunchedFromFileManager;
+			}
```
sevoku:
```suggestion
			} else if (!Platform.IsMac) {
				IdeApp.LaunchReason = IdeApp.LaunchType.Normal;
			}
```
To be sure we set it on other platforms

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
sevoku: > There's TrySetResult.
right! but the setter should be still internal

sevoku:

slluis:
There is a small delay after the icon stops bouncing and before the shell shows up, but it works fine. We can try to fix the delay in another PR.

PR #8033, opened 2019-06-26
[C#] Fix unhandled exception that is thrown when completion is canceled

Therzok:
Add a OnlyOnRanRanToCompletion option and execute synchronously.

Fixes VSTS #798074 - Unobserved exception in CSharpCompletionTextEditorExtension


Therzok:

PR #8034, opened 2019-06-27
[Debugger] Fix regression introduced when optimizing handlers

Therzok:
https://github.com/mono/monodevelop/commit/ef5ea322347547cb6d971b6097b41965312178e6 introduced a regression where we would load the children on a background thread.

That seems to race with debugger suspending/resuming. Go back to running on the UI thread. Also, fix up usage of the token source.

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValueTreeView.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValueTreeView.cs
@@ -947,15 +947,20 @@ namespace MonoDevelop.Debugger
 			}
 
 			int numberOfChildren = store.IterNChildren (iter);
-			Task.Run (() => {
+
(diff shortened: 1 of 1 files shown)


Therzok:
@mrward can you take a look and approve? We can get a VM not suspended dialog when expanding items in the treeview. Didn't see it during the first testing pass of the Task cleanup PR, not sure if we can unit test it.

Therzok:
@mrward Task.Factory.StartNew starts a task on the current synchronization context. If called from the UI thread, it'll queue an UI thread callback, if called from a background thread, it'll queue on the threadpool.

`LoadIEnumerableChildren` is definitely called on the UI thread. See [this](https://stackoverflow.com/a/55949173/2502519). TaskScheduler.Current is implied for most APIs except Task.Run, which uses TaskScheduler.Default (threadpool).

Therzok:
@slluis this got lost, we might want it for 8.2.

Therzok:

PR #8039, opened 2019-06-27
[Core] Fix null ref if RunTarget called after Project is disposed

mrward:
Project.RunTarget now uses BindTask so the target will complete
before the Project is disposed. This will also prevent a null
reference exception that could occur when the MSBuildProject was
accessed in RunMSBuildTarget when RunTarget is called after the
Project is disposed. A TaskCancelationException is thrown if
RunTarget is called after the Project has been disposed.

Fixes VSTS #935875 - High exception count in MSBuild as run by the
Android Designer

Could not reproduce the above bug testing the designer. Assumption
based on the exception callstack is that RunTarget is being called
after the Project is disposed.


Therzok (approved):

slluis (approved):

mrward:

mrward:

PR #8040, opened 2019-06-27
Fix 937834: In the new editor, the Chinese input repeats.

DavidKarlas:
Removed `<Map` for Backspace command/key. So it’s not handled by MonoDevelop, hence its passed to native editor so `KeyDown` method is called which calls `InterpretKeyEvents` so native IME processing logic updates its internal state so IME input after typing more characters after Backspace is correct, otherwise IME reinserts deleted chars


abock:

abock:

slluis:
@monojenkins backport master

PR #8048, opened 2019-06-27
[A11y] Only show the VoiceOver notification once.

iainx:
Fixes VSTS #729402

### DIFF
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -378,17 +378,15 @@ namespace MonoDevelop.MacIntegration
 		// The Enabled key needs to be controlled through NSUserDefaults so that it can be
 		// set from the command line. This lets users who require accessibility features to
 		// enabled to control it through the commandline, if it has been disabled from inside the software
-		static bool voiceOverNoticeShown;
 		const string EnabledKey = "com.monodevelop.AccessibilityEnabled";
-		const string VoiceOverNoticeShownKey = "com.monodevelop.VoiceOverShown";
+		const string VoiceOverNoticeShownKey = "MonoDevelop.VoiceOver.Show";
 		static void ShowVoiceOverNotice ()
 		{
-			if (voiceOverNoticeShown || PropertyService.Get<bool> (VoiceOverNoticeShownKey, false)) {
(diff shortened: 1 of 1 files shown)


iainx:

iainx:

iainx:
Going to move the property to the Propertyservice instead of using NSUserDefaults

mrward:
I was assuming NSUserDefaults would handle multiple instances running. The property service does not reload the settings after the IDE is started.

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -370,10 +370,22 @@ static void NewInstanceMenuItem_Activated (object sender, EventArgs e)
...
 		const string EnabledKey = "com.monodevelop.AccessibilityEnabled";
+		const string VoiceOverNoticeShownKey = "com.monodevelop.VoiceOverShown";
 		static void ShowVoiceOverNotice ()
 		{
+			if (voiceOverNoticeShown || PropertyService.Get<bool> (VoiceOverNoticeShownKey, false)) {
+				return;
+			}
+
+			// Show the VoiceOver notice once
+			voiceOverNoticeShown = true;
```
mrward: Do we need to use both the PropertyService and the voiceOverNoticeShown? Seems like we could just use the PropertyService now.

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -370,10 +370,22 @@ static void NewInstanceMenuItem_Activated (object sender, EventArgs e)
 				LoggingService.LogError ($"Failed to start new instance: {error.LocalizedDescription}");
 		}
 
-
+		// The Enabled key needs to be controlled through NSUserDefaults so that it can be
+		// set from the command line. This lets users who require accessibility features to
+		// enabled to control it through the commandline, if it has been disabled from inside the software
+		static bool voiceOverNoticeShown;
 		const string EnabledKey = "com.monodevelop.AccessibilityEnabled";
+		const string VoiceOverNoticeShownKey = "com.monodevelop.VoiceOverShown";
```
Therzok: Nit: Key naming style in propertyservice is like:
```suggestion
const string VoiceOverNoticeShownKey = "MonoDevelop.VoiceOver.Shown";
```

sevoku:

PR #8059, opened 2019-06-28
Fix 935181: TryGetBreakpointSpan called with invalid position

DavidKarlas:
I also added logic to fallback to whole line breakpoint in case Roslyn fails to find matching statement to place breakpoint on.


Therzok (approved):

jstedfast (approved):

slluis:

PR #8063, opened 2019-06-28
Avoid doing expensive work to trigger OnPathChanged when it might not be subscribed to

Therzok:
Fixes VSTS #936533 - ProjectFile.OnPathChanged computes values that are not used


mrward (approved):

Therzok:
There's still a problem with this. ProjectFileCollection subscribes the events before the Project sets itself as the owner, so here's a small penalty due to that, but not sure if it's easily fixable. This will do for now.

Therzok:
@slluis this is green, can we merge?

slluis (approved):

PR #8068, opened 2019-06-28
Fixes VSTS Bug 937031: [FATAL] System.NullReferenceException exception

mkrueger:
in Mono.TextEditor.TextArea.FocusOut()

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/937031

FocusOut should never produce a fatal crash.


sevoku:

PR #8073, opened 2019-06-28
Fix 935197: Exception in TextFileNavigationPoint.OnDocumentClosing

DavidKarlas:
I'm not exactly sure how that exception happened(maybe TextBuffer of TextView changed). But with this change problem of snapshots coming from different textbuffers can't happen again.


abock:

PR #8081, opened 2019-06-28
Fix the workspace logic to apply new text to a buffer.

KirillOsenkov:
The previous logic relied on the new Roslyn text deriving from the old Roslyn text via a chain of changes. We have seen a case in XAML where an extra change came from somewhere, deleting a random character. The change couldn't be explained by diffing between the old text and the new text.

The right thing to do here (and that's what VS Windows does) is to just give the desired new text to the Editor and let the Editor diff it. We pass the ComputeMinimalChange option so the editor will arrive at a minimal set of correct edits without Roslyn help.

We shouldn't worry about performance here because it's an infrequent scenario and the editor is really good at diffing efficiently, and the files of average size aren't a concern in terms of allocating the string.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
@@ -690,18 +690,7 @@ namespace MonoDevelop.Ide.TypeSystem
 		{
 			using (var edit = buffer.CreateEdit (options, reiteratedVersionNumber: null, editTag: null)) {
 				var oldSnapshot = buffer.CurrentSnapshot;
-				var oldText = oldSnapshot.AsText ();
-				var changes = newText.GetTextChanges (oldText);
-				//if (Microsoft.CodeAnalysis.Workspace.TryGetWorkspace(oldText.Container, out var workspace))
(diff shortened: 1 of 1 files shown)


KirillOsenkov:
@mgoertz-msft

mgoertz-msft:
Now it just needs to go into 8.2  ;-)

KirillOsenkov:
Edit: I was wrong, looks like the VS implementation did exactly what our code used to do:
http://source.roslyn.io/#Microsoft.VisualStudio.LanguageServices/Implementation/ProjectSystem/TextEditApplication.cs,bc561af5da7facbf

Still, that only works for the assumption that the new text was computed manually, and not via an opaque edit that replaces everything:
http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces/Workspace/Workspace.cs,1443
(in our scenario both the old text and the new text existed, so that last branch was taken).

KirillOsenkov:

KirillOsenkov:
@slluis I've tested the .dmg and the lightbulb works OK. How do we get this merged?

slluis:
@KirillOsenkov somebody needs to review, then I can merge in 8.2

KirillOsenkov:
@slluis we've decided to not take it for 8.2, so master only please, so we can dogfood it properly

slluis:
Sounds good.

PR #8084, opened 2019-06-30
[Ide] Fix crash on shutdown when just opening GTC window and then qui…

Therzok:
…tting

Fixes VSTS #940663 - Opening IDE then closing it exits with a crash

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -563,7 +563,10 @@ namespace MonoDevelop.Components.Commands
 				if (cinfo.Enabled && cinfo.Visible) {
 					if (!dispatched)
 						dispatched = DispatchCommand (commands [i].Id, null, null, CommandSource.Keybinding, ev.Time, cinfo);
-					if (!bindings.BindingExists (binding))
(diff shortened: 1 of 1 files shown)


Therzok:

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -563,7 +563,7 @@ bool ProcessKeyEventCore (Gdk.EventKey ev)
 				if (cinfo.Enabled && cinfo.Visible) {
 					if (!dispatched)
 						dispatched = DispatchCommand (commands [i].Id, null, null, CommandSource.Keybinding, ev.Time, cinfo);
-					if (!bindings.BindingExists (binding))
+					if (disposed || !bindings.BindingExists (binding))
```
Therzok: Due to the way async events work and continuations are queued, this is the simplest fix. If we do a pumping wait in any of the update handlers, we end up running other callbacks, like shutdown's CommandManager.Dispose being called while we're waiting.

```suggestion
					// A pumping wait could possibly let CommandManager.Dispose be called before continuing here, check for disposed checks.
					if (disposed || !bindings.BindingExists (binding))
```

The proper fix would be to have a CancellationTokenSource that's cancelled on dispose that is used when waiting on the update task, and using JTF.RunAsync on the wait above.

Therzok:

PR #8087, opened 2019-07-01
Fixes VSTS Bug 776934: TreeBuilderContext.GetTreeBuilder() null

mkrueger:
reference exception

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/776934

Setting the store to null after the node builders are destoyed should
fix the issue theoretically. I couldn't reproduce it on my machine so
I added some insanity checks in the event handlers as well.


slluis:

PR #8089, opened 2019-07-01
Fixes VSTS Bug 938346: [Feedback] How to use subverision over svn+ssh

mkrueger:
WITHOUT a port?

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/938346

Port -1 is now the value to omit the port in the uri.


sevoku:
@mkrueger this fixes just the spinner, but we should also fix the "Url" label generation, it looks like it doesn't always update correctly which adds more confusion for the user.

PR #8097, opened 2019-07-01
Fixes VSTS 935137: Null Reference Exceptions when disposing TextViewContent

avodovnik:
This is an attempt to fix Null Reference Exceptions that happen when `TextViewContent` is disposed.

There were two potential cases where this would have potentially happened, one, where `sourceEditorOptions` was null, and second, when `TextView` was null in the call to `TextView.Options`.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/935137/


abock:

PR #8103, opened 2019-07-02
[Core] Fix null ref in DotNetProject.OnGetReferencedAssemblies

mrward:
Fixed a race condition in WorkspaceObject.BindTask. If a project is
disposed during the BindTask call but before the task was added to
activeTasks the MSBuildProject would be set null causing a null
reference exception in DotNetProject.OnGetReferencedAssemblies. This
would also resulted in an unobserved task exception since BindTask
would return a different canceled task.

To prevent this BindTask registers a TaskCompletionSource to the
list of activeTasks before it calls the Func that returns the Task.
This ensures that the project will not be disposed whilst that Func
is being called.

Also if an exception being thrown by the Func the TaskCompletionSource
task is removed from the active tasks to avoid the project waiting
forever for all active tasks to finish and the project's OnDispose
method never being called.

Fixes VSTS #849584 - [Watson] NullReferenceException in
DotNetProject.OnGetReferencedAssemblies

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
@@ -98,15 +98,30 @@ namespace MonoDevelop.Projects
 		/// <typeparam name="T">Task return type</typeparam>
 		public Task<T> BindTask<T> (Func<CancellationToken, Task<T>> f)
 		{
-			var t = f (disposeCancellation.Token);
+			var taskCompletionSource = new TaskCompletionSource<bool> ();
 			lock (activeTasks) {
+				// Need to register a task before the Func is called otherwise the Dispose could be called whilst the
+				// Func's Task is being run and before the Func's task is registered.
 				if (disposeCancellation.IsCancellationRequested)
 					return Task.FromCanceled<T> (disposeCancellation.Token);
-				activeTasks.Add (t);
+				activeTasks.Add (taskCompletionSource.Task);
+			}
+			Task<T> t;
+			try {
+				t = f (disposeCancellation.Token);
+			} catch (Exception ex) {
+				// Do not prevent Project.OnDispose being called by leaving the task on the activeTasks list.
+				lock (activeTasks) {
+					taskCompletionSource.SetResult (true);
+					activeTasks.Remove (taskCompletionSource.Task);
+				}
+				return Task.FromException<T> (ex);
 			}
 			t.ContinueWith (tr => {
-				lock (activeTasks)
-					activeTasks.Remove (t);
+				lock (activeTasks) {
+					taskCompletionSource.SetResult (true);
+					activeTasks.Remove (taskCompletionSource.Task);
+				}
 			});
 			return t;
 		}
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
```diff
@@ -98,15 +98,30 @@ protected void AssertMainThread ()
...
-				activeTasks.Add (t);
+				activeTasks.Add (taskCompletionSource.Task);
+			}
+			Task<T> t;
+			try {
+				t = f (disposeCancellation.Token);
+			} catch (Exception) {
+				// Do not prevent Project.OnDispose being called by leaving the task on the activeTasks list.
+				lock (activeTasks) {
+					taskCompletionSource.SetResult (true);
```
Therzok: return Task.FromException?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
```diff
@@ -98,15 +98,30 @@ protected void AssertMainThread ()
 		/// <typeparam name="T">Task return type</typeparam>
 		public Task<T> BindTask<T> (Func<CancellationToken, Task<T>> f)
 		{
-			var t = f (disposeCancellation.Token);
+			var taskCompletionSource = new TaskCompletionSource<bool> ();
```
Therzok: The TaskCompletionSource is never used?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
```diff
@@ -98,15 +98,30 @@ protected void AssertMainThread ()
...
+			Task<T> t;
+			try {
+				t = f (disposeCancellation.Token);
+			} catch (Exception) {
+				// Do not prevent Project.OnDispose being called by leaving the task on the activeTasks list.
+				lock (activeTasks) {
+					taskCompletionSource.SetResult (true);
+					activeTasks.Remove (taskCompletionSource.Task);
+				}
+				throw;
```
Therzok: Okay, to clarify my poor review comment, instead of throwing here:
```suggestion
				return Task.FromException (e);
```

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs
```diff
@@ -98,15 +98,30 @@ protected void AssertMainThread ()
...
+				// Func's Task is being run and before the Func's task is registered.
 				if (disposeCancellation.IsCancellationRequested)
 					return Task.FromCanceled<T> (disposeCancellation.Token);
-				activeTasks.Add (t);
+				activeTasks.Add (taskCompletionSource.Task);
+			}
+			Task<T> t;
+			try {
+				t = f (disposeCancellation.Token);
+			} catch (Exception) {
```
Therzok:
```suggestion
			} catch (Exception e) {
```

main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/ProjectTests.cs
```diff
@@ -1153,5 +1154,78 @@ protected internal override void OnModified (SolutionItemModifiedEventArgs args)
...
+
+				try {
+					await p.BindTask (arg => throw new ApplicationException ("Test"));
+					Assert.Fail ("Should not reach this line");
+				} catch (ApplicationException) {
+				}
+
+				try {
+					var result = p.BindTask<string> (arg => throw new ApplicationException ("Test 2"));
+					Assert.Fail ("Should not reach this line - BindTask<T>");
```
Therzok: This line should be reached. Waiting for the task should trigger the exception.

rodrmoya (approved):

mrward:

slluis (approved):

PR #8105, opened 2019-07-02
Fixes VSTS 917028: Calling ScrollToRequestedCaretLocation with an invalid line number caused an exception

avodovnik:
In some cases (like a deletion of code from the document, then trying to double click an existing Search Result), this method might have been called with a line number that no longer exists. This resulted in a call to the API that threw an exception. Now, the method checks the line count in the snapshot, and aborts. Both cases (a graceful failure, and an exception) are logged, so we can see if this continues to happen.

I was able to repro the steps to get the exception, but I couldn’t repro a crash of the IDE:

1. Have a class with a field, reference that field in a bunch of places
1. Select Find References, and keep the results
1. Delete most of the document, save it, and close it
1. Double click a reference

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentManager.cs
@@ -657,11 +657,24 @@ namespace MonoDevelop.Ide.Gui.Documents
 				} else {
 					var offset = info.Offset;
 					if (offset < 0) {
-						var line = textView.TextSnapshot.GetLineFromLineNumber (info.Line - 1);
-						if (info.Column >= 1) {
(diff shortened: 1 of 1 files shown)


avodovnik:
Small note: this exception _originated_ in the VS Platform part, shared with VS Windows. I’ve checked some of the 400+ references of the method in the Windows side of this, and most of them (the relevant ones, that accept potential input or don’t have an up-to-date state) check/guard for either line-count or catch the exception and return an error state. I felt the right thing to do here, is to do the same and guard it where one of the calls happen. 

We **might** see more exceptions originating from that API, and we should add guards accordingly, although I think most of them treat `DocumentManager` as a facade to it anyway, and call `ScrollToRequestedCaretLocation` to get to the part of the code requested.

abock:

slluis:
@abock @avodovnik this needs a manual backport

PR #8106, opened 2019-07-02
Fixes VSTS Bug 937839: [Feedback] Text Editor: font weight ignored

mkrueger:
Fixes VSTS #937839
Fixes VSTS #752348
Fixes VSTS #752341
Fixes VSTS #752307
Fixes VSTS #987295

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Fonts/FontChooserPanelWidget.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Fonts/FontChooserPanelWidget.cs
@@ -71,22 +71,6 @@ namespace MonoDevelop.Ide.Fonts
 			base.OnDestroyed ();
 		}
 
-#if MAC
-		class FontPanelDelegate : NSWindowDelegate
-		{
-			[Export ("validModesForFontPanel:")]
-			NSFontPanelMode ValidModesForFontPanel (NSFontPanel panel)
-			{
-				return NSFontPanelMode.CollectionMask | NSFontPanelMode.FaceMask | NSFontPanelMode.SizeMask;
-			}
-
-			public override void WillClose (NSNotification notification)
-			{
-				NSApplication.SharedApplication.StopModal ();
-			}
-		}
-
-#endif
 		public FontChooserPanelWidget ()
 		{
 			this.Build ();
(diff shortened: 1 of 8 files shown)


mkrueger:
One proposal to "fix" that - we should use the native font chooser. 
But parsing the font description string needs to be centralized. I've implemented it in FontService.

Thoughts ?

sevoku:
My feeling is that we should consider using Xwt for the conversions and make sure that it does this right, cause we use Xwt in many other different places to convert fonts and colors (incl. designers etc.). Another "pro" would be to have all "proper" conversion code in one place.

mkrueger:
Requires: https://github.com/mono/xwt/pull/960

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Fonts/FontChooserPanelWidget.cs
```diff
@@ -80,46 +84,50 @@ public FontChooserPanelWidget ()
...
 				};
 				hBox.PackStart (setFontButton, true, true, 0);
 
 				var setDefaultFontButton = new Button ();
 				setDefaultFontButton.Label = GettextCatalog.GetString ("Set To Default");
 				setDefaultFontButton.Clicked += delegate {
-					SetFont (desc.Name, IdeServices.FontService.GetFont (desc.Name).FontDescription);
-					setFontButton.Label = IdeServices.FontService.FilterFontName (GetFont (desc.Name));
+					SetFont (desc.Name, IdeServices.FontService.GetFontDescriptionCodon (desc.Name).FontDescription);
+					setDefaultFontButton.Label = GetDisplayDescription (IdeServices.FontService.GetFont (desc.Name));
```
sevoku:
```suggestion
					setFontButton.Label = GetDisplayDescription (IdeServices.FontService.GetFont (desc.Name));
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Fonts/FontService.cs
```diff
@@ -285,8 +312,14 @@ public static Xwt.Drawing.Font ToXwtFont (this FontDescription font)
...
 			var toolkit = withToolkit ?? Xwt.Toolkit.CurrentEngine;
+
 			Xwt.Drawing.Font xwtFont = null;
-			toolkit.Invoke (() => xwtFont = Xwt.Drawing.Font.FromName (font.ToString ()));
+			toolkit.Invoke (() => {
+				xwtFont = Xwt.Drawing.Font.FromName (font.Family + " " + (int)(font.Size / Pango.Scale.PangoScale))
+					.WithWeight ((Xwt.Drawing.FontWeight)font.Weight)
+					.WithStyle ((Xwt.Drawing.FontStyle)font.Style)
+					.WithStretch ((Xwt.Drawing.FontStretch)font.Stretch);
+		});
```
sevoku:
```suggestion
			});
```

sevoku:
@abock or @sandyarmstrong could you guys verify this to make sure that it works as expected with the native editor?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Fonts/FontService.cs
```diff
@@ -274,8 +283,26 @@ public static FontDescription ToPangoFont (this Xwt.Drawing.Font font)
...
 				return backend.Copy ();
-			return FontDescription.FromString (font.ToString ());
+			var description = FontDescription.FromString (font.Family + " " + font.Size);
+			description.Weight = (Pango.Weight)font.Weight;
+			description.Style = (Pango.Style)font.Style;
+			description.Stretch = (Pango.Stretch)font.Stretch;
+			return description;
 		}
+#if MAC
+		public static NSFont ToNSFont (this Xwt.Drawing.Font font)
```
netonjm: question, is this method safe from null cases or exceptions? we should try/catch and return default system font?

sandyarmstrong:
@sevoku this already worked as expected in the new editor. I'll download a build and make sure it still does.

sandyarmstrong:
@sevoku @mkrueger this is great, thanks! Love that we use native font picker now. Works fine in my testing.

abock:
The new editor maps these legacy Pango font descriptions appropriately to NSFont, but I would love for us to use the native font selector via `NSFontManager.SharedFontManager. OrderFrontFontPanel`, etc.

https://github.com/mono/monodevelop/blob/f9839cdd26fd7710315b7ad7fcb7bc7b585223c9/main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor.Cocoa/CocoaTextViewDisplayBinding.cs#L105-L132

abock:
To clarify - this is fantastic that ultimately Pango.FontDescription is not in play on mac now, and the native font panel is used (finally).

My only request is that for bringing the font panel up we actually use `NSFontManager` and its methods for presenting the panel and getting/setting fonts. It is an asynchronous API insomuch as there is an event to listen to for when the user changes the font selection.

The important part being that `ISelectFontDialogBackend` for the Cocoa implementation explicitly plays with an `NSFontPanel` and presents it as a modal dialog - you do not see modals for font panels in well designed macOS apps.

The desirable user experience for a font panel in the IDE is that it should be presented (non-modal) and allow the user to try out new fonts by having them update as the selection changes, much like the theme selector does. The XWT approach stifles this UX.

sevoku:
This PR is about fixing the font mapping bug and making font selection accessible.

> It is an asynchronous API insomuch as there is an event to listen to for when the user changes the font selection.

We can have a discussion about implementing that (in native or additional Xwt API, doesn't really matter). That'd be more than "listening to that event" though, cause it involves prefs service handling, cancellation etc. But this is not the topic and purpose of this PR here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Fonts/FontService.cs
Therzok: This is where nullability annotations shine, but we don't have nullability annotations in xammac/xwt.

abock:
Sure. Let's merge this. Then let's fix the UX. I don't know what you mean by prefs service or cancellation. I think we're talking about 5-10 lines of code here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Fonts/FontService.cs
sevoku: This seems to be safe, the callers do a null check and Xwt.Drawing.Font.FromName is safe (will return the system font if it fails).

sevoku:

PR #8115, opened 2019-07-03
Fixes VSTS Bug 933992: Add new folder dialog should be able to add

mkrueger:
existing folder

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/933992


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewFolderDialog.cs
```diff
@@ -137,16 +140,22 @@ string GetDefaultFolderName ()
 		void FolderNameTextEntryChanged (object sender, EventArgs e)
 		{
 			FilePath directoryPath = GetNewFolderPath ();
-
+			addButton.Label = GettextCatalog.GetString ("Create");
 			if (folderNameTextEntry.Text.Length == 0) {
```
mrward: Not sure about the label name change. The original UI was approved by the design board. This would need a review.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewFolderDialog.cs
```diff
@@ -137,16 +140,22 @@ string GetDefaultFolderName ()
...
 				addButton.Sensitive = false;
 				HidePopoverMessage ();
 			} else if (!IsValidFolderName (directoryPath, folderNameTextEntry.Text)) {
 				ShowWarning (GettextCatalog.GetString ("The name you have chosen contains illegal characters. Please choose a different name."));
 				addButton.Sensitive = false;
 			} else if (Directory.Exists (directoryPath)) {
-				ShowWarning (GettextCatalog.GetString ("Folder name is already in use. Please choose a different name."));
-				addButton.Sensitive = false;
+				var alreadyContainsDirectory = project.Files.Any (f => f.FilePath.ToString ().StartsWith (directoryPath, FilePath.PathComparison));
```
mrward: FilePath.IsChildPathOf?

mkrueger: Can be the same path.

mkrueger:
https://devdiv-design.visualstudio.com/D3%20Studio/_workitems/edit/7286

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewFolderDialog.cs
mkrueger: https://devdiv-design.visualstudio.com/D3%20Studio/_workitems/edit/7286

mrward:
The inline rename code was removed and replaced with an Add New Folder dialog because of various issues, such as when Git was enabled. The label edit would stop and you would have to select the folder again and select Rename.

mrward:
Whilst it is still possible for the new folder node to lose its edit state, if anything refreshes the Solution pad whilst the tree node is in the edit state, such as say a NuGet restore or a NuGet install, I cannot repro the original problem that affected projects that used git.

There is a slight delay before the label can be edited now that we wait for the project to be saved. Not sure if this is a big problem.

NewFolderDelay

slluis:
@sevoku I think you tried to fix the original problem, and at the end we decided to use a dialog because there was no simple solution to the problem.

Therzok:
AddNodeCallback would previously trigger a rename session, which ended the up being cancelled immediately, thus we couldn't rename the folder, afaik

mrward:
The Add New Folder dialog was added due to this bug - https://devdiv.visualstudio.com/DevDiv/_workitems/edit/721437 - there are some comments there.

Any Solution pad tree refresh can cause the label edit to be cancelled. Saving the project before doing the callback, which is done in this pull request, seems to solve the original problem in the above bug, which affected a solution using Git where the version control addin would trigger several Solution pad tree refreshes. However there are cases where a refresh could occur and cancel the refresh - one example is a NuGet install finishing.

mkrueger:
First I implemented another solution but I think I solved the old bug. However with a small delay but I think it's acceptable.

sevoku:
Oh yeah #7003, I was not directly involved, so I forgot about that. I think in general a dialog is not a good solution, but we should make sure not to bring back the issues we had. @mkrueger could you verify the NuGet scenario @mrward mentioned?

mrward:
Anything that triggers the refresh of the tree will cancel the label edit. It is not specific to NuGet. I picked NuGet since that is something that is relatively easy to trigger - there are likely to be other ways. No idea how you can prevent all of them. Before the version control addin was refreshing the tree a lot when a new folder was created, from what I remember.

mkrueger:
@sevoku the nuget scenario is a problem on "rename" as well - it's not specific to that one and in no way a problem here IMO. The git update was another story because that was triggered in all git projects and focus lost was a common thing.

sevoku:
So, I tried some different scenarios with this build and IMO it works better than before we introduced the new dialog and definitely much better than _with_ the dialog.

First of all in the regular use case "create a folder without any operation running" this is definitely the state of the art. A special dialog just for this operation is very dated and unexpected nowadays, so simply a bad UX and not aligned with regular OS behaviour. This was also the concensus on yesterdays UXBoard (with different opinions on wording of the new folder name tho).

If some background operation (i.e. save) is ongoing, it'll delay the node creation, which is not too bad IMO and won't be so unexpected if the status bar says "Saving project...". This seems to be an improverment to what we had before.

The other argument for fixing 721437 was that it's "frustrating to require an extra click", which is definitely not solved by adding a new dialog (it needs at least 2 extra clicks).

Another scenario mentioned in the discussion:
_Close node > refresh > select Add Folder > doesn't expand to the new folder node_: I can't reproduce this with this build. Works for me with .NET and Core project types. @mrward could you verify?

WRT Version Control: Git is now freezing FSW during operations, so that should not affect this as much as it did before. And it'll get even better with the new VCS API.

For sure there are still scenarios where the editing might get canceled, but it's much more unlikely to happen then it did back then when we tried to fix this.
_BUT my main point is:_ even then, it'd add just 2 more clicks (or single CMD+R) required to rename the folder in this rare case. So we reduce UX friction in almost all cases.

Another thing to note: same can happen in many other apps, like Finder, Explorer (on Windows), even VS on Windows. If something refreshes the view or requests attention for something, the operation canceles and the folder gets created with the default name. So it's not so unexpected and IMO still much better than a UX with a dialog.

slluis:
The problem is that this solution is very fragile. It may work now for the chain of refresh events we currently have, but it can break easily since *anything* at *any* time can cause the tree to be refreshed, and we can't be chasing down those refreshes every time this happens.

I only see two ways of correctly fixing this: 1) change the way the tree is refreshed, so that existing nodes are not recreated if they don't need to (so that if a node is being edited and it doesn't have changes it would keep being edited). 2) Freeze tree updates while a node is being edited.

Option 1) is probably the best solution, but it may not be easy to implement. We currently refresh the tree (or branches of the tree) by deleting all the nodes and reconstructing them. Instead of that, we should go node by node and do a differential update.

Option 2) is easier (just a matter of calling Lock/UnlockUpdates), although it may have weird side effects in some cases, since the tree won't show any updates until the editing session ends.

My suggestion is to try 2), and if the experience is not good, then we can consider 1).

sevoku:
Agree on 2), but I don't see the general issue with this being fragile in general (sry, maybe I fully misunderstand something). Are there any side effects besides the creation being cancelled? If not (like crasher or data loss) then it's imo not a big problem.

mkrueger:
"but it can break easily since anything at any time"

The same argument applies for: 

Add Solution folder
Rename

This is not specific to that change.

slluis:
> Agree on 2), but I don't see the general issue with this being fragile in general (sry, maybe I fully misunderstand something). Are there any side effects besides the creation being cancelled? If not (like crasher or data loss) then it's imo not a big problem.

By fragile I don't mean that it can crash, I mean that the user experience may easily break again. That's something you should really be concerned about.

slluis:
> "but it can break easily since anything at any time"
> 
> The same argument applies for:
> 
> Add Solution folder
> Rename
> 
> This is not specific to that change in any way.
> 

It is true that the problem is not specific to this use case. There are other uses cases where this could also happen. That's why we need a global solution.

mkrueger:
Found an easy fix : just the node that's edited can't be updated - update on other nodes work. If the node that's edited is updated the update is canceled. After an edit the node refreshes anyways, regardless of the edit is canceled or not.

slluis:
Unfortunately this won't work if a parent node is refreshed, since then all the children are removed and created again, so there won't be an UpdateNode call.

PR #8119, opened 2019-07-03
[BuildOutput] Simplifies and removes delay cancellation in SearchCanBeCanceled test

netonjm:
Simplifies and removes the delay cancellation in SearchCanBeCanceled test

Fixes VSTS #941229 - MonoDevelop.Ide.BuildOutputTests.CustomProject_SearchCanBeCanceled very unreliable

### DIFF
--- a/main/tests/Ide.Tests/MonoDevelop.Ide/BuildOutputTests.cs
+++ b/main/tests/Ide.Tests/MonoDevelop.Ide/BuildOutputTests.cs
@@ -132,7 +132,7 @@ namespace MonoDevelop.Ide
 		}
 
 		[Test]
-		public async Task CustomProject_SearchCanBeCanceled ()
+		public async Task CustomProject_StressSearchCanBeCanceled ()
 		{
 			BuildOutputNode firstMatch = null;
 
@@ -142,10 +142,27 @@ namespace MonoDevelop.Ide
 			for (int i = 0; i < 100; i++) {
 				await Task.WhenAll (Task.Run (async () => firstMatch = await search.FirstMatch ("Message ")),
 									Task.Delay (100).ContinueWith (t => search.Cancel ()));
-
-				Assert.Null (firstMatch, "Got a first match while search was canceled");
 				Assert.True (search.IsCanceled, "Search was not canceled");
 			}
+		}
+
+		[Test]
+		public async Task CustomProject_SearchCanBeCanceled ()
+		{
+			var bo = GenerateCustomBuild (10);
+			var search = new BuildOutputDataSearch (bo.GetRootNodes (true));
+
+			var searchTask = search.FirstMatch ("Message ");
+			search.Cancel ();
+
+			var firstMatch = await searchTask;
+			Assert.Null (firstMatch, "Got a first match, but the canceled search did not expect records");
+			Assert.True (search.IsCanceled, "Search was not canceled, but we expect cancel the operation");
+
+			searchTask = search.FirstMatch ("Message ");
+			firstMatch = await searchTask;
+			Assert.NotNull (firstMatch, "No record was found in the search, but one was expected to be found.");
+			Assert.False (search.IsCanceled, "Search was canceled for some unexpected reason");
 		}
 
 		[Test]


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs
```diff
@@ -424,14 +425,20 @@ public BuildOutputDataSearch (IReadOnlyList<BuildOutputNode> rootNodes)
...
+			//we set the new cancellation token (this is what we want to real cancel
+			cancellation = new CancellationTokenSource ();
+
+			//if there are other search in progress we want to cancel it
+			parentCancellation?.Cancel ();
+			//we set the parent cancellation for the next iteration
+			parentCancellation = cancellation;
+
+			//we want to clear any old result
 			currentSearchMatches.Clear ();
```
mrward: Not really sure I understand how this fixes the problem. The original code was cancelling the previous CancellationTokenSource.

The only difference here seems to be that the cancellationTokenSource is created before the currentSearchMatches are cleared. Would moving the `cancellation = new CancellationTokenSource ()` line to above the currentSearchMatches.Clear line fix the problem instead of adding this new parentCancellation?

mrward: Or do we only need to change the test code? It does not look like moving the `cancellation = new CancellationTokenSource ();` line would change anything here.

netonjm: you are right ! fixed!

main/tests/Ide.Tests/MonoDevelop.Ide/BuildOutputTests.cs
```diff
@@ -134,18 +134,20 @@ public async Task CustomProject_DataSearch ()
...
+
+			var firstMatch = await searchTask;
+			Assert.Null (firstMatch, "Got a first match while search was canceled");
+			Assert.True (search.IsCanceled, "Search was not canceled");
+
+			searchTask = search.FirstMatch ("Message ");
+			firstMatch = await searchTask;
+			Assert.NotNull (firstMatch, "Got a first match while search was canceled");
+			Assert.False (search.IsCanceled, "Search was not canceled");
 		}
```
rodrmoya: The idea on this test was, IIRC, to stress test search functionality to simulate a user typing, that's why there was that `for` loop, to start/cancel lots of searches in little time to simulate search being started/canceled as users type on the search box, so is there any way to keep the `for` loop instead of just doing a single cancellation?

rodrmoya: That said, maybe `BuildOutputSearch`'s `FirstMatch`/`NextMatch` methods should take into account the cancellation status -> https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.BuildOutputView/BuildOutput.cs#L425

sevoku: So, if the goal of the test was to stress test this, we can keep the for loop. However the main problem here is:

```Assert.Null (firstMatch, "Got a first match while search was canceled");```

If the search cancellation is defined as "return whatever has been found before cancellation" then this assertion is wrong, otherwise the search is broken cause it returns a result even after cancellation.

So the frist quesiton @rodrmoya: is a result after cancellation expected?

IF the search CAN return whatever has been found before cancellation, then the await is wrong since `firstMatch` is undefined here for the test (it can contain a result or not), we should test with a sane timeout. 

```
var result = search.FirstMatch ("Message ").Wait (100) //or 1000?
[...]
Assert.IsTrue (result, "Search did not complete after cancellation");
```

Comment (rodrmoya) [reply main/tests/Ide.Tests/MonoDevelop.Ide/BuildOutputTests.cs]: Sorry, missed the notification for this. So looking at the current code, indeed a non-null result can happen even after a cancellation, not ideal, but that's how the current code seems to behave. So yes, I guess we should just remove checking the result in the test, and just check for the cancellation

Comment (netonjm) [reply main/tests/Ide.Tests/MonoDevelop.Ide/BuildOutputTests.cs]: Fixed

File: main/tests/Ide.Tests/MonoDevelop.Ide/BuildOutputTests.cs [hunk 3]
```diff
@@ -142,12 +142,29 @@ public async Task CustomProject_SearchCanBeCanceled ()
...
+			var searchTask = search.FirstMatch ("Message ");
+			search.Cancel ();
+
+			var firstMatch = await searchTask;
+			Assert.Null (firstMatch, "Got a first match while search was canceled");
+			Assert.True (search.IsCanceled, "Search was not canceled");
+
+			searchTask = search.FirstMatch ("Message ");
+			firstMatch = await searchTask;
+			Assert.NotNull (firstMatch, "Got a first match while search was canceled");
```
sevoku: So this tests whether a BuildOutputDataSearch can be reused after cancellation? The messages seem to be wrong here then.

netonjm: fixed

netonjm:
rebased and fixes messages

sevoku:

PR #8135, opened 2019-07-04
Fixes VSTS Bug 942353: [FATAL] System.NullReferenceException exception

mkrueger:
in Mono.MHex.HexEditor.Dispose()

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/942353

Can't really determine the cause - Options should always be != null
but it's the only thing that can throw a null ref exception. But in
any case Dispose shouldn't cause a fatal exception.


mkrueger:
Y was one of my guesses as well :) - we'll see if it works.

mkrueger:

PR #8137, opened 2019-07-05
Fixes VSTS Bug 940831: "Authors" (Blame) tab fails to load

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/940831

Blame/Diff now works without an text editor attached. The default view
handlers hide the diff/blame/merge view and commands because they need
the current text from somewhere - that's done through the ITextBuffer
API which needs to be provided by the document controller. But for
read-only views it would be possible to use these commands with
potential custom view handlers that enable these controllers on
specific view contents.

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/BlameWidget.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/BlameWidget.cs
@@ -39,6 +39,7 @@ using MonoDevelop.Ide.Fonts;
 using MonoDevelop.Ide.Editor;
 using Microsoft.VisualStudio.Text.Editor;
 using Microsoft.VisualStudio.Text;
+using MonoDevelop.Core.Text;
 
 namespace MonoDevelop.VersionControl.Views
 {
@@ -115,7 +116,6 @@ namespace MonoDevelop.VersionControl.Views
 			GtkWorkarounds.FixContainerLeak (this);
 			
 			this.info = info;
-			var textView = info.Controller.GetContent<ITextView> ();
 
 			vAdjustment = new Adjustment (0, 0, 0, 0, 0, 0);
 			vAdjustment.Changed += HandleAdjustmentChanged;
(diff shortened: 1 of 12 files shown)


main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/BlameWidget.cs
```diff
@@ -129,10 +129,19 @@ public BlameWidget (VersionControlDocumentInfo info)
...
 			AddChild (hScrollBar);
 
-			var doc = new TextDocument (textView.TextSnapshot.GetText()) {
-				IsReadOnly = true,
-				MimeType = IdeServices.DesktopService.GetMimeTypeForContentType (textView.TextBuffer.ContentType)
-			};
+			var textView = info.Controller.GetContent<ITextView> ();
+			TextDocument doc;
+			if (textView != null) {
+				doc = new TextDocument (textView.TextSnapshot.GetText ()) {
```
Therzok: Nit, reduce duplicated code:

```csharp
string text;
string mimeType;

if (textView != null) {
    text = textView.TextSnapshot.GetText ();
    mimeType = IdeServices.DesktopService.GetMimeTypeForContentType (textView.TextBuffer.ContentType);
} else {
    text = TextFileUtility.GetText (info.Item.Path);
    mimeType = IdeServices.DesktopService.GetMimeTypeForUri (info.Item.Path);
}

var doc = new TextDocument (text, mimeType) {
    IsReadOnly = true,
};
```

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/EditorCompareWidgetBase.cs
```diff
@@ -608,9 +608,16 @@ void HandleInfoDocumentTextEditorDataDocumentTextReplaced (object sender, TextCh
 		public void UpdateLocalText ()
 		{
 			var textBuffer = info.Controller.GetContent<ITextBuffer> ();
+			string localText = null;
```
Therzok: Style nit:

```
string localText = textBuffer?.CurrentSnapshot.GetText ()
	?? TextFileUtility.GetText (info.Item.Path);
```

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/DefaultBlameViewHandler.cs
```diff
@@ -29,16 +29,23 @@
...
 
 namespace MonoDevelop.VersionControl
 {
 	public class DefaultBlameViewHandler : IVersionControlViewHandler
 	{
-		public bool CanHandle (VersionControlItem item, DocumentController controller)
+		public static bool DefaultVCSViewCanHandle (VersionControlItem item, DocumentController controller)
 		{
-			return (controller is FileDocumentController || controller == null) && item.Repository.GetFileIsText (item.Path);
+			if (controller == null)
```
Therzok:
```suggestion
			if (controller is null)
```

slluis:

PR #8138, opened 2019-07-05
[DO NOT MERGE] log termination

iainx:
DO NOT MERGE just for debugging 942401


main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -683,30 +688,6 @@ void GlobalSetup ()
 		{
 			//FIXME: should we remove these when finalizing?
 			try {
-				ApplicationEvents.Quit += delegate (object sender, ApplicationQuitEventArgs e)
-				{
-					// We can only attempt to quit safely if all windows are GTK windows and not modal
-					if (!IsModalDialogRunning ()) {
```
Therzok: We lose the ability for handlers to stop quitting the IDE, like save dirty files. Maybe we should queue a continuation to quit after `Exit()` is done.

PR #8139, opened 2019-07-05
[Shell] Added Telemetry Counter to detect slow update command operations

jsuarezruiz:
Added Telemetry Counter to detect slow update command operations

Issue VSTS #623224

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -66,7 +66,8 @@ namespace MonoDevelop.Components.Commands
 		KeyboardShortcut[] chords;
 		string chord;
 		internal const int SlowCommandWarningTime = 25;
-		
+		internal const int SlowUpdateCommandWarningTime = 20;
+
 		Dictionary<object,Command> cmds = new Dictionary<object,Command> ();
 		Hashtable handlerInfo = new Hashtable ();
 		List<ICommandBar> toolbars = new List<ICommandBar> ();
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -553,8 +554,16 @@ bool ProcessKeyEventCore (Gdk.EventKey ev)
 
 			for (int i = 0; i < commands.Count; i++) {
 				CommandInfo cinfo = GetCommandInfo (commands [i].Id, new CommandTargetRoute ());
-				if (cinfo.IsUpdatingAsynchronously)
-					cinfo.UpdateTask.Wait (); // Not nice, but we need a synchronous result here
+				if (cinfo.IsUpdatingAsynchronously) {
+					// Not nice, but we need a synchronous result here
+					if (!cinfo.UpdateTask.Wait (SlowUpdateCommandWarningTime)) {
```
Therzok: Before, we would wait indefinitely for the task to end: see https://github.com/mono/monodevelop/issues/4917

Now, we are waiting 20ms, and then continuing. This means we are giving every update handler a maximum of 20ms to run. This will most likely cause handler updates to run _after_ the 20ms, without us doing anything about it.

For example, let's say an update task takes 25ms to run and has a continuation which does does something on the main thread.

Due to this code change, the continuation might run after the actual run handler is called.

Therzok: We need to also cancel the async update.

jsuarezruiz: Right. I think I should review all this block and look for a better option.

PR #8143, opened 2019-07-05
[Mac] Don't hang when terminating through the dock icon

iainx:
It turns out the answer to

// FIXME: could this block in rare cases?

was "No, it blocks every time", so make the delegate async and await on the call instead.

Also remove the old Carbon handler and move the important bits into the Cocoa handler instead as it hasn't been deprecated for at least a decade

Fixes VSTS #942401


slluis:

slluis:
Shouldn't we keep the IsModalDialogRunning check?

I'm merging since it fixes the hang, but I feel we should keep the modal check.

PR #8145, opened 2019-07-05
Move FileNesting operations on projects to NodeBuilder's

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/933930
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/918300
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/943240
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/923128
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/943248
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/948329


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -63,5 +65,15 @@ public static FilePath GetParentFile (Project project, FilePath inputFile)
 
 			return null;
 		}
+
+		public static bool HasChildren (Project project, FilePath inputFile)
```
Therzok: These two will be called from code which is supposed to run fast and not be allocation heavy, as it happens when we build the solution tree. Can we drop the LINQ?

Also, can we not do O(N) lookup?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
```diff
@@ -134,7 +142,7 @@ public override int CompareObjects (ITreeNavigator thisNode, ITreeNavigator othe
 		public override bool HasChildNodes (ITreeBuilder builder, object dataObject)
 		{
 			ProjectFile file = (ProjectFile) dataObject;
-			return file.HasChildren;
+			return file.HasChildren ? true : FileNestingService.HasChildren (file.Project, file.FilePath);
```
Therzok:
```csharp
return file.HasChildren || FileNestingService.HasChildren (file.Project, file.FilePath);
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
```diff
@@ -143,6 +151,12 @@ public override void BuildChildNodes (ITreeBuilder treeBuilder, object dataObjec
 			ProjectFile file = (ProjectFile) dataObject;
 			if (file.HasChildren)
 				treeBuilder.AddChildren (file.DependentChildren);
+			else {
+				var children = FileNestingService.GetChildren (file.Project, file.FilePath);
```
Therzok: GetChildren never returns null. Either we check for empty enumerable or change the API from IEnumerable.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -63,5 +65,15 @@ public static FilePath GetParentFile (Project project, FilePath inputFile)
...
+		{
+			return project.Files.Any (x => GetParentFile (project, x.FilePath) == inputFile);
+		}
+
+		public static IEnumerable<ProjectFile> GetChildren (Project project, FilePath inputFile)
```
Therzok: Why return IEnumerable?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: Yes, need to add a cache, will be doing that in the next commits

Therzok:
2019-07-08T18:10:35.7451160Z 1) MonoDevelop.Ide.PerfTests.TestSolutionLoad.TestLoad
2019-07-08T18:10:35.7451380Z Performance regression. Baseline: 8.248, Result: 14.326 (+0.74)

This is a pretty big regression (taking into account others have perf improvements)

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -52,16 +54,32 @@ static void HandleRulesProviderExtension (object sender, ExtensionNodeEventArgs
...
 					return parentFile;
 				}
 			}
 
 			return null;
 		}
+
+		public static bool HasChildren (ProjectFile inputFile)
+		{
+			return inputFile.Project.Files.Any (x => GetParentFile (x) == inputFile);
```
Therzok: O(N) lookup.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: I've added now a sort of "cache", still in progress, but makes performance much better than before, where we were just making lots and lots of calls to the NestingRuleProvider's

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: Because I wanted to use `ProjectFileCollection` internally, which only implements IEnumerable. Will look at improving it

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: Yes, this is gone now

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -52,16 +55,137 @@ static void HandleRulesProviderExtension (object sender, ExtensionNodeEventArgs
...
+			}
+		}
+
+		public ProjectFile GetParentForFile (ProjectFile inputFile)
+		{
+			EnsureInitialized ();
+			return projectFiles.TryGetValue (inputFile, out var nestingInfo) ? nestingInfo.Parent : null;
+		}
+
+		public ProjectFileCollection GetChidlrenForFile (ProjectFile inputFile)
```
slluis: Typo: GetChidlrenForFile -> GetChildrenForFile

slluis: ProjectFileCollection allows adding/removing items. I don't think we support that in this data structure.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreNestingRulesProvider.cs
```diff
@@ -37,5 +40,11 @@ public AspNetCoreNestingRulesProvider ()
 		{
 			SourceFile = AddinManager.CurrentAddin.GetFilePath (Path.Combine ("Resources", "AspNetCore.filenesting.json"));
 		}
+
+		protected override bool AppliesToProject (Project project)
+		{
+			var dotnetProj = project as DotNetProject;
+			return dotnetProj != null && dotnetProj.MSBuildProject.GetReferencedSDKs ().FirstOrDefault (x => x.IndexOf ("Microsoft.NET.Sdk.Web", StringComparison.OrdinalIgnoreCase) != -1) != null;
```
Therzok: [UsingMicrosoftNETSdkWeb](https://github.com/aspnet/websdk/blob/e8cb5ed1357d232251f26cf79413838ea62ba501/src/Web/Sdk/Sdk.props#L17) should be a more reliable way to do this.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -25,14 +25,17 @@
...
+using System.Linq;
 using Mono.Addins;
-using MonoDevelop.Core;
 
 namespace MonoDevelop.Projects.FileNesting
 {
-	internal static class FileNestingService
+	public static class FileNestingService
 	{
+		static readonly ConcurrentDictionary<Project, ProjectNestingInfo> loadedProjects = new ConcurrentDictionary<Project, ProjectNestingInfo> ();
```
Therzok: We indefinitely leak projects now. What we want here is a `ConditionalWeakTable`

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -52,16 +55,137 @@ static void HandleRulesProviderExtension (object sender, ExtensionNodeEventArgs
...
+			if (!loadedProjects.TryGetValue (project, out var nestingInfo)) {
+				nestingInfo = loadedProjects [project] = new ProjectNestingInfo (project);
+			}
+
+			return nestingInfo;
+		}
+
+		internal static ProjectFile InternalGetParentFile (ProjectFile inputFile)
 		{
+			if (inputFile.Project.UserProperties.HasValue ("FileNesting") && !inputFile.Project.UserProperties.GetValue<bool> ("FileNesting")) {
```
Therzok: Any reason we check for the value existing first?

```suggestion
			if (!inputFile.Project.UserProperties.GetValue<bool> ("FileNesting", false)) {
```

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -52,16 +55,137 @@ static void HandleRulesProviderExtension (object sender, ExtensionNodeEventArgs
...
+			return (children?.Count ?? 0) > 0;
+		}
+
+		public static IEnumerable<ProjectFile> GetChildren (ProjectFile inputFile)
+		{
+			return GetProjectNestingInfo (inputFile.Project).GetChidlrenForFile (inputFile);
+		}
 	}
+
+	class ProjectNestingInfo : IDisposable
```
Therzok:
```suggestion
	sealed class ProjectNestingInfo : IDisposable
```

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -52,16 +55,137 @@ static void HandleRulesProviderExtension (object sender, ExtensionNodeEventArgs
...
+	{
+		class ProjectFileNestingInfo
```
Therzok:
```suggestion
		sealed class ProjectFileNestingInfo
```

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
```diff
@@ -59,71 +59,74 @@ public NestingRule (NestingRuleKind kind, string appliesTo, IEnumerable<string>
 
 		public string AppliesTo { get; private set; }
 
-		bool CheckParentForFile (Project project, FilePath inputFile, FilePath parentFile)
+		ProjectFile CheckParentForFile (ProjectFile inputFile, FilePath parentFile)
```
Therzok: From the looks of the callers, this is hot code. Avoid LINQ.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
```diff
@@ -59,71 +59,74 @@ public NestingRule (NestingRuleKind kind, string appliesTo, IEnumerable<string>
...
+			return null;
 		}
 
-		public FilePath GetParentFile (Project project, FilePath inputFile)
+		public ProjectFile GetParentFile (ProjectFile inputFile)
 		{
 			string parentFile, inputExtension;
 
-			inputExtension = Path.GetExtension (inputFile);
+			inputExtension = inputFile.FilePath.Extension;
```
Therzok: We have [FilePath.HasExtension](http://source.monodevelop.com/#MonoDevelop.Core/MonoDevelop.Core/FilePath.cs,dff70377a8fc9935)

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
```diff
@@ -59,71 +59,74 @@ public NestingRule (NestingRuleKind kind, string appliesTo, IEnumerable<string>
...
+					return parent;
+				}
 			}
 
-			return false;
+			return null;
 		}
 
-		public FilePath GetParentFile (Project project, FilePath inputFile)
+		public ProjectFile GetParentFile (ProjectFile inputFile)
```
Therzok: This is again, potentially unbounded hot code. Avoid string allocations where possible. ParentDirectory is done in a loop multiple times.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRulesProvider.cs
```diff
@@ -121,8 +121,14 @@ static bool LoadFromFile (NestingRulesProvider provider)
 			return false;
 		}
 
-		public FilePath GetParentFile (Project project, FilePath inputFile)
+		protected virtual bool AppliesToProject (Project project) => true;
```
Therzok: Shouldn't this use the Pad configuration property added [here](https://github.com/mono/monodevelop/pull/8145/files#diff-a07347c1eebf4a28b82422c42025f40fR52) ?

Therzok: Either the FileNestingService has to use it (which seems wrong) or the node builders should. Marking as resolved and commenting there.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
```diff
@@ -92,7 +93,7 @@ void GetFolderContent (Project project, string folder, out List<ProjectFile> fil
 				
 				if (file.Subtype != Subtype.Directory) {
 					// If file depends on something other than a directory, continue
-					if (file.DependsOnFile != null && file.DependsOnFile.Subtype != Subtype.Directory)
+					if ((file.DependsOnFile != null && file.DependsOnFile.Subtype != Subtype.Directory) || FileNestingService.HasParent (file))
```
Therzok: New pad option isn't being used?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
```diff
@@ -96,11 +97,17 @@ public override object GetParentObject (object dataObject)
...
 			var dir = !file.IsLink ? file.FilePath.ParentDirectory : file.Project.BaseDirectory.Combine (file.ProjectVirtualPath).ParentDirectory;
-			
+
 			if (!string.IsNullOrEmpty (file.DependsOn)) {
 				ProjectFile groupUnder = file.Project.Files.GetFile (file.FilePath.ParentDirectory.Combine (file.DependsOn));
 				if (groupUnder != null)
 					return groupUnder;
+			} else {
+				// File nesting
+				var parentFile = FileNestingService.GetParentFile (file);
```
Therzok: New pad option isn't being used?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
```diff
@@ -134,7 +141,7 @@ public override int CompareObjects (ITreeNavigator thisNode, ITreeNavigator othe
 		public override bool HasChildNodes (ITreeBuilder builder, object dataObject)
 		{
 			ProjectFile file = (ProjectFile) dataObject;
-			return file.HasChildren;
+			return file.HasChildren || FileNestingService.HasChildren (file);
```
Therzok: New pad option isn't being used?

Therzok:
Done with first review pass

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: yes, was going to add code to listen to ProjectClosed events, but I guess a `ConditionalWeakTable` looks even better, will change to that

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: Yeah, good one, fixed now, with a default of `true` instead, as that's why I was calling HasValue + GetValue, didn't see the overloaded method with the default

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: Changed

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
rodrmoya: hmm, but this is to retrieve the extension, not to check if it has an extension, so what do you mean?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
rodrmoya: Fixed in https://github.com/mono/monodevelop/pull/8145/commits/dd64e9ba181098eebedb6949099b57f29fc440b0

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
rodrmoya: Yes, it's used everywhere via `InternalGetParentFile`, which is what's used to fill the nesting information. These other methods just retrieve the already initialised data from the cache

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
rodrmoya: ditto

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
rodrmoya: and ditto :)

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -60,7 +60,8 @@ static void HandleRulesProviderExtension (object sender, ExtensionNodeEventArgs
 		static ProjectNestingInfo GetProjectNestingInfo (Project project)
 		{
 			if (!loadedProjects.TryGetValue (project, out var nestingInfo)) {
-				nestingInfo = loadedProjects [project] = new ProjectNestingInfo (project);
+				nestingInfo = new ProjectNestingInfo (project);
+				loadedProjects.Add (project, nestingInfo);
```
Therzok: This can throw an exception.

Therzok: Do a TryGetValue then a [GetOrCreateValue](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2.getvalue?view=netframework-4.8#System_Runtime_CompilerServices_ConditionalWeakTable_2_GetValue__0_System_Runtime_CompilerServices_ConditionalWeakTable__0__1__CreateValueCallback_) to ensure we're exception free from races.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -29,6 +29,8 @@
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Runtime.CompilerServices;
+using System.Web.UI.WebControls;
```
Therzok: Huh?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -52,16 +56,225 @@ static void HandleRulesProviderExtension (object sender, ExtensionNodeEventArgs
...
+
+			return projectFiles.TryGetValue (inputFile, out var nestingInfo) ? nestingInfo.Children : null;
+		}
+
+		public void Dispose ()
+		{
+			Project.FileAddedToProject -= OnFileAddedToProject;
+			Project.FileRemovedFromProject -= OnFileRemovedFromProject;
+			Project.FileRenamedInProject += OnFileRenamedInProject;
+			Project.ParentSolution.UserProperties.Changed -= OnUserPropertiesChanged;
```
Therzok: Maybe also clear `projectFiles`

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: Yeah, autocompleting to wrong types added this, fixed now

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: Fixed with a single call to `GetValue`

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/NestingRule.cs
Therzok: Old code used to work differently.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -52,16 +56,210 @@ static void HandleRulesProviderExtension (object sender, ExtensionNodeEventArgs
...
+		void AddFile (ProjectFile projectFile)
+		{
+			var tmp = projectFiles.GetOrAdd (projectFile, new ProjectFileNestingInfo (projectFile));
+			tmp.Parent = FileNestingService.InternalGetParentFile (projectFile);
+			if (tmp.Parent != null) {
+				var parent = projectFiles.GetOrAdd (tmp.Parent, new ProjectFileNestingInfo (tmp.Parent));
+				if (parent.Children == null) {
+					parent.Children = new ProjectFileCollection ();
+				}
+				if (!parent.Children.Contains (projectFile)) {
```
Therzok: Maybe make this an O(1):

```csharp
var existing = parent.Children.GetFile (projectFile.FilePath);
if (existing == null) {
    parent.Children.Add (projectFile);
}
```

Therzok (approved):
Other than comment, LGTM

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: Coming in the next commit

Therzok (approved):

Therzok:
cc @sevoku @mrward 👀

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -52,16 +56,210 @@ static void HandleRulesProviderExtension (object sender, ExtensionNodeEventArgs
...
 		}
+
+		public static event Action<Project> NestingRulesChanged;
+
+		internal static void NotifyNestingRulesChanged (Project project) => NestingRulesChanged?.Invoke (project);
+
+		public static bool IsEnabledForProject (Project project)
+		{
+			return project?.ParentSolution?.UserProperties?.GetValue<bool> ("MonoDevelop.Ide.FileNesting.Enabled", true) ?? true;
+		}
```
mrward: Still wondering if we should move this to MonoDevelop.Ide. Seems a bit odd to me to be looking for a MonoDevelop.Ide.FileNesting.Enabled property here.

Also presumably if it was in MonoDevelop.Ide the FileNestingService class would not need to be public?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
```diff
@@ -52,16 +56,210 @@ static void HandleRulesProviderExtension (object sender, ExtensionNodeEventArgs
...
+		public static bool HasChildren (ProjectFile inputFile)
+		{
+			var children = GetProjectNestingInfo (inputFile.Project).GetChildrenForFile (inputFile);
+			return (children?.Count ?? 0) > 0;
+		}
+
+		public static IEnumerable<ProjectFile> GetChildren (ProjectFile inputFile)
+		{
+			return GetProjectNestingInfo (inputFile.Project).GetChildrenForFile (inputFile);
+		}
```
mrward: Returning null from an IEnumerable seems wrong to me. GetChildrenForFile seems to return null when file nesting is disabled. Can it instead return an empty enumerable?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
```diff
@@ -0,0 +1,266 @@
...
+
+		public static bool HasChildren (ProjectFile inputFile)
+		{
+			var children = GetProjectNestingInfo (inputFile.Project).GetChildrenForFile (inputFile);
+			return (children?.Count ?? 0) > 0;
+		}
+
+		public static IEnumerable<ProjectFile> GetChildren (ProjectFile inputFile)
+		{
+			return GetProjectNestingInfo (inputFile.Project).GetChildrenForFile (inputFile);
```
Therzok:
```suggestion
			return GetProjectNestingInfo (inputFile.Project).GetChildrenForFile (inputFile) ?? Array.Empty<ProjectFile> ();
```

mrward: We could add this Array.Empty to the GetChildrenForFile method. Then no need for any null checks anywhere for the IEnumerable<ProjectFile> items.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
```diff
@@ -0,0 +1,266 @@
...
+using System.Collections.Immutable;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using Mono.Addins;
+using MonoDevelop.Projects;
+
+namespace MonoDevelop.Ide.Projects.FileNesting
+{
+	public static class FileNestingService
+	{
```
mrward: Internal?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
Therzok: GetChildrenForFile returns a ProjectFileCollection though.

mrward: Good point 😄 I should have checked first.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
```diff
@@ -143,6 +148,12 @@ public override void BuildChildNodes (ITreeBuilder treeBuilder, object dataObjec
 			ProjectFile file = (ProjectFile) dataObject;
 			if (file.HasChildren)
 				treeBuilder.AddChildren (file.DependentChildren);
+			else {
+				var children = FileNestingService.GetChildren (file);
+				if (children?.Any () ?? false) {
```
mrward: Would not be null with Marius's suggested code change.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
rodrmoya: Yes, moved to internal in the last commit

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.FileNesting/FileNestingService.cs
rodrmoya: Both done

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
rodrmoya: I was going to revisit using a IEnumerable there, but don't really want to publicise the internal `ProjectFileCollection`, so I'm thinking about maybe using an `ImmutableList` internally and return a IReadOnlyList in the public API. I think it's better to check for null than calling `Any`, `Count`, etc on the `IEnumerable`, as this code gets called a lot.

Makes sense?

Therzok:
LGTM

mrward (approved):

rodrmoya:

rodrmoya:
Backported manually at https://github.com/mono/monodevelop/pull/8495

PR #8153, opened 2019-07-08
[GetToCode] Adds ShowAsync method in NewProjectController with TaskCo…

netonjm:
This PR fixes a problem waiting the response from NewProject window probably introduced by 
https://github.com/mono/monodevelop/commit/1578ca3ed2ebfc96890e23fe580f6ede097221b8#diff-e3e1db8132f317c043c737963f36d108

The problem is now we close the dialog very quickly without wait the creation of the project, then the variable is modified after return the value.

This is necessary to fix, because GTC in case of project creation needs to disable the window to don't allow the user to click buttons meanwhile we open the project.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
@@ -679,45 +679,46 @@ namespace MonoDevelop.Ide.Projects
 			}
 
 			dialog.CloseDialog ();
-		
-			if (ParentFolder != null)
-				await IdeApp.ProjectOperations.SaveAsync (ParentFolder.ParentSolution);
-			else
-				await IdeApp.ProjectOperations.SaveAsync (processedTemplate.WorkspaceItems);
 
-			CreateVersionControlItems ();
-
-			if (OpenSolution) {
-				DisposeExistingNewItems ();
-				TemplateWizard wizard = wizardProvider.CurrentWizard;
-				if (await OpenCreatedSolution (processedTemplate)) {
-					var sol = IdeApp.Workspace.GetAllSolutions ().FirstOrDefault ();
-					if (sol != null) {
-						if (wizard != null)
-							wizard.ItemsCreated (new [] { sol });
-						InstallProjectTemplatePackages (sol);
-					}
-				}
-			}
-			else {
-				// The item is not a solution being opened, so it is going to be added to
-				// an existing item. In this case, it must not be disposed by the dialog.
-				RunTemplateActions (processedTemplate);
-				if (wizardProvider.HasWizard)
-					wizardProvider.CurrentWizard.ItemsCreated (processedTemplate.WorkspaceItems);
+			try {
 				if (ParentFolder != null)
-					InstallProjectTemplatePackages (ParentFolder.ParentSolution);
-			}
+					await IdeApp.ProjectOperations.SaveAsync (ParentFolder.ParentSolution);
+				else
+					await IdeApp.ProjectOperations.SaveAsync (processedTemplate.WorkspaceItems);
 
-			wizardProvider.Dispose ();
-			IsNewItemCreated = true;
-			UpdateDefaultSettings ();
+				CreateVersionControlItems ();
 
-			Gtk.Application.Invoke ((sender, args) => {
+				if (OpenSolution) {
+					DisposeExistingNewItems ();
+					TemplateWizard wizard = wizardProvider.CurrentWizard;
+					if (await OpenCreatedSolution (processedTemplate)) {
+						var sol = IdeApp.Workspace.GetAllSolutions ().FirstOrDefault ();
+						if (sol != null) {
+							if (wizard != null)
+								wizard.ItemsCreated (new [] { sol });
+							InstallProjectTemplatePackages (sol);
+						}
+					}
+				} else {
+					// The item is not a solution being opened, so it is going to be added to
+					// an existing item. In this case, it must not be disposed by the dialog.
+					RunTemplateActions (processedTemplate);
+					if (wizardProvider.HasWizard)
+						wizardProvider.CurrentWizard.ItemsCreated (processedTemplate.WorkspaceItems);
+					if (ParentFolder != null)
+						InstallProjectTemplatePackages (ParentFolder.ParentSolution);
+				}
+
+				wizardProvider.Dispose ();
+				IsNewItemCreated = true;
+				UpdateDefaultSettings ();
+
 				projectCreated.SetResult (true);
-				ProjectCreated?.Invoke (this, EventArgs.Empty);
-			});
-			await projectCreated.Task;
+				await Runtime.RunInMainThread (() => ProjectCreated?.Invoke (this, EventArgs.Empty));
+			} catch (Exception ex) {
+				projectCreated.SetException (ex);
+				throw;
+			}
 		}
 
 		public Task<bool> ProjectCreation => projectCreated?.Task;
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
```diff
@@ -54,6 +54,7 @@ class NewProjectDialogController : INewProjectDialogController
 	{
 		public event EventHandler ProjectCreationFailed;
 		public event EventHandler ProjectCreated;
+		public event EventHandler ProjectCreationWindowClosed;
```
mrward: Just wondering if this would be better as a WindowClosed or DialogClosed event with a bool that indicates that post project creation actions are still running so you could decide whether or not to show/hide the welcome page/window.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
```diff
@@ -184,6 +186,28 @@ public bool Show ()
...
+
+			dialog.ShowDialog ();
+
+			await EnsureCreationFinishedAsync ();
+
+			imageProvider.Dispose ();
+
+			return IsNewItemCreated;
+		}
+
```
mrward: Another idea might be for us to fix the original `bool Show()` method so it returned the correct bool flag (true if project will be created, false if not) and instead add a Task (ProjectCreation?) on the controller which you could optionally await if you wanted to wait until all the post project creation actions are complete.

Code would be:

    if (newProjectDialog.Show ()) {
        HideWelcomePage ();
        await newProjectDialog.ProjectCreation;
    }

Main reason here is that ShowAsync seems to be Showing the dialog but also waiting for things to complete after the dialog.Show has finished.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
```diff
@@ -710,10 +710,19 @@ public async Task<bool> NewSolution (string defaultTemplate, bool showTemplateSe
 
 			IdeApp.Workbench.EnsureLayout ();
 			var newProjectDialog = new NewProjectDialogController ();
+
+			static void projectCreationWindowClosed (object s, EventArgs e)
+			{
+				WelcomePage.WelcomePageService.HideWelcomePageOrWindow ();
```
mrward: Should ProjectOperations know about the WelcomePage? Can this be done elsewhere? When the main IDE workbench loads the solution does that not hide the welcome page?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
netonjm: good idea!

sevoku:
Soo, we have two different problems here, which are kind of related.

**ProjectOperations.NewSolution always returns false**

-> fix ProjectOperations.NewSolution as proposed by @mrward in https://github.com/mono/monodevelop/pull/8153#discussion_r301269277, means:
  * add a `public Task NewProjectController.ProjectCreation` (with a backing `private TaskCompletionSource<bool> projectCreated` and run it to completion with `true` after project creation or `false` on dialog cancellation (close without creating a project).
  * the result of `NewProjectDialog.Show` should return bool/false depending on the _state_ (not necessarily result) of the new `NewProjectController.ProjectCreation`, at the end: `return !(ProjectCreation.IsCompleted && ProjectCreation.Result = false)` (ran to completion with `false`)
  * after `Show()` completes (optionally check if it retuned `true`) await `NewProjectController.ProjectCreation`

**WelcomeWindow needs to be closed right after the new project dialog _without_ waiting for the shell to appear and hide it** (there is a certain amount of time between NPD being closed and the main IDE Window trigger, leading to an undefined GTC state where it doesn't know what to do).

The problem here is that GTC is using `ProjectOperations` which diesn't expose it's `NewProjectController`, so even having the event there, won't solve the issue (GTC simply has no access to the instance). The easiest solution is to make ProjectOperations hide the welcome window:

with the fix from above NewSolution would do this at the end:

```
- return newProjectDialog.Show ();
+ if (newProjectDialog.Show ()) {
+     WelcomePageService.HideWelcomePageOrWindow ();
+     return await newProjectDialog.ProjectCreation;
+ }
+ return false;
```

sevoku:
Ok the new logic is as follows:
* `NewProjectController.Show ()` returns once the dialog is closed with a result:
  * `false`: dialog closed without creating a projects (close/cancel)
    * `NewProjectController.ProjectCreation` is `null` (indicating that no project creation has been started)
  * `true`: user chose to create a project
    * dialog finished santiy checks and kicked off project creation
    * actual creation can be awaited with `await NewProjectController.ProjectCreation`
      * `true` once the project is loaded
      * `false` creation failed and or project not loaded
 * ProjectOperations.NewSolution will:
   * Call `NewProjectController.Show ()`
   * Close the WelcomePage
   * Await project creation
   * Return the result of the whole operation

netonjm:
rebased and squased

sevoku:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
```diff
@@ -667,15 +670,16 @@ public async Task Create ()
...
 					}
 					File.Delete (solutionFileName);
 				}
 			}
 
-			dialog.CloseDialog ();
-
+			dialog.CloseDialog ();
+		
 			if (ParentFolder != null)
```
mrward: Can we test what happens if say SaveAsync throws an exception here? It looks like we would be stuck waiting for the projectCreated.Task since it will never be set.

Should we have a try/catch around everything after CloseDialog till the end of the method?

sevoku: yeah, we should try/catch the whole thing

sevoku: Sooo, it looks like we actually never checked or catched this at all. Everything is using `NewProjectController.Show ()` which is decoupled from the actual project creation, so if that fails nothing would happen. Before we started closing the dialog earlier in the process, it would have just been kept open without notification and now since we close it we can end up without a visible window at all. Fixed in 4915a980fc01708e56b16943a511dfe97c94701d now.

mrward: Do you know what happens now if there is inside that try/catch exception? Does anything catch that exception that is rethrown?

sevoku: When using commands, it'll be logged without notification. But we really need to redesign how the controller works (including making all ProjectOperations async and properly handling errors).

sevoku:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs
```diff
@@ -181,9 +181,9 @@ public bool Show ()
 
 			imageProvider.Dispose ();
 
-			return IsNewItemCreated;
-		}
-
+			return ProjectCreation != null && !(ProjectCreation.IsCompleted && !ProjectCreation.Result);
```
Therzok: Why do we synchronously wait here?

Therzok: This is basically invalidating the await in ProjectOperations by doing a synchronous wait here.

Therzok:
```csharp
if (ProjectCreation == null || !ProjectCreation.IsCompleted)
    return false;

return ProjectCreation.Result;
```

Would've probably been clearer.

PR #8155, opened 2019-07-09
Fix svn probe in Xcode.app

Therzok:
Fixes VSTS #935540 - Subversion addin could not load libsvn_client, so it will be disabled.


sevoku:

PR #8156, opened 2019-07-09
[Mac] Don't reshow the popover when the view has been removed from th…

iainx:
…e window

ViewDidMoveToWindow can be called when removing the view from a parent window,
in which case we should not try to redisplay any popovers

Fixes VSTS #929008


iainx:

sevoku:

PR #8163, opened 2019-07-10
Fixes VSTS Bug 940369: [Git] Preserve username when checking out a

mkrueger:
repo with username and password

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/940369

Password is not saved because of security issues. We can't just store
the hash we only can store it as text which is a security risk.

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
@@ -1402,25 +1402,30 @@ namespace MonoDevelop.VersionControl.Git
 				monitor.BeginTask ("Cloning...", 2);
 
 				RunOperation (() => RetryUntilSuccess (monitor, credType => {
-					RootPath = LibGit2Sharp.Repository.Clone (Url, targetLocalPath, new CloneOptions {
-						CredentialsProvider = (url, userFromUrl, types) => {
-							transferProgress = checkoutProgress = 0;
-							return GitCredentials.TryGet (url, userFromUrl, types, credType);
-						},
-						RepositoryOperationStarting = ctx => {
-							Runtime.RunInMainThread (() => {
-								monitor.Log.WriteLine ("Checking out repository at '{0}'", ctx.RepositoryPath);
-							});
-							return true;
-						},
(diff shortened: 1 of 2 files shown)


mkrueger:
The second part may be fixed by:
https://github.com/mono/monodevelop/pull/8155

I don't know what the check out hang could cause so I added some logging for it.

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
```diff
@@ -1402,25 +1402,29 @@ protected override void OnCheckout (FilePath targetLocalPath, Revision rev, bool
...
+							OnTransferProgress = (tp) => OnTransferProgress (tp, monitor, ref transferProgress),
+							OnCheckoutProgress = (path, completedSteps, totalSteps) => {
+								OnCheckoutProgress (completedSteps, totalSteps, monitor, ref checkoutProgress);
+								Runtime.RunInMainThread (() => {
+									monitor.Log.WriteLine ("Checking out file '{0}'", path);
+								});
+							},
+						});
+					} catch (Exception e) {
+						LoggingService.LogInternalError ("Error while cloning repository " + rev + " recuse: " + recurse, e);
```
Therzok: We will stop reporting to the user that checkouts failed here.

mkrueger: fixed

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
```diff
@@ -1402,25 +1402,30 @@ protected override void OnCheckout (FilePath targetLocalPath, Revision rev, bool
...
+						throw e;
```
Therzok: We need a VersionControlException to be thrown, otherwise the command handler won't report it.

mkrueger: I re-throw the exception here - regardless what it is. It can already be a VersionControlException.

PR #8166, opened 2019-07-10
Updated reference to xamarin/md-addins@e342963

xamarin-release-manager:
Updated reference to xamarin/md-addins@e342963


sevoku:
this is already done in https://github.com/mono/monodevelop/commit/31c1b9be0062d010dc2513af1b691e6540434e8f

PR #8171, opened 2019-07-10
[debugger] Use store instead of Model

sgmunn:
A bit of a long shot, but it should/could fix the NRE here. We fixed one possible NRE with DisplayValue but we see the same error since that fix. The offset is +0 so I'm thinking that it's early in the method and the only thing I can think of here is that Get_Model is not returning the store from Gtk. We can use store here in its place since they are (or should be) the same object.

Fixes VSTS 849589
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/849589

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValueTreeView.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValueTreeView.cs
@@ -1670,7 +1670,7 @@ namespace MonoDevelop.Debugger
 
 		bool ValidObjectForPreviewIcon (TreeIter it)
 		{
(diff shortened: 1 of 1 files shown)


sgmunn:
other offsets in the stack trace were present, but ¯\_(ツ)_/¯

The Model property calls into Gtk to get the value, so, yes, it is assigned in the constructor but it provide a code path where it could return null. I don't see anywhere else we can get an NRE in this method.

jstedfast:
I guess this change can't hurt anything...

sgmunn:
yeah :(

sgmunn:
what on earth is wrong with building this branch.

sgmunn:
good lord, now what

sgmunn:
unrelated test failure

slluis:

PR #8174, opened 2019-07-10
[dotnetcore] Uses DotNetCoreSdkPaths resolver instead of DotNetCoreSdk

jtorres:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/934237

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSystemInformation.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSystemInformation.cs
@@ -49,7 +49,7 @@ namespace MonoDevelop.DotNetCore
 				return GenerateDotNetCoreUpdateInfo ();
 			return null;
 		}
-
+		
 		UpdateInfo GenerateDotNetCoreUpdateInfo ()
 		{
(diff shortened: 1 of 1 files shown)


jtorres:
@mrward DotNetCoreSdk contains its own DotNetCoreSdkPath reference (for Solution level) so that the `DotNetCoreSdk.Version` is fed by DotNetCoreSdkPath and does not depend on the Solution as a result it will be the same DotNetCoreSdk.Version = DotNetCoreSdkPath.SdkVersions

mrward:
I thought the [DotNetCoreSdk.Update method](https://github.com/mono/monodevelop/blob/1914d97a4eb38446a5c3067ed04cb9e0d49fc5cf/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs#L54) was called by the [DotNetCoreProjectExtension](https://github.com/mono/monodevelop/blob/1914d97a4eb38446a5c3067ed04cb9e0d49fc5cf/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs#L298) which would change the versions shown in the About dialog based on the current solution?

jtorres:
Exactly, but Versions does not depend on the current solution, but on SDK Location preferences... and that's why they are sync'ed when calling Update (); in any case, if you want I can change it to use dotnetcoresdkpath everywhere to avoid confusions..

PR #8176, opened 2019-07-11
Added Telemetry Counter to detect slow update command operations

jsuarezruiz:
Added Telemetry Counter to detect slow update command operations

Issue VSTS #623224

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Services.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Services.cs
@@ -268,8 +268,8 @@ namespace MonoDevelop.Ide
 
 	class UpdateCommandInfoCounterMetadata : CounterMetadata
 	{
-		public object CommandId {
-			get => GetProperty<object> ("CommandId");
+		public string CommandId {
+			get => GetProperty<string> ();
 			set => SetProperty (value);
 		}
 	}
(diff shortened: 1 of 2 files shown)


sevoku:
To not loose the discussion here: https://github.com/mono/monodevelop/pull/8139/files#r300710641

@Therzok wrote:
> For example, let's say an update task takes 25ms to run and has a continuation which does does something on the main thread.
> 
> Due to this code change, the continuation might run after the actual run handler is called.

So, I think with the cancellation of the async update it should be fine then, right? And we increased the timeout to 250ms (maybe we should even use 500ms, should be still safe to avoid beachballs). The main problem here is that update handlers should not run anything on the main thread if called from the key handler, otherwise we have no way to interrupt Gtk key events (we can't switch the sync context here).

@Therzok do you have a better idea? Maybe we could add an additional param to update handlers indicating that it's a key-press? With that context the handlers could do alternative/different checks... not sure.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Services.cs
```diff
@@ -250,5 +252,13 @@ public CompositionLoadMetadata (Dictionary<string, long> timings)
 			set => SetProperty (value);
 		}
 	}
+
+	class UpdateCommandInfoCounterMetadata : CounterMetadata
+	{
+		public object CommandId {
```
Therzok: This should be a string, we don't support reporting arbitrary objects.

Therzok:
This LGTM.

sevoku:

PR #8183, opened 2019-07-11
[Mac] Check if the Url path is null

iainx:
a url's path property might be null if the referenced resource no longer exists

Fixes VSTS #940763


iainx:

iainx:

PR #8195, opened 2019-07-12
Fixes VSTS FeedbackTicket 943824: Diff Incorrect

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/943824

Fixes VSTS Bug 946944: [Feedback] Diff Incorrect

Diff View could be loaded before the text view had the text - that
caused that the "local" text was empty. The blame view had the same
issue (but crashed) - fixed that as well.

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/BlameCommand.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/BlameCommand.cs
@@ -26,6 +26,7 @@
 
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.VisualStudio.Text.Editor;
 using Mono.Addins;
 using MonoDevelop.Ide;
 using MonoDevelop.Ide.Gui;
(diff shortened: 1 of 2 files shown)


mkrueger:

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/BlameCommand.cs
```diff
@@ -52,7 +53,9 @@ public static async Task<bool> Show (VersionControlItemList items, bool test)
 			
 			foreach (var item in items) {
 				var document = await IdeApp.Workbench.OpenDocument (item.Path, item.ContainerProject, OpenDocumentOptions.Default | OpenDocumentOptions.OnlyInternalViewer);
-				document?.GetContent<VersionControlDocumentController> ()?.ShowBlameView ();
+				document.RunWhenContentAdded<ITextView> (tv => {
```
slluis: A null check on document is missing here.

An alternative solution using a different api:
```
if (document != null) {
    await document.GetContentWhenAvailable<ITextView> ();
    document.GetContent<VersionControlDocumentController> ()?.ShowBlameView ();
}
```

Therzok: That makes the whole method not return until the content is available. What happens if a content is not added?

Therzok: Yeah, if a content is never added, the Task will just stay queued forever. We have 2 problems:
1. A design issue in the document model, where we never cancel the registration on document dispose.
2. A contract change if we do that, because the old one didn't.

Therzok: Filed https://devdiv.visualstudio.com/DevDiv/_workitems/edit/946429

mkrueger: I think I prefer the RunWhenContentAdded API - Marius brought up a valid issue.

mkrueger:

PR #8197, opened 2019-07-12
Refactoring of ObjectValueTreeView

jstedfast:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/866545

This refactors ObjectValueTreeView, or more aptly creates a side-by-side replacement for the existing ObjectValueTreeView, so that we can have multiple implementations of the UI - one Gtk and one native cocoa. ObjectValueTreeView drives the locals pad, the watches pad, debugger tooltips, pinned debugger values and the properties view in the exception dialog.
The motivation for the refactoring is to allow native tooltips for the new text editor but we also need to continue providing a UI for other platforms so we needed to split some functionality out from the UI to a controller.

We created a side-by-side replacement because this was a large change with a component that we were not familiar with, it provided us the ability to switch to the "reference" implementation easily and allows us to incrementally migrate users to the new refactored base.

That said, the design principle is based on MVC. We have a controller that implements common functionality and interacts with a view that renders the debug objects. Some model classes to wrap and make some of the code more testable and some mock objects that we can use that have specific behaviors which are hard to reproduce when debugging normally - cases like when the debugger takes a while to load the locals values and returns the proxy object that it uses.


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValueTreeViewController.cs
```diff
@@ -70,6 +70,8 @@ public class ObjectValueTreeViewController
 		/// </summary>
 		readonly Dictionary<string, CheckpointState> oldValues = new Dictionary<string, CheckpointState> ();
 
+		public readonly CancellationTokenSource CancellationTokenSource = new CancellationTokenSource ();
```
Therzok: We shouldn't expose the CancellationTokenSource as API.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/LocalsPad.cs
```diff
@@ -44,8 +48,31 @@ void ReloadValues ()
...
+				xx.Add (new FakeIsImplicitNotSupportedObjectValueNode ());
+
+				xx.Add (new FakeEvaluatingGroupObjectValueNode (1));
+				xx.Add (new FakeEvaluatingGroupObjectValueNode (0));
+				xx.Add (new FakeEvaluatingGroupObjectValueNode (5));
+
+				xx.Add (new FakeEvaluatingObjectValueNode ());
+				xx.Add (new FakeEnumerableObjectValueNode (10));
+				xx.Add (new FakeEnumerableObjectValueNode (20));
+				xx.Add (new FakeEnumerableObjectValueNode (23));
```
slluis: Test code?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValueTreeViewModels.cs
```diff
@@ -0,0 +1,737 @@
...
+		/// <summary>
+		/// Replaces the given child node with a new set of nodes
+		/// </summary>
+		void ReplaceChildNode (IObjectValueNode node, IObjectValueNode [] newNodes);
+	}
+
+	/// <summary>
+	/// Base class for IObjectValueNode implementations
+	/// </summary>
+	public abstract class AbstractObjectValueNode : IObjectValueNode
```
slluis: This IObjectValueNode / AbstractObjectValueNode split looks redundant. AbstractObjectValueNode is the only implementation of IObjectValueNode, and all node classes inherit from AbstractObjectValueNode. We could rename AbstractObjectValueNode to ObjectValueNode and get rid of the interface.

slluis:
The approach looks good in general. Good job!

There is something in the controller/view pattern implementation that is not entirely correct though. The `ObjectValueTreeViewController` class defines a set methods that are invoked by the view when something happens (for example `ExpandNodeAsync`), and defines a set of events to which the view responds when the controller communicates something (for example `NodeExpanded`). It has to be the other way around. You need to define an interface for the view, which provides the api of everything the UI can do, and then the controller will *control* the view using that api. The view interface should define events that are raised when the user interacts with the UI, and the controller should subscribe and react to those. Then, implementing the Cocoa view will be just a matter of implementing the view interface.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValueTreeViewModels.cs
sgmunn: agreed. we started with the interface to provide the initial abstraction and as we went we found that we needed AbstractObjectValueNode so the interface did become redundant.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValueTreeViewController.cs
jstedfast: ok, it's internalized now.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValueTreeViewModels.cs
```diff
@@ -200,14 +102,42 @@ protected AbstractObjectValueNode (string name)
 			}
 		}
 
-		public IReadOnlyList<IObjectValueNode> Children => children;
-		public virtual bool IsExpanded { get; set; }
+		/// <summary>
+		/// Gets the collection of children that have been loaded from the debugger
+		/// </summary>
+		public IReadOnlyList<AbstractObjectValueNode> Children => children;
```
Therzok: What operations do we need on this list? Maybe ImmutableArray can be used?

jstedfast: What's the difference between these?

Therzok: Well, ImmutableArray means it won't mutate over time.

`IReadOnlyList<T>` only ensures that API users of `Children` cannot mutate it, but does not guarantee that the list will not mutate over time. (`List<T>` and `Collection<T>` are mutable and implement `IReadOnlyList<T>` see [this](
https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ireadonlylist-1?view=netframework-4.8)).

I guess this is alright, since `AbstractObjectValueNode` is mutable, but `List<T> children` is not exposed anywhere.

Therzok: So yeah, keep as-is.

jstedfast: Thanks for the clarification.

sgmunn:
@jstedfast the one thing we need that is not in this branch is a way to toggle this so that some users use the refactored Gtk version and some the old version. I am thinking we can enable this for users with a Microsoft account which we might have to enable from within md-addins.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebugValueWindow.cs
```diff
@@ -36,10 +36,13 @@ namespace MonoDevelop.Debugger
 {
 	class DebugValueWindow : PopoverWindow
 	{
-		public ObjectValueTreeView Tree { get; }
-		ScrolledWindow sw;
+		readonly ObjectValueTreeViewController controller;
```
Therzok: We should check whether we have any kind of ref cycle here. `readonly` means we can't set to null, thus we can have a UI hierarchy cycle like `Parent <-> Child`

sgmunn:
@jstedfast conflicts in DebuggerOptionsPanelWidget.cs

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ExceptionCaughtDialog.cs
```diff
@@ -620,51 +609,61 @@ void UpdateSelectedException (ExceptionInfo ex)
...
+			}
+
 			if (!ex.IsEvaluating && ex.Instance != null) {
 				var opts = DebuggingService.GetUserOptions ().EvaluationOptions.Clone ();
 				opts.FlattenHierarchy = true;
-				ExceptionValueTreeView.AddValues (ex.Instance.GetAllChildren (opts));
+
+				var values = ex.Instance.GetAllChildren (opts);
+
+				if (useNewTreeView) {
```
Therzok: Not a blocker, but I think we could've had a simple helper class that encapsulates the logic for `useNewTreeView`:

```csharp
class FeatureSwitchObjectValueTreeViewController : ObjectValueTreeViewController
{
    readonly ObjectValueTreeViewController controller;
    readonly ... // other fields here

    public FeatureSwitchObjectValueTreeViewController(...) {}
}
```

Would've simplified the implementation by not sprinkling these checks everywhere, but contained into this class

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/ObjectValueNode.cs
```diff
@@ -0,0 +1,257 @@
...
+		/// the debugger will not be re-queried
+		/// </summary>
+		public async Task<int> LoadChildrenAsync (CancellationToken cancellationToken)
+		{
+			if (!ChildrenLoaded) {
+				var loadedChildren = await OnLoadChildrenAsync (cancellationToken);
+				AddChildren (loadedChildren);
+				ChildrenLoaded = true;
+
+				return loadedChildren.Count ();
```
Therzok: Is this going to evaluate children twice?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/ObjectValueNode.cs
```diff
@@ -0,0 +1,257 @@
...
+		/// Asynchronously loads a range of at most count children for the node into Children.
+		/// Subsequent calls will load an additional count children into Children until all children
+		/// have been loaded.
+		/// The task will complete immediately if all the children have previously been loaded and
+		/// the debugger will not be re-queried. 
+		/// </summary>
+		public async Task<int> LoadChildrenAsync (int count, CancellationToken cancellationToken)
+		{
+			if (!ChildrenLoaded) {
+				var loadedChildren = await OnLoadChildrenAsync (children.Count, count, cancellationToken);
```
Therzok: Same as above

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/ObjectValueNodeEventArgs.cs
```diff
@@ -0,0 +1,92 @@
...
+{
+	public class ObjectValueNodeEventArgs : EventArgs
+	{
+		public ObjectValueNodeEventArgs (ObjectValueNode node)
+		{
+			Node = node;
+		}
+
+		public ObjectValueNode Node {
+			get; private set;
```
Therzok: Minor style nit:
```suggestion
			get;
```

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/ObjectValueNodeEventArgs.cs
```diff
@@ -0,0 +1,92 @@
...
+
+	public sealed class ObjectValueEditEventArgs : ObjectValueNodeEventArgs
+	{
+		public ObjectValueEditEventArgs (ObjectValueNode node, string newValue) : base (node)
+		{
+			NewValue = newValue;
+		}
+
+		public string NewValue {
+			get; private set;
```
Therzok: Minor style nit:
```suggestion
			get;
```

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/ObjectValueNodeEventArgs.cs
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	public sealed class ObjectValueDisplayEventArgs : ObjectValueNodeEventArgs
+	{
+		public ObjectValueDisplayEventArgs (ObjectValueNode node) : base (node)
+		{
+		}
+
+		public string DisplayValue {
+			get; private set;
```
Therzok: Always null?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/ObjectValueNodeEventArgs.cs
```diff
@@ -0,0 +1,92 @@
...
+
+	public sealed class ObjectValueExpressionEventArgs : ObjectValueNodeEventArgs
+	{
+		public ObjectValueExpressionEventArgs (ObjectValueNode node, string expression) : base(node)
+		{
+			Expression = expression;
+		}
+
+		public string Expression {
+			get; private set;
```
Therzok: Minor style nit:
```suggestion
			get;
```

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor.Cocoa/MonoDevelop.TextEditor.Cocoa.csproj
```diff
@@ -45,8 +45,6 @@
       <Private>False</Private>
     </Reference>
     <ProjectReference Include="$(MdAddinsDirectory)Xamarin.Designer\external\xwt\Xwt\Xwt.csproj">
-      <Project>{92494904-35FA-4DC9-BDE9-3A3E87AC49D3}</Project>
```
Therzok: I think this wasn't intended?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/ObjectValueTreeViewController.cs
```diff
@@ -0,0 +1,971 @@
...
+					if (evalGroupNode.IsEvaluatingGroup) {
+						var replacementNodes = evalGroupNode.GetEvaluationGroupReplacementNodes ();
+
+						foreach (var newNode in replacementNodes) {
+							RegisterNode (newNode);
+						}
+
+						OnEvaluationCompleted (sender as ObjectValueNode, replacementNodes);
+					} else {
+						OnEvaluationCompleted (sender as ObjectValueNode);
```
Therzok: This and line 814 seem like common code on both codepaths.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/ObjectValueTreeViewController.cs
```diff
@@ -0,0 +1,971 @@
...
+			}
+
+			return 0;
+		}
+		#endregion
+
+		#region Evaluation watches
+		/// <summary>
+		/// Registers the ValueChanged event for a node where IsEvaluating is true. If the node is not evaluating, and
+		/// sendImmediatelyIfNotEvaulating is true, then fire OnEvaluatingNodeValueChanged immediately
```
Therzok:
```suggestion
		/// sendImmediatelyIfNotEvaluating is true, then fire OnEvaluatingNodeValueChanged immediately 
```

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/ObjectValueNodeEventArgs.cs
sgmunn: actually, I think I left this class in by accident. There are a couple of irksome code locations where we call back to the controller to use common display value logic that needs to be ironed out but I wasn't really liking the approach so much. I don't really want to call back to the controller if we can avoid it but I we don't really want the nodes to know about the controller (to get to the debugger/frame) either so I was using callback events but I'm not excited by that too much either. for now I left it but planned to revisit.

sgmunn:
thanks @Therzok we can continue to tweak this in the second pass which is where we will be adding native UI and will most likely have to address some of the quirks in the initial refactoring.

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor.Cocoa/MonoDevelop.TextEditor.Cocoa.csproj
jstedfast: yea, not intended

jstedfast:
@Therzok ok, I think all of your issues have been fixed

jstedfast:
🥂

PR #8201, opened 2019-07-15
Support .NET Core 3.0 as the default

rodrmoya:


nosami (approved):

Therzok (approved):

Therzok (approved):

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkInstalledCondition.cs
```diff
@@ -66,6 +68,26 @@ public override bool Evaluate (NodeElement conditionNode)
...
+			string requiredRuntimeVersion = conditionNode.GetAttribute ("runtimeVersion");
+			if (string.IsNullOrEmpty (requiredRuntimeVersion))
+				return false;
+
+			requiredRuntimeVersion = requiredRuntimeVersion.Replace ("*", string.Empty);
+			DotNetCoreVersion requiredDotNetCoreVersion;
+			DotNetCoreVersion.TryParse (requiredRuntimeVersion, out requiredDotNetCoreVersion);
+
+			if (requiredDotNetCoreVersion == null)
+				return versions.Any (version => version.ToString ().StartsWith (requiredRuntimeVersion, StringComparison.OrdinalIgnoreCase));
```
Therzok: This looks wrong. Should we choose the latest non-preview?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkInstalledCondition.cs
```diff
@@ -66,6 +68,26 @@ public override bool Evaluate (NodeElement conditionNode)
...
+
+			return versions.Any (version => version.Major == requiredDotNetCoreVersion.Major && version.Minor == requiredDotNetCoreVersion.Minor);
```
Therzok:
```
requiredDotNetCoreVersion = 2.3.1
installedVersion1 = 2.3.0
installedVersion2 = 2.3.2
```

This will end up choosing the wrong version.

Therzok (dismissed):
Requesting changes on the method to choose the right version

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkInstalledCondition.cs
nosami: Shouldn't it be first version >= required?

mrward: Right now we do not use 3 digits for the requiredDotNetCoreVersion. Only use the first two. This is used for templates.

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "main/external/debugger-libs"]
 	path = main/external/debugger-libs
-	url = git://github.com/mono/debugger-libs.git
+	url = https://github.com/nosami/debugger-libs.git
```
Therzok: Submodule got retargeted by mistske

nosami: It's not a mistake. I can't push to mono/debugger-libs and we need to make the MD/md-addins/debugger-libs merges at the same time

PR #8203, opened 2019-07-16
ui: Remove trailing period from exception dialog

hbons:
Anything else that needs to happen here? Update the translatable string?


hbons:
ugh, line endings. :/

PR #8208, opened 2019-07-16
Fixes VSTS Bug 941853: Unknown icon shown in status bar when opening a

mkrueger:
file

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/941853

The BeginProgress API is the cause of that - IconID can beconverted to
a string. Other calls of that method are ok.


slluis:

PR #8213, opened 2019-07-16
[Mac] Fix NSImageCacheException exception when getting the app icon

sevoku:
SharedApplication.ApplicationIconImage is not meant to be used
to retrieve the actual bundle icon, but only to change the default icon.

Instead of SharedApplication.ApplicationIconImage get the icon
directly from the bundle by default and fall-back to BrandingService
if not running from an app bundle.

Fixes VSTS #905844

### DIFF
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -765,6 +765,22 @@ namespace MonoDevelop.MacIntegration
 		[DllImport ("/usr/lib/libobjc.dylib", EntryPoint = "objc_msgSend")]
 		public extern static IntPtr IntPtr_objc_msgSend_IntPtr (IntPtr receiver, IntPtr selector, IntPtr arg1);
 
+		private static NSImage applicationIcon;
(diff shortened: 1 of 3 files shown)


sevoku:

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -765,6 +765,22 @@ void GlobalSetup ()
...
+				if (applicationIcon == null) {
+					// use the bundle icon by default
+					// if run from a bundle, this will be the icon from the bundle icon file.
+					// if not run from a bundle, this will be the default file icon of the mono framework folder.
+					applicationIcon = NSWorkspace.SharedWorkspace.IconForFile (NSBundle.MainBundle.BundlePath);
+				}
+				return applicationIcon;
+			}
+			set {
+				applicationIcon = value;
```
Therzok: Should this set `NSApplication.SharedApplication.ApplicationIconImage`?

sevoku: Yeah, better logic: 091d8c8

sevoku:

PR #8219, opened 2019-07-16
[dotnetcore] Turns DotNetCoreNotInstalledDialog into a InfoBar

jtorres:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/697714

Screenshot 2019-07-16 16 19 18

Screenshot 2019-07-16 16 43 58

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreNotInstalledInfoBar.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreNotInstalledInfoBar.cs
@@ -1,5 +1,5 @@
 ﻿//
-// DotNetCoreNotInstalledDialog.cs
+// DotNetCoreNotInstalledInfoBar.cs
 //
 // Author:
 //       Matt Ward <matt.ward@xamarin.com>
@@ -26,70 +26,38 @@
 
 using MonoDevelop.Core;
 using MonoDevelop.Ide;
-using MonoDevelop.Ide.Gui;
-using System;
+using MonoDevelop.Ide.Gui.Components;
 
 namespace MonoDevelop.DotNetCore
 {
-	class DotNetCoreNotInstalledDialog : IDisposable
+	class DotNetCoreNotInstalledInfoBar
 	{
-		static readonly string defaultMessage = GettextCatalog.GetString (".NET Core SDK is not installed. This is required to build and run .NET Core projects.");
-
-		GenericMessage message;
-		AlertButton downloadButton;
 		string downloadUrl = DotNetCoreDownloadUrl.GetDotNetCoreDownloadUrl ();
 
-		public DotNetCoreNotInstalledDialog ()
+		public string Message { get; set; }
+		public bool IsUnsupportedVersion { get; set; }
+		public bool IsNetStandard { get; set; }
+		public DotNetCoreVersion RequiredDotNetCoreVersion { get; set; }
+		public string CurrentDotNetCorePath { get; set; }
+
+		public void Prompt ()
 		{
-			Build ();
-		}
-
-		void Build ()
-		{
-			message = new GenericMessage {
-				Text = defaultMessage,
-				DefaultButton = 1,
-				Icon = Stock.Information
-			};
-
-			downloadButton = new AlertButton (GettextCatalog.GetString ("Download .NET Core..."));
-			message.Buttons.Add (AlertButton.Cancel);
-			message.Buttons.Add (downloadButton);
-
-			message.AlertButtonClicked += AlertButtonClicked;
-		}
-
-		void AlertButtonClicked (object sender, AlertButtonEventArgs e)
-		{
-			if (e.Button == downloadButton)
-				IdeServices.DesktopService.ShowUrl (downloadUrl);
-		}
-
-		public void Dispose ()
-		{
-			message.AlertButtonClicked -= AlertButtonClicked;
-		}
-
-		public void Show ()
-		{
-			if (IsUnsupportedVersion || IsNetStandard) //for .net standard we'll show generic message
+			var items = new InfoBarItem [] {
+				new InfoBarItem (GettextCatalog.GetString ("Download .NET Core"), InfoBarItemKind.Button, DownloadButtonClicked, true)
+				};
+			
+			if (IsUnsupportedVersion || IsNetStandard || RequiredDotNetCoreVersion == null) //for .net standard we'll show generic message
 				Message = DotNetCoreSdk.GetNotSupportedVersionMessage ();
 			else {
 				Message = DotNetCoreSdk.GetNotSupportedVersionMessage (RequiredDotNetCoreVersion.OriginalString);
 				downloadUrl = DotNetCoreDownloadUrl.GetDotNetCoreDownloadUrl (RequiredDotNetCoreVersion);
 			}
 
-			MessageService.GenericAlert (message);
+			IdeApp.Workbench.ShowInfoBar (false, new InfoBarOptions (Message) {
+				Items = items
+			});
 		}
 
-		public string Message {
-			get { return message.Text; }
-			set { message.Text = value; }
-		}
-
-		public bool IsUnsupportedVersion { get; set; }
-		public bool IsNetStandard { get; set; }
-		public DotNetCoreVersion RequiredDotNetCoreVersion { get; set; }
-		public string CurrentDotNetCorePath { get; set; }
+		void DownloadButtonClicked () => IdeServices.DesktopService.ShowUrl (downloadUrl);
 	}
 }
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
@@ -184,10 +184,10 @@ namespace MonoDevelop.DotNetCore
 		{
 			string GetMessage (DotNetCoreVersion currentVersion)
 			{
-				return GettextCatalog.GetString ("NET Core {0}.{1} SDK version {2} is not compatible with this version of Visual Studio for Mac. Install the latest update to the .NET Core {0}.{1} SDK by visiting {3}.", currentVersion.Major, currentVersion.Minor, currentVersion.ToString (), DotNetCoreDownloadUrl.GetDotNetCoreDownloadUrl (currentVersion));
+				return GettextCatalog.GetString (".NET Core {0}.{1} SDK version {2} is not compatible with this version of Visual Studio for Mac. Install the latest update to the .NET Core {0}.{1} SDK by visiting {3}", currentVersion.Major, currentVersion.Minor, currentVersion.ToString (), DotNetCoreDownloadUrl.GetDotNetCoreDownloadUrl (currentVersion));
 			}
 
-			var installedVersion = Versions.OrderByDescending (x => x).FirstOrDefault ();
+			var installedVersion = Versions?.OrderByDescending (x => x).FirstOrDefault ();
 			if (installedVersion != null) {
 				if (installedVersion < DotNetCoreVersion.MinimumSupportedSdkVersion) {
 					return GetMessage (installedVersion);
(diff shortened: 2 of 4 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreNotInstalledInfoBar.cs
```diff
@@ -26,70 +26,38 @@
...
+		public string CurrentDotNetCorePath { get; set; }
 
-		public void Dispose ()
+		public void Prompt ()
 		{
-			message.AlertButtonClicked -= AlertButtonClicked;
-		}
+			var items = new InfoBarItem [] {
+				new InfoBarItem (GettextCatalog.GetString ("Download .NET Core..."), InfoBarItemKind.Hyperlink, DownloadButtonClicked, true)
+				};
```
mrward: Should be using the proper ellipsis character here not three dots or Vaclav will be sad 😄 

https://gist.github.com/vancura/27fad9793840ceabae14

Therzok: Did the analyzer not suggest this?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreNotInstalledInfoBar.cs
```diff
@@ -26,70 +26,38 @@
...
-
-			message.AlertButtonClicked += AlertButtonClicked;
-		}
-
-		void AlertButtonClicked (object sender, AlertButtonEventArgs e)
-		{
-			if (e.Button == downloadButton)
-				IdeServices.DesktopService.ShowUrl (downloadUrl);
-		}
+		public string Message { get; set; }
```
Therzok: Should these be publicly settable?

Therzok: I'm wondering if setting them at constructor time or via Prompt would be a better option.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
```diff
@@ -184,7 +184,7 @@ internal static string GetNotSupportedVersionMessage (string version = "")
 		{
 			string GetMessage (DotNetCoreVersion currentVersion)
 			{
-				return GettextCatalog.GetString ("NET Core {0}.{1} SDK version {2} is not compatible with this version of Visual Studio for Mac. Install the latest update to the .NET Core {0}.{1} SDK by visiting {3}.", currentVersion.Major, currentVersion.Minor, currentVersion.ToString (), DotNetCoreDownloadUrl.GetDotNetCoreDownloadUrl (currentVersion));
+				return GettextCatalog.GetString (" .NET Core {0}.{1} SDK version {2} is not compatible with this version of Visual Studio for Mac. Install the latest update to the .NET Core {0}.{1} SDK by visiting {3}. ", currentVersion.Major, currentVersion.Minor, currentVersion.ToString (), DotNetCoreDownloadUrl.GetDotNetCoreDownloadUrl (currentVersion));
```
Therzok: Why a space in front?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
```diff
@@ -198,7 +198,7 @@ string GetMessage (DotNetCoreVersion currentVersion)
 				}
 			}
 
-			return GettextCatalog.GetString (".NET Core {0} SDK is required to build this application, and is not installed. Install the latest update to the .NET Core {0} SDK by visiting {1}.", version, DotNetCoreDownloadUrl.GetDotNetCoreDownloadUrl (version));
+			return GettextCatalog.GetString (" .NET Core {0} SDK is required to build this application, and is not installed. Install the latest update to the .NET Core {0} SDK by visiting {1}. ", version, DotNetCoreDownloadUrl.GetDotNetCoreDownloadUrl (version));
```
Therzok: Space in front here too.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
jtorres: yes, there is a bug in the InfoBar when drawing so that it arranges the whole bar to the left depending on the string -IIRC it depends on the `.`the string contains-.  By adding a white space at the beginning or a `.` at the end sorts out it. There are a reported bug about that. I forgot to mention it in there.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
jtorres: ditto

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
Therzok: Can we add a comment in the code so we avoid someone removing the space in the future?

mrward:
Is there any pattern we should be following here? iOS seems to use buttons but here we have a hyperlink //cc @vancura 

iOS

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreNotInstalledInfoBar.cs
slluis: I don't think we need the ellipsis.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
slluis: Also please prepend the space to the result of GetString call, that is:

`return " " + GettextCatalog.GetString (".NET Core ...`

mrward:
Is it possible to make the url in the info bar a clickable hyperlink?

Installed

 If this is not possible then I think we should remove it from the infobar but keep it in the other locations (solution pad and build errors)

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdk.cs
jtorres: the issue is here: https://github.com/mono/monodevelop/blob/fbf698b7ec05d906ef158396cd80892e19f7d26b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/InfoBar.cs#L136

will try to fix it, and forget adding or removing whitespaces...

mrward:
Not seeing any change which would fix the null reference when trying to run a .NET Core project. The RequiredDotNetCoreVersion is currently null when you try to run a project and the prompt is displayed.

vancura:
We don't have any pattern nor there is anything on Mac and HIG I believe. But we already have a [design and specification](https://mseng.visualstudio.com/_git/DevDivDesign?path=%2Fspec%2Fvsmac-editor-toolbars%2FREADME.md&version=GBmaster&createIfNew=true&anchor=button&_a=preview).

A couple of questions:

1. Shouldn't we use native controls for all new UI? These buttons look like Gtk, and it would be much better to use AppKit ones.

2. Please make "Download Xcode 10.2 now" default (blue).

jtorres:
@vancura what you mean by "Download Xcode 10.2 now"? which button?

vancura:
@jtorres I commented on https://github.com/mono/monodevelop/pull/8219#issuecomment-512169433

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/InfoBar.cs
```diff
@@ -104,6 +102,9 @@ public XwtInfoBar (string description, Action onDispose, params InfoBarItem[] it
 					mainBox.PackStart (toAdd);
 			}
 
+			//sets it as expanded horizontally, OnBoundChanged we determine other properties
+			descriptionLabel.ExpandHorizontal = true;
```
sevoku: Changing the expansion behaviour here is wrong. The original behaviour is not ideal (in fact the label should not expand at all, but that's another issue... the bounding box should expand rather than the label, I'll review that code), but this change enforces expansion, always adding a huge space between the text and the first Hyperlink item, which is actually supposed to be "connected" to the text with the [En-dash](https://gist.github.com/vancura/27fad9793840ceabae14#-en-dash): _"some long message text **–** [Link with more details](url)"_

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreNotInstalledInfoBar.cs
```diff
@@ -26,70 +26,38 @@
...
+		public bool IsNetStandard { get; set; }
+		public DotNetCoreVersion RequiredDotNetCoreVersion { get; set; }
+		public string CurrentDotNetCorePath { get; set; }
 
-		public void Show ()
+		public void Prompt ()
 		{
-			if (IsUnsupportedVersion || IsNetStandard) //for .net standard we'll show generic message
+			var items = new InfoBarItem [] {
+				new InfoBarItem (GettextCatalog.GetString ("Download .NET Core"), InfoBarItemKind.Hyperlink, DownloadButtonClicked, true)
```
sevoku: Use a button type here:
```suggestion
				new InfoBarItem (GettextCatalog.GetString ("Download .NET Core"), InfoBarItemKind.Button, DownloadButtonClicked, true)
```

sevoku: Idiomatically a button is an action and a hyperlink gives the user more information/context. "_Download .NET Core_" is definitely a call to "action", hence it should be a button (even tho it opens a browser with the download page, where the user can download the package).

Whereas something like "How to Install .NET Core" might be a hyperlink, indicating that the user can get more information/details about the problem identified by the infobar and optionally download .NET Core.

So if it really needs to be a hyperlink, rephrase the text.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/InfoBar.cs
jtorres: setting `ExpandHorizontal` on `OnBoundChanged` creates unexpected behaviour because of it modifies `descriptionLabel.Size.Width` value,, moving it to the constructor, prevents that and leaves IIRC the dash at the right,, will double check though

sevoku: As I wrote, it should not expand at all, to properly align the text with/behind the en-dash. I'll review that...

sevoku: What I mean is that it should be like this:

`[Text] – [Hyperlink] {    e x p a n s i o n    } [other][buttons][X]`

but right now it's this:

`[Text] {    e x p a n s i o n    } – [Hyperlink][X]`

or this:

`[Text] {    e x p a n s i o n    } – [Hyperlink] {    e x p a n s i o n    } [other][buttons][X]` 

and sometime even this:

`[Text] – [Hyperlink][other][buttons] {    e x p a n s i o n    } [X]`

vancura: Sorry about the confusion, it's totally right this button cannot be blue (default). Thanks @sevoku for the clarification.

jtorres:
rebase once this #8260 is merged

jtorres:
rebased, ready to merge after all green

PR #8222, opened 2019-07-17
Make `TypeSystemService.GetCodeAnalysisProjectAsync` return actual value when solution is still loading

DavidKarlas:
Main difference between `TypeSystemService.GetCodeAnalysisProjectAsync` and `TypeSystemService.GetCodeAnalysisProject` should be that `Async` version can be called before project/solution is fully loaded and `await` for it to be loaded and then return correct result, `non-Async` version should return project if currently or else `null`.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService_WorkspaceHandling.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService_WorkspaceHandling.cs
@@ -431,18 +431,22 @@ namespace MonoDevelop.Ide.TypeSystem
 			var parentSolution = project.ParentSolution;
 			var workspace = await GetWorkspaceAsync (parentSolution, cancellationToken);
 			var projectId = workspace.GetProjectId (project);
-			if (projectId == null)
-				return null;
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService_WorkspaceHandling.cs
```diff
@@ -431,18 +431,22 @@ public async Task<Microsoft.CodeAnalysis.Project> GetCodeAnalysisProjectAsync (M
 			var parentSolution = project.ParentSolution;
 			var workspace = await GetWorkspaceAsync (parentSolution, cancellationToken);
 			var projectId = workspace.GetProjectId (project);
-			if (projectId == null)
-				return null;
-			var proj = workspace.CurrentSolution.GetProject (projectId);
-			if (proj != null)
-				return proj;
+			Microsoft.CodeAnalysis.Project proj = null;
+			if (projectId == null) {
```
slluis: shouldn't it be `projectId != null`?

DavidKarlas: 🤦‍♂ yes it should, updated

PR #8224, opened 2019-07-17
[WIP] File service pre-filters paths before notification

Therzok:
Attempt #1 at fixing filtering of notifications when using our big cone file system watcher.

See API usage in test files.

Fixes VSTS #643326 - PathTree should be able to remove a node directly
Fixes VSTS #947787 - Design API for consuming filtered FileService+FileWatcher notifications

This is a WIP, as creating too many FileSystemWatchers is another bottleneck we can hit, as corefx will do a lot of work to filter and map the events, so handling this with one watcher on our side might be more efficient.

TODO items:
* [x] ~934618 - Possible optimization of FileSystemWatcher by using 2 watchers per actual watched path~
* [ ] 947789 - Use filtered FileService+FileWatcher APIs in slow handlers
* [ ] 947788 - Implement API design for FileService+FileWatcher filtered notifications


Therzok:
On the test scenario I had, because the files were created in a folder that's not under any project:

FileCreated
```
MonoDevelop.Core.FileService+WatchingHandler
00:00:00.9028459 - MonoDevelop.Core.FileService+WatchingHandler Void NotifyIfNeeded(MonoDevelop.Core.FileEventArgs)
```

Versus the old one:
```
00:02:34.7219796 - MonoDevelop.Projects.Project Void OnFileCreated(System.Object, MonoDevelop.Core.FileEventArgs)
```

main/src/core/MonoDevelop.Core/MonoDevelop.Core/FileService.WatchingHandler.cs
```diff
@@ -0,0 +1,75 @@
...
+	public static partial class FileService
+	{
+		public sealed class WatchingHandler
+		{
+			readonly PathTree watchingTree = new PathTree ();
+			readonly ReaderWriterLockSlim readerWriterLock = new ReaderWriterLockSlim (LockRecursionPolicy.NoRecursion);
+
+			internal WatchingHandler () {}
+
+			public IDisposable WatchForFilesCreated (FilePath path, Action<FileEventArgs> handler)
```
slluis: Since FileService can send file change notifications in batches, this api should also support batches. That is, the handler should be `Action<FileEventArgs[]> handler`.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/FileService.WatchingHandler.cs
```diff
@@ -0,0 +1,75 @@
...
+			{
+				// 2 options:
+				// * one tree per notification kind - more memory but fewer traversal iterations when checking whether to notify
+				// * use the same tree, register the notification kind in the registration/id.
+				return new WatchingRegistration (watchingTree, path, handler, readerWriterLock);
+			}
+
+			internal void NotifyIfNeeded (FileEventArgs args)
```
slluis: What will call this?

main/src/core/MonoDevelop.Core/MonoDevelop.Core/FileService.WatchingHandler.cs
Therzok: FileEventArgs is an EventArgsChain

main/src/core/MonoDevelop.Core/MonoDevelop.Core/FileService.WatchingHandler.cs
Therzok: The FileSystemWatcher itself.

Therzok:
After discussing with @mhutch , FileService will have a new API that would define a watcher registration. The end goal is to limit the files we're monitoring.

Thus, we will have 2 kinds of watcher registration:
a) File registration, where we monitor project.Files, for example. That means we can then assign directoriees to watch, extensions, etc, allowing us to trim the total number of files watched
b) Directory registration - i.e. globs

File registration will be something like: WatchFiles (files). That way, we can have a simple hashset of files to validate against, and if the file is included, notify the file registrations.

Directory registration will be something like: WatchDirectory (path, includeGlobs, excludeGlobs).

If no includeGlobs are passed or the includeGlobs are just files in a directory, the directory will be watched non-recursively, with the given filter - i.e. extension.

If the includeGlobs are recursive, then we register a FileSystemWatcher that watches recursively, applying the glob information to filter notifications.

The notification algorithm would first check the file set and dispatch a notification to the file handlers if the file is included.

Then it would try to get the PathTree registration for a directory being watched, then traverse registered nodes towards he root and dispatch notifications.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/FileWatcherInformation_V2.cs
```diff
@@ -0,0 +1,151 @@
...
+// THE SOFTWARE.
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using MonoDevelop.FSW;
+using Roslyn.Utilities;
+
+namespace MonoDevelop.Core
+{
+	internal sealed class SingleFileWatcher : IDisposable
```
Therzok: This will be split out for the second phase of the PR, and it will be included as part of FileWatcherService.

main/src/core/MonoDevelop.Core/MonoDevelop.Core/FileWatcherInformation.cs
```diff
@@ -0,0 +1,123 @@
...
+// THE SOFTWARE.
+using System;
+using System.Diagnostics;
+using System.Threading;
+using Microsoft.Extensions.ObjectPool;
+using MonoDevelop.FSW;
+
+namespace MonoDevelop.Core
+{
+	sealed class FileWatcherInformation
```
Therzok: TODO: add doc comments about the purpose of this class.

It should abstract out each file change kind and handle pre-filtering of notifications, if any.

Therzok:
Superseded

PR #8225, opened 2019-07-18
Fixed Bug 820341: [Feedback] Unable to print code

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/820341

The only way to ensure that the theme can be printed in a good quality
is to use the default light theme for printing. The same with the font
size - it used to be the editor zoom factor but setting it to a
constant zoom facotr gives the print code the control over the size.
Fonts currently are way bigger than they should be in printing.


mkrueger:

PR #8231, opened 2019-07-19
Fixes VSTS Bug 948635: [Feedback] Tab doesn't work in text files

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/948635

EditorExtension no longer works on that level - it's moved to IDE
TextEditor for handling. Using the TextEditor from the document is
safer.


mkrueger:

PR #8235, opened 2019-07-19
[IDE] Handle bindings returning null commands

iainx:
KeyBindingManager.Commands can return null, so handle a null return. Wrap the
call to ProcessKeyEventCore in a try catch block to make sure we handle and log
any other possible problems with it.

Fixes VSTS #944669

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -478,8 +478,14 @@ namespace MonoDevelop.Components.Commands
 			}
 #endif
 			// Handle the GDK key via MD commanding
-			if (ProcessKeyEventCore (ev))
-				return true;
+			try {
+				if (ProcessKeyEventCore (ev)) {
+					return true;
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -478,8 +478,14 @@ internal bool ProcessKeyEvent (Gdk.EventKey ev)
...
 			// Handle the GDK key via MD commanding
-			if (ProcessKeyEventCore (ev))
-				return true;
+			try {
+				if (ProcessKeyEventCore (ev)) {
+					return true;
+				}
+			} catch (Exception ex) {
+				LoggingService.LogError ("Exception while parsing command", ex);
+				return false;
```
mrward: LogInternalError? Otherwise I don't think this will go to telemetry.

slluis: @iainx can you please fix?

sevoku:

sevoku:

PR #8237, opened 2019-07-19
[mdtool] Call terminate when ending

iainx:
Fixes VSTS #725899

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/GtkWorkarounds.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/GtkWorkarounds.cs
@@ -45,19 +45,9 @@ using System.Windows.Input;
 
 namespace MonoDevelop.Components
 {
-	public static class GtkWorkarounds
+	public static partial class GtkWorkarounds
 	{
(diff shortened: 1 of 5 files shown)


slluis:
Adding MonoDevelop.Ide as reference means loading a ton of unnecessary code. Can we include GtkWorkarounds by referencing the source file?

sevoku:
Including GtkWorkarounds.cs would pull other dependencies in (such as Gtk). I'd rather just add the termination code here like:

```csharp

const string LIBOBJC ="/usr/lib/libobjc.dylib";

[DllImport (LIBOBJC, EntryPoint = "objc_msgSend")]
static extern void objc_msgSend_NSInt64_NSInt32 (IntPtr klass, IntPtr selector, int arg);

static void MacTerminate ()
{
	objc_msgSend_NSInt64_NSInt32 (sharedApp, sel_terminate, 0);
}

static void Shutdown (bool terminate)
{
	Runtime.Shutdown ();

	if (terminate && Runtime.IsMac) {
		MacTerminate ();
	}
}
```

sevoku:
@iainx the build failed, no sure if that's related to the change or just needs a rebase.

iainx:
@sevoku yeah, the code wasn't falling through correctly anymore, fixed and rebased.

PR #8238, opened 2019-07-19
[IDE] Check the configuration merger exists

iainx:
Fixes VSTS #944654

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/MainToolbarController.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/MainToolbarController.cs
@@ -279,8 +279,14 @@ namespace MonoDevelop.Components.MainToolbar
 		void FillRuntimesForProject (List<RuntimeModel> list, SolutionItem project, ref int runtimes)
 		{
 			ExecutionTarget previous = null;
+			ConfigurationMerger merger;
 
-			foreach (var target in configurationMergers [project].GetTargetsForConfiguration (IdeApp.Workspace.ActiveConfigurationId, configurationMergers.Count < 2)) {
+			if (!configurationMergers.TryGetValue (project, out merger)) {
+				LoggingService.LogWarning ($"No configuration merger found for {project.Name}");
+				return;
+			}
+
+			foreach (var target in merger.GetTargetsForConfiguration (IdeApp.Workspace.ActiveConfigurationId, configurationMergers.Count < 2)) {
 				if (target is ExecutionTargetGroup) {
 					var devices = (ExecutionTargetGroup)target;
 


iainx:
Like Sevo says in VSTS, not sure if this is a problem with the project system and there should be a `ConfigurationMerger` for every project, or if the shell should be more careful. This is the shell being more careful

Therzok:
Agreed with @slluis. Just checking that there could end up in broken behaviour, like resetting the user's execution target without any notification.

Based on a few searches, there's a few places where the update code may be called on a background thread.

* [iOS load](https://github.com/xamarin/vsmac-xamarin-extensions/blob/33ef4d23961f092466828a5c54f70fe26b679539/MonoDevelop.IPhone/MonoDevelop.IPhone/Project/IPhoneProject.cs#L328) via [ios devicewatcher](https://github.com/xamarin/vsmac-xamarin-extensions/blob/33ef4d23961f092466828a5c54f70fe26b679539/MonoDevelop.IPhone/MonoDevelop.IPhone/Project/IPhoneProject.cs#L328)
* [iOS plist cache refresh](https://github.com/xamarin/vsmac-xamarin-extensions/blob/33ef4d23961f092466828a5c54f70fe26b679539/MonoDevelop.IPhone/MonoDevelop.IPhone/Project/IPhoneProject.cs#L1705) via [file removed from project notification](https://github.com/xamarin/vsmac-xamarin-extensions/blob/e6bc43db787c6713d96c1648abdebb9cfbc09e70/MonoDevelop.MacDev/MonoDevelop.MacDev/MacProject.cs#L563) as the latter uses AsyncEvents compared to Changed which uses FileService (cc @mrward)

* Unity also might be calling this on a background thread when debugging cc @jbevain 

The workload tags suggest that this is happening in aspnetcore scenarios too, which puzzles me.

Therzok:
We should have a Runtime.AssertMainThread check in there.

Therzok:
Probably related, I can trigger the following log message with this repro:
```
ERROR [2019-07-23 11:14:59Z]: No runtimeModel for dncapp2
```

1. Create a solution with 2 startup projects
2. Create a run config that contains both of them
3. Unload one project
4. Switch to that run config
5. Observe above error

slluis:
> @slluis since this fixes a crash, shouldn't we merge this and keep the bug open (moving it to the project system area) for further investigation?

We don't have a guarantee this fixes anything. There is something inconsistent in the configuration selector, so it may be crashing in many different ways. Let's do the investigation and fix it for real.

sevoku:
Although this change would fix the crasher, it'll still hide the underlying issue. Since we didn't get any new reports about this, it might have been fixed as part of some other change in the project system. Let's close this until we get more reports and we can reopen and dig deeper then.

PR #8239, opened 2019-07-19
Sourcelink

nosami:
Bug fixes:

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/957623

Fixes VSTS 695466


nosami:
depends on https://github.com/mono/debugger-libs/pull/235

nosami:
NetCoreBreakpointsAndSteppingTests tests failing

slluis (changes requested):
Test scenarios and use cases need to be [documented](https://github.com/xamarin/md-addins/tree/master/test-scenarios).

nosami:
NetCoreBreakpointsAndSteppingTests tests fixed

mrward (approved):

mrward:
There are some problems with the UI thread hanging but this looks like a problem in the text editor.

jstedfast (approved):

nosami:
Editor UI hang has gone since rebasing on master

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggerOptionsPanelWidget.cs
```diff
@@ -157,7 +190,8 @@ public void Store ()
 			ops.EvaluationTimeout = (int)spinTimeout.Value;
 
 			options.StepOverPropertiesAndOperators = checkStepOverPropertiesAndOperators.Active;
-			options.ProjectAssembliesOnly = checkProjectCodeOnly.Active;
+			options.ProjectAssembliesOnly = !checkProjectCodeOnly.Active;
```
Therzok: Hmmm, why was this inverted?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggerOptionsPanelWidget.cs
```diff
@@ -133,7 +163,9 @@ public DebuggerOptionsPanelWidget ()
 			Build ();
 
 			options = DebuggingService.GetUserOptions ();
-			checkProjectCodeOnly.Active = options.ProjectAssembliesOnly;
+			checkProjectCodeOnly.Active = !options.ProjectAssembliesOnly;
```
Therzok: same here.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggerOptionsPanelWidget.cs
```diff
@@ -67,10 +70,39 @@ public class DebuggerOptionsPanelWidget : VBox
 
 		void Build ()
 		{
-			checkProjectCodeOnly = new CheckBox (GettextCatalog.GetString ("Enable Just My Code"));
-			PackStart (checkProjectCodeOnly);
+			PackStart (new Label { Markup = "<b>" + GettextCatalog.GetString ("Scope") + "</b>" });
 			checkStepOverPropertiesAndOperators = new CheckBox (GettextCatalog.GetString ("Step over properties and operators"));
 			PackStart (checkStepOverPropertiesAndOperators);
+			checkProjectCodeOnly = new CheckBox (GettextCatalog.GetString ("Step into external code")) {
```
Therzok: The name of the checkbox should change too. This would reflect why we inverted the values for users.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Initializer.cs
```diff
@@ -41,14 +45,16 @@ public class Initializer: CommandHandler
...
+		string symbolCachePath;
+
 		protected override void Run ()
 		{
 			DebuggingService.CallStackChanged += OnStackChanged;
 			DebuggingService.CurrentFrameChanged += OnFrameChanged;
 			DebuggingService.DisassemblyRequested += OnShowDisassembly;
 
 			IdeApp.CommandService.RegisterGlobalHandler (new GlobalRunMethodHandler ());
+			symbolCachePath = UserProfile.Current.CacheDir.Combine ("Symbols");
```
Therzok: I'm wondering if this should be user configurable. The VSMac cache folder could get pretty big, as we do no invalidation here based on size.

Therzok: I'm wondering whether we should dump the symbol server information in the `.vs` directory and maintain a mapping in the cache folder if we want to de-duplicate symbol server downloads?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Initializer.cs
```diff
@@ -63,28 +69,73 @@ async void OnFrameChanged (object s, EventArgs a)
...
+					// ~/Library/Caches/VisualStudio/8.0/Symbols/org/projectname/git-sha/path/to/file.cs
+					if (!File.Exists (downloadLocation)) {
+						if (automaticSourceDownload == AutomaticSourceDownload.Always) {
+							doc = await DownloadAndOpenFileAsync (frame, line, sourceLink);
+						} else {
+							var hyperlink = $"<a href='{ sourceLink.Uri }'>{  Path.GetFileName (sourceLink.RelativeFilePath) }</a>";
+							var stackframeText = $"<b>{frame.FullStackframeText}</b>";
+
+							var text = GettextCatalog.GetString
+								("{0} is a call to external source code. Would you like to get {1} and view it?", stackframeText, hyperlink);
```
Therzok:
```suggestion
								("{0} is a call to external source code. Would you like to get '{1}' and view it?", stackframeText, hyperlink);
```

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Initializer.cs
```diff
@@ -119,7 +170,36 @@ async void OnFrameChanged (object s, EventArgs a)
 				disassemblyDoc.Select ();
 			}
 		}
-		
+
+		async System.Threading.Tasks.Task<Document> DownloadAndOpenFileAsync (StackFrame frame, int line, SourceLink sourceLink)
+		{
+			var pm = IdeApp.Workbench.ProgressMonitors.GetStatusProgressMonitor (
+				GettextCatalog.GetString ("Downloading ") + sourceLink.Uri,
```
Therzok:
```suggestion
				GettextCatalog.GetString ("Downloading {0}", sourceLink.Uri),
```

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Initializer.cs
```diff
@@ -119,7 +170,36 @@ async void OnFrameChanged (object s, EventArgs a)
...
+				Stock.StatusWorking,
```
Therzok:
```suggestion
				Stock.StatusDownload,
```

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Initializer.cs
```diff
@@ -119,7 +170,36 @@ async void OnFrameChanged (object s, EventArgs a)
...
+				true
+			);
+
+			Document doc = null;
+			try {
+				var downloadLocation = sourceLink.GetDownloadLocation (symbolCachePath);
+				Directory.GetParent (downloadLocation).Create ();
```
Therzok:
```suggestion
				Directory.CreateDirectory (Path.GetDirectoryName (downloadLocation));
```

Avoids a redundant DirectoryInfo allocation.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Initializer.cs
```diff
@@ -119,7 +170,36 @@ async void OnFrameChanged (object s, EventArgs a)
...
+				DocumentRegistry.SkipNextChange (downloadLocation);
+				var client = HttpClientProvider.CreateHttpClient (sourceLink.Uri);
+				using (var stream = await client.GetStreamAsync (sourceLink.Uri)) {
```
Therzok: Do we need UpdateSourceFile to run on the UI thread? Otherwise, I think we can safely do:
```csharp
using (var stream = await client.GetStreamAsync (sourceLink.Uri).ConfigureAwait (false))
using (var fs = new FileStream (downloadLocation, FileMode.CreateNew)) {
	await stream.CopyToAsync (fs).ConfigureAwait (false);
}
frame.UpdateSourceFile (downloadLocation);
doc = await Runtime.RunInMainThread (() => IdeApp.Workbench.OpenDocument (downloadLocation, null, line, 1, OpenDocumentOptions.Debugger));
```

Therzok: We should also pass `FileMode.Create`, as we could have a broken file hash if a download failure occurred, so we want to truncate and update the current file.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Initializer.cs
```diff
@@ -119,7 +170,36 @@ async void OnFrameChanged (object s, EventArgs a)
...
+					using (var fs = new FileStream (downloadLocation, FileMode.CreateNew)) {
+						await stream.CopyToAsync (fs);
+					}
+				}
+				frame.UpdateSourceFile (downloadLocation);
+				doc = await IdeApp.Workbench.OpenDocument (downloadLocation, null, line, 1, OpenDocumentOptions.Debugger);
+			} catch (Exception ex) {
```
Therzok: We should also do something to notify the user if a download fails or there's no connectivity?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Initializer.cs
```diff
@@ -119,7 +170,36 @@ async void OnFrameChanged (object s, EventArgs a)
 				disassemblyDoc.Select ();
 			}
 		}
-		
+
+		async System.Threading.Tasks.Task<Document> DownloadAndOpenFileAsync (StackFrame frame, int line, SourceLink sourceLink)
```
Therzok: $0.02 nitpick: This method should probably receive a monitor and pass the filepath, so we can have it free-threaded on a background thread.

Then, the caller would call `OpenDocument` and handle being on the UI thread.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DocumentRegistry.cs
```diff
@@ -137,7 +130,7 @@ public static void Remove (Document sourceEditorView)
 		public static void IgnoreAllChangedFiles ()
 		{
 			foreach (var view in GetAllChangedFiles ()) {
-				view.LastSaveTimeUtc = File.GetLastWriteTime (view.Document.FileName);
+				view.LastSaveTimeUtc = File.GetLastWriteTimeUtc (view.Document.FileName);
```
Therzok: Nice catch!

Therzok (commented):
Done with first review pass

Therzok:
Done with review pass 1.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Initializer.cs
nosami: > I'm wondering if this should be user configurable. The VSMac cache folder could get pretty big, as we do no invalidation here based on size.

On VS Windows, the cache path is configurable. I think @hbons was wanting to minimize the amount of configuration required here.

> I'm wondering whether we should dump the symbol server information in the .vs directory and have no problems with a global cache?

Sharing symbols seems like a good idea to me - less to download when symbols are shared between solutions (probably common for things like runtime code or Newtonsoft :) )

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggerOptionsPanelWidget.cs
nosami: I think I meant to do this but forgot :)

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Initializer.cs
nosami: UpdateSourceFile can run on any thread, I'll apply your suggestion

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Initializer.cs
```diff
@@ -63,28 +69,75 @@ async void OnFrameChanged (object s, EventArgs a)
...
+
 			int line = frame.SourceLocation.Line;
 			if (line != -1) {
-				if (!file.IsNullOrEmpty && System.IO.File.Exists (file)) {
+				if (!file.IsNullOrEmpty && File.Exists (file)) {
 					var doc = await IdeApp.Workbench.OpenDocument (file, null, line, 1, OpenDocumentOptions.Debugger);
 					if (doc != null)
 						return;
 				}
+				bool alternateLocationExists = false;
```
Therzok: We don't need this boolean, imo. From the BCL docs:

Create | 2 | Specifies that the operating system should create a new file. If the file already exists, it will be overwritten. This requires Write permission. FileMode.Create is equivalent to requesting that if the file does not exist, use CreateNew; otherwise, use Truncate. If the file already exists but is a hidden file, an UnauthorizedAccessException exception is thrown.

Therzok:
Took it for a spin, it works for me.

Minor nits, which I leave to the design folks to look at:
Screenshot 2019-08-07 at 14 05 42

I think we should disable interaction with the combobox when the external code check is untoggled.

Therzok:
cc @vancura @hbons

nosami:
> I think we should disable interaction with the combobox when the external code check is untoggled.

Done

togglecombo

Therzok (approved):

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "main/external/debugger-libs"]
 	path = main/external/debugger-libs
-	url = git://github.com/mono/debugger-libs.git
+	url = https://github.com/mono/debugger-libs.git
```
Therzok: Revert this change

Therzok (changes requested):
Revert the git -> https change in gitmodules and revert the vs-editor bump.

.gitmodules
nosami: done

Therzok (approved):

sgmunn:
as I understand it, the .net core tests are failing with a mono bug

PR #8253, opened 2019-07-21
[Core] Add some extensions to ReaderWriterLockSlim and showcase usage

Therzok:
Moves the assembly name cache to a RW lock, instead of a simple lock, allowing for concurrent reads from the dictionary when it has data

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/SystemAssemblyService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/SystemAssemblyService.cs
@@ -234,22 +234,26 @@ namespace MonoDevelop.Core.Assemblies
 			return aname;
 		}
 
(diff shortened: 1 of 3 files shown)


slluis (approved):

mrward (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/SystemAssemblyService.cs
```diff
@@ -234,22 +234,26 @@ public static AssemblyName ParseAssemblyName (string fullname)
...
 				if (assemblyNameCache.TryGetValue (file, out name))
 					return name;
 			}
 
 			try {
 				name = AssemblyName.GetAssemblyName (file);
-				lock (assemblyNameCache) {
-					assemblyNameCache [file] = name;
+
+				using (readerWriterLock.Write ()) {
```
Therzok: This should actually be an upgradeable read.

PR #8255, opened 2019-07-22
[A11y] Check selector before trying to call accessibility method

mrward:
Check accessibilityServesAsTitleForUIElements selector is available
for UI element when trying to remove the accessibility element
from the title.

Unhandled ObjC Exception
MonoDevelop.MacIntegration.MacPlatformService+MarshalledObjCException: NSInvalidArgumentException: -[ACAccessibilityElement accessibilityServesAsTitleForUIElements:]: unrecognized selector sent to instance

	  at <unknown> <0xffffffff>
	  at MonoDevelop.Components.Mac.Messaging:IntPtr_objc_msgSend <0x00111>
	  at MonoDevelop.Components.AtkCocoaHelper.AtkCocoaMacExtensions:RemoveElementFromTitle <0x0027a>
	  at MonoDevelop.Ide.WelcomePage.WelcomePageSection:RemoveAccessibiltyTitledWidget <0x0018a>
	  at MonoDevelop.Ide.WelcomePage.WelcomePageRecentProjectsList:RecentFilesChanged <0x00282>
	  at MonoDevelop.Ide.Desktop.RecentFileStorage:<OnRecentFilesChanged>b__31_0 <0x000fa>

A fix was added where we check when adding an accessibility element
for a title but not when removing.

Related commit: 96d23ffac8f497aad89ba21d67020d2eec9b2be4

Fixes VSTS #950645 - Unrecognised selector crash

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs
@@ -384,6 +384,12 @@ namespace MonoDevelop.Components.AtkCocoaHelper
 				return;
 			}
 
+			// bug #940756 suggests that some elements do not respond to this selector
+			// so check before sending so it doesn't throw an ObjC exception
+			if (!((NSObject)titleNsa).RespondsToSelector (selAccessibilityServesAsTitleForUIElements)) {
+				return;
+			}
+
 			IntPtr ptr = Messaging.IntPtr_objc_msgSend (titleNsa.Handle, selAccessibilityServesAsTitleForUIElements_Handle);
 			using (var array = Runtime.GetNSObject<NSArray> (ptr)) {
 				if (array == null)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs
```diff
@@ -384,6 +384,12 @@ public static void RemoveElementFromTitle (this Atk.Object title, Atk.Object o)
 				return;
 			}
 
+			// bug #940756 suggests that some elements do not respond to this selector
```
Therzok:
```suggestion
			// bug #950645 suggests that some elements do not respond to this selector
```

😄

mrward: I could remove the bug number completely 😄

iainx:
The C# API in theory allows any object to implement  INSAccessible, but in practice all the objects that implement it will be NSObject. We can add the checks if you like

PR #8256, opened 2019-07-22
[Projects] Avoid running out of file handles in EnumerateFiles

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/943580

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
@@ -924,7 +924,7 @@ namespace MonoDevelop.Projects.MSBuild
 				else if (!baseDir.EndsWith ("\\", StringComparison.Ordinal))
 					baseDir += '\\';
 				var recursiveDir = baseRecursiveDir.IsNullOrEmpty ? FilePath.Null : basePath.ToRelative (baseRecursiveDir);
-				res = res.Concat (Directory.EnumerateFiles (basePath, path).Select (f => func (f, baseDir + Path.GetFileName (f), recursiveDir)));
+				res = res.Concat (Directory.GetFiles (basePath, path).Select (f => func (f, baseDir + Path.GetFileName (f), recursiveDir)));
 			} else {
 				// Directory specifier
 				// Look for matching directories.


mrward (commented):
I thought Files.GetFiles was enough to fix the problem?

rodrmoya:
> I thought Files.GetFiles was enough to fix the problem?

yes, just changed it to just do that change

rodrmoya:

Therzok (approved):

mrward (approved):

slluis (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/DefaultMSBuildEngine.cs
```diff
@@ -924,7 +924,7 @@ static IEnumerable<T> ExpandWildcardFilePath<T> (MSBuildProject project, FilePat
 				else if (!baseDir.EndsWith ("\\", StringComparison.Ordinal))
 					baseDir += '\\';
 				var recursiveDir = baseRecursiveDir.IsNullOrEmpty ? FilePath.Null : basePath.ToRelative (baseRecursiveDir);
-				res = res.Concat (Directory.EnumerateFiles (basePath, path).Select (f => func (f, baseDir + Path.GetFileName (f), recursiveDir)));
+				res = res.Concat (Directory.GetFiles (basePath, path).Select (f => func (f, baseDir + Path.GetFileName (f), recursiveDir)));
```
Therzok: I was reading through the docs of GetFiles, and saw a peculiar note on the remarks section:

```
If the specified extension is exactly three characters long, the method returns files with extensions that begin with the specified extension. For example, "*.xls" returns both "book.xls" and "book.xlsx".
```

This may impact the glob include behaviour. Will open a bug for it, as EnumerateFiles has the same problem

https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.getfiles?view=netframework-4.8

Therzok: Added as a comment to https://devdiv.visualstudio.com/DevDiv/_workitems/edit/950486/

It should fix this issue.

mrward: Mono 6.0 does not seem to have this behaviour. Mono 6.0 seems to only return book.xls from Directory.GetFiles and Directory.EnumerateFiles. Not tried any other Mono versions.

I wonder if this is just for Windows.

mrward: On Windows this does happen for both GetFiles and EnumerateFiles. So I don't think anything is more broken than it was before here 😄

Therzok: Ah, good find. We need to find out if corefx will change this behaviour in the future.

PR #8259, opened 2019-07-23
[Core] Simplify Task usage in the Projects layer

Therzok:
Touches only a small part, but removes superflous Task.Runs sprinkled everywhere.


mrward (approved):

Therzok:
@slluis ping

slluis (approved):

PR #8260, opened 2019-07-23
[Ide] Fix InfoBar button alignments

sevoku:
* Align all buttons to the right side
* Align the first hyperlink item to the end of the text
* Fix jumping buttons on window resize

Screencast (this is just an example based on existing code):

InfoBar Bug Fixed

Fixes VSTS #948301


vancura:
Very nice! :)

PR #8261, opened 2019-07-23
Fixes VSTS 950315: Code Folding menu toggle does not work in new editor 

avodovnik:
I’ve modified the way the editor reacts to Toggle Folding/Outlining.

We used to rely on the command being remapped to a handler in VSEC, but that caused a few additional problems, like the menu text not always being in sync. 

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/950315/

### DIFF
--- a/main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewContent.cs
+++ b/main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewContent.cs
@@ -369,6 +369,9 @@ namespace MonoDevelop.TextEditor
 		{
 			UpdateLineNumberMarginOption ();
 
+			var foldMargin = PropertyService.Get<bool> ("ShowFoldMargin");
+			Imports.OutliningManagerService.GetOutliningManager (TextView).Enabled = foldMargin;
+
 			var newPolicyContainer = (Owner as IPolicyProvider)?.Policies;
 			if (newPolicyContainer != policyContainer) {
 				if (policyContainer != null)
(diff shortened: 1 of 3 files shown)


main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewContent.cs
```diff
@@ -440,6 +440,9 @@ private Task UpdateOptionsFromEditorConfigAsync (object sender, CodingConvention
 			if (editorConfigContext.CurrentConventions.UniversalConventions.TryGetAllowTrailingWhitespace (out var allowTrailingWhitespace))
 				EditorOptions.SetOptionValue (DefaultOptions.TrimTrailingWhiteSpaceOptionName, !allowTrailingWhitespace);
 
+			var foldMargin = PropertyService.Get<bool>("ShowFoldMargin");
+			Imports.OutliningManagerService.GetOutliningManager (this.TextView as ITextView).Enabled = foldMargin;
```
sandyarmstrong: This should go in `UpdateTextEditorOptionsAsync` instead, near the call to `UpdateLineNumberMarginOption`, because it has nothing to do with .editorconfig. Moving it shouldn't have any functional difference.

sandyarmstrong: is the `as` necessary? I thought `TextView` was already defined as `ITextView`?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/ViewCommandHandlers.cs
```diff
@@ -542,6 +550,7 @@ protected void ToggleAllFoldings ()
 		[CommandHandler (EditCommands.FoldDefinitions)]
 		protected void FoldDefinitions ()
 		{
+			Microsoft.VisualStudio.Text.Editor.ITextView textView;
```
sandyarmstrong: This can be deleted

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewContent.cs
avodovnik: Yeah, it failed to compile for me the first time, which I think was a problem with the state of my VS install (I just did an upgrade of Mono, etc.). I removed it now!

PR #8263, opened 2019-07-23
Use support assemblies when trying to resolve

nosami:
Support assemblies should always be used if they have already
been loaded.

Fixes UI tests when running locally.

### DIFF
--- a/main/src/addins/MonoDevelop.UnitTesting.NUnit/NUnit3Runner/NUnitTestRunner.cs
+++ b/main/src/addins/MonoDevelop.UnitTesting.NUnit/NUnit3Runner/NUnitTestRunner.cs
@@ -56,6 +56,7 @@ namespace NUnit3Runner
 		{
 			engine = TestEngineActivator.CreateInstance ();
 			filterService = engine.Services.GetService<ITestFilterService>();
+			AppDomain.CurrentDomain.AssemblyResolve += CurrentDomain_AssemblyResolve;
 		}
 
 		[MessageHandler]
@@ -223,7 +224,13 @@ namespace NUnit3Runner
 				}
 			}
 		}
-		
+
+		Assembly CurrentDomain_AssemblyResolve (object sender, ResolveEventArgs args)
+		{
+			var name = new AssemblyName (args.Name);
+			return AppDomain.CurrentDomain.GetAssemblies ().FirstOrDefault (a => AssemblyName.ReferenceMatchesDefinition (name, a.GetName ()));
+		}
+
 		private TestFilter CreateTestFilter (string[] nameFilter)
 		{
 			ITestFilterBuilder builder = filterService.GetTestFilterBuilder();


nosami:
Currently seeing this error because Autofac.Extras.AggregateService references Autofac 3.3.0.0 but the test project itself uses 4.2.1. Seems like we should resolve using any support assemblies that have already been loaded.

The tests run without modification using nunit3-console.exe but not from within the IDE.

image

mrward:
Not an expert on the Unit Test runner but we have some assembly loading of support assemblies if they are referenced:

https://github.com/mono/monodevelop/blob/305bddc1e9dcd90caf25f8bed1a042efc317528b/main/src/addins/MonoDevelop.UnitTesting.NUnit/NUnitRunner/NUnitTestRunner.cs#L155-L160

https://github.com/mono/monodevelop/blob/bfce2125b6a1a09795d1f30787ba43317aa4c11b/main/src/addins/MonoDevelop.UnitTesting.NUnit/MonoDevelop.UnitTesting.NUnit/NUnitProjectTestSuite.cs#L219

Do we know why this is not working for a PackageReference project but is working for a packages.config project? Just wondering if for a packages.config project the support assemblies includes AutoFac.dll but for a PackageReference project it does not.

Is AutoFac assembly local copied?

nosami:
> Not an expert on the Unit Test runner but we have some assembly loading of support assemblies if they are referenced:

My fix piggy backs on that code. That only supports direct reference loading, not transitive references. My fix uses the direct references that have already been loaded if there is an assembly name match.

nosami:
> Do we know why this is not working for a PackageReference project but is working for a packages.config project? Just wondering if for a packages.config project the support assemblies includes AutoFac.dll but for a PackageReference project it does not.

Autofac.dll _is_ included in the support assemblies. The problem was that there were 2 different versions of the same assembly. I think PackageReference vs packages.config is probably a red herring and something else changed that has stopped those tests from loading. It is an issue that could affect other users too though so needed fixing regardless.

mrward:
Not sure I really understand the transitive references problem here if AutoFac.dll is being returned as a direct reference 😄

I don't think Project.GetReferences includes transitive references currently for non-SDK style PackageReference projects (there is a hack/workaround I added in the SdkProjectExtension for this). This problem should be fixed on the multi-targeting branch - the hack in the SdkProjectExtension has been removed there.

The fix looks OK - just would like to understand the problem better 😄

nosami:
> I don't think Project.GetReferences includes transitive references currently for non-SDK style PackageReference projects (there is a hack/workaround I added in the SdkProjectExtension for this). This problem should be fixed on the multi-targeting branch - the hack in the SdkProjectExtension has been removed there.

I don't think it would work even if the transitive references list was fixed. There would be a different error when running the Assembly.LoadFrom line that you linked too before when trying to load a 2nd copy of the assembly. You can try running the tests on your branch though to see if it works :)

nosami:
You can see that there is try/catch in real NUnit when doing assembly resolve [here](https://github.com/nunit/nunit-console/blob/aaef18fb80c1a656d0ff194fd0039107cf55de12/src/NUnitEngine/nunit.engine/Internal/ProvidedPathsAssemblyResolver.cs#L74-L94). I think this the reason that nunit3-console.exe works and the IDE doesn't

PR #8265, opened 2019-07-24
Fixes VSTS 945987: Format document doesn't work from Navigate To when using the new editor

avodovnik:
This is a two-part problem. One is that, when a user searches for the command with something like “Format Doc”, there is one result. However, internally there are two - one of which is disabled. It seems to be that when it iterates across them, it invokes the default handler, which reports it being “disabled”. The disabled one is the one we care about, the other one is `WebToolingAddin.WebTooling.Web.Commands.WebCommands.FormatDocument`. 

image

I made a ~small~ change to enable the one we care about to show up, but I believe:
* this might still be a problem for other commands as well that are mapped to the new handlers, and 
* it’ll still be confusing for the user.

What happens underneath:
The [SearchResult](http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/SearchResult.cs,275) calls down to the default handler. It ends up calling into [this](http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs,3b39c23799a56c43) which, in the case of Format Document, means it gets to [`CodeFormattingCommands`](url).

~For now, this is a _small_ “hack” that might help. It might still be confusing for the end user until we replace the Web Tools, from where the document comes up (the other action is `WebToolingAddin.WebTooling.Web.Commands.WebCommands.FormatDocument`).~

I’ve discovered at least one more issue that this **likely** fixes.

Fixes VSTS https://devdiv.visualstudio.com/DevDiv/_workitems/edit/945987
Fixes VSTS https://devdiv.visualstudio.com/DevDiv/_workitems/edit/933997

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -1567,22 +1567,24 @@ namespace MonoDevelop.Components.Commands
 							if (cmd.CommandArray) {
 								handlers.Add (delegate {
 									OnCommandActivating (commandId, info, dataItem, localTarget, source);
-									var t = DateTime.Now;
+									var t = Stopwatch.StartNew ();
 									try {
 										chi.Run (localTarget, cmd, dataItem);
 									} finally {
-										OnCommandActivated (commandId, info, dataItem, localTarget, source, DateTime.Now - t);
+										t.Stop ();
+										OnCommandActivated (commandId, info, dataItem, localTarget, source, t.Elapsed);
 									}
 								});
 							}
 							else {
 								handlers.Add (delegate {
 									OnCommandActivating (commandId, info, dataItem, localTarget, source);
-									var t = DateTime.Now;
+									var t = Stopwatch.StartNew ();
 									try {
 										chi.Run (localTarget, cmd);
 									} finally {
-										OnCommandActivated (commandId, info, dataItem, localTarget, source, DateTime.Now - t);
+										t.Stop ();
+										OnCommandActivated (commandId, info, dataItem, localTarget, source, t.Elapsed);
 									}
 								});
 							}
@@ -1603,7 +1605,40 @@ namespace MonoDevelop.Components.Commands
 					UpdateToolbars ();
 					return true;
 				}
-	
+
+
+				// There is a special case, where some commands get called when we are running in the new editor. For
+				// example, if the user would be focused on a node in Solution Explorer and press Ctrl + I. Our walk
+				// across the controls wouldn't ever land us on the new editor itself, which is a command target. In
+				// those instances, we want to try and get that target and see if we can actually run the command in
+				// the new editor. If not, we'll revert to the old way. This is the same problem (and solution)
+				// that we use in <see cref="DefaultUpdateCommandInfo"/>. 
+				if (IdeApp.Workbench?.ActiveDocument?.GetContent<ICustomCommandTarget> () is ICustomCommandTarget customCommandTarget) {
+					var updater = customCommandTarget.GetCommandUpdater (commandId);
+					var handler = customCommandTarget.GetCommandHandler (commandId);
+
+					if(updater != null) {
+						updater?.Run (cmdTarget, info);
+						info.CancelAsyncUpdate ();
+					}
+
+					if (info.Enabled && handler != null) {
+						var stopwatch = Stopwatch.StartNew ();
+						OnCommandActivating (commandId, info, dataItem, cmdTarget, source);
+						try {
+							handler.Run (cmdTarget, CurrentCommand);
+							return true;
+						} catch (Exception ex) {
+							// if we got here, something went wrong
+							LoggingService.LogInternalError (ex);
+							return false;
+						} finally {
+							stopwatch.Stop ();
+							OnCommandActivated (commandId, info, dataItem, cmdTarget, source, stopwatch.Elapsed);
+						}
+					}
+				}
+
 				if (DefaultDispatchCommand (cmd, info, dataItem, cmdTarget, source)) {
 					UpdateToolbars ();
 					return true;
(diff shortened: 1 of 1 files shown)


slluis:
The fix is not right. Command update handlers are always paired with their run handlers. If the command is enabled in CodeFormattingCommands, it will also be executed by CodeFormattingCommands, which is not what you want.

The SearchResult class looks for command handlers starting at the last active widget (http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/SearchResult.cs,272). If the last active widget as the editor, then it should find the handler implemented (or mapped) by the new editor. If the handler is not found, it may be because MainToolbar.LastCommandTarget doesn't have the right value.

Therzok:
Problem seems to be in CommandManager. [GetActiveWidget](http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs,d3e3d371b958c5bc,references) only saves Gtk.Widgets as the last target of a command, so [CommandSearchResult](http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/SearchResult.cs,7cbdce00bf6f8a92,references) will activate on the wrong widget.

CommandManager needs to be modified to also store non-Gtk widgets as last active

avodovnik:
> The fix is not right. Command update handlers are always paired with their run handlers. If the command is enabled in CodeFormattingCommands, it will also be executed by CodeFormattingCommands, which is not what you want.
> 
> The SearchResult class looks for command handlers starting at the last active widget (http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/SearchResult.cs,272). If the last active widget as the editor, then it should find the handler implemented (or mapped) by the new editor. If the handler is not found, it may be because MainToolbar.LastCommandTarget doesn't have the right value.

So, what happens is that when the command is invoked from within `Navigate to` the Editor does not surface as a control:

```
Dispatching a command "MonoDevelop.Ide.CodeFormatting.CodeFormattingCommands.FormatBuffer"
GetActiveWindow/LastFocused: {MonoDevelop.Ide.Gui.DefaultWorkbench}
IsEmbeddedNSView ({<NSTextView: 0x7fbf1ed89740>
    Frame = {{26.50, 4.00}, {214.00, 22.00}}, Bounds = {{0.00, 0.00}, {214.00, 22.00}}
    Horizontally resizable: YES, Vertically resizable: YES
    MinSize = {214.00, 22.00}, MaxSize = {40000.00, 40000.00}
}), calling superview {<_NSKeyboardFocusClipView: 0x7fbf1e37c510>}
IsEmbeddedNSView ({<_NSKeyboardFocusClipView: 0x7fbf1e37c510>}), calling superview {<MonoDevelop_MacIntegration_MainToolbar_SearchBar: 0x7fbf1e584d40>}
IsEmbeddedNSView ({<MonoDevelop_MacIntegration_MainToolbar_SearchBar: 0x7fbf1e584d40>}), calling superview {<MonoDevelop_MacIntegration_MainToolbar_AwesomeBar: 0x7fbf1ea6af30>}
IsEmbeddedNSView ({<MonoDevelop_MacIntegration_MainToolbar_AwesomeBar: 0x7fbf1ea6af30>}), calling superview {<NSToolbarItemViewer: 0x7fbf1eaa5120 'AwesomeBarToolbarItem'>}
IsEmbeddedNSView ({<NSToolbarItemViewer: 0x7fbf1eaa5120 'AwesomeBarToolbarItem'>}), calling superview {(<NSToolbarView: 0x7fbf1ea966c0>: XSMainToolbar)}
IsEmbeddedNSView ({(<NSToolbarView: 0x7fbf1ea966c0>: XSMainToolbar)}), calling superview {<NSTitlebarView: 0x7fbf1ea3f8b0>}
IsEmbeddedNSView ({<NSTitlebarView: 0x7fbf1ea3f8b0>}), calling superview {<NSTitlebarContainerView: 0x7fbf1ea3d910>}
IsEmbeddedNSView ({<NSTitlebarContainerView: 0x7fbf1ea3d910>}), calling superview {<NSThemeFrame: 0x7fbf1ea3e7c0>}
GetActiveWidget, Mac (plain return): {<NSTextView: 0x7fbf1ed89740>
    Frame = {{26.50, 4.00}, {214.00, 22.00}}, Bounds = {{0.00, 0.00}, {214.00, 22.00}}
    Horizontally resizable: YES, Vertically resizable: YES
    MinSize = {214.00, 22.00}, MaxSize = {40000.00, 40000.00}
}
```

However, if called directly from the editor:

```
Dispatching a command "MonoDevelop.Ide.CodeFormatting.CodeFormattingCommands.FormatBuffer"
GetActiveWindow/LastFocused: {MonoDevelop.Ide.Gui.DefaultWorkbench}
IsEmbeddedNSView ({<Microsoft_VisualStudio_Text_Editor_Implementation_CocoaTextViewControl: 0x7fbf084ed960>}), calling superview {<Microsoft_VisualStudio_Text_Editor_Implementation_CocoaEditorGridView: 0x7fbf0b002ae0>}
IsEmbeddedNSView ({<Microsoft_VisualStudio_Text_Editor_Implementation_CocoaEditorGridView: 0x7fbf0b002ae0>}), calling superview {<GdkQuartzView: 0x7fbf1e5774b0>}
GetActiveWidget Mac, IsEmbeddedView: {<Microsoft_VisualStudio_Text_Editor_Implementation_CocoaTextViewControl: 0x7fbf084ed960>}
```

Personally, I’d expect the commanding to always target the editor, but I must be missing something. I’ll explore a bit more to see if I can find a reasonable solution.

avodovnik:
I’ve updated the PR.

If commands were invoked outside the editor surface (i.e. from the Navigate To widget), they would have focus set to that UI element and we would never walk through the hierarchy in a way that would get us to the editor. Now, if the walk-through fails, we simply try and get the custom command target implemented by the new editor, and see if it can handle the command.

Also, we replicate the same behavior that happens across the rest of the dispatching, which is that we invoke the Activating and Activated events.

I also replaced the usage of DateTime.Now with a Stopwatch as it's much more precise.

Fixes VSTS https://devdiv.visualstudio.com/DevDiv/_workitems/edit/945987
Fixes VSTS https://devdiv.visualstudio.com/DevDiv/_workitems/edit/933997

avodovnik:
Fix implemented differently in #8323, closing this PR.

PR #8268, opened 2019-07-24
[Debugger] Fixed .NET Core debugger to allow use of SHA256 and MD5 ha…

jstedfast:
…shes

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/951772

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeStackFrame.cs
+++ b/main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeStackFrame.cs
@@ -58,25 +58,58 @@ namespace MonoDevelop.Debugger.VsCodeDebugProtocol
 			this.frameId = frame.Id;
 		}
 
-		static byte [] HexToByteArray (string hex)
+		static byte ToXDigit (char c)
 		{
-			if (hex.Length % 2 == 1)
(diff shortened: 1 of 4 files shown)


main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeStackFrame.cs
```diff
@@ -58,25 +58,58 @@ public VsCodeStackFrame (VsFormat format, int threadId, int frameIndex, VsStackF
...
 			if (hex.Length % 2 == 1)
-				throw new ArgumentException ();
-			byte [] bytes = new byte [hex.Length / 2];
-			for (int i = 0; i < bytes.Length; i++) {
-				bytes [i] = Convert.ToByte (hex.Substring (i * 2, 2), 16);
+				return null;
+
+			try {
+				var bytes = new byte[hex.Length / 2];
+				for (int i = 0; i < bytes.Length; i += 2) {
```
mrward: This looks like it will skip over a byte and not process part of the hex information. i++?

For "FFAA" it returns 255, 00 instead of 255, 170

jstedfast: Thanks for that catch

jstedfast:
It has unit tests now.

PR #8275, opened 2019-07-25
Build: Use a property to access the vs-editor-api directory

sandyarmstrong:
This should make it easier for us to have all vs-editor-api references
point to vs-editor-core when available. There are some blockers to that
change, so for now just take this one preparatory step.

### DIFF
--- a/main/Directory.Build.props
+++ b/main/Directory.Build.props
@@ -4,7 +4,8 @@
     <RootDirectory>$(MSBuildThisFileDirectory)</RootDirectory>
     <PackagesDirectory>$(RootDirectory)packages</PackagesDirectory>
     <MdAddinsDirectory>$(MSBuildThisFileDirectory)..\..\md-addins\</MdAddinsDirectory>
-    <VSEditorCoreDirectory>$(MdAddinsDirectory)external\vs-editor-core\</VSEditorCoreDirectory>
+    <VSEditorCoreDirectory Condition="'$(VSEditorCoreDirectory)' == ''">$(MdAddinsDirectory)external\vs-editor-core\</VSEditorCoreDirectory>
+    <VSEditorApiDirectory Condition="'$(VSEditorApiDirectory)' == ''">$(MSBuildThisFileDirectory)external\vs-editor-api\</VSEditorApiDirectory>
     <ReferencesVSEditor Condition=" '$(OS)' == 'Windows_NT' ">$(RootDirectory)\msbuild\ReferencesVSEditor.Windows.props</ReferencesVSEditor>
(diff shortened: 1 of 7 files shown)


main/Directory.Build.props
```diff
@@ -5,6 +5,7 @@
     <PackagesDirectory>$(RootDirectory)packages</PackagesDirectory>
     <MdAddinsDirectory>$(MSBuildThisFileDirectory)..\..\md-addins\</MdAddinsDirectory>
     <VSEditorCoreDirectory>$(MdAddinsDirectory)external\vs-editor-core\</VSEditorCoreDirectory>
```
abock: Would probably be a good idea to go ahead and condition both of these properties to allow them to be overridden:

```xml
<VSEditorCoreDirectory Condition="'$(VSEditorCoreDirectory)' == ''">$(MdAddinsDirectory)external\vs-editor-core\</VSEditorCoreDirectory>
<VSEditorApiDirectory Condition="'$(VSEditorApiDirectory)' == ''">external\vs-editor-api\</VSEditorApiDirectory>
```

Additionally, wouldn't _really_ want something like this?


```xml

<VSEditorCoreDirectory Condition="'$(VSEditorCoreDirectory)' == ''">$(MdAddinsDirectory)external\vs-editor-core\</VSEditorCoreDirectory>
<VSEditorApiDirectory Condition="'$(VSEditorApiDirectory)' == '' And Exists('$(VSEditorCoreDirectory)'">$(VSEditorCoreDirectory)</VSEditorApiDirectory>
<VSEditorApiDirectory Condition="'$(VSEditorApiDirectory)' == ''">external\vs-editor-api\</VSEditorApiDirectory>
```

sandyarmstrong: Cool, will add the first condition suggested.

And yes, the original idea was to do as you suggested, so that `vs-editor-api` stuff gets overridden by `vs-editor-core` stuff. But there are various problems there:

1. `vs-editor-core` stuff wouldn't be in the solution, which means restore and debugging wouldn't work without additional steps.
2. `vs-editor-api` has a `ProjectReferences.projitems` file used by MD build that doesn’t exist in `vs-editor-core`.
3. More generally, making the `monodevelop` build depend on `md-addins` already being checked out and synchronized is going to be a big source of confusion for VSmac devs, and makes me think we should wait for either editor nugetization or repo consolidation to take any further steps.

Anyway, this PR is just a baby step to add some consistency.

abock: Sounds good. The second condition will make sense when we revisit the consolidation PR that introduces `.projitems` files that make sense across the repo split. (3) is a continually sunk cost and we have other plans to address it.

PR #8280, opened 2019-07-25
Updated reference to xamarin/md-addins@d726360

dalexsoto:


sevoku:
Already bumped using Wrench: https://github.com/mono/monodevelop/pull/8287

PR #8281, opened 2019-07-25
[Project] Change the default DotNetProject platform to anycpu

sgmunn:
Fixes VSTS 934343
[CSharp] Bad CPU type in executable trying to Debug/Run a default template of a CSharp project in catalina
https://vsmac.dev/934343

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectService.cs
@@ -53,7 +53,7 @@ namespace MonoDevelop.Projects
 
 		TargetFramework defaultTargetFramework;
 		
-		string defaultPlatformTarget = "x86";
+		string defaultPlatformTarget = "anycpu";
 		static readonly TargetFrameworkMoniker DefaultTargetFrameworkId = TargetFrameworkMoniker.NET_4_7;
 		
 		public const string BuildTarget = "Build";


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectService.cs
```diff
@@ -53,7 +53,7 @@ public class ProjectService
 
 		TargetFramework defaultTargetFramework;
 		
-		string defaultPlatformTarget = "x86";
+		string defaultPlatformTarget = "anycpu";
```
Therzok: This isn't going to fix projects retroactively.

https://github.com/mono/monodevelop/blob/ad32b280b6e7d2fdfa1ba99920d72da11c4d890d/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/AssemblyUtilities.cs#L93-L96

We should apply a change here to return x64 on >= Catalina.

mrward (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectService.cs
sgmunn: this fix wasn't intending to fix things retrospectively. I'm not sure I'm following what you mean here, doesn't the code you linked do what you're suggesting, or is that an example only?

Therzok: The code set preferred 32 bit to run as 64bit. We should make it return 64bit only on mac+.

Maybe if it requests specific 32bit, warn once per project?

sgmunn:
Pretty sure the Microsoft.VSForMac.Tests.AzureFunctionsTemplateTests.TestCreateBuildHttpTrigger test is unrelated, it has been timing out on many different PRs

Therzok:
Old projects will still fail to run on Catalina. We need to either force 64bit ness or provide guidance to migrate.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectService.cs
Therzok: Filed https://devdiv.visualstudio.com/DevDiv/_workitems/edit/960622

mrward:

PR #8282, opened 2019-07-25
[Debugger] Added debugger logging to try and track down a bug

jstedfast:


sgmunn:
hard to spot error, but might be this

> SoftDebuggerBacktrace.cs(90,1): error CS0023: Operator '+' cannot be applied to operand of type 'string' [/Users/vsts/agent/2.154.3/work/1/s/monodevelop/main/external/debugger-libs/Mono.Debugging.Soft/Mono.Debugging.Soft.csproj]

jstedfast:
Thanks

PR #8290, opened 2019-07-26
[MSBuild] Bump MSBuild Tasks to target .NET 4.7.2

manish:
`msbuild -t:Pack -p:Configuration=Debug -p:TargetFramework=net472` fails


main/msbuild/MDBuildTasks/MDBuildTasks.csproj
```diff
@@ -5,6 +5,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Build.Tasks.Core" Version="15.9.20" />
+    <PackageReference Include="System.IO.Compression" />
```
mrward: I think this should have a specific version.

PR #8293, opened 2019-07-27
[release-8.2] Fix ObjectDisposedException with TouchBar – Xwt bump

sevoku:
Updated reference to xwt@86ccb40 https://github.com/mono/xwt/pull/955

This is a hack, macOS seems to be calling conformsToProtocol:
after a window has been released, resulting in a crash.

Fixes VSTS #935146


slluis:
Already bumped.

PR #8294, opened 2019-07-27
Change Xwt project type to multi-targeted SDK

sevoku:
For future a11y work we need to conditionally support different .NET versions and therefore we need to change the project type of Xwt and Xwt.WPF projects.

This depends on / bumps to: https://github.com/mono/xwt/pull/954


Therzok:
I guess it was already bumped.

Therzok:
Yup!

PR #8298, opened 2019-07-29
[Debugger] Fixed toggling of breakpoints at the end of a line

jstedfast:
Just like line numbers, column numbers are also 1-based, but when
calculating the byte position, we need to adjust for 0-based indexes.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/917922/


sgmunn:
Test failure in web tools is unrelated.

sgmunn:
Microsoft.VSForMac.Tests.AzureFunctionsTemplateTests.TestCreateBuildHttpTrigger seems to be timing out everywhere atm, unrelated.

PR #8061, opened 2019-06-28
Fixes VSTS Bug 938343: [Feedback] Diff Option in Version Control not

mkrueger:
working as expected.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/938343

PR #8301, opened 2019-07-30
Fixes VSTS Bug 952599: Version control document views are missing with

mkrueger:
the new editor

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/952599

The base class of the new editor has changed so it no longer works.
Checking the file model is the faster solution in any case.


sevoku:
VCS views not reflecting unsaved changes sometimes is another issue, not related to enabled/disabled commands. In fact we need to switch to the new editor for these views.
So this change LGTM.

sevoku:
Hhm, looks like #8061 also regressed

PR #8309, opened 2019-07-31
Fix stack overflow at startup

slluis:
The IdeApp.DispatchIdleActions method is sometimes called just after startup.
When that happens, if the user has not had any interaction with the ide,
the CommandService.LastUserInteration property will have the default DateTime
value. The old code was substracting that value from the current time, which
would be a huge amount of milliseconds, and would overflow when converting
to int. It all would result with DispatchIdleActions being called over and over
with a negative number, finally causing a stack overflow.

Fixes VSTS #958148


slluis:

PR #8311, opened 2019-07-31
[Project Model] Shortcut for add new class in C# projects

ketttY:
Create shortcut for New Class... from the folder context-menu within the solution pad. From the Add menu, if New Class... is clicked, in the New File dialog, General | Empty Class is always selected.

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/643343

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/ProjectCommands.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/ProjectCommands.cs
@@ -62,7 +62,6 @@ namespace MonoDevelop.Ide.Commands
 		
 		AddReference,
 		AddNewFiles,
-		AddEmptyClass,
 		AddFiles,
 		NewFolder,
 		AddFilesFromFolder,
@@ -90,7 +89,8 @@ namespace MonoDevelop.Ide.Commands
 		SelectActiveRuntime,
 		EditSolutionItem,
 		Unload,
-		SetStartupProjects
+		SetStartupProjects,
+		AddEmptyClass
 	}
 
 	internal class SolutionOptionsHandler : CommandHandler
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/ProjectCommands.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/ProjectCommands.cs
@@ -62,6 +62,7 @@ namespace MonoDevelop.Ide.Commands
 		
 		AddReference,
 		AddNewFiles,
+		AddEmptyClass,
 		AddFiles,
 		NewFolder,
 		AddFilesFromFolder,
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/ProjectCommands.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/ProjectCommands.cs
@@ -90,7 +90,7 @@ namespace MonoDevelop.Ide.Commands
 		EditSolutionItem,
 		Unload,
 		SetStartupProjects,
-		AddEmptyClass
+		AddEmptyClass,
 	}
 
 	internal class SolutionOptionsHandler : CommandHandler
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
@@ -452,7 +452,7 @@ namespace MonoDevelop.Ide.Gui.Pads.ProjectPad
 		{
 			var project = (Project)CurrentNode.GetParentDataItem (typeof (Project), true);
 			if (project != null) {
-				info.Visible = project.SupportedLanguages.Contains ("C#");
+				info.Visible = IdeApp.ProjectOperations.CanCreateProjectFile (project, GetFolderPath (CurrentNode.DataItem), "EmptyClass");
 			}
 		}
 
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
@@ -435,7 +435,24 @@ namespace MonoDevelop.Ide.Gui.Pads.ProjectPad
 			if (IdeApp.Workbench.ActiveDocument != null)
 				IdeApp.Workbench.ActiveDocument.Select ();
 		}
-		
+
+		[CommandHandler (ProjectCommands.AddEmptyClass)]
+		protected void OnAddCSharpClass ()
+		{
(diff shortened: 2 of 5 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
```diff
@@ -435,7 +435,24 @@ public void AddNewFileToProject()
...
-		
+
+		[CommandHandler (ProjectCommands.AddEmptyClass)]
+		protected void OnAddCSharpClass ()
+		{
+			var project = (Project)CurrentNode.GetParentDataItem (typeof (Project), true);
+			if (IdeApp.ProjectOperations.CreateProjectFile (project, GetFolderPath (CurrentNode.DataItem), "EmptyClass")) {
+				CurrentNode.Expanded = true;
+			}
+		}
```
mrward: AddNewFiles selects the ActiveDocument after a new file is created. Not sure if we need to do the same here? I guess it gives focus to the new file that was added.

			if (IdeApp.Workbench.ActiveDocument != null)
				IdeApp.Workbench.ActiveDocument.Select ();

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
```diff
@@ -435,7 +435,24 @@ public void AddNewFileToProject()
...
+
+		[CommandUpdateHandler (ProjectCommands.AddEmptyClass)]
+		protected void UpdateAddEmptyClass (CommandInfo info)
+		{
+			var project = (Project)CurrentNode.GetParentDataItem (typeof (Project), true);
+			info.Visible = project.SupportedLanguages.Contains ("C#");
+		}
```
mrward: Wondering if there is a better way to do this than hard code 'C#'.

Currently the EmptyClass file template supports [VB.NET and C#](https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/templates/EmptyClass.xft.xml#L8). 

Be nice if there was a way to do something like:

    if (IdeApp.ProjectOperations.CanCreateProjectFile (project, "EmptyClass")) {

Then have some code that checks the language is supported by the file template.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/ProjectCommands.cs
```diff
@@ -62,6 +62,7 @@ public enum ProjectCommands
 		
 		AddReference,
 		AddNewFiles,
+		AddEmptyClass,
 		AddFiles,
```
mrward: Probably should add this at the end of the enum. I believe this is an API break since this is an int and putting it in the middle of the list changes what the original ints were assigned to.

https://stackoverflow.com/questions/3422178/is-changing-the-number-of-an-enum-a-breaking-change

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
ketttY: Good idea, I will add it now.

ketttY: I have just tested it, it seems like it is already automatically giving focus to the new file. Do you think we still need these lines to make sure?

mrward: If it is working then no need to add them. We could remove them from the other method but that could be done later in a separate pull request.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
slluis: Also, as a sanity check, it wouldn't hurt to do a null check on project.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/ProjectCommands.cs
```diff
@@ -89,7 +89,8 @@ public enum ProjectCommands
 		SelectActiveRuntime,
 		EditSolutionItem,
 		Unload,
-		SetStartupProjects
+		SetStartupProjects,
+		AddEmptyClass
```
Therzok: Style nit:
```suggestion
		AddEmptyClass,
```

Even though the IDE might show it as redundant, it makes it easier for future diffs if we just add a trailing comma, as the next change will just add one line, instead of adding one and modifying one.

Therzok:
Mostly LGTM!

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
```diff
@@ -435,7 +435,28 @@ public void AddNewFileToProject()
 			if (IdeApp.Workbench.ActiveDocument != null)
 				IdeApp.Workbench.ActiveDocument.Select ();
 		}
-		
+
+		[CommandHandler (ProjectCommands.AddEmptyClass)]
+		protected void OnAddCSharpClass ()
+		{
```
mrward: I think this method name can be changed to be `OnAddEmptyClass` since we are not necessarily adding a C# class now.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ProjectOperations.cs
```diff
@@ -884,6 +885,21 @@ public bool CreateProjectFile (Project parentProject, string basePath, string se
 			}
 		}
 
+		public bool CanCreateProjectFile (Project parentProject, string basePath, string selectedTemplateId)
+		{
+			var projectLangs = parentProject.SupportedLanguages;
+			var template = FileTemplate.GetFileTemplateByID (selectedTemplateId);
```
mrward: One small thing here - I guess we should check the template is not null and return false if it is null.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
```diff
@@ -435,7 +435,28 @@ public void AddNewFileToProject()
...
+			}
+		}
+
+		[CommandUpdateHandler (ProjectCommands.AddEmptyClass)]
+		protected void UpdateAddEmptyClass (CommandInfo info)
+		{
+			var project = (Project)CurrentNode.GetParentDataItem (typeof (Project), true);
+			if (project != null) {
+				info.Visible = IdeApp.ProjectOperations.CanCreateProjectFile (project, GetFolderPath (CurrentNode.DataItem), "EmptyClass");
+			}
```
mrward: Probably need to debug this to check but if project is null I think the Add New Class menu item will be visible. I think we should be setting info.Visible to false if the project is null.

PR #8312, opened 2019-07-31
Tasks: Stop highlighting "undone"

sandyarmstrong:
I don't know if this was originally supposed to be "undome", but as-is
it doesn't seem like a word that denotes a task.

I looked to see when this was added to see if I could learn more, but
there's not much info in https://github.com/mono/monodevelop/commit/05db27cf6c06423d8dae5e71e0b9d1dc60388af7.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Tasks/CommentTag.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Tasks/CommentTag.cs
@@ -44,7 +44,7 @@ namespace MonoDevelop.Ide.Tasks
 		
 		public static event EventHandler SpecialCommentTagsChanged;
 		
-		const string defaultTags = "FIXME:2;TODO:1;HACK:1;UNDONE:0";
+		const string defaultTags = "FIXME:2;TODO:1;HACK:1";
 		static List<CommentTag> specialCommentTags;
 		
 		public static List<CommentTag> SpecialCommentTags {


sandyarmstrong:
I did this as a quick web edit just to open the question, but it looks like some minor unit test changes will be needed, as well as (maybe) an adjustment to `FSharpSyntaxMode.xml`.

sandyarmstrong:
Incidentally, based on comments in unit tests, and on existing "UNDONE" comments in MonoDevelop source code, I'm pretty sure we have been taking it to mean "not yet done".

But then in https://en.wikipedia.org/wiki/Comment_(computer_programming)#Tags (as pointed out by @mrward), the following definition is given:

> UNDONE – a reversal or "roll back" of previous code.

Very confusing.

Anyway, I think "undone" is too common of a word to highlight, and should not be a default. Users can always add it.

mhutch:
I suspect we'd eliminate false positives if it was a case sensitive match on the first word on the line...

avodovnik:
> I suspect we'd eliminate false positives if it was a case sensitive match on the first word on the line...

That becomes a really computationally intensive task though, doesn't it? I mean, it means separating trigger words at the very least, or using a RegEx, which we've established is much slower than `IndexOf` in this case. Optionally, we might want to ONLY highlight uppercased words, like `TODO`, `FIXME`, etc., and not `todo`, `fixme`.

mhutch:
Is it computationally intensive? Just grab the first token out of every line comment and, if it's made out of uppercase letters only, check against a set

avodovnik:
Well... Maybe. Right now it's just using `.IndexOf` instead of tokenizing the string. Happy to have a chat if you have a better/faster suggestion! 

As for `undone`, I don't personally see the value in it, but that doesn't mean there isn't any. I think Sandy has a point, we should allow users to add it as a keyword (later).

mhutch:
I don't necessarily disagree about removing undone, just saying that the same kind of false positive could easily apply to other tags, and a more robust check would be good.

Even if continuing to use IndexOf, you could require uppercase and check each match for whether it's at the start of the line.

sandyarmstrong:
I like the idea of only checking for the first word in the comment. I don't agree about checking case, though.

And yeah, with that change I wouldn't care as much about "undone".

mhutch:
Well, maybe only handling uppercase is excessive, but doing a case sensitive match on the registered tag seems reasonable to me. Folks could register additional casings if they wanted.

Therzok:
cc @vancura as this is a defaults change

vancura:
This is not a design change, but source analysis one, I am not sure I can provide any meaningful opinion here.

But personally I don't recall seeing this token ever, it's utterly new to me. I don't have any issues undoneing the highlighting :)

Therzok:
@monojenkins rebase

Therzok:
Yep!

Therzok:
Wrong PR

PR #8317, opened 2019-08-01
Bug 948354: Unpinned docked pads appear unexpectedly after a brief

mkrueger:
mouse over

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/935140


slluis:
I'm now thinking that if this fails, the whole IDE is going to be broken. Something is not working right in the add-in engine, it shouldn't crash in this way. Let me take a look at mono-addins before merging this.

PR #8319, opened 2019-08-01
Fixes VSTS Bug 944655: [FATAL] System.ArgumentException exception in

mkrueger:
System.Threading.Tasks.Task.WhenAll[PEFile]()

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/944655

That exact crash shouldn't happen anymore in the current version
(definitions is now an ImmutableList and never null) but a crash at
that point should not crash the IDE in any case.

### DIFF
--- a/main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserWidget.cs
+++ b/main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserWidget.cs
@@ -917,31 +917,36 @@ namespace MonoDevelop.AssemblyBrowser
 		
 		internal void Open (string url, AssemblyLoader currentAssembly = null, bool expandNode = true)
 		{
-			Task.WhenAll (this.definitions.Select (d => d.LoadingTask)).ContinueWith (d => {
-				// At least one of them failed.
-				if (d.IsFaulted) {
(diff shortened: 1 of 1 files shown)


sevoku:
Hhm, looks like #8061 also regressed

Therzok:
It is still possible we have a null loader task in the immutablelist.

Therzok:
Maybe instead of setting [LoadingTask](http://source.monodevelop.com/#MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyLoader.cs,238) to null, we could set it to `Task.CompletedTask`

mkrueger:
The crash was not because of a single loading task beeing null. LoadingTask is never null at that point btw.

ad3d7e5565d8aaf729a91ae60ca48cf8f67aac88

was the commit btw.

Therzok:
@mkrueger The problem is not the list of tasks, it's _one task_. It can be null, in a use-after-dispose scenario. Not sure how open would be triggered _after_ the widget is destroyed, but it is highly plausible, due to lack of cancellation of various tasks on Dispose.

Maybe there's a bug caused by [this code](http://source.monodevelop.com/#MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserWidget.cs,848). As we can see there, after a new assembly browser view is opened, the `Open` call _still happens on the current one_

mkrueger:
Y could be but can't happen in current code. 
Fixed in 4ff15f7fbe2abac3b188805ee22dd1b2c59d5316

PR #7521, opened 2019-05-09
Added support to native dialogs (Cocoa) in the CommandManager

jsuarezruiz:
Added support to native dialogs (Cocoa) in the CommandManager. Added to be able to manage commands from native Windows (Example: GTC).

Fixes VSTS #796177
Fixes VSTS #801677

PR #8323, opened 2019-08-02
Fixes VSTS 945987: Format document doesn't work from Navigate To when using the new editor

avodovnik:
This is the second attempt at fixing VSTS 945987, where Format Document doesn't work if invoked from the Navigate To window. The first PR https://github.com/mono/monodevelop/pull/8265 attempted to resolve in a way that would allow commands to execute correctly on the handler, but as was pointed out, we also needed a way to allow native widgets to be focused. 

I built on changes from here #7521 and extended the fix as it kept throwing exceptions. 

Fixes https://vsmac.dev/945987

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -2280,11 +2280,15 @@ namespace MonoDevelop.Components.Commands
 				if (nw != null) {
 					var v = nw.FirstResponder as AppKit.NSView;
 					if (v != null && !IsRootGdkQuartzView (v)) {
+
+						ChangeActiveWidget (this, v);
+
 						if (IsEmbeddedNSView (v))
 							// FIXME: since there is no way to get the parent GTK widget of an embedded NSView,
 							// here we return a ICommandDelegatorRouter object that will cause the command route
 							// to continue with the active gtk widget once the NSView hierarchy has been inspected.
 							return new NSViewCommandRouter { ActiveView = v, ParentWidget = GetFocusedChild (widget) };
+
 						return v;
 					}
 				}
@@ -2299,13 +2303,18 @@ namespace MonoDevelop.Components.Commands
 
 				widget = GetFocusedChild (widget);
 			}
-			if (widget != lastActiveWidget) {
-				if (ActiveWidgetChanged != null)
-					ActiveWidgetChanged (this, new ActiveWidgetEventArgs () { OldActiveWidget = lastActiveWidget, NewActiveWidget = widget });
-				lastCommandTarget = new WeakReference (lastActiveWidget);
-				lastActiveWidget = widget;
-			}
+
+			ChangeActiveWidget (this, widget);
 			return widget;
+
+			static void ChangeActiveWidget (CommandManager cmdManager, Control newWidget)
+			{
+				if (newWidget == cmdManager.lastActiveWidget) return;
+
+				cmdManager.ActiveWidgetChanged?.Invoke (cmdManager, new ActiveWidgetEventArgs () { OldActiveWidget = cmdManager.lastActiveWidget, NewActiveWidget = newWidget });
+				cmdManager.lastCommandTarget = new WeakReference (cmdManager.lastActiveWidget);
+				cmdManager.lastActiveWidget = newWidget;
+			}
 		}
 
 #if WINDOWS
@@ -2339,7 +2348,7 @@ namespace MonoDevelop.Components.Commands
 			Gtk.Container container;
 
 			do {
-				container = widget.GetNativeWidget<Gtk.Container> ();
+				container = widget?.nativeWidget is Gtk.Container ? widget.GetNativeWidget<Gtk.Container> () : null;
 				if (container != null) {
 					Gtk.Widget child = container.FocusChild;
 					if (child != null)
@@ -2349,11 +2358,11 @@ namespace MonoDevelop.Components.Commands
 				}
 			} while (container != null);
 
-			return widget;
+			return widget.nativeWidget is Gtk.Widget ? widget : null;
 		}
 
-		#if MAC
-		class NSViewCommandRouter: ICommandDelegatorRouter
+#if MAC
+		class NSViewCommandRouter : ICommandDelegatorRouter
 		{
 			public AppKit.NSView ActiveView;
 			public Gtk.Widget ParentWidget;
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -2238,9 +2238,8 @@ namespace MonoDevelop.Components.Commands
 						if (win == null)
 							win = w;
 					}
-					if (lastFocused.nativeWidget == w) {
-						lastFocusedExists = true;
-					}
+
+					lastFocusedExists |= lastFocused?.nativeWidget == w;
 				}
 			}
 
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -335,8 +335,13 @@ namespace MonoDevelop.Components.Commands
 					// If the window is a gtk window and is registered in the command manager
 					// process the events through the handler.
 					var gtkWindow = Mac.GtkMacInterop.GetGtkWindow(window);
-					if (gtkWindow != null && !TopLevelWindowStack.Any (w => w.nativeWidget == gtkWindow))
+					if (gtkWindow != null &&
+						!TopLevelWindowStack.Select (x => x.nativeWidget).Contains (gtkWindow)) {
+						// the above is slightly more contrived than using a simple .Any statement
+						// because of a Roslyn bug that could potentially affect our performance
+						// see here: https://github.com/dotnet/roslyn/issues/20777
 						return null;
+					}
 				}
 			}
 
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Control.cs
```diff
@@ -87,8 +87,10 @@ public T GetNativeWidget<T> ()
 			}
 			if (nativeWidget is T resultWidget)
 				return resultWidget;
-			else
-				throw new NotSupportedException ();
+
+			// we used to throw a NotSupportedException, however, that prevented us from
+			// doing anything with native controls, see: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/945987
+			return null;
```
sevoku: I'm ok with being less strict here, but I'm not sure if there was a specific reason to throw here. @Therzok ?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Window.cs
```diff
@@ -37,6 +37,30 @@ internal protected Window (object window) : base (window)
 		{
 		}
 
+		public bool IsRealized {
```
sevoku: The naming here might be confusing, `IsRealized` means in the Gtk world that the widget is "ready"/initialized and has been added to a parent, but not necessarily visible (in fact showing a widget will realize it first), for a Gtk.Window this means that it has a Gdk.Window. The better match from the Cocoa world would be `NSView.Superview != null`, but I'm not sure about `NSWindow`...

Is this used somwhere? If not, maybe let's remove that? Otherwise `IsVisible` would make more sense.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -334,8 +334,8 @@ AppKit.NSEvent OnNSEventKeyPress (AppKit.NSEvent ev)
 				{
 					// If the window is a gtk window and is registered in the command manager
 					// process the events through the handler.
-					var gtkWindow = MonoDevelop.Components.Mac.GtkMacInterop.GetGtkWindow(window);
-					if (gtkWindow != null && !TopLevelWindowStack.Contains(gtkWindow))
+					var gtkWindow = Mac.GtkMacInterop.GetGtkWindow(window);
+					if (gtkWindow != null && !TopLevelWindowStack.Any (w => w.nativeWidget == gtkWindow))
```
Therzok: Can we avoid a lambda closure allocation on every keypress? Roslyn generates an [indiscriminate allocation](https://sharplab.io/#gist:0bc4f6e975a4d61b4db392b165002339) here, regardless of the code blocks.

See [roslyn issue](https://github.com/dotnet/roslyn/issues/20777)

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -2231,7 +2233,9 @@ Gtk.Window GetActiveWindow (Gtk.Window win)
 			if (!hasFocus) {
 				var nsWindow = AppKit.NSApplication.SharedApplication.KeyWindow;
 				hasFocus = nsWindow != null;
-				lastFocused = win = Mac.GtkMacInterop.GetGtkWindow (nsWindow);
+                if (hasFocus) {
```
Therzok:
```suggestion
				if (hasFocus) {
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -2315,15 +2319,21 @@ Windows.GtkWPFWidget GetFocusedWpfWidget ()
...
-				Gtk.Widget child = ((Gtk.Container)widget).FocusChild;
-				if (child != null)
-					widget = child;
-				else
-					break;
-			}
+			Gtk.Container container;
+
+			do {
+				container = widget.GetNativeWidget<Gtk.Container> ();
```
Therzok: I wonder if we should just do:
```csharp
while (widget.nativeWidget is Gtk.Container)
```

To preserve semantics and not deal with the extra `CommandRouterContainer`'s `GetNativeWidget` could create

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Control.cs
Therzok: `GetNativeWidget` should be called only when we know it should contain that type. This change can be possibly reverted if the [suggested change](https://github.com/mono/monodevelop/pull/8323/files#r310346207) here is applied.

Therzok: Basically, if we are requesting the bad control type, we should throw.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
avodovnik: This also throws an exception, as it seems this does a conversion in the background, and checks the result. In our case, however, that results in calling the `GetNativeWidget` from the operator, and throwing an exception. I don't want to be using Exceptions for code flow, though.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
avodovnik: Not sure if this is a GitHub rendering quick, but this seems to be aligned fine in the IDE.

sandyarmstrong: Other lines here use tabs; this added line uses spaces for indentation instead. So it looks off in GitHub because their tab width happens to be different from yours. Change to tabs and all shall be well!

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Window.cs
avodovnik: @sevoku looking at the code, it feels IsRealized is probably the subset of both methods - as in, it will call `IsRealized` in GTK world, and `IsVisible` in Cocoa world.

sandyarmstrong: Maybe we just add an XML comment on here explaining, and leave as-is, since we're already in this GTK-focused frame of mind here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -2315,15 +2324,21 @@ Windows.GtkWPFWidget GetFocusedWpfWidget ()
 		}
 #endif
 
-		Gtk.Widget GetFocusedChild (Gtk.Widget widget)
+		Gtk.Widget GetFocusedChild (Control widget)
```
Therzok:
```suggestion
		Gtk.Widget GetFocusedChild (Gtk.Container widget)
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -2280,11 +2289,11 @@ object GetActiveWidget (Gtk.Window win)
 
 				widget = GetFocusedChild (widget);
```
Therzok: Over here, pass `win.nativeWidget as Gtk.Container`

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -2249,15 +2258,15 @@ Gtk.Window GetActiveWindow (Gtk.Window win)
 				return null;
 		}
 		
-		object GetActiveWidget (Gtk.Window win)
+		object GetActiveWidget (Window win)
 		{
 			win = GetActiveWindow (win);
 
-			Gtk.Widget widget = win;
+			Control widget = win;
```
Therzok: This is redundant.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -2249,15 +2258,15 @@ Gtk.Window GetActiveWindow (Gtk.Window win)
...
 			if (win != null) {
 
 				#if MAC
-				var nw = MonoDevelop.Components.Mac.GtkMacInterop.GetNSWindow (win);
+				var nw = win.nativeWidget as AppKit.NSWindow;
 				if (nw != null) {
 					var v = nw.FirstResponder as AppKit.NSView;
 					if (v != null && !IsRootGdkQuartzView (v)) {
```
Therzok: You need to save the lastActiveWidget here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Control.cs
```diff
@@ -87,8 +87,8 @@ public T GetNativeWidget<T> ()
 			}
 			if (nativeWidget is T resultWidget)
 				return resultWidget;
-			else
-				throw new NotSupportedException ();
+
+			throw new NotSupportedException ($"Cannot get native widget {typeof (T)}");
```
Therzok: :heart:

sandyarmstrong:
@slluis this still requires a codeowner review from you, when you get a chance

sandyarmstrong:
@avodovnik let's hold this until test failures are figured out. Sevo says DDRITs and Azure Integration test failures all start with:

> Xamarin.TestEnvironment.TestFailureException : Failed to open New Project Dialog

So it seems like something regressed here for test automation.

slluis:
I relaunched the tests. Let's see what happens.

sandyarmstrong:
@monojenkins  backport release-8.3

Therzok:

PR #8325, opened 2019-08-02
Artificially rank the commands from the current add-in higher in Navigate To

avodovnik:
This will help surface IDE commands higher than those coming from addins, which is helpful when the addins don't follow convention or create the same commands under new ids (i.e. Format Document).

image

This is an example of how the WebAddin contributes commands that somehow overpower the Format Document command we surface. I feel we should rank _internal_ ones higher. 

This will surface with users when this is merged: https://github.com/mono/monodevelop/pull/8323

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/CommandSearchCategory.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/CommandSearchCategory.cs
@@ -38,15 +38,19 @@ namespace MonoDevelop.Components.MainToolbar
 	class CommandSearchCategory : SearchCategory
 	{
 		static readonly List<Tuple<Command, string>> allCommands;
-		private static readonly Mono.Addins.RuntimeAddin currentRuntimeAddin;
+		static readonly Mono.Addins.RuntimeAddin currentRuntimeAddin;
+
 		static CommandSearchCategory ()
 		{
+			// The AddinManager is not thread safe, so we need to make sure we're
+			// calling from the main thread.
+			Runtime.AssertMainThread ();
 			currentRuntimeAddin = Mono.Addins.AddinManager.CurrentAddin;
+
 			var hiddenCategory = GettextCatalog.GetString ("Hidden");
 			allCommands = IdeApp.CommandService.GetCommands ()
 			                    .Where (cmd => (cmd as ActionCommand)?.CommandArray != true && cmd.Category != hiddenCategory)
 			                    .Select(cmd => Tuple.Create (cmd, cmd.DisplayName))
-								.OrderByDescending(cmd => (cmd.Item1 as ActionCommand)?.RuntimeAddin == currentRuntimeAddin)
 			                    .ToList();
 		}
 
@@ -54,7 +58,7 @@ namespace MonoDevelop.Components.MainToolbar
 		{
 		}
 
-		string[] validTags = new [] { "cmd", "command", "c" };
+		readonly string[] validTags = { "cmd", "command", "c" };
 
 		public override string [] Tags {
 			get {
@@ -85,7 +89,7 @@ namespace MonoDevelop.Components.MainToolbar
 
 						if (matcher.CalcMatchRank (matchString, out var rank)) {
 							if ((cmd as ActionCommand)?.RuntimeAddin == currentRuntimeAddin)
-								rank += 100; // we prefer commands comming from the addin
+								rank += 1; // we prefer commands comming from the addin
 							searchResultCallback.ReportResult (new CommandResult (cmd, null, route, pattern.Pattern, matchString, rank));
 						}
 					}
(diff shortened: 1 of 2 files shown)


slluis:
I don't think this is right. This will make commands from the IDE to show at the top even when they are worse matches. I don't see an easy solution for this.

sandyarmstrong:
@slluis can't @avodovnik just change the commit to increase the rank by a smaller amount (say, 1 or 10) so that IDE commands are only favored when the relevancy is otherwise identical?

slluis:
We can try with 1 and see what happens

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/CommandSearchCategory.cs
```diff
@@ -26,31 +26,27 @@
...
-
+		private static readonly Mono.Addins.RuntimeAddin currentRuntimeAddin;
 		static CommandSearchCategory ()
 		{
+			currentRuntimeAddin = Mono.Addins.AddinManager.CurrentAddin;
 			var hiddenCategory = GettextCatalog.GetString ("Hidden");
 			allCommands = IdeApp.CommandService.GetCommands ()
 			                    .Where (cmd => (cmd as ActionCommand)?.CommandArray != true && cmd.Category != hiddenCategory)
 			                    .Select(cmd => Tuple.Create (cmd, cmd.DisplayName))
+								.OrderByDescending(cmd => (cmd.Item1 as ActionCommand)?.RuntimeAddin == currentRuntimeAddin)
```
Therzok: OrderByDescending on bool is weird :p

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/CommandSearchCategory.cs
```diff
@@ -86,10 +82,12 @@ public override Task GetResults (ISearchResultCallback searchResultCallback, Sea
 							break;
 						var cmd = cmdTuple.Item1;
 						var matchString = cmdTuple.Item2;
-						int rank;
 
-						if (matcher.CalcMatchRank (matchString, out rank))
+						if (matcher.CalcMatchRank (matchString, out var rank)) {
+							if ((cmd as ActionCommand)?.RuntimeAddin == currentRuntimeAddin)
```
Therzok: Didn't we already sort them in allCommands?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/CommandSearchCategory.cs
```diff
@@ -26,31 +26,27 @@
...
 	class CommandSearchCategory : SearchCategory
 	{
 		static readonly List<Tuple<Command, string>> allCommands;
-
+		private static readonly Mono.Addins.RuntimeAddin currentRuntimeAddin;
 		static CommandSearchCategory ()
 		{
+			currentRuntimeAddin = Mono.Addins.AddinManager.CurrentAddin;
 			var hiddenCategory = GettextCatalog.GetString ("Hidden");
 			allCommands = IdeApp.CommandService.GetCommands ()
```
Therzok: Unrelated, but this is cached once, and we don't listen to any changes on Extension Paths

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/CommandSearchCategory.cs
```diff
@@ -26,31 +26,27 @@
...
 namespace MonoDevelop.Components.MainToolbar
 {
 	class CommandSearchCategory : SearchCategory
 	{
 		static readonly List<Tuple<Command, string>> allCommands;
-
+		private static readonly Mono.Addins.RuntimeAddin currentRuntimeAddin;
 		static CommandSearchCategory ()
 		{
+			currentRuntimeAddin = Mono.Addins.AddinManager.CurrentAddin;
```
Therzok: The AddinManager is not thread safe. The static constructor can be called from any thread. If we keep this version, we need to assert we're on the Main Thread, otherwise weird stuff might happen.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/CommandSearchCategory.cs
avodovnik: Actually, this can probably be removed, since it's just a list.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/CommandSearchCategory.cs
avodovnik: We did, but we didn't need to (or shouldn't have, actually), since we **actually** sort them here, by rank. I've removed the other sort, and left this here.

avodovnik:
> We can try with 1 and see what happens

Sure, we can try 1. In the `Format Document` case that will work, as both get ranked equally, and the +1 will actually push ours to the top. We can evaluate others as we go along.

PR #8326, opened 2019-08-02
[Ide] Document.Select no longer calls GrabFocus.

mkrueger:
GrabFocus focuses the editor causing a selected sub view change. This is a side effect happening in VCS view commands.
Selecting a view programatically shouldn't automatically change sub view selections or focused controls.

Fixes VSTS #963585 – Diff/View commands broken in SolutonPad


sevoku:
we should make sure not to regress #7529

mkrueger:
#7529 works for me. OpenDocument now grabs focus only when a line/column is provided that basically queries a 'jump to that position' in that case the GrabFocus is correct.

mkrueger:
I think it is right. When open is called with line/column the editor should be selected IMO.

slluis:
A less intrusive fix: https://github.com/mono/monodevelop/pull/8571

PR #8329, opened 2019-08-02
Add .NET 4.7.2 target moniker, set it to default for templates and update NUnit Library project to use NUnit 3.12.0

manish:
- Bug 960218: .NET templates defaults to v4.7 instead of v4.7.2
- Bug 960219: NUnit Library Project should use NUnit 3.12 instead of 2.6.x

Fixes https://dev.azure.com/devdiv/DevDiv/_workitems/edit/960218
Fixes https://dev.azure.com/devdiv/DevDiv/_workitems/edit/960219

### DIFF
--- a/main/src/addins/MonoDevelop.UnitTesting.NUnit/templates/NUnitProjectVBNet.xpt.xml
+++ b/main/src/addins/MonoDevelop.UnitTesting.NUnit/templates/NUnitProjectVBNet.xpt.xml
@@ -32,7 +32,7 @@
 			</References>
 
 			<Packages>
-				<Package ID="NUnit" Version="2.6.4" />
+				<Package ID="NUnit" Version="3.12.0" />
 			</Packages>
 
 			<Files>
--- a/main/src/addins/MonoDevelop.UnitTesting.NUnit/templates/NUnitProject.xpt.xml
+++ b/main/src/addins/MonoDevelop.UnitTesting.NUnit/templates/NUnitProject.xpt.xml
@@ -32,7 +32,7 @@
 			</References>
 
 			<Packages>
-				<Package ID="NUnit" Version="2.6.4" />
+				<Package ID="NUnit" Version="3.12.0" />
 			</Packages>
 
 			<Files>
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/TargetFrameworkMoniker.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/TargetFrameworkMoniker.cs
@@ -333,6 +333,10 @@ namespace MonoDevelop.Core.Assemblies
 			get { return new TargetFrameworkMoniker ("4.7.1"); }
 		}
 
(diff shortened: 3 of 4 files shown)


main/src/addins/MonoDevelop.UnitTesting.NUnit/templates/NUnitProjectVBNet.xpt.xml
```diff
@@ -32,7 +32,7 @@
 			</References>
 
 			<Packages>
-				<Package ID="NUnit" Version="2.6.4" />
+				<Package ID="NUnit" Version="3.12.0" />
```
mrward: Did you try a project that uses NUnit 3.12.0 in VS Mac? VS Mac does not seem to show any tests in the Unit Tests window when NUnit 3.12.0 is used.

manish: We use 3.12 in ui testing and can run it from test pad
https://github.com/xamarin/md-addins/blob/master/Tests/Microsoft.VSForMac.Tests/Microsoft.VSForMac.Tests.csproj#L141

mrward: Does not work for me in VS Mac 8.2 nor in an 8.4 build. Note that the VSForMac.Tests.cspro is using a PackageReference - whether that makes any difference. I just created a new NUnit library project and upgraded the NUnit NuGet package.

mrward: I think there is some odd behaviour on updating the NuGet package since the nunit.framework is maybe also available in the GAC with Mono?

mrward (approved):

main/src/addins/MonoDevelop.UnitTesting.NUnit/templates/NUnitProjectVBNet.xpt.xml
manish: Weird. It has always worked even when using packages.config 
https://github.com/xamarin/QualityAssurance/blob/master/Automation/XQA/XQA.XS.Tests/XQA.XS.Tests.csproj#L54

mrward: Problem seems to be updating the package. The HintPath is not updated for some reason.

manish: I saw that too. I had to manually change the HintPath for it to work. Let's see if it works when we don't have to update, but just restore

nosami (approved):

manish:
image

cc: @mrward

Therzok (approved):

slluis:
@manish please use proper bug fix messages, one for each bug fixed.

PR #8333, opened 2019-08-02
 TextEditor: "Use Selection for Find" fixes

sandyarmstrong:
1. Rename "Use Selection for *" commands to have proper macOS labels
2. Move cmd+e keybinding from `FindNextSelection` command to
   `UseSelectionForFind`.
3. Add `UseSelectionForFind` command to main menu.
4. Add support for missing find commands to new editor.

It's unfortunate that we have two similar commands here but it's
important that cmd+e behaves as the user expects (it should not have
"find next" behavior...it just sets search text and keeps focus in
editor).

Fixes https://dev.azure.com/devdiv/DevDiv/_workitems/edit/947125


Therzok:
@sevoku anything we need to change in [KeyBindingScheme](https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/options/KeyBindingSchemeVisualStudio-mac.xml) ?

sandyarmstrong:
@Therzok I didn't think to check there, but no, it looks fine.

PR #8336, opened 2019-08-05
[MacOS] Specify dlopen library paths

iainx:
With the hardened runtime, we now need to specify the paths of all libraries
that are dlopened in the binary's rpath

### DIFF
--- a/main/build/MacOSX/Makefile.am
+++ b/main/build/MacOSX/Makefile.am
@@ -27,6 +27,10 @@ EXTRA_DIST = dmg-bg.png DS_Store Info.plist.in make-dmg-bundle.sh render.cs
 MONOSTUB_EXTRA_SOURCEFILES = monostub-utils.h
 export MACOSX_DEPLOYMENT_TARGET=10.12
 
+# With the hardened runtime, we need to specify the location of all libraries
+# that we dlopen
+MONOSTUB_RPATH=-Wl,-rpath,/Library/Frameworks/Mono.framework/Libraries/ -Wl,-rpath,@executable_path/../Resources/lib/monodevelop/bin/
+
 all: monostub monostub-nogui monostub-test
 
 render.exe: render.cs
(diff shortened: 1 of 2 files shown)


main/build/MacOSX/Makefile.am
```diff
@@ -27,6 +27,10 @@ EXTRA_DIST = dmg-bg.png DS_Store Info.plist.in make-dmg-bundle.sh render.cs
 MONOSTUB_EXTRA_SOURCEFILES = monostub-utils.h
 export MACOSX_DEPLOYMENT_TARGET=10.12
 
+# With the hardened runtime, we need to specify the location of all libraries
+# that we dlopen
+MONOSTUB_RPATH=-Wl,-rpath,/Library/Frameworks/Mono.framework/Libraries/ -Wl,-rpath,@executable_path/../Resources/lib/monodevelop/bin/
```
Therzok: I guess we need all the paths we have [here](https://github.com/mono/monodevelop/blob/c019361d83a2820ef0f8a3f3911070595bdb976c/main/build/MacOSX/monostub-utils.h#L53) to be included in this list.

And to be sure, this bit of code removed? https://github.com/mono/monodevelop/blob/c019361d83a2820ef0f8a3f3911070595bdb976c/main/build/MacOSX/monostub-utils.h#L151

Therzok: As a result, we might also need to modify [the test](https://github.com/mono/monodevelop/blob/c019361d83a2820ef0f8a3f3911070595bdb976c/main/build/MacOSX/monostub-test.m#L160)

iainx: I think we only need the non-standard library paths, so we don't need the /usr ones. Are we dlopening libraries from the bin dir or the Xcode library?

iainx: (also, we shouldn't really have binaries in the Resources/ folder, because that's just for data. It leads to an amusing situation where ThirdPartyNotices.txt and COPYING.LGPL need to be signed, because they're data inside the binary folder, and libxammac.dylib not needing signed because it's in the data folder)

Therzok: @iainx we dlopen libsvn from Xcode, we have libgit2 being dlopen'd from the `addins/VersionControl` dir.

Most likely some other stuff will be impacted.

iainx: (and whatever decision/changes are made here, will probably need to be made in the md-addins version of the Makefile too)

iainx: @Therzok ok, I'll get those added then. It might be worthwhile moving all our dlopened libraries into one standard folder (underneath the `Contents/MacOS` path, to make sure they get signed)

PR #8338, opened 2019-08-05
[Ide] Run all WelcomePageService operations on main UI thread

sevoku:
Fixes VSTS #957299

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageService.cs
@@ -110,17 +110,17 @@ namespace MonoDevelop.Ide.WelcomePage
 
 			// Try to get a dialog version of the "welcome screen" first
 			if (!await ShowWelcomeWindow (options)) {
-				ShowWelcomePage (true);
+				await Runtime.RunInMainThread (() => ShowWelcomePage (true));
 			}
 		}
 
 		public static async void HideWelcomePageOrWindow ()
 		{
 			if (WelcomeWindowProvider != null) {
-				await WelcomeWindowProvider.HideWindow ();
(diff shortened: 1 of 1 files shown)


sevoku:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.WelcomePage/WelcomePageService.cs
```diff
@@ -102,60 +102,64 @@ internal static async Task Initialize (bool hideWelcomePage)
 
 		public static bool HasWindowImplementation => WelcomeWindowProvider != null;
 
-		public static async void ShowWelcomePageOrWindow (WelcomeWindowShowOptions options = null)
+		public static async Task ShowWelcomePageOrWindow (WelcomeWindowShowOptions options = null)
 		{
```
mrward: Public API break? The WelcomePageService class is public so these Task changes are changing the public API.

sevoku: Yes, that was intended, we have changed that several times already, this should be the final thing here. But I actually wanted to split the PR, will do that asap.

sevoku: moved to #8345

mrward:
The ShowWelcomePage and HideWelcomePage methods are public and could be run from a background thread. Should we add a Runtime.AssertMainThread in those methods?

Should the events fired here be fired only on the UI thread? It looks like WelcomeWindowShown and WelcomeWindowHidden could be fired on a background thread. For IDE events we normally only fire those on the UI thread.

sevoku:
done @mrward 👍

sevoku:

sevoku:

sevoku:

PR #8340, opened 2019-08-06
Fixes VSTS Bug 958221: System.ArgumentNullException exception in

mkrueger:
System.Linq.Enumerable.OfType[DocumentControllerExtension]()

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/958221

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ExtensionChain.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ExtensionChain.cs
@@ -68,7 +68,7 @@ namespace MonoDevelop.Projects
 
 		public IEnumerable<ChainedExtension> GetAllExtensions ()
 		{
-			return extensions;
+			return extensions ?? Enumerable.Empty<ChainedExtension> ();
 		}
 
 		public IDisposable BatchModify () => batchModifier = new BatchModifier (this);
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Document.cs
@@ -130,16 +130,22 @@ namespace MonoDevelop.Ide.Gui
 
 		public T GetContent<T> (bool forActiveView) where T : class
 		{
+			if (IsDisposed)
+				return null;
 			return (T)GetContent (forActiveView, typeof (T));
 		}
 
 		public T GetContent<T> () where T : class
 		{
+			if (IsDisposed)
+				return null;
 			return GetContent<T> (false);
 		}
 
 		public IEnumerable<T> GetContents<T> (bool forActiveView) where T : class
 		{
+			if (IsDisposed)
+				return Enumerable.Empty<T> ();
 			if (forActiveView)
 				return view.GetActiveControllerHierarchy ().Select (controller => controller.GetContents<T> ()).SelectMany (c => c);
 			return GetControllersForContentCheck ().Select (controller => controller.GetContents<T> ()).SelectMany (c => c);
@@ -147,6 +153,8 @@ namespace MonoDevelop.Ide.Gui
 
 		public IEnumerable<T> GetContents<T> () where T : class
 		{
+			if (IsDisposed)
+				return Enumerable.Empty<T> ();
 			return GetContents<T> (false);
 		}
 
@@ -637,8 +645,11 @@ namespace MonoDevelop.Ide.Gui
 			Close ().Ignore ();
 		}
 
+		public bool IsDisposed { get; private set; }
+
 		void Dispose ()
 		{
+			IsDisposed = true;
 			DocumentRegistry.Remove (this);
 			callbackRegistration?.Dispose ();
 			var workspace = Runtime.PeekService<RootWorkspace> ();


Therzok (approved):

mrward (approved):
If GetAllExtensions can never be null then I guess this code can be changed:

https://github.com/mono/monodevelop/blob/d167b8289b3c9149932b1726bf80e5aafab5d830/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs#L675

slluis (changes requested):
`extensions` can only be null if the extension chain has been disposed, which happens when the document controller is disposed. If we are trying to access a disposed object, the bug is somewhere else.

slluis:
I'm wondering if MonoDevelopDocumentTrackingService is keeping a reference to a disposed document.

Therzok:
@slluis there's some other traces, not just MonoDevelopDocumentTrackingService:

```
System.Linq.Enumerable.OfType[DocumentControllerExtension](IEnumerable)
MonoDevelop.Ide.Gui.Documents.DocumentController.<GetContents>d__XXX.MoveNext()
System.Linq.Enumerable.TryGetFirst[Object](IEnumerable`1,Boolean&)
System.Linq.Enumerable.FirstOrDefault[Object](IEnumerable`1)
MonoDevelop.Ide.Gui.Documents.DocumentController.GetContent(Type)
MonoDevelop.Ide.Gui.Document.<>c__XXX.<GetContentIncludingAllViews>b__XXX(DocumentController)
System.Linq.Enumerable.SelectEnumerableIterator`2[[MonoDevelop.Ide.Gui.Documents.DocumentController, MonoDevelop.Ide, Version=2.6.0.0, Culture=neutral, PublicKeyToken=3ead7498f347467b],[System.Object, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]].MoveNext()
System.Linq.Enumerable.TryGetFirst[Object](IEnumerable`1,Func`2,Boolean&)
System.Linq.Enumerable.FirstOrDefault[Object](IEnumerable`1,Func`2)
MonoDevelop.Ide.Gui.Document.GetContentIncludingAllViews(Type)
MonoDevelop.Ide.Gui.Document.GetContent(Boolean,Type)
MonoDevelop.Ide.Gui.Document.GetContent[TextEditor](Boolean)
MonoDevelop.Ide.Gui.Document.GetContent[TextEditor]()
MonoDevelop.Ide.Gui.Document.get_Editor()
<addin>.ActiveDocumentChanged(Object,EventArgs)
System.CoreExtensions.SafeInvoke[DocumentEventArgs](EventHandler`1,Object,DocumentEventArgs)
```

slluis:
@Therzok yes, there may be more than one problem.

This is not right though:

https://github.com/mono/monodevelop/blob/50fbe0a7e65c5439e3313c6b50e7ef927f5f1fe9/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices/MonoDevelopDocumentTrackingService.cs#L77-L78

`activeDocument` should be assigned before rising the event. 

In any case, caching the active document here is a potential problem. It may happen that `TryGetActiveDocument` is called before  `MonoDevelopDocumentTrackingService` gets the `ActiveDocumentChanged` event, in which case it would return a stale document.

slluis:
Maybe `IdeApp.Workbench.ActiveDocument` can return a disposed document when the document is closed? worth checking different scenarios.

Therzok:
Looks like the code was changed compared to my [initial caching commit](https://github.com/mono/monodevelop/commit/65e918570a43773e189ef12f1f657204ffdd4642#diff-33b21cec6f657d2acae35005f9e1ea14). This commit [changed it to be assigned after](https://github.com/mono/monodevelop/commit/e466141ef7b2c39585e29f94c28f6deb82e8be97#diff-33b21cec6f657d2acae35005f9e1ea14)

mkrueger:
/fixed the activeDocument assignment - that was wrong.

Y there are different scenarios where that can happen.

mkrueger:
How about that ? Just giving cutting the GetContent  pipe on disposed Documents ?

Therzok (approved):

sevoku (approved):

sevoku:
@mkrueger tests are failing now

slluis:
My question still stands. Why is GetContent() being called on a disposed document?

Therzok:
Honestly, I think this is a fallout from asynchronizing all the document code. Given we can have a situation where:
```csharp
var doc = await GetDocument();
await SomethingElse(); // User closes the document in the meantime
...
// This is now disposed
DoSomething (doc);
```

and the user closes the document in the meantime, it can

slluis:
@Therzok we have an event that is raised when a document is closed, so that case can be avoided. I don't agree this is a problem of asynchronizing code.

Therzok:
Okay, I think I understand now what's going on, at least for the roslyn stack.

1. The order of assignment did cause most of the bugs.
2. SolutionCrawler runs on a background thread, so it might race with the UI thread disposing the document.

Therzok:
We also should probably remove the activeDocument from the list when it's closed.

mkrueger:
No idea what you mean with remove active document from the list. 
At that point it's no longer an IDE document at that point it's a roslyn document. And does it really make sense if a document gets closed that it's removed from the crawler ? 

Just because the IDE document is closed it's not affecting it's contents. Interesting would just be closing a document with changes and discard the changes on close.

Therzok:
@mkrueger We need to modify this test and ensure it passes:
https://github.com/mono/monodevelop/blob/6f21173f7286daeb64934b4dec9af2a03831dea9/main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/ExtensionChainTests.cs#L137

Therzok:
Can we split this PR? The roslyn active doc fix is good to merge. Stalling this too long is not really a good idea...

mkrueger:
In general I don't think that this change is harmful. Even if there is a deeper problem it makes the code more solid.

Therzok:
I still think Lluis is right, in that it hides a problem, not fixes it.

PR #8342, opened 2019-08-06
[Ide] Fix unobserved task exception in CustomToolService

mrward:
If an exception is thrown by a custom tool then this may result in
an unobserved task exception. The exception is set on a task
completion source whose task is only observed if a build is waiting
for custom tools to complete or another custom tool is run again for
the same file.

There does not seem to be any need to set the exception on the task
completion source since all errors are reported in the IDE and
callers of the CustomToolService Update or UpdateAsync do not get
access to this task. Now the task completion source has its result
set instead of setting the exception since the only thing needed is
to ensure the task is completed.

Fixes VSTS #958223 - System.Threading.Tasks.TaskCanceledException exception
in System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess()


mrward:

mrward:

mrward:
.NET Core failing tests are due the known problem with Mono 6.4.0.148.

PR #8349, opened 2019-08-07
[Mac] Add some more entitlements

iainx:
Add entitlements to allow MonoDevelop to use AppleScript and debug apps


iainx:
apparently we may not need the debugger one, because mono debugging doesn't use mach ports. Going to remove it for now

PR #8353, opened 2019-08-07
[Core] Cache ProjectFile.Include to improve project save performance

mrward:
With an SDK style project containing 1500 C# files, adding a new
C# class to the project would result in the saving taking about 20
seconds. The majority of this time was spent in the Project's
SaveProjectItems as it tried to determine if an MSBuild remove item
should be added for the new file. Caching the ProjectFile.Include
takes the save time down to around 300 ms.

Fixes VSTS #947103 - Class creation is very slow in projects with
hundreds of classes

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectFile.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectFile.cs
@@ -75,14 +75,22 @@ namespace MonoDevelop.Projects
 			BuildAction = buildAction;
 		}
 
+		string cachedInclude;
+
 		public override string Include {
 			get {
 				if (Project != null) {
+					if (cachedInclude != null)
+						return cachedInclude;
+
 					string path = MSBuildProjectService.ToMSBuildPath (Project.ItemDirectory, FilePath);
 					if (path.Length > 0) {
 						//directory paths must end with '/'
 						if ((Subtype == Subtype.Directory) && path [path.Length - 1] != '\\')
 							path = path + "\\";
+						// Cache the include path to avoid recalculating MSBuildProjectService.ToMSBuildPath
+						// which can slow down saving SDK style projects that contain thousands of files.
+						cachedInclude = path;
 						return path;
 					}
 				}
@@ -188,6 +196,8 @@ namespace MonoDevelop.Projects
 				// If the file is a link, rename the link too
 				if (IsLink && Link.FileName == oldPath.FileName)
 					link = Path.Combine (Path.GetDirectoryName (link), filename.FileName);
+
+				cachedInclude = null;
 
 				// If a file that belongs to a project is being renamed, update the value of UnevaluatedInclude
 				// since that is used when saving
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectFile.cs
```diff
@@ -485,12 +495,31 @@ internal bool ResolveParent ()
...
 			base.OnProjectSet ();
 			if (Project != null) {
 				base.Include = Include;
+
+				project = Project;
+				project.Modified += OnProjectModified;
+
 				VirtualPathChanged?.Invoke (this, new ProjectFileVirtualPathChangedEventArgs (this, FilePath.Null, ProjectVirtualPath));
+			} else {
+				if (project != null)
```
Therzok: We should unsubscribe this event handler always, not just on this code-path, to prevent leaking the project when `Project` is set multiple times.

Therzok: Also, since ProjectItem does not compare the project instances being set, even setting the same value twice will subscribe `OnProjectModified` _twice_ to the project.

mrward: Good catch 👍

main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/DotNetCoreProjectTests.cs
```diff
@@ -696,6 +696,44 @@ public async Task LoadProject_UnknownNuGetSDKPackage_SDKResolutionErrorsReported
...
+				p.AddFile (fileName, BuildAction.Compile);
+
+				var timer = Stopwatch.StartNew ();
+				await p.SaveAsync (Util.GetMonitor ());
+				timer.Stop ();
+
+				// This was taking 20 seconds before.
+				// Takes about 300ms with ProjectFile.Include caching. Here we use 2 seconds
+				// in case the build server is slow.
+				Assert.That (timer.ElapsedMilliseconds, Is.LessThan (2000));
```
Therzok: Should we add a new TimerCounter for `SaveAsync` and use that information here?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectFile.cs
```diff
@@ -485,15 +495,31 @@ internal bool ResolveParent ()
 			}
 		}
 
+		Project project;
+
 		protected override void OnProjectSet ()
 		{
 			base.OnProjectSet ();
+			if (project != null)
+				project.Modified -= OnProjectModified;
```
Therzok:
```csharp
if (project != null) {
    project.Modified -= OnProjectModified;
    project = null;
}
```

mrward: 😄

slluis:
An idea to simplify the code and avoid too many handlers for the `Project.Modified` event: keep a reference of the `Project.ItemDirectory` value that was used to create `cachedInclude` in ProjectFile, and return `cachedInclude` only if  `Project.ItemDirectory` has not changed.

mrward:
I originally tried keeping the original value of the Project.ItemDirectory (both as a FilePath and as a string) and comparing the new value with the old value, to see if it had changed, but that did not seem to fix the performance problem enough. Doing that dropped the save timing down to about 7 or 8 seconds. So I switched to using the Project.Modified event for better performance.

slluis:
Wow, that means this property is invoked a huge amount of times. It is a perfect candidate for a performance test.

slluis (approved):

Therzok (approved):

slluis:

PR #8354, opened 2019-08-07
Fixes VSTS Bug 961599: [Feedback] Visual Studio crashes/exists when

mkrueger:
attempting to commit to Git

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/961599

MessageService was causing a stack Overflow it could add a dialog as
child to itself. However teh git commit dialog should set itself as
parent for the git config dialog.

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/CommitDialogExtension.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/CommitDialogExtension.cs
@@ -6,8 +6,9 @@ namespace MonoDevelop.VersionControl
 	/// <summary>
 	/// Base class for commit dialog extensions.
 	/// </summary>
-	public class CommitDialogExtension: Gtk.EventBox
+	public class CommitDialogExtension : Gtk.EventBox
 	{
+		internal VersionControl.Dialogs.CommitDialog CommitDialog  { get; set; }
 		/// <summary>
 		/// Initialize the extension.
 		/// </summary>
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
@@ -395,6 +395,8 @@ namespace MonoDevelop.Ide
 			NSWindow nativeParent = null;
 			try {
 				nativeParent = parent;
+				if (nativeParent.Handle == nsdialog.Handle)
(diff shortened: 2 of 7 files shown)


mkrueger:

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/CommitDialogExtension.cs
```diff
@@ -6,8 +6,9 @@ namespace MonoDevelop.VersionControl
 	/// <summary>
 	/// Base class for commit dialog extensions.
 	/// </summary>
-	public class CommitDialogExtension: Gtk.EventBox
+	public class CommitDialogExtension : Gtk.EventBox
 	{
+		public VersionControl.Dialogs.CommitDialog { get; set; }
```
Therzok: Why public?

mkrueger:

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/CommitDialogExtension.cs
mkrueger: Was too late yesterday - I changed it to internal. Needed to access that in another project.

But y we shouldn't change the public API. IVT all the way :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/MessageService.cs
```diff
@@ -330,7 +330,7 @@ public static int RunCustomDialog (Dialog dlg, Window parent)
 				if (dialog.TransientFor != null)
 					parent = dialog.TransientFor;
 				else
-					parent = IdeServices.DesktopService.GetFocusedTopLevelWindow ();
```
sevoku: This is not right, `GetFocusedTopLevelWindow ()` has an additional check to not parent on pads, while `GetParentForModalWindow ()` just uses the top window (which might be also a pad) which is not necessarily focused. Actually we should deprecate `GetParentForModalWindow ()`, but I need to review that ...

sevoku: The right fix would be to check that `parent != dialog`, if the window is the same one throw an InvalidOperationException

mkrueger: fixed.

slluis:

sevoku:

sevoku:

PR #8357, opened 2019-08-07
TextEditor: Allow opening C# files with certain custom themes

sandyarmstrong:
Fixes https://dev.azure.com/devdiv/DevDiv/_workitems/edit/962445


mrward:
It looks to me that SyntaxHighlightingService.GetEditorTheme will not return null. If the name cannot be found it falls back to using a default theme name. If that cannot be found then it would get stuck in an infinite loop.

https://github.com/mono/monodevelop/blob/f9839cdd26fd7710315b7ad7fcb7bc7b585223c9/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/SyntaxHighlightingService.cs#L179-L194

sandyarmstrong:
Yeah, `setting` wasn't `null` in any of my testing, I was just being overly defensive. Will change, thanks!

PR #8359, opened 2019-08-08
Fixes VSTS Bug 960629: CommandSearchCategory potentially works on inv…

mkrueger:
…alid data

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/960629

Removed the cache - that should now run on the main thread. But there are too few commands to justify a task there in any case.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/Command.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/Command.cs
@@ -40,6 +40,7 @@ namespace MonoDevelop.Components.Commands
 		object id;
 		string text;
 		string displayName;
+		string cachedDisplayName;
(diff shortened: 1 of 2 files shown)


Therzok:
What I'm asking for is [change the replace call from get to set](http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Components.Commands/Command.cs,138)

mkrueger:
Commands should cache DisplayName if it's a problem. But in any case - how many commands we've ? Creating ~100 strings is in no way a problem for a computer program.

Therzok:
@mkrueger we don't need both displayName and cachedDisplayName.

Therzok:
using `mdtool extension-tool` and going to `/MonoDevelop/Ide/Commands`, I see 503 commands.

sevoku:

PR #8360, opened 2019-08-08
Fixes VSTS Bug 959146: System.NullReferenceException exception in

mkrueger:
Mono.TextEditor.Utils.TextBreaker.BreakLinesIntoWords()

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/959146

I can't find the cause of it. I checked the parameters of
BreakLinesIntoWords now. The diff must be out of sync with the text
document that should now not happen, it's checked before paint that
the versions match. In any case that implementation is more solid.


mkrueger:

sevoku:

PR #8362, opened 2019-08-08
[NuGet] Fix current version label in Manage NuGet packages dialog

mrward:
The Updates tab would show 'Multiple' for the current version when the
same package version was installed in multiple projects.

Fixes VSTS #963252 - Manage NuGet packages showing multiple current
versions when only one is installed


mrward:

PR #8372, opened 2019-08-09
[Debugger] Protect against null document text buffers

jstedfast:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/953602

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
@@ -1505,13 +1505,21 @@ namespace MonoDevelop.Debugger
 		{
 			var doc = IdeApp.Workbench.GetDocument (document.FilePath);
 			IBreakpointSpanResolver resolver = null;
+			ITextBuffer buffer;
 
-			if (doc != null)
+			if (doc != null) {
 				resolver = doc.GetContent<IBreakpointSpanResolver> ();
+				buffer = doc.TextBuffer;
+			} else {
+				buffer = document.TextBuffer;
+			}
+
+			if (buffer == null)
+				return Task.FromResult (default (Span));
 
 			resolver = resolver ?? new DefaultBreakpointSpanResolver ();
 
-			return resolver.GetBreakpointSpanAsync (document.TextBuffer, position, cancellationToken);
+			return resolver.GetBreakpointSpanAsync (buffer, position, cancellationToken);
 		}
 	}
 


jstedfast:
@sandyarmstrong @DavidKarlas why is the `document.TextBuffer == null`? Is that ever valid? And if not, why is it null?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
```diff
@@ -1504,12 +1504,11 @@ public static async Task<CompletionData> GetCompletionDataAsync (StackFrame fram
 		public static Task<Span> GetBreakpointSpanAsync (ITextDocument document, int position, CancellationToken cancellationToken = default (CancellationToken))
 		{
 			var doc = IdeApp.Workbench.GetDocument (document.FilePath);
-			IBreakpointSpanResolver resolver = null;
 
-			if (doc != null)
-				resolver = doc.GetContent<IBreakpointSpanResolver> ();
+			if (doc == null || doc.TextBuffer == null)
+				return Task.FromResult (default (Span));
```
sandyarmstrong: I don't think this fix is correct. Shouldn't the only necessary change be to return early if `document.TextBuffer` (NOT `doc.TextBuffer`) is `null`?

If I understand correctly, `document` can be from old or new editor, but `doc` will always be `null` with new editor.

I don't know what could be causing `document.TextBuffer` to be `null` in the first place, though.

jstedfast: oh, wow, you are right... I missed doc vs document

jstedfast: I've modified the code to use `doc.TextBuffer` if `doc != null` and to use `document.TextBuffer` otherwise.

Maybe that was the issue? I have no idea how to reproduce this situation so I have no way of testing :-\

sandyarmstrong: I honestly have no idea if this is correct. My assumption would be that `doc.TextBuffer` and `document.TextBuffer` should be identical, so I'm not sure if the extra change here makes a difference? Maybe @DavidKarlas knows.

sandyarmstrong: Oh, wait. 🤦‍♂

`doc` won't be always be `null` with new editor. I was thinking of something else. I am sorry for misleading you.

jstedfast: I think it should be fixed now btw

Therzok:
LGTM, can we get this green?

DavidKarlas:
Looked at this bug and PR for past hour... My conclusion is... this TextBuffer should never be null, it gets null when Document is Disposed but in that case `IdeApp.Workbench.GetDocument (document.FilePath);` would return `null` as well... So imo, just another Mono GC screw up...

Therzok:
How would mono be at fault here? I think this is another use-after-dispose due to async problem. There is another PR mike opened that changes some behaviour WRT the active document passed to roslyn

Therzok:
Hum, there's no async code here. Hmmm.

PR #8374, opened 2019-08-09
[8.3][Debugger] Protect against null document text buffers

jstedfast:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/953602


jstedfast:
This is a back-port of https://github.com/mono/monodevelop/pull/8372

PR #8378, opened 2019-08-10
[Ide] Show welcome page after preseting workbench

sevoku:
When Workbench.Present() is executed for the first time
it will set up the layout, hiding the welcome page.


Therzok:
@sevoku what's the impact of this change?

sevoku:
@Therzok this fixes the MonoDevelop welcome page, Present () sets up the layout and hides the welcome page when doing that. This change makes sure that the main IDE window is properly initialized before showing the welcome page overlay.

Therzok:
LGTM iteration 2

PR #8379, opened 2019-08-10
[DotNetCore] Add custom .NET Core sdk package sources

mrward:
See if this fixes the project template tests.

Temporary change - should not be needed when the packages
are published on nuget.org.


rodrmoya:

PR #8380, opened 2019-08-12
[Ide] Don't log an error if getting last write time for new documents fails

sevoku:
A new - not yet saved - document can't have a last write time. Although it's not a crasher, we should not even try to get the write time if we know that the file doesn't exist.


slluis:

PR #8386, opened 2019-08-12
Fixes VSTS Bug 958249: System.InvalidOperationException exception in

mkrueger:
MonoDevelop.Ide.Gui.Components.LogViewProgressMonitor.OnWriteLog()

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/958249

OnWriteLog is called on a disposed LogViewProgressMonitor. The removal
of the event handler would fix that exception. But it's better to
check for disposal on the context.Post calls. From the exception text
that is the cause and there are more cases where a disposal check
makes sense.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.ProgressMonitoring/LogTextWriter.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.ProgressMonitoring/LogTextWriter.cs
@@ -76,7 +76,11 @@ namespace MonoDevelop.Core.ProgressMonitoring
 		public override void Close ()
 		{
 			if (context != null)
-				context.Post ((o) => ((LogTextWriter)o).OnClosed (), this);
+				context.Post ((o) => {
+					if (((LogTextWriter)o).IsDisposed)
+						return;
+					((LogTextWriter)o).OnClosed ();
+				}, this);
 			else
 				OnClosed ();
 		}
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs
```diff
@@ -981,6 +981,8 @@ static Exception GetDisposedException ()
 		protected override void OnDispose (bool disposing)
 		{
 			base.OnDispose (disposing);
+			internalLogger.TextWritten -= WriteConsoleLogText;
```
Therzok: Do this before calling base.OnDispose?

main/src/core/MonoDevelop.Core/MonoDevelop.Core.ProgressMonitoring/LogTextWriter.cs
```diff
@@ -89,7 +93,11 @@ void OnClosed ()
 		public override void Write (char[] buffer, int index, int count)
 		{
 			if (context != null)
-				context.Post ((o) => ((LogTextWriter)o).OnWrite (buffer, index, count), this);
+				context.Post ((o) => {
+					if (((LogTextWriter)o).IsDisposed)
```
Therzok: Just do the Dispose check in `OnWrite`?

mkrueger: Possible, but you should never call something on a disposed object.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs
mkrueger: Doesn't matter that much in that case. There is more cleanup code after that. Splitting up clean up code is bad style.

Therzok: We should unregister this before calling Base.OnDispose.

Therzok (approved):

Therzok:
Let's get this green.

mkrueger:

sevoku (approved):

sevoku:

PR #8387, opened 2019-08-12
Fixes VSTS Bug 958247: System.NullReferenceException exception in

mkrueger:
MonoDevelop.AssemblyBrowser.AssemblyBrowserWidget.<>c__XXX.<EnsureDefinitionsLoaded>b__XXX()

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/958247

It could be that this exception didn't happen anymore however it's
hard to tell - this is basically the last thing that potentially could
be null in that method.

### DIFF
--- a/main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserWidget.cs
+++ b/main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserWidget.cs
@@ -1263,19 +1263,24 @@ namespace MonoDevelop.AssemblyBrowser
 		{
 			if (ensuredDefinitions == null)
 				throw new ArgumentNullException (nameof (ensuredDefinitions));
-			foreach (var def in ensuredDefinitions) {
-				if (!definitions.Contains (def)) {
-					definitions = definitions.Add (def);
 
-					Application.Invoke ((o, args) => {
-						if (ensuredDefinitions.Count + projects.Count == 1) {
-							TreeView.LoadTree (def.LoadingTask.Result);
-						} else {
-							TreeView.AddChild (def.LoadingTask.Result);
+			Runtime.RunInMainThread (() => {
+				foreach (var def in ensuredDefinitions) {
+					try {
+						if (!definitions.Contains (def)) {
+							definitions = definitions.Add (def);
+							var peFile = def.Assembly;
+							if (ensuredDefinitions.Count + projects.Count == 1) {
+								TreeView.LoadTree (peFile);
+							} else {
+								TreeView.AddChild (peFile);
+							}
 						}
-					});
+					} catch (Exception e) {
+						LoggingService.LogError ($"Can't load definition {def}. File name: {def?.FileName}", e);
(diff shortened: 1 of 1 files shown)


mkrueger:

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserWidget.cs
```diff
@@ -1264,6 +1264,10 @@ internal void EnsureDefinitionsLoaded (ImmutableList<AssemblyLoader> ensuredDefi
 			if (ensuredDefinitions == null)
 				throw new ArgumentNullException (nameof (ensuredDefinitions));
 			foreach (var def in ensuredDefinitions) {
+				if (def == null || def.LoadingTask == null) {
+					LoggingService.LogWarning ($"AssemblyLoader {def} had no loading task. File name: {def.FileName}");
```
bratsche: Can't `"{def.FileName}"` generate a NRE here?

Therzok: Good catch!

Therzok: @mkrueger basically what we're doing here is killing reporting this crash in case it happens.

So, we either need to figure out if the problem is actually fixed and close the bug, otherwise deploy a fix, logging a warning like this won't help much, as we don't get this information via `LogWarning`

mkrueger: I don't think that this is a real bug anymore.

Therzok: Yes, but we don't have access to that data anymore. I'd keep the error as-is and observe if the exception happens in newer releases.

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserWidget.cs
```diff
@@ -1263,19 +1263,24 @@ internal void EnsureDefinitionsLoaded (ImmutableList<AssemblyLoader> ensuredDefi
...
+							var peFile = def.Assembly;
+							if (ensuredDefinitions.Count + projects.Count == 1) {
+								TreeView.LoadTree (peFile);
+							} else {
+								TreeView.AddChild (peFile);
+							}
 						}
-					});
+					} catch (Exception e) {
+						LoggingService.LogError ($"Can't load definition {def}. File name: {def?.FileName}", e);
```
Therzok:
```suggestion
						LoggingService.LogInternalError ($"Can't load definition {def}. File name: {def?.FileName}", e);
```

sevoku:

PR #8389, opened 2019-08-12
[NuGet] Fix failing test due to change in .NET Core 8.0 SDK preview 8

mrward:
The NetCoreAndPackageReferenceProjectWithLibLogPackage test was
failing when checking the generated source files from the LibLog
NuGet package were available for .NET Standard projects. With .NET
Core SDK 3.0 preview 8 the RunProduceContentAssets now has a
condition that checks to see if there are content files to process.

    <Target Name="RunProduceContentAssets"
          DependsOnTargets="ResolvePackageAssets"
          Condition="'@(_ContentFilesToPreprocess)' != ''">

These content files are generated by ResolvePackageAssets but since
the condition is false it is not called. To fix this when getting
the source files for the type system, via the CoreCompileDependsOn
property value, the NuGet addin now explicitly runs the
ResolvePackageAssets target before the RunProduceContentAssets
target. This ensures there are content files to be processed.
The ResolvePackageAssets target exists in .NET Core 2.1 so this does
not cause a problem if only the older SDK is installed.


mrward:

PR #8395, opened 2019-08-12
[IDE] [a11y] Rewrite new configuration dialog in Xwt

bratsche:
This is to fix a11y issues with the combo entries, but it's a small
enough dialog to just rewrite from Gtk to Xwt now.

Fixes VSTS #753601


bratsche:
One thing that's kind of confusing is `TextEntry_Changed()`. I must be overlooking something really obvious because this makes no sense to me. If I don't add this handler to `comboName.TextEntry.Changed` then there is no text in `comboName.TextEntry`. But I'm not doing that for `comboPlatform.TextEntry` and it has a value. If anyone sees something obvious let me know so I can fix it. :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewConfigurationDialog.cs
```diff
@@ -0,0 +1,172 @@
...
+
+using System;
+using System.Collections.Generic;
+using MonoDevelop.Projects;
+using MonoDevelop.Core;
+using Xwt;
+
+namespace MonoDevelop.Ide.Gui.Dialogs
+{
+	public class NewConfigurationDialog : Dialog
```
mrward: Can we make this internal?

mrward:
The name text/combo box should get focus when the dialog opens. The old dialog did this but the new one does not seem to do that.

mrward:
Voice over in the dialog does not seem to work. Apart from the buttons and the title of the dialog, voice over does not say anything about the combo boxes or their labels, and also the check box is not announced.

mrward:
Voice over with tabbing seems to be OK. I was expecting to be able to use the voice over Ctrl+Option and arrow keys to navigate around the dialog - but that seems to only find the buttons.

mrward:
Might be an XWT problem. See similar problems with, for example, the NuGet license acceptance dialog. Other Mac native dialogs do not have this problem.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewConfigurationDialog.cs
```diff
@@ -0,0 +1,172 @@
...
+			Buttons.Add (okButton);
+
+			DefaultCommand = okButton.Command;
+
+			Content = mainVBox;
+		}
+
+		private void TextEntry_Changed (object sender, EventArgs e)
+		{
+			// This does nothing. For some reason without having this attached to
```
sevoku: This should not be required, let's fix Xwt.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewConfigurationDialog.cs
bratsche: I made it private just now, it seems to be what some other dialogs do.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewConfigurationDialog.cs
bratsche: > The name text/combo box should get focus when the dialog opens. The old dialog did this but the new one does not seem to do that.

I just made it so `comboName` gets focus when the dialog opens now. However, I noticed this only seems to happen when voiceover is disabled - when I turn voiceover on then I see the focus on `comboName` briefly and then it jumps to the dialog's Cancel button. @sevoku any idea, or is this normal with voiceover?

sevoku: The fix is correct, VO focus behavior is broken for most native dialogs with a modal parent, see https://devdiv.visualstudio.com/DevDiv/_workitems/edit/648643, it's probably an AtkCocoa issue.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewConfigurationDialog.cs
```diff
@@ -0,0 +1,174 @@
...
+			}
+
+			comboPlatform.TextEntry.Text = MultiConfigItemOptionsPanel.GetPlatformName ("");
+			if (!(item is Solution)) {
+				createChildrenCheck.Active = false;
+				createChildrenCheck.Visible = false;
+			}
+		}
+
+		void OkButtonClicked (object sender, EventArgs e)
```
sevoku: Sorry, I missed that: this is not the right approach, after the click has been handled Respond will be always called, so the dialog will be closed in any case and if the name is valid, it will respond twice.

The right way to do that:
* override `OnCommandActivated (Command cmd)`
* if `cmd == Command.Ok` do the logic and decide to respond or not
* otherwise call base `OnCommandActivated (cmd)`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewConfigurationDialog.cs
```diff
@@ -0,0 +1,174 @@
...
+			comboName.SetFocus ();
+		}
+
+		void Build ()
+		{
+			Padding = 0;
+			Resizable = false;
+			Title = GettextCatalog.GetString ("New Configuration");
+
+			var mainVBox = new VBox () { Spacing = 6, Margin = 6 };
```
sevoku: Use `this.Padding = 6` instead of `Content.Margin = 6` to fix VO not being able to focus the combo boxes as a workaround, or wait for the Xwt fix: https://github.com/mono/xwt/pull/963

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewConfigurationDialog.cs
bratsche: Instead of having any of this logic I just made it so that the Ok button is sensitive/insensitive depending on if the state is valid. Is that ok?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewConfigurationDialog.cs
bratsche: Fixed using `Margin = 6`, thanks!

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewConfigurationDialog.cs
bratsche: So fwiw, I ended up still using `TextEntry_Changed` when I implemented the validation so I no longer have this event handler that does nothing on here.

mrward:
Tabbing from the Name combo box does not go to the Platform combo box. It seems like it is tabbing to the warning icon but there is no visible sign that the image has focus. Should the popover get focus? It seems like it should not be able to get the focus.

Moving to the warning icon with voice over and the announcement is just 'image'. It should announce the warning text.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewConfigurationDialog.cs
```diff
@@ -0,0 +1,190 @@
...
+			comboPlatform = new ComboBoxEntry ();
+			comboPlatform.TextEntry.Changed += ComboTextChanged;
+			comboPlatform.Accessible.LabelWidget = label;
+			comboPlatform.WidthRequest = 250;
+			table.Add (label, 0, 1);
+			table.Add (comboPlatform, 1, 1);
+
+			popover = new InformationPopoverWidget ();
+			popover.Visible = false;
+			popover.CanGetFocus = false;
```
sevoku: The focus should not be skipped, otherwise it's an MAS violation. All elements should be keyboard accessible.
```suggestion
			popover.CanGetFocus = true;
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewConfigurationDialog.cs
```diff
@@ -0,0 +1,190 @@
...
+
+			label = new Label { Text = GettextCatalog.GetString ("Platform:") };
+			comboPlatform = new ComboBoxEntry ();
+			comboPlatform.TextEntry.Changed += ComboTextChanged;
+			comboPlatform.Accessible.LabelWidget = label;
+			comboPlatform.WidthRequest = 250;
+			table.Add (label, 0, 1);
+			table.Add (comboPlatform, 1, 1);
+
+			popover = new InformationPopoverWidget ();
```
sevoku: move this declaration up and set the a11y label to the Name label.
```suggestion
			popover = new InformationPopoverWidget ();
			popover.Accessible.LabelWidget = label;
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewConfigurationDialog.cs
```diff
@@ -0,0 +1,190 @@
...
+
+		void ValidateText ()
+		{
+			var name = comboName.TextEntry.Text.Trim ();
+			var isOk = false;
+
+			if (name.Length == 0 || name.IndexOf ('|') != -1) {
+				isOk = false;
+				popover.Message = GettextCatalog.GetString ("Please enter a valid configuration name.");
+				popover.Severity = Tasks.TaskSeverity.Warning;
```
sevoku: Severity is always the same, let's set it in `Build ()` when we create the popover.

bratsche:
@sevoku This has been rebased now.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewConfigurationDialog.cs
```diff
@@ -0,0 +1,190 @@
...
+			}
+
+			comboPlatform.TextEntry.Text = MultiConfigItemOptionsPanel.GetPlatformName ("");
+			if (!(item is Solution)) {
+				createChildrenCheck.Active = false;
+				createChildrenCheck.Visible = false;
+			}
+		}
+
+		ItemConfigurationCollection<ItemConfiguration> configurations;
```
Therzok: please put fields at the top.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/NewConfigurationDialog.cs
```diff
@@ -0,0 +1,190 @@
...
+
+			okButton.Sensitive = isOk;
+		}
+
+		protected override void Dispose (bool disposing)
+		{
+			base.Dispose (disposing);
+
+			if (disposing) {
+				comboName.TextEntry.Changed -= ComboTextChanged;
```
Therzok: Also set the fields to null if disposing, otherwise we have a parent <-> child ref cycle.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.OptionPanels/SolutionItemConfigurationsPanel.cs
```diff
@@ -107,21 +107,21 @@ void OnCopyConfiguration (object sender, EventArgs args)
...
 				bool done = false;
+
 				do {
-					if (MessageService.RunCustomDialog (dlg, Toplevel as Gtk.Window) == (int) Gtk.ResponseType.Ok) {
-						var cc = configData.AddConfiguration (dlg.ConfigName, copyFrom, dlg.CreateChildren);
-						store.AppendValues (cc, cc.Id);
-						done = true;
-					} else
-						done = true;
+					using (var dlg2 = new MonoDevelop.Ide.Gui.Dialogs.NewConfigurationDialog (configData.Entry, configData.Configurations)) {
```
Therzok: This should be `dlg` :D

sevoku:

PR #8396, opened 2019-08-12
Enable notarization

duncanmak:
This puts in the necessary changes to make VSMac function after turning on the hardened runtime flag. There's a corresponding PR in https://github.com/xamarin/md-addins/pull/5177.

This needs to land on both `master` and the current release branches.

@iainx Can you look over this (I just rebased it on top of this morning's master) and get this merged in. Also, I don't know where the current release branches are, so this needs to be backported there as well.

cc @sevoku 

Fixes VSTS #962402

### DIFF
--- a/main/build/MacOSX/Makefile.am
+++ b/main/build/MacOSX/Makefile.am
@@ -27,6 +27,10 @@ EXTRA_DIST = dmg-bg.png DS_Store Info.plist.in make-dmg-bundle.sh render.cs
 MONOSTUB_EXTRA_SOURCEFILES = monostub-utils.h
 export MACOSX_DEPLOYMENT_TARGET=10.12
 
+# With the hardened runtime, we need to specify the location of all libraries
+# that we dlopen
+MONOSTUB_RPATH=-Wl,-rpath,/Library/Frameworks/Mono.framework/Libraries/ -Wl,-rpath,@executable_path/../Resources/lib/monodevelop/bin/
+
 all: monostub monostub-nogui monostub-test
 
 render.exe: render.cs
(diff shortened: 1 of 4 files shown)


sevoku:
@duncanmak this needs a rebase to fix the bot provisioning for unit tests

main/build/MacOSX/Makefile.am
```diff
@@ -27,6 +27,19 @@ EXTRA_DIST = dmg-bg.png DS_Store Info.plist.in make-dmg-bundle.sh render.cs
 MONOSTUB_EXTRA_SOURCEFILES = monostub-utils.h
 export MACOSX_DEPLOYMENT_TARGET=10.12
 
+# With the hardened runtime, we need to specify the location of all libraries
+# that we dlopen
+MONOSTUB_RPATH=-Wl,-rpath,/Library/Frameworks/Mono.framework/Libraries/ \
```
wjk: Why do we need this? I haven't seen any documentation about Hardened Runtime creating a whitelist of dlopen paths. If so, this would put a real damper on the ability of an app (such as this one) to load third-party native-code plugins. I would appreciate a bit of clarification here. Thanks!

sevoku: I think @iainx can tell more next week (he's on vacation). My understanding is that it's exactly the case, macOS refused to run without this. We even need to specify the pathes to libs we ship inside our own bundle.

duncanmak: This restriction is something that we discovered during our testing. Googling around, it seems like this post confirms our finding: https://forums.developer.apple.com/message/365393#365393

iainx: Yes, basically, it seems like we need to whitelist everything we want to dlopen. It might be worth confirming our findings with a contact at Apple.

AppBundles have two special folders inside `Contents`; `MacOS` and `Resources`
`MacOS` is supposed to be for executable binaries/libraries, and `Resources` for data. Currently we have a number of libraries and executables inside the `Resources` folder. It might be that if we moved those libraries into the "correct" place inside `MacOS` that we wouldn't need to have an rpath for those libraries, but I've not checked yet.

None of this helps with the potential third-party native-code plugins problem however.

version-checks
```diff
@@ -17,8 +17,8 @@ DEP[0]=md-addins
 DEP_NAME[0]=MDADDINS
 DEP_PATH[0]=${top_srcdir}/../md-addins
 DEP_MODULE[0]=git@github.com:xamarin/md-addins.git
-DEP_NEEDED_VERSION[0]=64cc05e40930d8c2a7b6725c62d9be26511e9beb
-DEP_BRANCH_AND_REMOTE[0]="master origin/master"
+DEP_NEEDED_VERSION[0]=0b00e667f07af78978ce2f5800e4ef2ed56de1f6
+DEP_BRANCH_AND_REMOTE[0]="enable-notarization origin/enable-notarization"
```
iainx: How do we handle this, because it shouldn't get merged over to the branches?

sevoku: I'll do the merge manually

main/build/MacOSX/Makefile.am
Therzok: Really nitty here, but we should probably have a `Makefile.include` in monodevelop's MacOSX folder so we deduplicate all this code.

wjk: Technically speaking, dylibs and frameworks belong in `Contents/Frameworks`.

Therzok: > Binaries using @loader_path as the base of their install name are apparently always allowed.

That is excellent news. We have quite a few dylibs we ship which use this mechanism, and we can definitely use install_name_tool -id for those which don't comply.

PR #8397, opened 2019-08-12
[Debugger] Let VoiceOver read "Add breakpoint" and "Remove breakpoint"

ketttY:
Change toggle breakpoint into "add breakpoint" and "remove breakpoint" according to whether a breakpoint exists or not on the line with shortcut "Command + \", so that voice over can read the properly

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/753575


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.addin.xml
```diff
@@ -141,7 +141,17 @@
...
+				icon = "md-breakpoint"
+				defaultHandler = "MonoDevelop.Debugger.AddBreakpointHandler"
+				shortcut = "F9"
+				macShortcut = "Meta|\ F9" />
+		<Command id = "MonoDevelop.Debugger.DebugCommands.RemoveBreakpointOnLine"
+				_label = "Remove Breakpoint"
+				icon = "md-breakpoint"
+				defaultHandler = "MonoDevelop.Debugger.RemoveBreakpointOnLineHandler"
 				shortcut = "F9"
 				macShortcut = "Meta|\ F9" />
```
mrward: Main problem here is that we now have two commands with the same keyboard shortcut. This will show up as a conflict in preferences - Environment - Key bindings.

Can we instead change the label of the ToggleBreakpoint command based on the value of `col.Count` in the ToggleBreakpointHandler?

ketttY: I tried, seems like VoiceOver gets the msg to read before the execution. I tried both info.Text and info.Description. But it is still reading "Toggle Breakpoint"

mrward:
@iainx Is there another way to do this? The bug report mentions that voice over should say 'Breakpoint added' or 'Breakpoint removed' which does not sound like it is reading the label but some other text just for voice over. Is there a way to implement that?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.addin.xml
mrward: @iainx - Any ideas?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.addin.xml
```diff
@@ -141,7 +141,17 @@
...
 				icon = "md-breakpoint"
-				defaultHandler = "MonoDevelop.Debugger.ToggleBreakpointHandler"
+				defaultHandler = "MonoDevelop.Debugger.ToggleBreakpointHandler" />
+		<Command id = "MonoDevelop.Debugger.DebugCommands.AddBreakpoint"
+				_label = "Add Breakpoint"
+				icon = "md-breakpoint"
+				defaultHandler = "MonoDevelop.Debugger.AddBreakpointHandler"
+				shortcut = "F9"
+				macShortcut = "Meta|\ F9" />
+		<Command id = "MonoDevelop.Debugger.DebugCommands.RemoveBreakpointOnLine"
```
jstedfast: I'm thinking that the add/remove command names should use the same naming convention.

e.g.

AddBreakpointOnLine and RemoveBreakpointOnLine

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebugCommands.cs
```diff
@@ -361,7 +363,45 @@ protected override void Run ()
...
+			info.Enabled = IdeApp.Workbench.ActiveDocument != null &&
+					IdeApp.Workbench.ActiveDocument.GetContent<ITextView> (true) != null &&
+					IdeApp.Workbench.ActiveDocument.FileName != FilePath.Null &&
+					!DebuggingService.Breakpoints.IsReadOnly &&
+					(col.Count > 0);
+		}
+	}
+
+	[Obsolete]
+	class ToggleBreakpointHandler : BreakpointHandler
```
jstedfast: Instead of marking ToggleBreakpointHandler as Obsolete, should we just remove it?

It would be best to remove it unless there is code that still uses it somewhere... and if there is, that code should probably be updated.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.addin.xml
jstedfast: @mrward oh, good catch... I didn't even think of that.

Therzok: http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Components/Mac/MDMenuItem.cs,208

Therzok: It's using either CommandInfo.Text or CommandInfo.OverrideLabel, preferring the latter if it exists.

sevoku:
I think there is a simpler approach, we could let the UpdateHandler set `info.Text` to `Add Breakpoint` or `Remove Breakpoint` depending on whether there is a breakpoint at the focusted loaction or not (the result of `breakpoints.GetBreakpointsAtFileLine ()`). Then we don't need separate commands for this and can keep existing behaviour.

> Is there another way to do this? The bug report mentions that voice over should say 'Breakpoint added' or 'Breakpoint removed' which does not sound like it is reading the label but some other text just for voice over. Is there a way to implement that?

We could try to use acessibility announcements, but that's imo too complex and we don't know how that will interfere with regular menu handling. The regular behaviour is to read the menu item title, that's it. I'll also comment on the bug#.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.addin.xml
ketttY: Info.Text didn't work, it will still read the label. Any suggestion how can we access CommandEntry.OverrideLabel here?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebugCommands.cs
ketttY: If we remove the handler, should we also remove the command from MonoDevelop.Debugger.addin.xml? But we do need to keep it in the enum for the api, right?

jstedfast:
@sevoku @ketttY says she tried setting info.Text and that didn't change the VoiceOver. Is there something else she needs to do to make that work?

I think ideally if we could just make a slight adjustment to the ToggleBreakpointHandler like you are suggesting, that would be the best solution.

sevoku:
> @sevoku @ketttY says she tried setting info.Text and that didn't change the VoiceOver. Is there something else she needs to do to make that work?

Hhm, then it looks like Cocoa is caching the menu titles (it does the same for the command search inside the help menu), unfortunately, we can't do much about that and changing `info.Text` won't work in fact.

**Alternative:** making an announcement is actually not as hard as I expected, @ketttY try simply this after adding/removing the breakpint in `Run()`:

```
var msg = bp == null ? GettextCatalog.GetString ("Breakpoint removed") : GettextCatalog.GetString ("Breakpoint added") 
IdeApp.Workbench.RootWindow.Accessible.MakeAccessibilityAnnouncement (msg);
```

ketttY:
@sevoku I tried adding the announcement, but seems like it is still reading "Toggle Breakpoint", not sure what is happening here... It is using `MonoDevelop.Components.AtkCocoaHelper`, right?

sevoku:
@ketttY yes, `using MonoDevelop.Components.AtkCocoaHelper`, I just tried my proposed change and it worked perfectly fine. Well, it still reads the menu title, so it becomes:
"Toggle Breakpoint, Breakpoint added"
"Toggle Breakpoint, Breakpoint removed"
and sometimes it even skips the title (especially if you quickly activate it several times), so it also has the benefit that Cocoa would always tell the user what it did, even if VO decides to skip the title.

iainx:
As @sevoku said, Cocoa caches the menu titles refreshing the cache (I think) every time the menu is opened. I'm happy to go with the "Toggle Breakpoint, Breakpoint added" announcement method.

mrward:
Failing dotnetcore test is unrelated.

mrward:
Failing .NET Core test is not related. Due to this branch being a bit behind master.

PR #8398, opened 2019-08-12
[Core] Add new MSBuild item to ItemGroup without condition

mrward:
When a new item is added to the project and then the project is saved
the MSBuildProject will try to add the new item to an ItemGroup with
the same MSBuild item types. When adding a PackageReference this may
result in the package being added to an ItemGroup with a condition
which is not expected. Now ItemGroups with conditions are not used
when adding a new MSBuild item to a project.

Fixes VSTS #957474 - Adding a new PackageReference does not consider
conditions


mrward:
@monojenkins  backport release-8.3

Therzok (approved):

bratsche (approved):

mrward:

PR #8401, opened 2019-08-12
[DotNetCore] Don't overwrite DefaultTargets if it previously existed

bratsche:
Fixes VSTS #658903

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
@@ -549,6 +549,8 @@ namespace MonoDevelop.Projects.MSBuild
 			set { AssertCanModify (); defaultTargets = value; NotifyChanged (); }
 		}
 
+		public bool DefaultTargetsSpecified { get; set; }
+
 		string toolsVersion;
 
 		public string ToolsVersion
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
@@ -549,8 +549,6 @@ namespace MonoDevelop.Projects.MSBuild
 			set { AssertCanModify (); defaultTargets = value; NotifyChanged (); }
 		}
 
-		public bool DefaultTargetsSpecified { get; set; }
-
 		string toolsVersion;
 
 		public string ToolsVersion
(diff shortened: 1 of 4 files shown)


mrward:
Can we have a unit test for this please?

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildSdkProject.cs
```diff
@@ -90,7 +90,8 @@ public void WriteProject (MSBuildProject project, TargetFrameworkMoniker framewo
 			if (!hasDescription)
 				RemovePropertyIfHasDefaultValue (globalPropertyGroup, "Description", "Package Description");
 
-			project.DefaultTargets = null;
+			if (!project.DefaultTargetsSpecified)
+				project.DefaultTargets = null;
```
mrward: I think the DefaultTargetsSpecified should be moved inside this class and not be a part of MSBuildProject. The DefaultTargetsSpecified seems odd to me on the MSBuildProject since it is not being used when reading or writing values - just a flag.

Originally I was assuming we could detect this on reading the header in the MSBuildSdkProject, set a flag there, and then handle this on writing in the MSBuildSdkProject.

mrward:
There are some existing tests for the MSBuildSdkProject - https://github.com/mono/monodevelop/blob/master/main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/MSBuildSdkProjectTests.cs

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs
```diff
@@ -549,6 +549,8 @@ public string DefaultTargets
 			set { AssertCanModify (); defaultTargets = value; NotifyChanged (); }
 		}
 
+		public bool DefaultTargetsSpecified { get; set; }
```
Therzok: Do we want this to be internal set?

bratsche: @mrward Any idea why these tests don't show up in the unit test explorer pad?

bratsche:
> There are some existing tests for the MSBuildSdkProject - https://github.com/mono/monodevelop/blob/master/main/tests/MonoDevelop.Core.Tests/MonoDevelop.Projects/MSBuildSdkProjectTests.cs

@mrward Any idea why these tests don't show up in the unit test explorer pad?

mrward (approved):
👍

Therzok (approved):

PR #8407, opened 2019-08-13
Directory.Build.props: Remove NuGetVersionRoslyn

sandyarmstrong:
This property has moved to its own props file.

It was accidentally restored here via bad merge/rebase in
https://github.com/mono/monodevelop/pull/8201, causing
the version in the props file to be overridden.

Fixes https://dev.azure.com/devdiv/DevDiv/_workitems/edit/959442


sandyarmstrong:

PR #8409, opened 2019-08-13
[NuGet] Make references dialog aware of multi-target projects

mrward:
The Edit References dialog uses the Project.CanReferenceProject method
to determine if a project can be referenced. This does not consider
multi-target framework projects. Now a DotNetProjectExtension has
been added to the NuGet addin that uses NuGet to determine the
compatibility of all target frameworks the projects use.

Report all target frameworks as incompatible in the Edit References
dialog instead of just the first target framework.

Fixes VSTS #959111 - Unable to add references with multi-target
framework project


mrward:

mrward:

PR #8411, opened 2019-08-13
[IDE] Add "Add Reference" under project context menu

ketttY:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/947128/

Fixes VSTS #974919  - Change "Add" to "Add Reference," Remove "Add" in front of items in Add menu 
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/974919

Fixes VSTS #975256 - Change title of window from "Edit References" to "References"
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/975256

Screenshots:

AddSubMenu
ReferencesContextMenu
AddReferencesDependenciesMenu
ReferencesDialog
ProjectMainMenu

### DIFF
--- a/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
+++ b/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
@@ -239,7 +239,7 @@
 	<Command id = "MonoDevelop.Ide.Commands.ProjectCommands.AddReference"
 			_description = "Add and remove project references"
 			icon = "md-reference"
-			_label = "_Edit References..."
+			_label = "_Add References..."
 			defaultHandler = "MonoDevelop.Ide.Commands.EditReferencesHandler" />
 	<Command id = "MonoDevelop.Ide.Commands.ProjectCommands.AddNewFiles"
 			_label = "New _File..."
--- a/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
+++ b/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
@@ -239,7 +239,7 @@
 	<Command id = "MonoDevelop.Ide.Commands.ProjectCommands.AddReference"
 			_description = "Add and remove project references"
 			icon = "md-reference"
-			_label = "_Add References..."
+			_label = "_Add Reference..."
 			defaultHandler = "MonoDevelop.Ide.Commands.EditReferencesHandler" />
 	<Command id = "MonoDevelop.Ide.Commands.ProjectCommands.AddNewFiles"
 			_label = "New _File..."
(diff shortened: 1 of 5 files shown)


ketttY:
Screen Shot 2019-08-13 at 2 00 26 PM
This is the current location of this command, please let me know if you think it will be better to put it somewhere else :)

ketttY:
Screen Shot 2019-08-13 at 2 00 26 PM
This is the new layout.

bratsche (approved):
👍

mrward (approved):
I sort of wonder how that `insertbefore="MonoDevelop.WebReferences.WebReferenceCommands.Add"` in the NuGet addin is being used since the screenshot looks good to me - however this insertbefore would suggest a different order. I am guessing the web references addin is loaded later after the NuGet addin.

Not suggesting we look at that here though 😄

Therzok (approved):

jstedfast:
GitHub has been stalling out all day yesterday afternoon and this morning for me so I can't review/approve this PR (little blue progress bar gets 90% of the way across my screen and then just stops when I click the "Add your review" button), so I'll just comment here...

Looks good to me!

jstedfast (approved):

vancura:
Looks good, but I can't help myself, any idea why "Add Web Reference" has no ellipsis?

mrward:
@vancura We can fix up the Add Web Reference.

Should this be run by the UX design board? We now have two different labels for the same action. Add Reference and the existing Edit References. Which both open the Edit References dialog.

vancura:
Yes, we need an UXBoard, this is a rather significant change.

It's not the right thing to have two commands with the same action, we should kill one of them. But the problem is that if we kill _Edit References_, it won't be clear how to actually edit references, the _Add Reference_ doesn't tell the story about editing capabilities. It seems we're over-complicating the issue.

What about _Add or Edit References..._ instead of the simple _Add Reference..._, and killing the _Edit References..._?

/c @hbons

ketttY:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/975256/

vancura (approved):

Therzok (approved):

main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
```diff
@@ -239,7 +239,7 @@
 	<Command id = "MonoDevelop.Ide.Commands.ProjectCommands.AddReference"
 			_description = "Add and remove project references"
 			icon = "md-reference"
-			_label = "_Edit References..."
+			_label = "_Add References..."
 			defaultHandler = "MonoDevelop.Ide.Commands.EditReferencesHandler" />
```
mrward: Should this be `_Add Reference...` to be consistent with the other menu which would be `Add -> Reference...`. I think I added the wrong name to the bug report. It would also match VS on Windows //cc @vancura

vancura: Right, singular, I completely agree.

mrward (approved):

PR #8414, opened 2019-08-14
Fixes VSTS Bug 962517: [Feedback] Application output window is bigger

mkrueger:
than the screen

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/962517

It's easy to reproduce - just make the window big, hover -> make
window small - hover. The validation fixes that issue.


sevoku:

PR #8415, opened 2019-08-14
Fixes VSTS Bug 950651: VS4MAc does not handle switching branches well

mkrueger:
Hard to tell what did go wrong for the user which opened the bug request.

I made the switch branch code async and cleaned it up a bit - ThawEvents may've been called too soon by RunBlockingOperation and BranchSelectionChanged is now in sync with the whole operation.

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git.Tests/BaseGitRepositoryTests.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git.Tests/BaseGitRepositoryTests.cs
@@ -317,7 +317,7 @@ namespace MonoDevelop.VersionControl.Git.Tests
 
 				await Task.Run (() => repo2.CreateBranch ("branch1", null, null));
 
-				await Task.Run (() => repo2.SwitchToBranch (monitor, "branch1"));
+				await repo2.SwitchToBranchAsync (monitor, "branch1");
 				// Nothing could be stashed for master. Branch1 should be popped in any case if it exists.
 				Assert.IsFalse (repo2.GetStashes ().Any (s => s.Message == GetStashMessageForBranch ("master")));
 				Assert.IsFalse (repo2.GetStashes ().Any (s => s.Message == GetStashMessageForBranch ("branch1")));
@@ -328,7 +328,7 @@ namespace MonoDevelop.VersionControl.Git.Tests
 				await AddFileAsync ("file2", "text", true, false);
 				repo2.CreateBranch ("branch2", null, null);
 
-				repo2.SwitchToBranch (monitor, "branch2");
+				await repo2.SwitchToBranchAsync (monitor, "branch2");
 				if (automaticStashCreation) {
 					// Branch1 has a stash created and assert clean workdir. Branch2 should be popped in any case.
 					Assert.IsTrue (repo2.GetStashes ().Any (s => s.Message == GetStashMessageForBranch ("branch1")));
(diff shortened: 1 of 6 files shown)


mkrueger:
Switching branch seems to work a bit better with that implementation but I didn't measure numbers.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectSystemHandler.cs
```diff
@@ -115,8 +115,11 @@ internal static IEnumerable<string> GetFrameworks (MonoDevelop.Projects.Project
 				ProjectCacheInfo cacheInfo,
 				string framework)
 			{
+				if (p.ParentSolution == null) {
+					LoggingService.LogError ("Parent solution == null. " + Environment.StackTrace);
```
Therzok: How are we getting into this state?

mkrueger: I don't know saw that once

sevoku: Might be same as https://github.com/mono/monodevelop/pull/8440 ?

sevoku:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectSystemHandler.cs
sevoku: BTW shouldn't this be LogInternalError ?

mkrueger: There will be a null reference exception later that gets tracked - but in that case the log file contains the stack trace from above that'll hopefully tell us more about why it happened.

Therzok: We can't get diagnostic information from LogError. We either try to find why the exception happens - i.e. go to definition in misc workspace - or do not just shadow it

mkrueger: No chance at that point from my side to find out what happened. 
In any case that does not shadow anything - why do you think it does ? This adds information to the log.

Therzok: To the _log_, not reported telemetry. how are we going to access that information from the logs?

A NRE would most likely be noticed by us as a functional failure, but a log of a null value should not.

Imo, we shouldn't just log something in case we end up in an unexpected state. A Debug.Assert if we want to figure out more things to diagnose or get to the root cause.

sevoku: I agree with Marius. How is that related to the actual bug being fixed here btw? Does branch switching fail when this occurs? Logging this just to the log, doesn't help us at all. If we have evidence that this can happen, we should track it. Also this should be handeled by the caller properly and not here.

mkrueger: /removed it's not really related to the bug.

sevoku:

PR #8416, opened 2019-08-14
[NuGet] Use different command ids for manage packages commands

mrward:
Using the same command ids as those in the separate NuGet extensions
addin can cause odd behaviour if an old NuGet extensions addin is
disabled in the IDE. This can result in the Manage NuGet Packages
not being shown on the Solution window's context menu.


mrward:

mrward:
Mmm, looks like I missed the .NET Core addin here. It is still using the old commands. Will fix it in another pull request.

PR #8417, opened 2019-08-14
[AspNetCore] Switch to HttpClient from TcpClient

rodrmoya:
For some reason (so far, seems to be a bug in the latest Mono we bumped
to), TcpClient stopped working on this several years old code that has had
no change that could have broken it. So, as a workaround, switch to use
HttpClient, which works.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/962315

### DIFF
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreExecutionHandler.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreExecutionHandler.cs
@@ -27,10 +27,10 @@
 using System;
 using System.Linq;
 using System.Threading.Tasks;
-using System.Net.Sockets;
 using MonoDevelop.Core;
 using MonoDevelop.Core.Execution;
 using MonoDevelop.Ide;
+using MonoDevelop.Core.Web;
 
 namespace MonoDevelop.AspNetCore
 {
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreExecutionHandler.cs
```diff
@@ -86,12 +87,15 @@ public static async Task LaunchBrowserAsync (string appUrl, string launchUrl, Ex
...
-				using (var tcpClient = new TcpClient ()) {
+				using (var httpClient = HttpClientProvider.CreateHttpClient (launchUri.AbsoluteUri)) {
 					try {
-						await tcpClient.ConnectAsync (launchUri.Host, launchUri.Port).ConfigureAwait (false);
-						// pause briefly to allow the server process to initialize
-						await Task.Delay (TimeSpan.FromSeconds (1)).ConfigureAwait (false);
-						break;
+						using (var response = await httpClient.GetAsync (launchUri.AbsoluteUri, System.Net.Http.HttpCompletionOption.ResponseHeadersRead).ConfigureAwait (false)) {
+							if (response.StatusCode == System.Net.HttpStatusCode.OK) {
+								// pause briefly to allow the server process to initialize
```
mrward: Will checking for OK work in all cases? For example if you do not enable ssl cert properly will you get an OK status code? In the past VS Mac would open the page and you would see an error displayed the browser. The original code just checks it can make a connection to the port.

I am just wondering if instead it should just check that there was a response?

rodrmoya: yes, right

rodrmoya:
Mono bug -> https://github.com/mono/mono/issues/16234

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreExecutionHandler.cs
rodrmoya: Had a try at sending a "HEAD" request, but this works for all templates except for empty, so just moved it back to just make a request and not check the status, if we got a response, break the loop

rodrmoya:

PR #8424, opened 2019-08-14
[Debugger] Add Debug All Test command

ketttY:
Make it possible to create a keyboard shortcut to Debug all unit tests.

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/802586/

UnitTestWindow


mrward:
UnitTestsWindow

@vancura The Debug All button is new in the Unit Tests window in this change.

main/src/addins/MonoDevelop.UnitTesting/Gui/TestPad.cs
```diff
@@ -105,6 +105,14 @@ public override void Initialize (NodeBuilder[] builders, TreePadOption[] options
 			buttonRunAll.TooltipText = GettextCatalog.GetString ("Run all tests");
 			topToolbar.Add (buttonRunAll);
 			
+			buttonDebugAll = new Button (GettextCatalog.GetString ("Debug All"));
+			buttonDebugAll.Accessible.Name = "TestPad.DebugAll";
+			buttonDebugAll.Accessible.Description = GettextCatalog.GetString ("Debug all the tests");
+			buttonDebugAll.Clicked += new EventHandler (OnDebugAllClicked);
```
jstedfast: Not sure why the other code in this file is using the `x.Clicked += new EventHandler (...);` pattern, but it's not actually necessary.

You can actually just do this:

```csharp
buttonDebugAll.Clicked += OnDebugAllClicked;
```

That said, maybe it's best to just follow the same pattern throughout the rest of the file, in which case just leave it as you have it.

main/src/addins/MonoDevelop.UnitTesting/Gui/TestPad.cs
```diff
@@ -554,6 +562,24 @@ void OnRunAllClicked (object sender, EventArgs args)
...
+				return;
+
+			foreach (var mode in debugModeSet.ExecutionModes) {
+				if (test.CanRun (mode.ExecutionHandler)) {
+					RunTests (TreeView.GetSelectedNodes (), mode.ExecutionHandler);
+				}
+			}
+			
+		}
+
```
jstedfast: I'm not sure that the implementation of `OnDebbugAllClicked` is correct.

I was expecting it to look more like `OnRunAllClicked` other than needing to get the `debugModeSet`.

The `OnRunAllClicked` handler gets the root tree node (which presumably runs all of the tests under the root node of the tree) whereas `OnDebugAllClicked` seems to run only the selected test?

ketttY: The function for debugging is `RunTests (ITreeNavigator[] navs, IExecutionHandler mode, bool bringToFront = true)` which requires `ITreeNavigator[]`. I found this function from the CommandHandler of `DebugTest`. I am not sure if there's any other way of running it in debug mode, but if there is a more proper way of doing it please tell me! Thanks

mrward: The DebugAllTestHandler seems to use the root test - can you use similar code in the TestPad?

    UnitTestService.RunTest (test, context);

Or just use the TestPad's RunTest method? That seems to take an UnitTest.

    public AsyncOperation RunTest (UnitTest test, IExecutionHandler mode)

Or use the TestPad's:

    AsyncOperation RunTests (IEnumerable<UnitTest> tests, IExecutionHandler mode, bool bringToFront)

Looking at the RunAllTestsClicked method maybe something like this?
   
    			var debugModeSet = Runtime.ProcessService.GetDebugExecutionMode ();
				if (debugModeSet == null)
					return;
	
				var test = TreeView.GetRootNode ()
				if (test == null)
					return;
	
				foreach (var mode in debugModeSet.ExecutionModes) {
					if (test.CanRun (mode.ExecutionHandler)) {
						RunTest (test, mode.ExecutionHandler);
					}
				}

jstedfast: I would propose something like this:

```csharp
void OnDebugAllClicked (object sender, EventArgs args)
{
	var nav = TreeView.GetRootNode ();
	if (nav == null)
		return;

	var test = nav.DataItem as UnitTest;
	if (test == null)
		return;

	var debugModeSet = Runtime.ProcessService.GetDebugExecutionMode ();
	if (debugModeSet == null)
		return;

	foreach (var mode in debugModeSet.ExecutionModes) {
		if (test.CanRun (mode.ExecutionHandler)) {
			RunTests (new UnitTest [] { test }, mode.ExecutionHandler);
			return;
		}
	}
}
```

main/src/addins/MonoDevelop.UnitTesting/Gui/TestPad.cs
```diff
@@ -565,6 +591,7 @@ void OnTestSessionCompleted ()
 			runningTestOperation = null;
 			this.buttonRunAll.Sensitive = true;
 			this.buttonStop.Sensitive = false;
+			this.buttonDebugAll.Sensitive = true;
```
jstedfast: As I was looking at the rest of the code, I noticed a few other places that need `button.DebugAll.Sensitive` to be changed. Look for all the places where `button.RunAll.Sensitive` is changed and set `button.DebugAll.Sensitive` to the same value in the same places.

main/src/addins/MonoDevelop.UnitTesting/Gui/TestPad.cs
bratsche: Speaking of the event handlers, do we not need to remove them later?

main/src/addins/MonoDevelop.UnitTesting/Gui/TestPad.cs
```diff
@@ -554,6 +563,25 @@ void OnRunAllClicked (object sender, EventArgs args)
 			RunTest (TreeView.GetRootNode (), null);
 		}
 		
+		void OnDebugAllClicked (object sender, EventArgs args)
+		{
+			var debugModeSet = Runtime.ProcessService.GetDebugExecutionMode ();
+			if (debugModeSet == null)
+				return;
+
+			UnitTest test = GetSelectedTest ();
```
mrward: Going on what Jeff has commented, I think this needs to change, since it is getting the selected test not the root node.

main/src/addins/MonoDevelop.UnitTesting/Gui/TestPad.cs
jstedfast: Good point.

vancura:
I was assigned to review this feature, but I don't see anything about design other than there has to be a new default keyboards shortcut, am I right? Or is there anything else?

mrward:
@vancura Yes, the original design was about adding a keyboard shortcut. However the DebugAllTests command did not exist before so was not available anywhere in the UI. We thought it would make sense to make it available the same way that the RunAllTests command is which is via a button in the Unit Tests pad so a new button was added - it has no icon. Another idea was to just have it available in the context menu for Unit Tests.

vancura:
Right, I see it now.

Can we have just three icon buttons without labels? _Run All_, _Debug All_ and _Stop_ (disabled when not running)? These labels are not really needed if we still have the title available in tooltips.

We still need an icon, which we'll do through the Jelly Squad.

main/src/addins/MonoDevelop.UnitTesting/Commands/UnitTestCommands.cs
```diff
@@ -82,7 +84,44 @@ protected override void Run ()
...
+				var debugModeSet = Runtime.ProcessService.GetDebugExecutionMode ();
+				if (debugModeSet == null)
+					return;
+
+				foreach (UnitTest test in tests) {
+					foreach (var mode in debugModeSet.ExecutionModes) {
+						if (test.CanRun (mode.ExecutionHandler)) {
+							ExecutionContext context = new ExecutionContext (mode.ExecutionHandler, IdeApp.Workbench.ProgressMonitors.ConsoleFactory, null);
+							await UnitTestService.RunTests (new UnitTest [] { test }, context, true).Task;
+						}
```
mrward: If we have multiple execution modes this will run the tests for each execution mode. Should we be continuing the loop after we have found an execution handler that a test can be run with?

main/src/addins/MonoDevelop.UnitTesting/Gui/TestPad.cs
```diff
@@ -553,7 +559,38 @@ void OnRunAllClicked (object sender, EventArgs args)
...
+			this.buttonRunAll.Sensitive = false;
+			this.buttonDebugAll.Sensitive = false;
+			this.buttonStop.Sensitive = true;
+
+			foreach (UnitTest test in tests) {
+				foreach (var mode in debugModeSet.ExecutionModes) {
+					if (test.CanRun (mode.ExecutionHandler)) {
+						ExecutionContext context = new ExecutionContext (mode.ExecutionHandler, IdeApp.Workbench.ProgressMonitors.ConsoleFactory, null);
+						await UnitTestService.RunTests (new UnitTest [] { test }, context, true).Task;
+					}
```
mrward: Similar thing here.

PR #8436, opened 2019-08-15
[aspnetcore] Disables User-specific run config for asp.net core projects

jtorres:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/965794

TODO: Clean up commit history


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectRunConfiguration.cs
```diff
@@ -150,7 +151,9 @@ void GetProps (IPropertySet p, Dictionary<string,string> dict)
 
 		internal MSBuildProjectInstance ProjectInstance { get; set; }
 
-		public bool StoreInUserFile { get; set; } = true;
+		public bool StoreInUserFile { get; set; } 
+
+		public bool UserSpecificUnSupported { get; set; }
 	}
```
mrward: I think we should use `UserSpecificUnsupported` instead of `UserSpecificUnSupported`. `Unsupported` is used elsewhere in the code base.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectRunConfiguration.cs
```diff
@@ -37,6 +37,7 @@ public class ProjectRunConfiguration: SolutionItemRunConfiguration
 
 		public ProjectRunConfiguration (string name): base (name)
 		{
+			StoreInUserFile = !UserSpecificUnSupported;
 		}
```
mrward: Here we seem to have a relationship between these two properties, but I could create a run configuration and do this:

```
var config = new ProjectRunConfiguration ("Test");
config.UserSpecificUnsupported = true;
config.StoreInUserFile = true;
```

Right now that would disable the check box in the project options but the check box would be enabled. Not sure if that would be expected based on the name of the property.

Not a big problem - just wondering. Right now it seems that this UserSpecificUnsupported property seems to mean that you cannot change the value in project options.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectRunConfiguration.cs
rodrmoya: Or even better, IMO, why not have a "FeaturesSupported" or "SupportsFeature (feature)" with a list of flags? I think that's better than having a field for each new "setting"

So, something like

```
[Flags]
enum ProjectRunConfigurationFeatures {
  UserSpecificSupported
  ...
}
```
?

jtorres: hmm what is the advantage? there is just one enum value `UserSpecificSupported` right now

rodrmoya: Right, but whenever we need to do something similar, flags is better than adding a property for each "setting". Just thinking about breaking API in the future as little as possible, since this is public

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectRunConfiguration.cs
```diff
@@ -30,13 +30,22 @@
 
 namespace MonoDevelop.Projects
 {
+	[Flags]
+	public enum ProjectRunConfigurationFeatures
+	{
+		None,
+		UserSpecificSupported
+	}
+
```
mrward: I think we may want to [explicitly list the enum values here](https://github.com/mono/monodevelop/blob/305bddc1e9dcd90caf25f8bed1a042efc317528b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectFeature.cs#L31). If we add more enums here then they will increment by 1 by default which is not going to work for flags.

Also should we just use UserSpecific here instead of UserSpecificSupported?

rodrmoya (approved):
LGTM and works, just probably let's use the existing ProjectFeatures enum, as @mrward suggests

rodrmoya (approved):

mrward (approved):

rodrmoya:
@monojenkins backport pr-8.3-launch-settings-fix

PR #8440, opened 2019-08-15
Fix NRE when accessing null ParentSolution property on DotNetCoreProject

iantoalms:
Repro steps:

1) Create a solution with a .NET Core app project.
2) Add a .Net Core API project to this existing solution.

During project creation the API project will try to access the ParentSolution property in the FileService_FileChanged callback, before it has been set on the project.

This change is the least risky and minimal fix to avoid the NRE exception. A subsequent PR should include broader changes to avoid unnecessary duplication effort of global.json processing at the project level, among other improvements.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/944667


rodrmoya:
From @Therzok in the bug:

```When we fix this bug, can we also change the extension to subscribe to Project.FileChangedInProject rather than FileService? That way, we only get notifications for files in the project, rather than files in all projects.
```

so I guess we should do that?

Therzok:
@iantoalms Can you create a VSTS item tracking the improvements so we don't lose them?

rodrmoya:

PR #8449, opened 2019-08-16
Updated build scripts for Windows.

jeremy7nelson:
Call common winbuild.bat from debug and release scripts.


Therzok:
Thanks for the PR!

PR #8465, opened 2019-08-18
[VersionControl] Started to implement a git based client library and

mkrueger:
use it for the clone/checkout operation.

It's just a start. Checkout doesn't really work with libgit2 so using the command line client for checking out is a way to work around that. In theory we can use the git command line client for everything.

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git.ClientLibrary/MonoDevelop.VersionControl.Git.ClientLibrary/Branches/BranchUtil.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git.ClientLibrary/MonoDevelop.VersionControl.Git.ClientLibrary/Branches/BranchUtil.cs
@@ -34,6 +34,40 @@ using System.Collections.Generic;
 
 namespace MonoDevelop.VersionControl.Git.ClientLibrary
 {
+	public class CreateTagOptions
+	{
+		/// <summary>
+		/// The name of the new tag.
+		/// </summary>
+		public string Name { get; }
+
+		/// <summary>
+		/// Use the given tag message.
+		/// </summary>
+		public string Message { get; set; }
+
+		/// <summary>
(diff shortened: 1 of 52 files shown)


main/src/addins/VersionControl/MonoDevelop.VersionControl.Git.ClientLibrary/MonoDevelop.VersionControl.Git.ClientLibrary/Branches/BranchUtil.cs
```diff
@@ -42,10 +43,51 @@ public static async Task<GitLocalBranch> GetCurrentBranchAsync (string rootPath,
...
 			arguments.AddArgument ("--short");
 			arguments.AddArgument ("HEAD");
-			Console.WriteLine (arguments.Arguments);
 			await new GitProcess ().StartAsync (arguments, handler, false, cancellationToken);
-			Console.WriteLine (handler.Output);
 			return new GitLocalBranch (handler.Output.TrimEnd ());
 		}
+
+		#region Tags
+		public static async Task<GitResult> CreateNewTagAsync (string rootPath, string tagName, CancellationToken cancellationToken = default)
```
Therzok: Can we refactor the number of entry points we have? Maybe:

```
public static async Task<GitResult> TagAsync (TagOptions tagOptions)
{

}

class TagOptions
{
    public string RootPath { get; }
    public string TagName { get; }
    public CancellationToken Token { get; }
    public bool Delete { get; set; }

    public GitArguments ToArgs () {}
}
```

mkrueger: Working on that right now :). funny

mkrueger:
Closing we're switching to GIT.Api.

PR #8468, opened 2019-08-19
[NuGet] Fix duplicate projects in select projects dialog

mrward:
Checking more than one NuGet package in the Manage NuGet packages
dialog when that NuGet package is installed in multiple projects
could result in projects being listed multiple times in the Select
Projects dialog. This was because each checked package id was
being checked against all the packages installed in the project and
for each match the project was being returned. This would result
in the same project being listed multiple times in the dialog which
was confusing.

Fixes VSTS #966636 - NuGet Package Manager dialog after selecting
multiple packages to uninstall unclear

Also added more tests for the Manage NuGet packages dialog.

Fixes VSTS #961926 - [test coverage] nuget - solution wide nuget management


mrward:

PR #8474, opened 2019-08-19
Onboard arcade

agocke:


agocke:
@sandyarmstrong It looks like your build layout is very complicated -- make is building things to different directories and then invoking with a relative path. The Arcade transition is probably going to be pretty complicated. Is there someone on your team who can takeover?

sandyarmstrong:
@agocke sure. This may not happen until we make some upcoming changes to our build, and there is no specific timeframe for that.

agocke:
@sandyarmstrong Let's talk about this at the next sync. Roslyn provides new builds to its consumers via Arcade, so if VSfM wants regular updates I think adopting Arcade should be a prerequisite.

PR #8479, opened 2019-08-20
[Core] Some miscellaneous changes in the counter code

Therzok:
Not big changes, just removal of some unused APIs, some memory optimizations and code style.

Review each commit individually.

Mostly optimizes counter memory usage, reliability of time trackers and cleans up API. API removals are pretty straight-forward.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/MemoryProbe.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/MemoryProbe.cs
@@ -34,8 +34,7 @@ namespace MonoDevelop.Core.Instrumentation
 		
 		public MemoryProbe (Counter c)
 		{
-			this.c = c;
-			c++;
+			this.c = ++c;
 		}
 		
 		~MemoryProbe ()
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/MemoryProbe.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/MemoryProbe.cs
@@ -34,12 +34,13 @@ namespace MonoDevelop.Core.Instrumentation
 		
 		public MemoryProbe (Counter c)
 		{
-			this.c = ++c;
+			this.c = c;
+			c.Inc ();
 		}
 		
 		~MemoryProbe ()
 		{
-			c--;
+			c.Dec ();
 		}
 	}
 }
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimerCounter.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimerCounter.cs
@@ -64,7 +64,7 @@ namespace MonoDevelop.Core.Instrumentation
 		}
 
 		public TimeSpan AverageTime {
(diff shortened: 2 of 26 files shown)


Therzok:
I realize this PR is getting huge. I'll chunk it.

Therzok:
Actually, quite easy to review commit by commit.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/MemoryProbe.cs
```diff
@@ -34,8 +34,7 @@ public class MemoryProbe
 		
 		public MemoryProbe (Counter c)
 		{
-			this.c = c;
-			c++;
+			this.c = ++c;
```
iainx: should this not be `c++`?

Therzok: Same thing, really, but I can change it.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimerCounter.cs
```diff
@@ -93,24 +93,6 @@ internal void AddTime (TimeSpan time)
...
-					}
-				}
-			} else if (LogMessages) {
-				if (lastTimer != null)
-					lastTimer.Trace (message);
-				else if (message != null)
-					InstrumentationService.LogMessage (message);
-			}
-		}
-
```
mrward: Not sure I understand why this is not needed. Doesn't this mean lastTimer is not being used here anymore?

Not that I completely follow how the timer counter worked before :)

mrward: Oh I see. lastTimer is removed in a later commit.

mrward (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/MemoryProbe.cs
mrward: I was wondering the same thing but the Counter operator returns the same counter so it does not seem to matter. For a minute I was assuming the code was different like if an int was being incremented here.

iainx: Changing the `++` semantics is probably not a good thing here, because it clearly causes confusion unless you look deeper. Maybe deprecate the operator overload and use a much clearer 'Inc' method?

Therzok: Lemme go for that option, thanks @iainx

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimerCounter.cs
Therzok: Yeah, lastTimer is completely unreliable, possibly causing timers to have unexpected values. Best remove it rather than deprecate it.

If something still uses it, I can add dummies and obsolete the API, but I preferred to remove it in case nothing broke.

Therzok:
Weird f# failure, hm

Therzok:
Will do a manual test of the dmg

rodrmoya (approved):

jtorres (approved):

slluis (approved):

PR #8482, opened 2019-08-20
[NuGet] Allow checkboxes to be selected by keyboard in packages dialog

mrward:
[NuGet] Allow checkboxes to be selected by keyboard in packages dialog  …
Previously pressing space or enter would cause the selected package,
or multiple packages if checked, to be installed. Now pressing space
or enter checks or unchecks the package.

Fixes VSTS #750399 - Accessibility: Onfocus issue has been observed
while selecting packages from the list

Fixes VSTS #750407 - Accessibility: Checkboxes for selecting multiple
packages are not accessible through keyboard

[NuGet] Do not install package when enter pressed in search text box

No longer installing the selected package in the Manage NuGet Packages
dialog when the enter key is pressed. This has caused a bit of
confusion and can cause a package to be installed when the user is
trying to run the search. Note that the search is a search as you type
so pressing the enter key has no affect.

Fixes VSTS #968790 - Hitting return in a search field installs a package


mrward:
I can look at adding a new UI test. The existing TestAddAndUpdatePackages tests what happens when a single package is installed so I do not want to change that since it is testing one way of using the dialog. Using a checkbox is more useful when adding more than one package at a time.

PR #8485, opened 2019-08-20
[Core] Add a way to use spindump as a profiler backend

Therzok:
This commit introduces spindump as a profiling backend, alongside sample.

spindump has the advantage of being able to distinguish running threads
from waiting threads, thus will only generate information about things
spinning on the CPU.

On the other hand, spindump (even the normal report) is not loadable in Instruments.
But given the fact that the data is way smaller:

On the same scenario:
Sample, ~1450 stacktrace lines: https://gist.github.com/Therzok/d81908c632a3b93b30727a2b98c03d0b
Spindump, ~1200 stacktrace lines: https://gist.github.com/Therzok/94a6d3ee44ec26ce76b9266e76fa6dda

But the latter contains only actual running time, nothing that is actually spent waiting on synchronization
primitives.

Introduced a few new regexes to handle spindump format. Keep output that is sample compatible.

Also handle stackframes that are in kernel code by removing them, therefore simplifying the output.

Fixes VSTS #897635 - Consider having a spindump version for profiling, not just sample

### DIFF
--- a/main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/Commands.cs
+++ b/main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/Commands.cs
@@ -67,7 +67,7 @@ namespace PerformanceDiagnosticsAddIn
 			if (!File.Exists (spinDumpFile)) {
 				var message = new ConfirmationMessage (
 					GettextCatalog.GetString ("{0} requires administrative privileges to run spindump", BrandingService.ApplicationName),
-					GettextCatalog.GetString ("Continue with installing '{0}' so 'spindump' can be run without a password for this user?", spinDumpFile),
+					GettextCatalog.GetString ("Continue with installing '{0}' so 'spindump' can be run without a password for the current user?", spinDumpFile),
 					AlertButton.Proceed) {
 					Icon = MonoDevelop.Ide.Gui.Stock.Warning,
 				};
--- a/main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/Commands.cs
+++ b/main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/Commands.cs
@@ -72,8 +72,7 @@ namespace PerformanceDiagnosticsAddIn
 					Icon = MonoDevelop.Ide.Gui.Stock.Warning,
 				};
 
-				if (!MessageService.Confirm (message) || !CreatePrivilegedHelperFile ())
-					return false;
+				return MessageService.Confirm (message) && CreatePrivilegedHelperFile ();
 			}
 
 			return true;
--- a/main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/Commands.cs
+++ b/main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/Commands.cs
@@ -11,7 +11,7 @@ using MonoDevelop.Ide.Gui.Dialogs;
 
 namespace PerformanceDiagnosticsAddIn
 {
-	public class StartStopListeningUIThreadMonitorHandler : CommandHandler
+	class StartStopListeningUIThreadMonitorHandler : CommandHandler
 	{
 		protected override void Run ()
 		{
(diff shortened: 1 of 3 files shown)


Therzok:
To be able to actually use the functionality, you need to have sudo rights for the process. Not sure exactly how to enforce that from .NET.

```
➜  monodevelop git:(spindump-prof) cat /etc/sudoers.d/spindump
therzok ALL=(ALL:ALL) NOPASSWD:/usr/sbin/spindump
```

abock:
@Therzok see https://github.com/microsoft/workbooks/blob/master/Clients/Xamarin.Interactive.Client.Mac/AppDelegate.cs#L161-L196 for how to authorize the writing of `/etc/sudoers.d/spindump`.

Therzok:
Interesting, I'll try that tomorrow. Thanks a lot!

main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/Commands.cs
```diff
@@ -33,7 +34,21 @@ protected override void Update (CommandInfo info)
 
 		protected override void Run ()
 		{
-			UIThreadMonitor.Instance.Profile (5);
+			UIThreadMonitor.Instance.Profile (spinDump: false, 5);
+		}
+	}
+
+	public class SpinDumpFor5SecondsHandler : CommandHandler
```
mrward: I think this can be non-public.

main/src/core/MonoDevelop.Core/MonoDevelop.Utilities/SampleProfiler.cs
```diff
@@ -41,59 +41,87 @@ namespace MonoDevelop.Utilities
...
-			sampleProcess.EnableRaisingEvents = true;
-			sampleProcess.Exited += delegate {
-				ConvertJITAddressesToMethodNames (outputPath, outputFilePath, "Profile");
+			const int millisBetweenSamples = 1;
+
+			// We need to delete the file before using it as an output target, otherwise it will error.
+			File.Delete (outputFilePath);
+
+			return new ProcessStartInfo ("sudo") {
+				UseShellExecute = false,
```
mrward: How does VS Mac need to be run here for this to work? You need to `sudo VS Mac`?

Just wondering if we could prompt to elevate like we do for dotnet dev-certs?

Therzok: @mrward interesting, where do we do that?

I just edited `/etc/sudoers.d` to add an entry for `spindump`.

mrward: We have a separate console app that uses a Mac API to request authorization, then the console app calls itself, uses setuid to ensure it can run the same as sudo, then runs the command to install the dev certs. It is not that straightforward but maybe we could use that here on the Mac?

https://github.com/mono/monodevelop/blob/37e7bedb14ea06790e6002dfc22ce889229cb006/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/DotNetCoreDevCertsTool.cs#L87-L108

https://github.com/mono/monodevelop/blob/305bddc1e9dcd90caf25f8bed1a042efc317528b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.DevCertInstaller/Program.cs#L76

mrward: I guess editing sudoers.d is less work here. We could add a comment about this somewhere?

Therzok:
Added spindump sudoers file installation

main/src/core/MonoDevelop.Core/MonoDevelop.Utilities/SampleProfiler.cs
Therzok: Added an implementation that writes to sudoers file. I wonder if we should not try to do this for users that are not internal.

PR #8492, opened 2019-08-21
Prevents API breakage

jtorres:


jtorres:
@monojenkins backport master

PR #8313, opened 2019-07-31
Thread safe BreakpointStore

sgmunn:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/949284

This removes all the locks around the breakpoint store now that it is thread safe and bumps to pick up the changes.

PR #8496, opened 2019-08-22
[release-8.2] Thread safe BreakpointStore

mrward:
Backport of #8313

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/949284
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/935168

This removes all the locks around the breakpoint store now that it is thread safe and bumps to pick up the changes.


mrward:
Re-running the build. For some reason Xamarin.Mac cannot be found. The changes here would have no affect on that.

mrward:
Seems to be building ExtensionTools.csproj earlier on this branch - not sure why.

ExtensionTools.csproj tries to reference Xamarin.Mac.dll from build/bin which is not going to work if it builds before MonoDevelop.Ide. If the rebuild does not work I will change ExtensionTools.csproj to reference Xamarin.Mac.dll from external.

PR #8504, opened 2019-08-23
[release-8.2] [DotNetCore] Check for unsupported newest SDK versions

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/968797


mrward:
I think this is one part of the fix. It does not prevent 3.0 from being listed in the new project dialog so does not fix the bug.

mrward:
I think we need to make some changes in [DotNetCoreProjectSupportedTargetFrameworks](https://github.com/mono/monodevelop/blob/master/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectSupportedTargetFrameworks.cs) which I believe is used by the New Project dialog when showing the list of versions.

rodrmoya:
> I think this is one part of the fix. It does not prevent 3.0 from being listed in the new project dialog so does not fix the bug.

right, if users have 3.0 preview7, it would still work. Do we want to just hide 3.0 ignoring what version the user has installed?

rodrmoya:
Right, latest commit contains that, so we only show 3.0 templates if preview7 or lower is installed

PR #8510, opened 2019-08-23
[Mac] Remove the code for changing application name

iainx:
The Carbon code for changing the application name is private and unsupported,
and fragile, and isn't really needed anymore as we just want to use the app
name from the app bundle.

The downside of this is that the app name displayed when debugging from inside
the IDE is "mono64" and when run from 'make run' it is MonoDevelop. Run from
an appbundle, it gets the correct name

Fixes vSTS #899321


iainx:

PR #8518, opened 2019-08-26
[Mac] Use actual type from signature, float instead of int

Therzok:
SetContentHuggingPriorityForOrientation and SetContentCompressionResistancePriority both have `float priority`, so use that.


Therzok:
Funnily enough, the runtime supports _float_ enum underlying types, C# doesn't

PR #8519, opened 2019-08-26
[Mac] Use actual type from signature, float instead of int

Therzok:
SetContentHuggingPriorityForOrientation and SetContentCompressionResistancePriority both have `float priority`, so use that.

This prevents some conversions done at IL level.


Therzok:
Dupe PR

PR #8535, opened 2019-08-27
[Core] Fix null reference accessing target framework monikers

mrward:
Creating a .NET Core console project, then adding a new non-SDK style
C# library project, it was not possible to open the Edit References
dialog for the library project. The IDE log would have a null
reference exception:

```
  at System.Collections.Immutable.ImmutableArray`1[T].ThrowNullRefIfNotInitialized () [0x00000] in <6518d0e60ca84f6b915c07341a031a32>:0
  at System.Collections.Immutable.ImmutableArray`1[T].GetEnumerator () [0x00007] in <6518d0e60ca84f6b915c07341a031a32>:0
  at MonoDevelop.PackageManagement.PackageManagementCanReferenceProjectExtension.CanReferenceProject
```

The problem was that the project's TargetFramework setter is not
called when adding a new project to an existing solution so the
project's TargetFrameworkMonikers array was not initialized. This is
now handled.

Fixes VSTS #973344 - Unable to open Edit References dialog for new
project

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
@@ -98,6 +98,17 @@ namespace MonoDevelop.Projects
 				languageName = MSBuildProjectService.GetLanguageFromGuid (TypeGuid);
 		}
 
+		protected override void OnItemReady ()
+		{
+			// This is done here not in OnInitialize since it needs the extension chain initialized to
+			// be able to access the TargetFramework if it is not set. OnItemReady is always called both
+			// when loading an existing project or creating a new one.
+			if (!HasMultipleTargetFrameworks)
+				targetFrameworkMonikers = ImmutableArray.Create (TargetFramework.Id);
+
+			base.OnItemReady ();
+		}
+
 		ImmutableArray<TargetFrameworkMoniker> targetFrameworkMonikers;
 
 		void EvaluateTargetFrameworkMonikers ()
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
@@ -446,8 +446,8 @@ namespace MonoDevelop.Projects
 					return;
 				bool updateReferences = targetFramework != null;
 				targetFramework = value;
-				if (targetFrameworkMonikers.IsDefaultOrEmpty)
-					targetFrameworkMonikers = ImmutableArray.Create<TargetFrameworkMoniker> (targetFramework.Id);
+				if (!HasMultipleTargetFrameworks)
+					targetFrameworkMonikers = ImmutableArray.Create (targetFramework.Id);
 				if (updateReferences)
 					UpdateSystemReferences ();
 				NotifyModified ("TargetFramework");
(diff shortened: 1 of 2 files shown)


mrward:

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs
```diff
@@ -444,7 +444,13 @@ public string StripImplicitNamespace (string ns)
 		}
 
 		public ImmutableArray<TargetFrameworkMoniker> TargetFrameworkMonikers {
-			get { return targetFrameworkMonikers; }
+			get {
+				// Adding a new project to an existing solution will not call the TargetFramework's setter which
+				// initializes this array, so handle this here if the array is not initialized.
+				if (targetFrameworkMonikers.IsDefaultOrEmpty)
+					targetFrameworkMonikers = ImmutableArray.Create<TargetFrameworkMoniker> (TargetFramework.Id);
```
Therzok: Shouldn't we be doing this in `OnInitialize`?

Therzok: See https://github.com/mono/monodevelop/blob/dffcf18d0387f7930640d7fbd34314cd4952c5be/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs#L95

mrward: This does not work. OnInitialize is called too early before the extension chain is initialized so trying to use DotNetProject.TargetFramework at this point throws an exception.

```
  System.InvalidOperationException : The extension chain can't be used before OnExtensionChainInitialized() method is called
  at MonoDevelop.Projects.WorkspaceObject.AssertExtensionChainCreated () [0x00043] in projects/monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/WorkspaceObject.cs:379 
  at MonoDevelop.Projects.DotNetProject.get_ProjectExtension () [0x0000e] in projects/monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs:253 
  at MonoDevelop.Projects.DotNetProject.GetDefaultTargetFrameworkId () [0x00001] in projects/monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs:464 
  at MonoDevelop.Projects.DotNetProject.get_TargetFramework () [0x0000f] in projects/monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs:426 
  at MonoDevelop.Projects.DotNetProject.OnInitialize () [0x00061] in projects/monodevelop/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/DotNetProject.cs:97 
```

Will see if there is any other method that can be used instead.

mrward: OnItemReady might be something we can use instead. This is called when creating a new project and when loading an existing one. OnEndLoad is not called when creating a new project.

We may need to revisit what OnEndLoad is doing since this is initializing the file watcher - seems like that would not happen when adding a new C# project to an existing solution.

mrward: Failed to get them both in the same method. Multi-target framework information needs to be available before the extension chain gets involved otherwise various tests break. But I cannot do the same for single target framework projects - so that bit is done in OnItemReady.

Therzok (approved):
Looks great :+1:

slluis (approved):

PR #8538, opened 2019-08-27
[Ide] Remove some ref cycles, hopefully simplifying the object graph for leaks

Therzok:


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Shell/DefaultWorkbench.cs
```diff
@@ -77,6 +78,7 @@ internal partial class DefaultWorkbench : WorkbenchWindow, ICommandRouter, IInfo
 		List<SdiWorkspaceWindow> viewContentCollection = new List<SdiWorkspaceWindow> ();
 		Dictionary<PadCodon, IPadWindow> padWindows = new Dictionary<PadCodon, IPadWindow> ();
 		Dictionary<IPadWindow, PadCodon> padCodons = new Dictionary<IPadWindow, PadCodon> ();
+		static readonly ConditionalWeakTable<DockItem, PadWindow> itemToWindow = new ConditionalWeakTable<DockItem, PadWindow> ();
```
netonjm: We should clean or remove the association at some point?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Shell/DefaultWorkbench.cs
```diff
@@ -1425,37 +1427,56 @@ void AddPad (PadCodon padCodon, string placement, DockItemStatus defaultStatus)
 			item.DefaultVisible = false;
 			item.DefaultStatus = defaultStatus;
 			window.Item = item;
+
+			itemToWindow.Add (item, window);
```
netonjm: We should clean or remove the association at some point?

Therzok: ConditionalWeakTable is exactly to not need clearing associations. It uses weak keys for `item`, so this is automatically removed when a GC triggers and `item` is collected.

But definitely can add eager removal in `RemovePad`.

Therzok:
@slluis ping

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Shell/DefaultWorkbench.cs
```diff
@@ -1425,37 +1429,56 @@ void AddPad (PadCodon padCodon, string placement, DockItemStatus defaultStatus)
...
 		{
-			IPadWindow window = (IPadWindow) source;
-			if (!padCodons.ContainsKey (window)) 
+			var window = (PadWindow) source;
+			if (!window.Workbench.padCodons.TryGetValue (window, out PadCodon codon))
 				return;
-			PadCodon codon = padCodons [window];
-			DockItem item = GetDockItem (codon);
+			DockItem item = window.Workbench.GetDockItem (codon);
 			if (item != null) {
```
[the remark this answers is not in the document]
Therzok: GetDockItem requires an instance reference to DefaultWorkbench, which I'm trying to avoid because otherwise we can't make the event handler static, thus eliminate a strong reference path.

PR #8544, opened 2019-08-27
Fixes 940776: System.ArgumentOutOfRangeException on selection change in drop down

avodovnik:
If appears that in some circumstances, the snapshot changes as we get to calling the ActivateItem method on the provider, causing an argument out of range exception. This attempts to fix it by (1) boxing the offset to the current length of the current (valid) snapshot, and (2) catching any exception that would still flow to (this particular) UI element.

Fixes: https://vsmac.dev/940776

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/DropDownBoxListWindow.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/DropDownBoxListWindow.cs
@@ -30,6 +30,7 @@ using Gtk;
 using System.Text;
 using MonoDevelop.Components.AtkCocoaHelper;
 using MonoDevelop.Ide.Fonts;
+using MonoDevelop.Core;
 
 namespace MonoDevelop.Components
 {
@@ -91,9 +92,14 @@ namespace MonoDevelop.Components
 			list.SelectItem += delegate {
 				var sel = list.Selection;
 				if (sel >= 0 && sel < DataProvider.IconCount) {
-					DataProvider.ActivateItem (sel);
-					// This is so parent window of dropdown regains focus
-					TransientFor?.Present ();
+					try {
+						DataProvider.ActivateItem (sel);
+						// This is so parent window of dropdown regains focus
+						TransientFor?.Present ();
+					} catch (Exception ex) {
+						LoggingService.LogInternalError ("Offset seems to be out of sync with the snapshot.", ex);
+					}
+
 					Destroy ();
 				}
 			};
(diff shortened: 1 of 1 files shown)


main/src/addins/CSharpBinding/MonoDevelop.CSharp/CSharpPathedDocumentExtension.DataProvider.cs
```diff
@@ -241,8 +241,18 @@ public void ActivateItem (int n)
...
-				//FIXME: use the snapshot that the nodes came from
-				var point = new VirtualSnapshotPoint (editor.TextBuffer.CurrentSnapshot, offset);
+				// FIXME: use the snapshot that the nodes came from
+				var snapshot = editor.TextBuffer.CurrentSnapshot;
+				// HACK: it seems that before we fix the above, we seem to be hitting a bug
+				// where the offset is outside the snapshot's range, see: https://vsmac.dev/940776
+				if (offset > snapshot.Length || offset < 0) {
+					// even if that fixes the crash, we still want to log this!
+					LoggingService.LogInternalError (
+						$"Offset would have been called with a different number ({offset}) than the length of the snapshot ({snapshot.Length}).", new ArgumentOutOfRangeException ());
```
Therzok: This won't give us a stacktrace.

Therzok: Any exception that isn't thrown and bubbled has no stacktrace value. You could possible use a hack similar to [MarshalledObjCException](http://source.monodevelop.com/#MacPlatform/MacPlatform.cs,253)

avodovnik: I know. I wouldn't even pass the exception, but it's my understanding that we won't log anything other than this.

We don't really *need* a stack trace, as it's useless from this perspective. We know _how_ it gets called, but we won't know _why_. 

Do we (actually send) logs for any other call? If so, I'd change this.

Therzok: Then just `LogError` if you want log information. LogInternalError is used to Log + Telemetry.

avodovnik: So what you're saying is, that logging doesn't actually matter in this case. I'm half tempted to actually just drop that part - if we *still* get an exception, we'll log it later down the line anyway. We loose the count. Option two, is to keep this throwing an exception, which is what @sandyarmstrong suggested in the first place when I talked to him. When we go about fixing the actual cause, that'll help us equally, or more.

PR #8546, opened 2019-08-27
Check that SDK root path exists before versions

nosami:
Fixes VSTS #948292

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
@@ -190,6 +190,9 @@ namespace MonoDevelop.DotNetCore
 
 		IEnumerable<DotNetCoreVersion> GetInstalledSdkVersions (string sdkRootPath)
 		{
+			if (!Directory.Exists (sdkRootPath))
+				return Enumerable.Empty<DotNetCoreVersion> ();
+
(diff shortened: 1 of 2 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
```diff
@@ -190,6 +190,9 @@ bool CheckIsSupportedSdkVersion (string sdkDirectory)
 
 		IEnumerable<DotNetCoreVersion> GetInstalledSdkVersions (string sdkRootPath)
 		{
+			if (!Directory.Exists (sdkRootPath))
+				return Enumerable.Empty<DotNetCoreVersion> ();
+
```
mrward: If we are doing a Directory.Exists check here then we do not have do one in the constructor before GetInstalledSdkVersions() is called.

nosami: OK. I will remove the one in the constructor. The check here is needed.

nosami:
@monjenkins backport release-8.3

rodrmoya:
hmm, a related test is failing

nosami:
@rodrmoya I just checked with @mrward. The test is invalid so I removed it.

nosami:

nosami:

nosami:
@rodrmoya the failing dotnetcore test was due to cached SDK versions persisting in between tests. So the tests could fail depending on the order that they were run in. I guess this could have affected other branches too

PR #8550, opened 2019-08-28
[CSharp] Don't hide formatting dialog in ctor

sevoku:
The issue has already been fixed in MessageDialog, nevertheless
Hide, Show and ShowAll should never be called in the ctor
of a Gtk Dialog or Window. Even Hide will realize the window
and make it the key window on mac for the current UI iteration,
which will confuse MessageService and probably other logic
relying on a valid focused key window.

Fixes VSTS #943186


sevoku:

PR #8561, opened 2019-08-28
[Ide] If MEF is queried before startup handlers are run, throw an exc…

Therzok:
…eption

We should prevent extensions from slowing down the IDE startup if they request the composition before startup even occurs.
Mark anything after StartupHandlers as safe and only log a warning if queried too early

Fixes VSTS #973923 - Querying MEF in startup handlers should throw

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
@@ -315,10 +315,9 @@ namespace MonoDevelop.Ide
 			Counters.Initialization.Trace ("Running Startup Commands");
 			AddinManager.AddExtensionNodeHandler ("/MonoDevelop/Ide/StartupHandlers", OnExtensionChanged);
 
+			// Let extensions now access CompositionManager.Instance and start asynchronously composing the catalog
 			CompositionManager.ConfigureUninitializedMefHandling (throwException: false);
-
-			// Make sure the composition manager started initializing
-			Runtime.GetService<CompositionManager> ();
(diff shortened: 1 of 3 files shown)


Therzok:

Therzok:
Okay, perfect, tests failed. Need to change it so we opt-in for exception behaviour right before addin composition and opt out when command handlers are done.

Therzok:
Added 2 changes:
1. made the behaviour opt in and turn off after StartupHandlers are run
2. start composing mef earlier in the process initialization phase, at the same time we just start logging a warning

Therzok:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/Ide.cs
```diff
@@ -314,6 +314,10 @@ static void OnInitialized ()
 			// Startup commands
 			Counters.Initialization.Trace ("Running Startup Commands");
 			AddinManager.AddExtensionNodeHandler ("/MonoDevelop/Ide/StartupHandlers", OnExtensionChanged);
+
+			// Let extensions now access CompositionManager.Instance and start asynchronously composing the catalog
+			CompositionManager.ConfigureUninitializedMefHandling (throwException: false);
```
slluis: Shouldn't this be called before the above AddinManager.AddExtensionNodeHandler call? AddExtensionNodeHandler will cause startup handlers to be executed.

Therzok: Yes, we don't want addins to access mef before startup handlers are run.

PR #8565, opened 2019-08-28
[Mac] Fix not being able to open a file with external application

mrward:
[Mac] Handle filenames not just urls in GetApplications

MacPlatformService.GetApplications was using NSUrl.FromString which
works for hyperlinks but returns null for file that contains a space
character in its path. This was causing the open file dialog to
never close.

```
Parameter name: url
  at CoreServices.LaunchServices.GetDefaultApplicationUrlForUrl (Foundation.NSUrl url, CoreServices.LSRoles roles) [0x00030] in /Library/Frameworks/Xamarin.Mac.framework/Versions/5.16.1.17/src/Xamarin.Mac/CoreServices/LaunchServices.cs:97
  at MonoDevelop.MacIntegration.MacPlatformService.GetApplications (System.String filename) [0x00024] in /Users/vsts/agent/2.155.1/work/1/s/monodevelop/main/src/addins/MacPlatform/MacPlatform.cs:966
  at MonoDevelop.Ide.DesktopService.GetApplications (System.String filename) [0x00000] in /Users/vsts/agent/2.155.1/work/1/s/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/DesktopService.cs:139
  at MonoDevelop.Ide.Gui.DisplayBindingService.GetFileViewers (MonoDevelop.Core.FilePath filePath, MonoDevelop.Projects.Project ownerProject) [0x0017e] in /Users/vsts/agent/2.155.1/work/1/s/monodevelop/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DisplayBindingService.cs:131
```

Using NSUrl.FromString also meant that no applications were returned
for files when using the Open With context menu in the Solution pad.

To handle this a check is made to see if a url or a filename is
being passed then either NSUrl.FromString or NSUrl.FromFilename is
used.

[Mac] Fix not being able to open a file with external application

The Open With menu would launch the external application, such as
TextEdit.app, but not open the file. The problem was that the filename
was being converted using NSUrl.FromString which only works for
hyperlinks not files.

Fixes VSTS #970854 - GTC cannot open solutions via "open" button

### DIFF
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -963,11 +963,18 @@ namespace MonoDevelop.MacIntegration
 			checkUniqueName.Add ("MonoDevelop");
 			checkUniqueName.Add (BrandingService.ApplicationName);
 
-			var def = global::CoreServices.LaunchServices.GetDefaultApplicationUrlForUrl (NSUrl.FromString (filename));
+			NSUrl url = null;
+			if (Uri.TryCreate (filename, UriKind.Absolute, out Uri hyperlink) && !hyperlink.IsFile) {
+				url = NSUrl.FromString (filename);
+			} else {
+				url = NSUrl.FromFilename (filename);
+			}
+
+			var def = global::CoreServices.LaunchServices.GetDefaultApplicationUrlForUrl (url);
 
 			var apps = new List<DesktopApplication> ();
 
-			var retrievedApps = global::CoreServices.LaunchServices.GetApplicationUrlsForUrl (NSUrl.FromString (filename), global::CoreServices.LSRoles.All);
+			var retrievedApps = global::CoreServices.LaunchServices.GetApplicationUrlsForUrl (url, global::CoreServices.LSRoles.All);
 			if (retrievedApps != null) {
 				foreach (var app in retrievedApps) {
 					if (string.IsNullOrEmpty (app.Path) || !checkUniquePath.Add (app.Path))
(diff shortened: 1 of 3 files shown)


mrward:

mrward:

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -988,6 +990,14 @@ public override IEnumerable<DesktopApplication> GetApplications (string filename
 			return apps;
 		}
 
+		static NSUrl CreateNSUrl (string filename)
+		{
+			if (Uri.TryCreate (filename, UriKind.Absolute, out Uri hyperlink) && !hyperlink.IsFile)
```
Therzok: AspNetAppProjectFlavor passes in a relative url: http://source.monodevelop.com/#MonoDevelop.AspNet/Projects/AspNetAppProjectFlavor.cs,711

mrward: Mmm the problem with using 'test.html' is what is that? It could be a file or it could be a url. How can we know?

The problem here - using UriKind.RelativeOrAbsolute presumably will break the above logic. I just wonder how this worked before.

Therzok: My bad on being incomplete on the previous comment. I was saying we should probably change that to use `https://localhost` as [AspNetCoreProjectExtension does](http://source.monodevelop.com/#MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs,152)

mrward: Makes sense 👍

Right now it returns no apps using 'test.html'

mrward:

PR #8567, opened 2019-08-28
[Debugger] Fixed hang when showing the ExceptionCaughtDialog

jstedfast:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/968326/


mrward:
Verified. One of the debugger tests is failing but I am guessing it is unrelated since I doubt we have a test that shows the exception dialog.

PR #8569, opened 2019-08-28
[Test Tools] let unit tests support "Show Test Source Code"

ketttY:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/973891/

### DIFF
--- a/main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/VsTestTestClass.cs
+++ b/main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/VsTestTestClass.cs
@@ -23,16 +23,15 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
-
+using System;
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 using MonoDevelop.Core.Execution;
+using MonoDevelop.Ide;
+using MonoDevelop.Ide.TypeSystem;
 using MonoDevelop.Projects;
 using MonoDevelop.UnitTesting;
-using MonoDevelop.Ide;
-using System;
-using MonoDevelop.Ide.TypeSystem;
-using System.Linq;
 
 namespace MonoDevelop.UnitTesting.VsTest
 {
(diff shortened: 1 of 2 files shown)


main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/VsTestTestClass.cs
```diff
@@ -44,6 +49,25 @@ public VsTestTestClass (IVsTestTestRunner testRunner, Project project, VsTestUni
 			this.testRunner = testRunner;
 			FixtureTypeName = vsTestUnit.FixtureTypeName;
 			TestSourceCodeDocumentId = string.IsNullOrEmpty (vsTestUnit.FixtureTypeNamespace) ? FixtureTypeName : vsTestUnit.FixtureTypeNamespace + "." + FixtureTypeName;
+
+			if (sourceCodeLocation == null) {
+				IdeApp.TypeSystemService.GetCompilationAsync (Project).ContinueWith ((t) => {
```
mrward: Not sure we need the null check here. It will always be null at this point so I think we can remove it.

decriptor:
Does this also fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/961502/

decriptor:
Can this also be unit tested?

mrward:
@decriptor This fixes go to source for a class in the Unit Tests window for SDK style projects (e.g. xUnit, MSTest). Looking at your screen recording I would say it does not fix the problem you are seeing since it looks like you are using go to source for a test method.

mrward:
@manish Your test is a method - this change does not change the behaviour for test methods. This fix is when going to source for a test class.

main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/VsTestTestClass.cs
```diff
@@ -44,6 +49,21 @@ public VsTestTestClass (IVsTestTestRunner testRunner, Project project, VsTestUni
 			this.testRunner = testRunner;
 			FixtureTypeName = vsTestUnit.FixtureTypeName;
 			TestSourceCodeDocumentId = string.IsNullOrEmpty (vsTestUnit.FixtureTypeNamespace) ? FixtureTypeName : vsTestUnit.FixtureTypeNamespace + "." + FixtureTypeName;
+
+			IdeApp.TypeSystemService.GetCompilationAsync (Project).ContinueWith ((t) => {
```
Therzok: What's the threading story here? Where do we want this continuation to run? We should be explicit about this.

Also, since the class is disposable, we should also handle cancellation and not do any work if GetCompilationAsync runs after Dispose is done.

* Create a CancellationTokenSource which is Cancelled and Disposed on VsTestClass.Dispose.
* Pass the source's` CancellationToken` to `GetCompilationAsync`. Prefer setting it to a local variable first.
* Check for `token.IsCancellationRequested` in `ContinueWith` to avoid running after dispose.
* Specify an explicit task scheduler to run on to `ContinueWith`. `TaskScheduler.Default` for a background thread, `Runtime.MainTaskScheduler` for the UI thread.

mrward: All good questions. This code is based on the existing code in [VsTestUnitTest](https://github.com/mono/monodevelop/blob/8d7529292e9cdf9dcd8a9752494b42771d0323d8/main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/VsTestUnitTest.cs#L68) which David added.

PR #8571, opened 2019-08-29
Fix document focus issue

slluis:
When grabbing the focus of a view, the actual grab is done after
dispatching the event queue. It may happen that after the events are
executed, the view bacame hidden (for example, maybe an attached view
was set active). In that case, don't try to grab the focus since it
maye cause an unexpected re-activation of the view.

Fixes VSTS #963585 – Diff/View commands broken in SolutonPad


slluis:

PR #8585, opened 2019-08-30
[VCS] VSEditor uses 0 based indices, so adjust for that

Therzok:


Therzok:

PR #8590, opened 2019-08-30
[Git] Try to avoid libgit2 crashing on Dispose

sevoku:
This will probably not solve the crasher. We really need
to use a single threaded scheduler for all git operations.

Probably doesn't fix VSTS #975668


sevoku:

PR #8605, opened 2019-09-02
[Core] Refactor FileWatcherWrapper code to use 2 watchers per FSW

Therzok:
This is to allow us to validate whether an event is associated with
a file or a directory without having to do additional IO.

Also refactor a bit of the implementation to reuse FileEventArgs where
possible

Fixes VSTS #934618 - Possible optimization of FileSystemWatcher by using 2 watchers per actual watched path


mrward:
Code looks OK. Just wondering what two file watchers per directory will mean for file descriptors. Presumably this means we are using twice as many file descriptors for a particular directory being monitored.

Therzok:
@mrward yep, that's exactly what it means. The benefits here (knowing if we get file/directory notifications) outweigh the costs, imo.

WIth the handle leak on corefx side fixed, this shouldn't hit us as hard. 

I opened https://github.com/dotnet/corefx/issues/40751 for a way to fix this without this overhead, but it'll depend on corefx folks.

And when we migrate to native FSW with multiple root dirs passed as an argument, we can swap to another implementation.

mrward (approved):

Therzok:
Consider bumping file handle limit before merging this change. Related user story:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/992648

The disadvantage of increasing the handle limit is that it makes multiplexing syscalls like `select` have to traverse a larger set of handles every time, which may have a negative impact on performance.

sevoku (approved):

Therzok:
Adding do not merge until https://openradar.appspot.com/radar?id=5031201551155200 has a resolution.

Therzok:
This is obsolete with the new FSW implementation, we can probably apply a different kind of optimziation there.

PR #8606, opened 2019-09-02
[Core] Prefer Compile items over None items

mrward:
Project.GetSourceFilesAsync will add the Compile items it gets from
CoreCompileDependsOn to the HashSet before adding the evaluated items.
This prevents None items being used instead of Compile items where
an SDK project does not remove duplicate None items. A project that
does the equivalent of the following would result in code completion
not showing target frameworks in the text editor's navigation bar
for a multi-target project.

    <Compile Remove="**\*.cs" />
    <None Include="src\**\*.cs" />
    <Compile Include="src\Class1.cs" />

The None item was added to the HashSet first. The Compile item added
afterwards would be ignored.

Fixes VSTS #974214 - Multi-targeting: No target framework selector
shown in editor
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/974214

Other ideas here might be:

1. Always add None items last to the HashSet.
2. Provide a way for callers of Project.GetSourceFilesAsync to pass in a filter so the files can be filtered before the HashSet is used.
3. Provide a custom type system specific method (e.g. GetCompileSourceFilesAsync) that returns the files it needs and move all the filtering logic to this method. Problem here might be how to handle Projections.
4. Change the hash key to be build action + filename.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -532,8 +532,10 @@ namespace MonoDevelop.Projects
 			var coreCompileResult = await compileEvaluator.GetItemsFromCoreCompileDependenciesAsync (this, monitor, configuration);
 			var evaluatedCompileItems = coreCompileResult.SourceFiles;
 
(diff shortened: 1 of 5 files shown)


Therzok (approved):

mrward:

slluis (approved):

PR #8614, opened 2019-09-03
Switch from int to long

Therzok:
Interlocked has poor codegen for non-word size, so fix that.


mrward (approved):

Therzok:
@slluis ping

sevoku (approved):
This is also part of VSTS #977400 - Review usages of Interlocked ?

Therzok:
Yep.

PR #8618, opened 2019-09-03
CSharpCompilerParameters: Add support for "nullable"

sandyarmstrong:
This ensures that the editor and language service are aware of nullable context settings at the project level.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/970026

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CSharpCompilerParameters.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CSharpCompilerParameters.cs
@@ -81,6 +81,9 @@ namespace MonoDevelop.CSharp.Project
 		[ItemProperty ("WarningsNotAsErrors", DefaultValue = "")]
 		string warningsNotAsErrors = "";
 
+		[ItemProperty ("Nullable", DefaultValue = "")]
+		string nullableContextOptions = "";
+
 		protected override void Write (IPropertySet pset)
 		{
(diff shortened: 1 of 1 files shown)


main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CSharpCompilerParameters.cs
```diff
@@ -218,6 +222,35 @@ public override ParseOptions CreateParseOptions (DotNetProjectConfiguration conf
 			);
 		}
 
+		public NullableContextOptions NullableContextOptions {
+			get {
+				switch (nullableContextOptions.ToLower ()) {
```
Therzok: Maybe this would be [simpler](https://docs.microsoft.com/en-us/dotnet/api/system.enum.tryparse?view=netframework-4.8#System_Enum_TryParse__1_System_String_System_Boolean___0__)?

Not a dealbreaker, just less code.

sandyarmstrong: I considered that, but figured I don't have a guarantee that future enum items will be serialized in that way. I basically copied what's going on in https://github.com/dotnet/roslyn/blob/5010ecae8b4890e4a655959e2b2612c33eb2bd7c/src/Compilers/CSharp/Portable/CommandLine/CSharpCommandLineParser.cs#L341 . I suspect they might have done it that way because there used to be other valid options (like `safeonly` and `safeonlywarnings`), but I'm not really sure.

This seemed like a safer approach for now. But I'm not married to it.

sandyarmstrong:

PR #8619, opened 2019-09-03
Use a single-thread scheduler for all libgit2 operations 

sevoku:
This should solve most of the threading issues with libgit2.

Fixes VSTS #972329
Fixes VSTS #974907
Fixes VSTS #975668

**PR Review assistance:**

- Single Thread Scheduler: 2eaaf0c21d487cef4f936f983fdd0e65317ae9be
  - Tests: 4c4a19ccbd2f3a24dcd35569db09f315e853721e
  - BG thread: 1a303d5593e8b25b51693a835ad37cf0625f423c
- Git using single thread: e09e9d3450921b9d994c0ca541d21e8bb1c7bfb1, 0a2a352b7530797d70437a600651a836a1685fc5, 194d6f40e030d217d101dd0d0333be258e9b92f9
- FSW / Git lockfiles related fixes: 9cb3134bf5d70c93698cb5cee8e236d7ce0f731d, 260396f3fd0f8e7b789e0c271953d4228251745f and the **most important**: 2ca7910f0d5487899036628a0b3ca2d828ed6775
- GIT API mostly async to avoid UI deadlocks with the single thread scheduler: 90537534481b007be5d8c22d133a0265ef983345

**Important note:** not all UIs are really async right now (see 90537534481b007be5d8c22d133a0265ef983345), to make it kind of work properly I just replaced the sync calls with async.Wait at several places. We need to get this in ASAP, since 8.3 is crashin all over the place. the sync calls were crashing too, so if async.Wait() doesn't work (likely blocks rather than crashes), it'd not be a merge blocker, because other fixes here a real.

PR #8620, opened 2019-09-03
[Git] Make most GitRepository APIs async

sevoku:
Based on https://github.com/mono/monodevelop/pull/8619, this changes most GitRepository APIs to asnyc with appropriate handling from UIs to avoid deadlocks.


sevoku:
NOTE: Not all UIs are really async right now, to make it kind of work properly I just replaced the sync calls with `async.Wait` at several places. We need to get this in ASAP, since 8.3 is crashin all over the place. the sync calls were crashing too, so if `async.Wait()` doesn't work (maybe blocks rather than crashes), it'd not be a merge blocker, because other fixes here a real.

Therzok (approved):

Therzok (approved):

PR #8625, opened 2019-09-04
[DotNetCore] Add .NET Core runtime "component"

rodrmoya:


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreRuntimeSystemInformation.cs
```diff
@@ -0,0 +1,84 @@
...
+		public override string Description => GetDescription ();
+
+		public override string Version => DotNetCoreRuntime.Versions.FirstOrDefault ()?.ToString ();
+
+		public override string ApplicationId => "392e9bd0-7214-4d2f-8b38-420b38e3b20f";
+
+		public override UpdateInfo GetUpdateInfo ()
+		{
+			var latestRuntimeVersion = DotNetCoreRuntime.Versions.FirstOrDefault ();
+			if (latestRuntimeVersion != null) {
```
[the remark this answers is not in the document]
nosami: This logic needed to change to force the download of 2.2 runtime

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreRuntimeSystemInformation.cs
```diff
@@ -0,0 +1,89 @@
...
+		public override UpdateInfo GetUpdateInfo ()
+		{
+			var latestRuntimeVersion = 
+				DotNetCoreRuntime.Versions.FirstOrDefault(v => v.Major == 2 && v.Minor >= 2);
+			if (latestRuntimeVersion != null) {
+				return new UpdateInfo (ApplicationId, latestRuntimeVersion.Version);
+			}
+
+			// Force the install of the 2.2 runtime
+			return new UpdateInfo (ApplicationId, versionId: 0);
```
mrward: I think we should return null if .NET core is not installed at all. Return the 0 version if .NET Core installed.

Should be able to do that with the DotNetCoreRuntime.IsInstalled.

Otherwise we will be forcing it on users who never installed any dotnet cli and do not need it.

nosami: I think in this case we should merge the infobar PR too https://github.com/xamarin/md-addins/pull/5326

nosami:

PR #8631, opened 2019-09-04
[Ide] Rework GetWorkspaceAsync to cancel requests on any solution dispose

Therzok:
Refactor the current code to move the workspace registrations from TypeSystemService
into Solution.ExtendedProperties.

This way, we can keep track of all the registrations done for a solution, and cancel
all of them whenever a solution is disposed.

The second part of the PR is splitting each request into its own TaskCompletionSource.
Reusing the same task has unintended side effects due to the cancellation token being
passed as a parameter.

Using the same task meant that if some CancellationTokenSource decided to cancel its
request, it would cancel all the current active requests on the solution

Fixes VSTS #977144 - Leak detected on cleanup iteration in MonoDevelop.Projects.Solution class
Fixes VSTS #977335 - Quality of local testing

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
@@ -362,7 +362,8 @@ namespace MonoDevelop.Ide.TypeSystem
 
 			TypeSystemService.EnableSourceAnalysis.Changed -= OnEnableSourceAnalysisChanged;
 			TypeSystemService.Preferences.FullSolutionAnalysisRuntimeEnabledChanged -= OnEnableFullSourceAnalysisChanged;
-			desktopService.MemoryMonitor.StatusChanged -= OnMemoryStatusChanged;
+			if (desktopService !=null)
+				desktopService.MemoryMonitor.StatusChanged -= OnMemoryStatusChanged;
 
 			if (workspace != null) {
 				workspace.ActiveConfigurationChanged -= HandleActiveConfigurationChanged;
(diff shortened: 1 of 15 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -361,7 +361,8 @@ protected override void Dispose (bool finalize)
 
 			TypeSystemService.EnableSourceAnalysis.Changed -= OnEnableSourceAnalysisChanged;
 			TypeSystemService.Preferences.FullSolutionAnalysisRuntimeEnabledChanged -= OnEnableFullSourceAnalysisChanged;
-			desktopService.MemoryMonitor.StatusChanged -= OnMemoryStatusChanged;
+			if (desktopService !=null)
```
Therzok:
```suggestion
			if (desktopService != null)
```

Therzok:
Commits which modify test infrastructure were needing for tests to pass on my machine.

Therzok:
Consistently fixes leak tests

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService_WorkspaceRequestRegistration.cs
```diff
@@ -0,0 +1,84 @@
...
+					}
+				}
+			}
+
+			public void Dispose ()
+			{
+				// Requests are removed when canceled.
+				src.Cancel ();
+				src.Dispose ();
+				src = new CancellationTokenSource ();
```
mrward: Should we be creating a new CancellationTokenSource when the registration is disposed?

I was assuming this was maybe to handle something calling GetWorkspaceAsync after the Dispose is called but if the registration is disposed would we not want all calls to GetWorkspaceAsync fail with a cancellation exception? The registration Dispose, without it being removed from the extended properties, happens when unloading the workspace item.

Seems like we would never dispose this other instance of the CancellationTokenSource - not sure if that is a problem or not.

Therzok: Can change it. It was to allow us to clear the registration from both MonoDevelopWorkspace.Dispose and Solution.Dispose.

Therzok: But that might be a better way to handle it.

Therzok:
@slluis ping

PR #8636, opened 2019-09-05
[DotNetCore] Only attach FileChanged handler if we have a globalJsonPath

Therzok:
Nothing else sets this value so it's safe to assume that we need it set to non-null for the handler to have any purpose.

This should speed up headless test runs for dotnetcore projects when building

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
@@ -283,7 +283,6 @@ namespace MonoDevelop.DotNetCore
 		protected override void OnItemReady ()
 		{
 			base.OnItemReady ();
-			FileService.FileChanged += FileService_FileChanged;
 
 			if (!IdeApp.IsInitialized)
 				return;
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -283,7 +283,6 @@ async Task OnExecuteDotNetCoreCommand (ProgressMonitor monitor, ExecutionContext
 		protected override void OnItemReady ()
 		{
 			base.OnItemReady ();
-			FileService.FileChanged += FileService_FileChanged;
```
rodrmoya: hmm, IIRC you mentioned moving this code to use the Project's events instead, which was much better than handling `FileService` events?

mrward: Not sure that will work since the global.json file is not inside the project.

Therzok:
Yeah, need a better solution in FileWatcherService to watch singular files. Not sold on this, CI reports no improvements in perf tests/test runs

Therzok:
Closing, it made no difference.

PR #8637, opened 2019-09-05
[Ide] Switch command activation to use a stopwatch instead of DateTime

Therzok:
First of all, DateTime.Now is culture aware, and applies culture information to actually get the value, so it's really slow.
Secondly, it's really low resolution. Use Stopwatch to get better elapsed time for command activation


Therzok:
@slluis ping

PR #8638, opened 2019-09-05
Bump some nugets

Therzok:


decriptor:
@garuma This is probably something we should bump too?

Therzok:
@garuma I only bumped the analyzers, not vs-threading itself.

mrward (approved):
We should update md-addins too.

rodrmoya (approved):

PR #8642, opened 2019-09-05
Fixes IndexOutOfRange trying to get current tab by index in tabstrip

netonjm:
Fixes VSTS # 968261 - [FATAL] System.ArgumentOutOfRangeException exception in Gtk.Application.Run() "Index was out of range. Must be non-negative and less than the size of the collection. Parameter name: index"

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Tabstrip.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Tabstrip.cs
@@ -37,7 +37,6 @@ using MonoDevelop.Ide.Gui;
 using MonoDevelop.Ide.Fonts;
 using MonoDevelop.Ide;
 using MonoDevelop.Ide.Gui.Shell;
-using Gdk;
 
 namespace MonoDevelop.Components
 {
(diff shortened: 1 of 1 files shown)


mrward:
If the exception is in GetBounds then it seems to me that the tabs collection can only change if another thread is modifying the tabstrip at the same time this method is being called. Presumably only the UIThread is used when GetBounds is called. Presumably the TabStrip should not be called on a non-UI thread. So not sure this is the right fix here.

mrward:
I guess idx could be -1 if the tab is not in the set of tabs. Which would cause an out of range since the loop would never finish. The pubilc RemoveTab and InsertTab methods call GetBounds but the exception callstack is from OnMotionNotifyEvent which presumably has to be from the UI thread and I cannot see how the tab could not be in the tabs list when OnMotionNotifyEvent calls GetBounds unless a non-UI thread is changing the tabs.

mrward:
Mmm, OnMotionNotifyEvent does use a hoverTab which is passed to GetBounds. Is that maybe no longer in the tabs list?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Tabstrip.cs
```diff
@@ -80,7 +87,7 @@ class Tabstrip : DrawingArea
 		public Tabstrip ()
 		{
 			Accessible.SetRole (AtkCocoa.Roles.AXTabGroup);
-			Events |= Gdk.EventMask.ButtonPressMask | Gdk.EventMask.PointerMotionMask | Gdk.EventMask.LeaveNotifyMask | Gdk.EventMask.FocusChangeMask;
+			Events |= Gdk.EventMask.ButtonPressMask | Gdk.EventMask.PointerMotionMask | EventMask.EnterNotifyMask |  Gdk.EventMask.LeaveNotifyMask | Gdk.EventMask.FocusChangeMask;
```
mrward: Could use Gdk.EventMask.EnterNotifyMask to be consistent with the rest. Also could then remove the `using Gdk;` statement.

netonjm: done!

netonjm:
rebased

netonjm:
rebased!

PR #8647, opened 2019-09-05
Workspace: Prevent DocumentIds from changing on configuration change

Therzok:
It's possible for events like device (dis)connect to make several quick
calls to `ActiveConfigurationChanged`, some of which would be canceled,
ultimately causing old `DocumentId` data to get lost during project
reload.

Creating new `DocumentId`s could break multiple Roslyn services such as
intellisense. This change makes the whole process more robust so that
`DocumentId` data does not get lost.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/915494

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectData.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectData.cs
@@ -46,11 +46,17 @@ namespace MonoDevelop.Ide.TypeSystem
 				this.projectId = projectId;
 				workspaceRef = new WeakReference<MonoDevelopWorkspace> (ws);
 				DocumentData = new DocumentMap (projectId);
-				this.metadataReferences = new List<MonoDevelopMetadataReference> (metadataReferences.Length);
+				this.metadataReferences = new List<MonoDevelopMetadataReference> (metadataReferences);
+			}
 
-				lock (this.metadataReferences) {
+			internal void Connect ()
+			{
+				if (!workspaceRef.TryGetTarget (out var ws))
+					return;
+
+				lock (metadataReferences) {
 					foreach (var metadataReference in metadataReferences) {
-						AddMetadataReference_NoLock (metadataReference, ws);
+						metadataReference.SnapshotUpdated += OnMetadataReferenceUpdated;
 					}
 				}
 			}
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectDataMap.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectDataMap.cs
@@ -202,14 +202,33 @@ namespace MonoDevelop.Ide.TypeSystem
 				}
 			}
 
-			internal ProjectData CreateData (ProjectId id, ImmutableArray<MonoDevelopMetadataReference> metadataReferences)
+			/// <summary>
+			/// Construct, connect to, and return a new <see cref="ProjectData"/> based on
+			/// <paramref name="metadataReferences"/>. This replaces any old <see cref="ProjectData"/> for
+			/// <paramref name="id"/>.
+			/// </summary>
+			internal ProjectData ReplaceData (ProjectId id, ImmutableArray<MonoDevelopMetadataReference> metadataReferences, out ProjectData oldData)
+			{
+				var result = new ProjectData (id, metadataReferences, Workspace);
+				ReplaceData (id, result, out oldData);
+				return result;
+			}
+
+			/// <summary>
+			/// Remove the old <see cref="ProjectData"/> for <paramref name="id"/>, replace it with
+			/// <paramref name="newData"/>, and and connect to <paramref name="newData"/>.
+			/// </summary>
+			internal void ReplaceData (ProjectId id, ProjectData newData, out ProjectData oldData)
 			{
 				lock (updatingProjectDataLock) {
-					var result = new ProjectData (id, metadataReferences, Workspace);
-					projectDataMap [id] = result;
-					return result;
+					oldData = RemoveData (id);
+					if (newData != null) {
+						newData.Connect ();
+						projectDataMap [id] = newData;
+					}
 				}
 			}
+
 			internal ProjectId[] GetProjectIds ()
 			{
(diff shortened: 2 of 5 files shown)


Therzok:
Waiting for @cobey to confirm if https://github.com/mono/monodevelop/pull/8648 fixes it. seems like fix is not needed in master?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectData.cs
```diff
@@ -46,11 +46,17 @@ public ProjectData (ProjectId projectId, ImmutableArray<MonoDevelopMetadataRefer
 				this.projectId = projectId;
 				workspaceRef = new WeakReference<MonoDevelopWorkspace> (ws);
 				DocumentData = new DocumentMap (projectId);
-				this.metadataReferences = new List<MonoDevelopMetadataReference> (metadataReferences.Length);
+				this.metadataReferences = new List<MonoDevelopMetadataReference> (metadataReferences);
```
DavidKarlas: Is this changed needed to fix bug?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectData.cs
```diff
@@ -46,11 +46,17 @@ public ProjectData (ProjectId projectId, ImmutableArray<MonoDevelopMetadataRefer
...
+			internal void Connect ()
+			{
+				if (!workspaceRef.TryGetTarget (out var ws))
+					return;
 
-				lock (this.metadataReferences) {
+				lock (metadataReferences) {
 					foreach (var metadataReference in metadataReferences) {
-						AddMetadataReference_NoLock (metadataReference, ws);
+						metadataReference.SnapshotUpdated += OnMetadataReferenceUpdated;
```
DavidKarlas: Is this changed needed to fix bug?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectSystemHandler.cs
```diff
@@ -139,22 +139,9 @@ internal static IEnumerable<string> GetFrameworks (MonoDevelop.Projects.Project
...
-
-					var documents = await CreateDocuments (projectData, p, token, cacheInfo.SourceFiles, oldProjectData).ConfigureAwait (false);
-					if (documents == null)
-						return null;
-
-					mainDocuments = documents.Item1;
-					additionalDocuments = documents.Item2;
-				} finally {
-					workspace.LoadLock.Release ();
+				var (mainDocuments, additionalDocuments) = await ReplaceProjectDataAsync (projectId, p, cacheInfo.References, cacheInfo.SourceFiles, token);
```
DavidKarlas: Why move to different method? Makes reviewing change hard...

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectDataMap.cs
```diff
@@ -202,14 +202,25 @@ internal ProjectData RemoveData (ProjectId id)
...
 
-			internal ProjectData CreateData (ProjectId id, ImmutableArray<MonoDevelopMetadataReference> metadataReferences)
+			internal ProjectData ReplaceData (ProjectId id, ImmutableArray<MonoDevelopMetadataReference> metadataReferences, out ProjectData oldData)
+			{
+				var result = new ProjectData (id, metadataReferences, Workspace);
+				oldData = ReplaceData (id, result);
+				return result;
+			}
+
+			internal ProjectData ReplaceData (ProjectId id, ProjectData newData)
```
sandyarmstrong: These `ReplaceData` methods both need some doc comments, as it's not obvious from the signature what they return. I'd also consider making the second method return `void` and have an `out` parameter so that they both return `oldData` in the same way.

sandyarmstrong:
@backport release-8.2

sandyarmstrong:

sandyarmstrong:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectData.cs
sandyarmstrong: It makes sense, because of the event hookups moving to `Connect`, and otherwise there's no reference to the `metadataReferences` parameter.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectData.cs
sandyarmstrong: Because of how the list field is setup, we don't want to call `AddMetadataReference_NoLock` anymore because it would double-add everything to the list.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectSystemHandler.cs
sandyarmstrong: Undid method move to make diff easier here, since the idea is to backport to 8.2 and fewer changes are better.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectDataMap.cs
sandyarmstrong: Fixed.

sandyarmstrong:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.ProjectSystemHandler.cs
Therzok: Sorry, this was to actually test the method, but never got to get the test to fail so I didn't push that part.

PR #8649, opened 2019-09-06
[AspNet*] Filter HTTP handlers to viewer role

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/970420

This still shows iTerm as a browser, but this is the same the OS shows:

image

So, as discussed on the bug, provided this is what the OS shows, do the same on our side.

### DIFF
--- a/main/src/addins/MacPlatform/MacPlatform.cs
+++ b/main/src/addins/MacPlatform/MacPlatform.cs
@@ -942,6 +942,28 @@ namespace MonoDevelop.MacIntegration
 
 		public override IEnumerable<DesktopApplication> GetApplications (string filename)
 		{
+			return GetApplications (filename, DesktopApplicationRole.All);
+		}
+
+		public override IEnumerable<DesktopApplication> GetApplications (string filename, DesktopApplicationRole role)
+		{
+			static global::CoreServices.LSRoles ToLSRoles (DesktopApplicationRole role)
+			{
+				var lsRole = global::CoreServices.LSRoles.None;
+
+				if (role == DesktopApplicationRole.All)
+					return global::CoreServices.LSRoles.All;
+
+				if (role.HasFlag (DesktopApplicationRole.Viewer))
+					lsRole |= global::CoreServices.LSRoles.Viewer;
+				if (role.HasFlag (DesktopApplicationRole.Editor))
+					lsRole |= global::CoreServices.LSRoles.Editor;
+				if (role.HasFlag (DesktopApplicationRole.Shell))
+					lsRole |= global::CoreServices.LSRoles.Shell;
+
+				return lsRole;
+			}
+
 			//FIXME: we should disambiguate dupliacte apps in different locations and display both
 			//for now, just filter out the duplicates
 			var checkUniqueName = new HashSet<string> ();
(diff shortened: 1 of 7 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/PlatformService.cs
```diff
@@ -343,7 +343,7 @@ public virtual void StartUpdatesInstaller (FilePath installerDataFile, FilePath
 		{
 		}
 
-		public virtual IEnumerable<DesktopApplication> GetApplications (string filename)
+		public virtual IEnumerable<DesktopApplication> GetApplications (string filename, DesktopApplicationRole role = DesktopApplicationRole.All)
```
mrward: This looks like an API break. We should add a separate method with this parameter.

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -940,8 +940,24 @@ public override void OpenTerminal (FilePath directory, IDictionary<string, strin
...
+					return global::CoreServices.LSRoles.None;
+				case DesktopApplicationRole.Viewer:
+					return global::CoreServices.LSRoles.Viewer;
+				case DesktopApplicationRole.Editor:
+					return global::CoreServices.LSRoles.Editor;
+				case DesktopApplicationRole.Shell:
+					return global::CoreServices.LSRoles.Shell;
+				default:
+					return global::CoreServices.LSRoles.All;
+				}
```
mrward: The enum is a flags but we are using a switch here. So I could not do `DesktopApplicationRole.Viewer | DesktopApplicationRole.Editor` if I wanted to do that. Is that possible with LSRoles?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/PlatformService.cs
rodrmoya: That’s why I set the new argument as optional, but yes I guess it’s still a binary API break

main/src/addins/MacPlatform/MacPlatform.cs
rodrmoya: LSRoles is indeed a flags, so changing it

rodrmoya: Fixed now

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Desktop/PlatformService.cs
rodrmoya: Fixed now

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/DesktopService.cs
```diff
@@ -48,7 +48,7 @@ public class DesktopService : Service
 		PlatformService PlatformService {
 			get {
 				if (platformService == null)
-					throw new InvalidOperationException ("Not initialized");
+					throw new InvalidOperationException ("Not ingetitialized");
 				return platformService;
```
mrward: Typo?

rodrmoya: Yeah, fixed

rodrmoya:

Therzok:
I'm wondering whether we should just pass Viewer by default.

PR #8651, opened 2019-09-06
Disable "Show test source code" menu option when souceCodeLocation is…

ketttY:
… null

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/978849


main/src/addins/MonoDevelop.UnitTesting/Gui/TestNodeBuilder.cs
```diff
@@ -213,7 +213,14 @@ protected async void OnShowTest ()
 			if (loc != null)
 				await IdeApp.Workbench.OpenDocument (loc.FileName, null, loc.Line, loc.Column);
 		}
-		
+
+		[CommandUpdateHandler (TestCommands.ShowTestCode)]
+		protected void OnUpdateShowTest (CommandInfo info)
+		{
+			UnitTest test = CurrentNode.DataItem as UnitTest;
+			info.Enabled = test.SourceCodeLocation != null;
```
mrward: Probably should handle test being null here. `test?.SourceLocationLocation`

PR #8653, opened 2019-09-06
Fixes System.AggregateException trying to change a label from a tabview

netonjm:
Fixes VSTS #978857 - [HexEditor] View failed to load. System.AggregateException:EnsureUIThread

### DIFF
--- a/main/src/addins/MonoDevelop.HexEditor/MonoDevelop.HexEditor/HexEditorView.cs
+++ b/main/src/addins/MonoDevelop.HexEditor/MonoDevelop.HexEditor/HexEditorView.cs
@@ -52,7 +52,8 @@ namespace MonoDevelop.HexEditor
 
 		protected override async Task<Control> OnGetViewControlAsync (CancellationToken token, DocumentViewContent view)
 		{
-			view.Title = GettextCatalog.GetString ("Binary");
+			//title from document should be changed from ui thread
+			await Runtime.RunInMainThread (delegate { view.Title = GettextCatalog.GetString ("Binary"); });
 			hexEditor = new Mono.MHex.HexEditor ();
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.HexEditor/MonoDevelop.HexEditor/HexEditorView.cs
```diff
@@ -52,7 +52,8 @@ class HexEditorView : FileDocumentController, IUndoHandler, IBookmarkBuffer, IZo
 
 		protected override async Task<Control> OnGetViewControlAsync (CancellationToken token, DocumentViewContent view)
 		{
-			view.Title = GettextCatalog.GetString ("Binary");
+			//title from document should be changed from ui thread
+			await Runtime.RunInMainThread (delegate { view.Title = GettextCatalog.GetString ("Binary"); });
```
Therzok: Why is this not on the UI thread?

Therzok: All this code should be run on the UI thread.

Therzok:
Discussed with @netonjm on Slack. He investigated with @mrward  and it seems that the thread switches here:
https://github.com/mono/monodevelop/blob/f9839cdd26fd7710315b7ad7fcb7bc7b585223c9/main/src/addins/MonoDevelop.HexEditor/Mono.MHex.Data/Buffer.cs#L87-L89

Blocking until a bug is opened in Mono side, as this will just fix a symptom of an underlying bug, not the root cause.

netonjm:
Closing this because wrong fix. Seems to be a mono issue with the Synch context

PR #8661, opened 2019-09-06
Fix 980541: Switching themes in Preferences corrupts colors in editor 

DavidKarlas:
Using reflection is never pretty way to fix a bug, but themeing/coloring API of VS is just too old and ugly...
So for now this is solution, because alternative is so much code and potential bugs I would rather not do that...

### DIFF
--- a/main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/ThemeToClassification.cs
+++ b/main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/ThemeToClassification.cs
@@ -182,6 +182,15 @@ namespace MonoDevelop.TextEditor
 		{
 			var editorFormat = editorFormatMapService.GetEditorFormatMap (appearanceCategory);
 			editorFormat.BeginBatchUpdate ();
+
+			// Reason we do it this way to purge existing values is because there is many possible
+			// value that can be already in ResourceDictionary, Bold, Color, Font... So we would have to explicitly
+			// clear all this values and set them to default before setting new theme, since theme might depend on default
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/ThemeToClassification.cs
```diff
@@ -182,6 +182,10 @@ void UpdateEditorFormatMap (string appearanceCategory)
 		{
 			var editorFormat = editorFormatMapService.GetEditorFormatMap (appearanceCategory);
 			editorFormat.BeginBatchUpdate ();
+			try {
```
abock: Add a comment explaining why this is necessary and add the link to the bug in the comment and PR/commit message.

abock: Should also probably log the exception here in case the reflection breaks in the future.

Therzok: @abock maybe just add a test to verify the reflection endpoint works.

abock:
Still needs a bug and comment

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/ThemeToClassification.cs
abock: No need for a test now. I added `ClearProperties` to the editor.

abock:

PR #8662, opened 2019-09-07
[release-8.2] Bump android reference to 9.4.1.1

dalexsoto:


Aguilex:
XA 9.4.1.1 has already shipped in VS 16.2.4, which went out on September 5. So let's get this tee'd up so we can ship the same fix in the next VSMac 8.2 service release.

PR #8663, opened 2019-09-09
Map `ShowCompletionWindow` to `CommitUniqueCompletionListItem`

DavidKarlas:
Visual Studio Editor has 2 commands to show code completion `CommitUniqueCompletionListItemCommandArgs` and `InvokeCompletionListCommandArgs` there are 2 differences between two..
1) If there is unique item(only 1) in list just complete it without showing code completion
2) At least for web-tools difference is also, filter or not list of items

Since Cmd/Ctrl+Space on Windows but also in old editor behaved as `CommitUniqueCompletionListItemCommandArgs` this mapping is much more correct.


DavidKarlas:
I personally see no point... I have a feeling... `InvokeCompletionListCommandArgs` was kept around for backward compatibility, `CommitUniqueCompletionListItemCommandArgs` is what we always want...

abock:

PR #8664, opened 2019-09-09
[IDE] Check that events passed to the keypress handler are key events

iainx:
Keyboard remapper applications can cause invalid events to be
sent to the key event handlers. Check that events passed to the keypress
handler are actually key events

Fixes VSTS #935180

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -310,6 +310,13 @@ namespace MonoDevelop.Components.Commands
 #if MAC
 		AppKit.NSEvent OnNSEventKeyPress (AppKit.NSEvent ev)
 		{
+			// Protect against non-keyevents being passed here. See VSTS #935180. It seems that certain
+			// keyboard remapper applications such as Ukuele can cause non-keyevents to be sent to key event handlers.
+			if (ev.Type != AppKit.NSEventType.KeyDown && ev.Type != AppKit.NSEventType.KeyUp) {
+				LoggingService.LogInternalError (new Exception ($"Event is type {ev.Type} and not a KeyEvent"));
+				return null;
+			}
+
 			// If we have a native window that can handle this command, let it process
 			// the keys itself and do not go through the command manager.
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -310,6 +310,13 @@ bool CanUseBinding (KeyboardShortcut[] chords, KeyboardShortcut[] accels, out Ke
 #if MAC
 		AppKit.NSEvent OnNSEventKeyPress (AppKit.NSEvent ev)
 		{
+			// Protect against non-keyevents being passed here. See VSTS #935180. It seems that certain
+			// keyboard remapper applications such as Ukuele can cause non-keyevents to be sent to key event handlers.
+			if (ev.Type != AppKit.NSEventType.KeyDown && ev.Type != AppKit.NSEventType.KeyUp) {
```
Therzok: It's NSFlagsChanged

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -310,6 +310,18 @@ bool CanUseBinding (KeyboardShortcut[] chords, KeyboardShortcut[] accels, out Ke
...
+				LoggingService.LogInternalError (new Exception ($"Event is type {ev.Type} and not a KeyEvent"));
```
abock: I'm not sure we actually want to log this? We know now for a fact that because [GDK quartz is crazy](https://github.com/bratsche/gtk/blob/ff3424b5ef5177f32fa6933cee439dfa3d3d89b8/gdk/quartz/gdkevents-quartz.c#L1853-L1915) we will get an NSEvent masquerading as a key up or down that is really a flags changed event. We recently fixed a similar crash in the editor because of GDK's craziness here.

Therzok: This is pure native window handling code.

Therzok: This actual issue surfaced because of keyboard remappers.

slluis:

PR #8666, opened 2019-09-09
Check other projects for existing ports in use

nosami:


main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -115,6 +115,7 @@ private ExecutionCommand CreateAspNetCoreExecutionCommand (ConfigurationSelector
 			else
 				outputFileName = GetOutputFileName (configuration);
 
+			launchProfileProvider.FixPortNumbers ();
 			var applicationUrl = aspnetCoreRunConfiguration.CurrentProfile.TryGetApplicationUrl ();
```
mrward: Are we expecting the ports to change for old projects? I was assuming we would only change the ports for a new project when creating by the New Project dialog. I think we could have this fix ports logic in the new project template wizard.

If I have an existing solution with multiple ASP.NET Core projects and I am using port 5001 for each project this looks like it will change that value in some of the projects.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/LaunchProfileProvider.cs
```diff
@@ -92,6 +133,55 @@ public void LoadLaunchSettings ()
...
+			int port = validPortRange.Except (portsInUse).First (TryAllocatePort);
+			return port;
+		}
+
+		bool ShouldGenerateNewPort (string applicationUrl)
+		{
+			if (!allocatedPorts && project.ParentSolution != null && ApplicationUrlIsDefault (applicationUrl)) {
+				var allProjects = project.ParentSolution.GetAllProjects ().ToArray();
+				if (allProjects.Length > 1 && allProjects [0] == project) {
+					return false;
```
mrward: Looking at this I was assuming it was checking to see if the only project returned was itself. So should this be usig `allProjects.Length == 1` instead of `> 1`?

nosami: If there is only 1 project there is no need to do anything

nosami: But yeah, it looks odd :)

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
nosami: > Are we expecting the ports to change for old projects?

Yes, I think we need to. You can create two projects using `dotnet new`, then create a solution and they 2 projects would have conflicting ports.

nosami: Also, it will only modify the ports if they are the default ones. Once they have been changed, they stay changed.

nosami: > If I have an existing solution with multiple ASP.NET Core projects and I am using port 5001 for each project this looks like it will change that value in some of the projects.

If someone did this, they would still hit the error if they try and ran both projects at the same time, so not sure what the right answer is here. This scenario always holds if both projects are created with `dotnet new` and probably isn't what the user wanted.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/LaunchProfileProvider.cs
mrward: `If there is only 1 project there is no need to do anything` <- makes sense.

Not sure I understand the if statement.

`if (allProjects.Length > 1 && allProjects [0] == project) {`

One project => allProjects.Length == 1. Here we return true so we check the ports.

Two projects. allProjects.Length == 2. First project is current project. Then we return false and do not check the ports.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
mrward: Probably a question for @sayedihashimi 

I can see arguments both ways for either changing or not changing ports for an existing solution 😄

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/LaunchProfileProvider.cs
nosami: Another way to say it is `If we have more than one project and we are currently on the first, then don't do anything` .. If we are on the 2nd (or higher) project and that has default ports, then we need to change them

mrward: Can we add that as a comment to explain it for people like me? 😄

If there is only one project in the list we are checking the ports? Why do we need to do that?

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/LaunchProfileProvider.cs
```diff
@@ -92,6 +132,64 @@ public void LoadLaunchSettings ()
...
+			if (project.ParentSolution != null)
+				projects = project.ParentSolution.GetAllProjects ();
+
+			var runConfigurations = projects.SelectMany (p => p.RunConfigurations).OfType<AspNetCoreRunConfiguration> ();
+			var applicationUrls =
+				runConfigurations.Select (r => r.CurrentProfile.TryGetApplicationUrl ())
+				.Where (a => a != null)
+				.SelectMany (u => u.Split (';'));
+
+			var portsInUse = applicationUrls.Select (url => new Uri (url).Port).ToArray ();
```
Therzok: This is going to allocate a lot (unbounded collection) since we ToArray on a SelectMany. Is there a reason to do ToArray? Can we keep lazy evaluation till the end?

nosami: Yeah.... I missed that. `ToArray()` was debugging code

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
sayedihashimi: I think it's better to be conservative, we may cause unexpected results if we change existing projects. For ex; you have a web, api and mobile project in the same solution. The mobile site calls into the API, but the web is a standalone site that doesn't run at the same time as the others. In that case the user has already configured the mobile app to talk to the api through the localhost API. If we change the url to the api project then the mobile app will start to fail. We don't want to make changes to cause things to fail, especially if it's not obvious to the user why it's failing.

 If the user creates the project from the CLI then it's assumed that they know what they are doing.

nosami: OK. Done!

PR #8667, opened 2019-09-09
[release-8.2] Workspace: Prevent DocumentIds from changing on configuration change

sandyarmstrong:
It's possible for events like device (dis)connect to make several quick
calls to `ActiveConfigurationChanged`, some of which would be canceled,
ultimately causing old `DocumentId` data to get lost during project
reload.

Creating new `DocumentId`s could break multiple Roslyn services such as
intellisense. This change makes the whole process more robust so that
`DocumentId` data does not get lost.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/915494

Backport of #8647.

/cc @DavidKarlas  @Therzok


sandyarmstrong:
We're not merging this until there has been more testing of this change in the wild.

PR #8668, opened 2019-09-09
Fix 934342: When I press shift+tab, the selected code lines don't get un-indented

DavidKarlas:
Problem was that TextBuffer still had default `TabSize` of `4` so when indentions were calculated it it was wrong in some cases.


abock:

PR #8670, opened 2019-09-09
[Debugger] Added more logging and fixed NRE when loading scopes when …

jstedfast:
…local pdb's not available

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/LocalsPad.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/LocalsPad.cs
@@ -27,6 +27,8 @@
 
 using System.Linq;
 
+using Mono.Debugging.Client;
+
 namespace MonoDevelop.Debugger
 {
 	public class LocalsPad : ObjectValuePad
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/LocalsPad.cs
```diff
@@ -48,7 +50,13 @@ void ReloadValues ()
 			if (frame == null)
 				return;
 
-			var locals = frame.GetAllLocals ().Where (l => !string.IsNullOrWhiteSpace (l.Name) && l.Name != "?").ToArray();
+			var locals = frame.GetAllLocals ();
+
+			DebuggerLoggingService.LogMessage ("Begin Local Variables:");
+			foreach (var local in locals)
+				DebuggerLoggingService.LogMessage ("\t{0}", local.Name);
+			DebuggerLoggingService.LogMessage ("End Local Variables");
```
mrward: Do we want to filter out the locals after logging them? The original code was removing empty locals or if their name was ?

mrward: Where are these log messages displayed? Just wondering if we would want them to be translated (assuming they are not temporary).

jstedfast: They just end up in Ide.log and we don't need to filter the string.Empty or "?" locals, that seems to have been a relic from the distant past.

PR #8676, opened 2019-09-10
Increase timeout of TestTasksExecuteSequentially

slluis:
The tests is timing out sometimes. It might happen in slow machines,
or there might be a legitimate bug that is causing tasks to never be
dispatched. Let's find out.


slluis:

slluis:
ouch

slluis:

PR #8679, opened 2019-09-10
Prefer new editor by default for all file types

sandyarmstrong:
## LegacyEditorSupport

In order to be automatically opened in the legacy editor, file types
must explicitly register via the new `LegacyEditorSupport` extension point.

~For file types like C#, where both editors are supported, simply provide
the list of extensions to be supported, and the selected editor will be
determined by the "Use the legacy text editor where available" pref.~

For file types like F#, where only the legacy editor is supported, ~just
add `legacyOnly="true"`~ simply provide the list of extensions to be supported.
Then the legacy editor will be used ~regardless of preference~.

For file types like XML, where support must be determined dynamically
(because users can manually register file types), specify a
`providerType` that implements `ILegacyEditorSupportProvider`. This
provider will be queried to determine where the legacy editor is
preferred.

For file types like Android XML, continue using the old
`SupportedFileType` extension point to signify that the modern editor
is always preferred over the legacy editor. This will override any
"base" LegacyEditorSupport (again, like XML).

~Users will continue to be able to use the "Open With" menu to pick the
editor they prefer at that time. Of course when they pick unsupported
pairings, they may not have the best experience (missing features, etc).~

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/567102

## Terminology and Preferences

* ~Refer to new editor as "modern editor", and~ fix several places where
  it was still called the "experimental new editor".
* ~Refer to old editor as "legacy editor".~
* Users will only ever see "Source Code Editor" in the UI (via the context menu).
  That menu item will open the correct editor for that file.
* Remove link to "learn more about the New Editor", as the aka has not
  been maintained and there is no useful documentation page right now.
* ~Change the "Open C# files in the New Editor" preference to read
  "Use the legacy text editor where available (not recommended)", and
  invert its effect. Move this preference to the end of its section.~ Remove the
  "Open C# files in the New Editor" preference.


main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/LegacyEditorSupportExtensionNode.cs
```diff
@@ -0,0 +1,39 @@
...
+		[NodeAttribute ("legacyOnly", "Set this to true if modern editor is unsupported for files matching this node")]
+		public bool IsLegacyOnly { get; private set; }
+
+		[NodeAttribute("providerType", "Set this to an the type name for a MonoDevelop.TextEditor.ILegacyEditorSupportProvider implementation")]
+		public string ProviderType { get; private set; }
+	}
+
+	public interface ILegacyEditorSupportProvider
+	{
+		public bool PreferLegacyEditor (Ide.Gui.Documents.FileDescriptor modelDescriptor);
```
Therzok:
```suggestion
		bool PreferLegacyEditor (Ide.Gui.Documents.FileDescriptor modelDescriptor);
```

sandyarmstrong: Thanks, will fix. @abock suggested several other changes to me in a call so this is not the final form, just yet.

sandyarmstrong:
image

Therzok:
So what happens to F#?

sandyarmstrong:
@Therzok the legacy editor is used for F#. See its the F# addin.xml.

Therzok:
@manish @slluis @jtorres leak tests will fail consistently if the old editor support is removed for C# files. We need autotest support for the new editor, the leaks fixed and changing to the new leak infrastructure if we want to merge this

sandyarmstrong:
@Therzok I'll just add a feature flag that causes text files to always open with legacy editor, and autotest can use it until they switch to new editor. Then there should be no reason to block this PR.

sandyarmstrong:
@Therzok @manish @jtorres @abock you can now use `MD_FEATURES_ENABLED=AlwaysUseLegacyEditor` (or whatever your preferred method of enabling the `AlwaysUseLegacyEditor` feature switch is) to force the legacy editor to always be used for text files.

Therzok:
I'll push a commit to this PR for current leaktests and make a different PR for the new infrastructure.

main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewDisplayBinding.cs
```diff
@@ -21,53 +21,99 @@
...
+			var isTextByFilePath = modelDescriptor.FilePath.IsNotNull && IdeServices.DesktopService.GetFileIsText (modelDescriptor.FilePath, modelDescriptor.MimeType);
+			var isTextByMimeType = !string.IsNullOrEmpty (modelDescriptor.MimeType) && IdeServices.DesktopService.GetMimeTypeIsText (modelDescriptor.MimeType);
+
+			if (!isTextByFilePath && !isTextByMimeType)
+				yield break;
+
+			// If feature switch is enabled, don't even bother checking for modern editor support.
+			// This should only be used by automatic testing tooling that hasn't updated to support modern editor yet.
+			var preferLegacy = alwaysUseLegacyEditor.Value;
+			if (!preferLegacy) {
```
Therzok: Should this be `preferLegacy`?

sandyarmstrong: No. The feature switch makes it so there's no point in checking any of the extension nodes. Just always use legacy editor. The normal flow is for `preferLegacy` to be `false` here, though.

Therzok: Well, it does look like we're still not getting the old editor in leak tests.

PR #8687, opened 2019-09-11
[Ide] Add some .Ignore calls to WatchDirectories calls to look for ex…

Therzok:
…ceptions

This should help investigate issues like #980470 - Deadlock in FileWatcherService.WatchDirectories

From the looks of the code, there is no way for a deadlock to happen, as the async code is not
re-entrant.


Therzok:
Build finished, VSTS is acting up.

PR #8688, opened 2019-09-11
Fix hang in file watcher

slluis:
Enabling events has a sync wait on a flush of the IO event queue, and
that may take many seconds to complete if there is a lot of IO. To
avoid locking on this call, we now enable the events on a background
thread.

Fixes VSTS #952405 - Visual Studio Mac 2019 Freezes on Switch Branch
Fixes VSTS #980470 - Deadlock in FileWatcherService.WatchDirectories

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
@@ -209,8 +209,9 @@ namespace MonoDevelop.Projects
 				// Add the new ones.
 				foreach (var kvp in newWatchers) {
 					var path = kvp.Key;
-					// Don't modify a watcher that already exists.
-					if (watchers.ContainsKey (path)) {
+					// Don't modify a watcher that already exists, but ensure it is initialized
+					if (watchers.TryGetValue (path, out var existingWatcher)) {
+						tasks.Add (existingWatcher.EnableRaisingEventsAsync ());
 						continue;
 					}
 					var watcher = new FileWatcherWrapper (path, kvp.Value, readerWriterLock);
@@ -393,7 +394,7 @@ namespace MonoDevelop.Projects
 		public Task EnableRaisingEventsAsync ()
 		{
 			lock (enableEventLock) {
-				if (disposed || eventsEnabled)
+				if (disposed || (eventsEnabled && enablingEventsTask == null))
 					return Task.CompletedTask;
 				eventsEnabled = true;
 				if (enablingEventsTask == null) {
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
@@ -168,24 +168,27 @@ namespace MonoDevelop.Projects
 		static Dictionary<FilePath, PathTreeNode> newWatchers = new Dictionary<FilePath, PathTreeNode>();
 		static List<FilePath> toRemove = new List<FilePath> ();
 
-		static void UpdateWatchers (CancellationToken token)
+		static Task UpdateWatchers (CancellationToken token)
 		{
 			if (token.IsCancellationRequested)
-				return;
+				return Task.CompletedTask;
+
+			var tasks = new List<Task> ();
+
 			using (readerWriterLock.Write ()) {
 				if (token.IsCancellationRequested)
-					return;
+					return Task.CompletedTask;
 				newWatchers.Clear ();
 				foreach (var node in tree.Normalize (maxWatchers)) {
 					if (token.IsCancellationRequested)
-						return;
+						return Task.CompletedTask;
 					var dir = node.GetPath ().ToString ();
 					if (Directory.Exists (dir))
 						newWatchers.Add (dir, node);
 				}
 				if (newWatchers.Count == 0 && watchers.Count == 0) {
 					// Unchanged.
-					return;
+					return Task.CompletedTask;
 				}
 				toRemove.Clear ();
 				foreach (var kvp in watchers) {
@@ -196,7 +199,7 @@ namespace MonoDevelop.Projects
 
 				// After this point, the watcher update is real and a destructive operation, so do not use the token.
 				if (token.IsCancellationRequested)
-					return;
+					return Task.CompletedTask;
 
 				// First remove the watchers, so we don't spin too many threads.
 				foreach (var directory in toRemove) {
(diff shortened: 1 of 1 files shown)


slluis:

slluis:

Therzok (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
```diff
@@ -363,18 +367,71 @@ public FileWatcherWrapper (FilePath path, PathTreeNode rootNode, ReaderWriterLoc
...
+					if (disposed || value == eventsEnabled)
+						return;
+					eventsEnabled = value;
+					if (eventsEnabled && !enablingEvents) {
+						enablingEvents = true;
+
+						// Enabling events has a sync wait on a flush of the IO event queue, and that may take many
+						// seconds if there is a lot of IO. To avoid locking on this call, we enable the events
+						// on a background thread.
+						Task.Run (EnableEvents).Ignore ();
```
Therzok: Maybe we return this task so the service can return the task that actually waits for the watcher to be enabled.

Therzok: UpdateWatchers would then return a Task and UpdateWatchersAsync would return the Task.WhenAll for all of these.

Therzok: Otherwise we might see transient failures in VSTS because the directories aren't watched because of high IO.

mrward: Yeah. I wonder if this will affect some tests. Some tests assume the watcher is created when file watcher service methods return which might not happen now. Have to see.

https://github.com/mono/monodevelop/blob/dea5016c423bcc2f827a557d33c32cc2e5e75e22/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs#L309

mrward (approved):

Therzok (changes requested):
Marking as requesting changes. We need to handle the watcher being enabled by handling the task.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
```diff
@@ -163,29 +163,32 @@ static Task UpdateWatchersAsync ()
 			cancellationTokenSource = new CancellationTokenSource ();
 			CancellationToken token = cancellationTokenSource.Token;
 
-			return Task.Run (() => UpdateWatchers (token));
+			return Task.Run (async () => await UpdateWatchers (token));
```
Therzok:
```suggestion
			return Task.Run (() => UpdateWatchers (token)).Unwrap ();
```

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
```diff
@@ -212,24 +215,27 @@ static void UpdateWatchers (CancellationToken token)
...
-					}
+					var task = watcher.EnableRaisingEventsAsync ().ContinueWith (t => {
+						if (t.IsFaulted && t.Exception.InnerException is UnauthorizedAccessException e) {
+							LoggingService.LogWarning ("Access to " + path + " denied. Stopping file watcher.", e);
+							using (readerWriterLock.Write ()) {
+								watcher.Dispose ();
+								watchers.Remove (path);
+							}
+						}
+					}, TaskScheduler.Default);
```
Therzok:
```suggestion
					}, CancellationToken.None, TaskContinuationOptions.OnlyOnFaulted, TaskScheduler.Default);
```

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
```diff
@@ -212,24 +215,27 @@ static void UpdateWatchers (CancellationToken token)
...
 					watchers.Add (path, watcher);
-					try {
-						watcher.EnableRaisingEvents = true;
- 					} catch (UnauthorizedAccessException e) {
-						LoggingService.LogWarning ("Access to " + path + " denied. Stopping file watcher.", e);
-						watcher.Dispose ();
-						watchers.Remove (path);
-					}
+					var task = watcher.EnableRaisingEventsAsync ().ContinueWith (t => {
+						if (t.IsFaulted && t.Exception.InnerException is UnauthorizedAccessException e) {
```
Therzok: Not needed anymore with TaskContinuationOptions.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
```diff
@@ -363,18 +376,88 @@ public FileWatcherWrapper (FilePath path, PathTreeNode rootNode, ReaderWriterLoc
 		public FilePath Path { get; }
 
 		public bool EnableRaisingEvents {
-			get { return watcher.EnableRaisingEvents; }
-			set { watcher.EnableRaisingEvents = value; }
+			get { return eventsEnabled; }
+		}
+
+		public void DisableRisingEvents ()
```
Therzok:
```suggestion
		public void DisableRaisingEvents ()
```

and other places.

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
```diff
@@ -363,18 +376,88 @@ public FileWatcherWrapper (FilePath path, PathTreeNode rootNode, ReaderWriterLoc
...
+					// Enabling events has a sync wait on a flush of the IO event queue, and that may take many
+					// seconds if there is a lot of IO. To avoid locking on this call, we enable the events
+					// on a background thread.
+					return enablingEventsTask = semaphore.WaitAsync ().ContinueWith (EnableEvents, CancellationToken.None, TaskContinuationOptions.RunContinuationsAsynchronously, scheduler: TaskScheduler.Default);
+				}
+			}
+			return Task.CompletedTask;
 		}
 
+		void EnableEvents (Task t)
```
Therzok: t isn't used

Therzok (approved):
Looks good!

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
slluis: it is necessary to avoid a lambda in ContinueWith

Therzok: I'll be pedantic a bit, but roslyn codegen sucks.

This reuses a cached `Action<Task>`
```csharp
ContinueWith (t => EnableEvents (t))
```

This always allocates an `Action<Task>`.
```csharp
ContinueWith (EnableEvents);
```

slluis:

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
Therzok: Ah, nevermind! This is not a lambda without captures.

slluis:

slluis:

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
```diff
@@ -209,8 +209,9 @@ static Task UpdateWatchers (CancellationToken token)
 				// Add the new ones.
 				foreach (var kvp in newWatchers) {
 					var path = kvp.Key;
-					// Don't modify a watcher that already exists.
-					if (watchers.ContainsKey (path)) {
+					// Don't modify a watcher that already exists, but ensure it is initialized
+					if (watchers.TryGetValue (path, out var existingWatcher)) {
+						tasks.Add (existingWatcher.EnableRaisingEventsAsync ());
```
Therzok: :+1:

PR #8691, opened 2019-09-11
[DotNetCore] Do not show update count label on Frameworks folder

mrward:
The Frameworks folder for .NET Core 3.0 and .NET Standard 2.1
projects would show an update count label if the project had package
updates. Framework references cannot be updated in the IDE, and also
are not typically defined in the project itself, so the update count
should not be shown.

Fixes VSTS #980887 - Frameworks node under dependencies has a message
about updates next to it

Frameworks


mrward:

PR #8693, opened 2019-09-11
[Core] Fix adding derived Control items to the weak map

Therzok:
DropDownListBox is actually a Window, so use that as the type when adding to the table, not plain Control.
This prevents duplicates from being added to the map, and ensures the right type is added for this widget.

Fixes VSTS #976424 - Unhandled argument exception adding to ConditionalWeakTable in Components.Control

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp/CSharpPathedDocumentExtension.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp/CSharpPathedDocumentExtension.cs
@@ -154,7 +154,7 @@ namespace MonoDevelop.CSharp
 				MaxVisibleRows = 14
 			};
 			window.SelectItem (path [index].Tag);
-			return window;
+			return (Window)window;
 		}
 
 		PathEntry [] currentPath;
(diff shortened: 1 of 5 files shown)


Therzok:

Therzok:
Honestly, this is a mishap because `CreatePathWidget` requests a `Control` while `GetActiveWindow` requests `Window`. We could hard-code `GetImplicit` in `Control` to know about `Control`, `Window` and `Dialog`, but that could increase complexity unnecessarily.

Usually, if you want a `Window`, you will use the `Window` API. I am okay with adding code in `Control.GetImplicit`, just feels 'hacky'.

Therzok:
I don't know how we could prevent it in Control.cs. The C# compiler will choose this overload:
```csharp
		public static implicit operator Gtk.Widget (Control d)
		{
			return d?.GetNativeWidget<Gtk.Widget> ();
		}
```

Over this overload:
```csharp
		public static implicit operator Gtk.Window (Window d)
		{
			if (d is XwtWindowControl)
				return (XwtWindowControl)d;
			return d?.GetNativeWidget<Gtk.Window> ();
		}
```

So that is why this happens.

Therzok:
Added Debug.Assert as per @mrward 's suggestion

PR #8695, opened 2019-09-11
[Core] Enable nullability analysis in MonoDevelop.Utilities

Therzok:
Fixes VSTS #984188


mrward (commented):
Does this change the public API when we switch to nullable types?

mrward (approved):

jtorres (approved):

PR #8704, opened 2019-09-12
Enable FileNesting support on Razor Class Library projects

iantoalms:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/977852

The current logic which determines whether FileNesting is enabled checks for the existence of the UsingMicrosoftNETSdkWeb project property.
This isn't set in a Razor Class Library project, so we also look for the RazorTargetName property which is always set in this project type (see https://docs.microsoft.com/en-us/aspnet/core/razor-pages/sdk?view=aspnetcore-2.2)

### DIFF
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.Tests/MonoDevelop.AspNetCore.Tests/AspNetCoreProjectTests.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.Tests/MonoDevelop.AspNetCore.Tests/AspNetCoreProjectTests.cs
@@ -115,6 +115,15 @@ namespace MonoDevelop.AspNetCore.Tests
 		}
 
 		[Test]
+		public async Task RazorClassLib_Supports_FileNesting ()
+		{
+			string projectFileName = Util.GetSampleProject ("aspnetcore-razor-class-lib", "aspnetcore-razor-class-lib.csproj");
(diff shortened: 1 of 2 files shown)


main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.Tests/MonoDevelop.AspNetCore.Tests/AspNetCoreProjectTests.cs
```diff
@@ -114,6 +114,15 @@ public async Task AspNetCore_FileNesting ()
 			}
 		}
 
+		[Test]
+		public async Task RazorClassLib_Supports_FileNesting ()
+		{
+			string projectFileName = Util.GetSampleProject ("aspnetcore-razor-class-lib", "aspnetcore-razor-class-lib.csproj");
+			using (var project = (DotNetProject)await Services.ProjectService.ReadSolutionItem (Util.GetMonitor (), projectFileName)) {
+				Assert.True (FileNestingService.IsEnabledForProject (project));
+			}
```
rodrmoya: This should be testing `AppliesToProject` rather than `IsEnabledForProject`, which just checks a property in the solution and, if not present, returns true by default.

rodrmoya: Fixed in https://github.com/mono/monodevelop/pull/8729

PR #8707, opened 2019-09-12
[A11y] Set the background color of stacktrace items in Exception Caug…

ketttY:
…ht Dialog

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/648206/


mrward:
Code looks good. Just wondering if we can fix the contrast when using the dark theme and a row in the Properties list view is selected but it does not have focus.

Screenshot 2019-09-13 at 10 21 31

The contrast here is a bit low and does not equal or pass the 4.5:1 ratio. We could use a slightly darker grey for the selected background colour, or maybe change the foreground text to be black text.

Light theme looks good.

ketttY:
Screen Shot 2019-09-16 at 2 01 58 PM
Screen Shot 2019-09-16 at 2 01 18 PM

New color for properties rows

PR #8725, opened 2019-09-13
Enable Browser Selection on Publish Part 1

iantoalms:
Make AspNetCoreExecutionTarget public so that we can determine the active execution target and hence preferred browser from external code


iantoalms:

PR #8726, opened 2019-09-13
Fix all hangs

Therzok:
See each individual commit:

Fixes VSTS #980697 - Moving more than 2 nested files hangs VSfM
Fixes VSTS #982334 - IDE hang in version control when opening Storyboard in Xcode Interface Builder

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
@@ -167,6 +167,9 @@ namespace MonoDevelop.Projects
 						}
 					}
 
+					if (isNop)
+						return;
+
 					using (readerWriterLock.Write ()) {
 						Watch_NoLock (toWatch).Ignore ();
 						return;
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
@@ -60,8 +60,11 @@ namespace MonoDevelop.Projects
 		{
 			Debug.Assert (readerWriterLock.IsWriteLockHeld);
 
-			var toWatch = ComputeItems (item, registerEvent);
+			return Watch_NoLock (ComputeItems (item, registerEvent));
+		}
 
+		static Task Watch_NoLock (List<(object, HashSet<FilePath>)> toWatch)
+		{
 			bool modified = false;
 			foreach (var (id, set) in toWatch) {
 				modified |= RegisterDirectoriesInTree_NoLock (id, set);
@@ -146,8 +149,31 @@ namespace MonoDevelop.Projects
 
 		static void OnRootDirectoriesChanged (object sender, WorkspaceItem.RootDirectoriesChangedEventArgs args)
 		{
-			using (readerWriterLock.Write ()) {
-				if (args.SourceItem is WorkspaceObject item) {
+			if (args.SourceItem is WorkspaceObject item) {
+				if (!args.IsRemove && !args.IsAdd) {
+					// Don't recompute the items if the watched paths hasn't changed.
+					var toWatch = ComputeItems (item, false);
+
+					bool isNop = true;
+					using (readerWriterLock.Read ()) {
+						foreach (var (id, set) in toWatch) {
+							if (!monitoredDirectories.TryGetValue (id, out var currentSet))
+								continue;
+
+							if (!currentSet.SetEquals (set)) {
+								isNop = false;
+								break;
+							}
+						}
+					}
+
+					using (readerWriterLock.Write ()) {
+						Watch_NoLock (toWatch).Ignore ();
+						return;
+					}
+				}
+
+				using (readerWriterLock.Write ()) {
 					if (args.IsRemove) {
 						Remove_NoLock (item).Ignore ();
 					} else {
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
@@ -76,9 +76,12 @@ namespace MonoDevelop.Ide.Projects.FileNesting
 			return null;
 		}
 
-		public static event Action<Project> NestingRulesChanged;
+		/// <summary>
+		/// The project file which had modifications, in most cases, the parent of the added/removed projectfile
+		/// </summary>
+		public static event Action<ProjectFile, ProjectFile> NestingRulesChanged;
 
-		internal static void NotifyNestingRulesChanged (Project project) => NestingRulesChanged?.Invoke (project);
+		internal static void NotifyNestingRulesChanged (ProjectFile toUpdate, ProjectFile newParent) => NestingRulesChanged?.Invoke (toUpdate, newParent);
 
 		public static bool IsEnabledForProject (Project project)
 		{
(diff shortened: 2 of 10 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects/FileWatcherService.cs
```diff
@@ -145,16 +148,39 @@ static Task Remove_NoLock (WorkspaceObject item)
...
-			using (readerWriterLock.Write ()) {
-				if (args.SourceItem is WorkspaceObject item) {
+		{
+			if (args.SourceItem is WorkspaceObject item) {
+				if (!args.IsRemove && !args.IsAdd) {
+					// Don't recompute the items if the watched paths hasn't changed.
+					var toWatch = ComputeItems (item, false);
+
+					bool isNop = true;
+					using (readerWriterLock.Read ()) {
```
mrward: Should we be using isNop somewhere? We seem to set it but not use it.

Therzok: YES

mrward (approved):

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
```diff
@@ -76,9 +76,12 @@ internal static ProjectFile InternalGetParentFile (ProjectFile inputFile)
...
 		}
 
-		public static event Action<Project> NestingRulesChanged;
+		/// <summary>
+		/// The project file which had modifications, in most cases, the parent of the added/removed projectfile
+		/// </summary>
+		public static event Action<ProjectFile, ProjectFile> NestingRulesChanged;
 
-		internal static void NotifyNestingRulesChanged (Project project) => NestingRulesChanged?.Invoke (project);
+		internal static void NotifyNestingRulesChanged (ProjectFile toUpdate, ProjectFile newParent) => NestingRulesChanged?.Invoke (toUpdate, newParent);
```
rodrmoya: not sure this would work, as in some cases we are adding a file that becomes the parent of an existing file, that's why it just tries to recalculate the whole tree

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
```diff
@@ -192,45 +196,56 @@ void RemoveFile (ProjectFile projectFile, List<ProjectFile> originalFilesToRemov
...
+
+			return nestingInfo;
 		}
 
 		void OnFileAddedToProject (object sender, ProjectFileEventArgs e)
 		{
 			foreach (var file in e) {
-				AddFile (file.ProjectFile);
+				var nestingInfo = AddFile (file.ProjectFile);
+				NotifyNestingRulesChanged (nestingInfo);
```
rodrmoya: shouldn't this be called if `fileNestingEnabled` is true only, as before?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
```diff
@@ -192,45 +196,56 @@ void RemoveFile (ProjectFile projectFile, List<ProjectFile> originalFilesToRemov
...
 		void OnFileRenamedInProject (object sender, ProjectFileRenamedEventArgs e)
 		{
 			foreach (var file in e) {
-				RemoveFile (file.ProjectFile);
-				AddFile (file.ProjectFile);
+				var oldNestingInfo = RemoveFile (file.ProjectFile);
+				NotifyNestingRulesChanged (oldNestingInfo);
+
+				var nestingInfo = AddFile (file.ProjectFile);
+				NotifyNestingRulesChanged (nestingInfo);
```
rodrmoya: same here

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
Therzok: Which is why we compute this one's parent. See the treebuilder change

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
Therzok: Will change

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
Therzok: Why don't we have tests for this?

rodrmoya: We do -> https://github.com/mono/monodevelop/blob/master/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.Tests/MonoDevelop.AspNetCore.Tests/AspNetCoreProjectTests.cs#L67 although for sure they can be extended to cover more cases

Therzok: If the tests pass then... What's going on?

rodrmoya: Ok, just tested this build for this:

1. Create a ASP.NET Core project
2. Add a "Whatever.cshtml.cs" file
3. Add a "Whatever.cshtml" file

After the last step, .cshtml.cs file should become the child of the .cshtml file, but this doesn't work neither on the 8.3 build without the changes on this PR. I thought I had fixed that, but since it happens also on 8.3, I'll fix it in a separate PR

slluis:
The version control changes look ok to me. I'll let @rodrmoya review the file nesting changes.

slluis:

rodrmoya:
File nesting changes look good to me, except for avoiding the call to `NotifyRulesChanged` if `fileNestingEnabled` is false, as that would avoid unneeded calls when file nesting is disabled for that project

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
```diff
@@ -229,7 +229,7 @@ void OnFileRenamedInProject (object sender, ProjectFileRenamedEventArgs e)
 
 		void NotifyNestingRulesChanged (ProjectFileNestingInfo nestingInfo)
 		{
-			if (nestingInfo == null)
+			if (nestingInfo == null || !FileNestingService.AppliesToProject (nestingInfo.File.Project))
```
rodrmoya: with the changes in my PR, this extra condition is useless, as this would only be called for projects where file nesting applies.

rodrmoya: But yes, makes sense on this PR, I'll remove on my PR once I rebase on this

rodrmoya:
> File nesting changes look good to me, except for avoiding the call to `NotifyRulesChanged` if `fileNestingEnabled` is false, as that would avoid unneeded calls when file nesting is disabled for that project

which can be done on my PR (https://github.com/mono/monodevelop/pull/8730) so this is ok to merge from me

rodrmoya (approved):

slluis:

PR #8729, opened 2019-09-15
[AspNetCore] Fix Razor Class Lib file nesting test

rodrmoya:
It should be testing if file nesting applies to a Razor Class library
project, not if it is enabled, which is always true unless explicitly
disabled.


rodrmoya:

PR #8730, opened 2019-09-15
[FileNesting] Avoid computing file nesting tree for all projects

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/982815

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
@@ -95,17 +95,26 @@ namespace MonoDevelop.Ide.Projects.FileNesting
 
 		public static ProjectFile GetParentFile (ProjectFile inputFile)
 		{
+			if (!AppliesToProject (inputFile.Project)) {
+				return null;
+			}
 			return GetProjectNestingInfo (inputFile.Project).GetParentForFile (inputFile);
 		}
 
 		public static bool HasChildren (ProjectFile inputFile)
 		{
+			if (!AppliesToProject (inputFile.Project)) {
+				return false;
+			}
 			var children = GetProjectNestingInfo (inputFile.Project).GetChildrenForFile (inputFile);
 			return (children?.Count ?? 0) > 0;
 		}
 
 		public static ProjectFileCollection GetChildren (ProjectFile inputFile)
 		{
+			if (!AppliesToProject (inputFile.Project)) {
+				return null;
+			}
 			return GetProjectNestingInfo (inputFile.Project).GetChildrenForFile (inputFile);
 		}
 	}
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
```diff
@@ -61,7 +61,16 @@ static void HandleRulesProviderExtension (object sender, ExtensionNodeEventArgs
 
 		static ProjectNestingInfo GetProjectNestingInfo (Project project)
 		{
-			return loadedProjects.GetValue (project, p => new ProjectNestingInfo (p));
+			if (!loadedProjects.TryGetValue (project, out var pinfo)) {
```
Therzok: In case it doesn't apply to a project, we unnecessarily compute a Project's hashcode in the concurrent dictionary which will acquire locks.

Can we just query AppliesToProject before?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
```diff
@@ -61,7 +61,16 @@ static void HandleRulesProviderExtension (object sender, ExtensionNodeEventArgs
...
+				if (!AppliesToProject (project)) {
+					return null;
+				}
+
+				pinfo = new ProjectNestingInfo (project);
+				loadedProjects.Add (project, pinfo);
```
Therzok: This can race and throw dupe key exception

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
rodrmoya: This was the 1st thing I did, but thought it would be better to do it on `GetProjectNestingInfo`, but changed back to that now

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
rodrmoya: Fixed

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
```diff
@@ -92,17 +92,26 @@ public static bool AppliesToProject (Project project)
 
 		public static ProjectFile GetParentFile (ProjectFile inputFile)
 		{
+			if (!AppliesToProject (inputFile.Project)) {
```
Therzok: Can we also subscribe the event handlers based on project applicability?

rodrmoya: hmm, that's done only when we create a `ProjectNestingInfo`, which only happens now if it applies to the project, or do you mean something else?

Therzok: Makes sense!

rodrmoya:
Rebased, and includes now the changes from the code comments in https://github.com/mono/monodevelop/pull/8726

rodrmoya:

PR #8736, opened 2019-09-16
TextViewContent: Only handle commands if text view has keyboard focus

sandyarmstrong:
Allows various keyboard commands to continue working if, for example,
an editor margin contains a view that accepts input.


main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor/TextViewContent.Commands.cs
```diff
@@ -207,8 +207,11 @@ bool CanHandleCommand (object commandId)
 					return commandsSupportedWhenFindPresenterIsFocused.Contains (commandId);
 				}
 			}
-
+#if !WINDOWS
+			return (TextView as ITextView3)?.IsKeyboardFocused ?? true;
```
abock:
```csharp
#if !WINDOWS
if (TextView is ITextView3 textView3)
    return textView3.IsKeyboardFocused;
#endif

return true;
```

sandyarmstrong:
@monojenkins backport master

PR #8743, opened 2019-09-17
[AutoTest] NSObjectResult: Check for null runtimeName is added

OlgaBoiarshynova:
[Bug 983495:](https://dev.azure.com/devdiv/DevDiv/_workitems/edit/983495) ArgumentNullException is triggered in NSObjectResult on Active Runtime selection in Autotest

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest.Results/NSObjectResult.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest.Results/NSObjectResult.cs
@@ -378,9 +378,11 @@ namespace MonoDevelop.Components.AutoTest.Results
 				if (string.IsNullOrEmpty (runtimeName))
 					return false;
 
-				if (mutableModel.FullDisplayString.Contains (runtimeName) || mutableModel.DisplayString.Contains (runtimeName))
+				if (!string.IsNullOrWhiteSpace(mutableModel.FullDisplayString) && mutableModel.FullDisplayString.Contains (runtimeName))
 					return true;
-			
+				if (!string.IsNullOrWhiteSpace(mutableModel.DisplayString) && mutableModel.DisplayString.Contains (runtimeName))
+					return true;
+
 				var execTargetPInfo = r.GetType().GetProperty ("ExecutionTarget");
 				if(execTargetPInfo != null) {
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AutoTest.Results/NSObjectResult.cs
```diff
@@ -374,10 +374,13 @@ public override bool SetActiveRuntime (string runtimeName)
...
 				var mutableModel = r.GetMutableModel ();
 				LoggingService.LogDebug ($"[IRuntimeModel.IRuntimeMutableModel] FullDisplayString: '{mutableModel.FullDisplayString}' | DisplayString: '{mutableModel.DisplayString}'");
-				if (mutableModel.FullDisplayString.Contains (runtimeName))
-					return true;
-				if (mutableModel.DisplayString.Contains (runtimeName))
+
+				if (string.IsNullOrEmpty (runtimeName))
+					return false;
+
+				if (mutableModel.FullDisplayString.Contains (runtimeName) || mutableModel.DisplayString.Contains (runtimeName))
```
manish: the NRE comes from `FullDisplayString` or `DisplayString` being null, which means that NRE will still happen here.

so change

```csharp
if (mutableModel.FullDisplayString.Contains (runtimeName))
    return true;
if (mutableModel.DisplayString.Contains (runtimeName))
    return true;
```

to

```csharp
if (mutableModel.FullDisplayString != null && mutableModel.FullDisplayString.Contains (runtimeName))
    return true;
if (mutableModel.DisplayString !=null && mutableModel.DisplayString.Contains (runtimeName))
    return true;
```

OlgaBoiarshynova: @manish There is no NRE here but ArgumentNullException which is caused by no device is connected to bot and test passed null as parameter of SetActiveRuntime method

I can add null checks to FullDisplayString and DisplayString but currently it doesn't seem required.

Please let me know if I still need to add additional checks.

manish: It would be great to have both the checks in this case.

PR #8744, opened 2019-09-17
Break back compat of version control add-in

slluis:
Version control api has changed. This will prevent the load of version control extensions targeting old versions of the IDE.


slluis:

PR #8761, opened 2019-09-18
Fix attachment view display issue

slluis:
When attachments are added to a view, a new attachments container is
added, which has the main view as first item and the attachments as
additional items. Before inserting the main view as first item of the
attachments container we need to make sure it is detached from the
parent container, otherwise it can't be added to the new container.

Also fixed exception that may happen when a view is realized and then
quickly disposed.

Fixes VSTS #983862 - Editor goes blank if solution pad is used to move
an unsaved file to another folder

Backport of https://github.com/mono/monodevelop/pull/8760


slluis:
@monojenkins backport release-8.3-rtw

PR #8768, opened 2019-09-18
Fixes current navigation history and takes care about closed documents

netonjm:
Our navigation history seems to be broken because we are not refreshing current Line and Column of all TextFileNavigationPoint.cs items (is stays at 0).

This PR also fixes a case when the document is closed but has a TextFileNavigationPoint. We empty the doc variable for a good clean, but in case of go history of a closed document we need to re-set this property to set the cursor correctly in the line

someeyes

Fixes VSTS #984818 - "Navigate Back" shorcut it's not working correctly

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/DocumentNavigationPoint.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/DocumentNavigationPoint.cs
@@ -48,8 +48,11 @@ namespace MonoDevelop.Ide.Navigation
 		FilePath fileName;
 		string project;
 		
+		bool isDocument;
+
 		public DocumentNavigationPoint (Document doc)
 		{
+			isDocument = true;
 			SetDocument (doc);
 		}
 
@@ -113,7 +116,14 @@ namespace MonoDevelop.Ide.Navigation
 					break;
 				}
 			}
-			return await IdeApp.Workbench.OpenDocument (new FileOpenInformation (fileName, p, true));
+
+			//in case the document was reopened we want to set again
+			var document = await IdeApp.Workbench.OpenDocument (new FileOpenInformation (fileName, p, true));
+			if (isDocument) {
+				SetDocument (document);
+			}
+
+			return document;
 		}
 		
 		public override string DisplayName {
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/TextFileNavigationPoint.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/TextFileNavigationPoint.cs
@@ -53,6 +53,16 @@ namespace MonoDevelop.Ide.Navigation
 			: base (doc)
 		{
 			offset = textView.Caret.Position.BufferPosition;
+			RefreshWithCurrentOffset (textView);
+		}
+
+		void RefreshWithCurrentOffset (ITextView textView)
+		{
+			if (textView != null && offset.HasValue) {
+				var currentLine = textView.TextBuffer.CurrentSnapshot.GetLineFromPosition (offset.Value);
+				line = currentLine.LineNumber;
+				column = offset.Value.Position - currentLine.Start.Position;
+			}
 		}
 
 		protected override void OnDocumentClosing ()
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/TextFileNavigationPoint.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/TextFileNavigationPoint.cs
@@ -138,7 +138,7 @@ namespace MonoDevelop.Ide.Navigation
 
 		protected void JumpToCurrentLocation (ITextView textView)
 		{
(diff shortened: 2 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/DocumentNavigationPoint.cs
```diff
@@ -48,13 +48,16 @@ public class DocumentNavigationPoint : NavigationPoint
 		FilePath fileName;
 		string project;
 		
+		bool isDocument;
+
 		public DocumentNavigationPoint (Document doc)
 		{
+			isDocument = true;
```
sandyarmstrong: When is this ever `false`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/TextFileNavigationPoint.cs
```diff
@@ -128,7 +138,7 @@ protected override async Task<Document> DoShow ()
 
 		protected void JumpToCurrentLocation (ITextView textView)
 		{
-			textView.NavigateToLineAndColumn (Math.Max (0, Line - 1), Math.Max (0, Column - 1));
+			textView.NavigateToLineAndColumn (Math.Max (0, Line), Math.Max (0, Column));
```
sandyarmstrong: What happens if the old constructor is called, which takes line/column, and was expecting them to be 1-based?

What about how line/column are modified in `OnDocumentClosing`? Was that wrong before? Is it right now?

sandyarmstrong: If this was all wrong before, that's fine. I just want to be sure! :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/DocumentNavigationPoint.cs
netonjm: here : https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/DocumentNavigationPoint.cs#L62-L65

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/TextFileNavigationPoint.cs
netonjm: > What happens if the old constructor is called, which takes line/column, and (it seems) was expecting them to be 1-based?
> 
> What about how line/column are modified in `OnDocumentClosing`? Was that wrong before? Is it right now?

I don't know in deep this code, but it seems the old constructor doesn't expect a doc (correct me if I'm wrong) and then the subscription HandleClosed will not happen  (and there will be not call to OnDocumentClosing) in this case

> If this was all wrong before, that's fine. I just want to be sure! :)

I'm sure there is a reason for this (maybe @slluis knows better), but with Line - 1 breaks the feature for me. Any ideas?

sevoku:
@monojenkins backport release-8.3-rtw

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/TextFileNavigationPoint.cs
avodovnik: Also, did you test this change with the Errors Pad? Particularly show next/prev errors did not work, hence the `-1` change was made.

mrward: Yeah the show next/prev errors does not seem to work with this branch. VS Mac 8.3 seems to be OK. Is that something that could be fixed in the Errors pad?

avodovnik: That's what I was afraid of. That uses `TextFileNavigationPoint` too, hence the change. We can probably find a fix in there, but it should probably be addressed in, or immediately with, this PR.

avodovnik: I think we could get away with doing translation [here](http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Ide.Tasks/TaskStore.cs,422). That'd keep both logics the same, I think.

netonjm: Totally agree to move this code to ErrorPad, 👌🏻

avodovnik: I made the change. It looks like it keeps the error navigation working as well :-) Can someone double check this + the navigation behaviour?

avodovnik:
Just to make sure we keep track, this was the fix in the `TaskStore` that fixed error navigation: https://github.com/mono/monodevelop/commit/cce29309392aa0ceca5624f7a52aa593d7072958

netonjm:
tested the error pad fix and seems to work like expected, thanks @avodovnik!

slluis:

PR #8770, opened 2019-09-18
[Unit Tests] avoid running old tests in VsTestProjectTestSuite

ketttY:
When UnitTestService is running the tests, for VsTest, it is using the old tests before all test information is refreshed.
In UnitTestGroup.Refresh(), Tests collection is accessed which triggered VsTestProjectTestSuite’s `OnCreateTests()`. However it will return the control to the calling thread once `VsTestDiscoveryAdapter.Instance.DiscoverTestsAsync (Project);` is called. This causes the tests not properly refreshed.
For the fix, a new method `DiscoverTestAsync()`is created, which does the same thing as old `OnCreateTests()` does, but returns a Task. Local variable `testDiscoveryTask` is created and shared by `OnCreateTests()` and `Refresh()`. We use it to know when `Refresh()` can return.

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/967657/

### DIFF
--- a/main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/VsTestProjectTestSuite.cs
+++ b/main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/VsTestProjectTestSuite.cs
@@ -89,6 +89,7 @@ namespace MonoDevelop.UnitTesting.VsTest
 
 		protected override async void OnCreateTests ()
 		{
+			AddOldTests ();
 			if (testDiscoveryTask == null) {
 				testDiscoveryTask = DiscoverTestsAsync ();
 			}
@@ -164,7 +165,6 @@ namespace MonoDevelop.UnitTesting.VsTest
 	
 		public async override Task Refresh (CancellationToken ct)
 		{
-			AddOldTests ();
 			if (testDiscoveryTask == null) {
 				testDiscoveryTask = DiscoverTestsAsync ();
 			}
--- a/main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/VsTestProjectTestSuite.cs
+++ b/main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/VsTestProjectTestSuite.cs
@@ -44,6 +44,7 @@ namespace MonoDevelop.UnitTesting.VsTest
 		public Project Project { get; private set; }
 		DateTime? lastBuildTime;
 		UnitTest [] oldTests;
+		Task testDiscoveryTask;
 
 		public VsTestProjectTestSuite (Project project)
 			: base (project.Name, project)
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/VsTestProjectTestSuite.cs
```diff
@@ -118,7 +127,7 @@ string GetAssemblyFileName ()
 
 		void AddOldTests ()
 		{
-			if (oldTests != null) {
+			if (oldTests != null && Tests.Count == 0) {
 				foreach (var test in oldTests) {
```
mrward: Just wondering if the `Tests.Count == 0` check would be needed if we moved the `testDiscoveryTask = null` to inside the RefreshRequired if statement. I am guessing that AddOldTests is being called again before the tests have been cleared because DiscoveryTestsAsync is called when it does not need to be. I think moving the `testDiscoveryTask = null` line inside the RefreshRequired makes sense anyway. No need to run a new discovery if the assembly was not updated.

Previously it looks like if RefreshRequired returned false the Tests would not be set to null since UpdateTests was not called.

ketttY: I moved it into the if statement, but from debugging it seems like it is not because of this. Somehow after `Refresh()` initialized `testDiscoveryTask`, when `OnCreateTests()` is triggered from accessing `Tests` collection, `testDiscoveryTask` is still null.

ketttY:
Yea, I was working on locking yesterday. There is a racing condition for `testDiscoveryTask`, sometimes `DiscoverTestAsync()` is called multiple times because `OnCreateTests()` didn't realize the task is not null after `Refresh()` is called.

ketttY:
After checking that again, the reason `DiscoverTestAsync()` is called multiple times is that when `Refresh()` called `DiscoverTestAsync()`, before the method returns the task to `testDiscoveryTask`, `AddOldTests()` is called and it accessed the Tests collection which triggered `OnCreateTests()`. If we move `AddOldTests()` into `Refresh()` before it checks and assigns the Task, we can have `OnCreateTests()` start the task first. And then the null check in `Refresh()` can work properly.

mrward:
One problem here, which I think may be due to the OldTests being moved out of the DiscoverTestsAsync, is that when you build the project after making a change the Unit Tests tree collapses. VS Mac 8.3 does not collapse the tree after building.

Repro:

MSTest project.
Build the project so the tests are discovered.
Expand the tree so all tests are visible.
Make a small change to the tests.
Build - Build All
The unit tests tree collapses.

ketttY:
Good point, let me try to fix that

mrward:
Adding back the AddOldTests call to DiscoveryTestsAsync with the check you added `if (oldTests != null && Tests.Count == 0) {` I am guessing may fix it.

ketttY:
Yea I think so, but you think it is ok that we call `DiscoverTestAsync()` twice?

mrward:
Unless there is a way to avoid it. Maybe we can add a lock?

ketttY:
I moved `AddOldTests()` to `OnCreateTests()`, it's now working I think! I will push it so we can see if I missed something here.

PR #8779, opened 2019-09-19
Fixes VSTS Bug 983399: [VersionControl] Credential dialog is on top of

mkrueger:
all applications

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/983399

ProgressDialog didn't had the parent set. In VS4Mac the parent is a
NSWindow so it's needed to set the parent in that case after the gtk
window has been created.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/ProgressDialog.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/ProgressDialog.cs
@@ -30,6 +30,9 @@ using Gtk;
 using MonoDevelop.Components;
 using MonoDevelop.Core;
 using System.Threading;
+#if MAC
+using AppKit;
+#endif
 
 namespace MonoDevelop.Ide.Gui.Dialogs
 {
@@ -43,6 +46,7 @@ namespace MonoDevelop.Ide.Gui.Dialogs
 		List<TextTag> tags = new List<TextTag> ();
 		Stack<string> indents = new Stack<string> ();
 		CancellationTokenSource cancellationTokenSource;
+		private MonoDevelop.Components.Window componentsWindowParent;
 
 		public ProgressDialog (bool allowCancel, bool showDetails): this (null, allowCancel, showDetails)
 		{
@@ -53,6 +57,7 @@ namespace MonoDevelop.Ide.Gui.Dialogs
 			MonoDevelop.Components.IdeTheme.ApplyTheme (this);
 			this.Build ();
 			this.Title = BrandingService.ApplicationName;
+			this.componentsWindowParent = parent; 
 			HasSeparator = false;
 			ActionArea.Hide ();
 			DefaultHeight = 5;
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/ProgressDialog.cs
```diff
@@ -189,6 +194,27 @@ protected virtual void OnBtnCloseClicked (object sender, System.EventArgs e)
 		{
 			Destroy ();
 		}
+
+		protected override void OnShown ()
+		{
+			base.OnShown ();
+			try {
+#if MAC
+				var nsWindow = componentsWindowParent.GetNativeWidget<NSWindow> ();
```
Therzok: This could easily be an extension method somewhere in `MonoDevelop.Ide`.

Therzok: Or in Window.cs

mkrueger: If we've that exact problem at another area - y. Doing that before would just be another helper method nobody knows about and would never be used anywhere else.

Changing that at that point is easy. We in the shell team know now about that issue. This only happens because the small start window we use now is a NSView.

Therzok: Extending Control to support shown event and doing that in Control sounds better than having to fix in each implementation

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/ProgressDialog.cs
```diff
@@ -189,6 +194,11 @@ protected virtual void OnBtnCloseClicked (object sender, System.EventArgs e)
 		{
 			Destroy ();
 		}
+
+		protected override void OnShown ()
+		{
+			base.OnShown ();
+			this.SetParentToWindow (componentsWindowParent);
```
sevoku: This needs a null check

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/ProgressDialog.cs
```diff
@@ -43,6 +46,7 @@ public partial class ProgressDialog : Gtk.Dialog
 		List<TextTag> tags = new List<TextTag> ();
 		Stack<string> indents = new Stack<string> ();
 		CancellationTokenSource cancellationTokenSource;
+		private MonoDevelop.Components.Window componentsWindowParent;
```
sevoku: the name is confusing, let's call it just `parentWindow`.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Dialogs/ProgressDialog.cs
Therzok: Or maybe SetParentToWindow should handle the null scenario.

mkrueger: SetParentToWindow should log if it's null.

sevoku:

Issue #4727, opened 2018-05-04
quit re-defaulting to "case sensitive" search 


nddipiazza:
hit "ctrl + f" for search
note: case sensitive. So uncheck the case sensitive and make your search.

hit "ctrl + f" for search again
note: case sensitive again. 

should not redefault to that

> VS bug [#611793](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/611793)

xamarin-release-manager:
> [sync] [VS-22] Comment by **xamarinc**



Fixed in version 8.5.0.490 (master)



Pull Request #8785 merged by: Vsevolod Kukol  
 Author: mono  
 Commit: d2deae9cd363ee35946cc45dc2fad46d0af6a8fc (mono/monodevelop)

xamarin-release-manager:
> [sync] [VS-23] Comment by **xamarinc**



Fixed in version 8.4.0.1815 (release-8.4)



Pull Request #8785 merged by: Vsevolod Kukol  
 Author: mono  
 Commit: d2deae9cd363ee35946cc45dc2fad46d0af6a8fc (mono/monodevelop)

xamarin-release-manager:
> [sync] [VS-25] Comment by **xamarinc**



Fixed in version 8.6.0.802 (master)



Pull Request #8785 merged by: Vsevolod Kukol  
 Author: mono  
 Commit: d2deae9cd363ee35946cc45dc2fad46d0af6a8fc (xamarin/monodevelop)

PR #8785, opened 2019-09-20
Fixes VSTS #611793 - quit re-defaulting to case sensitive search #4727

jmatthiesen:
Fixes VSTS #611793 - quit re-defaulting to case sensitive search #4727

By default, in MonoDevelop preferences, the "Automatically Set Pattern Casing" setting is off. However, in the product code when retrieving this setting it's defaulting to enable it. This change modifies the default value (when retrieving this setting from the PropertyService) to match the default setting show in preferences. As I confirmed this worked in the legacy editor, I also found that the Search in Solution feature from the search box also displayed this bug, likewise for the Find for log views.

I've tested and confirmed that search is now case insensitive until Automatically Set Pattern Casing is enabled in preferences by the user. This also matches the behavior in the new editor.


Therzok:
cc @vancura @abock

vancura:
I think this is a meaningful change. @hbons, do you agree?

Therzok:
@sevoku PTAL

PR #8719, opened 2019-09-13
Fix flickering issue in results window

slluis:
Avoid refresing the search results window too often.

Fixes bug #972204 - Too much flickering while typing in searchbar

This is a less intrusive alternative of https://github.com/mono/monodevelop/pull/8703.

PR #8786, opened 2019-09-20
[Ide] Improve global search result reporting even more

sevoku:
This avoids unneeded global search hiding and redrawing
so that we actually don't need to throttle the incoming results
to avoid flickering.

This reverts 731915e557e92dfe7333a1fab6cd37335aae1583 introduced in https://github.com/mono/monodevelop/pull/8719

Fixes VSTS #972204
Fixes VSTS #982384
Fixes VSTS #984594
Fixes VSTS #985811

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/SearchPopupWindow.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/SearchPopupWindow.cs
@@ -553,7 +553,7 @@ namespace MonoDevelop.Components.MainToolbar
 			for (int i = 0; i < newResults.Count; i++) {
 				var tuple = newResults [i];
 				try {
-					if (tuple.Item2.Count == 0)
+					if (tuple.Item2.Count == 0 || tuple.Item1 is LoadingSearchProvidersCategory)
 						continue;
 					if (topResult == null || topResult.DataSource [topResult.Item].Weight < tuple.Item2 [0].Weight)
 						topResult = new ItemIdentifier (tuple.Item1, tuple.Item2, 0);
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/SearchPopupWindow.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/SearchPopupWindow.cs
@@ -353,8 +353,8 @@ namespace MonoDevelop.Components.MainToolbar
 
 		sealed class LoadingSearchProvidersCategory : SearchCategory
 		{
-			Dictionary<SearchCategory, ProviderSearchResult> data = new Dictionary<SearchCategory, ProviderSearchResult> ();
-			List<ProviderSearchResult> values = new List<ProviderSearchResult> ();
+			readonly Dictionary<SearchCategory, ProviderSearchResult> data = new Dictionary<SearchCategory, ProviderSearchResult> ();
+			readonly List<ProviderSearchResult> values = new List<ProviderSearchResult> ();
 			public IReadOnlyList<ProviderSearchResult> Values => values;
 
 			public int ProvidersLeft => values.Count;
@@ -375,12 +375,24 @@ namespace MonoDevelop.Components.MainToolbar
 				data.Add (provider, result);
 			}
 
+			public void AddRange (IEnumerable<SearchCategory> providers)
+			{
+				foreach (var provider in providers)
+					Add (provider);
+			}
+
 			public void Remove (SearchCategory provider)
 			{
 				if (data.TryGetValue (provider, out ProviderSearchResult result)) {
 					values.Remove (result);
 					data.Remove (provider);
 				}
+			}
+
+			public void Clear ()
+			{
+				values.Clear ();
+				data.Clear ();
 			}
 
 			public override Task GetResults (ISearchResultCallback searchResultCallback, SearchPopupSearchPattern pattern, CancellationToken token)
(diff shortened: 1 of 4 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/SearchPopupWindow.cs
```diff
@@ -73,7 +73,7 @@ public SearchPopupWindow () : base (PopupType.Tooltip)
 
 		public void Update (SearchPopupSearchPattern pattern)
 		{
-			Content.Update (pattern);
+			Task.Run (() => Content.Update (pattern)).Ignore ();
```
Therzok: Why Task.Run then?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/SearchPopupWindow.cs
```diff
@@ -428,27 +433,29 @@ public void Update (SearchPopupSearchPattern pattern)
...
 			isInSearch = true;
 			if (results.Count == 0) {
 				QueueDraw ();
 			}
 
-			var newResults = new List<Tuple<SearchCategory, IReadOnlyList<SearchResult>>> ();
-
-			//we want add the search results with a special behaviour
-			newResults.Add (new Tuple<SearchCategory, IReadOnlyList<SearchResult>> (searchProvidersCategory, searchProvidersCategory.Values));
+			var newResults = ImmutableList.Create (new Tuple<SearchCategory, IReadOnlyList<SearchResult>> (searchProvidersCategory, searchProvidersCategory.Values));
```
Therzok: Use ImmutableArray here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/SearchPopupWindow.cs
```diff
@@ -353,8 +353,8 @@ public override string GetMarkupText (bool selected)
 
 		sealed class LoadingSearchProvidersCategory : SearchCategory
 		{
```
mrward: The LoadingSearchProvidersCategory class is not thread safe. Some parts are called on the UI thread but the Remove method is called on the background thread since the TaskScheduler.Default is used with the ContinueWith. Also the Values is accessed on a background thread in the ContinueWith.

There is this bug which is one result of the class not being thread safe - https://devdiv.visualstudio.com/DevDiv/_workitems/edit/982384 - the fix @netonjm was trying here was to run the ContinueWith with the `Runtime.MainTaskScheduler`

sevoku: No that's a different problem fixed with e0874f231dedda128a9c70eea271f340c591ddbd. All operation on this specific category happen inside the lock, this specific category never gets passed to ShowResults.

mrward: The LoadingSearchProvidersCategory is not thread safe and its Values property is used in the tuple so this can change on the background thread whilst the UI thread is using it. So not sure that bug is fixed.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/SearchPopupWindow.cs
```diff
@@ -560,7 +570,7 @@ public void Update (SearchPopupSearchPattern pattern)
 			return (failedResults, topResult);
 		}
 
-		void ShowResults (List<Tuple<SearchCategory, IReadOnlyList<SearchResult>>> newResults, ItemIdentifier topResult)
+		void ShowResults (IList<Tuple<SearchCategory, IReadOnlyList<SearchResult>>> newResults, ItemIdentifier topResult)
```
Therzok:
```suggestion
		void ShowResults (ImmutableArray<Tuple<SearchCategory, IReadOnlyList<SearchResult>>> newResults, ItemIdentifier topResult)
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/SearchPopupWindow.cs
```diff
@@ -499,45 +507,47 @@ public void Update (SearchPopupSearchPattern pattern)
...
 					.Ignore ();
 			}
 		}
 
-		Task nextUpdate = Task.CompletedTask;
-
 		readonly object lockObject = new object ();
 
-		(List<Tuple<SearchCategory, IReadOnlyList<SearchResult>>> failedResults, ItemIdentifier topResult) GetTopResult (List<Tuple<SearchCategory, IReadOnlyList<SearchResult>>> newResults)
+		(List<Tuple<SearchCategory, IReadOnlyList<SearchResult>>> failedResults, ItemIdentifier topResult) GetTopResult (IList<Tuple<SearchCategory, IReadOnlyList<SearchResult>>> newResults)
```
Therzok:
```suggestion
		(List<Tuple<SearchCategory, IReadOnlyList<SearchResult>>> failedResults, ItemIdentifier topResult) GetTopResult (ImmutableArray<Tuple<SearchCategory, IReadOnlyList<SearchResult>>> newResults)
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.MainToolbar/SearchPopupWindow.cs
sevoku: Oh right, sorry I missed that, now it should be good: d7debf46c84c21c180deb90ec0b2362958757814

Therzok:
Some unit tests would be great, but that would take some time.

DavidKarlas:
Why am I seeing `FAKE search provider` quickly before `file:` results are returned?

sevoku:
@DavidKarlas: fixed with 56cdfd452c8c210a7b2fbe185c9adccfcbefe33a

netonjm:
detected overlap in loading section, the item it's not removed correctly when there are no providers left
image

sevoku:
@netonjm @DavidKarlas that was a glitch with top result logic, fixed: 23e3c66

sevoku:

sevoku:

PR #8791, opened 2019-09-20
[Debugger] Fixed VsCode backend to bypass the NRefactory expression r…

jstedfast:
…esolver

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/939728/


jstedfast:
These changes broke the unit tests that evaluate a type name where the expected result is the full type name.

Probably all for the best anyway since this would have only fixed it for the VsCode backend and presumably the SDB backend has the same issue.

The difference, though, was that SDB *needs* the expression resolver while the VsCode backend technically doesn't, but it *does* if we want it to match SDB's behavior.

The solution *may* require killing the NRefactory-based expression resolver and using Roslyn instead. I'm not sure if that will solve this problem or not though, because the main problem here is that the NRefactory expression resolver doesn't have enough context to decide if an Identifier expression is a variable name vs a type name and so to make that decision, it checks to see if it *could* be a type name and if so, assumes it must be.

PR #8792, opened 2019-09-20
Debugger: Set position when using New Breakpoint with new editor

sandyarmstrong:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/824300


sandyarmstrong:
@jstedfast yup! Old editor implements this API too. Tested with c# and f#.

PR #8794, opened 2019-09-23
Add support for Blazor server side projects

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/943227
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/944132
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/980398

### DIFF
--- a/main/src/addins/MonoDevelop.AspNetCore/Properties/MonoDevelop.AspNetCore.addin.xml
+++ b/main/src/addins/MonoDevelop.AspNetCore/Properties/MonoDevelop.AspNetCore.addin.xml
@@ -142,6 +142,7 @@
 				icon="md-netcore-empty-project"
 				imageId="md-netcore-empty-project"
 				wizard="MonoDevelop.AspNetCore.ProjectTemplateWizard"
+				supportedParameters="AspNetCoreWorker"
 				condition="UseNetCore30=true"
 				category="netcore/app/aspnet"
 				defaultParameters="IncludeLaunchSettings=true" />
@@ -194,6 +195,7 @@
 				icon="md-netcore-empty-project"
 				imageId="md-netcore-empty-project"
 				wizard="MonoDevelop.AspNetCore.ProjectTemplateWizard"
+				supportedParameters="AspNetCoreBlazor"
 				condition="UseNetCore30=true"
 				category="netcore/app/aspnet"
 				defaultParameters="IncludeLaunchSettings=true" />
--- a/main/src/addins/MonoDevelop.AspNetCore/Properties/MonoDevelop.AspNetCore.addin.xml
+++ b/main/src/addins/MonoDevelop.AspNetCore/Properties/MonoDevelop.AspNetCore.addin.xml
@@ -43,6 +43,8 @@
 		              file = "Templates/StartupClass.xft.xml"/>
 		<FileTemplate id   = "AspNetCoreWebApiController"
 		              file = "Templates/WebApiController.xft.xml"/>
+		<FileTemplate id   = "AspNetCoreRazorComponent"
+					  file = "Templates/RazorComponent.xft.xml"/>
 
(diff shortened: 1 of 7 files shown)


main/src/addins/MonoDevelop.AspNetCore/Properties/MonoDevelop.AspNetCore.addin.xml
```diff
@@ -183,6 +185,18 @@
...
+				_overrideName="Blazor Server App"
+				_overrideDescription="A project template for creating a Blazor server app that runs server-side inside an ASP.NET Core app and handles user interactions over a SignalR connection. This template can be used for web apps with rich dynamic user interfaces (UIs)."
+				path="${DotNetCoreSdk.3.0.Templates.Web.ProjectTemplates.nupkg}"
+				icon="md-netcore-empty-project"
+				imageId="md-netcore-empty-project"
+				wizard="MonoDevelop.AspNetCore.ProjectTemplateWizard"
+				condition="UseNetCore30=true"
+				category="netcore/app/aspnet"
+				defaultParameters="IncludeLaunchSettings=true" />
 			<Template
```
mrward: Need to check how the new project dialog works here. Will it show all .NET Core runtimes installed? May need to filter them.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Tests/MonoDevelop.DotNetCore.Tests/DotNetCoreProjectTemplateTests.cs
```diff
@@ -35,6 +35,7 @@
 using IdeUnitTests;
 using MonoDevelop.Ide.Projects.FileNesting;
 using MonoDevelop.Core.Execution;
+using ICSharpCode.Decompiler.IL;
```
mrward: I guess this is a mistake :)

rodrmoya: yeah, fixed

main/src/addins/MonoDevelop.AspNetCore/Properties/MonoDevelop.AspNetCore.addin.xml
rodrmoya: what do you mean, on project creation if it will ask for SDK even though this is only for 3.0?

mrward: In the New Project dialog there is a drop down list of target frameworks. By default we show all the runtimes installed.

https://github.com/mono/monodevelop/blob/237b9b23781b583bc70a122faeac6aae09b5347f/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Templating/DotNetCoreProjectTemplateWizard.cs#L87

I am assuming that showing .NET Core 2.1 in this list does not make sense for the Blazor project template? I think we need to special case this project template, like we do for the Razor templates with the SupportsNetCore1x but for NetCore 2x.

rodrmoya: Right, it offers 2.2, which is wrong, so I'll add a "Blazor" parameter to check for these, as well as for "Worker", recently added which also has the same issue.

rodrmoya: Fixed in https://github.com/mono/monodevelop/pull/8794/commits/b83534131efe43611d2dd85a5e2bea73c52a6c0b

PR #8795, opened 2019-09-23
[Ide] Mark NSViewExtensions as obsolete and remove it in future versions

Therzok:
The method is implemented in XamMac, no need to duplicate the implementation
Fixes VSTS #973862 - Type forward NSViewExtensions.SortSubviews to Xamarin.Mac


Therzok:
@sevoku PTAL

PR #8802, opened 2019-09-25
Fixes VSTS Bug 988755: [Feedback] iOS project will not build

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/988755

ErrorListPad is mandatory for the IDE to work. A null progress monitor
will prevent the crash here, but that needs to be logged. The earliest
point to look for the ErrorListPad is the initialization.
It may be that the pad gets removed after the initialization - the
logging will tell us if it's never loaded.


mkrueger:

mkrueger:
Atm it's based on tip of master - will rebase tomorrow to re run the build.

slluis:
Not merging this for now. The crash doesn't make any sense and the original log has some weird initialization exceptions. We'll reopen the bug if this is hit in a stable build.

PR #8808, opened 2019-09-26
Fix crash in global search

slluis:
Added some locks required when accessing data that can mutate while
running the search.

Hopefully fixes VSTS #984594


slluis:

slluis:

PR #8811, opened 2019-09-26
[F#] Fix recursion stack overflow when parsing big files

nosami:
Before this fix, the stack could overflow with ~12000 lines on my
machine.

Unit test is using 100000 lines without a crash
Fixes https://dev.azure.com/devdiv/DevDiv/_workitems/edit/988934


nosami:

PR #8816, opened 2019-09-30
Add null check for accessibility and remove redundant one

Therzok:
Runtime.GetNSObject does a null check


Therzok:
Without this, we don't have an editor when we open a document.

sevoku:
CI is broken

Issue #5052, opened 2018-06-13
NuGet addin has quadratic behaviour on update


Therzok:
https://github.com/mono/monodevelop/blob/master/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/CheckForNuGetPackageUpdatesTaskRunner.cs#L63-L85

This bit of code is doing unnecessary querying to nuget. Since we have the same repository provider, we should not be querying each package for each project.

We should be doing a set of distinct packages we should query, and only query those.

Found via device not being connected to internet, the log is spammed with 2800 of these when opening Main.sln.

```
ERROR [2018-06-13 12:19:50Z]: Check for updates error.
NuGet.Protocol.Core.Types.FatalProtocolException: Unable to load the service index for source https://dotnet.myget.org/F/roslyn-for-vs-for-mac/api/v3/index.json. ---> System.Net.Http.HttpRequestException: An error occurred while sending the request ---> System.Net.WebException: Error: NameResolutionFailure
  at System.Net.WebConnection+<Connect>d__16.MoveNext () [0x0002c] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-04/external/bockbuild/builds/mono-x64/mcs/class/System/System.Net/WebConnection.cs:137 
--- End of stack trace from previous location where exception was thrown ---
  at System.Net.WebConnection+<InitConnection>d__19.MoveNext () [0x000cc] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-04/external/bockbuild/builds/mono-x64/mcs/class/System/System.Net/WebConnection.cs:259 
--- End of stack trace from previous location where exception was thrown ---
  at System.Net.WebOperation+<Run>d__57.MoveNext () [0x00052] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-04/external/bockbuild/builds/mono-x64/mcs/class/System/System.Net/WebOperation.cs:268 
--- End of stack trace from previous location where exception was thrown ---
  at System.Net.WebCompletionSource`1+<WaitForCompletion>d__15[T].MoveNext () [0x0008e] in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-04/external/bockbuild/builds/mono-x64/mcs/class/System/System.Net/WebCompletionSource.cs:111 
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #8818, opened 2019-09-30
[NuGet] Update multiple packages together in a batch

mrward:
Updating two NuGet package references that have a strict dependency
on a single version of another NuGet package not explicitly added
to the project would fail in the Manage NuGet Packages dialog. This
was because the update would be done one package at a time. Doing
the update of the two NuGet packages together allows NuGet to
update both so they then use the new strict dependency.

Note that updating all packages in the project from the Solution
window does not have this problem since there the packages are
updated together in a batch.

Fixes VSTS #986960 - Solution nuget updater fails to update a set of
packages that all have a related dependency update


mrward:

Therzok:
Is this related? https://github.com/mono/monodevelop/issues/5052

mrward:
#5052 is about checking for updates and is not related.

mrward:

mrward:

Therzok:
@monojenkins rebase

PR #8824, opened 2019-09-30
bumps Xamarin.PropertyEditing

netonjm:
This bumps includes a fix in Proppy to fix layout issues in VS4Mac when native proppy is added

Fixes VSTS #976526 - VSMac layout is screwed up when Proppy initially activated


garuma:
Supersede by https://github.com/mono/monodevelop/pull/8828

PR #8832, opened 2019-10-01
Fix 985131: Unable to pull up the emoji selector

abock:
The check for an empty `NSEvent.Characters` is too exhaustive for attempting to validate an event. The explicit check for Key Up/Down should be enough.

Apple documentation states that dead keys will result in an empty string for `NSEvent.Characters`, but these keys should be allowed to be processed as well by commanding.

This fixes a regression introduced in #8664.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/985131

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -317,11 +317,6 @@ namespace MonoDevelop.Components.Commands
 				return null;
 			}
 
-			if (string.IsNullOrEmpty (ev.Characters)) {
-				LoggingService.LogInternalError (new Exception ($"Keyevent has no characters"));
-				return null;
-			}
-
 			// If we have a native window that can handle this command, let it process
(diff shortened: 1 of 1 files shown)


abock:

iainx:
The problem that the characters check was supposed to fix was that there is a crash when an event  arrives in `OnNSEventKeyPress` whose underlying Carbon event structure is completely missing a CharCode property. Not that the property is empty, it's not there, which, when passed into `PerformKeyEquivalent` throws an ObjC exception.

It seems like this is caused by key remapper apps (see https://github.com/tekezo/Karabiner-Elements/issues/608 for example), but I've not been able to reproduce it with Ukulele

Does checking `Characters` fix it? I don't know, it was just a bit of extra protection, because I didn't think that simply checking for KeyUp/Down was enough.

abock:
I think we need to take this PR and keep an eye on telemetry again. It's too defensive of a check and breaks downstream components (the editor) that handle expected events that have `.Characters` set to an empty string ([dead keys](https://en.wikipedia.org/wiki/Dead_key)).

PR #8839, opened 2019-10-02
Fixes VSTS Bug 810330: Accessibility: Version Control - Changes: Voice

mkrueger:
Over was reading the drop-down's in the window as only buttons. It
should provide description related to drop-down, role, selected item
and state of drop-down.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/810330

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/ComparisonWidget.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/ComparisonWidget.cs
@@ -36,6 +36,7 @@ using MonoDevelop.Core;
 using MonoDevelop.Ide.Gui;
 using MonoDevelop.Ide.Gui.Content;
 using System.Threading.Tasks;
+using Microsoft.TemplateEngine.Core.Contracts;
 
 namespace MonoDevelop.VersionControl.Views
 {
@@ -93,15 +94,23 @@ namespace MonoDevelop.VersionControl.Views
 				originalComboBox.Text = GettextCatalog.GetString ("Loading…");
 				originalComboBox.Sensitive = false;
 				originalComboBox.Tag = editors[1];
+				originalComboBox.TextChanged += UpdateAccessibility;
 
 				diffComboBox = new DropDownBox ();
 				diffComboBox.WindowRequestFunc = CreateComboBoxSelector;
 				diffComboBox.Text = GettextCatalog.GetString ("Loading…");
 				diffComboBox.Sensitive = false;
 				diffComboBox.Tag = editors[0];
-
+				diffComboBox.TextChanged += UpdateAccessibility;
 				this.headerWidgets = new [] { diffComboBox, originalComboBox };
+				UpdateAccessibility (this, EventArgs.Empty);
 			}
+		}
+
+		void UpdateAccessibility (object sender, EventArgs e)
+		{
+			originalComboBox.Accessible.Description = GettextCatalog.GetString ("Select original revision, current: {0}", originalComboBox.Text);
+			diffComboBox.Accessible.Description = GettextCatalog.GetString ("Select diff revision, current: {0}", diffComboBox.Text);
 		}
 
 		protected override void OnSetVersionControlInfo (VersionControlDocumentInfo info)
@@ -121,6 +130,10 @@ namespace MonoDevelop.VersionControl.Views
 		protected override void OnDestroyed ()
 		{
 			info.Updated -= OnInfoUpdated;
+			if (originalComboBox != null)
+				originalComboBox.TextChanged -= UpdateAccessibility;
+			if (diffComboBox != null)
+				diffComboBox.TextChanged -= UpdateAccessibility;
 			base.OnDestroyed ();
 		}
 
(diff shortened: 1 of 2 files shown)


main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/DropDownBox.cs
```diff
@@ -48,9 +48,12 @@ public class DropDownBox : Gtk.Button
 			}
 			set {
 				layout.SetText (value);
+				TextChanged?.Invoke (this, EventArgs.Empty);
 //				QueueResize ();
 			}
 		}
+
+		public event EventHandler TextChanged;
```
Therzok: Since this is going to be public API, should we have an explicit TextChangedEventArgs which contain the old text?

mkrueger: For a control like that it's unusual. I don't think it's useful here it's not a text container.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/ComparisonWidget.cs
```diff
@@ -36,6 +36,7 @@
 using MonoDevelop.Ide.Gui;
 using MonoDevelop.Ide.Gui.Content;
 using System.Threading.Tasks;
+using Microsoft.TemplateEngine.Core.Contracts;
```
Therzok: Should be removed?

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/DropDownBox.cs
Therzok: It just seems backwards to have an event here. Maybe a SetText method which also allows you to set the accessibility description, would be better or a way to pass an AccessibleFormatText and using string.Format when text is set in the dropdownbox.

sevoku: I agree, an `AccessibleFormatText` would simplify a lot, no event handling, and everything inside DropDownBox.

mkrueger: Text can be changed with an item as well. 

It's not just for accessiblity it's a component that fires when a property has changed here - nothing more. This is not unusual.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/ComparisonWidget.cs
mkrueger: +

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/DropDownBox.cs
```diff
@@ -48,9 +48,12 @@ public class DropDownBox : Gtk.Button
 			}
 			set {
 				layout.SetText (value);
+				TextChanged?.Invoke (this, EventArgs.Empty);
```
Therzok: Again, an event handler doesn't make much sense. Set a AccessibilityTextFormat property to:

`AccessibilityTextFormat = GettextCatalog.GetString ("Text here {0}");` and use string.Format everywhere where text is set should do the trick.

Therzok: In a previous comment you said that the text can be changed another way. This PR doesn't seem to handle this.

Therzok: It is the responsibility of the Dropdown control to set its accessibility properties, not of its user

mkrueger: This PR handles that. It fires on any text change.

AccessibilityTextFormat doesn't make sense. There is an Accessible object that handles that.

sevoku: I totally agree with Marius, the Accessible object is just an API for common use. The regular native behaviour is that if i.e. a label changes, that a narrator (here VO on Mac) will read the changed label without the need to explicitely set the a11y property. All objects should behave like that from a client perspective. So, in general, all custom widgets should follow that pattern: If `Text` changes, a11y changes automatically.

mkrueger: "not the responsibility of the widget creator to set the accessibility text"

Ah ok now I understand what you're meaning. Y I think that's right everywhere the DropDown is used the accessibility should work :).

mkrueger:
@iainx I don't want to change UI controls in an accessiblitiy PR. We need to redesign the UI in any case that's the time we'll change the controls.

sevoku:

PR #8840, opened 2019-10-02
[Core] Fix project modified fired for refs when they have not changed

mrward:
Originally the Project.Modified event was fired with a Hint set to
References when a ProjectReference was added to or removed from the
project. This was changed in commit 90d690e56f2e100d8dea11bf3ad690f9eefaeb4f
and now the event is fired for any ProjectItem being added/removed.
This can cause the type system to try to update its type information
when it should not. One example is updating a NuGet PackageReference
which would trigger this event before the updated reference information
will be available for the type system.

Fixes VSTS #992017 - Modified event fires for References when non
reference added/removed from project


Therzok (approved):

mrward:

Therzok (approved):

Therzok:
@monojenkins rebase

sgmunn (approved):

PR #8841, opened 2019-10-02
SaveJson - Serialize "protected List<CounterValue> values" in Counter

manish:
Fixes #992546

We wish to save all the counter data on the disk, but since `Counter.values` is protected, it can't be serialized. Create a custom `ContractResolver` named `CounterContractResolver` as per @Therzok suggestion.

This only helps in serialization and not deserialization (which should use JObject).

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/Counter.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/Counter.cs
@@ -472,15 +472,16 @@ namespace MonoDevelop.Core.Instrumentation
 		Line
 	}
 
-	public class CounterContractResolver : DefaultContractResolver
+	internal class CounterContractResolver : DefaultContractResolver
 	{
 		protected override List<MemberInfo> GetSerializableMembers (Type objectType)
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/Counter.cs
```diff
@@ -468,4 +471,26 @@ public enum CounterDisplayMode
 		Block,
 		Line
 	}
+
+	public class CounterContractResolver : DefaultContractResolver
```
Therzok:
```suggestion
	internal class CounterContractResolver : DefaultContractResolver
```

mrward (commented):
Just wondering if we can avoid using reflection here. Could we instead do one of the following?

1. Have Counter implement ISerializable?
2. Add an internal, or maybe even public, Values property?

Therzok:
We have a `GetValues()` public API already

Therzok:
Did not know that! Counter already is Serializable, so that does seem like the simpler approach. Good tip

mrward:
Replaced by #8846

PR #8844, opened 2019-10-03
[NuGet] Fix requests bypassing the native HTTP handler 

mrward:
Restoring or installing a NuGet package into a project that uses
PackageReferences would bypass the native HTTP handler. This was due
to NuGet's CachingSourceProvider creating its own set of
INuGetResourceProviders when it creates a set of new SourceRepository
instances. This resulted in the default HttpSourceResourceProvider
being used when uses Mono's HttpClientHandler. This would happen
when the .nupkg was downloaded.

NuGet 5.4 allows a custom NuGetResourceProvider factory to be
configured instead of only when creating instances of SourceRepository.
This is now used by the Package Management addin so all HTTP requests
use the native HTTP handler.

Fixes VSTS #918815 - Fixes NuGet requests bypassing native Http Handler
Update to NuGet 5.4.0.6315

Fixes VSTS #1001701 - Bump to NuGet 5.4


mrward:
Mono 6.8.0.87 is shipping NuGet 5.5 preview 1 with its MSBuild. It is shipping NuGet 5.4 for the `nuget` command. Updating to NuGet 5.4 final release here in VS Mac.

mrward:
Xamarin.Mac compat check failure are a known issue with pull requests currently.

PR #8845, opened 2019-10-03
[Core] Fix test to work with new gtk-sharp.dll.pdb

kdubau:
Mono 2019-08 v6.6.0.104 changed the compiler used to build gtk-sharp from `mcs` to `csc`.
As a result, we do not have .mdb files anymore, rather the normal .pdb.

This fixes the test to look for the right file.


kdubau:

PR #8846, opened 2019-10-03
SaveJson - Serialize "protected List<CounterValue> values" in Counter,TimerCounter using ISerializer

manish:
Supersedes #8841

Fixes #992546

We wish to save all the counter data on the disk, but since Counter.values is protected, it can't be serialized. Implement `ISerializable` over `Counter` and `TimerCounter`

This only helps in serialization and not deserialization (which should use JObject). Deserialization would be cumbersome and error prone since many fields are `read` only and we are also saving some state, which not not be valid if we deserialize.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimerCounter.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimerCounter.cs
@@ -143,7 +143,7 @@ namespace MonoDevelop.Core.Instrumentation
 			return c;
 		}
 
-		public new void GetObjectData (SerializationInfo info, StreamingContext context)
+		public override void GetObjectData (SerializationInfo info, StreamingContext context)
 		{
 			base.PopulateSerializableMembers (info, context);
 			info.AddValue (nameof (this.MinSeconds), this.MinSeconds);
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimerCounter.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimerCounter.cs
@@ -33,11 +33,12 @@ using System.Globalization;
 using System.Runtime.CompilerServices;
 using System.Collections.ObjectModel;
 using System.Linq;
+using System.Runtime.Serialization;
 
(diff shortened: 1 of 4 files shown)


Therzok (approved):

main/tests/MonoDevelop.Core.Tests/MonoDevelop.Core/InstrumentationTests.cs
```diff
@@ -219,5 +222,41 @@ public void CounterMetadataQueryingDoesNotCrash ()
 
 			Assert.AreEqual (default (int), metadata.SomeMeasure);
 		}
+
+		[Test]
+		public void SerializeCounters ()
+		{
+			InstrumentationService.SaveJson ("serialize_counters.json");
+			using (var textReader = new StreamReader("serialize_counters.json")) {
```
mrward: Can we test for specific values instead of just not null?

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/TimerCounter.cs
```diff
@@ -141,6 +142,17 @@ internal ITimeTracker<T> BeginTiming<T> (string? message, T? metadata, Cancellat
 			}
 			return c;
 		}
+
+		public new void GetObjectData (SerializationInfo info, StreamingContext context)
```
mrward: Could just make the base method virtual here.

manish: Yes, i can make this virtual and other as override.

main/tests/MonoDevelop.Core.Tests/MonoDevelop.Core/InstrumentationTests.cs
mrward: Can we not check the original counter properties?

mrward: Another way would be to just serialize the Counter and TimerCounter and check the values are written out as expected.

manish: yes, I can match the serialized values with the counter value

Therzok (approved):

PR #8847, opened 2019-10-03
Fix two sources of null reference exceptions

nosami:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/861001

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs
@@ -1002,22 +1002,22 @@ namespace MonoDevelop.Ide.Gui.Components
 
 			public override TextReader In {
 				get {
-					return monitor.inputReader;
+					return monitor?.inputReader;
 				}
 			}
 			public override TextWriter Out {
 				get {
-					return monitor.Log;
+					return monitor?.Log;
 				}
 			}
 			public override TextWriter Error {
 				get {
-					return monitor.ErrorLog;
+					return monitor?.ErrorLog;
 				}
 			}
 			public override TextWriter Log {
 				get {
-					return monitor.internalLogger;
+					return monitor?.internalLogger;
 				}
 			}
 
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
@@ -787,7 +787,7 @@ namespace MonoDevelop.Debugger
 
 			public void LogWriter (bool iserr, string text)
 			{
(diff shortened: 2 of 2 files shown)


Therzok:
We shouldn't ever have a null monitor.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs
```diff
@@ -1002,22 +1002,22 @@ public LogViewProgressConsole (LogViewProgressMonitor monitor)
...
 			public override TextReader In {
 				get {
-					return monitor.inputReader;
+					return monitor?.inputReader;
 				}
 			}
 			public override TextWriter Out {
 				get {
-					return monitor.Log;
+					return monitor?.Log;
```
Therzok: If monitor shouldn't be null, usually you pass a `new ProgressMonitor()`

nosami: the only place I can see this happening is after the monitor has been disposed.

Therzok: Then we have a task continuation executing code after dispose.

nosami: I can't repro this issue at all, it looks like it should never be null, but it looks like it could happen after the monitor has been intentionally disposed

Therzok: _looking at the bug report_

Therzok: @nosami to me, it looks like we're disposing things in the wrong order or a race somewhere.

DebuggingService.Cancel is hooked on the AsyncOperation cancellation [here](http://source.monodevelop.com/#MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs,d3d4ca45939ba1a4,references)

This is [LogViewProgressConsole.Dispose](http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Ide.Gui.Components/LogView.cs,538e6b5dfe066f4d,references) which sets the monitor to null. Given we have a NRE, these are properly disposed, not sure if at the right time.

Is any output sent after [TargetExited](http://source.monodevelop.com/#MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs,624) is called?

I'm wondering if a test app which writes messages every millisecond or something could trigger it.

nosami: @Therzok I'll see if I can try and create an app that triggers this bug tomorrow.

PR #8851, opened 2019-10-04
Fixes VSTS Bug 810342: Accessibility: Version Control - Log: Voice

mkrueger:
Over was taking the code getting displayed after expanding the row as
a row and reading it as empty it as row with empty content.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/810342


main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/LogWidget.cs
```diff
@@ -204,17 +206,26 @@ public LogWidget (VersionControlDocumentInfo info)
 			treeviewLog.Selection.Changed += TreeSelectionChanged;
 
 			treeviewFiles = new FileTreeView ();
-			treeviewFiles.DiffLineActivated += HandleTreeviewFilesDiffLineActivated;
+
 			scrolledwindowFiles.Child = treeviewFiles;
 			scrolledwindowFiles.ShowAll ();
 
+			treeviewFileContents = new FileTreeView ();
+			treeviewFileContents.DiffLineActivated += HandleTreeviewFilesDiffLineActivated;
```
Therzok: Need to remove this handler.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/LogWidget.cs
```diff
@@ -239,10 +250,14 @@ public LogWidget (VersionControlDocumentInfo info)
...
+			treeviewFileContents.HeadersVisible = false;
+			treeviewFileContents.AppendColumn (colChangedPath);
+			treeviewFileContents.Model = contentspathstore;
+			treeviewFileContents.Events |= Gdk.EventMask.PointerMotionMask;
+
+
 			treeviewFiles.Model = changedpathstore;
-			treeviewFiles.TestExpandRow += HandleTreeviewFilesTestExpandRow;
-			treeviewFiles.Events |= Gdk.EventMask.PointerMotionMask;
+			treeviewFiles.Selection.Changed += SetDiff;
```
Therzok: This handler needs to be removed.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/LogWidget.cs
mkrueger: Why was this not a problem before ?

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/LogWidget.cs
mkrueger: same as above /fixed

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/LogWidget.cs
Therzok: It is a problem though. It's just not something I've specifically profiled for in the past.

mkrueger:
1. We only have the cell renderer therefore only the tree view can display it. We should've a text editor based diff view for showing diffs.

2. This is very difficult maybe just make a group for each line ?

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/DiffRendererWidget.cs
```diff
@@ -537,5 +573,72 @@ static int ParseCurrentLine (string line)
...
+			public DiffLineAccessible (DiffRendererWidget widget, int x, int y, BlockType blockType, int line, string text)
+			{
+				this.widget = widget;
+				this.line = line;
+
+				Accessible = AccessibilityElementProxy.ButtonElementProxy ();
+				Accessible.GtkParent = widget;
+				Accessible.PerformPress += PerformPress;
+				string msg;
+				switch (blockType) {
```
Therzok: Can we extract this into a method?

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/DiffRendererWidget.cs
```diff
@@ -0,0 +1,644 @@
...
+		public event EventHandler<int> DiffLineActivated;
+
+		protected virtual void OnDiffLineActivated (int line)
+		{
+			DiffLineActivated?.Invoke (this, line);
+		}
+
+		public DiffRendererWidget ()
+		{
+			Events |= EventMask.PointerMotionMask | EventMask.LeaveNotifyMask | EventMask.ButtonPressMask;
```
sevoku: The widget should have the `Group` accessibility role

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/LogWidget.cs
```diff
@@ -204,10 +204,15 @@ public LogWidget (VersionControlDocumentInfo info)
 			treeviewLog.Selection.Changed += TreeSelectionChanged;
 
 			treeviewFiles = new FileTreeView ();
-			treeviewFiles.DiffLineActivated += HandleTreeviewFilesDiffLineActivated;
+
 			scrolledwindowFiles.Child = treeviewFiles;
 			scrolledwindowFiles.ShowAll ();
 
+			diffRenderer = new DiffRendererWidget ();
```
sevoku: This should have a label like "Diff View" and once a file is selected on the left hand side, it should be updated with the file name.

sevoku:
@vancura @hbons @maryanne-exe could you guys take a quick look? Screenshot is here: https://github.com/mono/monodevelop/pull/8851#issuecomment-539537863

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/DiffRendererWidget.cs
mkrueger: In general I like to extract such things but here I don't see it as improvement because it would scatter the accessibility messages.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/DiffRendererWidget.cs
mkrueger: done.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/LogWidget.cs
mkrueger: done. But needed to be set to the scrolled window Accessibility.

vancura:
FYI I commented in the [DevOps thread](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/810342).

jmatthiesen:
Is there any way to find the full path to a file in this new view? That detail is important when files are spread across numerous folders in a commit, more so when there are files with the same name in multiple folders. I tried to check in the build associated with this PR, but the editor won't open in this build (I just get a blank tab).

iainx:
This is working really well from an accessibility point of view. From a UI design perspective, I think @jmatthiesen is right, we need the  full path somewhere in the UI, and we also need some padding around the `46,7 -+ 46,9` in the above screenshot

It might be able to fix both these issues at once by adding the file path above the numbers

sevoku:

sevoku:

sevoku:
@monojenkins rebase

PR #8852, opened 2019-10-04
Fixes VSTS Bug 810343: Accessibility: Document View Tabs: Voice Over

mkrueger:
was not providing any information related to whether the "Source",
"Changes" etc are selected or not.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/810343

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Tabstrip.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Tabstrip.cs
@@ -42,10 +42,8 @@ namespace MonoDevelop.Components
 {
 	class Tabstrip : DrawingArea
 	{
-
 		readonly List<Tab> tabs = new List<Tab> ();
 		readonly List<Cairo.PointD> tabSizes = new List<Cairo.PointD> ();
-
 
 		double mx, my;
 		Tab hoverTab;
@@ -109,8 +107,8 @@ namespace MonoDevelop.Components
 			else if (activeTab >= index)
 				activeTab++;
 
-			if (focusedTab >= index) {
-				focusedTab++;
+			if (FocusedTab >= index) {
+				FocusedTab++;
 			}
 
 			QueueResize ();
@@ -132,19 +130,19 @@ namespace MonoDevelop.Components
 					ActiveTab = index - 1;
 			}
 
-			if (focusedTab == index) {
+			if (FocusedTab == index) {
 				if (index == tabs.Count - 1)
-					focusedTab--;
-			} else if (focusedTab > index)
-				focusedTab--;
+					FocusedTab--;
+			} else if (FocusedTab > index)
+				FocusedTab--;
 
 			var tab = tabs [index];
 			tabs.RemoveAt (index);
 			tabSizes.RemoveAt (index);
 			if (activeTab > index)
 				activeTab--;
-			if (focusedTab >= index)
-				focusedTab = index;
+			if (FocusedTab >= index)
+				FocusedTab = index;
 
 			QueueResize ();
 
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Tabstrip.cs
```diff
@@ -345,36 +343,46 @@ protected override bool OnExposeEvent (Gdk.EventExpose evnt)
 		}
 
 		int focusedTab = -1;
+
+		int FocusedTab {
+			get => focusedTab;
+			set {
+				focusedTab = value;
+				if (focusedTab >= 0 && focusedTab < tabs.Count)
+					Accessible.Description = GettextCatalog.GetString ("Focused tab {0}", this.tabs[focusedTab].Label);
```
sevoku: I don't think that this is necessary, the native tab control (i.e. System Preferences > Keyboard) does not announce the currently selected tab when it receives focus. Also "Focused tab X" is misleading here, the focused tab is the one which has the VO focus, not the selected one, so if we add this to the description, it should be "Selected tab X".

mkrueger: /removed

y the "not announce" part drove me crazy - but ok it's not part of the original bug.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Tabstrip.cs
```diff
@@ -345,36 +343,44 @@ protected override bool OnExposeEvent (Gdk.EventExpose evnt)
 		}
 
 		int focusedTab = -1;
+
+		int FocusedTab {
```
Therzok: This property is now redundant. Revert these changes

Therzok: It doesn't make any sense to have a private property which is similar to an auto property. Either removed `focusedTab` or remove the property alltogether. It's just noise and error prone. It's really easy to set either focusedTab and FocusedTab, and if it's no longer an autoproperty, it can break.

netonjm:
The PR seems include 2 atomic changes, the FocusedTab refactoring and the fix accessibility fix. 
It would be great to separate to in case of regression detect easy the problem.

And related the accessibility fix in Accessible.SetRole (AtkCocoa.Roles.AXRadioButton, active ? "active tab" : "tab"); maybe we could change only the message instead set the role?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Tabstrip.cs
```diff
@@ -543,11 +555,12 @@ public Tab (Tabstrip parent, string label, TabPosition tabPosition)
 
 			if (AccessibilityElementProxy.Enabled) {
 				Accessible = AccessibilityElementProxy.ButtonElementProxy ();
-				Accessible.SetRole (AtkCocoa.Roles.AXRadioButton, "tab");
 				Accessible.Title = label ?? "";
+				Accessible.SetRole (AtkCocoa.Roles.AXRadioButton, "tab");
```
Therzok: This change is redundant since UpdateAccessibility does it.

mkrueger: removed - but was old code no idea why it moved :/

sevoku:

sevoku:

garuma:
See https://github.com/mono/monodevelop/pull/8925#pullrequestreview-304215669

mkrueger:
Comment such stuff on VSTS, it's not reflected there.

PR #8855, opened 2019-10-04
[PerfTests] Add create new solution tests

mrward:
- Add a test that creates a new console project.
 - Add a test that creates a Xamarin.Forms solution.
 - Fixed the NewProjectController UITest helper - was not selecting the template categories in the New Project dialog.

Need to get some timings and update the baseline xml before this can be merged.


mrward:
The new perf tests are failing to select the template category on VSTS. No idea why this is. They work locally. There seemed to be a problem with the New Project dialog not getting focus sometimes testing this locally.

```
System.TimeoutException : Timeout while executing WaitForElement: c => c.TreeView ().Marked ("templateCategoriesTreeView").Model ("templateCategoriesListStore__Name").Text ("Multiplatform", False).Children ().Text ("App", True).Selected ();
 ```

Using basically the same code as in md-addins here - with some retries.

PR #8861, opened 2019-10-07
[Core] Fix ReferencePath items being ignored by the type system

mrward:
The DotNetProject's GetReferencedAssemblies method was not returning
all references used by a project. A project may define extra
assemblies by adding ReferencePath items. These were being
filtered out since only the output parameters defined by the targets
ResolveAssemblyReferencesDesignTime and ResolveProjectReferencesDesignTime
which missed some references.

Fixes VSTS #991961 - Project model reports wrong references for
WebTools projects


mrward:

Therzok (approved):

sandyarmstrong (approved):

PR #8866, opened 2019-10-08
Fixes VSTS Bug 810293: Accessibility: Version Control - Changes:

mkrueger:
"Revert the changes" buttons in the central section of "Changes" is
not accessible with keyboard and with Voice Over turned On.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/810293


Therzok:
I have no idea what actually changed. @mkrueger can you please split the changes into two commits? One which moves the code and one which actually does the changes?

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/EditorCompareWidgetBase.cs
```diff
@@ -979,17 +990,151 @@ protected override bool OnExposeEvent (EventExpose evnt)
...
+					Accessible.FrameInGtkParent = new Rectangle (x, y, w, h);
+
+					var cocoaY = widget.Allocation.Height - y - h;
+
+					Accessible.FrameInParent = new Rectangle (x, cocoaY, w, h);
+				}
+
+				void PerformPress (object sender, EventArgs e)
+				{
+					widget.PerformRevert (hunk);
```
sevoku: if the revert succeeded the button should be removed, or at least hidden from a11y, otherwise it'll be still there and can be activated several times.

mkrueger: /fixed

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/EditorCompareWidgetBase.MiddleArea.cs
```diff
@@ -0,0 +1,457 @@
...
+					return true;
+				}
+				return base.OnButtonPressEvent (evnt);
+			}
+
+			void PerformRevert (Hunk hunk)
+			{
+				try {
+					widget.UndoChange (fromEditor, toEditor, hunk);
+					UpdateAccessiblity ();
```
sevoku: This is not enough, unfortunately, `UpdateAccessiblity ()` will recreate the a11y objects however the accessibility focus doesn't get updated and VO gets confused (it is still focused on the removed element and hence can't find any other in the chain, since it's not valid anymore). We could get the next element based on the removed object and pass it to `UpdateAccessiblity (focusElement)` and it could select that after resetting everything (using `Accessible.AccessibilityFocused`). If no hunk is left, focus the parent group.

mkrueger: The update doesn't recreate objects in all cases - I fixed that issue and hopefully saved some memory as well.

However I can't find a way to update the VO. Seems that there as api broken at least it's missing some things.

To access the Focused property you've to do:

((RealAccessibilityElementButtonProxy)obj.Accessible.Proxy).AccessibilityFocused

The proxy itself doesn't seem to have anything - but in any case the property alone doesn't do the trick. And if no buttons left there is no AccessibilityFocused to set.

I'll get back to that issue after contacting Iain.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/EditorCompareWidgetBase.cs
```diff
@@ -753,11 +755,13 @@ public MiddleArea (EditorCompareWidgetBase widget, MonoTextEditor fromEditor, Mo
 				this.toEditor = toEditor;
 				this.useLeft = useLeft;
 				this.toEditor.EditorOptionsChanged += HandleToEditorhandleEditorOptionsChanged;
+				Accessible.SetRole (AtkCocoa.Roles.AXGroup, GettextCatalog.GetString ("Revert changes margin"));
```
iainx: the text should be set as the label, not the role description.
`Accessible.SetRole (AtkCocoa.Roles.AXGroup);
Accessible.SetLabel (GettextCatalog.GetString ("Revert Changes Margin"));`

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/EditorCompareWidgetBase.cs
```diff
@@ -987,17 +997,151 @@ protected override bool OnExposeEvent (EventExpose evnt)
...
+				public ButtonAccessible (MiddleArea widget, Hunk hunk)
+				{
+					this.widget = widget;
+					this.hunk = hunk;
+
+					Accessible = AccessibilityElementProxy.ButtonElementProxy ();
+					Accessible.GtkParent = widget;
+					Accessible.PerformPress += PerformPress;
+
+					Accessible.SetRole (AtkCocoa.Roles.AXButton, widget.useLeft ? GettextCatalog.GetString ("Revert left changes button") : GettextCatalog.GetString ("Revert right changes button"));
```
iainx: Again, the text description should be used in `Accessible.SetLabel ()`
There are very few cases where SetRole should be used with a second parameter

iainx: This is an example of how Xcode sets properties for the fold buttons

Screenshot 2019-10-23 at 9 47 40 am

iainx:
The only other minor problem is that the focus goes back to the start when the buttons disappear, but I think that's a deeper level than this work which is something I need to look into, and would be a usability issue rather than an accessibility one, so I don't think that should hold up this patch

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/EditorCompareWidgetBase.cs
mkrueger: done.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Views/EditorCompareWidgetBase.cs
mkrueger: done.

sevoku:

PR #8869, opened 2019-10-09
Add support to Pack .net std lib projects

nosami:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/950073

### DIFF
--- a/main/src/addins/MonoDevelop.AspNetCore/Properties/MonoDevelop.AspNetCore.addin.xml
+++ b/main/src/addins/MonoDevelop.AspNetCore/Properties/MonoDevelop.AspNetCore.addin.xml
@@ -884,7 +884,7 @@
 	</Extension>
 
 	<Extension path = "/MonoDevelop/Ide/MainMenu/Build">
-		<CommandItem id = "MonoDevelop.AspNetCore.Commands.PublishToFolder" />
+		<CommandItem id = "MonoDevelop.AspNetCore.Commands.PublishToFolder" insertbefore="MonoDevelop.DotNetCore.Commands.Pack" />
 		<SeparatorItem id = "ProfilesSeparator" />
 		<CommandItem id = "MonoDevelop.AspNetCore.Commands.PublishToFolderProfiles" />
 	</Extension>
--- a/main/src/addins/MonoDevelop.AspNetCore/Properties/MonoDevelop.AspNetCore.addin.xml
+++ b/main/src/addins/MonoDevelop.AspNetCore/Properties/MonoDevelop.AspNetCore.addin.xml
@@ -884,7 +884,7 @@
 	</Extension>
 
 	<Extension path = "/MonoDevelop/Ide/MainMenu/Build">
-		<CommandItem id = "MonoDevelop.AspNetCore.Commands.PublishToFolder" insertbefore="MonoDevelop.DotNetCore.Commands.Pack" />
+		<CommandItem id = "MonoDevelop.AspNetCore.Commands.PublishToFolder" insertbefore="MonoDevelop.DotNetCore.Commands.PackSeparator" />
 		<SeparatorItem id = "ProfilesSeparator" />
 		<CommandItem id = "MonoDevelop.AspNetCore.Commands.PublishToFolderProfiles" />
 	</Extension>
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
@@ -13,14 +13,17 @@ namespace MonoDevelop.DotNetCore.Commands
 	{
 		protected override async void Run (object dataItem)
 		{
-			var project = IdeApp.ProjectOperations.CurrentSelectedProject as DotNetProject;
-			if (project == null)
-				return;
+			try {
+				var project = IdeApp.ProjectOperations.CurrentSelectedProject as DotNetProject;
+				if (project == null)
+					return;
 
-			var buildTarget = new PackProjectBuildTarget (project);
-			IdeApp.ProjectOperations.Build (buildTarget);
+				var buildTarget = new PackProjectBuildTarget (project);
+				IdeApp.ProjectOperations.Build (buildTarget);
+			} catch (Exception e) {
+				LoggingService.LogInternalError ("Error running Pack", e);
+			}
 		}
-
 
 		protected override void Update (CommandInfo info)
 		{
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
@@ -48,14 +48,9 @@ namespace MonoDevelop.DotNetCore.Commands
 
 		public async Task<BuildResult> Build (ProgressMonitor monitor, ConfigurationSelector configuration, bool buildReferencedTargets = false, OperationContext operationContext = null)
 		{
-			var result = new BuildResult ();
-
-			// Build the project and any dependencies first.
-			if (buildReferencedTargets && project.GetReferencedItems (configuration).Any ()) {
-				result = await project.Build (monitor, configuration, buildReferencedTargets, operationContext);
-				if (result.Failed)
-					return result;
-			}
+			var result = await project.Build (monitor, configuration, buildReferencedTargets, operationContext);
+			if (result.Failed)
+				return result;
 
 			// Run the "Pack" target on the project
 			var packResult = (await project.RunTarget (monitor, "Pack", configuration, new TargetEvaluationContext (operationContext))).BuildResult;
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
@@ -13,18 +13,18 @@ namespace MonoDevelop.DotNetCore.Commands
 	{
 		protected override void Run (object dataItem)
 		{
-            var project = IdeApp.ProjectOperations.CurrentSelectedProject as DotNetProject;
-            if (project == null)
-                return;
+			var project = IdeApp.ProjectOperations.CurrentSelectedProject as DotNetProject;
+			if (project == null)
+				return;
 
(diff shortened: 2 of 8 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
```diff
@@ -0,0 +1,105 @@
...
+		public string Name {
+			get { return project.Name; }
+		}
+
+		public async Task<BuildResult> Build (ProgressMonitor monitor, ConfigurationSelector configuration, bool buildReferencedTargets = false, OperationContext operationContext = null)
+		{
+			var result = new BuildResult ();
+
+			// Build the project and any dependencies first.
+			if (buildReferencedTargets && project.GetReferencedItems (configuration).Any ()) {
```
mrward: If buildReferencedTargets is false should we be building the project anyway?

Just wondering we should just always call project.Build here. If it was already built then it would be a no-op.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
```diff
@@ -0,0 +1,105 @@
...
+using MonoDevelop.Components.Commands;
+using MonoDevelop.Core;
+using MonoDevelop.Ide;
+using MonoDevelop.Projects;
+
+namespace MonoDevelop.DotNetCore.Commands
+{
+	class PackCommandHandler : CommandHandler
+	{
+		protected override async void Run (object dataItem)
```
Therzok: there's no actual await being used here, so not sure why we need async.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
```diff
@@ -0,0 +1,105 @@
...
+		{
+			try {
+				var project = IdeApp.ProjectOperations.CurrentSelectedProject as DotNetProject;
+				if (project == null)
+					return;
+
+				var buildTarget = new PackProjectBuildTarget (project);
+				IdeApp.ProjectOperations.Build (buildTarget);
```
Therzok: unless we want to await this to finish.

main/src/addins/MonoDevelop.AspNetCore/Properties/MonoDevelop.AspNetCore.addin.xml
```diff
@@ -715,7 +715,7 @@
 	</Extension>
 
 	<Extension path = "/MonoDevelop/Ide/MainMenu/Build">
-		<CommandItem id = "MonoDevelop.AspNetCore.Commands.PublishToFolder" />
+		<CommandItem id = "MonoDevelop.AspNetCore.Commands.PublishToFolder" insertbefore="MonoDevelop.DotNetCore.Commands.Pack" />
```
rodrmoya: this should be set to the "PublishToFolderProfiles" item below, although, there is a "PublishToAzureProfiles" placeholder in the Azure addin, so with this right now, it will show:

```
Publish to Folder
Pack
----
Folder profile 1
Folder profile 2
Azure profile 1
```

So, I guess we need a "PackSeparator" item in DotNetCore, and then have all these have "insertBefore" that?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
```diff
@@ -0,0 +1,105 @@
...
+			} catch (Exception e) {
+				LoggingService.LogInternalError ("Error running Pack", e);
+			}
+		}
+
+		protected override void Update (CommandInfo info)
+		{
+			var project = IdeApp.ProjectOperations.CurrentSelectedProject as DotNetProject;
+			info.Enabled = project != null && IsDotNetCoreProject(project);
```
rodrmoya: Shouldn't this check if the project is a library? or does the command apply to all kind of .NET Core projects?

rodrmoya: Also, it should be hidden if it's not a .NET Core project?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
nosami: It was needed at some point, I changed this code a lot :) Will remove it

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
nosami: > unless we want to await this to finish.

I don't think we need to

main/src/addins/MonoDevelop.AspNetCore/Properties/MonoDevelop.AspNetCore.addin.xml
nosami: Yep... haven't actually seen those folder profile menus yet.. will fix

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
mrward: I think you can pack all SDK style projects. You can pack a .NET Standard library. Worth checking what VS Win supports here.

nosami: > Shouldn't this check if the project is a library? or does the command apply to all kind of .NET Core projects?

As far as I know, the command can be applied to any .NET Core project. Could be wrong though, I don't see much information about it. 

It can be used by web developers to package up a web application ready for deployment for example.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
```diff
@@ -0,0 +1,101 @@
...
+			}
+
+			// Run the "Pack" target on the project
+			var packResult = (await project.RunTarget (monitor, "Pack", configuration, new TargetEvaluationContext (operationContext))).BuildResult;
+			return result.Append (packResult);
+		}
+
+		public bool CanBuild (ConfigurationSelector configuration)
+		{
+			throw new NotImplementedException ();
```
rodrmoya: Is it ok to throw on all these methods? They're never called?

nosami: They're not called so I guess it's ok.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
nosami: > Shouldn't this check if the project is a library? or does the command apply to all kind of .NET Core projects?

Answered by @sayedihashimi - It applies to all SDK style projects

main/src/addins/MonoDevelop.AspNetCore/Properties/MonoDevelop.AspNetCore.addin.xml
nosami: I think this is working fine as it is. Probably because all the "Publish" items are part of an `ItemSet` ?

image

nosami: @rodrmoya How do I get those folder profiles to appear so that I can test?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
nosami: We only ever invoke `Build` for the `PackProjectBuildTarget`, so we don't need to worry about implementing the other methods https://github.com/mono/monodevelop/pull/8869/files#diff-358e137c08560f67367062e4f9dab3eeR21

main/src/addins/MonoDevelop.AspNetCore/Properties/MonoDevelop.AspNetCore.addin.xml
nosami: image

Looks like I need a new separator, but the menu item appears in the correct place

nosami: image

mrward:
For some reason this does not work for multi-target projects if you have not built the solution first. It looks like only the first project is being built. Then the pack fails since the other dlls are not available.

We may need to add Pack to the list of targets here which are special cased - https://github.com/mono/monodevelop/blob/c48b30e307c9652ef5569bb32db63a75562f51a5/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs#L1595 - unless someone can think of a better way of doing this. Some way of indicating the target should have special handling via the context?

Also maybe we should add a test for the Pack command? Just verify it works and generates a .nupkg at the expected location.

nosami:
> Also maybe we should add a test for the Pack command? Just verify it works and generates a .nupkg at the expected location.

Yes... I'll add this now and test that it works for multi-targeting seeing as it looks like the most likely thing that could possibly break in the future.

rodrmoya (approved):

rodrmoya:
LGTM, but yes, we need a few tests for the pack command :D

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Commands/PackCommandHandler.cs
mrward: Testing this it seems that msbuild /t:pack will build the project. Presumably Pack depends on Build. The comment suggests that the project will be built by the code in the if statement which is not always the case here if there are no referenced items. So this code makes sense to me but maybe needs a comment as to why we are not building the main project here?

nosami:
@rodrmoya added a test for multi-targeted project which is I think the worst case scenario

mrward (approved):

nosami:

PR #8870, opened 2019-10-10
[DotNetCore] Add support for upcoming .NET Core 3.1

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/997815
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/997820


rodrmoya:
Only failing in the compat check, which is fixed in another PR, so merging

PR #8879, opened 2019-10-11
[VersionControl] Check repository state before certain operations.

mkrueger:
During rebase the version control seemed to do something
with the repository that caused an invalid state. It may be the add/remove file handling in version control file system extension to cause that.

This patch prevents the extension to handle add/remove events during a git operation..


Therzok:
The build failed

sevoku:
@mkrueger ping wrt build

PR #8883, opened 2019-10-11
Don't allow Reset current property pad in case of item focused changed

netonjm:
Fixes #999288 - Native Property Pad switches context when it loses focus

### DIFF
--- a/main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport/PropertyPadVisitor.cs
+++ b/main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport/PropertyPadVisitor.cs
@@ -73,7 +73,6 @@ namespace MonoDevelop.DesignerSupport
 						return;
 					}
 				}
-				DesignerSupport.Service.ReSetPad ();
 			}
 		}
 
@@ -85,7 +84,7 @@ namespace MonoDevelop.DesignerSupport
 			if (ob == ((DefaultWorkbench)IdeApp.Workbench.RootWindow).ActiveWorkbenchWindow)
 				visitedCurrentDoc = true;
 
-			if (ob is MonoDevelop.Components.Docking.AutoHideBox) {
+			if (ob is Components.Docking.AutoHideBox) {
 				found = true;
 				return true;
 			}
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport/PropertyPadVisitor.cs
```diff
@@ -73,7 +73,6 @@ public void End ()
 						return;
 					}
 				}
-				DesignerSupport.Service.ReSetPad ();
```
sevoku: I'm not sure why it was required here, it doesn't make much sense to me. @mkrueger or @slluis do you remember (this is more than 7 years old code)? Is it safe do remove this call here, or are we missing something?

Therzok: Seems to be done every time before a property pad needs to be refreshed:
http://source.monodevelop.com/#MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport/DesignerSupportService.cs,2e5f832abcff8e71,references

netonjm: Yeah, I'm not sure if it was to cover some specific case (which I did not find) when the property panel needs to be reseted, or maybe it's an old code.

sevoku:

main/src/addins/MonoDevelop.DesignerSupport/MonoDevelop.DesignerSupport/PropertyPadVisitor.cs
sevoku: yeah, that's clear, makes sense, but not really after walking the chain and not finding anything. This behaviour can be also observed with the Gtk pad (especially if it's autohiding), so it's not a regression with proppy.

Therzok:
@sevoku @slluis this PR introduced leaks in the IDE

https://dev.azure.com/DevDiv/DevDiv/_workitems/edit/1000899

Please prioritize, as it breaks CI

PR #8892, opened 2019-10-11
Add more file extensions to msbuild mimetype

sandyarmstrong:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/999409

### DIFF
--- a/main/src/core/MonoDevelop.Ide/ExtensionModel/MimeTypes.addin.xml
+++ b/main/src/core/MonoDevelop.Ide/ExtensionModel/MimeTypes.addin.xml
@@ -93,8 +93,13 @@
 		<File pattern="*.fsx" />
 		<File pattern="*.fsi" />
 	</MimeType>
-	<MimeType id="application/x-msbuild" _description="MSBuild targets file" baseType="application/xml">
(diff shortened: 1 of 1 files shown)


sandyarmstrong:

mhutch:
FWIW the MSBuild editor extension redefines the mimetype so this is redundant for folks who use it. And I would suggest that anyone who wants to edit MSBuild files should have it installed.

mhutch:
FYI https://github.com/mhutch/MonoDevelop.MSBuildEditor/blob/master/MonoDevelop.MSBuildEditor/Properties/Manifest.addin.xml#L16

decriptor:
FTW, Not sure of the current state, but I've had to disable or uninstall the MSBuildEditor extension because it was causing VSmac to hardcrash. I think there was probably some assembly ref mismatch.

knocte:
> I would suggest that anyone who wants to edit MSBuild files should have it installed.

Why not having this extention installed by default in VS4Mac. I would think at some point all .NET devs are going to edit msbuild files.

sandyarmstrong:
@knocte yes, that would be ideal. There are some blockers that need to be fixed first (like proper textmate bundle support in new editor).

PR #8897, opened 2019-10-11
[release-8.4][Core] Continue on errors when resolving assembly references

mrward:
Backport of #8896

If a PackageReference could not be resolved then no assembly
references would be given to the type system from the DotNetProject's
GetReferencedAssemblies. Whilst the NuGet restore may fail the
project.assets.json file is created and reference information is
available. The ResolvePackageAssets target was failing which then
stopped the ResolveAssemblyReferencesDesignTime target from being run.
Visual Studio on Windows avoids this problem by setting the
ContinueOnError property to ErrorAndContinue. This allows all the
targets to run even if some of them fail. It also allows references
to be returned and provided to the type system service for the
PackageReferences that can be determined.

Also set ContinueOnError for:
 - Resolving package dependencies.
 - Resolving framework references

Could not get a test working with the CoreCompileDependsOn run target - so did not add ContinueOnError there.

Fixes VSTS #998324 - NuGet Package Restore causes BCLs to not be
defined/imported


rodrmoya (approved):

iantoalms (approved):

Therzok (approved):

mrward:

PR #8906, opened 2019-10-14
[Git] Azure DevOps Git Clone Hangs

mkrueger:
SSH credentials are prefered now over username/password.
Fixed deadlock when getting credentials from extensions.

Fixes VSTS #993869

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitCredentials.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitCredentials.cs
@@ -112,30 +112,13 @@ namespace MonoDevelop.VersionControl.Git
 		{
 			bool result = false;
 			Uri uri = null;
-
 			GitCredentialsState state;
 			if (!credState.TryGetValue (type, out state))
 				credState [type] = state = new GitCredentialsState ();
 			state.UrlUsed = url;
+			Credentials cred = null;
 
-			// We always need to run the TryGet* methods as we need the passphraseItem/passwordItem populated even
-			// if the password store contains an invalid password/no password
-			if ((types & SupportedCredentialTypes.UsernamePassword) != 0) {
-				if (Uri.TryCreate (url, UriKind.RelativeOrAbsolute, out uri)) {
-					if (!state.NativePasswordUsed && TryGetUsernamePassword (uri, out var username, out var password)) {
-						state.NativePasswordUsed = true;
-						return new UsernamePasswordCredentials {
-							Username = username,
-							Password = password
-						};
-					}
-				}
-			}
-
-			Credentials cred;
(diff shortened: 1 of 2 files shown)


main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlTask.cs
```diff
@@ -88,8 +88,14 @@ void ReportError (string message, Exception exception)
 		{
 			string msg = GettextCatalog.GetString ("Version control operation failed");
 			tracker.ReportError ($"{msg}: {message}", exception);
-			if (IdeApp.Workbench.RootWindow?.Visible == false)
-				MessageService.ShowError (msg, message, exception);
+			if (IdeApp.Workbench.RootWindow?.Visible == false) {
+				// this ensures that the progress window has closed and that the
+				// message service is picking up the correct native greeter window as parent.
+				GLib.Timeout.Add (10, delegate {
```
sevoku: This smells a bit. The root cause is that first we show the alert (will parent to the MessageDialogProgressMonitor window) and then dispose the monitor (will close MessageDialogProgressMonitor). By closing the progress window we force the alert into an invalid state. Can't we just do the RootWindow check in https://github.com/mono/monodevelop/blob/02c66e32b8e0f5886628b0c157347590cc1ac5cd/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl/VersionControlTask.cs#L54 and then show the alert after calling `Wakeup()` (would need a tracker disposed check in ReportError)

mkrueger: outdated.

sevoku:

sevoku:
Also fixed the deadlock with getting the PAD from Azure DevOps.

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitCredentials.cs
```diff
@@ -237,7 +238,7 @@ static GitCredentialsProviderResult GetCredentialsFromProvider (IGitCredentialsP
 			if (type != SupportedCredentialTypes.UsernamePassword)
 				return GitCredentialsProviderResult.NotFound;
 
-			var (result, credentials) = gitCredentialsProvider.TryGetCredentialsAsync (uri).Result;
+			var (result, credentials) = Task.Run (() => gitCredentialsProvider.TryGetCredentialsAsync (uri)).Result;
```
Therzok: Does the keychain API work on a non-UI thread?

sevoku: yes, I verified that, even the password prompt popped up to allow access to the keychain, I presume that runs out of proc.

PR #8910, opened 2019-10-14
Bump libgit2 to 0.28.3

sevoku:
Another try to use latest stable libgit2 0.28.3, since we run all git operations on a dedicated thread now. Let's see how this behaves wrt memory corruption, because this should solve a bunch of other git related issues.


sevoku:
libgit2 crashes, closing.

PR #8911, opened 2019-10-14
[IDE] Add period to the template description of Shared Project

ketttY:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/984256/


ketttY:
Screen Shot 2019-10-14 at 2 29 48 PM
For this one, should I change it into "Creates a blank workspace." to make it consistent?

mrward:
`Creates a blank workspace.` sounds find to me 👍

main/src/core/MonoDevelop.Ide/templates/BlankCombine.xpt.xml
```diff
@@ -10,7 +10,7 @@
 		<Icon>md-solution</Icon>
 		<Image id="md-solution" />
 		<LanguageName/>
-		<_Description>A blank solution</_Description>
+		<_Description>Creates a blank solution</_Description>
 	</TemplateConfiguration>
```
mrward: Looks like we are missing a dot at the end here.

ketttY: Thanks lol, I feel like I lost track of the periods after staring at them for too long.

Therzok: I guess we could crawl all addin.xml files with a regex. Something in words is: line ending with <Description> or <_Description> and isn't preceded by a period

Therzok: Using grep on all .xml files with this regex: `[^.]<\/_?Description>$` will highlight which don't end with a period. Granted,  it might yield false positives, but we know where to look now!

Therzok: This is what I ran on the monodevelop directory:
`grep --extended-regexp --recursive --line-number --include *.xml '[^.]<\/_?Description>' .`

https://gist.github.com/Therzok/2777484d42c21db3547b21a8610006c6

Therzok: If we want to limit it to the templates, you can probably run the command twice, once for *.xft.xml, once for *.xpt.xml

Therzok: It can also be modified to check overrideDescription etc.

sgmunn:
the tests passed in the build, but for some reason the GH status did not get updated

sgmunn:

PR #8914, opened 2019-10-14
Fixes pin tab separation on smaller tabs

netonjm:
Fixes VSTS #984216 - [Tabs] Pin button is really close from Close button when there are many tabs in the TabStrip

### Before

image

### After

image

image


sgmunn:
Works much better for me now.

Screen Shot 2019-10-15 at 2 11 47 PM

PR #8917, opened 2019-10-14
[Debugger] Added code completion in the Watch pad

jstedfast:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/999418/


jstedfast:

mrward:
Not sure how much code completion should work here. Simple variable names are supported in the Watch window but dot completion for object properties does not seem to work.

Tested the 8.4 build of this since this was not built earlier.

```
Build Information
Release ID: 804001927
Git revision: 8e7f00e5de5e93bde439cc2d0b5489bd6c5b242a
Build date: 2019-10-14 18:58:52+00
Build branch: backport-pr-8917-to-release-8.4
Xamarin extensions: 05ab301120ffee4057e2506e996875f005ace939
```

jstedfast:
Well, I'm sure idealistically we'd want full "intellisense" support, but I'm afraid that it's not a realistic goal for 8.4. It will take a *massive* amount of work to support true code-completion for this.

sandyarmstrong:
I agree that this is good for 8.4. I would like to test it first before approving, but what I've seen so far looks reasonable. I think embedding the editor and adding a proper completion source is a better path moving forward, and should get us behavior closer to Windows, but I agree it is not the highest priority for 8.4.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacDebuggerObjectNameView.cs
```diff
@@ -201,17 +337,38 @@ protected override void UpdateContents ()
...
 
-			OptimalWidth += TextField.Frame.Width;
+			if (editable) {
+				codeCompletionTextField.PlaceholderAttributedString = GetAttributedPlaceholderString (placeholder);
+				textField = codeCompletionTextField;
 
-			constraints.Add (TextField.CenterYAnchor.ConstraintEqualToAnchor (CenterYAnchor));
-			constraints.Add (TextField.LeadingAnchor.ConstraintEqualToAnchor (ImageView.TrailingAnchor, RowCellSpacing));
+				if (!codeCompletionTextFieldVisible) {
+					TextField.RemoveFromSuperview ();
```
sandyarmstrong: oh, one question: why do we need two separate text fields? Might it be easier to always use the `codeCompletionTextField` but put it in a read-only mode when not `editable`?

jstedfast: Unfortunately, NSComboBox has that ugly down-arrow dingus so I decided to use a normal NSTextField for rendering strings that aren't editable.

sandyarmstrong: My bad, I thought that could be removed with a simple property, but looks like that isn't he case. :)

Therzok:
@monojenkins rebase

sgmunn:
@jstedfast we can close this one now right

PR #8923, opened 2019-10-15
[Catalina] Fixes broken accessory view in any NSSavePanel based panels (Cocoa bug)

netonjm:
This problem happens only in Catalina, in a first try I thought it was something about bindings or new behaviour but investigating I tested in a real xcode swift application and I can reproduce the issue. 

The problem only happens when we set the parent in any NSSavePanel. This breaks the accessory view (no matter the complexity). I think we'll need wait a fix in cocoa. 

switft code: https://gist.github.com/netonjm/c2f526e6df8143281448ccf461a64221

To try speed up this I also opened an issue in Apple Feedback : 

https://feedbackassistant.apple.com/feedback/7380159

That's how it feels now.

Open Dialog

Screenshot 2019-10-15 at 04 35 12

image

image

Fixes VSTS #1000498 - Open/Save File dialog accessory views not visible on Catalina


sevoku:

PR #8943, opened 2019-10-16
Fixes VSTS Bug 1002571: Regression: VSC checkout message dialogs shown below progress dialog window.

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1002571

+ Added missing API for AskQuestion - only method which didn't take an optional parent window.

### DIFF
--- a/main/src/addins/MacPlatform/Dialogs/MacAlertDialogHandler.cs
+++ b/main/src/addins/MacPlatform/Dialogs/MacAlertDialogHandler.cs
@@ -240,12 +240,17 @@ namespace MonoDevelop.MacIntegration
 							NSApplication.SharedApplication.StopModal ();
 						});
 
-						// pass parent and not alert so that the Runloop will change
-						// and processing will stop until the sheet is closed.
-						// If we pass alert, then it will run until a second alert is created
-						// which will be shown as a dialog and then the runloop changes and
-						// processing stops
-						NSApplication.SharedApplication.RunModalForWindow (parent);
+						if (IdeApp.Workbench.RootWindow?.Visible == true) {
+							// pass parent and not alert so that the Runloop will change
+							// and processing will stop until the sheet is closed.
+							// If we pass alert, then it will run until a second alert is created
+							// which will be shown as a dialog and then the runloop changes and
+							// processing stops
+							NSApplication.SharedApplication.RunModalForWindow (parent);
+						} else {
+							// fixes message dialog below other dialogs issue when no main window is present (at start)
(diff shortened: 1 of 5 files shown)


main/src/addins/MacPlatform/Dialogs/MacAlertDialogHandler.cs
```diff
@@ -240,12 +240,17 @@ public bool Run (AlertDialogData data)
 							NSApplication.SharedApplication.StopModal ();
 						});
 
-						// pass parent and not alert so that the Runloop will change
-						// and processing will stop until the sheet is closed.
-						// If we pass alert, then it will run until a second alert is created
-						// which will be shown as a dialog and then the runloop changes and
-						// processing stops
-						NSApplication.SharedApplication.RunModalForWindow (parent);
+						if (IdeApp.Workbench.RootWindow?.Visible == true) {
```
Therzok: deduplicate the method call.

```csharp
Window modal = IdeApp.Workbench.RootWindow?.Visible == true ? parent : alert.Window;
NSApplication.SharedApplication.RunModalForWindow (modal);
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.ProgressMonitoring/MessageDialogProgressMonitor.cs
```diff
@@ -39,7 +39,9 @@ public class MessageDialogProgressMonitor: ProgressMonitor
 		ProgressDialog dialog;
 		bool hideWhenDone;
 		bool showDetails;
-		
+
+		public ProgressDialog ProgressDialog { get => dialog; }
```
Therzok: Could be simpler:

```suggestion
		public ProgressDialog ProgressDialog => dialog;
```

main/src/addins/MacPlatform/Dialogs/MacAlertDialogHandler.cs
mkrueger: No, I want to state that this is a work around for something. That would be shadowed by that.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.ProgressMonitoring/MessageDialogProgressMonitor.cs
mkrueger: Is that only a getter ?

sevoku:

PR #8944, opened 2019-10-16
[Ide] Hide .NET Runtimes selection behind feature switch

mrward:
The .NET Runtimes page in Preferences - Projects is now not displayed
unless the RUNTIME_SELECTOR feature switch is enabled.

Other Mono runtimes, apart from the current runtime being used to run
the IDE, are ignored if the RUNTIME_SELECTOR feature switch is not set.
This prevents older Mono versions being run to check their version in
the MonoRuntimeInfo class which can trigger the mono-sgen32 is not
optimized dialog to be displayed on the Mac for Mono versions that do
not have 64 bit support.

If another runtime was configured to be used instead of the default
and the RUNTIME_SELECTOR is not set it will not be used since it will
not be returned from the MonoTargetRuntimeFactory. Instead the runtime
used by the IDE will be used.

Fixes VSTS #997928 - Disable support for multiple Mono runtimes
behind feature switch


mrward:

Therzok (approved):

sgmunn (approved):

PR #8948, opened 2019-10-16
Clear current cached allocations on BlankPad

netonjm:


Therzok:
Included as part of https://github.com/mono/monodevelop/pull/8955/files

PR #8966, opened 2019-10-16
[DotNetCore] Fix focus in SDK locations options page

mrward:
In Preferences - Projects - SDK Locations - .NET Core, opening
the open file dialog and then closing it would end up with the
Preferences dialog not having focus. This only happens when the IDE
window is open. When the Start Window is open then the focus goes
back to the Preferences dialog. To reproduce this tab to the browse
button, press space to open the open file dialog, then press
escape. The focus should be on the browse button, but the dialog
does not have the focus so tabbing does not move around the dialog.

To fix this the dialog is explicitly focussed after the select file
dialog is closed. Note that the SDK locations dialog uses the XWT
OpenFileDialog which does not set the focus back to its parent. The
MonoDevelop.Components.SelectFileDialog will set focus back to its
TransientFor using the DesktopService.FocusWindow method. The .NET
Core SDK locations page now does the same.

Fixes VSTS #752768 - Accessibility: After closing the Choose Location
window focus is not there on Choose Location button.


mrward:

mrward:

PR #8973, opened 2019-10-17
[NuGet] Fix dialog loss of focus on adding new package source

mrward:
In Preferences - NuGet - Sources, the Add Package Source dialog
would lose focus when the Browse button was used. A way to reproduce
this is to tab to the Add button, press space, then tab to the
Browse button in the Add Package Source dialog, press space, then
press escape. The Add Package Source dialog then no longer has focus
and tabbing does not work.

Fixed this by explicitly setting focus back to the parent dialog
when a child dialog is closed.

Fixes VSTS #1003430 - Accessibility: NuGet add package source dialog
browse button loses dialog focus


mrward:

mrward:

mrward:

sevoku:
This should not be required, is that still an issue with latest AtkCocoa at all? https://github.com/xamarin/AtkCocoa/commits/master

PR #8980, opened 2019-10-18
Fixes VSTS Bug 1003485: [Feedback] After Commit and Push changes

mkrueger:
remains

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1003485

The commit dialog always reponses with ok when commit is clicked.
ButtonCommitClicked handles that dialog but doesn't change the
response.
So the best solution is to let the commit dialog handle the commit
logic itself.

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/CommitDialog.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/CommitDialog.cs
@@ -44,6 +44,7 @@ namespace MonoDevelop.VersionControl.Dialogs
 {
 	partial class CommitDialog : Gtk.Dialog
 	{
+		Repository vc;
 		ListStore store;
 		HashSet<FilePath> selected = new HashSet<FilePath> ();
 		List<CommitDialogExtension> extensions = new List<CommitDialogExtension> ();
@@ -51,8 +52,9 @@ namespace MonoDevelop.VersionControl.Dialogs
 		string oldMessage;
 		bool responseSensitive;
 
-		public CommitDialog (ChangeSet changeSet)
+		public CommitDialog (Repository vc, ChangeSet changeSet)
 		{
+			this.vc = vc;
 			Build ();
 
 			store = new ListStore(typeof (Xwt.Drawing.Image), typeof (string), typeof (string), typeof(bool), typeof(object));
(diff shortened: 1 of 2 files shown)


main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/CommitDialog.cs
```diff
@@ -44,15 +44,17 @@ namespace MonoDevelop.VersionControl.Dialogs
...
+		Repository vc;
 		ListStore store;
 		HashSet<FilePath> selected = new HashSet<FilePath> ();
 		List<CommitDialogExtension> extensions = new List<CommitDialogExtension> ();
 		ChangeSet changeSet;
 		string oldMessage;
 		bool responseSensitive;
 
-		public CommitDialog (ChangeSet changeSet)
+		public CommitDialog (Repository vc, ChangeSet changeSet)
```
Therzok: ChangeSet already has the repository.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/CommitDialog.cs
```diff
@@ -342,5 +352,60 @@ void OnCommitToggledHandler(object o, ToggledArgs args)
...
 		}
+
+		private class CommitWorker : VersionControlTask
+		{
+			Repository vc;
+			ChangeSet changeSet;
+			CommitDialog dlg;
+			bool success;
+
+			public CommitWorker (Repository vc, ChangeSet changeSet, CommitDialog dlg)
```
Therzok: Duplicate information. ChangeSet already has the repository. CommitDialog already has the ChangeSet.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/CommitDialog.cs
mkrueger: Maybe but I don't want to change that since the commit API splits that.

main/src/addins/VersionControl/MonoDevelop.VersionControl/MonoDevelop.VersionControl.Dialogs/CommitDialog.cs
mkrueger: Maybe but I don't want to change that since the commit API splits that.

mkrueger:
Should be done in another commit. The repository/changeset split is something that's in the whole commit/status view API. Just removing it at that point would lead to an unused parameter and atm I can't predict the side effects changing it at that point.

The fix is a valid refactoring that doesn't alter the behavior beside of the bug. Everything else is not in scope of that PR. But I agree with Marius here that it looks redundant to have a repository and changeset object as parameter in the whole API (not only at that point).

PR #8991, opened 2019-10-18
[Ide] Fix incorrect language read by voice over in New Project dialog

mrward:
If F# was selected for one template in the New Project dialog but
the other templates did not support F# then moving to the other
templates F Sharp would be announced by Voice Over. The problem was
that the wrong language was being set for the Text associated with
the language selector.

Fixes VSTS #752317 - Accessibility: In Project description Voice over
is reading  letter C as F (C#(C number) as F#(F number))
Fixes VSTS #751297 - Accessibility: Voice over is not reading the  whole project details in recently used templates page

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkTemplateCellRenderer.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkTemplateCellRenderer.cs
@@ -73,11 +73,11 @@ namespace MonoDevelop.Ide.Projects
 
 		void SetAccessibilityText ()
 		{
-			if (template != null) {
-				var text = template.Name;
-				if (!string.IsNullOrEmpty (templateCategory))
-					text += ", " + templateCategory.Replace ("→", "–"); // we don't want narrators to read "right arrow"
-				Text = text;
+			Text = template?.Name ?? string.Empty;
+			if (!string.IsNullOrEmpty (templateCategory)) {
+				if (!string.IsNullOrEmpty (Text))
+					Text += ", ";
+				Text += templateCategory.Replace ("→", "–"); // we don't want narrators to read "right arrow"
 			}
 		}
 
@@ -133,7 +133,7 @@ namespace MonoDevelop.Ide.Projects
 				int textPixelWidth = widget.Allocation.Width - ((int)Xpad * 2);
 				layout.Width = (int)(textPixelWidth * Pango.Scale.PangoScale);
 
-				layout.SetMarkup (TemplateCategory);
+				layout.SetMarkup (MarkupTopLevelCategoryName (TemplateCategory));
 
 				int w, h;
 				layout.GetPixelSize (out w, out h);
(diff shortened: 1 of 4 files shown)


sevoku:
The problem is that the value is not stored in the model, maybe it'd be enough to reset the TemplateCategory value in the model to trigger an AtkCocoa update? @iainx ?

sevoku:
@mrward see for a final fix: #9002, feel free to merge and rebase (I'll be on vacation this week).

mrward:
@monojenkins rebase

mrward:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkTemplateCellRenderer.cs
```diff
@@ -40,16 +40,47 @@ class GtkTemplateCellRenderer : CellRendererText
...
+			set {
+				template = value;
+				SetAccessibilityText ();
+			}
+		}
+		public string SelectedLanguage {
+			get { return selectedLanguage; }
+			set {
+				selectedLanguage = value;
+				SetAccessibilityText ();
```
mykyta-bondarenko: why do we need this if we don't use `selectedLanguage `under `SetAccessibilityText ` method?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkTemplateCellRenderer.cs
```diff
@@ -40,16 +40,47 @@ class GtkTemplateCellRenderer : CellRendererText
 	{
 		const int iconTextPadding = 9;
 		int groupTemplateHeadingTotalYPadding = 24;
-		int recentTemplateHeadingTotalYPadding = 30;
+		int recentTemplateHeadingTotalYPadding = 30;
+		private SolutionTemplate template;
+		private string templateCategory;
+		private string selectedLanguage;
```
mykyta-bondarenko: do we need an extra `private` here?

mykyta-bondarenko:
I think we may add the global category name there ("iOS - General") 
image

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkTemplateCellRenderer.cs
sevoku: not really, probably that was added automatically

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkTemplateCellRenderer.cs
sevoku: right, this is not really necessary, but I'd keep it for consistency.

sevoku:
> I think we may add the global category name there ("iOS - General")
> image

Not necessarily, unlike in the recent templates view, here the user has already selected a category, so it'd be redundant (the user knows that it's "iOS - General" already, by selecting the category initially on the left-hand side).

sevoku:
@monojenkins rebase

sevoku:

sevoku:
The previously failing `Microsoft.VSForMac.Tests.AspDotNetTemplateTests` tests pass both locally for me with the last commit fcd3ab30e9ba40d7c4132232b7ac7b4eee4a5dc6. 1 of 2 failed here because of the known unrelated NameResolutionFailure issue: https://devdiv.visualstudio.com/DevDiv/_release?releaseId=495572&_a=release-contribution-tab-ms.vss-test-web.test-result-in-release-management

PR #8993, opened 2019-10-18
[AspNetCore] Work around certificates issue in .NET Core 3.1 Preview 1

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1002439
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1002551

### DIFF
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.DevCertInstaller/Program.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.DevCertInstaller/Program.cs
@@ -130,6 +130,15 @@ namespace MonoDevelop.AspNetCore.DevCertInstaller
 				return -3;
 			}
 
+			// Clean up certificates first, to work around an issue in .NET Core 3.1 P1
+			// with a bad certificate installation. See https://github.com/aspnet/AspNetCore/issues/15118
+			using (var process = Process.Start (dotNetCorePath, "dev-certs https --clean")) {
+				process.WaitForExit ();
+				if (process.ExitCode != 0) {
+					Console.WriteLine ($"Could not clean up current certificate, trusting might fail");
+				}
+			}
+
 			using (var process = Process.Start (dotNetCorePath, "dev-certs https --trust")) {
 				process.WaitForExit ();
 				return process.ExitCode;
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreCertificateManager.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreCertificateManager.cs
@@ -92,12 +92,12 @@ namespace MonoDevelop.AspNetCore
 				if (result == CertificateCheckResult.OK) {
 					IsDevelopmentCertificateTrusted = true;
 					return;
-				} else if (result == CertificateCheckResult.Error) {
-					// Check failed - Do not try to trust certificate since this
-					// will likely also fail.
-					return;
 				}
 
+				// If we got a failure, try trusting, as there are cases (3.1 Preview 1,
+				// see https://github.com/aspnet/AspNetCore/issues/15118) where this is just
+				// due to a bad certificate being installed in the system. Since our DevCert
+				// Installer tool will clean before trusting, just have a try.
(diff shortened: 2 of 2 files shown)


main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.DevCertInstaller/Program.cs
```diff
@@ -130,6 +130,15 @@ static int RunDotNetDevCerts (string dotNetCorePath)
...
 
+			// Clean up certificates first, to work around an issue in .NET Core 3.1 P1
+			// with a bad certificate installation. See https://github.com/aspnet/AspNetCore/issues/15118
+			using (var process = Process.Start (dotNetCorePath, "dev-certs https --clean")) {
+				process.WaitForExit ();
+				if (process.ExitCode != 0) {
+					Console.WriteLine ($"Could not clean up current certificate, trusting might fail");
+				}
+			}
+
```
mrward: With a bad certificate `dev-certs https --check` returns 0 so we would not run this. Not sure this fixes 1002439 nor 1002551. I think it needs to be fixed in .NET Core 3.1 preview 1 sdk. The bug linked to has a repro which does not involve VS Mac.

However, that being said, I am not against running a clean before trying to trust the certificates. That might help in some cases - just not sure it would fix the 3.1 preview 1 problem.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreCertificateManager.cs
```diff
@@ -92,12 +92,12 @@ public static async Task TrustDevelopmentCertificate (ProgressMonitor monitor)
 				if (result == CertificateCheckResult.OK) {
 					IsDevelopmentCertificateTrusted = true;
 					return;
-				} else if (result == CertificateCheckResult.Error) {
-					// Check failed - Do not try to trust certificate since this
-					// will likely also fail.
-					return;
 				}
```
mrward: Another idea here might to be to pass a clean parameter to the console dev-certs runner if CertificateCheckResult.Error was the exit code.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.DevCertInstaller/Program.cs
rodrmoya: Yeah, have been repro'ing the bug and seems I thought it was fixed because, I guess, ran it when the certificate was ok, so starting again :p

rodrmoya: Well, I thought I had been dreaming, but not, following the steps at https://github.com/aspnet/AspNetCore/issues/15118#issuecomment-543793272

after all those:
```
rodrigo@Rodrigos-MacBook-Pro:/tmp/webapp $ dotnet dev-certs https --check
A valid certificate was found.
rodrigo@Rodrigos-MacBook-Pro:/tmp/webapp $ echo $?
0
rodrigo@Rodrigos-MacBook-Pro:/tmp/webapp $ dotnet dev-certs https --trust --check
A valid certificate was found.
The following certificates were found, but none of them is trusted:
CN=localhost - EBFEC93EE823FBC3BF969E5753863E3CAED95AD1
rodrigo@Rodrigos-MacBook-Pro:/tmp/webapp $ echo $?
7
```
--trust --check is what we use -> https://github.com/mono/monodevelop/blob/9b6e959e662c67340854e249d9425e36c7d2d97f/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/DotNetCoreDevCertsTool.cs#L55

so the code that calls `CheckCertificate` indeed gets a non-0 error code so goes through the --clean --trust again.

rodrmoya: ok, and now it doesn't work :/

rodrmoya: Now can't repro anymore, I always get 0 as return code :(

PR #8999, opened 2019-10-20
This PR is the first step in the series that tries to simplify our solution build setup

Therzok:
It relies more on detecting which OS you're running on, rather than building for a specific configuration.

This makes most of the projects from the repository switch to using `Debug` rather than `DebugMac` or `DebugWin32`.

Trying to find a way (other than traversal projects) to actually conditionally include projects in the sln (like Xwt.XamMac).


Therzok:
There is a quality of life improvement for this PR. If you have a fresh clone of a solution, and you switch from `Debug` to `DebugMac`, we will not wastefully compute the workspace files cache once for the Debug config and once for DebugMac.

main/src/tools/mdtool/mdtool.csproj
```diff
@@ -29,6 +29,10 @@
       <Name>Mono.Addins.Setup</Name>
       <Private>False</Private>
     </ProjectReference>
+    <Reference Include="Xamarin.Mac" Condition="$(IsMac)">
```
Therzok: With this, calling Terminate actually is not a NOP anymore in mdtool.

Therzok: So, funnily, this surfaced an actual product bug! 💃

jtorres (approved):

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.MSBuildBuilder.csproj
```diff
@@ -9,15 +9,10 @@
...
-    <PlatformTarget>x64</PlatformTarget>
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'DebugWin32|x86' " />
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
-    <PlatformTarget>x64</PlatformTarget>
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'ReleaseWin32|x86' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' "/>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' "/>
   <!-- TODO: Linux -->
```
mrward: Do we need to define `<PlatformTarget>x64</PlatformTarget>` here? Or is x64 the default? Not sure it makes a different here anyway since even if x86 is used mono will run it with mono64 on the Mac.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.MSBuildResolver/MonoDevelop.MSBuildResolver.csproj
```diff
@@ -16,15 +16,10 @@
...
     <OutputPath>..\..\..\..\build\bin\MSBuild</OutputPath>
     <DefineConstants>PUBLIC_API</DefineConstants>
+    <PlatformTarget Condition="$([MSBuild]::IsOsPlatform('Windows'))">x86</PlatformTarget>
   </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
-    <PlatformTarget>x64</PlatformTarget>
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'DebugWin32|AnyCPU' " />
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
-    <PlatformTarget>x64</PlatformTarget>
```
mrward: Same question here about x64 platform target since it is not set now apart from on Windows.

mrward (commented):
Changes look fine to me. Just wondering about the x64 platform target, whether that is still needed.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.MSBuildBuilder.csproj
Therzok: `AnyCPU` means use `mono`. Since mono defaults to 64bit on Mac, not sure we need to modify it to be x64. If we want that, then we shouldn't build `AnyCPU` and set `x64` as the target.

I'm okay with both forcing it to 64bit and leaving it as it is.

main/src/core/MonoDevelop.Projects.Formats.MSBuild/MonoDevelop.MSBuildResolver/MonoDevelop.MSBuildResolver.csproj
Therzok: Yeah, we need x86 on Windows since the app is x86. Otherwise, AnyCPU == 64bit on mac.

mrward (approved):

Therzok:
That's a long way away. This just simplifies the available configuration matrix, to only have `Debug` and `Release` for monodevelop projects. I did this in my free time, I'll see when I can do the rest :)

PR #9003, opened 2019-10-21
[Ide] Fixed possible null reference exception.

mkrueger:
See: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/810343

https://github.com/mono/monodevelop/pull/8925

Didn't know that Accessibility can be == null.

Fixes VSTS #1005254 - File content isn't displayed after opening

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Tabstrip.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Tabstrip.cs
@@ -489,7 +489,7 @@ namespace MonoDevelop.Components
 
 		void UpdateAccessibility ()
 		{
-			Accessible.SetRole (AtkCocoa.Roles.AXRadioButton, active ? "active tab" : "tab");
+			Accessible?.SetRole (AtkCocoa.Roles.AXRadioButton, active ? "active tab" : "tab");
 		}
 
 		public bool IsSeparator {


mkrueger:
Added that change to the open PR https://github.com/mono/monodevelop/pull/8925

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Tabstrip.cs
```diff
@@ -489,7 +489,7 @@ class Tab : IDisposable
 
 		void UpdateAccessibility ()
 		{
-			Accessible.SetRole (AtkCocoa.Roles.AXRadioButton, active ? "active tab" : "tab");
+			Accessible?.SetRole (AtkCocoa.Roles.AXRadioButton, active ? "active tab" : "tab");
```
Therzok: How could accessible be null?

mkrueger: I don't know but it can - see the exception. There are more Accessible == null checks in that class.

Therzok: cc @iainx

mkrueger: In any case there seeems to be a crash on some systems so we need that.

sevoku: maybe not realized? But that doesn't make much sense...

Therzok:
Accessible being null only happen in this scenario. It shouldn't happen for other widgets.

PR #9009, opened 2019-10-21
Fixes VSTS Bug 1002619: System.NullReferenceException exception in

mkrueger:
MonoDevelop.Ide.FindInFiles.SearchResult.GetDocument()

GetDocument was not inside try…catch so an exception there could crash
the whole IDE. (async void should be avoided)

Why it crashes is hard to tell. There is a search & replace PR open that stabilizes that layer a bit.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/SearchResult.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/SearchResult.cs
@@ -85,14 +85,14 @@ namespace MonoDevelop.Ide.FindInFiles
 		internal string GetCopyData (SearchResultWidget widget)
 		{
 			if (copyData == null)
-				FillCache (widget);
+				FillCache (widget).Ignore ();
 			return copyData;
 		}
 
 		internal string GetMarkup(SearchResultWidget widget, bool isSelected)
 		{
 			if (markup == null)
-				FillCache (widget);
+				FillCache (widget).Ignore ();
 			return isSelected ? selectedMarkup : markup;
 		}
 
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/SearchResult.cs
```diff
@@ -274,8 +275,13 @@ public virtual Components.HslColor GetBackgroundMarkerColor (EditorTheme style)
 
 		TextEditor GetDocument ()
 		{
-			if (cachedEditor == null || cachedEditor.IsDisposed || cachedEditor.FileName != FileName || cachedEditorFileProvider != FileProvider) {
+			var fileProvider = FileProvider;
+			if (cachedEditor == null || cachedEditor.IsDisposed || cachedEditor.FileName != FileName || cachedEditorFileProvider != fileProvider) {
+				if (fileProvider == null)
+					throw new NullReferenceException ("FileProvider == null");
```
Therzok:
```suggestion
					throw new InvalidOperationException ("FileProvider == null");
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.FindInFiles/SearchResult.cs
```diff
@@ -274,8 +275,13 @@ public virtual Components.HslColor GetBackgroundMarkerColor (EditorTheme style)
...
 				var content = FileProvider.ReadString ();
+				if (content == null)
+					throw new NullReferenceException ("FileProvider.ReadString () == null");
```
Therzok:
```suggestion
					throw new InvalidOperationException ("FileProvider.ReadString () == null");
```

PR #9017, opened 2019-10-21
[Debugger] Don't Dispose() items until *after* we call RemoveItems()

jstedfast:
May fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1004450/

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacObjectValueTreeViewDataSource.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacObjectValueTreeViewDataSource.cs
@@ -96,14 +96,28 @@ namespace MonoDevelop.Debugger
 				Add (value, new LoadingObjectValueNode (node));
 		}
 
-		void Remove (MacObjectValueNode node)
+		void Remove (MacObjectValueNode node, List<MacObjectValueNode> removed)
 		{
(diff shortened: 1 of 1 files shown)


jstedfast:

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacObjectValueTreeViewDataSource.cs
```diff
@@ -127,41 +141,40 @@ public void Replace (ObjectValueNode node, ObjectValueNode[] replacementNodes)
 			if (index == -1)
 				return;
 
+			var removed = new List<MacObjectValueNode> ();
```
Therzok: Do we know how many we're removing?

jstedfast: We know how many top-level items we are removing, but that does not include the child nodes that we also need to dispose.

sgmunn:
this is hard to validate. I've not been able to reproduce this exact crash so I cannot tell if it helps or not. I do know that this doesn't help the other crasher that we have with collapsing nodes (in tooltips).

jstedfast:
Can this be merged?

Therzok:
Re-triggered test run, provisioning failed.

jstedfast:
Ping for reviews...

PR #9023, opened 2019-10-22
Fixes NRE focusing main toolbar in early initializations when the window is not yet attached to the view

netonjm:
Fixes NRE focusing main toolbar in early initializations when the window is not yet attached to the view

Fixes VSTS #1002143 - [FATAL] System.NullReferenceException exception in MonoDevelop.MacIntegration.MainToolbar.MainToolbar.Focus()


netonjm:

PR #9026, opened 2019-10-22
Fixes VSTS Bug 958242: System.NullReferenceException exception in

mkrueger:
MonoDevelop.SourceEditor.SourceEditorView.RunWhenRealized()

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/958242


mkrueger:

mkrueger:
This surely doesn't fix the cause. Something from code analysis is calling the document after it's been disposed.

Therzok:
Feels like we are masking a bug.

sevoku:
This is for the old editor only, so I'd rather just keep it as it is, @slluis thoughts?

sevoku:
yes, I meant not to take the fix at all

PR #9033, opened 2019-10-22
Fix 959380: TextMate tagger does not work with extensions (VS compat issue)

DavidKarlas:


DavidKarlas:
Did final test locally and worked fine...

DavidKarlas:

PR #9038, opened 2019-10-22
[AspNetCore] Avoid exception when saving run configurations

rodrmoya:
When loading a ASP.NET Core project, OnCreateRunConfiguration is called
for the "Default" config, and the ASP.NET Core implementation does the
syncing of the launchSettings.json file, in which case it enumerates
the existing RunConfigurations in the Project, which in turn calls
again OnCreateRunConfiguration to create the "Default" again.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1005086


rodrmoya:

PR #9042, opened 2019-10-22
[Debugger] Make sure tooltip request wasn't cancelled before showing

jstedfast:
After the context switch to the main thread in order to show the tooltip,
check that the request hasn't been cancelled due to the editor window
being closed in the meantime (for example).

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1002160/


jstedfast:

sgmunn:
@jstedfast I cannot get tooltips to work at all on this build

sgmunn:
oh, wait. that's right. I'm on catalina. will try mojave, but this is interesting because on Catalina with other builds I get tooltips but they are the wrong size.

jstedfast:
@sgmunn are you getting any log messages about the editor window being closed?

PR #9047, opened 2019-10-22
[DotNetCore] Fix detection of required runtime for the project being run

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/982889

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
@@ -204,21 +204,24 @@ namespace MonoDevelop.DotNetCore
 			TargetFrameworkMoniker framework,
 			SolutionItemRunConfiguration runConfiguration)
 		{
+			var requiredVersion = DotNetCoreVersion.Parse (Project.TargetFramework.Id.Version);
 			if (DotNetCoreRuntime.IsMissing) {
-				return ShowCannotExecuteDotNetCoreApplicationDialog ();
+				return ShowCannotExecuteDotNetCoreApplicationDialog (DotNetCoreRuntime.GetNotInstalledVersionMessage (requiredVersion.OriginalString), requiredVersion);
+			}
+			if (Project.TargetFramework.IsNetCoreApp () &&
+				!DotNetCoreRuntime.Versions.Any (x => x.OriginalString.StartsWith (Project.TargetFramework.Id.Version, StringComparison.OrdinalIgnoreCase))) {
+				return ShowCannotExecuteDotNetCoreApplicationDialog (DotNetCoreRuntime.GetNotInstalledVersionMessage (requiredVersion.OriginalString), requiredVersion);
 			}
 
 			return base.OnExecute (monitor, context, configuration, framework, runConfiguration);
 		}
 
-		Task ShowCannotExecuteDotNetCoreApplicationDialog ()
+		Task ShowCannotExecuteDotNetCoreApplicationDialog (string message, DotNetCoreVersion requiredVersion)
 		{
-			return Runtime.RunInMainThread (() => {
-				CreateInfoBarInstance ().Prompt ();
-			});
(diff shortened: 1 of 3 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -427,9 +427,13 @@ protected override void OnPrepareForEvaluation (MSBuildProject project)
 
 		public bool IsDotNetCoreSdkInstalled ()
```
Therzok: Can we have some tests around this logic?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -427,9 +427,13 @@ protected override void OnPrepareForEvaluation (MSBuildProject project)
...
 		{
-			if (DotNetCoreSdk.IsInstalled || MSBuildSdks.Installed)
-				return DotNetCoreSdk.IsSupported (Project.TargetFramework);
-			return false;
+			if (Project.TargetFramework.IsNetCoreApp ()) {
```
mrward: Not sure this logic is correct. DotNetCoreSdk.IsSupported checks that an sdk greater or equal to the .NET Core app used by the project is installed. That is all you need to be able to compile the project unless a global.json is used. So not sure this error would be correct in all cases. You should not need to install .NET Core SDK 2.1 to build a .NET Core 2.1 project. You would need the runtime though to run it.

The bug seems to be about a missing runtime not a missing sdk.

I think the logic should be:

1. No global.json.
 a) Ensure a more recent sdk is installed.
 b) User tries to run project - ensure the runtime is available.
2. Global.json
  a) Ensure sdk is available that supports the global.json sdk setting.
  b) User tries to run project - ensure the runtime is available.

Although for the bug itself, if we are only looking to fix that, we could just implement 1.a) which is the same as 2.b). Check in the execution handler code in the DotNetCoreProjectExtension that the runtime is available. Not sure we would need to do this on load since you may never run the project, just compile it.

rodrmoya:
image

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -204,40 +204,38 @@ protected override bool IsSupportedFramework (TargetFrameworkMoniker framework)
...
 		{
+			var requiredVersion = DotNetCoreVersion.Parse (Project.TargetFramework.Id.Version);
 			if (DotNetCoreRuntime.IsMissing) {
-				return ShowCannotExecuteDotNetCoreApplicationDialog ();
+				return ShowCannotExecuteDotNetCoreApplicationDialog (DotNetCoreRuntime.GetNotInstalledVersionMessage (requiredVersion.OriginalString), requiredVersion);
+			}
+			if (Project.TargetFramework.IsNetCoreApp () &&
+				!DotNetCoreRuntime.Versions.Any (x => x.OriginalString.StartsWith (Project.TargetFramework.Id.Version, StringComparison.OrdinalIgnoreCase))) {
+				return ShowCannotExecuteDotNetCoreApplicationDialog (DotNetCoreRuntime.GetNotInstalledVersionMessage (requiredVersion.OriginalString), requiredVersion);
 			}
```
mrward: This looks like it will be shown on opening the project. Do we want to show it when the project is run instead? I could open a solution and never run that particular project.

rodrmoya: hmm, no, it's shown when running, this is `OnExecute`, or what do you mean?

mrward: Sorry my mistake. I did not look at the entire file here. GitHub is showing IsSupported above this code so it looked like it was in that method 😄

rodrmoya:

PR #9049, opened 2019-10-22
Add example of how to crash from a pinvoke

Therzok:


main/build/MacOSX/monostub.mm
```diff
@@ -227,6 +227,12 @@
 	}
 }
 
+extern "C" void crash_me()
```
Therzok: Need to figure out how to include this only in debug builds.

Therzok: Adding do not merge until this is fixed.

main/src/addins/PerformanceDiagnostics/PerformanceDiagnostics/Commands.cs
```diff
@@ -201,5 +201,16 @@ protected override void Run ()
...
 	}
+
+	class InduceNativePInvokeCrashHandler : CommandHandler
+	{
+		[DllImport ("__Internal")]
+		static extern void debug_trigger_sigsegv ();
+
+		protected override void Run ()
+		{
+			crash_me ();
```
kdubau: Should be `debug_trigger_sigsegv ();` ?

Therzok: Yup!

Therzok:
```suggestion
			debug_trigger_sigsegv ();
```

PR #9051, opened 2019-10-23
Fixes VSTS Bug 977177: [FATAL] System.NullReferenceException exception

mkrueger:
in MonoDevelop.Components.Mac.GtkMacInterop.GetNSWindow()

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/977177


mkrueger:
I think in that code path window is always != null, but adding an argument null check is never wrong.

I guess that window.GdkWindow == null there but window.Show () is called in the code path. The centerwindow works if null is returned. Since it's a very rare exception I've no idea how to reproduce that or how it can happen.

netonjm:
Not sure if is the right fix, this problem happens when we try to get a NSWindow from a hidden Gtk.Window.

sevoku:
I agree with @netonjm the actual problem here is that something is trying to cast a not yet realized (GdkWindow==null) window to an NSWindow, but that's not supported and should not be done. If something is needing the NSWindow, it should do a realized check frist.

PR #9053, opened 2019-10-23
Fixes VSTS User Story 980492: Create unit test suite for

mkrueger:
DocumentRegistry

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/980492

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DocumentRegistry.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DocumentRegistry.cs
@@ -27,6 +27,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Threading.Tasks;
 using MonoDevelop.Core;
 
 namespace MonoDevelop.Ide.Gui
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/DocumentRegistry.cs
```diff
@@ -53,28 +54,32 @@ static DocumentRegistry ()
 			FileService.FileChanged += HandleFileServiceChange;
 		}
 
-		static void HandleFileServiceChange (object sender, FileEventArgs e)
+		static async void HandleFileServiceChange (object sender, FileEventArgs e)
```
Therzok: We should avoid async void.

Therzok: This changes the event handler to exit invoke when the first await is reached, which changes logic and does not ensure that all documents have been notified.

mkrueger: The old way doesn't ensure it either. I don't want to change the event.
The biggest reason against async void is the exception handling which I handled here.

Therzok:
@mkrueger the old way would exit only after _all_ the documents were triggered for reload. This exits as soon as the _first_ one is triggered for reload.

mkrueger:
In the old way it may never be reloaded now it does. This implementation is more correct.

What do you mean with exit after the first one triggered ? the loop isn't canceled after reload. it exists after all have really reloaded.

main/tests/Ide.Tests/MonoDevelop.Ide.Gui.Documents/DocumentRegistryTests.cs
```diff
@@ -0,0 +1,147 @@
...
+		[SetUp]
+		public async Task Setup ()
+		{
+			documentManager = await Runtime.GetService<DocumentManager> ();
+			documentControllerService = await Runtime.GetService<DocumentControllerService> ();
+			shell = (MockShell)await Runtime.GetService<IShell> ();
+		}
+
+		[Test]
+		public async Task TestVSTSBug8608 ()
```
jtorres, line 59: Could we add a description here?

mkrueger:
moved to vs for mac.

PR #9056, opened 2019-10-23
Use the right pasteboard type values for rtf/html atoms

Therzok:
NOTE: This patch by itself does not fix anything.

There are outdated pboard type values being used in gtk which may need updating.


Therzok:
cc @DavidKarlas @bratsche

Therzok:
Obsolete

PR #9057, opened 2019-10-23
[983297][NewProjectDialog] fixed cell's accessibility

mykyta-bondarenko:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/983297

- added a `Description` text to a VO text
- fixed recent item's reading text (before VO read only `iOS -> App` label). Now it does it in the same way for all cells

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
@@ -54,7 +54,7 @@ namespace MonoDevelop.Ide.Projects
 			// Set up the list store so the test framework can work out the correct columns
 			SemanticModelAttribute modelAttr = new SemanticModelAttribute ("templateCategoriesListStore__Name", "templateCategoriesListStore__Icon", "templateCategoriesListStore__Category");
 			TypeDescriptor.AddAttributes (templateCategoriesTreeStore, modelAttr);
-			modelAttr = new SemanticModelAttribute ("templateListStore__Name", "templateListStore__Icon", "templateListStore__Template");
+			modelAttr = new SemanticModelAttribute ("templateListStore__Name", "templateListStore__Category", "templateListStore__Icon", "templateListStore__Template");
 			TypeDescriptor.AddAttributes (templatesTreeStore, modelAttr);
 
 			templateCategoriesTreeView.Selection.Changed += TemplateCategoriesTreeViewSelectionChanged;
@@ -138,7 +138,7 @@ namespace MonoDevelop.Ide.Projects
 			var templateTextRenderer = (GtkTemplateCellRenderer)renderer;
 			templateTextRenderer.Template = template;
 			templateTextRenderer.TemplateIcon = model.GetValue (it, TemplateIconColumn) as Xwt.Drawing.Image;
-			templateTextRenderer.TemplateCategory = model.GetValue (it, TemplateNameColumn) as string;
+			templateTextRenderer.TemplateCategory = model.GetValue (it, TemplateCategoryPathColumn) as string;
 		}
 
 		static void SetLanguageCellData (TreeViewColumn col, CellRenderer renderer, TreeModel model, TreeIter it)
@@ -392,13 +392,15 @@ namespace MonoDevelop.Ide.Projects
 				var iter = templatesTreeStore.AppendValues (
 					MarkupTopLevelCategoryName (subCategory.Name),
 					null,
+					null,
 					null);
 
 				foreach (SolutionTemplate template in subCategory.Templates) {
 					if (template.HasProjects || controller.IsNewSolution) {
 						templatesTreeStore.AppendValues (
 							iter,
-							template.Name,
+							template.Name + "\n" + template.Description,
+							null,
 							GetIcon (template.IconId, IconSize.Dnd),
 							template);
 					}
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -303,6 +303,11 @@ protected override void OnDestroyed ()
 
 		void LoadTemplates ()
 		{
+			// set category path once to use it after 
+			foreach (var recentTemplate in controller.RecentTemplates) {
+				recentTemplate.CategoryPath = controller.GetCategoryPathText (recentTemplate);
+			}
```
mrward: This does not look right. The dialog should not be changing the SolutionTemplate.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -303,6 +303,10 @@ protected override void OnDestroyed ()
 
 		void LoadTemplates ()
 		{
+			foreach (var recentTemplate in controller.RecentTemplates) {
+				recentTemplate.CategoryPath = controller.GetCategoryPathText (recentTemplate);
+			}
+
```
mrward: I think instead of doing this I would add a TemplateCategoryPath to the GtkTemplateCellRender and [set it at the same time the TemplateCategory is set](https://github.com/mono/monodevelop/blob/85e0bf4519fa76a359c589b3d1ef8768ec107025/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs#L141).

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -419,7 +423,7 @@ void ShowRecentTemplates ()
 				if (template.HasProjects || controller.IsNewSolution) {
 					templatesTreeStore.AppendValues (
 						iter,
-						controller.GetCategoryPathText (template),
+						template.Name + "\n" + template.Description,
```
mrward: I think this breaks what Sevo did for Voice Over so it read out the category path for the recently used template. This looks like we now only get the template name and description. 

I think we should have Voice Over read out template name, then the category path, then the template description. I think these could be separate rows in the tree store if we wanted them. I believe Voice Over will read all the text columns set, but would need checking.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
mykyta-bondarenko: @mrward I've fixed it

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
mykyta-bondarenko: @mrward I've added a category to a VO text

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -392,13 +392,15 @@ void ShowTemplatesForCategory (TemplateCategory category)
...
 
 				foreach (SolutionTemplate template in subCategory.Templates) {
 					if (template.HasProjects || controller.IsNewSolution) {
 						templatesTreeStore.AppendValues (
 							iter,
-							template.Name,
+							template.Name + "\n" + template.Description,
+							null,
 							GetIcon (template.IconId, IconSize.Dnd),
 							template);
```
mrward: I am thinking we should split up name and description and use different columns. The problem here is that there is another change where we want the language to be read out by voice over. https://github.com/mono/monodevelop/pull/8991/files

We probably want voice over to read the template name, then category, then language, then description. If we have these as separate columns in the above order I think voice over should read them in that order.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -412,14 +414,17 @@ void ShowRecentTemplates ()
...
 				null,
 				null);
 			foreach (SolutionTemplate template in controller.RecentTemplates) {
 				if (template.HasProjects || controller.IsNewSolution) {
+					var category = controller.GetCategoryPathText (template);
 					templatesTreeStore.AppendValues (
 						iter,
-						controller.GetCategoryPathText (template),
+						template.Name + "\n" + category + "\n" + template.Description,
+						category,
```
mrward: This looks like voice over will read template name, category, description and then category again?

mykyta-bondarenko: @mrward it reads [name] -> [category] -> [description]
image

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
mykyta-bondarenko: >We probably want voice over to read the template name, then category, then language, then description. 
VO reads inner readers (if I understood correctly). There are `text-render` and `version-render`, that's why VO reads all 'text' content first.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
mrward: I think we would want language before description, since description could be a long piece of text. //cc @sevoku

PR #9058, opened 2019-10-23
[AspNetCore] Support static assets in Razor Class libraries

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/944135

Test case:
1. Create new Razor Class Lib project, call it `classlib`.
2. Check it targets .NetCoreApp
3. Remove `Areas` folder
4. Add `wwwroot` folder and a `index.js` file inside it with this content: `document.write('Hello from class library')`
5. Add Web project to the solution
6. Add a reference to the `classlib` project
7. Edit `Pages/Index.cshtml` and add `<script src="_content/classlib/index.js"></script>` at the end of the file
8. Run

### DIFF
--- a/main/src/addins/MonoDevelop.AspNetCore/Properties/MonoDevelop.AspNetCore.addin.xml
+++ b/main/src/addins/MonoDevelop.AspNetCore/Properties/MonoDevelop.AspNetCore.addin.xml
@@ -211,7 +211,7 @@
 				condition="UseNetCore31=true"
 				category="netcore/app/aspnet"
 				formatExclude="*.min.css|*.min.js|*.map"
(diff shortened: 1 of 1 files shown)


rodrmoya:
Not sure we want to merge this yet, as the scenario this was going to fix already works without any change (see https://devdiv.visualstudio.com/DevDiv/_testPlans?planId=939525&suiteId=1006562), so I guess we should rather add support to select either .NET Standard or .NETCoreApp for Razor class libs. Will have a look at what VS Win does.

So, marking this as don't merge for now

rodrmoya:
Closing this, as we should better add support for selecting `SupportsPagesAndViews` on the NPD

PR #9066, opened 2019-10-24
[Xwt] Bump Xwt to latest master

BretJohnson:
This bump gets the latest Xwt master branch, which contains 2 a11y fixes from me and a change from @GouriKumari .

The corresponding 8.4 bump is here: https://github.com/mono/monodevelop/pull/9067
Please review/approve both.


mrward:
Updated in xwt in master and md-addins.

PR #9067, opened 2019-10-24
[release-8.4][Xwt] Bump Xwt to latest 8.4

BretJohnson:
This bump gets the latest Xwt d16-4 branch, which contains 2 a11y fixes from me and a change from @GouriKumari .


mrward:
xwt also needs to be updated in md-addins. There is an internal website you can use to do this instead of opening pull requests yourself.

mrward:
Updated xwt in md-addins and monodevelop using the internal tool.

PR #9072, opened 2019-10-24
Fixes VSTS Bug 991658: Crashes from "Critical Gtk Errors" in log

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/991658

AssemblyBrowser navigation points were broken. The assembly
definitions are disposed on close so the definitions need to be
reloaded using file names.

### DIFF
--- a/main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserNavigationPoint.cs
+++ b/main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserNavigationPoint.cs
@@ -23,6 +23,8 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
+#nullable enable
+
 using System;
 using System.Linq;
 using MonoDevelop.Core;
@@ -37,9 +39,9 @@ namespace MonoDevelop.AssemblyBrowser
 	class AssemblyBrowserNavigationPoint : DocumentNavigationPoint
 	{
 		List<string> definitions = new List<string> ();
-		string idString;
+		string? idString;
 
-		public AssemblyBrowserNavigationPoint (ImmutableList<AssemblyLoader> definitions, AssemblyLoader assembly, string idString) : base (assembly?.FileName)
+		public AssemblyBrowserNavigationPoint (ImmutableList<AssemblyLoader> definitions, AssemblyLoader assembly, string? idString) : base (assembly?.FileName)
 		{
 			foreach (var def in definitions) {
 				if (def != null)
@@ -89,7 +91,7 @@ namespace MonoDevelop.AssemblyBrowser
 				if (!string.IsNullOrEmpty (idString)) {
 					if (!string.IsNullOrEmpty (FileName))
 						return String.Format ("{0} : {1}", base.DisplayName, idString);
-					return idString;
+					return idString ?? "";
 				}
 				return base.DisplayName;
 			}
--- a/main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserNavigationPoint.cs
+++ b/main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserNavigationPoint.cs
@@ -36,12 +36,15 @@ namespace MonoDevelop.AssemblyBrowser
 {
 	class AssemblyBrowserNavigationPoint : DocumentNavigationPoint
 	{
-		ImmutableList<AssemblyLoader> definitions;
+		List<string> definitions = new List<string> ();
 		string idString;
 
 		public AssemblyBrowserNavigationPoint (ImmutableList<AssemblyLoader> definitions, AssemblyLoader assembly, string idString) : base (assembly?.FileName)
 		{
-			this.definitions = definitions;
+			foreach (var def in definitions) {
+				if (def != null)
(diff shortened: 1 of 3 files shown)


main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserNavigationPoint.cs
```diff
@@ -36,12 +36,15 @@ namespace MonoDevelop.AssemblyBrowser
...
 	{
-		ImmutableList<AssemblyLoader> definitions;
+		List<string> definitions = new List<string> ();
 		string idString;
 
 		public AssemblyBrowserNavigationPoint (ImmutableList<AssemblyLoader> definitions, AssemblyLoader assembly, string idString) : base (assembly?.FileName)
 		{
-			this.definitions = definitions;
+			foreach (var def in definitions) {
+				if (def != null)
```
Therzok: Why would we pass null here?

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserWidget.cs
```diff
@@ -918,7 +918,7 @@ void HPaneExpose (object sender, Gtk.ExposeEventArgs args)
 		internal void Open (string url, AssemblyLoader currentAssembly = null, bool expandNode = true)
 		{
 			try {
-				Task.WhenAll (this.definitions.Select (d => d.LoadingTask)).ContinueWith (d => {
+				Task.WhenAll (this.definitions.Select (d => d.LoadingTask).Where (t => t != null)).ContinueWith (d => {
```
Therzok: Why would the task be null?

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserNavigationPoint.cs
Therzok: I don't see how `def` can be null. It's always a proper value.

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserWidget.cs
Therzok: Seems like loadingTask is set to `null` after dispose. Can we just set it to `Task.CompletedTask`?

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserNavigationPoint.cs
mkrueger: Was a bug during development. It shouldn't be null there but there is nothing that enforces this.

Therzok:
@mkrueger nullable annotations are a way to enforce non-null

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserWidget.cs
Therzok: Again, set LoadingTask to `Task.CompletedTask` instead of null and remove this null check.

Therzok: And you can mark the file with `#nullable enable` at the top and enable nullable annotations.

mkrueger: It's just making an invariant more visible - nothing more.

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserNavigationPoint.cs
Therzok: Null check is redundant.

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserWidget.cs
mkrueger: tried to do it with nullable enable - but that should be another language. I don't know if that's really an improvement.

Before it was a visible invariant now it's implicit.

main/src/addins/MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/AssemblyBrowserNavigationPoint.cs
```diff
@@ -36,12 +38,15 @@ namespace MonoDevelop.AssemblyBrowser
...
-		string idString;
+		List<string> definitions = new List<string> ();
+		string? idString;
 
-		public AssemblyBrowserNavigationPoint (ImmutableList<AssemblyLoader> definitions, AssemblyLoader assembly, string idString) : base (assembly?.FileName)
+		public AssemblyBrowserNavigationPoint (ImmutableList<AssemblyLoader> definitions, AssemblyLoader assembly, string? idString) : base (assembly?.FileName)
 		{
-			this.definitions = definitions;
+			foreach (var def in definitions) {
+				if (def != null)
```
Therzok: Null check is redundant.

sevoku:
@monojenkins rebase

sevoku:

PR #9074, opened 2019-10-24
[Debugger] When the user clicks the Refresh button, refresh the Objec…

jstedfast:
…tValue

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/999605/

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/DebuggerObjectValueNode.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/DebuggerObjectValueNode.cs
@@ -172,7 +172,9 @@ namespace MonoDevelop.Debugger
 
 		void OnDebuggerValueChanged (object sender, EventArgs e)
 		{
-			OnValueChanged (e);
+			Runtime.RunInMainThread (() => {
+				OnValueChanged (e);
+			}).Ignore ();
 		}
 
 		bool GetCanEdit ()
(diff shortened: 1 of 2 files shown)


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/DebuggerObjectValueNode.cs
```diff
@@ -172,7 +172,9 @@ static Task<ObjectValue[]> GetChildrenAsync (ObjectValue value, int index, int c
 
 		void OnDebuggerValueChanged (object sender, EventArgs e)
 		{
-			OnValueChanged (e);
+			Runtime.RunInMainThread (() => {
+				OnValueChanged (e);
```
Therzok: All async dispatches should be validated with a cancellationtokensource that's canceled and dispose at the same time the view is.

Scenario: pad is destroyed between `OnDebuggerValueChanged` and `OnValueChanged` called on the UI thread.

jstedfast: The nodes aren't disposable. That said, the NSView subclasses that use the nodes and subscribe to their ValueChanged events are all disposable and they all disconnect the even handlers when they are disposed.

Therzok: Dispose runs on the UI thread, so that should be okay. Although, nodes which are still registered for finalization might break.

PR #8965, opened 2019-10-16
[Project] Make Run Configurations accessible

sgmunn:
Fixes VSTS 752769 in conjunction with https://github.com/mono/xwt/pull/987
Accessibility: Project Options - Run Configuration: No on-screen or programmatically associated label provided for the edit fields.
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/752769

PR #9079, opened 2019-10-24
Update description for text and folder selector.

GouriKumari:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1007429
This is also a fix to an issue mentioned in bug 992658


netonjm:
@GouriKumari could you please include a screencast with the voice over popup announcing all the entries you changed?

GouriKumari:
@netonjm : https://www.dropbox.com/s/esntp7rbnv5vowj/ConsoleProject%20Options.mov?dl=0 .

I have changed the  description to label for file selector and folder selector, otherwise it won't read the description when edit text  in group is selected.

sgmunn:
I just realized that this conflicts with https://github.com/mono/monodevelop/pull/8965, although I like that there are longer descriptions for the widgets. One query I have is whether the phrasing is consistent with what is used elsewhere?

to answer my own question, by paraphrasing what @iainx just told me:
use `AtkCocoaHelper` and `control.Accessible.SetTitleUIElement (label.Accessible)`. the phrasing could be more succinct and something like "set additional arguments", but that doesn't set the cocoa accessibility relationships so using SetTitleUIElement is better. Also, no need to mention the type of the control in the label since cocoa uses the role information to pass that to the user.

I'm updating my PR to reflect this as there are other a11y fixes in there that we need as well.

sgmunn:
closing in favor of https://github.com/mono/monodevelop/pull/8965

PR #9080, opened 2019-10-24
[Debugger] null-protect nameView.SetPreviewIcon() in case nameView is…

jstedfast:
… null?

I'm not sure how this is possible... but yay Catalina?

Everything else is null-protected so the only way to get an NRE in this
tiny method is if nameView is null.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1006515/

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacObjectValueTreeView.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacObjectValueTreeView.cs
@@ -636,7 +636,7 @@ namespace MonoDevelop.Debugger
 			if (rowView != null) {
 				var nameView = (MacDebuggerObjectNameView) rowView.ViewAtColumn (0);
 
-				nameView.SetPreviewButtonIcon (icon);
+				nameView?.SetPreviewButtonIcon (icon);
 			}
 		}
 


jstedfast:

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacObjectValueTreeView.cs
```diff
@@ -636,7 +636,7 @@ void UpdatePreviewIcon (nint row, PreviewButtonIcon icon)
 			if (rowView != null) {
 				var nameView = (MacDebuggerObjectNameView) rowView.ViewAtColumn (0);
 
-				nameView.SetPreviewButtonIcon (icon);
+				nameView?.SetPreviewButtonIcon (icon);
```
Therzok:
```
➜  Frameworks ag viewAtColumn
AppKit.framework/Versions/C/Headers/NSTableRowView.h
100:- (nullable id)viewAtColumn:(NSInteger)column;

AppKit.framework/Versions/C/Headers/NSTableColumn.h
95:@property (strong) id /* NSCell * */ dataCell; // Use a view-based NSTableView, and see the -tableView:viewForTableColumn:row: delegate method and -viewAtColumn:row:makeIfNecessary:
96:- (id)dataCellForRow:(NSInteger)row; // Use a view-based NSTableView, and see -viewAtColumn:row:makeIfNecessary

AppKit.framework/Versions/C/Headers/NSTableView.h
435:/* View Based TableView: Access to a view (or row view) for a particular 'row' / 'column'. Returns 'nil' if there is no view at that location. An exception will be thrown if 'row' is not within the numberOfRows. The returned result is always an 'id' for easy assignment to local types, however, -viewAtColumn:row: will always return an NSView, and rowViewAtRow: will always return an NSTableRowView (or subclass thereof). The returned result should generally not be held onto for longer than the current run loop cycle.
440:- (nullable __kindof NSView *)viewAtColumn:(NSInteger)column row:(NSInteger)row makeIfNecessary:(BOOL)makeIfNecessary API_AVAILABLE(macos(10.7));
809:- (nullable NSCell *)preparedCellAtColumn:(NSInteger)column row:(NSInteger)row  API_DEPRECATED("Use View Based TableView and -viewAtColumn:row:", macos(10.5,10.10));
```

Therzok: Seems like it is nullable regardless!

PR #9082, opened 2019-10-24
[Debugger] When the BackgroundStyle is set, only UpdateContents() if …

jstedfast:
…Superview is non-null

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1005195/


jstedfast:

PR #9085, opened 2019-10-25
[NuGet]  Show license metadata

mrward:
- Show license metadata in Manage Packages dialog  …

If the NuGet package source has license metadata then show this
information instead of the View License link. License metadata
can either be an license expression or a file.

If the license is a file then clicking View License will open a separate
dialog with the license read from the file.

Fixes VSTS #1005394 - Support the new "license" properties in the NuGet
package manager UI

 - Show license metadata in License Acceptance dialog

If the NuGet package has associated license metadata then show this
in the dialog. License metadata can be an expression or a file.
A license file will now be displayed in a dialog if the View License
link is clicked. Warning icons and associated message will also be
displayed for deprecated licenses.

There are problems here. Could not get the text to wrap so ended up
making the dialog resizable and always showing the scrollbars. Not wrapping should be OK in the majority of cases since the text should fit unless a complex license expression is used. Scrollbars show with a whitebackground unlike the scrollbars when using the GTK# backend.

Hyperlinks in license expressions do not open the url in a browser
since the Xwt Xamarin.Mac's LabelBackend does not implement support
for this. Worked around this by using the gtk# backend instead.

Dialog no longer shrinks to fit. Using the GTK# backend the dialog is displayed before you can resize so you can see it shrink, unlike when Xamarin.Mac backend is used. The resize code has been removed.

 - Screenshots

MITLicense

LicenseExpression

LicenseWarnings

LicenseFileDialog

LicenseAcceptanceDialog

### DIFF
--- a/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement.Gui/LicenseFileDialog.cs
+++ b/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement.Gui/LicenseFileDialog.cs
@@ -24,14 +24,16 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using NuGet.PackageManagement.UI;
 using Xwt;
 using Xwt.Formats;
 
-namespace MonoDevelop.PackageManagement.Gui
+namespace MonoDevelop.PackageManagement
 {
-	class LicenseFileDialog : Dialog
+	sealed class LicenseFileDialog : Dialog
 	{
 		RichTextView textView;
 		LicenseFileText licenseFileText;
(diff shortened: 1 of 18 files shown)


karann-msft:
what does selecting "view license" do when the package is coming from a v3 feed like nuget.org?

mrward:
@karann-msft View License opens the license url in the default browser or the license file (if one is available) in a dialog.

main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement.Gui/LicenseFileDialog.cs
```diff
@@ -55,11 +53,16 @@ void Build ()
...
+			Content = scrollView;
+
 			textView = new RichTextView ();
-			Content = textView;
+			textView.ReadOnly = true;
+			scrollView.Content = textView;
 
-			Buttons.Add (Command.Ok);
-			DefaultCommand = Command.Ok;
+			var okCommand = new Command ("Ok", Core.GettextCatalog.GetString ("OK"));
```
Therzok: There is [Command.Ok](http://source.monodevelop.com/#Xwt/Xwt/Command.cs,1b7c5fd6fc2419b8,references) already

mrward: Yes, originally that was used. However in the UX review there were complaints about showing 'Ok' instead of 'OK' as the button text. I guess I could change the Command.Ok but did not want to make that change.

vancura: This is interesting. So this means all "OK" in MD are "Ok"?

mrward: Only dialogs that use Xwt and use the [Command.Ok](https://github.com/mono/xwt/blob/0b54869e24438f1b36aee977e799f0b82481ed7b/Xwt/Xwt/Command.cs#L60) will show 'Ok'.

vancura: I think it would make sense to reword this globally to "OK". @hbons?

mrward:
Bit stuck with this. Showing the license acceptance dialog with the native toolkit means the hyperlinks do not work on this dialog. You cannot click them.

Could not see how to support clickable links in XWT with the NSTextField - some research suggests to either use the NSTextView, since this has support to finding the character based on a mouse click, or maybe adding a bunch of code to register a layout manager with the NSTextField.

Using the non-native (gtk) license license acceptance dialog has bad UX. The list scrolls down and you get odd tearing of the UI:

NuGetLicenseDialog-Tearing

Also ended up removing the resize logic when using gtk since you see the dialog resize after it is displayed which does not not look great here.

Other problem is that license expressions (with more than one hyperlink in the label) is not accessible. In general NuGet packages only have a single license but NuGet supports expressions. The single license is OK with accessibility (currently you cannot open hyperlinks with Voice Over but that is not specific to this dialog).

PR #9088, opened 2019-10-25
[Debugger] ViewAtColumn can return null so protect against that

jstedfast:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1008045/


jstedfast:

PR #9093, opened 2019-10-25
[Ide] Handle unauthorized access exception on startup

mrward:
Handle an UnauthorizedAccessException when trying to determine if
the filename passed as a startup argument is a workspace item. The
exception is caught and logged and return a non-zero exit code.
Previously this would generate a fatal exception.

```
System.IO.FileStream..ctor(String,FileMode,FileAccess,FileShare,Int32,Boolean,FileOptions)
System.IO.FileStream..ctor(String,FileMode,FileAccess,FileShare,Int32,FileOptions)
System.IO.StreamReader..ctor(String,Encoding,Boolean,Int32)
System.IO.StreamReader..ctor(String,Boolean)
System.IO.StreamReader..ctor(String)
MonoDevelop.Projects.MSBuild.SlnFile.GetFileVersion(String)
MonoDevelop.Projects.MSBuild.SlnFileFormat.CanReadFile(String,MSBuildFileFormat)
MonoDevelop.Projects.MSBuild.MSBuildFileFormat.CanReadFile(FilePath,Type)
MonoDevelop.Projects.MSBuildSerializationExtension.CanRead(FilePath,Type)
MonoDevelop.Projects.ProjectService.GetObjectReaderForFile(FilePath,Type)
MonoDevelop.Projects.ProjectService.FileIsObjectOfType(FilePath,Type)
MonoDevelop.Projects.ProjectService.IsWorkspaceItemFile(FilePath)
MonoDevelop.Ide.IdeStartup.Run(MonoDevelopOptions)
MonoDevelop.Ide.IdeStartup.Main(String[],IdeCustomizer)
```

Fixes VSTS #1005197 - [FATAL] System.UnauthorizedAccessException
exception in System.IO.FileStream..ctor()

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
@@ -187,9 +187,14 @@ namespace MonoDevelop.Ide
 
 			if (!options.NewWindow && startupInfo.HasFiles) {
 				foreach (var file in startupInfo.RequestedFileList) {
-					if (MonoDevelop.Projects.Services.ProjectService.IsWorkspaceItemFile (file.FileName)) {
-						options.NewWindow = true;
-						break;
+					try {
+						if (MonoDevelop.Projects.Services.ProjectService.IsWorkspaceItemFile (file.FileName)) {
+							options.NewWindow = true;
+							break;
+						}
+					} catch (UnauthorizedAccessException ex) {
+						LoggingService.LogError (string.Format ("Unable to check startup file is a workspace item '{0}'", file.FileName), ex);
+						return 1;
 					}
 				}
 			}


mrward:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
```diff
@@ -187,9 +187,14 @@ int Run (MonoDevelopOptions options)
...
-						options.NewWindow = true;
-						break;
+					try {
+						if (MonoDevelop.Projects.Services.ProjectService.IsWorkspaceItemFile (file.FileName)) {
+							options.NewWindow = true;
+							break;
+						}
+					} catch (UnauthorizedAccessException ex) {
+						LoggingService.LogError (string.Format ("Unable to check startup file is a workspace item '{0}'", file.FileName), ex);
+						return 1;
```
Therzok: This will just exit the app?

Therzok: Isn't this the same case where the user has open previous solution set?

Therzok: Nope, that's not the case.

Therzok:
@monojenkins rebase

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
mrward: Not sure what the best approach is here. Either we give up since we cannot open the file (which is implemented), or we try to continue on, opening the IDE, or passing the sln file to another instance (which changes the behaviour of the code since we would not set NewWindow to true), then the IDE will fail (silently it seems) trying to open that solution.

I tested other areas here:

  - Recent projects - the main IDE window opens but the solution does not load. No error shown to the user.
  - File - Open - Nothing happens on clicking Open. Also seems to cause the dialog to break - unable to close it.

Ideally we would show an error to the user here but right now it seems we silently fail with no information given to the user.

PR #9096, opened 2019-10-25
[Ide] Unset some environment variables after they're no longer needed

Therzok:
Should prevent polluting child processes with these values

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1000552

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Gettext.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Gettext.cs
@@ -138,6 +138,9 @@ namespace MonoDevelop.Core
 			catch (Exception ex) {
 				Console.WriteLine (ex);
 			}
+
+			Environment.SetEnvironmentVariable ("MONODEVELOP_LOCALE_PATH", null);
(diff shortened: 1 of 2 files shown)


sgmunn (approved):

sgmunn:
This fixes the issue I saw with https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1000552

rodrmoya (approved):

sgmunn:
although, this doesn't seem to help .net core test projects, only mono

Therzok:
@sgmunn we have this bug in .netcore projects? how can that be related to the monostub commit?

mrward:
Did .NET Core tests work before? Checking older VS Mac versions (e.g. 8.3.1 and 8.3.0), which are OK when debugging NUnit tests with Mono, do not show the exception caught dialog with .NET Core xUnit and .NET Core NUnit test projects.

Also running a .NET Core console project that throws an exception in its Main does not show the exception caught tooltip, unlike a classic console project.

Possibly we are missing some support in our VS Debug Protocol implementation?

mrward (approved):

Therzok:
@monojenkins backport release-8?4

Therzok:

slluis (approved):

PR #9102, opened 2019-10-26
[DotNetCore] Allow selecting frameworks even if not installed

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/973338

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Gui/DotNetCoreRuntimeOptionsPanelWidget.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Gui/DotNetCoreRuntimeOptionsPanelWidget.cs
@@ -35,7 +35,7 @@ namespace MonoDevelop.DotNetCore.Gui
 {
 	partial class DotNetCoreRuntimeOptionsPanelWidget
 	{
-		readonly List<TargetFramework> frameworks;
+		readonly List<TargetFramework> installedFrameworks;
 		readonly DotNetProject project;
 		readonly DotNetCoreProjectExtension dotNetCoreProject;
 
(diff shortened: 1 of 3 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectSupportedTargetFrameworks.cs
```diff
@@ -60,6 +60,40 @@ public IEnumerable<TargetFramework> GetFrameworks ()
...
+		static string [] supportedNetCoreAppVersions = new string [] {
+			"3.1", "3.0", "2.2", "2.1"
+		};
+
+		public IEnumerable<TargetFramework> GetKnownFrameworks ()
+		{
+			static IEnumerable<TargetFramework> GetKnownNetStandardFrameworks ()
+			{
+				foreach (var v in supportedNetStandardVersions) {
+					yield return CreateTargetFramework (".NETStandard", v);
```
rodrmoya: By doing this, we get warnings in the log about that TFM not being known to the Assembly service, so I guess these should be registered in a different way @mrward ?

mrward: There is no support for that in the assembly service. There is a related bug filed about this causing all frameworks to be loaded. We may need to special case the assembly service for .net framework and .netstandard - just treat them as known frameworks to avoid the warning.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore.Gui/DotNetCoreRuntimeOptionsPanelWidget.cs
```diff
@@ -94,10 +96,10 @@ public void Store ()
 			if (project == null || runtimeVersionCombo.Active == -1 || project.HasMultipleTargetFrameworks)
 				return;
 
-			TargetFramework framework = frameworks [runtimeVersionCombo.Active];
+			TargetFramework framework = installedFrameworks [runtimeVersionCombo.Active];
```
rodrmoya: ugh, this is wrong, it should use `knownFrameworks` here. Changing...

rodrmoya:

PR #9105, opened 2019-10-26
Improve feature switch service

rodrmoya:


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreFeatureSwitchController.cs
```diff
@@ -0,0 +1,56 @@
...
+
+using System.Collections.Generic;
+using Mono.Addins;
+using MonoDevelop.Core;
+using MonoDevelop.Core.FeatureConfiguration;
+
+namespace MonoDevelop.DotNetCore
+{
+	[Extension (typeof (IFeatureSwitchController))]
+	public class DotNetCoreFeatureSwitchController : IFeatureSwitchController
```
nosami: This should be named VBNetDotCoreFeatureSwitchController?

rodrmoya: Not really, the idea is to put there all .NET Core related feature toggles, it happens to have only VB.NET for now, but it will have more for sure (Blazor, for instance)

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
```diff
@@ -25,13 +25,34 @@
...
 
 namespace MonoDevelop.Core.FeatureConfiguration
 {
 	public static class FeatureSwitchService
 	{
+		static ImmutableList<IFeatureSwitchController> featureControllers = ImmutableList<IFeatureSwitchController>.Empty;
+
+		static FeatureSwitchService ()
+		{
+			AddinManager.AddExtensionNodeHandler (typeof (IFeatureSwitchController), HandleFeatureSwitchExtension);
```
Therzok: This is going to force load a lot of assemblies on startup. What can we do (at addin.xml) to avoid this?

We need to register some kind of filters, because as the code is now, every assembly that defines an IFeatureSwitchController will be forcefully loaded when this constructor is hit.

Therzok: Instead of registering a class, maybe we could express everything in xml metadata items (`id`, `_description`, `defaultValue`).

Therzok (changes requested):
The assembly load issue should be fixed first, then I'll do another review pass.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
```diff
@@ -69,5 +89,29 @@ public static class FeatureSwitchService
 
 			return null;
 		}
+
+		#region Internal API for unit tests
+
+		internal static IEnumerable<FeatureSwitch> DescribeFeatures ()
```
mrward: Are we going to have a way to enable/disable features in the IDE? Just wondering we have a description if nothing is using it and there is no way using the FeatureSwitchService to get all the features provided, without using `AddinManager.GetExtensionObjects<IFeatureSwitchController>` and checking each controller.

Reason I ask is that I was looking at a [feature switch addin](https://github.com/mrward/monodevelop-feature-switch-addin) that allowed you to enable/disable features but there was no way to discover all the features apart from when one was requested.

Therzok: I like the approach in the feature switch addin!

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
rodrmoya: Yes, or maybe lazy loading?

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
rodrmoya: > Are we going to have a way to enable/disable features in the IDE? Just wondering we have a description if nothing is using it and there is no way using the FeatureSwitchService to get all the features provided, without using `AddinManager.GetExtensionObjects<IFeatureSwitchController>` and checking each controller.
> 
> Reason I ask is that I was looking at a [feature switch addin](https://github.com/mrward/monodevelop-feature-switch-addin) that allowed you to enable/disable features but there was no way to discover all the features apart from when one was requested.

yeah, was thinking about adding a UI for that, that's why I changed the implementation, but wasn't sure how to better add it, but your addin sounds a much better idea indeed! And indeed, the `DescribeFeatures` is exactly for that, I made it internal for now as nothing was using, apart from tests, but indeed, we can make it public, and even add a much better API to be used by that addin (like maybe events for when new features are added/removed/enabled, ...?)

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
Therzok: Lazy loading can still trigger the assembly load problem if something uses feature switch service earlier in the IDE.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitch.cs
```diff
@@ -26,8 +26,19 @@
 
 namespace MonoDevelop.Core.FeatureConfiguration
 {
-	public interface IFeatureSwitchController
+	public class FeatureSwitch
```
Therzok:
```suggestion
	internal class FeatureSwitch
```

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
```diff
@@ -25,15 +25,44 @@
...
+using System.Collections.Immutable;
 using System.Linq;
 using Mono.Addins;
+using MonoDevelop.Core.Addins;
 
 namespace MonoDevelop.Core.FeatureConfiguration
 {
 	public static class FeatureSwitchService
 	{
+		static ImmutableDictionary<string, FeatureSwitch> featureSwitches = ImmutableDictionary<string, FeatureSwitch>.Empty;
```
Therzok: Why ImmutableDictionary? Addin information is only processed on the UI thread, so there's no need for thread-safety.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
```diff
@@ -25,15 +25,44 @@
...
+
+			initialized = true;
+			AddinManager.AddExtensionNodeHandler ("/MonoDevelop/Core/FeatureSwitches", HandleFeatureSwitchExtension);
+		}
+
+		static void HandleFeatureSwitchExtension (object sender, ExtensionNodeEventArgs args)
+		{
+			var fs = args.ExtensionNode as FeatureSwitchExtensionNode;
+			if (fs != null) {
+				if (args.Change == ExtensionChange.Add && !featureSwitches.TryGetValue (fs.Id, out _)) {
```
Therzok: This is not quite correct. If the arg kind is `Add` and there already is a registered feature switch on that ID, it'll remove the existing one and not add the new one, making it so none are registered on the same ID.

Maybe log an error about it, and keep the first registered one.

Therzok: Maybe move the `TryGetValue` call inside `RegisterFeatureSwitch` so it's easier to reason about.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
```diff
@@ -25,15 +25,44 @@
...
 	public static class FeatureSwitchService
 	{
+		static ImmutableDictionary<string, FeatureSwitch> featureSwitches = ImmutableDictionary<string, FeatureSwitch>.Empty;
+		static bool initialized = false;
+
+		static void EnsureInitialized ()
+		{
+			if (initialized)
+				return;
+
```
Therzok:
```suggestion

			Runtime.AssertMainThread ();
```

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
```diff
@@ -49,25 +78,35 @@ public static class FeatureSwitchService
...
 
 			return null;
 		}
+
+		#region Internal API for unit tests
+
+		internal static IEnumerable<FeatureSwitch> DescribeFeatures ()
+		{
+			EnsureInitialized ();
+			return featureSwitches?.Values;
```
Therzok:
```suggestion
			return featureSwitches.Values;
```

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
```diff
@@ -49,25 +78,35 @@ public static class FeatureSwitchService
...
-				foreach (var ext in extensions) {
-					switch (ext.IsFeatureEnabled (featureName)) {
-					case true:
-						explicitlyEnabled = true;
-						break;
-					case false:
-						explicitlyDisabled = true;
-						break;
-					}
+			if (featureSwitches != null) {
```
Therzok: this null check looks redundant.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
rodrmoya: ok, moved now to be all in the addin.xml files, as indeed, all `IFeatureSwitchController` implementations were just returning the `defaultValue`

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitch.cs
rodrmoya: I set it as public so that it could be used by @mrward 's addin, but I think there's no need for that, we can just have public API in `FeatureSwitchService`to do it, so changing to internal then

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
rodrmoya: Ah ok, just copied from other similar services, but better to use a simple `Dictionary`

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
rodrmoya: yeah, changed now

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
rodrmoya: @mrward yeah, added -> https://github.com/mono/monodevelop/pull/9105/files#diff-9e619a97ca9961bdaeb48e7115c5cb91R90

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitch.cs
```diff
@@ -26,8 +26,21 @@
 
 namespace MonoDevelop.Core.FeatureConfiguration
 {
-	public interface IFeatureSwitchController
+	internal class FeatureSwitch
```
mrward: The problem here is that this is a public API break. This will break any addins that implement a feature switch controller, which is possibly is not going to be likely, but would break my addin, for example. IFeatureSwitchController has been available since v7.8.

Therzok: We can keep the old interface as back-compat, but obsolete it and remove it when possible. The old extensibility point causes force loads of assemblies due to having to instantiate the controllers at runtime.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.AddIns/FeatureSwitchExtensionNode.cs
```diff
@@ -0,0 +1,46 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+using System;
+using Mono.Addins;
+
+namespace MonoDevelop.Core.Addins
+{
+	[ExtensionNode (Description = "Feature switch information.")]
+	class FeatureSwitchExtensionNode : ExtensionNode
```
Therzok: Serialization optimization:
```suggestion
	sealed class FeatureSwitchExtensionNode : ExtensionNode
```

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitch.cs
Therzok:
```suggestion
	internal sealed class FeatureSwitch
```

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
Therzok: Add a main thread assertion, addinmanager can't be used from background threads.

rodrmoya: Yeah, well, this is called from `IsFeatureEnabled`, which should be possible to call from any thread, so just asserting will just throw an exception when so. Will look at improving the initialisation.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitch.cs
rodrmoya: I've added back the interface and code to query `IFeatureSwitchController` implementations, although just as fall back if all the other methods "fail", so this will still break, although not API-related, Matt's addin, as his implementation will never be called for the feature flags that are correctly registered via the .addin.xml.

To fix that, I could move the fallback code to be run before checking the xml-based ones, but that would mean having the assembly load issue Marius was talking about, as soon as `IsFeatureEnabled` is called.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
```diff
@@ -25,13 +25,35 @@
...
 
 namespace MonoDevelop.Core.FeatureConfiguration
 {
 	public static class FeatureSwitchService
 	{
+		static Dictionary<string, FeatureSwitch> featureSwitches = new Dictionary<string, FeatureSwitch> ();
+
+		static FeatureSwitchService ()
+		{
+			AddinManager.AddExtensionNodeHandler ("/MonoDevelop/Core/FeatureSwitches", HandleFeatureSwitchExtension);
```
Therzok: We need to run this on the UI thread.

Please add a Runtime.AssertMainThread().

Therzok (changes requested):

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitch.cs
mrward: Not that concerned about breaking my addin, I can update it. Supporting anybody else that has implemented IFeatureSwitchController properly is enough here.

mrward (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Core.FeatureConfiguration/FeatureSwitchService.cs
rodrmoya: But now that it is static, it should be run on the UI thread AFAICS from other similar implementations, do we still need it?

Therzok: This doesn't guarantee it's run on the UI thread. It's only run once, on the caller thread. Either add an `internal static Initialize()` and force call that on startup or add assert

Therzok: Or implement `IService`, http://source.monodevelop.com/#MonoDevelop.Core/MonoDevelop.Core/IService.cs,7a5d97d8a5b5df09,references

Therzok: Actually, might be simpler to extend `Service`:
http://source.monodevelop.com/#MonoDevelop.Core/MonoDevelop.Core/Service.cs,3f3b4ad9dcdc81c6,references

Therzok (approved):

nosami (approved):

PR #9110, opened 2019-10-28
[Core] Fix MSBuild functions not being evaluated

mrward:
Defining a property that used an MSBuild instrinsic function would
not evaluate if the case did not match. For example:

    <PropertyGroup>
      <IsMac>$([MSBuild]::IsOsPlatform('OSX'))</IsMac>
    </PropertyGroup>

This would not evaluate to true since the instrinsic function's
case is IsOSPlatform. However MSBuild on the command line would
evaluate this correctly.

Fixed this by making the method cache lookup case insensitive so it
matches MSBuild's behaviour.

Fixes VSTS #1008396 - DefineConstants not working right if they are
set in imported projects


mrward:

Therzok (approved):

sgmunn (approved):

PR #9115, opened 2019-10-28
[Debugger] Don't dereference statusIcon.Image for its size, just use …

jstedfast:
…ImageSize


jstedfast:

PR #9118, opened 2019-10-28
Prevents OpenFile dialog set a disposable parent window setting to IDE or GTC instead the last focused top level

netonjm:
GetFocusedTopLevelWindow gets the most top level window, but in some contexts this could be transients windows which could be closed unexpectly for a reason.

I changed to use IDE or GTC as parent to get a consistent not disposable windowlike parent of OpenFile dialog

Fixes VSTS #998575 - [FATAL] System.ObjectDisposedException exception in Foundation.NSObject.get_SuperHandle()


sevoku:

netonjm:
@monojenkins rebase

PR #9120, opened 2019-10-28
Prevents not disposable parent window in nuget license dialog setting to IDE or GTC instead use XwtRootWindow

netonjm:
I changed in nuget license dialog to use IDE or GTC as parent to get a consistent not disposable window like parent

Fixes VSTS #961624 - "System.ObjectDisposedException: Cannot access a disposed object" error when trying to access nuget license dialog


sevoku:

PR #9121, opened 2019-10-28
Only brings ErrorListPad to front in case of be loaded in Workbench

netonjm:
Only brings ErrorListPad to front in case of be loaded in Workbench

Fixes #941969 - [FATAL] System.NullReferenceException exception in MonoDevelop.MacIntegration.MainToolbar.BuildResultsView.MouseDown()

### DIFF
--- a/main/src/addins/MacPlatform/MainToolbar/StatusBar.cs
+++ b/main/src/addins/MacPlatform/MainToolbar/StatusBar.cs
@@ -242,7 +242,7 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 
 		public override void MouseDown (NSEvent theEvent)
 		{
-			IdeApp.Workbench.GetPad<MonoDevelop.Ide.Gui.Pads.ErrorListPad> ().BringToFront ();
+			IdeApp.Workbench.GetPad<MonoDevelop.Ide.Gui.Pads.ErrorListPad> ()?.BringToFront ();
 		}
 
 		string INSAccessibilityStaticText.AccessibilityValue {


Therzok:
Looking at WrapPad:
http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs,28ef42ba3aee7408

It seems that the pad is removed from the list when destroyed.

`GetPad<T>` goes through `Pads` which is only constructed once. So it is possible for it to be null because someone destroyed the window (via the x button)

Therzok:
I'm pretty sure we should be reconstructing the pad if not found.

netonjm:
Yeah @Therzok, I think it depends on what behaviour we want in case of the user closes the panel (and it's destroyed).

In theory (please correct me) if I'm wrong,  this brings to front when build finishes the panel to show the results. Then here the question is do want always want to show it even though the user closes it?

For me makes sense as @Therzok says to force open it's quite important at the end of the build. thoughts @sevoku ?

mrward:
In this case the user has clicked the error or warning icon in the status bar so we should always open the Errors window. If that means we need to re-create the Errors window then we should do that and then display it. Doing nothing would not be a good user experience.

There is a question of where this logic should go. I think the IdeApp.Workbench.GetPad should recreate the pad, if it needs to, since it knows about them instead of having the logic in several places.

netonjm:
Makes completely sense @mrward !

netonjm:
I was investigating with @mrward , and we don't have any kind of dispose or remove the panel. 
Hide panel don't destroy the object, it seems we only dispose panels on close the IDE. I added some additional logging in case of NRE to get more information related to that

netonjm:
@monojenkins rebase

PR #9126, opened 2019-10-28
[Debugger] Prevent pinned watches from disappearing during scroll/etc

sandyarmstrong:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/999608


jstedfast:
Do the pinned watches collapse if you expand a pinned watch and then scroll? I'll give this a try in a bit once I finish the patch I'm working on atm.

sandyarmstrong:
@jstedfast no, they do not collapse. What behavior do you want?

jstedfast:
If they don't collapse on scroll, that is *perfecto*!

sandyarmstrong:
Ok, cool. Fixed the other related issue I mentioned earlier, so I think this is done.

Perf can get stuttery when an expanded pinned watch scrolls back into view, but I think that can be addressed later.

Let me know if you have any other concerns!

jstedfast:
I see what you mean by stuttering, but I agree, that can be solved later. Right now this is a huge improvement!

sandyarmstrong:

PR #9131, opened 2019-10-28
Updated Visual Studio theme

jmatthiesen:
Updated the Visual Studio theme to fix the text selection color, which was too dark and made text hard to read when selected.

image

While making the change, I also updated a few other colors to match Windows:
* Usages (i.e. identifier highighting)
image

* Find results
image

* Link colors
image

Fixes VSTS #1009498.


sgmunn:
@jmatthiesen a screenshot here would help so that @vancura can easily see the changes

jmatthiesen:
@sgmunn Updated - good call. @vancura I've added in screen shots, the colors I updated are using the same colors as VS Windows Light theme.

sgmunn:
@hbons ^, I think @vancura might be FTO at the moment

PR #9139, opened 2019-10-29
Fix exception caught while debugging

DavidKarlas:
While I was editing .cshtml this code threw exception often, so I fixed it the way I think it should be to never throw exception

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/TextFileNavigationPoint.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/TextFileNavigationPoint.cs
@@ -58,10 +58,12 @@ namespace MonoDevelop.Ide.Navigation
 
 		void RefreshWithCurrentOffset (ITextView textView)
 		{
-			if (textView != null && offset.HasValue) {
-				var currentLine = textView.TextBuffer.CurrentSnapshot.GetLineFromPosition (offset.Value);
+			if (textView != null && offset is SnapshotPoint sp) {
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Navigation/TextFileNavigationPoint.cs
```diff
@@ -58,10 +58,12 @@ public TextFileNavigationPoint (Document doc, ITextView textView)
...
 		{
-			if (textView != null && offset.HasValue) {
-				var currentLine = textView.TextBuffer.CurrentSnapshot.GetLineFromPosition (offset.Value);
+			if (textView != null && offset is SnapshotPoint sp) {
+				var currentSnapshot = sp.Snapshot.TextBuffer.CurrentSnapshot;
+				var position = sp.TranslateTo (currentSnapshot, PointTrackingMode.Negative).Position;
+				var currentLine = currentSnapshot.GetLineFromPosition (position);
 				line = currentLine.LineNumber;
-				column = offset.Value.Position - currentLine.Start.Position;
+				column = position - currentLine.Start.Position;
```
netonjm: maybe we could ensure here we didn't get here a negative value or log in that case

DavidKarlas: This can't be negative

netonjm: then 👍

DavidKarlas:

PR #9147, opened 2019-10-29
Introduce IDynamicDocumentInfoProvider.

KirillOsenkov:
This is new infrastructure used by Blazor. For certain files in the project such as .razor a generated C# buffer is added to the workspace. We use a similar mechanism to AddVirtualDocuments but the distinction is that the list of .razor files needs to be read by the project system from the project.


KirillOsenkov:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/MonoDevelopWorkspace.cs
```diff
@@ -360,6 +366,8 @@ protected override void Dispose (bool finalize)
 			ProjectHandler.Dispose ();
 			MetadataReferenceManager.ClearCache ();
 
+			dynamicFileManager.UnloadWorkspace (this);
```
sandyarmstrong: Tests failing because apparently it's possibly for `dynamicFileManager` to be `null` here.

sandyarmstrong:
Test failures are same `NameResolutionFailure` problems plaguing all tests right now. I'd prefer to see this go green (will try redeploying them in the morning) but have removed `do-not-merge` label.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/DynamicFileManager.cs
```diff
@@ -0,0 +1,149 @@
...
+			{
+				var oldFiles = dynamicSourceFiles;
+				var newFiles = currentDynamicSourceFiles.ToImmutableHashSet ();
+				dynamicSourceFiles = newFiles;
+
+				var addedFiles = newFiles.Except (oldFiles);
+				var removedFiles = oldFiles.Except (newFiles);
+
+				foreach (var document in removedFiles) {
+					foreach (var dynamicFileProvider in dynamicFileManager.GetDynamicFileProviders (document)) {
```
Therzok: This is going to allocate on a loop. Is this called often?

PR #9163, opened 2019-10-30
Redoing MSBuild dep to avoid whitespace error

joj:


joj:
Closing because it's old and not needed

PR #9167, opened 2019-10-30
[ConnectedServices] Add accessibility support to the connected services view

sevoku:
This is not perfect but makes the connected services accessible. We can improve the usability in the next cycle.

Fixes VSTS #753572


sevoku:
@monojenkins rebase

sevoku:

PR #9171, opened 2019-10-30
Switch to `CompileDesignTime` for getting SourceFiles

DavidKarlas:
This fixes problem with Razor projects that have different set of SourceFiles between DesignTime and Compile time

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1010676

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
@@ -763,15 +763,7 @@ namespace MonoDevelop.Projects
 				}
 
 				if (startTask) {
-					var coreCompileDependsOn = project.sourceProject.EvaluatedProperties.GetValue<string> ("CoreCompileDependsOn");
-
-					if (string.IsNullOrEmpty (coreCompileDependsOn)) {
-						currentTask.SetResult (CoreCompileEvaluationResult.Empty);
-						return currentTask.Task.Result;
-					}
-
 					var result = CoreCompileEvaluationResult.Empty;
-					var dependsList = string.Join (";", coreCompileDependsOn.Split (new [] { ";" }, StringSplitOptions.RemoveEmptyEntries).Select (s => s.Trim ()).Where (s => s.Length > 0));
 					try {
 						// evaluate the Compile targets
 						var ctx = new TargetEvaluationContext ();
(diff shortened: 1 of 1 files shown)


KirillOsenkov:
@mrward this fixes our issue, as described in https://github.com/dotnet/project-system/blob/master/docs/design-time-builds.md

main/src/core/MonoDevelop.Core/MonoDevelop.Projects/Project.cs
```diff
@@ -783,12 +775,12 @@ public async Task<CoreCompileEvaluationResult> GetItemsFromCoreCompileDependenci
 						ctx.BuilderQueue = BuilderQueue.ShortOperations;
 						ctx.LogVerbosity = MSBuildVerbosity.Quiet;
 
-						var evalResult = await project.RunTargetInternal (monitor, dependsList, config.Selector, ctx);
+						var evalResult = await project.RunTargetInternal (monitor, "CompileDesignTime", config.Selector, ctx);
```
sandyarmstrong: This should be a constant.

KirillOsenkov:
We should test if this works for old-style non-SDK projects as they might not have the CompileDesignTime target at all.

Ideally we should be checking for the target's existence and fallback to the existing behavior if that target doesn't exist.

DavidKarlas:
Seems to be working for Main.sln, there is bunch of unit tests failing, fun

KirillOsenkov:
Trying to build `CompileDesignTime` on the test project results in this error:

```
 error : SkipCompilerExecution and ProvideCommandLineArgs should be set before calling the
 CompileDesignTime target
```

Which is reasonable, we should be setting those if we're not ^^

See https://github.com/dotnet/project-system/issues/4348

mrward:
If we are having problems here we could also do a short term workaround by amending the [CoreCompileDepends on target](https://github.com/mono/monodevelop/blob/246aedb81fefa6c2b9e64017a7f51fead7c26f5f/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/PackageManagementMSBuildExtension.cs#L94-L147) to avoid running the Razor target which is causing the problem.

Another idea would be to try setting the DesignTime flag to true instead of switching to CompileDesignTime, see if that avoids the Razor target being run. Probably worth trying this first before amending the .net core project extension to modify the CoreCompileDependsOn targets.

Longer term it would be good to use CompileDesignTime so we are doing the same as VS on Windows. I think VS Win uses a different compile target for other project types - at least it did the last time I checked.

mrward:
Looking at the Razor targets for .NET Core 3.1 preview it looks like setting DesignTimeBuild to true may avoid the .g.cs files being generated since RazorComponentGenerate would not be called, with VS Mac still using the CoreCompileDependsOn logic for now. Needs testing though.

```
    <CoreCompileDependsOn Condition="'$(DesignTimeBuild)'!='true'">
      RazorComponentGenerate;
      $(CoreCompileDependsOn)
    </CoreCompileDependsOn>

    <CoreCompileDependsOn Condition="'$(DesignTimeBuild)'=='true'">
      RazorGenerateComponentDeclarationDesignTime;
      $(CoreCompileDependsOn)
    </CoreCompileDependsOn>
```

DavidKarlas:
https://github.com/mono/monodevelop/pull/9176

PR #9172, opened 2019-10-30
MacIdeAppleEvents: add an apple event for getting current solution path

abock:
Unity needs to be able to query running instances of MD/VS Mac for the active/selected current solution path.

/cc @jbevain

### DIFF
--- a/main/src/addins/MacPlatform/query-vsmac.mm
+++ b/main/src/addins/MacPlatform/query-vsmac.mm
@@ -1,0 +1,204 @@
+//
+// Copyright (c) Microsoft Corp. (https://www.microsoft.com)
+//
+// Authors:
+//   Aaron Bockover <abock@microsoft.com>
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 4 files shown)


jbevain:
Previously Unity would open their C# project in the first VSM instance it could find.
I've verified that we can use this feature to detect the proper VSM instance to work with.
A back port to 8.5 would be appreciated as well.

main/src/addins/MacPlatform/query-vsmac.mm
```diff
@@ -0,0 +1,204 @@
+//
```
abock: Note that this is just a small reference utility/library to test the feature. This does not actually build or ship as part of VS Mac itself.

Therzok:
@jbevain we are still merging master to 8.5, so it will be picked up automatically.

Therzok:
LGTM, merge at will.

abock:
Would be good to get this for 8.4 though, which is what I think @jbevain was thinking.

abock:

PR #9188, opened 2019-11-01
New command called "Quick New File"

DavidKarlas:
Behaviour is same as VSCode Cmd+N which creates "Untitled-#" file I used Cmd+Option+N shortcut that was available, this is super useful when I want to inspect some log file that I copied from web or terminal output, that I want to search/read in editor...

Fixes https://github.com/mono/monodevelop/issues/9243

### DIFF
--- a/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
+++ b/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
@@ -401,6 +401,13 @@
 			macShortcut = "Meta|N"
 			_description = "Create a new file"
 			icon = "md-regular-file" />
+	<Command id = "MonoDevelop.Ide.Commands.FileCommands.QuickNewFile"
+			defaultHandler = "MonoDevelop.Ide.Commands.QuickNewFileHandler"
+			_label = "Quick New File..."
+			shortcut = "Control|Alt|N"
+			macShortcut = "Meta|Alt|N"
+			_description = "Create a new file without dialog"
+			icon = "md-regular-file" />
 	<Command id = "MonoDevelop.Ide.Commands.FileCommands.Save"
 			icon = "gtk-save"
 			shortcut = "Control|S"
--- a/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
+++ b/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
@@ -403,7 +403,7 @@
 			icon = "md-regular-file" />
 	<Command id = "MonoDevelop.Ide.Commands.FileCommands.QuickNewFile"
 			defaultHandler = "MonoDevelop.Ide.Commands.QuickNewFileHandler"
-			_label = "Quick New File..."
+			_label = "Quick New File"
 			shortcut = "Control|Alt|N"
 			macShortcut = "Meta|Alt|N"
 			_description = "Create a new file without dialog"
--- a/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
+++ b/main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
@@ -406,7 +406,7 @@
 			_label = "Quick New File"
 			shortcut = "Control|Alt|N"
 			macShortcut = "Meta|Alt|N"
-			_description = "Create a new file without dialog"
+			_description = "Create a new empty Text File"
 			icon = "md-regular-file" />
 	<Command id = "MonoDevelop.Ide.Commands.FileCommands.Save"
 			icon = "gtk-save"
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileCommands.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileCommands.cs
@@ -40,6 +40,7 @@ using MonoDevelop.Ide.Desktop;
 using System.Linq;
 using MonoDevelop.Components;
 using MonoDevelop.Components.Extensions;
+using System.Collections.Generic;
 
 namespace MonoDevelop.Ide.Commands
 {
@@ -73,7 +74,8 @@ namespace MonoDevelop.Ide.Commands
 		OpenContainingFolder,
 		SetBuildAction,
 		ShowProperties,
-		CopyToOutputDirectory
+		CopyToOutputDirectory,
+		QuickNewFile
 	}
 
 	// MonoDevelop.Ide.Commands.FileCommands.OpenFile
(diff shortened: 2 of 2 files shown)


netonjm:
Love it 👍

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileCommands.cs
```diff
@@ -126,6 +127,21 @@ protected override void Run ()
 		}
 	}
 
+	// MonoDevelop.Ide.Commands.FileCommands.QuickNewFile
+	public class QuickNewFileHandler : CommandHandler
```
mrward: Can be non-public.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileCommands.cs
```diff
@@ -50,6 +50,7 @@ public enum FileCommands
 	{
 		OpenFile,
 		NewFile,
+		QuickNewFile,
```
mrward: Should be added at the end since otherwise this is an API break.

DavidKarlas: Was thinking about this, is it realistic that someone depends on/uses this enum as integer?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileCommands.cs
DavidKarlas: I just copy pasted NewFileHandler :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileCommands.cs
mrward: It is public so could be used in an addin. You could use it in a command handler.

DavidKarlas: I still see why would anyone depend on integer value of it, btw, I updated PR ;)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileCommands.cs
```diff
@@ -126,6 +127,21 @@ protected override void Run ()
 		}
 	}
 
+	// MonoDevelop.Ide.Commands.FileCommands.QuickNewFile
+	class QuickNewFileHandler : CommandHandler
+	{
+		protected override void Run ()
+		{
+			var names = IdeApp.Workbench.Documents.Select (d => d.Name);
+			for (int i = 1; i < 1000; i++) {
```
mrward: If I create 1000 Untitled files then nothing happens here. I guess would not happen in practice - just wondering what should happen.

DavidKarlas: You run out of memory :P

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileCommands.cs
```diff
@@ -126,6 +127,21 @@ protected override void Run ()
 		}
 	}
 
+	// MonoDevelop.Ide.Commands.FileCommands.QuickNewFile
+	class QuickNewFileHandler : CommandHandler
+	{
+		protected override void Run ()
+		{
+			var names = IdeApp.Workbench.Documents.Select (d => d.Name);
```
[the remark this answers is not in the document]
Therzok:
```csharp
var names = new HashSet<string> (IdeApp.Workbench.Documents.Select (d => d.Name).Where (d => d.StartsWith ("Untitled-"));
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileCommands.cs
Therzok: _what if we iterated from `0` to `int.MaxValue` to run out of CPU too?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileCommands.cs
```diff
@@ -126,6 +127,21 @@ protected override void Run ()
...
+			for (int i = 1; i < 1000; i++) {
+				if (!names.Contains ("Untitled-" + i)) {
+					IdeApp.Workbench.NewDocument ("Untitled-" + i, "text/plain", "").Ignore ();
```
Therzok: Doesn't passing null automatically set the untitled tab name? If so, ignore the hashset comment.

DavidKarlas: Passing `null` throws exception in AssemblyBrowser... Staying with existing logic.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileCommands.cs
DavidKarlas: Now it goes to int.Max, just to be sure you run out of memory...

DavidKarlas:
@slluis should we get input from someone who cares about Keyboard bindings?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Commands/FileCommands.cs
Therzok: That's weird. The AssemblyBrowser is doing something wrong then

sevoku: @DavidKarlas @Therzok if there is a bug in AssemblyBrowser, pls file it!

sevoku:
> @slluis should we get input from someone who cares about Keyboard bindings?

It would be @sayedihashimi

DavidKarlas:
@monojenkins rebase

sayedihashimi:
I tried it a noticed a couple of issues:

- After saving the file, the tab doesn't show the new file name. It still shows "Untitled-1"
- After creating a quick new file, save, & close the file when invoking for a second time there is an error dialog that the file has been autosave.

I made a gif to show you what I'm seeing.

quick-new-file

sayedihashimi:
My only other concern with this feature is how does a user add a file to an existing project? For the projects that I own (.NET Core) it doesn't really matter since we auto include files from the project folder, but for other project types that's not the case. With that being said, I think after the bugs above are fixed, I think it's a great feature to have. We should watch out for feedback from users to see if adding it to the project becomes something that we should look closer at.

sandyarmstrong:
@sayedihashimi the intent of this new command is explicitly to make a new file _not_ in the project. It is to bring up a throwaway editor for pasting some junk or writing some quick notes, etc, the way users do in VS Code.

If a user wants to add a new file to an existing project, they almost definitely know that before creating the file, so they would use the existing New File, New Class, etc commands.

sayedihashimi:
@sandyarmstrong yes, I understand the intent. Just saying we should keep an eye for feedback in case our intent and users’ expectations don’t match.

DavidKarlas:
@sayedihashimi both bugs mentioned are fixed by https://github.com/mono/monodevelop/pull/9347 and https://github.com/mono/monodevelop/pull/9348
This should be merged once those two bugs are fixed.

main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
```diff
@@ -401,6 +401,13 @@
 			macShortcut = "Meta|N"
 			_description = "Create a new file"
 			icon = "md-regular-file" />
+	<Command id = "MonoDevelop.Ide.Commands.FileCommands.QuickNewFile"
+			defaultHandler = "MonoDevelop.Ide.Commands.QuickNewFileHandler"
+			_label = "Quick New File..."
```
sevoku: The `…` indicates an additional action (like a prompt or dialog), which is not the case here
```suggestion
			_label = "Quick New File"
```

main/src/core/MonoDevelop.Ide/ExtensionModel/Commands.addin.xml
```diff
@@ -401,6 +401,13 @@
...
+			shortcut = "Control|Alt|N"
+			macShortcut = "Meta|Alt|N"
+			_description = "Create a new file without dialog"
```
sevoku: wording:
```suggestion
			_description = "Create a new empty Text File"
```

PR #9098, opened 2019-10-25
[Mac] Mirror Xamarin.Mac launcher.m setup closer

Therzok:

PR #9191, opened 2019-11-01
[Mac] Call xamarin_initialize_cocoa_threads so objects constructed on…

Therzok:
… threads that are not the main one are released

Fixes VSTS #1012551 - Launcher does not call xamarin_initialize_cocoa_threads

Split the more important part out of https://github.com/mono/monodevelop/pull/9098


Therzok:
@monojenkins rebase

PR #9193, opened 2019-11-01
[Debugger] Fixed deleting/changing of expressions in Watch Pad

jstedfast:
For deleting, 2 bugs are fixed:

1. Deleting an expression now works between sessions (failed before because
   removing the node sets the Parent to null and thus 'is RootObjectValueNode'
   failed, causing no ExpressionRemoved event to fire.
2. If the user has multiple of the same expression, the old logic would remove
   the first instance even if the user was deleting the last instance.

A similar bug existed for editing an existing expression if the user had
identical expressions in the list.


jstedfast:

PR #9195, opened 2019-11-02
When opening files without solution open maximized(closed pads)

DavidKarlas:
Main reason behind this change is to make IDE more friendly to open single files and behave more like editor. Since default open pads are useless(empty solution and error pad).
If user later opens solution, we automatically revert this.

This fixes https://github.com/mono/monodevelop/issues/9242
Fixes VSTS #1017488

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
@@ -643,6 +643,47 @@ namespace MonoDevelop.Ide.Gui
 
 		public void ToggleMaximize ()
 		{
+			if (workbench.IsInFullViewMode) {
+				Normalize ();
+			} else {
+				Maximize ();
+			}
+		}
+
+		bool normalizeAutomatically = false;
+		/// <summary>
+		/// Maximizes editor area by hiding all other pads, e.g. Solution pad, Errors pad...
+		/// </summary>
+		/// <param name="normalizeAutomatically">When set to true, Workbench will reverse effect of maximazing whenever suitable, e.g. when solution loads</param>
+		public void Maximize (bool normalizeAutomatically = false)
+		{
+			this.normalizeAutomatically = normalizeAutomatically;
+			if (workbench.IsInFullViewMode)
+				return;
+
+			if (normalizeAutomatically) {
+				IdeApp.Workspace.SolutionLoaded += Workspace_SolutionLoaded;
+			}
+			workbench.ToggleFullViewMode ();
+		}
+
+		private void Workspace_SolutionLoaded (object sender, SolutionEventArgs e)
+		{
+			IdeApp.Workspace.SolutionLoaded -= Workspace_SolutionLoaded;
+			if (!normalizeAutomatically)
+				return;
+			normalizeAutomatically = false;
+			Normalize ();
+		}
+
+		/// <summary>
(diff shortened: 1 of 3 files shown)


DavidKarlas:
When opening single file.
Before:
image
After:
image

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
```diff
@@ -643,6 +643,47 @@ public Task<Document> OpenDocument (ModelDescriptor modelDescriptor, DocumentCon
...
+				Maximize ();
+			}
+		}
+
+		bool normalizeAutomatically = false;
+		/// <summary>
+		/// Maximizes editor area by hiding all other pads, e.g. Solution pad, Errors pad...
+		/// </summary>
+		/// <param name="normalizeAutomatically">When set to true, Workbench will reverse effect of maximazing whenever suitable, e.g. when solution loads</param>
+		public void Maximize (bool normalizeAutomatically = false)
```
Therzok: Please avoid optional parameters and add 2 overloads:
```
Maximize();
Maximize(bool);
```

DavidKarlas: Why?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
```diff
@@ -643,6 +643,47 @@ public Task<Document> OpenDocument (ModelDescriptor modelDescriptor, DocumentCon
...
+		{
+			this.normalizeAutomatically = normalizeAutomatically;
+			if (workbench.IsInFullViewMode)
+				return;
+
+			if (normalizeAutomatically) {
+				IdeApp.Workspace.SolutionLoaded += Workspace_SolutionLoaded;
+			}
+			workbench.ToggleFullViewMode ();
```
Therzok: Won't this mess with user preferences for hidden/shown pads?

DavidKarlas: Not sure what you mean by this? FullViewMode/Maximized is not stored in preferences(persisted on restart).

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
Therzok: It makes API easier to reason about when adding new overloads. Given this method with an optional parameter, and say you want to add another parameter.

```
public void Maximize (bool = false, bool = false)
```

This is an ABI break. And you can't have 2 overloads with optional parameters since that will yield a compiler warning.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui/Workbench.cs
Therzok: Yeah, I just re-read the code. I thought we would serialize the dock item status, not the original layout.

DavidKarlas:
No I did not open issue for this.

sevoku:
> No I did not open issue for this.

@DavidKarlas could you please create one and link from here? And see @Therzok's comment, I think it makes sense.

DavidKarlas:
I saw comment from Marius, I will update PR when I have time for this, probably this weekend.
But why do I need to open issue?

sevoku:
Because it's a significant UX change (improvement/papercut fix) that should be tracked and called out in release notes

PR #9196, opened 2019-11-02
[Error Pad] Optimize rendering and remove manual loop pumping

Therzok:
Removing manual loop pumping introduces an UI hang, which is why this is still a draft. Need to figure out how to optimize rendering, sorting, filtering for a scenario where 28000 warnings are added at once

Fixes VSTS #1006554 - Potential crash due to manual UI loop pumping when adding tasks


Therzok:
Chatted with @sevoku and we should move this to a native view. I'll migrate the non-intrusive pad changes to a separate PR and open a new user story for the native error pad

Therzok:
Going to split this PR so we get the UI agnostic changes. Without a native toolkit implementation, we can't optimize it further right now.

PR #9198, opened 2019-11-04
Attempt to make Welcome Page always appear

DavidKarlas:
`GLib.Idle.Add` doesn't always execute, hence this should fix it

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
@@ -61,7 +61,6 @@ namespace MonoDevelop.Ide
 
 		List<AddinError> errorsList = new List<AddinError> ();
 		bool initialized;
-		static bool hideWelcomePage;
 		static StartupInfo startupInfo;
 
 		static TimeToCodeMetadata ttcMetadata;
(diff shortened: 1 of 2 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
```diff
@@ -396,7 +394,13 @@ async Task<int> MainLoop (MonoDevelopOptions options, StartupInfo startupInfo)
 			IdeStartupTracker.StartupTracker.MarkSection ("PumpEventLoop");
 			IdeStartupTracker.StartupTracker.Stop (startupInfo);
 
-			GLib.Idle.Add (OnIdle);
+			if (!hideWelcomePage && !WelcomePage.WelcomePageService.HasWindowImplementation) {
```
Therzok: What do you mean by Idle.Add doesn't always execute?

DavidKarlas: Example given:
1) Start VSMac
2)MainMenu use File->Recent Files->SomeFile.cs
3) DO NOT MOVE YOUR MOUSE AFTER CLICK
4) https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/MDMenuItem.cs#L113 Will not be reached until you move your mouse

That is Glib.Timeout this is GLib.Idle, but its pretty much same...

sevoku: good catch! @Therzok since we changed CompositionManager init anyway (OnIdle was initially used for that, before we added the welcome window), I think this change should be safe, right?

Therzok: Yeah, it should be okay.

PR #9200, opened 2019-11-04
[DotNetCore] Don't allow selection of 3.1 when creating new projects

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1008333


rodrmoya:
> Should we also remove .NET Standard 2.1? Not sure if .NET Standard 2.1 works in 8.3.

2.1 is supported in 3.0 AFAICS, right?

mrward:
@rodrmoya Good point 😄

rodrmoya:
But this is not working with just 3.0 and 3.1 SDKs installed, so DON'T MERGE

rodrmoya:
Closing this, as we won't be doing this for 8.3, and will send a PR to master with a future-proof solution, as this will keep happening

PR #9210, opened 2019-11-04
[Mac][985660] Fix NSSavePanel crash on Catalina

sevoku:
The NSSavePanel.DidChangeToDirectory event is broken on Catalina and using a custom delegate doesn't seem to be supported by Xamarin.Mac (although we don't know if that'd actually fix the crash).

This fixes a potential race with the `NSSavePanel.DidChangeToDirectory` delayed continuation and adds an additional check for `NSOpenSavePanelUrlEventArgs.NewDirectoryUrl` being NSNull, which can happen on Catalina for some unknown reason.

~and disables `NSSavePanel.DidChangeToDirectory` handling on Catalina (introduced in 5a2894490ba43c9e1056d09f39255c794facbe8d for https://github.com/mono/monodevelop/pull/3119)~.

Fixes VSTS #985660
Fixes VSTS #1015827

### DIFF
--- a/main/src/addins/MacPlatform/Dialogs/MacOpenFileDialogHandler.cs
+++ b/main/src/addins/MacPlatform/Dialogs/MacOpenFileDialogHandler.cs
@@ -1,4 +1,4 @@
-﻿// 
+// 
 // MacSelectFileDialogHandler.cs
 //  
 // Author:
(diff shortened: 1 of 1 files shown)


main/src/addins/MacPlatform/Dialogs/MacOpenFileDialogHandler.cs
```diff
@@ -73,25 +73,33 @@ protected override NSSavePanel OnCreatePanel (OpenFileDialogData data)
...
-						data.SelectedFiles = new FilePath [] { selectedPath };
-						pathAlreadySet = true;
-
-						// We need to call Cancel on 1ms delay so it's executed after DidChangeToDirectory event handler is finished
-						// this is needed because it's possible that DidChangeToDirectory event is executed while dialog is opening
-						// in that case calling .Cancel() leaves dialog in weird state...
-						// Fun fact: DidChangeToDirectory event is called from Open on 10.12 but not on 10.13
-						System.Threading.Tasks.Task.Delay (1).ContinueWith (delegate { panel.Cancel (panel); }, Runtime.MainTaskScheduler);
-					};
+					//TODO: DidChangeToDirectory is broken on Catalina
```
Therzok: Need to create a bug in xammac for these things.

sevoku: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1015827

DavidKarlas: Xamarin.Mac, not Visual Studio for Mac

sevoku: @DavidKarlas since I can not reproduce the issue with a simple Mac app, I created the VS Mac issue for further investigation (see bug description). We can't block this PR only because we don't know the actual cause, because it fixes a crash which affects many customers.

sevoku:

PR #9214, opened 2019-11-05
[Mac] Fix binding for task_info

Therzok:
It seems that the API requires the count in natural_t units, not in bytes. Fixing this, fixes the crasher in Catalina
the reason it crashes in Catalina is that in Catalina task_info tries to write more data to the output struct (for Mojave the size of the struct and the maximum size written by Mojave was the same, so it didn't matter that the size argument was too big, because Mojave would never write more than the size of the buffer anyway)

Thanks to @rolfbjarne for the investigative work


slluis:

PR #9216, opened 2019-11-05
Fixes 1007912: Keyboard focus isn't given to Solution pad

avodovnik:
Focus wasn't given back to the Solution pad when the `View | Pads | Solution` command was invoked.


sandyarmstrong:

netonjm:
@sevoku @sandyarmstrong this PR introduces a infinite focus loop in panels with autohide mode

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1020127

PR #9218, opened 2019-11-05
Fix 988924: Some commands (save, undo) do not work if editor is detat…

DavidKarlas:
…ched from IDE into its own window

Problem is that when command comes in, it first goes to CocoaTextViewControl and then up NSView treeview, problem starts when migrating to GTK treeview because GtkNSViewHost is not set as it should be, after debugging why this happens it appears that `GetFocusedChild` method wasn't returning `GtkNSViewHost` as expected, upon further debugging it turns out thats because logic of drilling down GTK treeview doesn't work because `GetFocusedChild` parameter is `NSWindow` instead of `GtkWindow`. So fix is to convert `NSWindow` to `GtkWindow` and now everything works as expected. Reason we are getting `NSWindow` instead of `GtkWindow` is somewhere in `GetActiveWindow` method, but I'm not sure if its bug or not....


DavidKarlas:

sandyarmstrong:

PR #9223, opened 2019-11-05
[Debugger] Make sure colorPreview has a non-null Layer property

jstedfast:
(I hope)

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacDebuggerObjectValueView.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacDebuggerObjectValueView.cs
@@ -67,7 +67,8 @@ namespace MonoDevelop.Debugger
 			};
 
 			colorPreview = new NSView (new CGRect (0, 0, ImageSize, ImageSize)) {
-				TranslatesAutoresizingMaskIntoConstraints = false
+				TranslatesAutoresizingMaskIntoConstraints = false,
+				WantsLayer = true
 			};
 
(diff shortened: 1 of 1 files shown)


jstedfast:

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacDebuggerObjectValueView.cs
```diff
@@ -159,7 +159,7 @@ protected override void UpdateContents ()
 				strval = string.Empty;
 				editable = false;
 			} else {
-				strval = TreeView.Controller.GetDisplayValueWithVisualisers (Node, out showViewerButton);
+				strval = TreeView.Controller.GetDisplayValueWithVisualisers (Node, out showViewerButton) ?? string.Empty;
```
Therzok: Can we add a Debug.Assert on the returned value? Would be nice to know why `Node` is null.

jstedfast: We already know the Node itself isn't null since the top of this method checks that...

sgmunn:
either  node.GetInlineVisualisation () or  node.GetDisplayValue () are returning null it seems.

sgmunn:
@jstedfast or it's this

> colorPreview.Layer.BackgroundColor = GetCGColor (previewColor.Value);

Layer will be null unless you add colorPreview.WantsLayer = true;

jstedfast:
That seems more likely, thanks

jstedfast:

PR #9235, opened 2019-11-06
[Ide] Work around bug in Xamarin.Forms multi project templates

rodrmoya:
The templating engine creates all the outputs correctly, but PrimaryOutputs
contains the files containing the "sourceName" from the template
("NewApp" in the Xamarin.Forms case), not the real names of the files
on disk (NewApp.csproj instead of $ProjectName.csproj).

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1016597


slluis:

mrward:
Is this a bug in the newer templating engine? It looks like it might be since older VS Mac 8.4 versions work. The project templates look OK - they define `"sourceName": "NewApp",` so this should be automatically replaced. Might be worth looking at using maybe a later templating engine? It was updated to a 3.0 version, since the 3.1 versions were pre-release versions.

We should file a bug against the templating engine here too since this used to work.

rodrmoya:
> Is this a bug in the newer templating engine? It looks like it might be since older VS Mac 8.4 versions work. The project templates look OK - they define `"sourceName": "NewApp",` so this should be automatically replaced. Might be worth looking at using maybe a later templating engine? It was updated to a 3.0 version, since the 3.1 versions were pre-release versions.
> 
> We should file a bug against the templating engine here too since this used to work.

yes, it seems to be a bug in the templating engine. Seems the Xamarin.Forms templates are the only multi project templates out there, so the issue is that PrimaryOutputs is returning a list of files with the `sourceName` not replaced with `projectName`. Everything on disk is correct, with the correct names, but not the `PrimaryOutputs` in the template engine API.

PR #9257, opened 2019-11-08
[Core] Support generating a binlog for any MSBuild target

mrward:
Previously a binlog file would only be generated if a Build was run.
Now it is possible to have a binlog generated for any MSBuild target.
The filename for the bin log can be specified when running an MSBuild
target by setting the TargetEvaluationContext's BinLogFilePath.

Note that setting the BinLogFilePath if a build session is in progress
will not have any affect, the build session's binlog overrides this
setting.

Also support advanced scenarios where the running of MSBuild directly
can be intercepted and monitored. This is done by adding a handler
to the MSBuildProcessService which by default will call the
Runtime.ProcessService's StartProcess. This allows binlogs to be generated when MSBuild is run directly not in the build host.

Allows the Project System Tools addin (similar to that available for Visual Studio on Windows) to support showing a binlog for all MSBuild targets:

BuildLogging2

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProcessService.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProcessService.cs
@@ -34,6 +34,12 @@ namespace MonoDevelop.Projects.MSBuild
 {
 	public static class MSBuildProcessService
 	{
+		/// <summary>
+		/// Allows the MSBuild process start to be intercepted and monitored.
+		/// </summary>
+		public static Func<string, string, string, TextWriter, TextWriter, EventHandler, ProcessWrapper>
+			StartProcessHandler { get; set; } = Runtime.ProcessService.StartProcess;
+
 		public static ProcessWrapper StartMSBuild (string arguments, string workingDirectory, TextWriter outWriter, TextWriter errorWriter, EventHandler exited)
 		{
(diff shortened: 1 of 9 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProcessService.cs
```diff
@@ -34,6 +34,12 @@ namespace MonoDevelop.Projects.MSBuild
 {
 	public static class MSBuildProcessService
 	{
+		/// <summary>
+		/// Allows the MSBuild process start to be intercepted and monitored.
+		/// </summary>
+		public static Func<string, string, string, TextWriter, TextWriter, EventHandler, ProcessWrapper>
+			StartProcessHandler { get; set; } = Runtime.ProcessService.StartProcess;
```
Therzok: Can we just declare a delegate type?

mrward: Updated to use a delegate. That is better than using a Func since it is clear what the arguments are with the delegate.

Therzok: Yeah, that was my original problem too, pretty hard to figure out what each argument is.

Therzok (approved):

rodrmoya (approved):

PR #9259, opened 2019-11-08
[Unit Test] Show source code while debugging

ketttY:
1. Update the version of VSCodeDebugProtocol used
2. Update the exception request during `Lauch` and `Attach` to catch the unhandled exception as a "stopped" event
3. In `DebuggingService`, when `NotifyCallStackChanged()` is called when conditions satisfied, subsequent calls will occur: `OnStackChanged()` -> `SetSourceCodeFrame()` -> `NotifyCurrentFrameChanged()`. Therefore adding `noNeedUpdateFrame` to prevent duplicate calls to `NotifyCurrentFrameChanged()`

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1000552

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VSCodeDebuggerSession.cs
+++ b/main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VSCodeDebuggerSession.cs
@@ -156,6 +156,7 @@ namespace MonoDevelop.Debugger.VsCodeDebugProtocol
 		}
 
 		bool currentExceptionState = false;
+		bool unhandleExceptionRegistered = false;
 		void UpdateExceptions ()
 		{
 			//Disposed
@@ -163,11 +164,13 @@ namespace MonoDevelop.Debugger.VsCodeDebugProtocol
 				return;
 
 			var hasCustomExceptions = breakpoints.Select (b => b.Key).OfType<Catchpoint> ().Any (e => e.Enabled);
-			if (currentExceptionState != hasCustomExceptions) {
+			if (currentExceptionState != hasCustomExceptions || !unhandleExceptionRegistered) {
 				currentExceptionState = hasCustomExceptions;
-				protocolClient.SendRequest (new SetExceptionBreakpointsRequest (
-					Capabilities.ExceptionBreakpointFilters.Where (f => hasCustomExceptions || (f.Default ?? false)).Select (f => f.Filter).ToList ()
-				), null);
+				var exceptionRequest = new SetExceptionBreakpointsRequest (
+					Capabilities.ExceptionBreakpointFilters.Where (f => hasCustomExceptions || (f.Default ?? false) || (f.Label == "User-Unhandled Exceptions")).Select (f => f.Filter).ToList ());
+				exceptionRequest.ExceptionOptions = new List<ExceptionOptions> () { new ExceptionOptions (ExceptionBreakMode.Unhandled)};
+				protocolClient.SendRequest (exceptionRequest, null);
+				unhandleExceptionRegistered = true;
 			}
 		}
 
@@ -268,6 +271,7 @@ namespace MonoDevelop.Debugger.VsCodeDebugProtocol
 			LaunchRequest launchRequest = CreateLaunchRequest (startInfo);
 			protocolClient.SendRequestSync (launchRequest);
 			protocolClient.SendRequestSync (new ConfigurationDoneRequest ());
+			UpdateExceptions ();
 		}
 
 		protected void Attach (long processId)
@@ -277,6 +281,7 @@ namespace MonoDevelop.Debugger.VsCodeDebugProtocol
 			protocolClient.SendRequestSync (attachRequest);
 			OnStarted ();
 			protocolClient.SendRequestSync (new ConfigurationDoneRequest ());
+			UpdateExceptions ();
 		}
 
 		protected internal DebugProtocolHost protocolClient;
@@ -403,12 +408,7 @@ namespace MonoDevelop.Debugger.VsCodeDebugProtocol
 							// It's OK to evaluate expressions in external code
 							stackFrame = (VsCodeStackFrame)backtrace.GetFrame (0);
 						}
-
-						if (!breakpoints.Select (b => b.Key).OfType<Catchpoint> ().Any (c => ShouldStopOnExceptionCatchpoint (c, stackFrame.frameId))) {
-							OnContinue ();
-							return;
-						}
-						args = new TargetEventArgs (TargetEventType.ExceptionThrown);
+						args = new TargetEventArgs (TargetEventType.UnhandledException);
 						break;
 					default:
 						throw new NotImplementedException (body.Reason.ToString ());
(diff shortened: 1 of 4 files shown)


ketttY:
@monojenkins rebase

ketttY:
@monojenkins rebase

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Initializer.cs
```diff
@@ -67,30 +68,41 @@ void OnStackChanged (object s, EventArgs a)
 		
 		async void OnFrameChanged (object s, EventArgs a)
 		{
+			if (changingFrame) {
+				return;
+			} else {
```
Therzok: Else is redundant.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Initializer.cs
```diff
@@ -67,30 +68,41 @@ void OnStackChanged (object s, EventArgs a)
...
+				changingFrame = true;
+			}
 			if (disassemblyDoc != null && DebuggingService.IsFeatureSupported (DebuggerFeatures.Disassembly))
 				disassemblyView.Update ();
 		
 			var frame = DebuggingService.CurrentFrame;
-			if (frame == null)
+			if (frame == null) {
+				changingFrame = false;
```
Therzok: I would wrap most of the method after the early exit in a try/finally.

```csharp
if (changingFrame)
    return;

try {
    changingFrame = true;
    ...
} finally {
    changingFrame = false;
}
```

We don't want to end up with `changingFrame` being set to true and never being unset due to an exception

ketttY:
@monojenkins rebase

main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VSCodeDebuggerSession.cs
```diff
@@ -156,18 +156,21 @@ protected override BreakEventInfo OnInsertBreakEvent (BreakEvent breakEvent)
...
 
 			var hasCustomExceptions = breakpoints.Select (b => b.Key).OfType<Catchpoint> ().Any (e => e.Enabled);
-			if (currentExceptionState != hasCustomExceptions) {
+			if (currentExceptionState != hasCustomExceptions || !unhandleExceptionRegistered) {
 				currentExceptionState = hasCustomExceptions;
-				protocolClient.SendRequest (new SetExceptionBreakpointsRequest (
-					Capabilities.ExceptionBreakpointFilters.Where (f => hasCustomExceptions || (f.Default ?? false)).Select (f => f.Filter).ToList ()
-				), null);
+				var exceptionRequest = new SetExceptionBreakpointsRequest (
+					Capabilities.ExceptionBreakpointFilters.Where (f => hasCustomExceptions || (f.Default ?? false) || (f.Label == "User-Unhandled Exceptions")).Select (f => f.Filter).ToList ());
```
jstedfast: Can we be sure that the filter's Label will be `"User-Unhandled Exceptions"`? In other words, can we be sure that this label string will never be translated to other languages (e.g. using `GettextCatalog.GetString()`)

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/Initializer.cs
```diff
@@ -67,108 +68,119 @@ void OnStackChanged (object s, EventArgs a)
...
-			if (disassemblyDoc != null && DebuggingService.IsFeatureSupported (DebuggerFeatures.Disassembly))
-				disassemblyView.Update ();
-		
-			var frame = DebuggingService.CurrentFrame;
-			if (frame == null)
+			if (changingFrame) {
 				return;
-			
-			FilePath file = frame.SourceLocation.FileName;
+			} try {
```
jstedfast: The `try` should start on a new line

ketttY:
@monojenkins rebase

mrward:
This fixes the problem where the unhandled exception popup window was not being displayed for a .NET Core console project with an unhandled exception, which is very nice.

However test discovery seems to be broken for me for an NUnit .NET Core and an xUnit .NET Core project. No tests are shown in the Unit Tests window. Test discovery seems OK for me with VS Mac 8.4. I have .NET Core 3.1, 3.0 and 2.1 installed. Just wondering if this is related to the VS Test update made recently?

The .NET Core debugger tests are failing - not sure these are related to this change or not.

ketttY:
@monojenkins rebase

mrward:
May need this other change approved/merged to master which downgrades the VSTest NuGet package to workaround problems with the test discovery not working on the Mac - https://github.com/mono/monodevelop/pull/9417

ketttY:
@monojenkins rebase

ketttY:
@monojenkins rebase

ketttY:
@monojenkins rebase

ketttY:
@monojenkins rebase

ketttY:
@monojenkins rebase

ketttY:
@monojenkins rebase

main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VSCodeDebuggerSession.cs
```diff
@@ -156,18 +156,21 @@ protected override BreakEventInfo OnInsertBreakEvent (BreakEvent breakEvent)
...
 
 			var hasCustomExceptions = breakpoints.Select (b => b.Key).OfType<Catchpoint> ().Any (e => e.Enabled);
-			if (currentExceptionState != hasCustomExceptions) {
+			if (currentExceptionState != hasCustomExceptions || !unhandleExceptionRegistered) {
 				currentExceptionState = hasCustomExceptions;
-				protocolClient.SendRequest (new SetExceptionBreakpointsRequest (
-					Capabilities.ExceptionBreakpointFilters.Where (f => hasCustomExceptions || (f.Default ?? false)).Select (f => f.Filter).ToList ()
-				), null);
+				var exceptionRequest = new SetExceptionBreakpointsRequest (
+					Capabilities.ExceptionBreakpointFilters.Where (f => hasCustomExceptions || (f.Default ?? false) || (f.Label == GettextCatalog.GetString("User-Unhandled Exceptions"))).Select (f => f.Filter).ToList ());
```
mrward: Not sure this is going to work here. I am guessing this string is from the .NET Core debugger. If so we should see if we can find out if this is translated, and if so we should be using the string resources provided with the debugger.

The problem with using GettextCatalog.GetString is that it requires the translation team to use the same text as the .NET Core debugger uses. They will not know to do that and will just translate it as they see and would be unlikely to match presumably.

main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VSCodeDebuggerSession.cs
```diff
@@ -403,12 +408,15 @@ protected void HandleEvent (object sender, EventReceivedEventArgs obj)
 							// It's OK to evaluate expressions in external code
 							stackFrame = (VsCodeStackFrame)backtrace.GetFrame (0);
 						}
-
-						if (!breakpoints.Select (b => b.Key).OfType<Catchpoint> ().Any (c => ShouldStopOnExceptionCatchpoint (c, stackFrame.frameId))) {
-							OnContinue ();
-							return;
+						if (body.Text.Contains (GettextCatalog.GetString ("but was not handled in user code"))){
+							args = new TargetEventArgs (TargetEventType.UnhandledException);
```
mrward: Same thing here with this translation.

PR #9260, opened 2019-11-08
[CSharpBinding] Rely on MSBuild's OutputType for determining Roslyn's OutputKind

garuma:
- Fixes AB#1018498
- Context: http://sourceroslyn.io/#Microsoft.CodeAnalysis.CSharp/CommandLine/CSharpCommandLineParser.cs,f777bd5ac8b781b5,references
- Context: https://docs.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-properties?view=vs-2019

The previous strategy used a project model extension mechanism which was not accurate in a couple of scenarios like plain .NET/.NET standard library project where it would erroneously results in `OutputKind.ConsoleApplication` being used.

This new logic instead leverages MSBuild directly in the same fashion than other implementations of Roslyn's workspaces do to get the most accurate information.

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CSharpCompilerParameters.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CSharpCompilerParameters.cs
@@ -84,6 +84,8 @@ namespace MonoDevelop.CSharp.Project
 		[ItemProperty ("Nullable", DefaultValue = "")]
 		string nullableContextOptions = "";
 
+		string outputType;
+
 		protected override void Write (IPropertySet pset)
 		{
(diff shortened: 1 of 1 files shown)


main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CSharpCompilerParameters.cs
```diff
@@ -84,6 +84,8 @@ public class CSharpCompilerParameters : DotNetCompilerParameters
 		[ItemProperty ("Nullable", DefaultValue = "")]
 		string nullableContextOptions = "";
 
+		string outputType;
```
Therzok: Shiuld this not be an item property?

garuma: I followed the strategy of `optimize` and `warningLevel` thinking that maybe `ItemProperty` only works on non-config specific properties? (i.e. evaluation is done without taking active configuration into account)

PR #9267, opened 2019-11-11
Adds general KeyViewLoop custom implementation in CommandManager

netonjm:
This PR adds a new mechanism to create our own keyloop system to sync Gtk and Cocoa world focus chain.

xRCBhpk7dk

zXN2k3hSf7-1

To get some context about the problem, under the hood the native views are overlapped on top of Gtk and synchronized with a Gtk view container host (GtkNSViewHost)

In VS4Mac is the Gtk application is who captures the keypress events and the CommandManager ensures if the selected view is a native view it tunnels the event to cocoa world (making stop the escalation of the event in the Gtk) or otherwise it continues the gtk event flow.

Actually we were doing this tunneling to cocoa in an incomplete way, we were only calling to ".OnKeyDown" from the focused view, which is incomplete, since this is a KeyDown flow that starts from the NSWindow and ends into the view. This problem ends affecting to navigate pressing tab key (Keyloop) mechanism.... which sometimes Tab works and sometimes not depending in the View behaviour.

Right now our GtkQuarzWindow hack has some pain points: all the native embeded views shares the same NSWindow (GtkQuarzWindow) using this structure:

- GtkQuarzWindow
   - GtkQuarzView
     - Embeded widgets (ProppyPad, ToolboxPad, Source Editor Pad)

When the NextKeyView is not set in a native view, the NSWindow autocalculates what's is going to be the next or previous candidate element, but this doesn't work in VS4Mac context, and that's because we are under a Gtk Application and with a cocoa hack.

Another problem we had was how connect the Gtk focus chain to Cocoa, that's means we needed to handle cases when user press tab and next view is a GtkNSViewHost then calculate which is our first candidate for item focus.. and ... the same for pressing SHIFT+Tab (focus last item from this embedded view).

To try to fix this, I checked a lot of documentation, but it seems there are not much information about how this works under the hood or how replicate it... and after a lot of tests didn't found any good way to do it more easily. That's because I decided to rewrite current implementation to create our own mechanism to calculate this keyloop calculation.

And.. sorry about the hack.. but didn't found any idea about how do it in a better way living into a Gtk application :-(

Fixes #1001596 - A11Y_Xamarin Designers_Property pane_keyboard : User is unable to navigate from property filter to the next controls inside property panel window
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1001596

Fixes VSTS #1007912 - Keyboard focus isn't given to Solution pad when "Solution Pad" command is invoked
Fixes VSTS #1016070 - Solution Pad not getting focus, when ran via PadActivationHandler

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -442,7 +442,7 @@ namespace MonoDevelop.Components.Commands
 		void SimulateKeyDownInView (AppKit.NSView view, AppKit.NSEvent currentEvent, AppKit.NSWindow window)
 		{
 			if (currentEvent.KeyCode == (ushort)AppKit.NSKey.Tab) {
-				var expectedKeyView = FindValidKeyView (view);
+				var expectedKeyView = FindValidKeyView (view, defaultView: view);
 				AppKit.NSView next = null;
 				if (currentEvent.ModifierFlags.HasFlag (AppKit.NSEventModifierMask.ShiftKeyMask)) {
 					next = expectedKeyView.PreviousValidKeyView;
@@ -466,19 +466,22 @@ namespace MonoDevelop.Components.Commands
 			}
 		}
 
-		static AppKit.NSView FindValidKeyView (AppKit.NSView view)
+		static AppKit.NSView FindValidKeyView (AppKit.NSView view, AppKit.NSView defaultView)
 		{
 			if (view == null)
 				return null;
 
 			if (view.AcceptsFirstResponder ()) {
 				if (view.Superview?.Superview is AppKit.NSControl control && control.CurrentEditor == view) {
-					return FindValidKeyView (control);
+					return FindValidKeyView (control, defaultView);
 				}
 				return view;
 			}
 
-			return FindValidKeyView (view.Superview);
+			if (IsGtkQuartzView (view.Superview )) {
+				return defaultView;
+			}
+			return FindValidKeyView (view.Superview, defaultView);
 		}
 
 #endif
@@ -573,7 +576,7 @@ namespace MonoDevelop.Components.Commands
 				var gtkNSViewHost = GetGtkNSViewHostFromView (mainView);
 				if (gtkNSViewHost != null) {
 					//we are in a embeded view case
-					var expectedKeyView = FindValidKeyView (view);
+					var expectedKeyView = FindValidKeyView (view, defaultView: view);
 
 					if (currentEvent.KeyCode == (ushort)AppKit.NSKey.Tab) {
 						if (currentEvent.ModifierFlags.HasFlag (AppKit.NSEventModifierMask.ShiftKeyMask)) {
@@ -679,12 +682,15 @@ namespace MonoDevelop.Components.Commands
 					return null;
 				}
 			}
-			if (nSView.Superview.ToString ().StartsWith ("<GdkQuartzView")) {
+			if (IsGtkQuartzView (nSView.Superview)) {
 				return nSView;
 			}
 			var parent = GetGtkNSViewHostContentView (nSView.Superview);
 			return parent;
 		}
+
+		const string GdkQuartzViewTypeName = "<GdkQuartzView";
+		static bool IsGtkQuartzView (AppKit.NSView view) => view.ToString ().StartsWith (GdkQuartzViewTypeName);
 
 		readonly List<Gtk.GtkNSViewHost> viewHosts = new List<Gtk.GtkNSViewHost> ();
 		internal void RegisterEmbededView (Gtk.GtkNSViewHost gtkNSViewHost)
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -674,8 +674,11 @@ namespace MonoDevelop.Components.Commands
 			return null;
 		}
 
-		bool IsLastView (AppKit.NSView view, bool addViewBeforeChildren, bool removeContentView) => GetOrderedFocusableViews (view, addViewBeforeChildren, removeContentView).LastOrDefault () == view;
-		bool IsFirstView (AppKit.NSView view, bool addViewBeforeChildren, bool removeContentView) => GetOrderedFocusableViews (view, addViewBeforeChildren, removeContentView).FirstOrDefault () == view;
+		bool IsLastView (AppKit.NSView view, bool addViewBeforeChildren, bool removeContentView)
+			=> GetOrderedFocusableViews (view, addViewBeforeChildren, removeContentView).LastOrDefault () == view;
+
+		bool IsFirstView (AppKit.NSView view, bool addViewBeforeChildren, bool removeContentView)
+			=> GetOrderedFocusableViews (view, addViewBeforeChildren, removeContentView).FirstOrDefault () == view;
 
 		GtkNSViewHost GetGtkNSViewHostFromView (AppKit.NSView view) => view == null ? null : viewHosts.FirstOrDefault (s => s.Content == view);
 
@@ -723,14 +726,14 @@ namespace MonoDevelop.Components.Commands
 			if (addViewBeforeChildren && CanBecomeKeyView (view, contentView as IGtkViewHostContentView) && contentView != view)
 				toAddViews.Add (view);
 
-			if (view.IsFlipped) {
-				foreach (var subview in view.Subviews.OrderBy (s => s.Frame.Left).ThenBy (h => h.Frame.Top)) {
-					AddRecursivelyFocusableViews (subview, toAddViews, addViewBeforeChildren, contentView);
-				}
-			} else {
-				foreach (var subview in view.Subviews.OrderBy (s => s.Frame.Left).ThenByDescending (h => h.Frame.Top)) {
-					AddRecursivelyFocusableViews (subview, toAddViews, addViewBeforeChildren, contentView);
-				}
+			var possibleChildren = view.Subviews.OrderBy (s => s.Frame.Left);
+			if (view.IsFlipped)
+				possibleChildren = possibleChildren.ThenBy (h => h.Frame.Top);
+			else
+				possibleChildren = possibleChildren.ThenByDescending (h => h.Frame.Top);
+
+			foreach (var subview in possibleChildren) {
+				AddRecursivelyFocusableViews (subview, toAddViews, addViewBeforeChildren, contentView);
 			}
 
 			if (!addViewBeforeChildren && CanBecomeKeyView (view, contentView as IGtkViewHostContentView) && contentView != view)
@@ -758,7 +761,7 @@ namespace MonoDevelop.Components.Commands
 		internal void RegisterEmbededView (Gtk.GtkNSViewHost gtkNSViewHost)
 		{
 			if (gtkNSViewHost == null) {
-				throw new NullReferenceException ("cannot register a null GtkNSViewHost");
+				throw new ArgumentNullException ("cannot register a null GtkNSViewHost");
 			}
 			if (!viewHosts.Contains (gtkNSViewHost)) {
 				viewHosts.Add (gtkNSViewHost);
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -642,7 +642,7 @@ namespace MonoDevelop.Components.Commands
 			return true;
 		}
 
-		static bool HandlesFocus (AppKit.NSView view)
+		static bool AllowsRecursivityFocus (AppKit.NSView view)
 		{
 			if (view.Hidden) {
 				return true;
@@ -656,7 +656,7 @@ namespace MonoDevelop.Components.Commands
 
 		static void AddRecursivelyFocusableViews (AppKit.NSView view, List<AppKit.NSView> toAddViews, bool addViewBeforeChildren, AppKit.NSView contentView)
 		{
-			if (HandlesFocus (view)) {
+			if (AllowsRecursivityFocus (view)) {
 				return;
 			}
 
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
@@ -439,15 +439,33 @@ namespace MonoDevelop.Components.Commands
 			return false;
 		}
 
-		void SimulateKeyDownInView (AppKit.NSView view, AppKit.NSEvent currentEvent, AppKit.NSWindow window)
+		void SimulateKeyDownInView (AppKit.NSView view, GtkNSViewHost currentGtkViewHost, AppKit.NSEvent currentEvent, AppKit.NSWindow window)
 		{
 			if (currentEvent.KeyCode == (ushort)AppKit.NSKey.Tab) {
 				var expectedKeyView = FindValidKeyView (view, defaultView: view);
 				AppKit.NSView next = null;
 				if (currentEvent.ModifierFlags.HasFlag (AppKit.NSEventModifierMask.ShiftKeyMask)) {
 					next = expectedKeyView.PreviousValidKeyView;
+
+					//is our next from another embeded view? we don't allow to do that and we 
+					//we calculate what's the next keyview if there is no a valid candidate (null case) or if our
+					//previous is from another embeded view
+					bool isValidFocus = next != null
+					&& currentGtkViewHost.Content == GetGtkNSViewHostContentView (next) //is in our embeded view
+					&& CanBecomeKeyView (next);
+					if (!isValidFocus) {
+						next = GetPreviousKeyView (expectedKeyView, addViewBeforeChildren: false, removeContentView: true);
+					}
+
 				} else {
 					next = expectedKeyView.NextValidKeyView;
+					bool isValidFocus = next != null
+						&& currentGtkViewHost.Content == GetGtkNSViewHostContentView (next) //is in our embeded view
+						&& CanBecomeKeyView (next);
+					if (!isValidFocus) {
+						//we calculate what's the next keyview if there is no a valid candidate (null case)
+						next = GetNextKeyView (expectedKeyView, addViewBeforeChildren: true, removeContentView: true);
+					}
 				}
 
 				view.KeyDown (currentEvent);
@@ -586,7 +604,7 @@ namespace MonoDevelop.Components.Commands
 								backToGtk = true;
 							}
 						} else {
-							var isLastView = IsLastView (expectedKeyView, addViewBeforeChildren: false, removeContentView: true);
+							var isLastView = IsLastView (expectedKeyView, addViewBeforeChildren: true, removeContentView: true);
 							if (isLastView) {
 								//if we are in last element we change to next Gtk widget
 								backToGtk = true;
@@ -595,16 +613,16 @@ namespace MonoDevelop.Components.Commands
 					}
 
 					if (!backToGtk) {
-						SimulateKeyDownInView (view, currentEvent, window);
+						SimulateKeyDownInView (view, gtkNSViewHost, currentEvent, window);
 					} else {
 
 						if (!gtkNSViewHost.IsFocus)
 							gtkNSViewHost.IsFocus = true;
 
 						//Unfocus all focused views in the Gtkquartzwindow
+						//NOTE: Be careful! after this, native window.FirstResponder it will change from the current one
 						window.MakeFirstResponder (null);
-						//Be careful! after this, native window.FirstResponder it will change from the current one
-						
+					
 						//gtkNSViewHost.ProcessEvent (ev);
 						//GetGtkWindow (ev.Window).Present ();
 					}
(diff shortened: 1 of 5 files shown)


netonjm:
@sandyarmstrong in this PR I did some changes related to the focus and I think it fixed https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1020127 for free, can you please test the current build to double check if also fixes for you?

sandyarmstrong:
@netonjm yes, this does fix the toolbox input/focus issue, as far as I can tell. I need to do more review and testing, but one regression I noticed is that the "Focus Document" command no longer works (editor does not get focus).

sevoku:
@netonjm this seems to fix the toolbox flickering and focus to/from native widgets lgtm too now! However I can confirm the "Focus Document" issue @sandyarmstrong mentioned.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -438,46 +439,78 @@ bool IsCommandBinding (object commandId, string binding)
...
+					//we calculate what's the next keyview if there is no a valid candidate (null case) or if our
+					//previous is from another embeded view
+					bool isValidFocus = next != null
+					&& currentGtkViewHost.Content == GetGtkNSViewHostContentView (next) //is in our embeded view
+					&& CanBecomeKeyView (next);
+					if (!isValidFocus) {
+						next = GetPreviousKeyView (expectedKeyView, addViewBeforeChildren: false, removeContentView: true);
+					}
+
 				} else {
```
sevoku: shouldn't we do the else case only if no other `ModifierFlags` is set? Right now this would trigger also for Alt, Ctrl, etc.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -438,46 +439,78 @@ bool IsCommandBinding (object commandId, string binding)
...
+					if (!isValidFocus) {
+						//we calculate what's the next keyview if there is no a valid candidate (null case)
+						next = GetNextKeyView (expectedKeyView, addViewBeforeChildren: true, removeContentView: true);
+					}
 				}
 
 				view.KeyDown (currentEvent);
-				if (next != null && window?.FirstResponder != next) {
+
+				if (supportsTabKey && next != null && window?.FirstResponder != next) {
```
sevoku: Same here, what happens for other meta keys? Shouldn't we make sure to handle only Tab or Shift+Tab?

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -438,46 +439,78 @@ bool IsCommandBinding (object commandId, string binding)
...
+			if (supportsTabKey && currentEvent.KeyCode == (ushort)AppKit.NSKey.Tab) {
 				AppKit.NSView next = null;
 				if (currentEvent.ModifierFlags.HasFlag (AppKit.NSEventModifierMask.ShiftKeyMask)) {
 					next = expectedKeyView.PreviousValidKeyView;
+
+					//is our next from another embeded view? we don't allow to do that and we 
+					//we calculate what's the next keyview if there is no a valid candidate (null case) or if our
+					//previous is from another embeded view
+					bool isValidFocus = next != null
+					&& currentGtkViewHost.Content == GetGtkNSViewHostContentView (next) //is in our embeded view
```
sevoku: indent here, to indicate that it belongs to the declaration

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -438,46 +439,78 @@ bool IsCommandBinding (object commandId, string binding)
...
+				if (supportsTabKey && next != null && window?.FirstResponder != next) {
 					window.MakeFirstResponder (next);
 				}
 			} else {
 				view.KeyDown (currentEvent);
 			}
 		}
 
+		//cases from views where focus is not handled by IDe
+		bool HandlesMakeFirstResponder (AppKit.NSView view)
```
sevoku:
```suggestion
		static bool HandlesMakeFirstResponder (AppKit.NSView view)
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -567,14 +599,191 @@ internal bool ProcessKeyEvent (Gdk.EventKey ev)
...
+					//return true brokes the current event handling from Gtk, then if we back to gtk we should return false
+					return !backToGtk;
+				}
 			}
 #endif
 			return false;
 		}
 
+		const string sourceEditorTypeClass = ("Microsoft.VisualStudio.Text.Editor.Implementation.CocoaTextViewControl");
+		bool IsSourceEditor (AppKit.NSView view) => view.GetType ().FullName == sourceEditorTypeClass;
```
sevoku:
```suggestion
		static bool IsSourceEditor (AppKit.NSView view) => view.GetType ().FullName == sourceEditorTypeClass;
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/GtkNSViewHost.cs
```diff
@@ -315,7 +321,20 @@ protected override bool OnFocusInEvent (Gdk.EventFocus evnt)
 					return false;
 				}
 
-				acceptsFirstResponderView.Window.MakeFirstResponder (acceptsFirstResponderView);
+				AppKit.NSView nativeViewToFocus = null;
+				var currentEvent = AppKit.NSApplication.SharedApplication?.CurrentEvent;
+				if (currentEvent.Type == NSEventType.KeyDown && currentEvent.KeyCode == (ushort)AppKit.NSKey.Tab) {
+					if (currentEvent.ModifierFlags.HasFlag (AppKit.NSEventModifierMask.ShiftKeyMask)) {
+						nativeViewToFocus = MonoDevelop.Ide.IdeApp.CommandService.GetOrderedFocusableViews (Content, addViewBeforeChildren:false, removeContentView: true).LastOrDefault ();
+					} else {
```
sevoku: check other ModifierFlags here

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/GtkNSViewHost.cs
```diff
@@ -329,12 +348,9 @@ protected override bool OnFocusOutEvent (Gdk.EventFocus evnt)
 		{
 			LogEnter ();
 			try {
-				if (view?.Window?.FirstResponder is NSView firstResponder &&
-					view?.AncestorSharedWithView (firstResponder) == view)
-					firstResponder.Window?.MakeFirstResponder (firstResponder.Window.ContentView);
 				return base.OnFocusOutEvent (evnt);
-			} finally {
-				LogExit ();
+			} finally {
```
sevoku: remove tabs

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
netonjm: yep, you are right

netonjm: fixed

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
netonjm: Fixed

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
netonjm: fixed

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
netonjm: fixed

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/GtkNSViewHost.cs
netonjm: fixed

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/Mac/GtkNSViewHost.cs
netonjm: fixed

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -567,14 +591,182 @@ internal bool ProcessKeyEvent (Gdk.EventKey ev)
...
+					//return true brokes the current event handling from Gtk, then if we back to gtk we should return false
+					return !backToGtk;
+				}
 			}
 #endif
 			return false;
 		}
 
+		const string sourceEditorTypeClass = ("Microsoft.VisualStudio.Text.Editor.Implementation.CocoaTextViewControl");
+		static bool IsSourceEditor (AppKit.NSView view) => view.GetType ().FullName == sourceEditorTypeClass;
```
Therzok: Icky, but not sure whether we can find a better way. Maybe check for `ITextView3Container`, but that won't guarantee it's the actual hosted editor, rather than any implementation.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -567,14 +591,182 @@ internal bool ProcessKeyEvent (Gdk.EventKey ev)
...
+		{
+			var mainView = GetGtkNSViewHostContentView (view);
+			var focusableViews = new List<AppKit.NSView> ();
+			AddRecursivelyFocusableViews (mainView, focusableViews, addViewBeforeChildren, removeContentView ? mainView : null);
+			return focusableViews;
+		}
+
+		AppKit.NSView GetNextKeyView (AppKit.NSView view, bool addViewBeforeChildren, bool removeContentView)
+		{
+			var views = GetOrderedFocusableViews (view, addViewBeforeChildren, removeContentView);
```
Therzok: Can we make the recursive search stop when it's done processing children of our view? I imagine that a view that is early in the UI hierarchy could have an early exit.

This could be optimized, I'm okay with keeping this in until we prove it's problematic - it's just on focus change anyway.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -567,14 +591,182 @@ internal bool ProcessKeyEvent (Gdk.EventKey ev)
...
+			var currentIndex = views.IndexOf (view);
+			if (currentIndex > 0 && currentIndex < views.Count) {
+				return views [currentIndex - 1];
+			}
+			return null;
+		}
+
+		bool IsLastView (AppKit.NSView view, bool addViewBeforeChildren, bool removeContentView) => GetOrderedFocusableViews (view, addViewBeforeChildren, removeContentView).LastOrDefault () == view;
```
Therzok: newline before implementation.
```suggestion
		bool IsLastView (AppKit.NSView view, bool addViewBeforeChildren, bool removeContentView)
			=> GetOrderedFocusableViews (view, addViewBeforeChildren, removeContentView).LastOrDefault () == view;
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -567,14 +591,182 @@ internal bool ProcessKeyEvent (Gdk.EventKey ev)
...
+		}
+
+		const string GdkQuartzViewTypeName = "<GdkQuartzView";
+		static bool IsGtkQuartzView (AppKit.NSView view) => view.ToString ().StartsWith (GdkQuartzViewTypeName);
+
+		readonly List<Gtk.GtkNSViewHost> viewHosts = new List<Gtk.GtkNSViewHost> ();
+		internal void RegisterEmbededView (Gtk.GtkNSViewHost gtkNSViewHost)
+		{
+			if (gtkNSViewHost == null) {
+				throw new NullReferenceException ("cannot register a null GtkNSViewHost");
```
Therzok:
```suggestion
				throw new ArgumentNullException ("cannot register a null GtkNSViewHost");
```

`NullReferenceException` is usually thrown by the runtime to signal null access, and we're not doing a null access here, we're signaling API mis-use.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -567,14 +591,182 @@ internal bool ProcessKeyEvent (Gdk.EventKey ev)
...
+			}
+			if (!viewHosts.Contains (gtkNSViewHost)) {
+				viewHosts.Add (gtkNSViewHost);
+			}
+		}
+
+		internal void RemoveEmbededView (GtkNSViewHost gtkNSViewHost) => viewHosts.Remove (gtkNSViewHost);
```
Therzok: RegisterEmbeddedView does argument null checks. We should be consistent.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -567,14 +591,182 @@ internal bool ProcessKeyEvent (Gdk.EventKey ev)
...
+		static void AddRecursivelyFocusableViews (AppKit.NSView view, List<AppKit.NSView> toAddViews, bool addViewBeforeChildren, AppKit.NSView contentView)
+		{
+			if (AllowsRecursivityFocus (view)) {
+				return;
+			}
+
+			if (addViewBeforeChildren && CanBecomeKeyView (view, contentView as IGtkViewHostContentView) && contentView != view)
+				toAddViews.Add (view);
+
+			if (view.IsFlipped) {
```
Therzok: I think we can deduplicate the loops:
```csharp
var possibleChildren = views.Subviews.OrderBy (s => s.Frame.Left);
if (view.IsFlipped)
    possibleChildren = possibleChildren.ThenBy (h => h.Frame.Top);
else
    possibleChildren = possibleChildren.ThenDescendingBy (h => h.Frame.Top);

foreach (var subview in possibleChildren) {
    ddRecursivelyFocusableViews (subview, toAddViews, addViewBeforeChildren, contentView);
}
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
sandyarmstrong: `ITextView*Container` API isn't public either, though. I think this check is fine for now.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
netonjm: fixed

sevoku:

sevoku:
I can reproduce the "Focus Document" command issue, it might be focusing the outer container maybe? We could probably fix that in the command handler...
Besides that everything seems to be working as it should!

sevoku:

netonjm:
Please, for all the other people involved, approve the PR as a successful verification sign. Thank you!

mrward:
See some odd tabbing behaviour when the Properties pad is docked (OK when it is not docked). Tabbing seems to leave the Properties window and go to the pad toolbar and then around the main status bar. Was expecting it to remain in the Properties pad.

Toolbox tabbing seems to not work when shift+tabbing. Not seeing focus on the text snippets when forward tabbing. Shift tabbing does not seem to get back to the top part of the pad.

sevoku:
> See some odd tabbing behaviour when the Properties pad is docked (OK when it is not docked). Tabbing seems to leave the Properties window and go to the pad toolbar and then around the main status bar. Was expecting it to remain in the Properties pad.

This is expected and was the same without this change. However we should improve that and don't let the focus leave the pad at all (any pad), same behaviour as Xcode.

> Toolbox tabbing seems to not work when shift+tabbing. Not seeing focus on the text snippets when forward tabbing. Shift tabbing does not seem to get back to the top part of the pad. (Note same behaviour in VS Mac 8.5 so not due to the change here).

This might be an issue with NSOutlineView maybe and seems to be the same on master, so not a regression. Focus into the table will make the focus appear once you press up/down. But focusing back doesn't work at all.

garuma:
@sevoku is the intention to backport this to 8.4 as well? That seems like a big change to do on a .3 release

sevoku:
@garuma yes, we need to fix the focus chain in 8.4, the bug has a 60 days SLA so we need to service it.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -465,19 +487,22 @@ private void SimulateViewKeyActionBehaviour (AppKit.NSView view, AppKit.NSEvent
...
 			if (view.AcceptsFirstResponder ()) {
 				if (view.Superview?.Superview is AppKit.NSControl control && control.CurrentEditor == view) {
-					return FindValidKeyView (control);
+					return FindValidKeyView (control, defaultView);
 				}
 				return view;
 			}
 
-			return FindValidKeyView (view.Superview);
+			if (IsGtkQuartzView (view.Superview )) {
```
Therzok:
```suggestion
			if (IsGtkQuartzView (view.Superview)) {
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -567,14 +591,191 @@ internal bool ProcessKeyEvent (Gdk.EventKey ev)
...
+					if (!backToGtk) {
+						SimulateKeyDownInView (view, expectedKeyView, gtkNSViewHost, currentEvent, window);
+					} else {
+
+						if (!gtkNSViewHost.IsFocus)
+							gtkNSViewHost.IsFocus = true;
+
+						//Unfocus all focused views in the Gtkquartzwindow
+						//NOTE: Be careful! after this, native window.FirstResponder it will change from the current one
+						window.MakeFirstResponder (null);
```
Therzok: window can be null, see https://github.com/mono/monodevelop/pull/9267/files#diff-b09b2588c432fb3d2d826347c2dae4dcR561

Therzok: Actually, even the `FlagsChanged` handling seems to not care about window being null.

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.Commands/CommandManager.cs
```diff
@@ -567,14 +591,191 @@ internal bool ProcessKeyEvent (Gdk.EventKey ev)
...
+					
+						//gtkNSViewHost.ProcessEvent (ev);
```
Therzok: Commented code

garuma:
I tested this with the Android designer and there are a few issues I could find:
- Our toolbar Tab/Shift-tab behavior works fine initially but if you go focus the global search once for instance then only Shift-Tab works (Tab jumps somewhere else). That's also reproducible in the master build
- With this build, our tabbing behavior inside the surface (moving between various UI elements of the Android screen) is broken

BretJohnson:
@sevoku Given the rule that we're only requiring VO based keyboard navigation on Mac, not tab key, for MAS-C, isn't that sufficient justification from a process perspective for us to not have to take this to 8.4? That & the potential risk. Agreed on taking it to 8.5, though.

sevoku:
@BretJohnson the MAS-C exception with tabbing was true for reaching MAS-C formally. However we're already out of the exception for the product not being MAS-C (and therefore WCAG) compliant by End of 2019, we already extended it to 8.4 stable, but I'm not sure if we can extend it to 8.5 (more then 3 months too late). Additionally, all a11y bugs have a 60 day SLA, this was not an issue while we had the formal exception (the one until end of 2019), but now the SLA is real, so no matter whether we're compliant now or not, we've got still the SLA.

BretJohnson:
@sevoku Assuming we are all still in agreement that tabbing isn't required for MAS-C on Mac, then I would think we could follow the process you mentioned before, where we say  “It is accessible with VO. Please reopen issues with keyboard tabbing with the A11yUsable tag” & Alexandria (with Dante's approval) opens a new issue with the A11yUsable tag. Those issues are just needed for MAS-B and shouldn't count against the SLA.

That said, we should do whatever the "right thing" here is, like Matt suggested. If risk is high enough, just take to 8.5. If the usability gains are enough making tab nav work (for able bodied as well as disabled users), take to 8.4. Which is all a way of saying, I'm not sure :-) . But from the MAS-C compliance process perspective, I think we have an out.

BretJohnson:
I did more testing here.
The two issues Jeremie mentioned earlier are fixed now: https://github.com/mono/monodevelop/pull/9267#issuecomment-577218505

But I spotted a few other things:

1. I still see this issue: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1001596
where tab / shift-tab from the proppy Property Filter edit box does nothing.

2. After hitting tab in global search, the focus appears to go to solution explorer, as shown in the screenshot below. But the focus must actually go to some other unseen window, as hitting keys (e.g. tab or arrow) at that point don't do anything. And note that Packages below is grey, not blue as when it has focus.

3. When starting  from global search and hitting shift-tab a few times, past the run button, the editor then has focus and starts inserting tabs into the buffer. That may be by design though.

Beyond that, things looked good to me, in our designers and in the IDE UI I tested.

image

netonjm:
closed in favour of https://github.com/xamarin/monodevelop/pull/762

PR #9270, opened 2019-11-11
[Debugger] Optimize debugger tooltip resizing

jstedfast:
Fixes VSTS 1002713

Fixes issues with the tooltip sizing on catalina. Speeds up the calculation of the column widths and fixes a NRE when expanding using the "show values", "more values" buttons.

this also fixes VSTS 1011400

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacDebuggerObjectCellViewBase.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacDebuggerObjectCellViewBase.cs
@@ -158,6 +158,22 @@ namespace MonoDevelop.Debugger
 			return new NSAttributedString (text ?? string.Empty, strokeColor: NSColor.PlaceholderTextColor);
 		}
 
+		protected static nfloat GetWidthForString (NSFont font, string text, int sizeDelta = 0)
+		{
+			NSFont modified = null;
+			nfloat width;
+
+			if (sizeDelta != 0)
+				modified = NSFont.FromDescription (font.FontDescriptor, font.PointSize + sizeDelta);
+
+			using (var str = new NSAttributedString (text, font: modified ?? font))
+				width = str.Size.Width;
+
+			modified?.Dispose ();
+
+			return width;
+		}
+
 		protected void UpdateFont (NSControl control, int sizeDelta = 0)
 		{
 			var font = TreeView.CustomFont;
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacDebuggerObjectCellViewBase.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacDebuggerObjectCellViewBase.cs
@@ -171,7 +171,7 @@ namespace MonoDevelop.Debugger
 
 			modified?.Dispose ();
 
-			return width;
+			return NMath.Ceiling (width + 2);
 		}
 
 		protected void UpdateFont (NSControl control, int sizeDelta = 0)
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacDebuggerObjectNameView.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacDebuggerObjectNameView.cs
@@ -108,6 +108,7 @@ namespace MonoDevelop.Debugger
 			var editable = TreeView.AllowWatchExpressions && Node.Parent is RootObjectValueNode;
 			var selected = Superview is NSTableRowView rowView && rowView.Selected;
 			var iconName = ObjectValueTreeViewController.GetIcon (Node.Flags);
+			var wrapper = (MacObjectValueNode) ObjectValue;
 			var textColor = NSColor.ControlText;
 			var showAddNewExpression = false;
 			var placeholder = string.Empty;
@@ -165,9 +166,9 @@ namespace MonoDevelop.Debugger
 			TextField.TextColor = textColor;
 			TextField.Editable = editable;
 			UpdateFont (TextField);
-			TextField.SizeToFit ();
 
-			OptimalWidth += TextField.Frame.Width;
+			var value = editable && string.IsNullOrEmpty (name) ? placeholder : name;
+			OptimalWidth += GetWidthForString (TextField.Font, value);
 
 			constraints.Add (TextField.CenterYAnchor.ConstraintEqualToAnchor (CenterYAnchor));
 			constraints.Add (TextField.LeadingAnchor.ConstraintEqualToAnchor (firstView.TrailingAnchor, RowCellSpacing));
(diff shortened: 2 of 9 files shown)


jstedfast:
Pretty sure I can kill the override of `ViewDidEndLiveResize` because live resizing is not something that can be done with tooltips or pinned watches, so having that override doesn't make sense to have.

`ViewDidMoveToSuperview` can probably also be dropped since it only gets called when we pack the `MacObjectValueTreeView` into the parent `NSScrollView` which, itself, is not yet parented or "Windowed".

Also, we get a `ViewDidMoveToWindow` when it finally does get windowed, so that is the ideal time to calculate optimal dimensions, not when we get added to the `NSScrollVIew`.

sgmunn:
Collapsing doesn't seem to re-adjust the column width

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacObjectValueTreeView.cs
```diff
@@ -213,62 +213,65 @@ public class MacObjectValueTreeView : NSOutlineView, IObjectValueTreeView
...
 
-			if (nameColumn.Width != nameWidth) {
+			if ((int) nameColumn.Width != (int) nameWidth) {
 				nameColumn.MinWidth = nameColumn.Width = nameWidth;
 				changed = true;
 			}
 
-			if (valueColumn.Width != valueWidth) {
+			if ((int) valueColumn.Width != (int) valueWidth) {
 				valueColumn.MinWidth = valueColumn.Width = valueWidth;
```
sgmunn: I suspect that removing the assignment to MinWidth might be the fixer for the columns not resizing back down

sgmunn:
removing the assignment to MinWidth and removing the call to SizeToFit makes the columns resize down correctly from a quick test. I'm pretty sure this will also work on Catalina (I made similar changes on my test branch which works).

sgmunn:
I verified this on mojave and catalina and the tooltips are resizing just nicely now

https://share.getcloudapp.com/v1uyq2rk

ugh, can't upload the .mov file.

jstedfast:
Looks amazing, @sgmunn! Thanks for the help!

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacDebuggerObjectNameView.cs
```diff
@@ -201,6 +202,45 @@ protected override void UpdateContents ()
...
+			var editable = treeView.AllowWatchExpressions && node.Parent is RootObjectValueNode;
+			var placeholder = string.Empty;
+			var name = node.Name;
+
+			if (node.IsUnknown) {
+			} else if (node.IsError || node.IsNotSupported) {
+			} else if (node.IsImplicitNotSupported) {
+			} else if (node.IsEvaluating) {
+			} else if (node.IsEnumerable) {
+			} else if (node is AddNewExpressionObjectValueNode) {
```
sgmunn: @jstedfast there are probably some simplifications we can do here, for example, since we only care about the optimal width for tooltips, we know that we will not have the add new expression node in a tooltip.

sgmunn: although, if we're using these calculations later when we adjust the constraints for the view then ignore that.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacDebuggerObjectCellViewBase.cs
```diff
@@ -158,6 +158,22 @@ protected static NSAttributedString GetAttributedPlaceholderString (string text)
...
+
+			if (sizeDelta != 0)
+				modified = NSFont.FromDescription (font.FontDescriptor, font.PointSize + sizeDelta);
+
+			using (var str = new NSAttributedString (text, font: modified ?? font))
+				width = str.Size.Width;
+
+			modified?.Dispose ();
+
+			return NMath.Ceiling (width + 2);
```
sandyarmstrong: will probably want a comment down the road to explain why 2.

jstedfast: Yea, it's just 1px padding on each side of the text. Might need more (or less?), but it was a best guess.

sgmunn:

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacObjectValueTreeView.cs
```diff
@@ -223,52 +230,38 @@ void OptimizeColumnSizes (bool emitResized = true)
...
+				valueColumn.Width = valueWidth;
 				changed = true;
 			}
 
 			if (changed) {
-				SizeToFit ();
-
-				if (emitResized)
-					OnResized ();
+				var optimalTooltipWidth = nameWidth + valueWidth + pinColumn.Width + IntercellSpacing.Width * 2;
```
sandyarmstrong:
```suggestion
				var optimalTooltipWidth = nameWidth + valueWidth + (pinColumn?.Width ?? 0) + IntercellSpacing.Width * 2;
```

sandyarmstrong: @sgmunn I noticed while working on https://github.com/mono/monodevelop/pull/9284 that it is possible for `pinColumn` to be null. My PR has some other fixes to check for that, but this one was specific to this branch (which I originally based my work on).

jstedfast: hmmm, pinColumn should never be null for tooltips or adornments.

sandyarmstrong: It's null if allowpinning is false (maybe that never happens in reality).

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ObjectValue/Mac/MacDebuggerObjectValueView.cs
```diff
@@ -308,8 +310,102 @@ protected override void UpdateContents ()
...
 				constraint.Active = true;
 
-			OptimalWidth += TextField.Frame.Width;
 			OptimalWidth += MarginSize;
+
+			wrapper.OptimalValueFont = TreeView.CustomFont;
+			wrapper.OptimalValueWidth = OptimalWidth;
+		}
+
+		public static nfloat GetOptimalWidth (MacObjectValueTreeView treeView, ObjectValueNode node, bool hideValueButton)
```
slluis: Much of the logic in this method is a copy of what's in UpdateContents. This is not good, it is very fragile. We can merge as is for the sake of the cutoff, but this will have to be fixed.

PR #9271, opened 2019-11-11
[a11y] NuGet Package Management

sevoku:
This fixes some accessibility issues around package management. This is not ideal, and to make it really right we need to make this native. However the fixes here make it good enough to be MAS compliant.

Fixes VSTS #750371
Fixes VSTS #750376


main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement.Gui/ManagePackagesDialog.cs
```diff
@@ -105,9 +106,13 @@ public ManagePackagesDialog (ManagePackagesViewModel viewModel, string initialSe
 			imageLoader.Loaded += ImageLoaded;
 
 			browseLabel.ButtonPressed += BrowseLabelButtonPressed;
+			browseLabel.KeyPressed += BrowseLabelKeyPressed;
```
Therzok: Unsubscribe on dispose

sevoku:

sgmunn:
tried it out, VO appears to work as expected.

sevoku:
@mrward thanks! Yeah the spacing is not ideal, but TBH it was not perfect before either. The contrast problem affects all Gtk tables unfortunately, we need to improve colors there. I think the next step will be native, I tried to get it working, but the main problem for me was the toolbar (which is a custom backend thing), so I ran out of time for the fix.

PR #9272, opened 2019-11-11
[Ide] Allow to unfocus the Keybinding entry with keyboard

sevoku:
When/after recording a new key binding, there is now way
to finish the assignment for keyboard only users, because
a tab key combination will be always recorded and not executed.

This changes the behaviour and allows Gtk to handle the
focus event:

* Don't accept tab as chord and let Gtk move the focus
* For accels allow tab to be entered,
  but don't reset on a second tab and let Gtk handle it again
* If the accel is complete, move focus without resetting the entry,
  allowing the user to finish entering a binding

Fixes VSTS #751986


sevoku:

sevoku:

PR #9275, opened 2019-11-12
Go to Definition for razor projection buffers

KirillOsenkov:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1010683

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.Ide.Completion.Presentation/MonoDevelopContainedDocument.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.Ide.Completion.Presentation/MonoDevelopContainedDocument.cs
@@ -54,6 +54,8 @@ namespace MonoDevelop.Ide.Completion.Presentation
 		public IMonoDevelopContainedLanguageHost ContainedLanguageHost { get; private set; }
 		private readonly HostType _hostType;
 
+		public IProjectionBuffer TopBuffer => DataBuffer;
+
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentManager.cs
```diff
@@ -253,16 +253,17 @@ public async Task<Document> OpenDocument (DocumentController controller, bool br
 
 		public async Task<Document> OpenDocument (FileOpenInformation info)
 		{
-			if (string.IsNullOrEmpty (info.FileName))
+			FilePath filePath = GetActualFilePathToOpen(info.FileName);
```
sandyarmstrong: Does this prevent the user from ever opening the webtools `*.g.cs` files? Do they expect to have a way to do that?

sandyarmstrong:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentManager.cs
KirillOsenkov: You're right, I don't like this. I only want to do this on the codepath where the workspace is asking to open the .g.cs file.

I'll redo it.

KirillOsenkov:
OK we think this is in a good shape now

KirillOsenkov:

PR #9284, opened 2019-11-13
[Debugger] Simplify PinnedWatchView, allow scrolling children via popover

sandyarmstrong:
Currently, the editor intercepts scrolling events that an adornment
such as the PinnedWatchView might want to receive. This prevents such an
adornment from being properly scrollable.

Looking at how pinned watches work on Windows, however, the watch itself
cannot be expanded. Rather, the user can click an expand button to see
the child items in a popup. Inspired by this, I have removed scrolling
and expansion from the PinnedWatchView, and made it so hovering a watch
shows the same popover tooltip (which _is_ scrollable, as it's a child
window) that one sees when hovering a value in the editor.

A few notes:

* Changed the data source to not add child nodes at all when
  `AllowExpanding` is not set on the tree view. This ensures there will
  be no disclosure triangle.
* Pinned watches whose values do not have children will not get a
  tooltip.
* When hovering a pinned watch, the root node will always be expanded.
* The only way to dismiss the watch's popover tooltip is to press ESC
  or click elsewhere. This actually feels pretty good.
* Child expansion state in the tooltip is preserved as long as the watch
  remains pinned.

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/999603

---

this also fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/999610

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.VSTextView/PinnedWatches/PinnedWatchView.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.VSTextView/PinnedWatches/PinnedWatchView.cs
@@ -32,12 +32,15 @@ using Mono.Debugging.Client;
 
 namespace MonoDevelop.Debugger.VSTextView.PinnedWatches
 {
-	sealed class PinnedWatchView : NSScrollView
+	sealed class PinnedWatchView : NSView
 	{
+		readonly PinnedWatch watch;
+		readonly StackFrame frame;
 		readonly ObjectValueTreeViewController controller;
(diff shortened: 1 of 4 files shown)


sandyarmstrong:
pinned watch popover

vancura:
Don't merge it to master yet, please. I think an email to the board would be fine, we don't need the full meeting.

sgmunn:
@sandyarmstrong could I get you to email the UXboard for this?

sandyarmstrong:
@sgmunn check your inbox :)

sgmunn:
looks good when I ran it.

sandyarmstrong:
@monojenkins rebase

sandyarmstrong:

PR #9290, opened 2019-11-13
[Debugger] Add more Begin/End Updates around tree manipulation

sgmunn:
Simplest guess for VSTS 1004450

I'm not going to claim this as a fix, but it is possibly related and, without some larger refactoring, seems to be the simplest of the guesses that we have for those crashes. Also, related changes in column sizing code might be related and offer fixes (if the crashes were in tooltips or pinned watches, which we can't tell because it's the same code for all cases).

In any case, we should add these begin/end updates around code where we manipulate the tree items. 

---
related:

VSTS 1008044

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/LocalsPad.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/LocalsPad.cs
@@ -77,9 +77,13 @@ namespace MonoDevelop.Debugger
 			DebuggerLoggingService.LogMessage ("End Local Variables");
 
 			if (UseNewTreeView) {
-				controller.ClearValues ();
-				controller.AddValues (locals);
-
+				_treeview.BeginUpdates ();
+				try {
+					controller.ClearValues ();
+					controller.AddValues (locals);
+				} finally {
+					_treeview.EndUpdates ();
+				}
 #if ADD_FAKE_NODES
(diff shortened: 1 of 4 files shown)


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/LocalsPad.cs
```diff
@@ -77,9 +77,13 @@ void ReloadValues ()
...
 			if (UseNewTreeView) {
-				controller.ClearValues ();
-				controller.AddValues (locals);
-
+				_treeview.BeginUpdates ();
+				try {
+					controller.ClearValues ();
+					controller.AddValues (locals);
+				} finally {
+					_treeview.EndUpdates ();
```
sandyarmstrong: Should this go around the whole if/else?

sgmunn: the else uses the old gtk treeview (from 8.3 and earlier), so not really. "treeView" in the else case is the legacy / original ObjectValueTreeView

sandyarmstrong: Oh, whoops!

sgmunn:

sgmunn:
@monojenkins rebase

PR #9294, opened 2019-11-13
[Debugger] Improve Breakpoint dialog accessibility

sevoku:
Use LabelWidget rather than Label for combo boxes,
because Label is also used to reflect combo box
selections with AtkCocoa.

Also add some more accessibility label relations.

Fixes VSTS #1018518


sevoku:
@monojenkins rebase

sevoku:

PR #6926, opened 2019-01-18
[Ide] Let the references search be a child of the notebook

sevoku:
This is a partial fix for VSTS #750263, AtkCocoa needs to support the right focus order, including an Action widget.

PR #9297, opened 2019-11-13
Fix focus problems in Select References Dialog

bratsche:
This partially fixes VSTS #750236 (along with https://github.com/mono/monodevelop/pull/6926)

fe25a99ffeee2503c7f0be0a8cfbccdf27c80a8c fixes the initial focus of the dialog so it's on the notebook tabs and not the search entry

e48c2bf61005f5ab938bc6922583ec033aef300e fixes a focus chain issue that caused the notebook tab to get focused twice when tabbing through the dialog. It seemed like it was an AtkCocoa issue of the notebook being registered twice or something, but it turned out to be because of a keyboard filter.


bratsche:
I pushed these commits into https://github.com/mono/monodevelop/pull/6926 so I'm closing this and we'll just do it all in that PR.

PR #9298, opened 2019-11-14
Refactor how virtual and dynamic documents are added to the workspace.

KirillOsenkov:
Virtual documents come from Razor when a .cshtml file is opened and they have .cshtml.g.cs file extension. These are C# projection buffers generated by Razor and only exist while the corresponding .cshtml file is open in an editor.

Dynamic documents also come from Razor but these come from all .razor and .cshtml files in the project that have the MSBuild item type "Content". These don't have to be open in the editor.

The problem is that we weren't adding virtual and dynamic documents in all scenarios. Specifically when the solution was initially loaded our code was not running and so if a Razor document was open before the solution was loading it wouldn't get added to the workspace properly.

To mitigate this we've found a single codepath (LoadProject) that all the logic in the project system goes through. This includes the initial load as well as reload. This is the single place where a Roslyn ProjectInfo is created, and it's also a better location to add virtual and dynamic documents to the project being loaded.

There is now no need to call AddVirtualDocuments from three separate locations.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1012526

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/DynamicFileManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/DynamicFileManager.cs
@@ -141,7 +141,7 @@ namespace MonoDevelop.Ide.TypeSystem
 					}
 				}
 
-				projectInfo = projectInfo.WithAdditionalDocuments(documents);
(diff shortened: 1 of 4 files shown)


sandyarmstrong:
@monojenkins rebase

sandyarmstrong:

PR #9307, opened 2019-11-14
Implement Go To Definition for Razor

KirillOsenkov:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1010683

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices/MonoDevelopDocumentNavigationService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices/MonoDevelopDocumentNavigationService.cs
@@ -1,7 +1,7 @@
 using System;
 using System.Composition;
+using System.Linq;
 using System.Threading;
-using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Host;
 using Microsoft.CodeAnalysis.Host.Mef;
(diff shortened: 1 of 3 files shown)


KirillOsenkov:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.RoslynServices/MonoDevelopDocumentNavigationService.cs
```diff
@@ -99,13 +110,12 @@ public bool TryNavigateToSpan (Workspace workspace, DocumentId documentId, TextS
...
 			Runtime.AssertMainThread ();
 
-			var document = OpenDocument (workspace, documentId, options);
+			var document = workspace.CurrentSolution.GetDocument (documentId);
 			if (document == null) {
 				return false;
 			}
 
 			var text = document.GetTextSynchronously (CancellationToken.None);
-			var textBuffer = text.Container.GetTextBuffer ();
```
sandyarmstrong: This line is the only difference from before?

KirillOsenkov: yes, and it was causing the problem. We used to open the file twice, once in the OpenDocument call above, and once in NavigateTo below.

We removed the OpenDocument call above, but now the document isn't open by the time line 108 runs, so there was no buffer and we crashed. 

Good news is that the line wasn't used for anything so just deleting it solves it.

And as a bonus we only open the document once now, at the end.

PR #9309, opened 2019-11-14
[Core] Avoid showing duplicated files in project tree

rodrmoya:
There are some cases where, it seems, we don't evaluate conditions as we should,
resulting in duplicated items being added to the project and, thus, also shown
in the project tree (see aspnet/AspNetCore#17088).

Also, by just adding a <None Include="*"/>, duplicated files show also.

So, due to this limitation on the project system, work around it on the project
tree, by checking for duplicates based on the files' paths.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1005277

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
@@ -81,6 +81,7 @@ namespace MonoDevelop.Ide.Gui.Pads.ProjectPad
 		{
 			string folderPrefix = folder + Path.DirectorySeparatorChar;
 
+			var visitedFiles = new HashSet<FilePath> ();
 			files = new List<ProjectFile> ();
 			folders = new List<string> ();
 			
@@ -100,7 +101,8 @@ namespace MonoDevelop.Ide.Gui.Pads.ProjectPad
 						? project.BaseDirectory.Combine (file.ProjectVirtualPath).ParentDirectory
 						: file.FilePath.ParentDirectory;
 						
-					if (dir == folder) {
+					if (dir == folder && !visitedFiles.Contains (file.FilePath)) {
+						visitedFiles.Add (file.FilePath);
 						files.Add (file);
 						continue;
 					}
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
@@ -31,6 +31,7 @@ using System.Collections.Immutable;
 using System.Linq;
 using System.Runtime.CompilerServices;
 using Mono.Addins;
+using MonoDevelop.Core;
 using MonoDevelop.Projects;
 
 namespace MonoDevelop.Ide.Projects.FileNesting
@@ -166,7 +167,7 @@ namespace MonoDevelop.Ide.Projects.FileNesting
 			}
 		}
 
-		readonly ConcurrentDictionary<ProjectFile, ProjectFileNestingInfo> projectFiles = new ConcurrentDictionary<ProjectFile, ProjectFileNestingInfo> ();
+		readonly ConcurrentDictionary<FilePath, ProjectFileNestingInfo> projectFiles = new ConcurrentDictionary<FilePath, ProjectFileNestingInfo> ();
 		bool fileNestingEnabled;
 
 		public Project Project { get; }
@@ -195,10 +196,10 @@ namespace MonoDevelop.Ide.Projects.FileNesting
 
 		ProjectFileNestingInfo AddFile (ProjectFile projectFile)
 		{
-			var tmp = projectFiles.GetOrAdd (projectFile, new ProjectFileNestingInfo (projectFile));
+			var tmp = projectFiles.GetOrAdd (projectFile.FilePath, new ProjectFileNestingInfo (projectFile));
 			tmp.Parent = FileNestingService.InternalGetParentFile (projectFile);
 			if (tmp.Parent != null) {
-				var parent = projectFiles.GetOrAdd (tmp.Parent, new ProjectFileNestingInfo (tmp.Parent));
+				var parent = projectFiles.GetOrAdd (tmp.Parent.FilePath, new ProjectFileNestingInfo (tmp.Parent));
 				if (parent.Children == null) {
 					parent.Children = new ProjectFileCollection ();
 				}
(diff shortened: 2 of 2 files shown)


rodrmoya:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
```diff
@@ -81,6 +81,7 @@ void GetFolderContent (Project project, string folder, out List<ProjectFile> fil
 		{
 			string folderPrefix = folder + Path.DirectorySeparatorChar;
 
+			var visitedFiles = new HashSet<string> ();
```
Therzok: Use `HashSet<FilePath>` here for case insensitive duped

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
```diff
@@ -166,7 +166,7 @@ public ProjectFileNestingInfo (ProjectFile projectFile)
 			}
 		}
 
-		readonly ConcurrentDictionary<ProjectFile, ProjectFileNestingInfo> projectFiles = new ConcurrentDictionary<ProjectFile, ProjectFileNestingInfo> ();
+		readonly ConcurrentDictionary<string, ProjectFileNestingInfo> projectFiles = new ConcurrentDictionary<string, ProjectFileNestingInfo> ();
```
Therzok: Same here

Therzok: There might be a big perf hit from object reference equality vs full string match

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
Therzok: Also add w comment that it's a workaround for a project system limitation

rodrmoya:

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
rodrmoya: Done both

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
rodrmoya: Changed to have `FilePath` as the key, but not sure what to do with the perf, the work around works because we check the file paths, as the files are duplicated in `project.Files`, so can't do object reference equality.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/FolderNodeBuilder.cs
```diff
@@ -81,6 +81,7 @@ void GetFolderContent (Project project, string folder, out List<ProjectFile> fil
 		{
 			string folderPrefix = folder + Path.DirectorySeparatorChar;
 
+			var visitedFiles = new HashSet<FilePath> ();
```
Therzok: We still need a comment on the fact that this hashset is needed as a workaround.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
```diff
@@ -166,7 +167,7 @@ public ProjectFileNestingInfo (ProjectFile projectFile)
 			}
 		}
 
-		readonly ConcurrentDictionary<ProjectFile, ProjectFileNestingInfo> projectFiles = new ConcurrentDictionary<ProjectFile, ProjectFileNestingInfo> ();
+		readonly ConcurrentDictionary<FilePath, ProjectFileNestingInfo> projectFiles = new ConcurrentDictionary<FilePath, ProjectFileNestingInfo> ();
```
Therzok: Add a comment that the FilePath is used as a workaround, we should be using ProjectFile.

mrward:
Alternative fix here which should address the underlying problem of Conditions being ignored on the Content items - #9407

rodrmoya:
Closing this one then

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
```diff
@@ -305,7 +306,7 @@ public ProjectFileCollection GetChildrenForFile (ProjectFile inputFile)
 			if (!fileNestingEnabled)
 				return null;
 
-			return projectFiles.TryGetValue (inputFile, out var nestingInfo) ? nestingInfo.Children : null;
+			return projectFiles.TryGetValue (inputFile.FilePath, out var nestingInfo) ? nestingInfo.Children : null;
 		}
```
mrward: @rodrmoya Can you remember why you had to change the FileNestingService here for the duplicate files? Just wondering if the other pull request is enough to fix the problem since that change would only affect the files in the Solution pad window, not the files belonging to the project.

rodrmoya: hmm, so `FileNestingService` deals with the files in the project, not the solution pad really, so if the files in the project are still duplicated, I guess maybe we need this change after all?

mrward: I did not see any duplicate files in the Solution pad. Just wondering what would happen here with the nesting if Project.Files contained duplicate files, but with different conditions.

rodrmoya: Or well, maybe it's not needed at all, as `FileNestingService` is called from the node builders, so if those don't get the duplicated files, it should be ok. But not sure, as the service listens to FileAdded/Removed events on the project to keep the data up to date

rodrmoya: I think it might be affected indeed, as files are stored on a dictionary -> https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs#L198 with the `ProjectFile` being the key, so maybe it makes sense to get the FileNestingService's changes of this PR, although, as @Therzok mentioned, that adds a big perf hit :/

I'll have a look and see how/if it gets affected

PR #9314, opened 2019-11-14
Instead of using custom ISerializable to serialize private fields, just expose it via ReadOnlyCollection

manish:
Needs partial reverse of https://github.com/mono/monodevelop/pull/8846

Fixes #1019319

We can just expose the hidden `values` CounterValue using ReadOnlyCollection with only a getter. It prevents modification of the existing list. This is better since `ISerializable` has many issues esp `Counter.values` were not serialized but `TimerCounter.values` were. Using the additional read-only property provides a C# solution rather than relying on underlying implementation fo Newtonsoft.Json

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/Counter.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/Counter.cs
@@ -28,6 +28,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.IO;
 using System.Linq;
 using System.Runtime.Serialization;
@@ -114,7 +115,9 @@ namespace MonoDevelop.Core.Instrumentation
 		}
 
 		public virtual CounterDisplayMode DisplayMode => CounterDisplayMode.Block;
-		
+
+		public IReadOnlyList<CounterValue> AllValues => new ReadOnlyCollection<CounterValue> (values);
+
 		public IEnumerable<CounterValue> GetValues ()
 		{
 			lock (values) {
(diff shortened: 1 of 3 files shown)


manish:
@monojenkins rebase

Therzok (approved):

manish:
@monojenkins backport release-8.5

manish:

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/Counter.cs
```diff
@@ -114,7 +115,9 @@ public override string ToString ()
 		}
 
 		public virtual CounterDisplayMode DisplayMode => CounterDisplayMode.Block;
-		
+
+		public IReadOnlyList<CounterValue> AllValues => new ReadOnlyCollection<CounterValue> (values);
+
```
mrward: Do we need to lock values? Other code seems to lock values.

Therzok (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Instrumentation/Counter.cs
```diff
@@ -99,22 +100,30 @@ public override string ToString ()
...
 		public virtual CounterDisplayMode DisplayMode => CounterDisplayMode.Block;
-		
+
+		public IReadOnlyList<CounterValue> AllValues {
+			get {
+				lock (values) {
+					return new ReadOnlyCollection<CounterValue> (values);
+				}
+			}
+		}
```
mrward: Not sure this is giong to work. We are passing a list to the [ReadOnlyCollection](https://github.com/mono/mono/blob/a6db160596bb987af11fb2844d51c9a3a7c73133/mcs/class/referencesource/mscorlib/system/collections/objectmodel/readonlycollection.cs#L27-L32) and it looks like it just uses that list directly. We may need to clone it to a new list inside a lock and pass it to the ReadOnlyDictionary.

Therzok: Good point

Therzok (approved):

mrward (approved):

manish:
@monojenkins backport release-8.5

manish:

PR #9315, opened 2019-11-14
[Debugger] Fixed calculation for NSTextField width as used by the tre…

jstedfast:
…eview cells

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1006468/


jstedfast:

PR #9325, opened 2019-11-15
Revert "First pass to fix solution pad not regaining focus on shortcut press."

netonjm:
Revert "First pass to fix solution pad not regaining focus on shortcut press."
This reverts commit 7e01fd039a5c6705189f0a927868919e953e9c73 in this PR : https://github.com/mono/monodevelop/pull/9216

I reverted this commit because introduces new problems, in this case infinite focus loops on in Pads with autohide with embedded native views ( https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1020127)

Additionally all this logic is going to be obsolete by https://github.com/mono/monodevelop/pull/9267
which hopefully will fix this issue also.

Fixes VSTS #1020127 - Toolbox focus flickers when in autohide mode


netonjm:

PR #9335, opened 2019-11-15
[Debugger] Fixed an NRE exception in OVTVController.RemoveValue if no…

jstedfast:
…de.Parent is null

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1022024


jstedfast:

jstedfast:
I think the root cause of this NRE was already fixed by @sgmunn a few days ago in commit f143a20ac2752505191870ca5138258c41ed8bf9, but that's just a theory since I'm not sure how to reproduce this NRE.

jstedfast:

jstedfast:
I think that second commit is a more likely/better fix for this since I think what is happening is that the user had non-root nodes selected when he hit Delete

sgmunn:
I don't think my fix changes that. this I think is removing a pinned watch (on startup??) and my patch is in the button click code path

sgmunn:
Whilst not able to repro the crash, I did verify that the patch stops you deleting sub-nodes incorrectly. we do need to fix some enabled state issues on the context menu but that can be done later if we need to cut off for today.

PR #9342, opened 2019-11-15
[Debugger] Don't allow selection of the pinned watch row

jstedfast:
This prevents the row from getting into an odd
selected-but-not-focused state.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1021870/


jstedfast:

PR #9347, opened 2019-11-16
Fix 1023494: Unsaved new documents have `/` added to begining of file name

DavidKarlas:
I think that anything that needs Canonical comparison should make conversion itself.
I'm also not aware of any regressions this might cause... @slluis any ideas?

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/FileDocumentController.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/FileDocumentController.cs
@@ -25,6 +25,7 @@
 // THE SOFTWARE.
 
 using System;
+using System.IO;
 using System.Threading.Tasks;
 using MonoDevelop.Core;
 using System.Text;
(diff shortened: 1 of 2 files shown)


Therzok:
@davidkarlas maybe we could have a unit test with a file controller pointing to a symlinked file, and assert that the symlink is unchanged and we wrote to the real file.

DavidKarlas:
@Therzok what is bug# of bug you are referring to?

Therzok:
Can't find it, seems to be new code in the new document model. In any case, can we add that test? :D

DavidKarlas:
Test for something this PR doesn't fix?

Therzok:
Okay, for some reason, I thought CanonicalPath resolves symlinks, but it doesn't. LGTM

DavidKarlas:
Added unit test for new behavior.

slluis:
A better solution would be to only get the canonical path when the path is absolute.

DavidKarlas:
@slluis updated code as suggested to do Canonical only for AbsolutePath.

PR #9364, opened 2019-11-21
[Ide[ Configure Gtk# for slim startup

Therzok:


Therzok:
@slluis @sevoku PTAL. This fixes counter reporting for gtk# init, so it properly reports how much gtk# takes to init now.

And it enables some flag that when used in a newer gtk#, it will optimize startup performance a bit more for apps that opt into this functionality.

PR #9371, opened 2019-11-25
[AspNet*] Add a group for ASP.NET Core Browser execution targets

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/945598

### DIFF
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
@@ -195,14 +195,9 @@ namespace MonoDevelop.AspNetCore
 				}
 			}
 
-			if (result.Count > 0)
-				yield return result;
-			else {
-				// Fallback to base implementation
-				foreach (var target in base.OnGetExecutionTargets (ctx, configuration, runConfig)) {
-					yield return target;
-				}
-			}
+			return result.Count > 0
+				? new ExecutionTarget [] { result }
+				: base.OnGetExecutionTargets (ctx, configuration, runConfig);
 		}
 
 		IEnumerable<ExecutionTarget> GetMultipleTargetFrameworkExecutionTargets (DesktopApplication browser)
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
@@ -178,7 +178,7 @@ namespace MonoDevelop.AspNetCore
 
 		protected override IEnumerable<ExecutionTarget> OnGetExecutionTargets (OperationContext ctx, ConfigurationSelector configuration, SolutionItemRunConfiguration runConfig)
 		{
-			var result = new List<ExecutionTarget> ();
+			var result = new ExecutionTargetGroup (GettextCatalog.GetString ("Browser"), "MonoDevelop.AspNetCore.BrowserExecutionTargets");
 			foreach (var browser in IdeServices.DesktopService.GetApplications ("https://localhost", Ide.Desktop.DesktopApplicationRole.Viewer)) {
 				if (browser.IsDefault) {
 					if (Project.HasMultipleTargetFrameworks) {
@@ -195,7 +195,14 @@ namespace MonoDevelop.AspNetCore
 				}
 			}
 
-			return result.Count > 0 ? result : base.OnGetExecutionTargets (configuration);
+			if (result.Count > 0)
+				yield return result;
+			else {
+				// Fallback to base implementation
+				foreach (var target in base.OnGetExecutionTargets (ctx, configuration, runConfig)) {
+					yield return target;
+				}
+			}
 		}
 
 		IEnumerable<ExecutionTarget> GetMultipleTargetFrameworkExecutionTargets (DesktopApplication browser)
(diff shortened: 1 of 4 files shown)


main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -178,24 +178,28 @@ private ExecutionCommand CreateAspNetCoreExecutionCommand (ConfigurationSelector
...
+							result.Add (target);
+						}
 					} else {
 						result.Add (new AspNetCoreExecutionTarget (browser));
 					}
 				}
 			}
 
-			return result.Count > 0 ? result : base.OnGetExecutionTargets (configuration);
+			yield return result;
```
mrward: Before if there were no items we would call the base.OnGetExecutionTargets. Now we do not do that. Is that correct?

rodrmoya: Well, now we always return something, but yeah, I guess I should test if the group has any target, and if not, return `base.OnGetExecutionTargets`

rodrmoya: Fixed

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -195,7 +195,13 @@ protected override IEnumerable<ExecutionTarget> OnGetExecutionTargets (Operation
...
 			}
 
-			return result.Count > 0 ? result : base.OnGetExecutionTargets (configuration);
+			if (result.Count > 0)
+				yield return result;
+
+			// Fallback to base implementation
+			foreach (var target in base.OnGetExecutionTargets (ctx, configuration, runConfig)) {
+				yield return target;
+			}
```
mrward: This still looks different to what it was before since it now also returns the base.OnGetExecutionTargets after the group is returned. Before it would return either the browsers or the base.OnGetExecutionTargets.

rodrmoya: Ugh yeah, sorry, fixed now

mrward (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/ExecutionTarget.cs
```diff
@@ -220,6 +220,20 @@ IEnumerator IEnumerable.GetEnumerator ()
 		}
 
 		#endregion
+
+		#region Helper methods
```
iantoalms: These are used under the hood somewhere?

iantoalms (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/ExecutionTarget.cs
rodrmoya: Yes, here -> https://github.com/mono/monodevelop/pull/9371/files#diff-39987f1423a39aa8e26e70a4d295587cR185 and https://github.com/mono/monodevelop/pull/9371/files#diff-39987f1423a39aa8e26e70a4d295587cR191

nosami (approved):

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -195,7 +195,14 @@ protected override IEnumerable<ExecutionTarget> OnGetExecutionTargets (Operation
 				}
 			}
 
-			return result.Count > 0 ? result : base.OnGetExecutionTargets (configuration);
+			if (result.Count > 0)
+				yield return result;
```
Therzok: Can't we just do:
```csharp
return result.Count > 0 ? result : base.OnGetExecutionTargets (ctx, configuration, runConfig);
```

Does that change behaviour?

mrward: Yeah I think that code, which was used originally, would work.

Therzok: There's also a difference between the two OnGetExecutionTargets call, but I think the new version is actually correct.

rodrmoya: Yeah, tried that, but returning the `ExecutionTargetGroup`, which implements `IList` just returns inside the group, not the group itself, which is what we want

rodrmoya: > There's also a difference between the two OnGetExecutionTargets call, but I think the new version is actually correct.

yes, better to call it with all the arguments, right?

slluis:
Can you attach a screenshot of the result?

rodrmoya:
image

mrward:
Out of interest - what does this look like for a multi-target project? Not sure it is likely to have a multi-target ASP.NET Core project.

rodrmoya:
> Out of interest - what does this look like for a multi-target project? Not sure it is likely to have a multi-target ASP.NET Core project.

doesn't seem to work with ASP.NET Core projects

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
Therzok:
```csharp
return result.Count > 0 ? new ExecutionTarget [] { result } : base.OnGetExecutionTargets (ctx, configuration, runConfig);
```

Should work then.

mrward (approved):

PR #9373, opened 2019-11-25
[Debugger] Handle NRE when adding a watch when not debugging

sgmunn:
Fixes VSTS 1025766
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1025766


sgmunn:
@mrward that's two additional test scenarios we need to add :)

sgmunn:

PR #9374, opened 2019-11-25
[Unit Test] Fix the nesting for unit tests

ketttY:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/901156/

### DIFF
--- a/main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/DiscoveredTests.cs
+++ b/main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.VsTest/DiscoveredTests.cs
@@ -48,7 +48,7 @@ namespace MonoDevelop.UnitTesting.VsTest
 		{
 			tests.Sort (OrderByName);
 
-			var parentNamespace = new VsTestNamespaceTestGroup (projectTestSuite, null, projectTestSuite.Project, String.Empty);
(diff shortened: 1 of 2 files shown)


ketttY:
@monojenkins rebase

ketttY:
Current nesting:
**Solution -> Project -> Namespace -> Test class -> Test case**
Screen Shot 2019-11-26 at 10 51 44 AM

VS win doesn't have the solution layer displayed, wondering should we keep it or not?

mrward:
VS Win does not support multiple solutions so they can avoid showing the solution. The Run All Tests is fine in VS Win for a single solution. If you have multiple solutions open in VS Mac there is no easy way to run all tests for a single solution if there is no root solution tree node. Which probably means we need to keep the solution node at the top.

mrward:
Can we add some screenshots before and after the change?

Also might be worth adding some tests for this new logic. There are tests for the VS Test tree but since everything is green here it seems I did not add enough tests to cover the original namespace logic otherwise some would have failed.

https://github.com/mono/monodevelop/tree/master/main/src/addins/MonoDevelop.UnitTesting/MonoDevelop.UnitTesting.Tests/MonoDevelop.UnitTesting.Tests

ketttY:
Screen Shot 2019-12-03 at 4 56 55 PM

- **Project**: mstest

- **Namespaces**: 

   1. mstest (with **childNamespace**: childNamespace)
   2. mstest2

mrward:
Was hoping for some unit tests to be added before this was merged 😞

PR #9385, opened 2019-11-26
Invert icon color for selected scaffolder row

nosami:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1022976

### DIFF
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.Scaffolding/ScaffolderTemplateSelectPage.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.Scaffolding/ScaffolderTemplateSelectPage.cs
@@ -37,6 +37,7 @@ namespace MonoDevelop.AspNetCore.Scaffolding
 	{
 		ListBox listBox;
 		ScaffolderArgs args;
+		int selectedRow;
 
 		public event EventHandler ScaffolderSelected;
 		public ScaffolderTemplateSelectPage (ScaffolderArgs args) : base (args)
@@ -69,9 +70,11 @@ namespace MonoDevelop.AspNetCore.Scaffolding
 
 			var scaffolders = GetScaffolders ().Value;
 
+			var png = ImageService.GetIcon ("md-html-file-icon", Gtk.IconSize.Dnd);
+			var selectedPng = png.WithStyles ("sel");
+
 			foreach (var scaffolder in scaffolders) {
 				var row = listStore.AddRow ();
-				var png = ImageService.GetIcon ("md-html-file-icon", Gtk.IconSize.Dnd);
 
 				listStore.SetValue (row, icon, png);
 				listStore.SetValue (row, name, scaffolder.Name);
@@ -84,7 +87,12 @@ namespace MonoDevelop.AspNetCore.Scaffolding
 			listBox.DataSource = listStore;
 			listBox.HeightRequest = 300;
 			listBox.WidthRequest = 300;
-			listBox.SelectionChanged += (sender, e) => Args.Scaffolder = scaffolders [listBox.SelectedRow];
+			listBox.SelectionChanged += (sender, e) => {
+				Args.Scaffolder = scaffolders [listBox.SelectedRow];
+				listStore.SetValue (selectedRow, icon, png);
+				listStore.SetValue (listBox.SelectedRow, icon, selectedPng);
+				selectedRow = listBox.SelectedRow;
+			};
 			listBox.RowActivated += (sender, e) => ScaffolderSelected?.Invoke (sender, e);
 			listBox.SelectRow (0);
 			listBox.FocusedRow = 0;


nosami:

nosami:

sevoku:
Fixed in Xwt: https://github.com/mono/xwt/pull/1009

nosami:
> Just wondering if we want to remove the SelectionChanged event handler on disposing and store the images, so they can be used in the event handler. //cc @Therzok

why? it will be removed by GC

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.Scaffolding/ScaffolderTemplateSelectPage.cs
```diff
@@ -84,7 +87,12 @@ protected override Widget GetMainControl ()
 			listBox.DataSource = listStore;
 			listBox.HeightRequest = 300;
 			listBox.WidthRequest = 300;
-			listBox.SelectionChanged += (sender, e) => Args.Scaffolder = scaffolders [listBox.SelectedRow];
+			listBox.SelectionChanged += (sender, e) => {
+				Args.Scaffolder = scaffolders [listBox.SelectedRow];
```
Therzok: listBox creates a refcycle to itself via this event handler.

Therzok: Even before, it was creating a refcycle to itself via `this`.

`listBox` -> `SelectionChanged` -> `this` -> `listBox`

sevoku:
@nosami Matt is right, it's a ref cycle (listbox <> handler <> listbox), so GC won't touch this unless released manually. However I think we can close this anyway, this is the final fix: https://github.com/mono/monodevelop/pull/9396

Therzok:
Leak still happens :D

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.Scaffolding/ScaffolderTemplateSelectPage.cs
nosami: Really? TIL

I was working on assumption that ScaffolderTemplateSelectPage and listBox had the same lifetime scope so would be GC'd at the same together.... doesn't matter that listBox references itself if it will be GC'd anyway

nosami:
Not convinced, sorry

nosami:
ScaffolderTemplateSelectPage and listBox are both instances with same lifetime. If listBox is marked for collection (as would be the case when ScaffolderTemplateSelectPage is disposed), then it doesn't matter that there is a `listbox(marked for deletion) <> handler <> listbox(marked for deletion)` cycle. Why would it?

nosami:
On the other hand, it obviously doesn't hurt to just write the extra boiler plate code out to be on the _extremely safe_ side. But cargo cult can be harmful.

nosami:
Maybe I am wrong about this, in which case I take everything back and apologise. 

I can't believe that the GC would be so idiotic about this though, which makes me think I am right.

Therzok:
Yes, you are right, but the problem here is not the GC, but the toolkits. Good thing you guys brought this up, cause I completely forgot about this and my brain is hard-wired to Gtk# before leak fixes done in the toolkit. I just refreshed my memory on how the whole thing works, so here's what we have:

There's been a lot of work done in the past to fix this for Gtk#. Since 2017, handler removal is _not mandatory_ if `Destroy` is called, but I'm not sure when the actual patch was included in mono - iirc it was 2018 since the changes were considered high risk. We still had to maintain back compat since mono versions could be changed, etc.

https://github.com/mono/gtk-sharp/commit/f6fc2a437d173386f388b77ab928299e3e4d85df.

Wall of text in details with internal functionality and some disclaimers:
<details>
Gtk+ is reference counted. That is straightforward.

GOOD PART: On top, we add Gtk#, which replaces the initial unmanaged ref with a ToggleRef - which changes from `strong` to `weak` reference on the actual gtk# object based on refcount [here](https://github.com/mono/gtk-sharp/blob/gtk-sharp-2-12-branch/glib/ToggleRef.cs#L118-L127). So that means when refcount is 1, which would happen after destroy, we would keep the object only with a weak reference in a static `Dictionary<IntPtr, object>`. And the ToggleRef also keeps a reference to the signals that are attached as event handlers.

So, when calling [Destroy](https://developer.gnome.org/gtk2/stable/GtkWidget.html#gtk-widget-destroy), we're going to take the refcount down to 1 (remember, we switched our first refcount). In 2017, this whole situation was fixed and everything is better - there's a series of patches you can check on 2-12 log.

There is a [historical hack](https://github.com/mono/gtk-sharp/blob/gtk-sharp-2-12-branch/gtk/Object.custom#L98-L105), so there is no way to actually free an object immediately, simply because `Dispose()` would cause crashes if you tried to query Gtk# and preserving legacy app functionality was needed - as refcounting was completely wrong before, which meant a lot of weird code got written. So we leave it to the finalizer and not care too much.

One important patch which changed signal behaviour here: https://github.com/mono/gtk-sharp/commit/f6fc2a437d173386f388b77ab928299e3e4d85df. With this patch, we now don't _require_ handlers to be removed. But this patch was included really late in mono and for a while users might've switched mono versions and had memory leaks. We can probably stop being overly-defensive on Gtk# side if all users have the required gtk#.

_All good until here_. But if there's any strongref, it'll leak as long as the one ref-ing it does does. So in this case, it would be the `page`, all good. The above text gives insight into the internals of gtk# and how it ended up being sane after many iterations spent on improving its ref-ing mechanisms.
</details>

Xwt (which is used here) has a problem with unsubscribing on dispose - https://github.com/mono/xwt/pull/835. This patch got reverted a few days after - and I forgot about it until this comment. I should resurrect that PR without the finalizer codepath and that should be solved as well.

Manually unsubscribing should not be needed, and I agree. The reason it became a _pattern_ to unsubscribe is that Gtk# implementation was sub-par and not taking advantage of the recursive Destroy() propagation to widget children.

It is still possible to create lifetime cycles in managed, but that has nothing to do with native. Most importantly, it is still _good hygiene_ to unsubscribe handlers and null out fields - which are not WeakReference - in `OnDestroyed` simply to prevent future chains of huge objects being retained for a longer time.

But now that we have infrastructure for leak tests, such errors can be spotted in CI and the situation has changed.

Note: For Xamarin.Mac, there is _no way_ to actually avoid needing to unsubscribe. Unlike Gtk#, where Destroy() propagates to children, cocoa does not have that - and an event handler will create a strong ref cycle. See, and the fact that there's no way to clear refs between views: https://docs.microsoft.com/en-us/xamarin/ios/deploy-test/performance

So for Xamarin.Mac, it will be mandatory to use WeakReference between objects and event handler subscription removal - as those have a Delegate.Target, which will be one of the views.

Therzok:
Tl;dr: Gtk# bindings did not clear signals on destroy, so they needed to be manually removed. 

It's still possible to capture unwanted data and make it live longer via closures.

Therzok:
Currently, code does not leak under Gtk#.

If this code stays as is, it'll create leaks under Xwt (even with the fix) and Xamarin.Mac used as a backend. We're going to have to figure out a way to actually not to introduce leaks when doing the switch to Xwt using Xamarin.Mac.

Manually chaining dispose calls will surface bugs like the folks on the editor team are encountering, which are really hard to fix. Going forward, we should be properly using WeakReference and weak eventing for Xamarin.Mac.

sevoku:
@nosami Xwt has been bumped, could you verify the fix and close this PR then?

PR #9391, opened 2019-11-26
[Ide] Load AddinManager data on the UI thread.

Therzok:
We need to run this on the UI thread since AddinManager is not thread safe.
Invoking extension changed handlers on non-UI thread can lead to weird results

Fixes VSTS #1027414 - [FATAL] SigTerm signal in Mono.Addins.dll!Mono.Addins.RuntimeAddin::LoadModule+82"

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
@@ -124,7 +124,7 @@ namespace MonoDevelop.Ide.Composition
 				var fullTimer = System.Diagnostics.Stopwatch.StartNew ();
 				var stepTimer = System.Diagnostics.Stopwatch.StartNew ();
 
-				var mefAssemblies = ReadAssembliesFromAddins (timer);
+				var mefAssemblies = await Runtime.RunInMainThread (() => ReadAssembliesFromAddins (timer));
 				timings ["ReadFromAddins"] = stepTimer.ElapsedMilliseconds;
 				stepTimer.Restart ();
 
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
@@ -72,19 +72,19 @@ namespace MonoDevelop.Ide.Composition
 
 		protected override Task OnInitialize (ServiceProvider serviceProvider)
 		{
-			Runtime.AssertMainThread ();
+			return Runtime.RunInMainThread (() => {
+				var timings = new Dictionary<string, long> ();
+				var metadata = new CompositionLoadMetadata (timings);
 
-			var timings = new Dictionary<string, long> ();
-			var metadata = new CompositionLoadMetadata (timings);
+				var timer = Counters.CompositionLoad.BeginTiming (metadata);
+				var stepTimer = System.Diagnostics.Stopwatch.StartNew ();
 
-			var timer = Counters.CompositionLoad.BeginTiming (metadata);
-			var stepTimer = System.Diagnostics.Stopwatch.StartNew ();
+				var mefAssemblies = ReadAssembliesFromAddins (timer);
 
-			var mefAssemblies = ReadAssembliesFromAddins (timer);
+				timings ["ReadFromAddins"] = stepTimer.ElapsedMilliseconds;
 
-			timings ["ReadFromAddins"] = stepTimer.ElapsedMilliseconds;
-
-			return Task.Run (() => InitializeInstanceAsync (timer, mefAssemblies));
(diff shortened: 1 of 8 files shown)


Therzok:
Deadlocks in TextEditor and VersionControl.Git

Therzok:
One other solution I can think of is _preloading_ the addin assemblies somewhere else (read earlier in the startup sequence), and avoiding possible deadlocks. cc @slluis what do you think?

Therzok:
I'll open another PR where we eagerly load the addin assemblies that are required for the composition and we get rid of the main thread synchronization here. It might slightly regress startup time, but the cost of reliability is higher.

With this version of the PR, we can fail to initialize the IDE/extensions if extensions query MEF before it's built. By having the on-demand behaviour of the lazy init, we just eagerly load the addin assemblies from the extension point and start composing MEF in the background _right after_ that happens. Using that mechanism, we should be able to get the best of both worlds:

* reliable startup at the cost of a small perf hit
* we can start composing mef _earlier_ than after initialization is done.

Therzok:
Running 2 CI builds, one with changes in tests too, one without, with the v2 fix.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
```diff
@@ -56,23 +56,12 @@ public partial class CompositionManager: Service
...
-			=> throw new InvalidOperationException ("MEF queried while it was still being built");
-
-		internal static void ConfigureUninitializedMefHandling (bool throwException)
-			=> HandleMefQueriedBeforeCompletion = throwException ? new Action (UninitializedThrowException) : new Action (UninitializedLogWarning);
-
 		public static CompositionManager Instance {
 			get {
 				if (instance == null) {
 					var task = Runtime.GetService<CompositionManager> ();
-					if (!task.IsCompleted && Runtime.IsMainThread) {
```
slluis: Why not check IsMainThread anymore?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
```diff
@@ -56,23 +56,12 @@ public partial class CompositionManager: Service
...
-						HandleMefQueriedBeforeCompletion ();
+					if (!task.IsCompleted) {
+						LoggingService.LogWarning ("UI thread queried MEF while it was still being built:{0}{1}", Environment.NewLine, Environment.StackTrace);
 					}
```
mrward: The check for IsMainThread is gone now so the message looks wrong to me. Do we need the `Runtime.IsMainThread` part added back or the message changed?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
Therzok: Done in OnInitialize

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
Therzok: Check OnInitialize

mrward: OnInitialize checks for the UI thread being used - that is all good. What is to stop some code from accessing CompositionManager.Instance from a background thread? In this case the logged message would be wrong?  Just wondering since the warning message seems to suggest that a UI thread is accessing the Instance when this might not be the case.

Therzok: Well, the log message was for the scenarios where composing was not done, and we synchronously wait for the operation to finish, causing a hang.

Runtime.GetService synchronously runs OnInitialize. We call that right after OnInitialized of Ide. So this is why this warning exists. We start composing early so we avoid the hang.

We ensure that OnInitialize is called on the UI thread via AssertMainThread there.

mrward: Yeah I just mean the message is confusing to me now since it mentions the UI thread even though this may not be the case. Would it make sense to change the message to not mention the UI thread? We are only interested in something accessing it before it is initialized. Just removng 'UI ' would be enough or `MEF queried while it was still being built`.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Composition/CompositionManager.cs
Therzok: Yes, this is still needed, I'm an idiot

Therzok:
Something's wrong with the test runner it seems. If an exception is thrown, it hangs.

slluis:

PR #9394, opened 2019-11-26
[AspNetCore] Add support for running Blazor WASM projects

rodrmoya:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/965966


main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -141,6 +151,29 @@ private ExecutionCommand CreateAspNetCoreExecutionCommand (ConfigurationSelector
...
 
+		private ExecutionCommand CreateBlazorWasmExecutionCommand (ConfigurationSelector configSel, DotNetProjectConfiguration configuration, AspNetCoreRunConfiguration aspnetCoreRunConfiguration, FilePath outputFileName, string applicationUrl)
+		{
+			var blazorDevServerDll = Project.MSBuildProject.EvaluatedProperties.GetValue (BlazorDevServerDllProperty).Replace ('\\', '/');
+			if (File.Exists (blazorDevServerDll)) {
+				return new AspNetCoreExecutionCommand (
+					string.IsNullOrWhiteSpace (aspnetCoreRunConfiguration.StartWorkingDirectory) ? Project.BaseDirectory : aspnetCoreRunConfiguration.StartWorkingDirectory,
+					blazorDevServerDll,
+					$"serve --applicationpath {outputFileName}"
+				) {
```
mrward: Do you need to double quote the outputFileName path here? Might not work if the path has a space in it.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -141,6 +151,29 @@ private ExecutionCommand CreateAspNetCoreExecutionCommand (ConfigurationSelector
...
+					ExternalConsole = aspnetCoreRunConfiguration.ExternalConsole,
+					LaunchBrowser = aspnetCoreRunConfiguration.CurrentProfile.LaunchBrowser ?? false,
+					LaunchURL = aspnetCoreRunConfiguration.CurrentProfile.LaunchUrl,
+					ApplicationURL = applicationUrl.GetFirstApplicationUrl (),
+					ApplicationURLs = applicationUrl,
+					PipeTransport = aspnetCoreRunConfiguration.PipeTransport
+				};
+			}
+
+			return null;
```
mrward: Should we handle a missing blazor server dll in a better way? Would be nice to report this to the user somehow in the UI.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
rodrmoya: yeah, fixed

nosami:
Might want to look at ProcessArgumentBuilder class for escaping command line arguments. `\"  \"` might be enough though.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
mrward: Removing the File.Exists should be OK for now. As you already investigated an error is shown in the application output if blazor server dll is not found so that is good enough here for now.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -95,9 +97,16 @@ protected override bool SupportsObject (WorkspaceObject item)
 			return DotNetCoreSupportsObject (item) && IsWebProject ((DotNetProject)item);
 		}
 
+		bool IsBlazorWasm (DotNetProject project)
```
Therzok: Can be static

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -106,35 +108,38 @@ protected override bool IsSupportedFramework (TargetFrameworkMoniker framework)
...
-			FilePath outputFileName;
-			if (!(runConfiguration is AspNetCoreRunConfiguration aspnetCoreRunConfiguration))
-				return null;
-			if (aspnetCoreRunConfiguration.StartAction == AssemblyRunConfiguration.StartActions.Program)
-				outputFileName = aspnetCoreRunConfiguration.StartProgram;
-			else
-				outputFileName = GetOutputFileName (configuration);
-
-			var applicationUrl = aspnetCoreRunConfiguration.CurrentProfile.TryGetApplicationUrl ();
+			var runArguments = Project.MSBuildProject.EvaluatedProperties.GetValue (RunArgumentsProperty).Replace ('\\', '/');
```
mrward: This needs to be configuration specific. This will only work for the default output which I guess is Debug.

mrward: So you would need to do something like:

1. Get the active config from the Ide.Workspace.ActiveConfiguration
2. Get the project config from the project using `var conf = p.GetConfiguration (configuration) as DotNetProjectConfiguration;`
3. Then get the RunArguments from that DotNetProjectConfiguration.Properties.

Also may need to take into account multi-target frameworks here. It may just work if the active config has that framework but needs testing.

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
rodrmoya: No, because it was using a method from the base class, but this is gone now

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
rodrmoya: hmm, did 2 and 3 and seems to work, not sure why #1 is needed? Just calling `GetConfiguration ()` with the `ConfigurationSelector` that is passed to that method should be enough, right?

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -151,11 +153,15 @@ DotNetCoreExecutionCommand CreateDotNetCoreExecutionCommand (ConfigurationSelect
 				outputFileName = GetOutputFileName (configuration);
 
 			var workingDirectory =  Project.GetOutputFileName (configSel).ParentDirectory;
+			var currentConfig = Project.GetConfiguration (configSel) as ProjectConfiguration;
+			var runArguments = currentConfig?.Properties?.GetValue (RunArgumentsProperty)?.Replace ('\\', '/');
```
mrward: We could have an extension method here so this code is not duplicated in two places. So we could have code say like:

`var runArguments =. Project.GetRunArguments (configSel);`

Then everything is in that DotNetProjectExtension.GetRunArguments extension method.

rodrmoya: https://github.com/mono/monodevelop/pull/9394/commits/6e76e67b755be69e374279cd20c0699147616cc2

PR #9398, opened 2019-11-27
[Debugger] Exception caught dialog accessibility

mrward:
- Fixes VSTS #782188 - Voiceover doesn't give any information about the "eye" and "Pen" icon which are interactive
- Fixes VSTS #1025889 - VoiceOver does not announce the image name.
- Fixes VSTS #1025898 - ScreenReader focus does not go on the table.

Problems:

- Tabbing into and out of the Properties tree view is currently broken. Seems to be a problem with the interaction with the GtkNSViewHost and the native tree view control.

Small changes to the UI:
 - Replaced the Properties tree view the new native one created by Jeff.
 - Removed the expanders since they were causing problems with Voice Over.

UI Before:

ExceptionCaughtDialogBefore

UI After:

ExceptionCaughtDialogAfter

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ExceptionCaughtDialog.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ExceptionCaughtDialog.cs
@@ -30,6 +30,7 @@ using System.IO;
 using System.Linq;
 using System.Collections.Generic;
 
+using Foundation;
 using Gtk;
 
 using Mono.Debugging.Client;
@@ -56,6 +57,7 @@ namespace MonoDevelop.Debugger
 		VBox vboxAroundInnerExceptionMessage, rightVBox, container;
 		Button close, helpLinkButton, innerExceptionHelpLinkButton;
 		TreeView exceptionValueTreeView, stackTraceTreeView;
+		MacObjectValueTreeView macExceptionValueTreeView;
 		Expander expanderProperties, expanderStacktrace;
 		InnerExceptionsTree innerExceptionsTreeView;
 		ObjectValueTreeViewController controller;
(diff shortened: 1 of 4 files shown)


main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/ExceptionCaughtDialog.cs
```diff
@@ -199,77 +209,70 @@ Widget CreateExceptionValueTreeView ()
...
+					"transform", NSNull.Null,
+					"bounds", NSNull.Null);
+
+				var host = new GtkNSViewHost (scrolled);
+				host.ShowAll ();
+				scrolledWidget = host;
+			} else {
+				exceptionValueTreeView.ModifyBase (StateType.Normal, Styles.ExceptionCaughtDialog.ValueTreeBackgroundColor.ToGdkColor ());
+				exceptionValueTreeView.ModifyBase (StateType.Active, Styles.ObjectValueTreeActiveBackgroundColor.ToGdkColor ());
+				exceptionValueTreeView.ModifyFont (Pango.FontDescription.FromString (Platform.IsWindows ? "9" : "11"));
```
netonjm: constants in magic numbers

mrward: Yeah this is the original code. I did not change it just indented it 😄

mrward:

sgmunn:
@mrward I could not get VO to announce the name of the "eye" button. it seemed to just say "button"

mrward:
Yeah it just says button. With the old tree view it does not recognise the buttons only the text.

sgmunn:
I thought we named them, or are supposed to name them

mrward:
Probably. Was only looking at fixing the Voice Over problems specific to the Exception Caught dialog in this pull request.

Fixes in the tree view will fixes problems in that dialog and in the Watch/Locals pad. Not looked at  accessibility in the tree view itself.

sgmunn:
fair enough. fwiw, I think it's just missing `AccessibilityTitle` (which we have for one of them).

mrward:
I see the same behaviour in the locals/watch pad as in this pull request. Looking at the code it does not look like accessibility labels have been set for the preview button and eye button, if I am looking in the right place.

sgmunn:
yes - I was specifically looking because `Fixes VSTS #1025889 - VoiceOver does not announce the image name`

mrward:
Will take a look at fixing them here. Not seen any other bugs filed for the Locals/Watch pad.

That image name problem is fixed here. It is the lighting icon at the top of the dialog.

sgmunn:
or, is that the image that shows what kind of property the object is? I wonder if we should be updating the a11y title to match the property type or ignoring that image altogether?

sgmunn:
oh, _that_ image

mrward:
The images and buttons now have an accessibility title.

mrward:

PR #9401, opened 2019-11-28
[753577][StatusBar] announce changed status

mykyta-bondarenko:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1019006

### DIFF
--- a/main/src/addins/MacPlatform/MainToolbar/StatusBar.cs
+++ b/main/src/addins/MacPlatform/MainToolbar/StatusBar.cs
@@ -43,6 +43,7 @@ using System.Threading;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using MonoDevelop.MacInterop;
+using MonoDevelop.Components.AtkCocoaHelper;
 
 namespace MonoDevelop.MacIntegration.MainToolbar
 {
@@ -927,8 +928,21 @@ namespace MonoDevelop.MacIntegration.MainToolbar
 
 			bool changed = LoadText (message, isMarkup, statusType);
 			LoadPixbuf (image);
-			if (changed)
+			if (changed) {
 				ReconstructString ();
+				// announce new status if vo/a11y is enabled
+				if (MonoDevelop.Ide.Desktop.PlatformService.AccessibilityInUse) {
+					MakeAccessibilityAnnouncement (text);
+				}
+			}
+		}
+
+		static void MakeAccessibilityAnnouncement (string text)
+		{
+			using var message = new NSString (text);
+			using var dictionary = new NSDictionary (NSAccessibilityNotificationUserInfoKeys.AnnouncementKey, message,
+													 NSAccessibilityNotificationUserInfoKeys.PriorityKey, NSAccessibilityPriorityLevel.High);
+			NSAccessibility.PostNotification (NSApplication.SharedApplication.AccessibilityMainWindow, NSAccessibilityNotifications.AnnouncementRequestedNotification, dictionary);
 		}
 
 		bool LoadText (string message, bool isMarkup, MessageType statusType)


main/src/addins/MacPlatform/MainToolbar/StatusBar.cs
```diff
@@ -927,10 +927,22 @@ public void ShowMessage (IconId image, string message, bool isMarkup, MessageTyp
...
+				}
+			}
 		}
 
+		void AnnounceText ()
+		{
+			var dictionary = new NSDictionary (NSAccessibilityNotificationUserInfoKeys.AnnouncementKey, new NSString (text),
+											   NSAccessibilityNotificationUserInfoKeys.PriorityKey, NSAccessibilityPriorityLevel.High);
+			NSAccessibility.PostNotification (NSApplication.SharedApplication.AccessibilityMainWindow, NSAccessibilityNotifications.AnnouncementRequestedNotification, dictionary);
+		}
```
mrward: There is an existing announcment api. Can `IdeApp.Workbench.RootWindow.Accessible.MakeAccessibilityAnnouncement (msg)` be used instead of this code? It seems to be using the same code as is defined here.

mykyta-bondarenko: @mrward I've just tried it - it doesn't work. `GetNSAccessibilityElement` return `null` there

sevoku: yep that won't work for a Gtk Window.

sevoku:
This works really nice, not ideal since VO interrupts itself if the status changes quickly, but that's ok, sighted users will also miss the previous status in such a case. First of all this is a huge usability improvement, without this VO users don't get status updates at all right, so there is no way to know whether an operation has finished.

main/src/addins/MacPlatform/MainToolbar/StatusBar.cs
mrward: Not sure I completely understand. How does `IdeApp.Workbench.RootWindow.Accessible.MakeAccessibilityAnnouncement (msg)` work elsewhere? IdeApp.Workbench.RootWindow is a gtk window or does it change somehow? Just wondering why using that would not work everywhere?

netonjm: I agree with mrward, if doesn't work we should move this code to be reusable in MakeAccessibilityAnnouncement

main/src/addins/MacPlatform/MainToolbar/StatusBar.cs
```diff
@@ -927,10 +927,22 @@ public void ShowMessage (IconId image, string message, bool isMarkup, MessageTyp
...
+			if (changed) {
 				ReconstructString ();
+				// annonce new status if vo/a11y is enabled
+				if (MonoDevelop.Ide.Desktop.PlatformService.AccessibilityInUse) {
+					AnnounceText ();
+				}
+			}
 		}
 
+		void AnnounceText ()
```
netonjm: We should move this code to be reusable in AtkCocoaHelper

mykyta-bondarenko:
@netonjm it's done :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Components.AtkCocoaHelper/AtkCocoaHelperMac.cs
```diff
@@ -608,17 +608,23 @@ public static void AddLinkedUIElement (this Atk.Object o, params Atk.Object [] l
...
+		public static void MakeAccessibilityAnnouncement (this NSObject element, string message)
+		{
+			if (element == null)
 				return;
-			var dictionary =
-				new NSDictionary (NSAccessibilityNotificationUserInfoKeys.AnnouncementKey, new NSString (message),
-								  NSAccessibilityNotificationUserInfoKeys.PriorityKey, NSAccessibilityPriorityLevel.High);
-			NSAccessibility.PostNotification (nsObject, NSAccessibilityNotifications.AnnouncementRequestedNotification, dictionary);
+			
+			var dictionary = new NSDictionary (NSAccessibilityNotificationUserInfoKeys.AnnouncementKey, new NSString (message),
```
Therzok: wrap this in a using.

Therzok: Both the dictionary and the message

mykyta-bondarenko: @Therzok but they are local variables. Wouldn't they be destroyed after invocation ended?

Therzok: @mykyta-bondarenko there's a difference between disposing locally (frees native memory immediately), and waiting for the finalizer to kick in, and then queue a UI thread action to actually dispose it - which also incurs a greater GC impact.

It is usually better to dispose locals in a scope, rather than leave them to the finalizer

Therzok: See how it works [here](https://github.com/xamarin/xamarin-macios/blob/master/src/Foundation/NSObject2.cs#L781).

mykyta-bondarenko: @Therzok, I've added it

sevoku:
@monojenkins rebase

sevoku:
> @mykyta-bondarenko please move AnnounceText method to AtkCocoaHelperMac to a reusable method with something like signature:
> 
> public static void MakeAccessibilityAnnouncement (this NSObject o, string message)
> 
> and then remove the duplicated code and reuse it in MakeAccessibilityAnnouncement (this Atk.Object o, string message)

We can not do this, AtkCocoaHelperMac is only for AtkCocoa, and we don't have any extensions for native controls, since Xamarin.Mac is not public MonoDevelop.Ide API. I agree that it might be useful to have an extension for this, however that must be done on `MonoDevelop.Components.Control` level.

@mykyta-bondarenko as discussed on call, let's revert the last change and leave the code in MacPlatform for now.

mykyta-bondarenko:
@sevoku I've returned the previous version of code back.

main/src/addins/MacPlatform/MainToolbar/StatusBar.cs
```diff
@@ -927,8 +928,20 @@ public void ShowMessage (IconId image, string message, bool isMarkup, MessageTyp
...
+				// announce new status if vo/a11y is enabled
+				if (MonoDevelop.Ide.Desktop.PlatformService.AccessibilityInUse) {
+					MakeAccessibilityAnnouncement (text);
+				}
+			}
+		}
+
+		static void MakeAccessibilityAnnouncement (string message)
+		{
+			using var dictionary = new NSDictionary (NSAccessibilityNotificationUserInfoKeys.AnnouncementKey, new NSString (message),
```
Therzok:
```suggestion
			using var nsMessage = new NSString (message);
			using var dictionary = new NSDictionary (NSAccessibilityNotificationUserInfoKeys.AnnouncementKey, nsMessage,
```

mykyta-bondarenko: done

sevoku:
@monojenkins backport release-8.4-a11y

PR #9405, opened 2019-11-29
Fixes 1026106 - in some cases, NSEvent seems to have been null

avodovnik:
This introduces a null check, and also introduces additional debug levelchecks for others, just to make sure if we do run into this while debugging, that we are aware.

Fixes https://vsmac.dev/1026106

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/ContextMenuExtensionsMac.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Components/ContextMenuExtensionsMac.cs
@@ -139,12 +139,12 @@ namespace MonoDevelop.Components
 												null, 0, 0, 0);
 
 				// the following lines are here to dianose & fix VSTS 1026106 - we were getting
-				// a SigSegv from here and it is likely caused by NSEven being null, however
+				// a SigSegv from here and it is likely caused by NSEvent being null, however
 				// it's worth leaving Debug checks in just to be on the safe side
 				if (tmp_event == null) {
 					// since this is often called outside of a try/catch loop, we'll just
 					// log an error and not throw the exception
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Components/ContextMenuExtensionsMac.cs
```diff
@@ -136,6 +137,20 @@ public static void ShowContextMenu (NSView parent, int x, int y, NSMenu menu, bo
 												0, 0,
 												parent.Window.WindowNumber,
 												null, 0, 0, 0);
+
+				// the following lines are here to dianose & fix VSTS 1026106 - we were getting
+				// a SigSegv from here and it is likely caused by NSEven being null, however
```
Therzok:
```suggestion
				// a SigSegv from here and it is likely caused by NSEvent being null, however
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Components/ContextMenuExtensionsMac.cs
```diff
@@ -136,6 +137,20 @@ public static void ShowContextMenu (NSView parent, int x, int y, NSMenu menu, bo
...
+				// it's worth leaving Debug checks in just to be on the safe side
+				if (tmp_event == null) {
+					// since this is often called outside of a try/catch loop, we'll just
+					// log an error and not throw the exception
+					LoggingService.LogInternalError (new ArgumentNullException ("tmp_event"));
```
Therzok:
```suggestion
					LoggingService.LogInternalError (new ArgumentNullException (nameof(tmp_event)));
```

sandyarmstrong:

PR #9407, opened 2019-12-02
[Ide] Consider MSBuild item conditions in Solution pad

mrward:
Creating an ASP.NET Core project when .NET Core 3.1 SDK was installed
would result in .json files being displayed twice in the Solution pad.
.NET Core 3.1 SDK defines .json files twice.

    <Content Include="**\*.json" ... Condition="'$(ExcludeConfigFilesFromBuildOutput)'!='true'" />

    <Content Include="**\*.json" ... Condition="'$(ExcludeConfigFilesFromBuildOutput)'=='true'" />

Older .NET Core SDKs did not define the Content items more than once.
The Condition was not considered when showing files in the Solution
pad.

To support conditional files the Solution pad asks the project for
its visible files. The project uses the MSBuildEvaluationContext
to evaluate the condition to see if the file is visible or not.

Note that conditions on parent ItemGroups are currently not taken into
account. Also that visible files are not updated if the active config
is changed.

Out of scope for this change (to minimize changes for 8.4):

 - Handling ItemGroup definitions.
 - Changing visible files in the Solution pad on changing the active configuration.

Fixes VSTS #1005277 Create ASP.NET Core project, open Properties
folder, there are two launchSettings.json files.

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildPropertyGroupEvaluated.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildPropertyGroupEvaluated.cs
@@ -24,6 +24,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Xml.Linq;
 using MonoDevelop.Core;
@@ -32,7 +33,7 @@ namespace MonoDevelop.Projects.MSBuild
 {
 	class MSBuildPropertyGroupEvaluated: MSBuildNode, IMSBuildPropertyGroupEvaluated, IMSBuildProjectObject
 	{
-		protected Dictionary<string,IMSBuildPropertyEvaluated> properties = new Dictionary<string, IMSBuildPropertyEvaluated> (StringComparer.OrdinalIgnoreCase);
+		protected ConcurrentDictionary<string, IMSBuildPropertyEvaluated> properties = new ConcurrentDictionary<string, IMSBuildPropertyEvaluated> (StringComparer.OrdinalIgnoreCase);
 		MSBuildEngine engine;
 
 		internal MSBuildPropertyGroupEvaluated (MSBuildProject parent)
@@ -70,7 +71,7 @@ namespace MonoDevelop.Projects.MSBuild
 
 		internal void SetProperties (Dictionary<string,IMSBuildPropertyEvaluated> properties)
 		{
-			this.properties = properties;
+			this.properties = new ConcurrentDictionary<string, IMSBuildPropertyEvaluated> (properties, StringComparer.OrdinalIgnoreCase);
 		}
 
 		public IEnumerable<IMSBuildPropertyEvaluated> GetProperties ()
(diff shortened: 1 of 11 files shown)


mrward:

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildPropertyGroupEvaluated.cs
```diff
@@ -280,13 +301,19 @@ void IPropertyGroupListener.PropertyRemoved (MSBuildProperty prop)
...
+					}
 				}
 			}
 		}
 
 		public IEnumerable<IMSBuildPropertyEvaluated> Properties {
-			get { return properties.Values; }
+			get {
+				lock (properties) {
+					return properties.Values.ToArray ();
```
Therzok: :(

Therzok: I guess there's no way to avoid this.

Therzok (approved):

main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildPropertyGroupEvaluated.cs
mrward: The ToArray part? I did also try ConcurrentDictionary but that broke a test. There is some code that relies on the order of the items in the dictionary which works when using Dictionary. Using the lock was a way to make sure the test did not fail.

slluis: Why do we need this collection to allow concurrency?

mrward: When testing this change I ran into an InvalidOperationException `Collection was modified` being thrown when getting the properties for a project configuration (`ProjectConfiguration.Properties.GetProperties ()`).

Adding a new file to the project saves the project and the project configuration properties are modified. Values are added and then removed (if they are to be merged up to the parent group) when OnWriteProjectHeader is called. If that save happens at the same time as another thread accesses the project configuration properties then the InvalidOperationException can occur.

slluis: So this InvalidOperationException is not really related to the changes you are doing, isn't it? If that's the case, it would be better to do them in a separate PR.

mrward: It looks to be related as far as I can see. The change made here, to look at the MSBuild condition, makes use of the project configuration properties on the UI thread and the first time I tested adding a file it triggered the InvalidOperationException resulting in no files being displayed in a folder in the Solution pad window. The comment on the second commit has the exception callstack I ran into - https://github.com/mono/monodevelop/pull/9407/commits/78345eab59e39601c5070d6c074abbc0a3c5abae

slluis: I'm concerned about doing this change in 8.4, since it might have an important impact in performance.

rodrmoya:
@monojenkins rebase

PR #9417, opened 2019-12-03
[UnitTesting] Fix VS Test discovery failing

mrward:
Downgraded VS Test from 16.4 to 16.2. The 16.4 version fails to
discover tests. Test Discovery Console window has the error:

```
Microsoft.VisualStudio.TestPlatform.ObjectModel.TestPlatformException:
Testhost process exited with error: The application to execute does
not exist: ''
```

The more recent VS Test seems to not find the testhost.dll causing
test discovery to fail.

VS Test 16.2 is used by .NET Core 3.1 SDK.

Fixes VSTS #1030325 - VS Test discovery fails
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1030325


mrward:
Tested the dmg locally and everything seems OK here. .NET Core xUnit, NUnit and MSTest can be tested and debugged.

PR #9418, opened 2019-12-03
[Debugger] Fixed string quote/unquote logic for VsCode backend

jstedfast:
Also worked around a bug in the VsCode debugger protocol library
where it gives us invalid EvaluationName strings that cannot be
evaluated (due to casting errors) by using the existing string value.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1001902
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1023874

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
+++ b/main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
@@ -224,6 +224,33 @@ namespace MonoDevelop.Debugger.VsCodeDebugProtocol
 			return string.CompareOrdinal (value, index, message, 0, message.Length) == 0;
 		}
 
+		//static bool IsCSError (string text, out int code, out string message)
+		//{
+		//	message = null;
+		//	code = -1;
+
+		//	if (!text.StartsWith ("error CS", StringComparison.Ordinal))
+		//		return false;
+
+		//	int startIndex = "error CS".Length;
+
+		//	if (startIndex + 6 >= text.Length)
+		//		return false;
+
+		//	for (int i = startIndex; i < startIndex + 4; i++) {
+		//		if (!char.IsDigit (text[i]))
+		//			return false;
+		//	}
+
+		//	if (text[startIndex + 4] != ':' || text[startIndex + 5] != ' ')
+		//		return false;
+
+		//	code = int.Parse (text.Substring (startIndex, 4), NumberStyles.None, CultureInfo.InvariantCulture);
+		//	message = text.Substring (startIndex + 6);
+
+		//	return true;
+		//}
+
 		public VSCodeObjectSource (VSCodeDebuggerSession vsCodeDebuggerSession, int variablesReference, int parentVariablesReference, string name, string type, string evalName, int frameId, string val)
 		{
 			this.vsCodeDebuggerSession = vsCodeDebuggerSession;
@@ -294,7 +321,7 @@ namespace MonoDevelop.Debugger.VsCodeDebugProtocol
 
 			public RawString (string val)
 			{
-				this.val = val.Remove (val.Length - 1).Remove (0, 1);
+				this.val = val;
 			}
 
 			public int Length {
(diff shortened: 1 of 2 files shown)


main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
```diff
@@ -224,6 +224,33 @@ static bool IsCSError (int code, string message, string value, out string newVal
...
+
+		//	if (text[startIndex + 4] != ':' || text[startIndex + 5] != ' ')
+		//		return false;
+
+		//	code = int.Parse (text.Substring (startIndex, 4), NumberStyles.None, CultureInfo.InvariantCulture);
+		//	message = text.Substring (startIndex + 6);
+
+		//	return true;
+		//}
+
```
mrward: Should we remove this commented out code?

main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
```diff
@@ -315,22 +342,110 @@ public string Substring (int index, int length)
...
 
-				return rawValue.Substring (1, rawValue.Length - 2);
+				return new RawValueString (new RawString (rawValue));
 			}
 
+			//using (var timer = vsCodeDebuggerSession.EvaluationStats.StartTimer ()) {
+			//	rawValue = vsCodeDebuggerSession.protocolClient.SendRequestSync (new EvaluateRequest (evalName) { FrameId = frameId }).Result;
+			//	timer.Success = true;
+			//}
+
```
mrward: Remove this commented out code?

jstedfast: Eventually we'll need the commented out code here once we (presumably) support raw values other than strings. I think?

main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
jstedfast: I left it in because I was thinking that the above parsing code could be handy in the future when trying to handle evaluation exceptions like we *were* getting here until I discovered that the real problem was the bug in the library and not us.

The VsCode debugger backend has a lot of missing bits it seems :-\

jstedfast:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1023874/

main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
```diff
@@ -315,22 +342,110 @@ public string Substring (int index, int length)
 			}
 		}
 
-		public object GetRawValue (ObjectPath path, EvaluationOptions options)
+		static string Quote (string text)
 		{
-			string rawValue = null;
+			var quoted = new StringBuilder (text.Length + 2);
```
Therzok: Maybe worth using [StringBuilderCache](http://source.monodevelop.com/#MonoDevelop.Core/MonoDevelop.Core/StringBuilderCache.cs,ef8ac79fdf7f94be)

main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
```diff
@@ -315,22 +342,110 @@ public string Substring (int index, int length)
...
+
+			quoted.Append ('"');
+			for (int i = 0; i < text.Length; i++) {
+				char c = text [i];
+
+				switch (c) {
+				case '\0': quoted.Append ("\\0"); break;
```
Therzok: I wonder if we should be checking `char.IsControl` and quotes or we should just escape every instance of `\` and `"`

Therzok: If this is for passing to other processes, we have [ProcessArgumentBuilder](http://source.monodevelop.com/#MonoDevelop.Core/MonoDevelop.Core.Execution/ProcessArgumentBuilder.cs,0246c0e6529b1641)

main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
```diff
@@ -315,22 +342,110 @@ public string Substring (int index, int length)
...
+			return quoted.ToString ();
+		}
+
+		static string Unquote (string text)
+		{
+			var unquoted = new char[text.Length - 2];
+			bool escaped = false;
+			int count = 0;
+
+			for (int i = 1; i < text.Length - 1; i++) {
```
Therzok: maybe:
```csharp
foreach (char c in text.AsSpan(1, text.Length - 1)) {

}
```

main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
jstedfast: ProcessArgumentBuilder doesn't do what we need it to do. What we need here is to quote strings the way you'd quote a string for use in C# code.

i.e.

```
This is
a
multi-line string with
    tabs and bells
```

=>

```csharp
var text = "This is\r\na\r\nmulti-line string with\r\n\ttabs and bells\a";
```

main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
jstedfast: meh

main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
jstedfast: Maybe, but these strings can be any length (and quite long). StringBuilderCache looks like it's better suited for smaller strings. That said, I didn't know about StringBuilderCache... will have to take advantage of that more in the future...

StringBuilderCache could probably be improved by providing an API that took an expected capacity so that it could try and find a suitable StringBuilder with said capacity already rather than just returning the first one in the list.

main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
Therzok: Ohhhh, I get it. Makes sense, thanks for clarifying!

main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VsCodeObjectSource.cs
Therzok: I'm still waiting on [ValueStringBuilder](https://github.com/dotnet/corefx/issues/28379) to be productized

sgmunn:
this has a repro from a user, we should turn that into a test case and link it here

jstedfast:
Test Case: https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1047490

jstedfast:
ping

jstedfast:
This has been obsoleted by https://github.com/mono/monodevelop/pull/9513

PR #9421, opened 2019-12-04
Fixes VSTS Bug 1027417: [FATAL] SigTerm signal in

mkrueger:
MonoDevelop.VersionControl.Git.dll!MonoDevelop.VersionControl.Git.TaskFailureExtensions::RunWaitAndCapture+0

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1027417

Was the only command call left using the non async update which may
cause dead locks.

PR #9425, opened 2019-12-04
Fixes VSTS Bug 1027415: [FATAL] SigTerm signal in MonoDevelop.VersionControl.Git.dll!MonoDevelop.VersionControl.Git.GitRepository::Dispose+49

mkrueger:
MonoDevelop.VersionControl.Git.dll!MonoDevelop.VersionControl.Git.TaskFailureExtensions::RunWaitAndCapture+0

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1027415

Hard to reproduce the issue (only 4 Hits). The GitConfigurationDialog
does some long running operations where the cancellation token isn't
passed to these. That got fixed.

But the main issue is the Dispose that could dead lock on the
GitRepository. Since destroy is done on the dedicated operation thread
it may dead lock in some situations.
That can only happen if something long running blocks the
DedicatedOperationFactory. Now the DedicatedOperationFactory
operations get linked to the disposeToken which should cancel these
operations.
But all actions need to get the combined token and use that for their
cancellations.

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
@@ -35,6 +35,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using LibGit2Sharp;
+using Microsoft.TemplateEngine.Core.Util;
 using MonoDevelop.Core;
 using MonoDevelop.Core.Text;
 using MonoDevelop.Ide;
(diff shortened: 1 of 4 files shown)


sevoku:
@mkrueger some tests are still failing and this needs a manual rebase. Also it seems that 1027417 has been fixed with https://github.com/mono/monodevelop/pull/9421, did you mean to fix 1027415 here?

sevoku:
@monojenkins rebase

sevoku:

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
```diff
@@ -284,46 +285,47 @@ void ThawEvents ()
 		{
 			FileService.ThawEvents ();
 		}
-
+		object disposeLock = new object ();
```
Therzok:
```suggestion
		readonly object disposeLock = new object ();
```

main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs
```diff
@@ -122,9 +123,9 @@ public GitRepository (VersionControlSystem vcs, FilePath path, string url) : thi
 
 		Task<LibGit2Sharp.Repository> CreateSafeRepositoryAsync (string path)
 		{
-			if (Thread.CurrentThread == GitScheduler.DedicatedThread)
+			//if (Thread.CurrentThread == GitScheduler.DedicatedThread)
```
Therzok: Commented code here, if we don't need it, might as well delete it.

PR #9429, opened 2019-12-04
[Debugger] Move the Download-via-SourceLink prompt into the NoSourceV…

jstedfast:
…iew page

Instead of annoying the user with a dialog prompt and then,
when the user clicks Cancel, showing them the NoSourceView
document - just add a "Download via SourceLink" button to
the NoSourceView document tab allowing the user to download
the source there.

This simplifies the process and makes it much less annoying.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/966732/

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/NoSourceView.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/NoSourceView.cs
@@ -23,49 +23,79 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
+
 using System;
 using System.IO;
-using MonoDevelop.Components;
-using MonoDevelop.Core;
-using MonoDevelop.Ide;
-using MonoDevelop.Ide.Gui;
-using Xwt;
-using MonoDevelop.Ide.Gui.Documents;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+
+using Xwt;
+
+using Mono.Debugging.Client;
+
+using MonoDevelop.Ide;
+using MonoDevelop.Core;
+using MonoDevelop.Ide.Gui;
+using MonoDevelop.Core.Web;
+using MonoDevelop.Components;
+using MonoDevelop.Ide.Gui.Documents;
 
 namespace MonoDevelop.Debugger
 {
 	public class NoSourceView : DocumentController
 	{
-		XwtControl xwtControl;
-		ScrollView scrollView = new ScrollView ();
+		readonly ScrollView scrollView = new ScrollView ();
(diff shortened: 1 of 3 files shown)


jstedfast:
download-via-sourcelink

nosami:
"Step into External Code" is a feature that needs to be enabled manually first so not sure why it would be annoying?

And now, there is no easy to find option to automatically download? I think if you wanted to the UI to work this way then there should be a 3rd button to set the "always download" property.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/NoSourceView.cs
```diff
@@ -23,49 +23,71 @@
...
+
 				var actionsBox = new HBox ();
+
 				var buttonBrowseAndFind = new Button (GettextCatalog.GetString ("Browse and find {0}", fileName));
 				buttonBrowseAndFind.Clicked += OpenFindSourceFileDialog;
 				actionsBox.PackStart (buttonBrowseAndFind);
+
+				if (frame.SourceLocation?.SourceLink != null && options.AutomaticSourceLinkDownload == AutomaticSourceDownload.Ask) {
+					var buttonDownload = new Button (GettextCatalog.GetString ("Download via Source Link"));
+					buttonDownload.Clicked += DownloadSourceLink;
```
Therzok: Handler should be unsubscribed.

jstedfast:
@vancura can you review this UI change and make suggestions?

nosami:
> @vancura can you review this UI change and make suggestions?

@hbons probably should take a look too seeing as he was the person involved initially with the design.

sgmunn:
@vancura the original bug suggested to add an option to automatically do this, which was on the original dialog. should we add that here for completeness?

vancura:
@sgmunn Yes please

jstedfast:
Updated screenshot:

sourcelink-checkbox

vancura:
LGTM. @hbons?

jstedfast:
The Leak Tests failure seems to be unrelated (failure to install graphviz)

hbons:
@jstedfast I like your idea. I just made a few tweaks to the presentation:
image
image

thoughts?

jstedfast:
@hbons I think I like your design better than the existing UI.

jstedfast:
Updated UI based on @hbons comment:

sourcelink

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/NoSourceView.cs
```diff
@@ -23,91 +23,173 @@
...
+						MarginBottom = 20,
+						MarginTop = 10,
+						Spacing = 10,
+					};
+
+					var label = new Label {
+						Markup = GettextCatalog.GetString ("External source code is available. Would you like to download {0} and view it?", $"<a href=\"clicked\">{fileName}</a>"),
+						Name = "SourceLinkLabel"
+					};
+					label.LinkClicked += DownloadSourceLink;
```
Therzok: Event handlers need to be unsubscribed.

PR #9430, opened 2019-12-04
[MacPlatform] Add JumpList MenuItems

CartBlanche:


main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -372,6 +373,25 @@ void AppDelegate_ShowDockMenu (object sender, ShowDockMenuArgs e)
...
+
+			menu.AddItem (NSMenuItem.SeparatorItem);
+
+			// Jump List
+			var projects = IdeServices.DesktopService.RecentFiles.GetProjects ();
+			int count = 0;
+			foreach (var item in projects) {
+				var menuItem = new NSMenuItem (item.DisplayName);
+				menuItem.Activated += (o, e) => {
+					var actionUrl = "project://" + item.FileName;
```
slluis: You need to do a local capture of `item`

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -372,6 +373,25 @@ void AppDelegate_ShowDockMenu (object sender, ShowDockMenuArgs e)
 			newInstanceMenuItem.Title = GettextCatalog.GetString ("New Instance");
 			newInstanceMenuItem.Activated += NewInstanceMenuItem_Activated;
 			menu.AddItem (newInstanceMenuItem);
+
+			menu.AddItem (NSMenuItem.SeparatorItem);
+
+			// Jump List
+			var projects = IdeServices.DesktopService.RecentFiles.GetProjects ();
+			int count = 0;
+			foreach (var item in projects) {
```
slluis: Better use `foreach (var item in projects.Take(9))` instead of a counter.

CartBlanche: Thanks! didn't realise this was added way back in .NET 3.5! Fixed now.

main/src/addins/MacPlatform/MacPlatform.cs
CartBlanche: Done

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -372,6 +372,21 @@ void AppDelegate_ShowDockMenu (object sender, ShowDockMenuArgs e)
 			newInstanceMenuItem.Title = GettextCatalog.GetString ("New Instance");
 			newInstanceMenuItem.Activated += NewInstanceMenuItem_Activated;
 			menu.AddItem (newInstanceMenuItem);
+
+			menu.AddItem (NSMenuItem.SeparatorItem);
```
Therzok: We shouldn't add a separator item in case we don't have any projects.

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -372,6 +372,21 @@ void AppDelegate_ShowDockMenu (object sender, ShowDockMenuArgs e)
...
+
+			// Jump List
+			var projects = IdeServices.DesktopService.RecentFiles.GetProjects ().Take (9);
+			foreach (var item in projects) {
```
Therzok: There is a built-in way to handle this:
https://mattrajca.com/2011/07/23/appkit-quick-tip-populating-the-open-recent-menu-in-non-document-based-applications.html

Therzok: But not sure we want to pursue that route just yet.

main/src/addins/MacPlatform/MacPlatform.cs
```diff
@@ -372,6 +372,21 @@ void AppDelegate_ShowDockMenu (object sender, ShowDockMenuArgs e)
 			newInstanceMenuItem.Title = GettextCatalog.GetString ("New Instance");
 			newInstanceMenuItem.Activated += NewInstanceMenuItem_Activated;
 			menu.AddItem (newInstanceMenuItem);
+
+			menu.AddItem (NSMenuItem.SeparatorItem);
+
+			// Jump List
+			var projects = IdeServices.DesktopService.RecentFiles.GetProjects ().Take (9);
```
Therzok: There's 2 things that come to mind here:
1. This won't show unless there already is an instance of VSMac open
2. Isn't 9 a bit too many items? By default, the macOS dock menus have 3 items. Maybe this should all be packed in a submenu or we should use the native handling - otherwise there's no way to clear this list of recent items.

main/src/addins/MacPlatform/MacPlatform.cs
CartBlanche: Good point. I'll check if we have any 1st.

main/src/addins/MacPlatform/MacPlatform.cs
CartBlanche: 1. Yes, by design. It is an alternative way to open a "recent" project. With less mouse travel in a lost of cases. 
2. Xcode lists the last 10 most recently opened projects. VS4M has over ten most recent items both in it's `File->Recent Solutions` list and the initial Launch screen (up to 25!). I think 10, makes sense from the Jump List. As to "clearing", that is already taken care of at the bottom of the `File->Recent Solutions` list, where there's a menu item to do just that.

I want to also implement `NSDockTilePlugIn` so that it would be possible to launch a recent project directly from the Icon, without VS4M running. I'm investigating that in my spare time.

main/src/addins/MacPlatform/MacPlatform.cs
CartBlanche: I'm simply re-using what the VS4M team were already using. I didn't want to rewrite/refactor the internals, for now.

PR #9439, opened 2019-12-05
[Debugger] Removed DebuggerSession.set_Breakpoints to reduce complexity

jstedfast:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1026591/


mrward:
Are we OK with having an API break here?

slluis:
I prefer https://github.com/mono/debugger-libs/pull/267, it is less intrusive and it doesn't break the api.

PR #9442, opened 2019-12-06
[Core] Convert strings to paths when calling methods on the string

mrward:
An MSBuild expression such as the following will fail since the path
is not converted to a native path when not on Windows.

    $(SrcRelativeProjectDirectory.IndexOf('/'))

Arguments passed were converted but the original string was not. To
match MSBuild's behaviour the conversion of the slash characters is
only done if the directory exists.

Fixes VSTS #1028961 - MSBuild Properties are not fully evaluated causing
build to fail

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildEvaluationContext.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildEvaluationContext.cs
@@ -745,6 +745,18 @@ namespace MonoDevelop.Projects.MSBuild
 				}
 
 				// Invoke the method
+				if (instance is string instanceString &&
+					instanceString != null &&
+					instanceString.IndexOf ('\\') >= 0 &&
+					Path.DirectorySeparatorChar != '\\') {
+					// Ensure expressions such as $(SomePath.IndexOf('/')) work by using the native path separator.
+					// The directory must exist.
+					string convertedInstance = instanceString.Replace ('\\', Path.DirectorySeparatorChar);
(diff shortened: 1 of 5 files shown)


main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildEvaluationContext.cs
```diff
@@ -745,6 +745,18 @@ bool EvaluateMethod (ReadOnlySpan<char> str, MemberInfo[] member, object instanc
 				}
 
 				// Invoke the method
+				if (instance is string instanceString &&
```
Therzok: How do other paths work though? Is this an MSBuild bug?

mrward: Both dotnet's msbuild and mono's msbuild has the same behaviour here. If the directory exists it treats the path as a native path, otherwise it treats it as a string.

ConvertArgs already does this slash replacement which is how I got the idea to try the same thing for the instance object.

mrward: ConvertArg - https://github.com/mono/monodevelop/blob/3761fb76a8985c959d1f8f01b7c35535660fc156/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildEvaluationContext.cs#L978-L979

Therzok (approved):

sandyarmstrong (approved):

DavidKarlas (approved):

PR #9451, opened 2019-12-09
Improve FileNesting performance by avoiding costly IO

iantoalms:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1034938

Given the current FileNesting logic, we check for a potential parent which matches a filepath spec. We filter out these parents if they are directories using FilePath.IsDirectory, which causes IO via Directory.Exists.  As we already know whether a filepath in the project filelist is a directory, we should delay this filtering and avoid the IO.

Testing:

This change has a 2x speedup (~2.5mins to ~1.1min) for a React.js & Redux Asp.Net Core project which has a node_modules folder added to the project.


iantoalms:

PR #9456, opened 2019-12-10
[Debugger] Added telemetry for stepping and Locals/Watch/CallStack/To…

jstedfast:
…oltips

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1010671/


sgmunn:
@jstedfast what is with these tests? any time we touch this code these exact tests fail

PR #9457, opened 2019-12-11
Sync vs-editor-api and adopt code to changes

DavidKarlas:


knocte:
Shouldn't the !WINDOWS defines be changed to MAC? otherwise compilation will fail on Linux.

sandyarmstrong:
@knocte this is new editor infrastructure, which is not yet supported at all on Linux. If it was, it would use the same APIs as Mac, which sometimes differ from what is available on Windows. So these defines are correct.

PR #9460, opened 2019-12-12
[NuGet] Allow license acceptance dialog to be disabled

mrward:
Added a feature switch to allow the license acceptance dialog
to be disabled. This can be done by setting the environment
variable:

    MD_FEATURES_DISABLED=NuGetLicenseAcceptanceDialog

By default the license acceptance dialog will be displayed. The
feature switch can be used by automation tests to disable the
dialog and the license acceptance for the NuGet package will
occur without the user being prompted.

Fixes VSTS #1036006 - Disable license acceptance dialog with
feature switch


manish:
If there is more than one disabled feature, how do we do it?

mrward:
You can specify multiple features by adding a semi-colon between each feature.

    export MD_FEATURES_DISABLED=NuGetLicenseAcceptanceDialog;Another;SomeOtherCoolSecretFeature

PR #9462, opened 2019-12-12
Adding an Edit Project File item to the root project context menu.

jmatthiesen:
Fixes VSTS #617837

This work adds an Edit Project File command to the context menu shown in the Solution Pad when right-clicking on a project file.

image

Based on a discussion with @hbons, I placed this menu in the same section as "Options" since they're conceptually used for the same purposes.


mrward:
Where does the menu appear for Solutions? It seems like it would not be next to Options like it is with projects.

jmatthiesen:
Currently I left it in the Tools submenu for solutions. We hear requests for "Edit Project File" a lot and haven't heard any requests for Solutions.

mrward:
Is it in the Tools submenu? It looks like it is outside the Tools menu for a solution.

mrward:
The context menu for a solution now looks like this: 

EditFile-Solution

Edit File has moved out of the Tools submenu here.

jmatthiesen:
Good catch, thanks @mrward. I just pushed a fix to keep the Edit File menu under the Tools submenu like it was before. So now only the Edit Project File command is outside the submenu.

Therzok:
Anything else needed to merge this?

PR #9467, opened 2019-12-13
[Core] Handle odd file watcher events causing solution to be closed

mrward:
The file watcher seems to sometimes produce strange events which
look like a directory being renamed to a file. This can cause the
IDE to believe the solution file has been deleted since the file
service will raise a FileRemoved event for a rename. To avoid this
check that the directory does not exist and if it does the FileRemoved
event is not fired with this type of rename.

Fixes VSTS #1027550 - VSMac keeps closing my solution


Therzok (approved):

migueldeicaza:
So far this looks fine - have not tested as much as I would like to, but I have not triggered this problem yet.

slluis (approved):

PR #9470, opened 2019-12-13
Fix dialog parenting issue

slluis:
When a message dialog is shown, we try to get the currently focused
top level window if no parent is provided. This is done by the
GetFocusedTopLevelWindow() method that is implemented in MacPlatform.
This method should not return auto-hide pads as top level windows,
and there is a check for that, but the check doesn't work when
the auto-hide pad is implemented with Cocoa, since the pad may have
the Cocoa focus but not the GTK focus. It may happen then that an
auto-hide pad is used for parenting a message dialog, and if the
pad is auto-hidden, the modal dialog will also be hidden and
will block the whole IDE. The patch adds an additional check to
the condition for detecting if a top level is focused or not.

Fixes bug #1027840


slluis:

PR #9497, opened 2019-12-27
[AspNetCore] Support new error code for dev-certs --check

rodrmoya:
Due to the issue we found with 3.1 Preview2 on notarization, the .NET Core
team have added a new error code to detect that situation, so VSMac needs
to take it into account and run "dev-certs https --trust" as with other
known error codes, which should fix the issue.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1032644

### DIFF
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.DevCertInstaller/Program.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.DevCertInstaller/Program.cs
@@ -122,6 +122,14 @@ namespace MonoDevelop.AspNetCore.DevCertInstaller
 			}
 		}
 
+		static int RunDevCertsCommand (string dotNetCorePath, string command)
+		{
+			using (var process = Process.Start (dotNetCorePath, $"dev-certs https {command}")) {
+				process.WaitForExit ();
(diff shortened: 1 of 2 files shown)


rodrmoya:
cc @javiercn

Didn't fully test yet, need to add a clear set of steps for repro'ing and, if possible, add a new test for the issue. @javiercn is https://github.com/aspnet/AspNetCore/issues/15118#issuecomment-544909104 enough for reproducing the issue?

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore.DevCertInstaller/Program.cs
```diff
@@ -130,10 +138,10 @@ static int RunDotNetDevCerts (string dotNetCorePath)
 				return -3;
 			}
 
-			using (var process = Process.Start (dotNetCorePath, "dev-certs https --trust")) {
-				process.WaitForExit ();
-				return process.ExitCode;
-			}
+			// Clean the certificates to avoid issues with invalid ones
+			RunDevCertsCommand (dotNetCorePath, "--clean");
```
Therzok: Don't we need to handle clean's exit code?

PR #9504, opened 2020-01-03
[NuGet] Fix analyzers not used by SDK style projects

mrward:
Analyzer items were not being resolved so no analyzers were being
passed to the type system for SDK style projects. Non-SDK style
PackageReference projects were resolving Analyzer items. Projects
that use a packages.config have to explicitly use Analzyer items
in the project file.

To fix this the CoreCompileDependsOn targets is changed to include
the ResolveLockFileAnalyzers which resolves the Analyzer items.
This was already working for non-SDK style PackageReference projects
since the NuGet addin was adding the ResolveNuGetPackageAssets
target to the CoreCompileDependsOn targets and that resolves the
Analyzer items.

Fixes VSTS #1047893 - Analyzers not used by SDK style projects


Therzok (approved):

jstedfast (approved):

mrward:
The Xamarin.Mac compat check failure is a known problem with PRs currently.

PR #9505, opened 2020-01-03
[PackageManagement] Register credential providers coming from secure plugins

garuma:
This will do a barebone registration of available, plugin-based credential providers (like the Azure DevOps one) without implementing any authentication flow. The idea is that users can pre-authenticate via another mean (manual MSBuild invocation or NuGet binary) and the IDE will, at least, be able to retrieve those cached credentials.

Fixes VSTS #632121 - NuGet credential providers

### DIFF
--- a/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/PackageManagementCredentialService.cs
+++ b/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement/PackageManagementCredentialService.cs
@@ -32,6 +32,7 @@ using MonoDevelop.Core;
 using NuGet.CommandLine;
 using NuGet.Credentials;
 using NuGet.Protocol;
+using NuGet.Protocol.Plugins;
 
 namespace MonoDevelop.PackageManagement
 {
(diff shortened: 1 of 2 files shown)


mrward:
Need to check the VS Win implementation. From what I remember they put the credential provider at the end since it has bad performance compared with other implementations.

mrward:
Yeah it looks like they have it after their own VS provider which prompts - https://github.com/NuGet/NuGet.Client/blob/06903595bc47328222a882be9346660b11a7cc00/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/DefaultVSCredentialServiceProvider.cs#L73

So I think you would still see a prompt with VS Win here but not had the time to verify this.

garuma:
Yep, the reason for putting it there is to circumvent the MonoDevelop implementation (which is currently the one that would show the pesky dialog). Since the AzDO provider is the only one I can think of that would be returned today, if someone didn't go out of his/her way to set it up there should be no providers returned thus no-op'ing.

garuma:
And yeah relevant spec part is https://github.com/NuGet/Home/wiki/NuGet-cross-plat-authentication-plugin#how-does-visual-studio-work-with-the-plugin

mrward:
Yeah not sure why VS Win puts the command line credential provider last, after the one that prompts. It always seemed odd to me since you would potentially have a provider that could work but you were prompting before that for some reason.

PR #9507, opened 2020-01-04
[DesignerSupport] Support CustomTypeDescriptors in new property editor

mrward:
An object that derives from CustomTypeDescriptor would not show the
custom properties in the new native property editor but would with
the old property editor.

The problem was that the type was being passed to the TypeDescriptor's
GetProperties method instead of the object providing the properties.

Fixes VSTS #1047889 - New Property Editor does not support
CustomTypeDescriptors


mrward:

PR #9510, opened 2020-01-06
High Contrast Icons

vancura:
High contrast versions of all icons.

**Don't merge, work in progress.**

Fixes VSTS #1015990


main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor.Cocoa/MonoDevelop.TextEditor.Cocoa.csproj
```diff
@@ -108,57 +108,62 @@
     <IncludeCopyLocal Include="Microsoft.VisualStudio.Text.DifferenceViewer.Implementation.dll" />
     <IncludeCopyLocal Include="Microsoft.VisualStudio.TextMate.Core.dll" />
   </ItemGroup>
+  <None Remove="**\*.png" />
   <ItemGroup>
```
mrward: Looks like this is breaking the build. This None Remove item needs to be in an ItemGroup.

```
/Users/runner/runners/2.162.0/work/1/s/monodevelop/main/src/addins/MonoDevelop.TextEditor/MonoDevelop.TextEditor.Cocoa/MonoDevelop.TextEditor.Cocoa.csproj(111,3): error MSB4067: The element <None> beneath element <Project> is unrecognized.

```

vancura: Ah, ok. It compiled for me locally, I did exactly what Sevo told me on Slack. Let me fix that.

vancura:
Please, any idea why the build fails? Take a look: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3360370&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=852c8e5f-4ddc-570e-624a-b7cb34d65d98&l=815

From what I understand it claims the last three files are duplicated, but I can't see it in the source, all of them are there only once. Any idea?

(Locally it builds w/o any problems)

mrward:
The check for duplicates does a hash of each file and checks they are unique.

Cannot view the 3 images that are considered duplicates on github but can view the others.

```
[["Visual Studio.app/Contents/Resources/lib/monodevelop/AddIns/AspNet/Templates/images/project-aspnet-32~contrast~dark.png",
  "Visual Studio.app/Contents/Resources/lib/monodevelop/AddIns/AspNet/Templates/images/project-aspnet-32~dark~sel.png",
  "Visual Studio.app/Contents/Resources/lib/monodevelop/AddIns/AspNet/Templates/images/project-aspnet-32~sel.png"]]
```

mrward:
Yeah those three pngs are identical. Same md5 hash and `diff` reports no binary differences. Is that correct? Seems like they should be different 😄

mrward:
Looks like some of those images are corrupt maybe? They do not seem to show anything if you look at them with Finder:

aspnetimages

vancura:
They're white :)

vancura:
Just open them in your favourite viewer. You'll see :)

mrward:
Since the pngs are identical you would need to mark them as duplicates in the check-for-duplicates.rb in md-addins to make the build happy.

vancura:
Wait, so the compiler checks for the file contents, via CRC? That's crazy, we have TONS of duplicated files everywhere, it would take ages to list them in the script. Isn't there a better way?

mrward:
Currently every duplicate file is listed in check-for-duplicates.rb. Globs are not supported.

mrward:
Marius pointed out that since the images are EmbeddedResources they do not need to be shipped in the app bundle. The real problem is that they are being local copied into the .app.

I think the problem is this line in the project:

https://github.com/mono/monodevelop/blob/b25a554c5df09dd3656c1d327068a8e416e68489/main/src/addins/AspNet/MonoDevelop.AspNet.csproj#L251

I believe this is causing the AspNet/Template/images to be local copied when we do not need them to be copied. Some of the AspNet/Template/**/*.* files should be copied but the images are not needed. This does not seem to be a new problem - VS Mac 8.4 has these template images in its .app bundle.

mrward:
Probably can fix this with something like:

     <ExtensionContent Include="Templates\**\*.*" Exclude="images/**" />

Or:

    <ExtensionContent Include="Templates\**\*.*" Exclude="**\*.png" />

sevoku:
@monojenkins rebase

VladyslavKostin:
@vancura @mrward I found one issue with Step Over icon on the light theme
Screen Shot 2020-01-16 at 7 05 02 PM

sevoku:
@VladyslavKostin thanks for testing!

Wrt designer colors: I see the same on 8.4 even without high contrast mode enabled. If you scroll down, you'll see some colors. I presume that it's just "white"? @garuma ?

garuma:
Those are generated on the fly and not coming from icons so not being affected by those changes

vancura:
@VladyslavKostin Thanks for reporting the issue! Fixed in the commit above.

sevoku:

PR #9511, opened 2020-01-06
[Ide] Fix Remove item added for file that is deleted from disk

mrward:
Deleting a file from the Solution pad would sometimes add a Remove
item for that file which should only happen if the file was excluded
from the project. A way to reproduce this is to start the IDE,
create a new .NET Standard project, then right click the Class1.cs
file in the Solution pad, and select Delete. Loading an existing
solution does not seem to reproduce this.

The problem is that the file is deleted from disk after the file is
removed from the Project.Files collection. The removal of the file
causes the type system service to request updated project information.
This can then cause the project file to be updated in memory before
the file is deleted from disk. Since the file exists on disk the
Project.SaveProjectItems method adds a Remove item since it is
treating the file as been excluded. This Remove item will not be
removed when the file is deleted from disk and the project is saved
later since it is not possible to distinguish this case from the case
where a file has been explicitly excluded.

To fix this the simplest approach is to delete the file before
modifying the project's Files collection. Also fixed this problem
that can occur when deleting a folder.

Fixes VSTS #593642 - Deleting file from SDK style project adds
unnecessary data into csproj file


mrward:
@monojenkins rebase

slluis (approved):

Therzok (approved):

PR #9512, opened 2020-01-06
WIP [Core] Fix update item metadata being saved in project

mrward:
Saving a project that had MSBuild items associated with MSBuild Update
items would have the Update item metadata added to the project file
when the original MSBuild item did not have these values. The problem
was that all the metadata from all MSBuild items is stored with the
ProjectItem and then on saving all values are applied to the MSBuild
item in the project. This caused extra metadata properties to be added
when they were not needed.

Fixes VSTS #1032617 VStudio for Mac unexpectedly updates
PackageReference in csproj file


mrward:
@monojenkins rebase

PR #9514, opened 2020-01-06
[Ide] Fix intellisense not updated after adding target frameworks

mrward:
Editing an SDK project that targeted a single target framework so it
targets multiple target frameworks would result in the intellisense
not being updated from that point when the project changed, such as
when a new file was added.

The problem was that the Project.Modified event handler was removed
for the project when the old project was removed since the Roslyn
project id to project mapping had already been updated during the
reload to point to the new project. To avoid this a check is made to
ensure that no other project ids exist in the mapping before the
event handler is removed.

Fixes VSTS #1035311 - Wrong version of multi target project referenced
in .net framework project

Also fixed a null reference for new project added to existing solution. Adding a new project to an existing solution would trigger a null
reference exception in the CSharpCompilerParameters when the type
system service tried to use the project's configuration. This was then
breaking code completion for the NUnit library project when it was
added to an existing solution.

This was due to a change 18cf33d where
the OutputType was only being set if the project file was read from disk
which is not the case when adding a new project from a template to an
existing solution.

Fixes VSTS #1048443 - No intellisense for NUnit types when adding a new
NUnit project to an existing solution

### DIFF
--- a/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CSharpCompilerParameters.cs
+++ b/main/src/addins/CSharpBinding/MonoDevelop.CSharp.Project/CSharpCompilerParameters.cs
@@ -141,7 +141,7 @@ namespace MonoDevelop.CSharp.Project
 					ParentConfiguration.OutputDirectory
 			);
 
-			var outputKind = OutputTypeToOutputKind (outputType);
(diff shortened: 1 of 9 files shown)


mrward:
@monojenkins rebase

PR #9518, opened 2020-01-08
[NuGet] Register credential providers coming from secure plugins

mrward:
This will do a barebone registration of available, plugin-based credential providers (like the Azure DevOps one) without implementing any authentication flow. The idea is that users can pre-authenticate via another mean (manual MSBuild invocation or NuGet binary) and the IDE will, at least, be able to retrieve those cached credentials.

Fixes VSTS #632121 - NuGet credential providers

Still some odd behaviour here. If you open the Manage Packages dialog whilst check for updates is running it seems that you get the dialog asking for credentials. It seems that the request is being treated as a retry so the NuGet credential provider removes it from the session token cache, if that is being used.

### DIFF
--- a/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement.Gui/ManagePackagesDialogRunner.cs
+++ b/main/src/addins/MonoDevelop.PackageManagement/MonoDevelop.PackageManagement.Gui/ManagePackagesDialogRunner.cs
@@ -38,6 +38,7 @@ namespace MonoDevelop.PackageManagement
 		public void Run (IDotNetProject project = null, string initialSearch = null)
 		{
 			try {
+				PackageManagementCredentialService.Reset ();
 				bool configurePackageSources = false;
 				do {
 					using (ManagePackagesDialog dialog = CreateDialog (initialSearch, project)) {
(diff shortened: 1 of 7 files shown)


mrward:
The initial check for updates puts the credential service into retry mode. The NuGet credential provider will clear its session token cache for the url if retry is true. This behaviour is different to other providers, such as the settings provider, which return the credentials even if retry is true.

One change here when checking for updates which may make sense independent of this is to use a new credential service and make it not prompt. Restore would prompt if packages are missing but if check for updates fails then it is not a big problem and is going to be better than getting a lot of credential prompts for each enabled package source.

However this retry mode problem needs some investigation. Will check how VS Win works since they use the same credential service and it seems to be single instance.

mrward:
VS Win does not seem to get to using the command line credential providers. Its built-in vsts provider is reached first and this attempts to handle it, which makes sense here, and later providers are not used.

mrward:
Does not seem feasible to have check for updates use a different non-interactive credential service since the credential service is shared state so it would affect all NuGet actions.

mrward:
Clearing the retry cache seems to fix the credential prompt problem.

PR #9526, opened 2020-01-09
[Mac] Improve startup reason detection

iainx:
The AppDelegate always knows what the start up reason was, so check it there as
a final check if we missed the startup notification

Fixes VSTS #984249


sevoku:

sevoku:

PR #9527, opened 2020-01-09
Fixes VSTS 1005823: Adds icon for Go To Line navigation UI.

avodovnik:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1005823


avodovnik:
This is the MD side of https://github.com/xamarin/vs-editor-core/pull/460

sandyarmstrong:
We can ignore the binary compat failure, as it's nothing to do with the icon insertion.

The editor bumps will be done properly and binary compat won't be an issue then.

PR #9533, opened 2020-01-10
Fixes CommandHandler routing in native widgets

netonjm:
This PR fixes the routing of CommandHadlers into a native widgets. Basically we were not getting the active window correctly in cases when a native widget was focused then the routing was failing, affecting to all the key shortcuts combinations.

In the current example shows how Copy shorcut now works on the native cocoa Watch panel

ptsif2fujl-1

image-1

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/999606/


netonjm:
@monojenkins rebase

jstedfast:
The failure is the BinaryCompatibility.txt thing which doesn't appear to be related to this PR (it's happening for other PR's as well)

netonjm:
rebased!

PR #9537, opened 2020-01-10
[AspNetCore] Don't create browser-based execution targets for Worker services

rodrmoya:
With the enabling of launchSettings.json support for Worker services,
we added support, by mistake, to run it on the user's installed
browsers, which is wrong, as Worker services are just console
applications.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1050466

### DIFF
--- a/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
@@ -178,26 +178,30 @@ namespace MonoDevelop.AspNetCore
 
 		protected override IEnumerable<ExecutionTarget> OnGetExecutionTargets (OperationContext ctx, ConfigurationSelector configuration, SolutionItemRunConfiguration runConfig)
 		{
-			var result = new ExecutionTargetGroup (GettextCatalog.GetString ("Browser"), "MonoDevelop.AspNetCore.BrowserExecutionTargets");
-			foreach (var browser in IdeServices.DesktopService.GetApplications ("https://localhost", Ide.Desktop.DesktopApplicationRole.Viewer)) {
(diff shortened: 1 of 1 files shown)


rodrmoya:
@monojenkins backport backport-pr-9498-to-release-8.4

main/src/addins/MonoDevelop.AspNetCore/MonoDevelop.AspNetCore/AspNetCoreProjectExtension.cs
```diff
@@ -178,26 +178,30 @@ private ExecutionCommand CreateAspNetCoreExecutionCommand (ConfigurationSelector
 
 		protected override IEnumerable<ExecutionTarget> OnGetExecutionTargets (OperationContext ctx, ConfigurationSelector configuration, SolutionItemRunConfiguration runConfig)
 		{
-			var result = new ExecutionTargetGroup (GettextCatalog.GetString ("Browser"), "MonoDevelop.AspNetCore.BrowserExecutionTargets");
-			foreach (var browser in IdeServices.DesktopService.GetApplications ("https://localhost", Ide.Desktop.DesktopApplicationRole.Viewer)) {
-				if (browser.IsDefault) {
-					if (Project.HasMultipleTargetFrameworks) {
-						result.InsertRange (0, GetMultipleTargetFrameworkExecutionTargets (browser));
+			if (IsWeb) {
```
mrward: Just wondering why this is done here. Does it make sense to not use the AspNetCoreProjectExtension if IsWeb is false? If so then we could do this check in SupportsObject?

Or am I missing something?

rodrmoya: Yes, see the full change on the 8.4 backport -> https://github.com/mono/monodevelop/pull/9499

rodrmoya:
This is failing on the known compat check failure only, so ignoring

PR #9539, opened 2020-01-13
[Ide] Support adding/updating .editorconfig with Roslyn

mrward:
Right clicking in the text editor and selecting an item underneath
'Configure and Supress issues' would not add an .editorconfig file
or update the existing one.

The problem was that this feature requires changes in the type system
service.

Some odd behaviour with multiple projects in the solution, which looks like a Roslyn bug, where the editorconfig file is added as a link to a different project to the one the file being worked on belongs to.

Fixes VSTS #1001982 - Attempting to add analyzer configs fails silently
(to the user)
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1001982

Test cases linked to on above bug report.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/IMonoDevelopHostDocument.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/IMonoDevelopHostDocument.cs
@@ -32,7 +32,12 @@ namespace MonoDevelop.Ide.TypeSystem
 			textBuffer?.Properties.RemoveProperty (typeof (IMonoDevelopHostDocument));
 		}
 
-		internal static IMonoDevelopHostDocument FromDocument(Document document)
+		internal static IMonoDevelopHostDocument FromDocument (Document document)
+		{
+			return FromDocument ((TextDocument)document);
+		}
+
+		internal static IMonoDevelopHostDocument FromDocument(TextDocument document)
 		{
 			IMonoDevelopHostDocument containedDocument = null;
 			if (document.TryGetText (out SourceText sourceText)) {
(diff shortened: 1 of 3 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/IMonoDevelopHostDocument.cs
```diff
@@ -32,7 +32,12 @@ internal static void UnRegister(ITextBuffer textBuffer)
 			textBuffer?.Properties.RemoveProperty (typeof (IMonoDevelopHostDocument));
 		}
 
-		internal static IMonoDevelopHostDocument FromDocument(Document document)
+		internal static IMonoDevelopHostDocument FromDocument (Document document)
```
DavidKarlas: Since `Document` inherits `TextDocument`, can we just remove `Document` overload?

mrward: I did do that originally but I was getting errors from the text editor, from what I remember, so I added the overload.

mrward: Not the text editor but there are a few classes that want to use the Document overload. A quick search and the MonoDevelopContainedDocument class and theMonoDevelopFormattingRuleFactoryServiceFactory class still want to use Document. So did not want to change the world and added in the overload.

mrward: Also looks like that factory service implements a Roslyn interface that expects Document to be used - https://github.com/dotnet/roslyn/blob/master/src/Workspaces/Core/Portable/Formatting/Rules/IHostDependentFormattingRuleFactoryService.cs

PR #9540, opened 2020-01-13
[Ide.Tests] Retry writing to file in multi-target test

mrward:
Try to fix the sharing violation that may happen on the build server
when the MultiTargetFramework_ReloadProject_TargetFrameworksChanged
test tries to write the project file to disk directly using
File.WriteAllText.

Fixes VSTS #1050360 - MultiTargetFramework_ReloadProject_TargetFrameworksChanged
test reliability

### DIFF
--- a/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SdkProjectReader.cs
+++ b/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/SdkProjectReader.cs
@@ -65,7 +65,8 @@ namespace MonoDevelop.Projects
 				// 2) An Sdk node as a child of the Project node
 				// 3) An Sdk attribute on any Import node
 				var document = new XmlDocument ();
-				document.Load (new StreamReader (file));
+				using (var sr = new StreamReader (file));
+				document.Load (sr);
 				XmlNode projectNode = document.SelectSingleNode ("/Project");
 				if (projectNode != null) {
 					XmlAttribute sdkAttr = projectNode.Attributes ["Sdk"];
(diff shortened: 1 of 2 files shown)


radical:
Maybe add a CWL to see if we got any sharing violation exception, because when I was trying a new branch, with no changes - it wasn't failing! So, I'm not sure if it failed here at all.

mrward:
Not expecting it to fail here. The md-addins build would be the one failing since that uses Mono 6.10.

I can add some Console.WriteLines here to see if it is happening with the md-addins build.

radical:
Hm so, the file seems to get "locked" after `using (var sol = (Solution)await Services.ProjectService.ReadWorkspaceItem (Util.GetMonitor (), solFile)) {`.
Trying to follow the code paths, IIUC, it gets read here - https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildProject.cs#L232 .

I'm trying to find other places where the file might be getting opened and not disposed.

radical:
hm that seems to have fixed the test. But the question is - how does this test pass on master with 6.8.0? @mrward this does test run, as is, on master, correct?

radical:
err.. my mistake. This PR is against master!

radical:
@mrward Was this test failing only sometimes on master+mono-6.8.0?

radical:
@Therzok thanks for the review. I have opened a fresh PR for this though - https://github.com/mono/monodevelop/pull/9554 .

mrward:
Replaced with #9554

PR #9543, opened 2020-01-14
Fixes VSTS Bug 1027417: [FATAL] SigTerm signal in

mkrueger:
MonoDevelop.VersionControl.Git.dll!MonoDevelop.VersionControl.Git.TaskFailureExtensions::RunWaitAndCapture+0

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1027417

Not connected to the original bug.

### DIFF
--- a/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitService.cs
+++ b/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitService.cs
@@ -124,8 +124,14 @@ namespace MonoDevelop.VersionControl.Git
 
 		public static void ShowStashManager (GitRepository repo)
 		{
-			using (var dlg = new StashManagerDialog (repo))
-				MessageService.ShowCustomDialog (dlg);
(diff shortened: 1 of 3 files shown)


main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitService.cs
```diff
@@ -124,8 +124,14 @@ public static void ShowMergeDialog (GitRepository repo, bool rebasing)
 
 		public static void ShowStashManager (GitRepository repo)
 		{
-			using (var dlg = new StashManagerDialog (repo))
-				MessageService.ShowCustomDialog (dlg);
+			var dlg = new StashManagerDialog ();
+			Task.Run (async delegate {
```
Therzok: ❌ `void` return and `Task.Run` here. 

Callers should be responsible for synchronization context management, not callees.

mkrueger: fixed.

PR #9546, opened 2020-01-14
[Ide] Improve ImageService.CustomImageLoader

sevoku:
* sort resources by name, to ensure same sequence for all icons
* use a global static dictionary for loaded resources
* filter by requested name to avoid Xwt.Image parser scanning all resources over and over again

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ImageService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ImageService.cs
@@ -913,7 +913,7 @@ namespace MonoDevelop.Ide
 	class CustomImageLoader : Xwt.Drawing.IImageLoader
 	{
 		RuntimeAddin addin;
-		Dictionary<System.Reflection.Assembly, string []> resources = new Dictionary<System.Reflection.Assembly, string[]> ();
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ImageService.cs
```diff
@@ -913,7 +913,7 @@ static byte PixelClamp (int val)
 	class CustomImageLoader : Xwt.Drawing.IImageLoader
 	{
 		RuntimeAddin addin;
-		Dictionary<System.Reflection.Assembly, string []> resources = new Dictionary<System.Reflection.Assembly, string[]> ();
+		static Dictionary<System.Reflection.Assembly, List<string>> resources = new Dictionary<System.Reflection.Assembly, List<string>> ();
```
Therzok: Is this used only on the UI thread? Otherwise, are there any race condition concerns?

sevoku: no concerns, all calls to image loader come from ImageService which should never be called from a non-ui thread. Actually, it might even make sense to assert in all public ImageService methods. Xwt.Image creation must run on UI thread.

PR #9548, opened 2020-01-14
[IDE] Allow deleting missing files from project

ketttY:
- After a file is deleted from Finder outside VS Mac, right click on the file in the IDE will show "Remove" instead of "Delete" and a dialog will confirm user's choice to remove the file and its child from the project. If any child exists, it will be removed (not deleted)

Discussion: Should we directly delete the child?

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1051207

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
@@ -326,7 +326,7 @@ namespace MonoDevelop.Ide.Gui.Pads.ProjectPad
 				}
 
 				var result = MessageService.AskQuestion (question, new [] { AlertButton.Cancel, fileExists ? AlertButton.Delete : AlertButton.Remove });
-				if(result == AlertButton.Cancel)
+				if (result == AlertButton.Cancel)
 					return;
 				else
 					RemoveFilesFromProject (fileExists, files);
@@ -344,7 +344,7 @@ namespace MonoDevelop.Ide.Gui.Pads.ProjectPad
 				var pf = (ProjectFile)node.DataItem;
 				files.Add (pf);
 			}
-			if(!CheckAnyFileExists(files))
+			if (!CheckAnyFileExists(files))
 				info.Text = GettextCatalog.GetString ("Remove");
 		}
 
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
@@ -290,25 +290,42 @@ namespace MonoDevelop.Ide.Gui.Pads.ProjectPad
 			if (CheckAllLinkedFile (files)) {
 				RemoveFilesFromProject (false, files);
 			} else {
-				string behavior = fileExists ? "delete" : "remove";
 				if (hasChildren) {
-					if (files.Count == 1)
-						question = GettextCatalog.GetString ("Are you sure you want to {0} the file {1} and " +
-															 "its code-behind children from project {2}?",
-															 behavior, Path.GetFileName (files [0].Name), files [0].Project.Name);
-					else
-						question = GettextCatalog.GetString ("Are you sure you want to {0} the selected files and " +
-															 "their code-behind children from the project?", behavior);
+					if (files.Count == 1) {
+						if (fileExists)
+							question = GettextCatalog.GetString ("Are you sure you want to delete the file {0} and " +
+															 "its code-behind children from project {1}?",
+															 Path.GetFileName (files [0].Name), files [0].Project.Name);
+						else
+							question = GettextCatalog.GetString ("Are you sure you want to remove the file {0} and " +
+															 "its code-behind children from project {1}?",
+															 Path.GetFileName (files [0].Name), files [0].Project.Name);
+					} else {
+						if (fileExists)
+							question = GettextCatalog.GetString ("Are you sure you want to delete the selected files and " +
+															 "their code-behind children from the project?");
+						else
+							question = GettextCatalog.GetString ("Are you sure you want to remove the selected files and " +
+															 "their code-behind children from the project?");
+					}
 				} else {
-					if (files.Count == 1)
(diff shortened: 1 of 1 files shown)


ketttY:
EmptyClass is deleted from Finder, the context menu shows "Remove" and a dialog pops up for confirmation
Screen Shot 2020-01-14 at 11 44 53 AM
Screen Shot 2020-01-14 at 11 45 00 AM

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
```diff
@@ -285,35 +285,34 @@ public override void DeleteMultipleItems ()
...
-				bool filesExist = CheckAnyFileExists (files);
-
+				string behavior = fileExists ? "delete" : "remove";
 				if (hasChildren) {
 					if (files.Count == 1)
-						question = GettextCatalog.GetString ("Are you sure you want to delete the file {0} and " +
-															 "its code-behind children from project {1}?",
-															 Path.GetFileName (files [0].Name), files [0].Project.Name);
+						question = GettextCatalog.GetString ("Are you sure you want to {0} the file {1} and " +
+															 "its code-behind children from project {2}?",
```
mrward: Not sure this is going to work. I think we would need the entire string here without placeholders. The translation might be wrong if the translator just translates delete and remove without the knowledge that this is going to be inside this particular string.

Also you can use GettextCatalog.GetPluralString here - not sure if that helps. Or maybe move the code inside the if and else to another method and just have that it take a bool which indicates which string to use?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
```diff
@@ -323,18 +322,13 @@ public override void DeleteMultipleItems ()
...
+				files.Add (pf);
 			}
-		}
-
-		static AlertButton [] GetDeleteConfirmationButtons (bool includeDelete)
-		{
-			if (includeDelete)
-				return new [] { AlertButton.Cancel, AlertButton.Delete };
-			return new [] { AlertButton.Cancel };
+			if( !CheckAnyFileExists(files) )
```
mrward: Spacing here is a bit odd :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
```diff
@@ -285,35 +285,34 @@ public override void DeleteMultipleItems ()
...
 				}
-		 
-				var result = MessageService.AskQuestion (question, GetDeleteConfirmationButtons (filesExist));
-				if (result != AlertButton.Delete)
-					return;
 
-				RemoveFilesFromProject (true, files);
+				var result = MessageService.AskQuestion (question, new [] { AlertButton.Cancel, fileExists ? AlertButton.Delete : AlertButton.Remove });
+
+				if(result == AlertButton.Cancel)
```
mrward: Minor formatting here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
ketttY: Seems like moving the code into a method might require passing 5 parameters (3 bools, file name, project name), therefore I just coded the entire string :)  Do you think this will be easier to read, or will you prefer a separate method?

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
ketttY: Did you mean the empty line here?

mrward: Just meant the missing space between the if and the opening brace. `if(result == AlertButton.Cancel)`

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
```diff
@@ -323,18 +339,13 @@ public override void DeleteMultipleItems ()
...
 			}
-		}
-
-		static AlertButton [] GetDeleteConfirmationButtons (bool includeDelete)
-		{
-			if (includeDelete)
-				return new [] { AlertButton.Cancel, AlertButton.Delete };
-			return new [] { AlertButton.Cancel };
+			if(!CheckAnyFileExists(files))
+				info.Text = GettextCatalog.GetString ("Remove");
```
mrward: Looking at the original code it would set the menu item text to be Remove is ProjectFile.IsLink was true but now we check for the file existing. Does this mean that the menu item text is now Delete when a link is selected?

mrward: OK, looking at CheckAnyFileExists that returns false if the ProjectFile.IsLink is true.

mrward:
Looks like this needs a rebase since there seems to be some conflicts with ProjectFileNodeBuilder.

ketttY:
@monojenkins rebase

mrward:
The pull request is showing changes to the xwt, debugger-libs and vs-editor-api submodules which does not look right.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Pads.ProjectPad/ProjectFileNodeBuilder.cs
mrward: What you implemented looks OK to me.

mrward:
Tested this and everything looks good to me. If we can fix the conflict and remove the submodule bumps then we can approve this.

ketttY:
Test Case: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1055515

PR #9550, opened 2020-01-15
Bump vs-editor-api

DavidKarlas:


mrward:
Should this have a comment with the usual fixes message and then the bug number?

mrward:
> Should this have a comment with the usual fixes message and then the bug number?

Answer is that it is not needed. The bug fix comment is read from the submodule. Thanks @DavidKarlas for clarifying this 😄

DavidKarlas:
Replaced by https://github.com/mono/monodevelop/pull/9557 with branch that doesn't have slashes.

PR #9551, opened 2020-01-15
Fixes compatbaseline check for HttpHeaderKind change

josegallardo:
This is required for insertion PR https://github.com/xamarin/md-addins/pull/6190


mrward:
Not sure this is correct. PRs are currently failing since they are building on Mac 10.13 but the master branch is building with Catalina where the compat check is working (was updated by @dalexsoto and this change here seems to revert this change made by Alex - https://github.com/mono/monodevelop/pull/9525/files)

josegallardo:
> Not sure this is correct. PRs are currently failing since they are building on Mac 10.13 but the master branch is building with Catalina where the compat check is working (was updated by @dalexsoto and this change here seems to revert this change made by Alex - https://github.com/mono/monodevelop/pull/9525/files)

Then we shouldn't merge this change and just merge the md-addins PRs even with CompatBaseCheck failing?

Looks like that's the case, and we should just close this PR and merge https://github.com/xamarin/md-addins/pull/6190

Thanks @mrward  for the quick catch! @dalexsoto  can you please confirm?

josegallardo:
Adding the `do-not-merge` label for now until we can confirm this should be closed.

mrward:
> Then we shouldn't merge this change and just merge the md-addins PRs even with CompatBaseCheck failing?
Correct. @josegallardo Yeah I would just merge the md-addins change.

PR #9557, opened 2020-01-16
Bump vs-editor-api

DavidKarlas:


mrward:
Similar here. The vs-editor-core bump is to a PR branch not master. Is that expected?

https://github.com/microsoft/vs-editor-api/pull/21/files

PR #9564, opened 2020-01-17
[1032534] [About VS for Mac] Add localization to "Show loaded assemblies" string

Semptra:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1032534

This button should now be localized

image


Semptra:
@mrward could you please check and confirm that this is all the required changes to make this string localized?

PR #9565, opened 2020-01-17
[1019742] Disable VBox controls before project creation

Semptra:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1019742

Another possible fix could be moving `dialog.CloseDialog ()` (https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/NewProjectController.cs#L683) at the beginning of `Create` method.

Here is how this dialog looks like after clicking on "Create" button:
image

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
@@ -590,8 +590,12 @@ namespace MonoDevelop.Ide.Projects
 			if (controller.IsLastPage) {
 				try {
 					CanMoveToNextPage = false;
+					// disable all controls on this dialog to prevent users actions
+					VBox.Sensitive = false;
 					await controller.Create ();
 				} catch {
+					// if something goes wrong, we need to enable dialog contols
+					VBox.Sensitive = true;
 					throw;
 				} finally {
 					CanMoveToNextPage = true;


Semptra:
@sevoku also, this part of the code

https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs#L644-L647

 seems to be newer executed, because of this check

https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs#L590-L600

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
```diff
@@ -590,6 +590,7 @@ public async Task MoveToNextPage ()
 			if (controller.IsLastPage) {
 				try {
 					CanMoveToNextPage = false;
+					centreVBox.Children [0].Sensitive = false;
 					await controller.Create ();
```
mrward: Seems like we could make all controls non-sensitive in the dialog here - not just the first child of the centerVBox. That would include the buttons (I think there used to be a bug here where you could click the button repeatedly).

If an exception is thrown should we set Sensitive back to true?

mrward:
Please also can we have the commit message indicate why the change is being made here? Currently the message is telling me what is being changed not why :)

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
Semptra: @mrward the first child of vBox contains all buttons (as on a screenshot).
Yes, seems like a good idea to bring it back to `true` in `catch`.

mrward: If you look at the Build method centerVBox does not seem to contain all the UI widgets. 

https://github.com/mono/monodevelop/blob/ef6da1668b427ce9a3e06edc6f55fc0db79a32c3/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.UI.cs#L108

The NextButton for example here is outside it.

However the NextButton seems to be explicitly disabled in code.

mrward: Can we just do `centreVBox.Sensitive = false` instead of the first child if we do not need to do the entire dialog? :)

mrward:
In general I think we should disable (make them non-sensitive) all UI widgets on the dialog when the Create button is clicked. Currently you can click the Previous button after clicking the Create button if you are quick enough, and that moves you back a page.

Semptra:
@mrward I've added the suggested changes.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs
Semptra: Doing the `centreVBox.Sensitive = false` will not make `Cancel` and `Previous` buttons inactive. `centreVBox.Parent.Sensitive = false`seems to be the solution.

mrward: If we are doing `centreVBox.Parent.Sensitive = false` then you should be able to replace that with `VBox.Sensitive = false`. VBox seems to be the parent here.

Semptra: @mrward agree.

mrward:
> @sevoku also, this part of the code
> 
> https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs#L644-L647
> 
> seems to be newer executed, because of this check
> 
> https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects/GtkNewProjectDialogBackend.cs#L590-L600

That code is executed as far as I can see. The GetFinalPageControls method enables the UITest check box for those templates that support it - which is working. If you look at the code there is a call to MoveToNextPage after the IsLastPage check which could then cause IsLastPage to be true. Then GetWidgetToDisplay would execute that logic.

mrward:

PR #9566, opened 2020-01-17
VSTS 1049891:  aspx files no intellisense 

avodovnik:
This switches the aspx file editing to the old editor an and includes .aspx, .ashx, .asmx, .ascx, .master, .asax, ~.cshtml, .html, .htm~ extensions.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1049891


avodovnik:
I'm actually not sure we should also be including `cshtml` and `html` in this.

mrward:
Yeah I am not sure either - I added Kirill as a reviewer to check that :)

avodovnik:
I've gone and removed the cshtml, html and html files, as they worked fine (i.e. had syntax highlighting) already, so I assume they were wired as they should be. I can always revert that commit if @KirillOsenkov feels that's the better option, all things considered.

KirillOsenkov:
Yes, html and cshtml are new editor. @toddgrun

mrward:
This fixes aspx code completion. There seems to be no code completion for html but that seems to affect 8.5 not 8.4 so not relevant here. Will open a bug for html completion against VS Mac 8.5.

mrward:

mrward:

PR #9567, opened 2020-01-17
[Debugger] Tooltips should not be shown for conditional method invoca…

jstedfast:
…tions

When hovering the mouse cursor over most method invocation
expressions, the logic returns a null expression to evaluate
(thereby bypassing evaluation of said expression), but when
hovering over conditional method invocations, the logic failed.

Since the logic is identical to VS Windows, do what VS Windows does
in this scenario as well, which is to ignore evaluation errors
and not show them.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1054139

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.VSTextView/QuickInfo/DebuggerQuickInfoSource.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger.VSTextView/QuickInfo/DebuggerQuickInfoSource.cs
@@ -135,7 +135,7 @@ namespace MonoDevelop.Debugger.VSTextView.QuickInfo
 				if (cancellationToken.IsCancellationRequested)
 					return;
 
-				if (val == null || val.IsUnknown || val.IsNotSupported)
+				if (val == null || val.IsUnknown || val.IsNotSupported || val.IsError)
 					return;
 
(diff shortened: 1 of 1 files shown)


jstedfast:
Sample test app:

```csharp
using System;
using System.Threading;
using System.Threading.Tasks;

namespace DebuggerTestApp
{
	class Program
	{
		public static void Main (string[] args)
		{
			var command = new Command ();

			TestStuffAsync (command).GetAwaiter ().GetResult ();
		}

		static async Task TestStuffAsync (object command)
		{
			var canExecute = (command as Command)?.ChangeCanExecute ();
			var canExecute2 = (command as Command)?.GetCommand ().GetCommand ()?.GetCommand ().GetCommand ().ChangeCanExecute ();
			var property = (command as Command).GetCommand ().Property;
		}
	}

	class Command
	{
		public string Property {
			get { return "property"; }
		}

		public Command GetCommand ()
		{
			return this;
		}

		public bool ChangeCanExecute ()
		{
			return true;
		}
	}
}
```

jstedfast:
Test case: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1055493

mrward:
Can we put the code in the test case so it is self-contained? Could have the code in a file attachment on the test case.

PR #9568, opened 2020-01-17
[Debugger] Added more logging for VsCodeDebugger protocol requests

jstedfast:
For OnSetNextStatement(), convert exceptions into NotSupportedException
so that higher-level code properly handles this case.

For Stepping, just drop & log the exception like SoftDebuggerSession does.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1042237

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VSCodeDebuggerSession.cs
+++ b/main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VSCodeDebuggerSession.cs
@@ -66,12 +66,21 @@ namespace MonoDevelop.Debugger.VsCodeDebugProtocol
 
 		protected override void OnContinue ()
 		{
-			protocolClient.SendRequestSync (new ContinueRequest (currentThreadId));
+			try {
+				protocolClient.SendRequestSync (new ContinueRequest (currentThreadId));
+			} catch (Exception ex) {
+				if (!HandleException (ex))
+					OnDebuggerOutput (true, ex.ToString ());
+			}
 		}
 
 		protected override void OnDetach ()
 		{
-			protocolClient.SendRequestSync (new DisconnectRequest ());
+			try {
+				protocolClient.SendRequestSync (new DisconnectRequest ());
+			} catch (Exception ex) {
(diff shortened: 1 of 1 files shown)


main/src/addins/MonoDevelop.Debugger.VSCodeDebugProtocol/MonoDevelop.Debugger.VsCodeDebugProtocol/VSCodeDebuggerSession.cs
```diff
@@ -126,9 +148,15 @@ protected override void OnSetNextStatement (long threadId, string fileName, int
...
 			var source = new Source { Name = Path.GetFileName (fileName), Path = fileName };
 			var request = new GotoTargetsRequest (source, line) { Column = column };
-			var response = protocolClient.SendRequestSync (request);
+			GotoTargetsResponse response;
 			GotoTarget target = null;
 
+			try {
+				response = protocolClient.SendRequestSync (request);
+			} catch (Exception ex) {
+				throw new NotSupportedException (ex.Message);
```
mrward: Should we pass the ex as an InnerException? Just wondering if the NotSupportedException would be logged and we would miss any useful exception information from the protocol client's error.

Also does an exception here just mean it is not supported? Can we detect a real error versus something not supported?

jstedfast: The UI just uses NotSupportedException as a way to show an error to the user about "can't set the next statement to that location"

mrward: My next question is should we be logging the exception using DebuggerLoggingService.LogError as you are doing elsewhere?

jstedfast: Yea, I guess we could. I was looking thru telemetry (which is how I discovered this issue) and the only errors reported so far are all "You cannot set the next statement to another function"

jstedfast: Added

mrward:

PR #9571, opened 2020-01-20
[786228] [Telemetry] Added TargetSdkVersion to ProjectEventMetadata

Semptra:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/786228

`GetProjectSdkVersion` method could now be used to get information about the project SDK. This is required to populate this information to telemetry.
After this will be merged, some additional code needs to be added here in md-addins:
https://github.com/xamarin/md-addins/blob/27ede114756638d3a48d719cd3d2aaa52760a32c/Xamarin.Ide/Xamarin.Ide/Xamarin.Ide.Insights/ProjectTelemetryExtension.cs#L130

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
@@ -83,6 +83,18 @@ namespace MonoDevelop.DotNetCore
 			return base.SupportsObject (item) && HasSupportedFramework ((DotNetProject)item);
 		}
 
+		protected override ProjectEventMetadata OnGetProjectEventMetadata (ConfigurationSelector configurationSelector)
+		{
+			var metadata = base.OnGetProjectEventMetadata (configurationSelector);
+
+			var projectSdkVersion = sdkPaths.TargetSdkVersion ?? sdkPaths.GetLatestSdk ();
+			var projectSdkVersionString = projectSdkVersion?.ToString ();
+
+			metadata.TargetSdkVersion = projectSdkVersionString;
+
+			return metadata;
+		}
+
 		/// <summary>
 		/// Cannot check TargetFramework property since it may not be set.
 		/// Currently support .NET Core and .NET Standard.
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
@@ -40,6 +40,7 @@ namespace MonoDevelop.DotNetCore
 		static readonly object locker = new object ();
 
 		public DotNetCoreVersion [] SdkVersions { get; internal set; } = Array.Empty<DotNetCoreVersion> ();
+		public DotNetCoreVersion TargetSdkVersion { get; private set; }
 		public string SdkRootPath { get; internal set; }
 		public string GlobalJsonPath { get; set; }
 
@@ -68,8 +69,7 @@ namespace MonoDevelop.DotNetCore
 		{
 			if (!SdkVersions.Any ())
 				return;
-					
-			DotNetCoreVersion targetVersion = null;
+
 			if (forceLookUpGlobalJson) {
 				GlobalJsonPath = LookUpGlobalJson (workingDir);
 			}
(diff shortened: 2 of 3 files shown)


main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
```diff
@@ -56,6 +56,15 @@ public DotNetCoreProjectExtension ()
...
 		}
 
+		public string GetProjectSdkVersion (DotNetProject project)
+		{
+			var projectReferencedSdks = project.MSBuildProject.GetReferencedSDKs ();
+			var projectSdkPaths = DotNetCoreSdk.FindSdkPaths (projectReferencedSdks);
+			projectSdkPaths.ResolveSDK (project.ParentSolution?.BaseDirectory);
+			var projectSdkVersion = projectSdkPaths.TargetSdkVersion ?? projectSdkPaths.GetLatestSdk ();
+			return projectSdkVersion.ToString ();
+		}
```
mrward: We already have the sdkPaths stored in the DotNetCoreProjectExtension. So I do not think we need to do this resolving again.

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreSdkPaths.cs
```diff
@@ -40,6 +40,7 @@ class DotNetCoreSdkPaths
 		static readonly object locker = new object ();
 
 		public DotNetCoreVersion [] SdkVersions { get; internal set; } = Array.Empty<DotNetCoreVersion> ();
+		public DotNetCoreVersion TargetSdkVersion { get; private set; }
```
mrward: 👍

main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
Semptra: @mrward yes, but the only place where this field is being populated is `OnPrepareForEvaluation` method. In the `md-addins Xamarin.Ide.Insights` we need to create a new instance of `DotNetCoreProjectExtension` in order to use it, and since `OnPrepareForEvaluation` is not public, we cannot access it.

Semptra: Is there any other way how could this method be invoked without creating a new instance of `DotNetCoreProjectExtension`?

mrward: Still not understanding. You do not need to call OnPrepareForEvaluation. The DotNetCoreProjectExtension has the sdkPaths.

Is the problem that the telemetry information is taken before the sdkPaths is called?

mrward: My main concern here is that we are doing this resolving twice when we do not need to.

mrward: I also do not really like this returning a string. Ideally if we need this information it would be a property and would be the DotNetCoreVersion but that class is internal.

Another idea here would be to use the project's ExtendedProperties, since this is internal data for telemetry. Which avoids having to make everything public.

rodrmoya: Or another idea would be to fill in the counter we use for building (we do right?) with the SDK version and just map it in md-addins?

mrward: I think the problem with using it in the build metadata is that the telemetry event is for project used which happens on OnEndLoad currently.

That could be changed to OnItemReady but would need the entire extension chain to be called before the telemetry event is sent. The DotNetCoreProjectExtension configures sdkPaths at the end of its OnItemReady method - https://github.com/mono/monodevelop/blob/666a40983eeb6bc45d25a9bfcd071a338e28974c/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs#L306

Semptra: Currently, `Xamarin.Ide` project does not even have a reference to `MonoDevelop.DotNetCore` project. Inside this method (https://github.com/xamarin/md-addins/blob/27ede114756638d3a48d719cd3d2aaa52760a32c/Xamarin.Ide/Xamarin.Ide/Xamarin.Ide.Insights/ProjectTelemetryExtension.cs#L130) we need to create a new instance of `DotNetCoreProjectExtension` class. `sdkPaths` is not populated in ctor.

Semptra: > Another idea here would be to use the project's ExtendedProperties, since this is internal data for telemetry. Which avoids having to make everything public.

Yes, we could do this, but since `DotNetCoreVersion` class is internal, we will not be able to cast this property from object back to this type in the caller method.

mrward: You do not need to create a DotNetCoreProjectExtension class in md-addins.

You can either put this data into the project's ExtendedProperties, which is a simple fix here and requires no API changes.

Another perhaps better long term idea here might be to have a way for the project or project extension to add extra metadata properties for the project used event (but maybe out of scope for now). Similar to what we do for build events. Or maybe even use the existing OnGetProjectEventMetadata which does not indicate it is only to be used for build events.

mrward: > Yes, we could do this, but since DotNetCoreVersion class is internal, we will not be able to cast this property from object back to this type in the caller method.

You can use the same string you are returning in ExtendedProperties - it does not need to be the DotNetCoreVersion

Semptra: @mrward I updated this part of the code to use ExtendedProperties.

mrward: Yeah but this is a public method. The idea behind using ExtendedProperties was that we do not need to change the API.

I think I prefer Rodrigo's idea now of using the ProjectEventMetadata instead since that is not specific to the DotNetCoreProjectExtension.

mrward (approved):
👍

mykyta-bondarenko (approved):

Therzok (approved):

rodrmoya (approved):

nosami (approved):

PR #9576, opened 2020-01-20
Fix information popover flickering

sevoku:
When the TooltipPopoverWindow is visible, the parent
might receive mouse in/out events rapidly, depending
on the parenting status of the parent dialogs. This might
be caused by some Gtk Loop issue.

Delaying the hiding of the popup when the mouse leaves
effectively workarounds the issue.


sevoku:

PR #9577, opened 2020-01-21
Fixes VSTS Bug 1016381: Can't save or easily close a solution

mkrueger:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1016381

The AndroidDesignerController is the issue here but the base class
shouldn't crash in that scenario.
(base.Dispose() called before disposing the derived class. base.Dispose () should be called last.)

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
@@ -438,6 +438,7 @@ namespace MonoDevelop.Ide.Gui.Documents
 				disposedTokenSource.Cancel ();
 				disposedTokenSource.Dispose ();
 				disposedTokenSource = null;
+				extensionChain = null;
 				disposed = true;
 			}
 		}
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Gui.Documents/DocumentController.cs
```diff
@@ -574,8 +575,10 @@ public IEnumerable<object> GetContents (Type type)
 				yield return AssertValid (c, type);
 
 			if (extensionChain != null) {
-				foreach (var ext in extensionChain.GetAllExtensions ().OfType<DocumentControllerExtension> ()) {
-					foreach (var c in ext.GetContents (type))
+				foreach (var ext in extensionChain.GetAllExtensions ()) {
+					if (!(ext is DocumentControllerExtension documentControllerExtension))
```
mrward: The bug has the exception callstack:

```
Parameter name: source
  at System.Linq.Enumerable.OfType[TResult] (System.Collections.IEnumerable source) [0x00003] in /Users/builder/jenkins/workspace/build-package-osx-mono/2019-08/external/bockbuild/builds/mono-x64/external/corefx/src/System.Linq/src/System/Linq/Cast.cs:16 
```

The OfType cast is throwing an ArgumentNullException since presumably the GetAllExtensions is returning a null enumerable? With this new code it looks like the foreach will throw a null reference exception?

mkrueger:
ported to vsmac.

PR #9580, opened 2020-01-21
[1050440]Fix description of Razor files

GouriKumari:
Fix https://dev.azure.com/DevDiv/DevDiv/_workitems/edit/1050440/

Tasks Done:
Change description of Razor view  to "Razor View Page"

### DIFF
--- a/main/src/addins/MonoDevelop.AspNetCore/Templates/MVCViewImportsPage.xft.xml
+++ b/main/src/addins/MonoDevelop.AspNetCore/Templates/MVCViewImportsPage.xft.xml
@@ -10,7 +10,7 @@
 		<_Category>ASP.NET Core</_Category>
 		<LanguageName>C#</LanguageName>
 		<DefaultFilename>_ViewImports</DefaultFilename>
-		<_Description>Creates a MVC View Import Page.</_Description>
+		<_Description>Creates a Razor View Import Page.</_Description>
 	</TemplateConfiguration>
 
 	<Conditions>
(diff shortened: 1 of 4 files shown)


GouriKumari:
@nosami : When I tested this the changes didn't reflect. Can you check my changes?

nosami:
@GouriKumari just test the dmg from the build on this page once it's built

nosami:
Maybe this isn't the right change

`Create a MVC View page.` vs `Creates an MVC Razor View page.`

mrward:
Yeah the description is not consistent with all the other file templates.

Also I think this is the wrong file template. The bug is about ASP.NET Core projects whilst this file template if for classic ASP.NET projects.

The ASP.NET Core file template that matches Razor View is this one - https://github.com/mono/monodevelop/blob/release-8.5/main/src/addins/MonoDevelop.AspNetCore/Templates/MVCViewPage.xft.xml#L13 - however I cannot see that in the New File dialog - not sure why that is.

mrward:
Ah, I see. @rodrmoya  already changed the name of this file template here - https://github.com/mono/monodevelop/commit/8857b5200ad471c61f0975d1ceeca6eb40ec0eea for another bug.

mrward:
I think we should just ensure the file template name is consistent with the description for all the file templates changed in this commit - https://github.com/mono/monodevelop/commit/8857b5200ad471c61f0975d1ceeca6eb40ec0eea

Keeping the description so it is similar to 'Creates a Razor View' - replacing the Razor View with the name of the file template.

GouriKumari:
@mrward : It was my mistake. I was looking at the wrong code base.  In Visual Studio For Windows, Razor template description do not have "Creates a" prefix. The change I made here is what we have in Visual Studio For Windows. Is it okay to go ahead and bring in the same description for other Razor files or  should I make it " Creates a Razor View Page".

mrward:
I guess my only problem with this is that now the descriptions of the ASP.NET Core file templates are inconsistent with all the other file templates we have.

GouriKumari:
Reverted the extra changes.

PR #9581, opened 2020-01-22
[Debugger] Provide SolutionItem to DebuggerSession mapping

mrward:
The DebuggerService now has two methods that can map from a SolutionItem
to a DebuggerSession or the other way around. The ExecutionCommand now
has a SolutionItem property which is used by the DebuggerService to
create this mapping from DebuggerSession to SolutionItem.

Fixes VSTS #1012614 - Improvements to DebuggerSession for XAML Hot Reload

### DIFF
--- a/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
+++ b/main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
@@ -67,8 +67,8 @@ namespace MonoDevelop.Debugger
 		static readonly BreakpointStore breakpoints = new BreakpointStore ();
 		static readonly DebugExecutionHandlerFactory executionHandlerFactory;
 
-		static Dictionary<long, SourceLocation> nextStatementLocations = new Dictionary<long, SourceLocation> ();
-		static Dictionary<DebuggerSession, SessionManager> sessions = new Dictionary<DebuggerSession, SessionManager> ();
+		static ConcurrentDictionary<long, SourceLocation> nextStatementLocations = new ConcurrentDictionary<long, SourceLocation> ();
+		static ConcurrentDictionary<DebuggerSession, SessionManager> sessions = new ConcurrentDictionary<DebuggerSession, SessionManager> ();
 		static Backtrace currentBacktrace;
 		static SessionManager currentSession;
 		static int currentFrame;
@@ -151,7 +151,7 @@ namespace MonoDevelop.Debugger
 
 		public static DebuggerSession GetSession (IRunTarget runTarget)
 		{
-			foreach (KeyValuePair<DebuggerSession, SessionManager> item in sessions.ToArray ()) {
+			foreach (KeyValuePair<DebuggerSession, SessionManager> item in sessions) {
 				if (item.Value.RunTarget == runTarget) {
 					return item.Key;
 				}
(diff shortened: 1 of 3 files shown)


mrward:
I have some concerns about thread safety in DebuggerService. No locks are used when accessing the sessions dictionary. Some methods check the GUI thread is used but others do not and it looks like they would not always be called on the UI thread. Whilst ToArray is being used in some places I am not sure that is enough to solve the problem.

Using a ConcurrentDictionary might be a simple solution here, however the ordering of the items in a ConcurrentDictionary is not guaranteed and behaves different to the Dictionary type.

main/src/core/MonoDevelop.Core/MonoDevelop.Core.Execution/ExecutionCommand.cs
```diff
@@ -46,5 +46,11 @@ public abstract class ExecutionCommand
 		/// Execution target. For example, a specific device.
 		/// </summary>
 		public ExecutionTarget Target { get; set; }
+
+		/// <summary>
+		/// Solution item associated with this execution command. This allows the DebuggerSession to be
+		/// associated with a Solution item.
+		/// </summary>
+		public SolutionItem SolutionItem { get; set; }
```
Therzok: Should this be [IRunTarget](http://source.monodevelop.com/#MonoDevelop.Core/MonoDevelop.Projects/IRunTarget.cs,bf4f319052985c6c)?

Therzok (approved):

drasticactions (approved):
lgtm! 👍

Therzok (approved):

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
```diff
@@ -149,6 +149,26 @@ public static IExecutionHandler GetExecutionHandler ()
...
+				if (item.Value.RunTarget == runTarget) {
+					return item.Key;
+				}
+			}
+			return null;
+		}
+
+		public static IRunTarget GetRunTarget (DebuggerSession session)
+		{
+			foreach (KeyValuePair<DebuggerSession, SessionManager> item in sessions.ToArray ()) {
```
jstedfast: Couldn't this implementation just use `sessions.TryGetValue (session, out var runTarget)`?

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
```diff
@@ -149,6 +149,26 @@ public static IExecutionHandler GetExecutionHandler ()
 			return sessions.Keys.ToArray ();
 		}
 
+		public static DebuggerSession GetSession (IRunTarget runTarget)
+		{
+			foreach (KeyValuePair<DebuggerSession, SessionManager> item in sessions.ToArray ()) {
```
Therzok: This ToArray() should not be needed. Is there any thread-safe way to modify sessions? Otherwise, nothing is executing code that could potentially modify the collection on the current running thread here.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
mrward: Indeed it could. Must have just based this code on the code above. Still a bit worried about thread safety here. I am assuming TryGetValue might not be happy if another thread is adding/removing sessions at the same tim.

main/src/addins/MonoDevelop.Debugger/MonoDevelop.Debugger/DebuggingService.cs
mrward: That was my worry. There is no locking done in the DebuggerService. Some code seems to assuming that the UI thread is being used, but then you see Runtime.RunInMainThread being called directly. There is also no checks made to ensure the main UI thread is being used. So not sure what the locking strategy should be in the DebuggerService.

ConcurrentDictionary is an option but then the order of the keys will be different which would affect the CurrentSession property which uses `sessions.Keys.FirstOrDefault()`.

jstedfast (approved):

PR #9584, opened 2020-01-22
[Debugger] Fix InvalidCastException in MacObjectValueTreeView.OnCopy()

jstedfast:
Based on "Find References" on GetRawValue(), it seems that this can
sometimes return RawValueString instead of string. Make sure that
we handle this instead of assuming the value returned by GetRawValue()
is always a string.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1054038/


jstedfast:

mrward:
Compat check failure is not related to this change.

PR #9586, opened 2020-01-22
[Debugger] Update exception options to include unhandled exception

ketttY:
NETCoreExceptionUnhandled

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1051257/
Fixes VSTS #1051257

(The exceptionCaughtDialog's position is a bit weird, filing another bug to record it)


ketttY:
Test Case: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1056181

mrward:
With the second test case I see the full exception caught dialog (not the text editor popup window) since it looks like the source for the exception could not be found since it is in external code. Can we update the test case to indicate say this? Also maybe a screenshot in both test cases. Currently they both say `Verify that the exception caught dialog pops up with null reference exception` which made me think something was wrong.

PR #9590, opened 2020-01-23
[Ide] Dynamically add ~contrast~dark~sel resource mappings

sevoku:
to avoid adding duplicate EmbeddedResource entries to project files.

Background:

To support loading the correct icons in selected state with dark theme and high contrast mode, we'd need to add duplicate resources entries to all csproj file like this:

```diff
     <EmbeddedResource Include="icons\project-32~dark~sel.png" />
+    <EmbeddedResource Include="icons\project-32~dark~sel.png" >
+      <LogicalName>project-32~contrast~dark~sel.png</LogicalName>
+    </EmbeddedResource>
     <EmbeddedResource Include="icons\project-32~dark~sel%402x.png" />
+    <EmbeddedResource Include="icons\project-32~dark~sel%402x.png" >
+      <LogicalName>project-32~contrast~dark~sel@2x.png</LogicalName>
+    </EmbeddedResource>
```

this could be automated, but would be hard to maintain afterwards (it's already quite hard with all icons we have). With this change we don't need any duplicates.

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ImageService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ImageService.cs
@@ -929,12 +929,45 @@ namespace MonoDevelop.Ide
 				resourceList.Sort (); // sort resources by name
 			}
 
-			return resourceList.Where (r => r.StartsWith (baseName) && r.EndsWith (ext));
+			return GetFinalFilelist (resourceList, baseName, ext);
 		}
+
+		IEnumerable<string> GetFinalFilelist (IEnumerable<string> source, string baseName, string ext)
+		{
+			foreach (var name in source) {
+				if (name.StartsWith (baseName) && name.EndsWith (ext)) {
+					yield return name;
+
+					var tagPortion = name.Substring (baseName.Length, name.Length - baseName.Length - ext.Length);
+					// to avoid duplicate resource entries in project files, add virtual file mappings for
+					// high contrast icons in selected state
+					if (tagPortion == "~dark~sel" || tagPortion == "~dark~sel@2x") {
+						var map = name.Replace ("~dark~sel", "~contrast~dark~sel");
+						if (virtualMappings == null) {
+							virtualMappings = new Dictionary<string, string> ();
+						}
+						if (!virtualMappings.ContainsKey (map)) {
+							virtualMappings.Add (map, name);
+						}
+						yield return map;
+					} else {
+						// remove existing mapping if a resource with the same name exists.
+						// note: we know that the source is sorted
+						virtualMappings?.Remove (name);
+					}
+				}
+			}
+		}
+
+		Dictionary<string, string> virtualMappings;
 
 		public Stream LoadImage (string fileName)
 		{
-			return addin.GetResource (fileName, true);
+			// load original resource if a mapping exists
+			if (virtualMappings != null && virtualMappings.TryGetValue (fileName, out var mapsTo))
+				return addin.GetResource (mapsTo, true);
+			else
+				return addin.GetResource (fileName, true);
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ImageService.cs
```diff
@@ -926,15 +926,49 @@ public IEnumerable<string> GetAlternativeFiles (string fileName, string baseName
 
 			if (!resources.TryGetValue (r.ReferencedAssembly, out var resourceList)) {
 				resourceList = resources [r.ReferencedAssembly] = r.ReferencedAssembly.GetManifestResourceNames ().ToList ();
+				var virtualNames = new List<string> ();
```
Therzok: unused.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ImageService.cs
```diff
@@ -926,15 +926,49 @@ public IEnumerable<string> GetAlternativeFiles (string fileName, string baseName
...
+			return GetFinalFilelist (resourceList, baseName, ext);
 		}
 
+		IEnumerable<string> GetFinalFilelist (IEnumerable<string> source, string baseName, string ext)
+		{
+			foreach (var name in source) {
+				if (name.StartsWith (baseName) && name.EndsWith (ext)) {
+					yield return name;
+
+					var tagPortion = name.Substring (0, name.Length - ext.Length).Substring (baseName.Length);
```
Therzok: Maybe use ReadOnlySpan here.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ImageService.cs
```diff
@@ -926,15 +926,49 @@ public IEnumerable<string> GetAlternativeFiles (string fileName, string baseName
...
+					// to avoid duplicate resource entries in project files, add virtual file mappings for
+					// high contrast icons in selected state
+					if (tagPortion == "~dark~sel" || tagPortion == "~dark~sel@2x") {
+						var map = name.Replace ("~dark~sel", "~contrast~dark~sel");
+						if (virtualMappings == null) {
+							virtualMappings = new Dictionary<string, string> ();
+						}
+						if (!virtualMappings.ContainsKey (map)) {
```
Therzok:
```suggestion
						else if (!virtualMappings.ContainsKey (map)) {
```

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ImageService.cs
vancura: May I have a question I've been thinking about since I joined Xamarin? How's possible that you see these things? Do you static analyze the PR, build the branch, or you just see it? :)

Therzok: My brain is hard-wired to see redundancy and eliminate it :D

Therzok: Plus, it's a new declaration and nothing uses it in the diff block, so it's easy to spot.

vancura: 🤯

sevoku: @Therzok ftw!

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ImageService.cs
sevoku: Stupid me. Now changed to one Substring, which is faster than Span.

sevoku:

PR #9591, opened 2020-01-23
Update fallback feed to match live feed

nosami:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1022827


nosami:

nosami:
> Can we fix the line endings? Had to use `?w=1` to see the small diff here :)

@mrward I think the line endings are now fixed and were wrong before :) I just copy and pasted the contents from here https://webpifeed.blob.core.windows.net/webpifeed/Partners/ScaffoldingPackageVersions_2108718.json

PR #9595, opened 2020-01-24
[786228] [Telemetry] Added TargetSdkVersion to ProjectEventMetadata

Semptra:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/786228

GetProjectSdkVersion method could now be used to get information about the project SDK. This is required to populate this information to telemetry.
After this will be merged, some additional code needs to be added here in md-addins:
https://github.com/xamarin/md-addins/blob/27ede114756638d3a48d719cd3d2aaa52760a32c/Xamarin.Ide/Xamarin.Ide/Xamarin.Ide.Insights/ProjectTelemetryExtension.cs#L130

Renamed branch from https://github.com/mono/monodevelop/pull/9571

### DIFF
--- a/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
+++ b/main/src/addins/MonoDevelop.DotNetCore/MonoDevelop.DotNetCore/DotNetCoreProjectExtension.cs
@@ -83,6 +83,18 @@ namespace MonoDevelop.DotNetCore
 			return base.SupportsObject (item) && HasSupportedFramework ((DotNetProject)item);
 		}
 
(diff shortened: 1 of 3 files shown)


mrward:
Another idea would be to have some way to detect if DotNetCoreSdkPaths has checked for the global json file path and resolved the sdk with it. And if not then that is done in the GetProjectEventMetadata method when it is called. Then the OnItemReady method would also need to see if that was done instead of doing it again.

Looking at the DotNetCoreSdkPaths maybe we could just check to see if GlobalJsonPath is not null. It seems to be set to the path or an empty string when the path is resolved.

At some point we probably want to look at where this global sdk paths logic is done. Right now we do it for every .NET Core project but really it only needs to be done once for the solution. However that is out of scope for this change.

PR #9596, opened 2020-01-24
[Templates] Check IdeApp.IsInitialized before opening document

Semptra:
When running unit tests, `IdeApp.Workbench` is null which causing NRE. This should not affect calls from other places.
This affects https://github.com/xamarin/md-addins/pull/6242


Semptra:
https://github.com/xamarin/vsmac/pull/82

PR #9599, opened 2020-01-24
Build output is not cleared when loading or creating a new project. 

GideonCheruiyot:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/573061

Build output is now empty every time a new project is created.


mrward:
Can we clean up the commits so it is just a single commit? Two commits seem to be the same :)

GideonCheruiyot:
deleting this to fix multiple commits issue

PR #9600, opened 2020-01-24
Allow the startup to complete without taking control of the runloop.

iainx:
Separate the initialization/shutdown sequence into separate methods just on the offchance that a separate runloop might be used. You never know...


Therzok:
I don't have enough context to know how to review this.

main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/IdeStartup.cs
```diff
@@ -761,7 +784,9 @@ public static int Main (string[] args, IdeCustomizer customizer = null)
 						BrandingService.ApplicationName
 					), ex);
 			} finally {
-				Runtime.Shutdown ();
+				if (controlMainLoop) {
```
Therzok: Do we need this controlMainLoop code? What if we had 2 separate runloop runners (i.e. in IdeCustomizer), with different implementations and simplify all the signatures that have changed around here?

PR #9601, opened 2020-01-24
[Debugger] Set the font for tooltips and pinned watches

jstedfast:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1051838


jstedfast:
tooltip-fonts

jstedfast:
Adding @vancura - this will need UX review before merging.

I have a local patch which makes it use the same font as the editor, but fixed-width fonts in tooltips look horrible IMHO, **but** I just had a thought...

What font do the editor's tooltips use? Can we use that font?

vancura:
@sevoku See the question about the tooltip font above.

@jstedfast You're right about the monospace font in tooltips, but we should use it on all other places (i.e. dialogs) **where code is printed** (e.g. `String` visualizer).

PR #9602, opened 2020-01-24
Exclude Contents/MonoBundle from BinaryCompatChecker

KirillOsenkov:


main/build/MacOSX/BinaryCheckerConfig.txt
```diff
@@ -6,13 +7,14 @@
 !Contents/Resources/lib/monodevelop/AddIns/docker/MonoDevelop.Docker/mscorlib.dll
 !Contents/Resources/lib/monodevelop/AddIns/DotNetCore.Debugger/Adapter
 !Contents/Resources/lib/monodevelop/AddIns/MonoDevelop.AzureFunctions/azure-functions-cli
+!Contents/Resources/lib/monodevelop/AddIns/MonoDevelop.MonoDroid/Microsoft.CodeAnalysis.Workspaces.MSBuild.dll
```
Therzok: This shouldn't be there. cc @josegallardo

KirillOsenkov: yeah I was wondering if someone added this recently

Therzok: It should not be copied

https://github.com/xamarin/vsmac-xamarin-extensions/blob/master/MonoDevelop.MonoDroid/MonoDevelop.MonoDroid/MonoDevelop.MonoDroid.csproj#L224

KirillOsenkov: Good catch; I've pulled the .dmg and the file isn't even there. Removing it.

main/build/MacOSX/BinaryCheckerConfig.txt
```diff
@@ -13,6 +14,6 @@
 !Contents/Resources/lib/monodevelop/AddIns/Xamarin.Interactive.XS/Xamarin Inspector.app
 !Contents/Resources/lib/monodevelop/bin/Microsoft.VisualStudio.Imaging.dll
 !Contents/Resources/lib/monodevelop/bin/ServiceHub
-!Contents/Resources/lib/monodevelop/AddIns/MonoDevelop.MonoDroid/Microsoft.CodeAnalysis.Workspaces.MSBuild.dll
+!Contents/Resources/lib/monodevelop/bin/Xamarin.Mac.dll
```
Therzok: Does this mean we don't want to validate Xamarin.Mac compatibility? What if some addin is built against a newer Xamarin.Mac than what we ship?

KirillOsenkov: We'll re-enable after everything moves to Catalina. Right now it flip flopped depending on whether you're building on Catalina or not, so I want to remove the noise. Otherwise people lose trust in the compat checker.

From @dalexsoto:

```
I think I have tracked down the Compatibility Check issue happening in MonoDevelop master, this all goes down to [...] where we brought HttpClient implementation from CoreFX but more specificly to these lines [...], the offending types and methods are there. Now this did not show up earlier because it seems we just recently started building with mojave so we just started building with the new Xamarin.Mac 6.14.x  [...] thus not hitting the new code paths and not getting Compatibility Check issues related to this until now.
Since we were already expecting this checks in Compatibility Check I think this is the right fix https://github.com/mono/monodevelop/pull/9525, if you are ok with it we should backport this to release-8.5 once merged. (edited) 
```

PR #9604, opened 2020-01-26
[Ide] Calculate file nesting on demand

rodrmoya:
Instead of loading all files at once, just calculate file nesting information
on demand, that is, when the node builders ask for it.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1030656

### DIFF
--- a/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
+++ b/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.FileNesting/FileNestingService.cs
@@ -187,9 +187,6 @@ namespace MonoDevelop.Ide.Projects.FileNesting
 				initialized = true;
 
 				fileNestingEnabled = FileNestingService.IsEnabledForProject (Project);
-				foreach (var file in Project.Files) {
-					AddFile (file);
-				}
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


rodrmoya:
This reduces the number of times `AddFile` is called, as it was being called duplicated times for some files, but seems the node builders retrieve information about all files in the project when loading the project, is that so @mrward @Therzok ? Seems so from my debugging of this, so while this improves a little bit the performance, it seems to not be enough if the node builders ask for the file nesting information for all files at once.

mrward:
Where does this happen? `the node builders retrieve information about all files in the project when loading the project`.

rodrmoya:
> Where does this happen? `the node builders retrieve information about all files in the project when loading the project`.

well, not sure, that's why I'm asking, but I get the file nesting service called for all the files in the project at initial load, so I assume it is the node builders, which is what calls into the file nesting service.

mrward:
Pretty sure the node builders are lazy. Could it be the OnFileAddedToProject method in the file nesting service? That may be called on load for each file added to the project.

rodrmoya:
> Pretty sure the node builders are lazy. Could it be the OnFileAddedToProject method in the file nesting service? That may be called on load for each file added to the project.

those events are only called when files are added/removed *after* the project has been loaded AFAICS, but maybe I'm wrong

PR #9637, opened 2020-05-20
Please provide precompiled binaries!

somerandompiggo:
I'm pretty shit at building other people's software and it seems like most open source projects also have binaries. MonoDevelop doesnt for windows though. On my linux machine, I love it! But, I just want to use it on windows too. I know I can compile it with VS, i have tried (failed, mostly because of errors). But why isn't there a binary?


My-Source-Code-Base:
Needs MonoDevelop binary under windows 10 x64 too.

PR #9664, opened 2020-12-21
fix #9633

### LINKED ISSUE #9633: Gnome terminal now installed in different location
Hi,

I use Ubuntu 20.04 so it might be different on your system, but `gnome-terminal` is in another directory now:

Is no longer located at `/usr/lib/gnome-terminal/gnome-terminal-server` but instead at `/usr/libexec/gnome-terminal-server`

I see this is the code that runs tries the old file:

https://github.com/mono/monodevelop/blob/master/main/src/addins/GnomePlatform/GnomePlatform.cs#L176

KisaragiEffective:


knocte:
@KisaragiEffective can you move this PR to https://github.com/dotdevelop/dotdevelop/ please?

KisaragiEffective:
@knocte sure, I'll do it.

